PR #2709, opened 2021-07-07
Support for Altair

zah:
Includes implementation of sync committees and other Altair-related changes.

PR #2761, opened 2021-08-04
Initial implementation of sync committees (squashed and rebased)

zah:
Also includes:

* Dynamic transition to the Altair gossip topics
* Metadata for the Altair devnets
* Implementation of the Altair DiscV5 spec changes


scripts/launch_local_testnet.sh
```diff
@@ -38,7 +38,7 @@ LONGOPTS="help,preset:,nodes:,data-dir:,with-ganache,stop-at-epoch:,disable-htop
 NUM_NODES="10"
 DATA_DIR="local_testnet_data"
 USE_HTOP="1"
-USE_VC="1"
+USE_VC="0"     # TODO until REST Altair fixes come in
```
tersec, line 41: This does seem, indeed, like the least bad approach to proceeding.

beacon_chain/validators/validator_pool.nim
```diff
@@ -152,6 +157,65 @@ proc signAggregateAndProof*(v: AttachedValidator,
                                                   aggregate_and_proof)
       await signWithRemoteValidator(v, root)
 
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.7/specs/altair/validator.md#prepare-sync-committee-message
```
tersec, line 160: My inclination as regards all these spec ref URLs is that they're harmless to leave this way for a bit. One sequencing is to merge this, and rebase what remains (if anything) of the `altair` branch, and then do the noisy-but-empty changes in `unstable` once they're off any critical path.

beacon_chain/spec/datatypes/base.nim
```diff
@@ -947,3 +964,4 @@ template isomorphicCast*[T, U](x: U): T =
     doAssert sizeof(T) == sizeof(U)
     doAssert getSizeofSig(T()) == getSizeofSig(U())
   cast[ptr T](unsafeAddr x)[]
+
```
tersec, line 967: Empty trailing line

beacon_chain/spec/forkedbeaconstate_helpers.nim
```diff
@@ -278,3 +278,22 @@ func shortLog*(x: ForkedSignedBeaconBlock | ForkedTrustedSignedBeaconBlock): aut
...
+    cfg.GENESIS_FORK_VERSION
+  else:
+    cfg.ALTAIR_FORK_VERSION
+
+proc nextForkEpochAtSlot*(cfg: RuntimeConfig, slot: Slot): Epoch =
+  if slot.epoch < cfg.ALTAIR_FORK_EPOCH:
+    cfg.ALTAIR_FORK_EPOCH
+  else:
+    FAR_FUTURE_EPOCH
+
```
tersec, line 299: Empty trailing line

beacon_chain/rpc/rest_utils.nim
```diff
@@ -150,6 +150,11 @@ func match(data: openarray[char], charset: set[char]): int =
       return 1
   0
 
+proc `==`*(lhs, rhs: RestValidatorIndex): bool {.borrow, noSideEffect.}
```
tersec, line 153: Could comment in an explicit, greppable way that https://github.com/nim-lang/Nim/issues/14440 is why this is `proc ... {.borrow, noSideEffect.}` so that it can be easily found later once we switch to Nim 1.6.

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -440,6 +442,24 @@ when false:
   proc `==`*(x, y: ParticipationFlags) : bool {.borrow, noSideEffect.}
 
 chronicles.formatIt BeaconBlock: it.shortLog
+chronicles.formatIt SyncCommitteeIndex: uint8(it)
+
+template asInt*(x: SyncCommitteeIndex): int = int(x)
+template asUInt8*(x: SyncCommitteeIndex): uint8 = uint8(x)
+template asUInt64*(x: SyncCommitteeIndex): uint64 = uint64(x)
+
+proc `==`*(lhs, rhs: SyncCommitteeIndex): bool {.borrow, noSideEffect.}
```
tersec, line 451: See other comment about making it easy to find these https://github.com/nim-lang/Nim/issues/14440 workarounds

beacon_chain/spec/helpers.nim
```diff
@@ -195,3 +201,4 @@ func add_flag*(flags: ParticipationFlags, flag_index: int): ParticipationFlags =
 func has_flag*(flags: ParticipationFlags, flag_index: int): bool =
   let flag = ParticipationFlags(1'u8 shl flag_index)
   (flags and flag) == flag
+
```
tersec, line 204: Empty trailing line

beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -20,7 +20,7 @@ import
     attestation_pool, exit_pool,
     block_pools_types, spec_cache
   ],
-  ".."/[beacon_node_types, ssz, beacon_clock]
+  ".."/[beacon_node_types, ssz, beacon_clock, beacon_chain_db]
```
tersec, line 23: This is a less-than-thrilling dependency.

beacon_chain/rpc/rpc_beacon_api.nim
```diff
@@ -412,7 +412,9 @@ proc installBeaconApiHandlers*(rpcServer: RpcServer, node: BeaconNode) {.
       # It was not integrated into the beacon node's database.
       return 202
     else:
-      let res = await proposeSignedBlock(node, head, AttachedValidator(), blck)
+      let res = await proposeSignedBlock(
+        node, head, AttachedValidator(),
+        ForkedSignedBeaconBlock(kind: BeaconBlockFork.Phase0, phase0Block: blck))
```
tersec, line 417: `ForkedSignedBeaconBlock.init` from https://github.com/status-im/nimbus-eth2/blob/65f6cd6282abbdddf963e8ee55f04d2cf7e75e65/beacon_chain/spec/forkedbeaconstate_helpers.nim#L54-L57

beacon_chain/validators/validator_duties.nim
```diff
@@ -312,63 +360,124 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
-      eth1Proposal.deposits,
-      node.exitPool[].getProposerSlashingsForBlock(),
-      node.exitPool[].getAttesterSlashingsForBlock(),
-      node.exitPool[].getVoluntaryExitsForBlock(),
-      default(ExecutionPayload),
-      restore,
-      cache)
+      return none(ForkedSignedBeaconBlock)
+
+    let doPhase0 = slot.epoch < node.dag.cfg.ALTAIR_FORK_EPOCH
```
tersec, line 365: I wrote this, but still am not happy with its level of duplication. It was intended as a semi-placeholder. `makeBeaconBlock` is this function with literally more than a dozen parameters, all of which are identical, and some not completely trivial to write, between two versions of this call, except for that the Altair version has `node.sync_committee_msg_pool[].produceSyncAggregate(head)` and the phase 0 version does not.

Unless it's prohibitively slow, the easiest fix here is to add a dummy `SyncAggregate` parameter to the phase 0 `makeBeaconBlock()` function, enabling fairly easy metaprogramming at worst to avoid this duplication. `block_sim` has the same problem, but I'm generally less averse there to hacky code such as this than in the main `nimbus_beacon_node` flow.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1199,3 +1305,14 @@ proc getProposer*(
...
+template validateCommitteeIndexOr*(idxParam: uint64,
+                                   epochRef: EpochRef,
+                                   failureCase: untyped): CommitteeIndex =
+  let committeesCount = get_committee_count_per_slot(epochRef)
+  let idx = idxParam
+  if idx < committeesCount:
+    IHaveVerifiedThis(CommitteeIndex, idx, "Just verified")
+  else:
+    failureCase
+
```
tersec, line 1318: Empty trailing line

beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
```diff
@@ -0,0 +1,167 @@
...
+    try:
+      produceSyncAggregateAux(pool.bestAggregates[target.root])
+    except KeyError:
+      raiseAssert "We have checked for the key upfront"
+  else:
+    SyncAggregate(sync_committee_signature: ValidatorSig.infinity)
+
+  debug "SyncAggregate produced",
+         target = target.root, value = shortLog(result)
+
```
tersec, line 167: Empty trailing line

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -345,3 +353,49 @@ proc voluntaryExitValidator*(
...
+  let v = validateSignedContributionAndProof(self.dag, self.syncCommitteeMsgPool,
+                                             contributionAndProof, wallTime,
+                                             checkSignature)
+  if v.isErr():
+    debug "Dropping sync committee contribution and proof",
+           validationError = v.error
+    return v.error[0]
+
+  ValidationResult.Accept
+
```
tersec, line 401: Empty trailing line

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -763,3 +769,214 @@ proc validateVoluntaryExit*(
...
+
+      return err((ValidationResult.Reject, cstring(
+        "validateSignedContributionAndProof: aggregate signature fails to verify")))
+
+    syncCommitteeMsgPool[].addSyncContribution(
+      msg.message.contribution,
+      cookedSignature.get)
+
+  ok()
+
```
tersec, line 979: Empty trailing line

beacon_chain/networking/eth2_network.nim
```diff
@@ -1855,19 +1851,47 @@ proc updateStabilitySubnetMetadata*(
...
+    debug "ENR fork id changed", value
+
 # https://github.com/ethereum/eth2.0-specs/blob/v1.0.1/specs/phase0/validator.md#phase-0-attestation-subnet-stability
 func getStabilitySubnetLength*(node: Eth2Node): uint64 =
   EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION +
     node.rng[].rand(EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION.int).uint64
 
 func getRandomSubnetId*(node: Eth2Node): SubnetId =
   node.rng[].rand(ATTESTATION_SUBNET_COUNT - 1).SubnetId
+
```
tersec, line 1897: Empty trailing line (obviously, these all got added in some systemic way)

beacon_chain/spec/network.nim
```diff
@@ -68,19 +75,31 @@ func compute_subnet_for_attestation*(
...
+    current_fork_version = cfg.forkVersionAtSlot(slot)
+    fork_digest = compute_fork_digest(current_fork_version,
+                                      genesis_validators_root)
   ENRForkID(
     fork_digest: fork_digest,
-    next_fork_version: current_fork_version,
-    next_fork_epoch: FAR_FUTURE_EPOCH) # TODO altair-transition
+    next_fork_version: cfg.ALTAIR_FORK_VERSION,
+    next_fork_epoch: cfg.nextForkEpochAtSlot(slot))
+
```
tersec, line 105: Empty trailing line

stefantalpalaru:
For the record, I like "empty trailing lines" because they separate the last line of code from my editor's status bar.

tersec:
> For the record, I like "empty trailing lines" because they separate the last line of code from my editor's status bar.

(1) Seems like that should be configurable within an editor, e.g., by creating additional vertical space between the status bar and contents of the edited file on the editor side.

(2) If one decides that source files should adopt that as a standard, it's at least worth doing so consistently. This, however, comes across as a little bit of random fallout of using probably some editor which inserts such lines easily, but it hasn't been done in any systematic way across the codebase. It creates a kind of sloppiness across the source code. Unless such a decision is made, it seems better to actively maintain a lack of empty trailing lines.

beacon_chain/rpc/rest_utils.nim
arnetheduck: `borrow` doesn't work with generic code in certain situations in 1.2, should not be used  at all (I've had to remove it from several places because buggy code was generated)

arnetheduck: it's broken when the calling code is generic (it's not about the borrowed code)

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -35,17 +35,23 @@ import
...
+    weak_subjectivity, signatures],
   ./consensus_object_pools/[
     blockchain_dag, block_quarantine, block_clearance, block_pools_types,
-    attestation_pool, exit_pool, spec_cache],
+    attestation_pool, sync_committee_msg_pool, exit_pool, spec_cache],
   ./eth1/eth1_monitor
 
 from eth/common/eth_types import BlockHashOrNumber
 
+from spec/datatypes/altair import
```
arnetheduck: altair already imported

beacon_chain/beacon_chain_db.nim
```diff
@@ -34,6 +34,10 @@ type
     eth1Block*: Eth2Digest
     depositContractState*: DepositContractState
 
+  ImmutableValidatorsMemoryStore* = object
+    byIndex*: seq[ImmutableValidatorData2]
+    byPubKey: Table[ValidatorPubKey, ValidatorIndex]
```
arnetheduck, line 39: unused?

beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
```diff
@@ -0,0 +1,167 @@
...
+    beaconBlockRoot: Eth2Digest,
+    signature: CookedSig,
+    committeeIdx: SyncCommitteeIndex,
+    positionInCommittee: uint64) =
+
+  if beaconBlockRoot notin pool.blockVotes:
+    pool.blockVotes[beaconBlockRoot] = @[]
+
+  try:
+    pool.blockVotes[beaconBlockRoot].add TrustedSyncCommitteeMsg(
```
arnetheduck, line 43: `mGetOrPut` gets rid of the exception and the double lookup

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -216,11 +216,17 @@ when hasGenesisDetection:
...
       let pubkey = deposit.pubkey
       if pubkey notin m.genesisValidatorKeyToIndex:
-        let idx = ValidatorIndex m.genesisValidators.len
+        let idx = m.genesisValidators.len
         m.genesisValidators.add ImmutableValidatorData(
           pubkey: pubkey,
           withdrawal_credentials: deposit.withdrawal_credentials)
-        m.genesisValidatorKeyToIndex[pubkey] = idx
+        m.genesisValidatorKeyToIndex[pubkey] =
+          IHaveVerifiedThisValue(ValidatorIndex, idx, """
```
arnetheduck, line 224: This looks pretty odd - `ValidatorIndex` is a distinct type which should _already_ raise awareness that it needs to be validated _every time_ before use - although very explicit, the extra stuff here does not actually find places where you _forgot_ to check - ie a comment and conversion would have been equally informative without being .. odd and redundant.

We could potentially introduce a `ValidValidatorIndex` distinct type which should be even more clear, and use that in places where assumptions are being made about validity - this would at least give a compiler error when someone uses it forgetfully (as opposed to sloppily)

beacon_chain/gossip_processing/batch_validation.nim
arnetheduck: a better question is why is it needed? any dag/key/state-related stuff should go via chaindag which is responsible for keeping keys and caches up to date

beacon_chain/beacon_chain_db.nim
tersec: Already discussed in https://github.com/status-im/nimbus-eth2/pull/2709#discussion_r679914541?

beacon_chain/beacon_chain_db.nim
```diff
@@ -591,6 +628,15 @@ proc getAltairBlock*(db: BeaconChainDB, key: Eth2Digest):
   else:
     result.err()
 
+template validateValidatorIndexOr*(idxParam: uint64, dbParam: BeaconChainDB,
```
arnetheduck, line 631: validating the validator index should be done based on the chaindag and the state against which the validator index is being validated, not the beaconchaindb (which is an internal implementation detail of the chaindag)

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -372,6 +373,10 @@ proc validateAggregate*(
     if v.isErr():
       return err(v.error)
 
+  let aggregatorIdx = aggregate_and_proof.aggregator_index
+    .validateValidatorIndexOr(pool.dag.db):
```
arnetheduck, line 377: ditto - this shouldn't use the db directly (just like `validatorKey` is part of `ChainDag`

beacon_chain/validators/validator_duties.nim
```diff
@@ -312,63 +360,124 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
+  # TODO altair
+  #if node.config.dumpEnabled:
+  #  dump(node.config.dumpDirOutgoing, newBlock)
 
-  node.network.broadcast(
-    getBeaconBlocksTopic(node.dag.forkDigests.phase0), newBlock)
+  case newBlock.kind:
+  of BeaconBlockFork.Phase0:
+    node.network.broadcast(
+      getBeaconBlocksTopic(node.dag.forkDigests.phase0), newBlock.phase0Block)
```
tersec, line 477: might be better to use `withBlck` for this, though it doesn't trivially handle the `node.dag.forkDigests.phase0` vs `node.dag.forkDigests.altair` on its own and barring that, could be better to keep the choices obviously synchronized in this manual way.

beacon_chain/validators/validator_duties.nim
```diff
@@ -171,9 +176,11 @@ proc sendAttestation*(
 
   return case ok
     of ValidationResult.Accept:
+      # TODO altair: this probably is just a function of wallslot's altair, not
```
tersec, line 179: This is my comment, so I'll follow up on this -- I'm pretty convinced this is true now. This should be a function of wall clock time, not the attestation, as with the other gossip broadcasts. At least it should be consistent with them, and in some of those cases (especially slashings) there's no specific slot to use.

beacon_chain/ssz/bitseqs.nim
```diff
@@ -265,6 +265,11 @@ func isZeros*(x: BitSeq): bool =
     if w != 0: return false
   return true
 
+func isZeros*(x: BitArray): bool =
```
arnetheduck, line 268: tests

beacon_chain/spec/network.nim
```diff
@@ -68,19 +75,31 @@ func compute_subnet_for_attestation*(
...
+                            committeeIdx: SyncCommitteeIndex): string =
   ## For subscribing and unsubscribing to/from a subnet.
-  eth2Prefix(forkDigest) & "beacon_attestation_" & $uint64(subnet_id) & "/ssz"
+  eth2Prefix(forkDigest) & "sync_committee_" & $(committeeIdx.asUInt8) & "/ssz"
 
-func getENRForkID*(fork_version: Version,
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.8/specs/altair/p2p-interface.md#topics-and-messages
+func getSyncCommitteeContributionAndProofTopic*(forkDigest: ForkDigest): string =
+  ## For subscribing and unsubscribing to/from a subnet.
+  eth2Prefix(forkDigest) & "sync_committee_contribution_and_proof" & "/ssz"
```
tersec, line 92: Could use tests like the others in https://github.com/status-im/nimbus-eth2/blob/c4bf4f8fffa10fed030db013e173dcb3b8ba24a3/tests/test_honest_validator.nim#L13-L19

beacon_chain/validators/validator_duties.nim
```diff
@@ -312,63 +360,124 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
+    warn "Unable to add proposed block to block pool"
+      # TODO altair
+      #newBlock = shortLog(newBlock.message),
+      #blockRoot = shortLog(newBlock.root)
 
     return head
 
   notice "Block proposed",
-    blck = shortLog(newBlock.message),
+    # TODO altair
```
arnetheduck, line 465: these kind of todos obviously need fixing before merging

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -468,10 +476,10 @@ proc validateAggregate*(
     var x = await cryptoFuts.slotCheck
     case x
     of BatchResult.Invalid:
-      return errReject(cstring("validateAggregate: invalid slot signature"))
+      return errReject("validateAggregate: validateAggregate: invalid slot signature")
     of BatchResult.Timeout:
       beacon_aggregates_dropped_queue_full.inc()
-      return err((ValidationResult.Reject, cstring("validateAggregate: timeout checking slot signature")))
+      return errReject("validateAggregate: timeout checking slot signature")
```
tersec, line 479: https://github.com/status-im/nimbus-eth2/blob/c4bf4f8fffa10fed030db013e173dcb3b8ba24a3/beacon_chain/gossip_processing/gossip_validation.nim#L162-L168 notes that `errReject` is meant explicitly not to arise from timeouts, because they could occur stochastically, without an underlying logic bug. `errReject` is meant to catch for-sure logic bugs in a CI context in a way more difficult to ignore and temporally closer to identifying root causes.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -480,11 +488,10 @@ proc validateAggregate*(
     var x = await cryptoFuts.aggregatorCheck
     case x
     of BatchResult.Invalid:
-      return errReject(cstring(
-        "validateAggregate: invalid aggregator signature"))
+      return errReject("validateAggregate: invalid aggregator signature")
     of BatchResult.Timeout:
       beacon_aggregates_dropped_queue_full.inc()
-      return err((ValidationResult.Reject, cstring("validateAggregate: timeout checking aggregator signature")))
+      return errReject("validateAggregate: timeout checking aggregator signature")
```
tersec, line 491: Another `errReject` on a timeout.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -493,11 +500,10 @@ proc validateAggregate*(
     var x = await cryptoFuts.aggregateCheck
     case x
     of BatchResult.Invalid:
-      return errReject(cstring(
-        "validateAggregate: invalid aggregate signature"))
+      return errReject("validateAggregate: invalid aggregate signature")
     of BatchResult.Timeout:
       beacon_aggregates_dropped_queue_full.inc()
-      return err((ValidationResult.Reject, cstring("validateAggregate: timeout checking aggregate signature")))
+      return errReject("validateAggregate: timeout checking aggregate signature")
```
tersec, line 503: `errReject` on a timeout.

tersec:
needs a ``proc `$`(foo: RestValidatorIndex)``?

beacon_chain/eth1/eth1_monitor.nim
zah: Yes, my intention was to gradually transition to an even more strict version of this type that rejects the coercion `ValidatorIndex(foo)`. It's yet to be seen how practical this is.

beacon_chain/beacon_chain_db.nim
zah: While it's true that a `ValidatorIndex` is a valid against a certain `BeaconState`, this property is difficult to describe within the Nim type system. Many of the `ValidatorIndex` values are long-lived and created far away from the position where they are used.

The less strict guarantee that a `ValidatorIndex` is something that you could use to obtain information about a validator from the database is easier to enforce in practice and it's still very useful in many of the contexts where validator indices are used.

zah: I would also argue that a single-responsibility object such as the "Immutable validator data store" is likely to survive many more types of refactorings than the multi-responsibility ChainDAG object, so there is nothing wrong to introduce it as a dependency in more places and thus reduce the API area of the ChainDAG object.

beacon_chain/eth1/eth1_monitor.nim
zah: After the changes in this branch, `ValidatorIndex` is assumed to be `ValidValidatorIndex`

beacon_chain/rpc/rest_utils.nim
zah: I merely copied the existing practice used for the other distinct types in the datatypes folder, but sure, I tend to use templates over borrow in fresh code as well.

beacon_chain/validators/validator_duties.nim
zah: Considering the time when attestations are produced and sent, the wallSlot and the attestation.slot are very unlikely to disagree? Furthermore, a consequential disagreement is possible only during the precise slot of the transition, so this is unlikely to matter in practice and we can save ourselves the trouble of getting the current time.

beacon_chain/gossip_processing/gossip_validation.nim
zah: What are the things that are known to time out in local sim?

beacon_chain/validators/validator_duties.nim
zah: Some TODOs were addressed and all remaining items are related to the REST API.

beacon_chain/beacon_chain_db.nim
arnetheduck: You can argue that, but then you need to refactor the code accordingly - the chaindag is currently responsible for cache coherence of the state stuff so all accesses should go through there - that's because the responsibility of updating the cache happen at the exact point where a _valid_ block is added to the chain dag - ergo, it's a natural responsibility for it.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: we should document it as such then and not use it in any types that pass through the magic serialization layer which doesn't do this check - but I'd still remove `IHaveVerifiedThisValue` which just looks onerous

beacon_chain/beacon_chain_db.nim
arnetheduck: Notably, the `BeaconChainDB` exists mostly to convert nim objects to key-value and back and not worry about logical assumptions around spec-based validity - it's part of the "trusted inner city" - ie anything given to it should already be "valid" and the `ChainDag` has the responsibility to make it so, with `clearance` being the moat.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: Also, it's worth considering how we're using the other aliases - `Slot` etc - whether those types are before or after validity checks - that's a problem with these magic serialization layers and reusing serialized types directly in logic - they hide conversions all too easily and you can't tell what's been validated and what hasn't.

zah: The current code demonstrates that the "valid vs wire" distinction is not that difficult to work with. All network messages carry types such as `uint64` and the application logic has to use the `validateX` functions to make these values usable in the rest of the codebase. Slots and Epochs are not subjected to this just because they don't have particularly invalid values.

zah: Otherwise, I'll introduce `ValidatorIndex.verifiedValue(x)` as a middle-ground solution.

arnetheduck: > The current code demonstrates that the "valid vs wire" distinction is not that difficult to work with

Bug count and severity tell a different story. 

> ValidatorIndex.verifiedValue(x)

we used `load` for public keys which is possible because they're self-contained - if we were to introduce a validation function, it would ideally take enough data as arguments for it to perform the validation - in this case, a chaindag and a uint64 that gives a `ValidatorIndex` etc

beacon_chain/validators/validator_duties.nim
tersec: That's reasonable. I still think the wall clock's the most correct approach, but also agree that it doesn't much matter for the reasons you state.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: Haven't seen one specifically yet, but I introduced `errReject` to be unambiguous. One doesn't have to guess, or ever wonder with it if, maybe, something just ran a little slow. If that changes, that undermines its purpose.

scripts/launch_local_testnet.sh
arnetheduck: it's not so hard to fix the VC actually, it doesn't do anything much different to the "normal" mode

arnetheduck:
work resumes in #2709 after all

PR #2764, opened 2021-08-06
Fix #2751

### LINKED ISSUE #2751: REST parser too strict on json with extra fields
Generally, in json it is customary that extra fields can be added to a response as a backwards-compatible change that doesn't require a new version - in order to be future-proof against such additions, the rest client should not raise an error in such cases.

zah:


arnetheduck:
is it possible to discover which fields were skipped? ie to display a warning / notification?

zah:
It will require more complicated bookkeeping involving a bitmask. I do have some plans for enhacements to the `FieldReadersTable` that will make it simpler.

zah:
Ah, sorry, I replied too quickly thinking this was addressed to the new `requireAllFields` feature in json_serialization that was submitted yesterday:
https://github.com/status-im/nim-json-serialization/commit/4f3775ddf48d9abee30c51a53862cea84a09fa78

It's possible to add support for a callback that will be executed when an "unknown field" is skipped.

arnetheduck:
> It's possible to add support for a callback that will be executed when an "unknown field" is skipped.

sounds reasonable

arnetheduck:
superceded by https://github.com/status-im/nimbus-eth2/pull/3005 for now - this might need revisiting in the future, but with better feedback on unexpected fields

PR #2770, opened 2021-08-09
altair enr switching

tersec:
and nothing else (well, it picks up a stray removal-of-redundant-type-conversion that seemed pointless to exclude)

Based on `altair`

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1586,11 +1586,11 @@ proc createEth2Node*(rng: ref BrHmacDrbgContext,
                      netKeys: NetKeyPair,
                      cfg: RuntimeConfig,
                      forkDigests: ForkDigestsRef,
-                     headSlot: Slot,
+                     headEpoch: Epoch,
                      genesisValidatorsRoot: Eth2Digest): Eth2Node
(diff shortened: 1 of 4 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -1584,16 +1586,11 @@ proc createEth2Node*(rng: ref BrHmacDrbgContext,
                      netKeys: NetKeyPair,
                      cfg: RuntimeConfig,
                      forkDigests: ForkDigestsRef,
+                     headSlot: Slot,
```
arnetheduck: `headEpoch` technically

arnetheduck: shouldn't this be .. wall time epoch btw?

tersec: https://github.com/status-im/nimbus-eth2/pull/2770/commits/13d78a35473dcad3094686118cb0b469665691c1 addresses the unambiguous point about using slots or epochs.

tersec: I agree, regarding walltime: https://github.com/status-im/nimbus-eth2/pull/2770/commits/10399322fca6af7a533b1aa4f0ea26c56eac7dfb

PR #2676, opened 2021-06-25
Implement the new Altair req/resp protocols

zah:
Also fixes the altair message-id computation by providing the correct
forkdigest prefix in `isAltairTopic`.

PR #2773, opened 2021-08-10
send attestations and exit messages on fork-appropriate topic

tersec:


beacon_chain/validators/validator_duties.nim
```diff
@@ -184,18 +185,18 @@ proc sendAttestation*(
       false
 
 proc sendVoluntaryExit*(node: BeaconNode, exit: SignedVoluntaryExit) =
```
arnetheduck: let's move all these helpers to eth2_network, and have it choose the right topic - then we don't need to spread the `getXxxTopic` logic all over the place

beacon_chain/validators/validator_duties.nim
```diff
@@ -171,9 +171,10 @@ proc sendAttestation*(
 
   return case ok
     of ValidationResult.Accept:
+      let forkPrefix = node.dag.forkDigestAtEpoch(
+        node.beaconClock.now.slotOrZero.epoch)
```
arnetheduck: let's comment here why we choose wall clock over attestation slot

beacon_chain/validators/validator_duties.nim
tersec: - They're in (a) `nimbus_beacon_node.nim` for sub/unsub purposes, (b) `validator_duties` for sending, and (c) a couple of RPC API files which shouldn't depend much on the rest of the codebase. But sure, there's reason to consolidate where possible.

- Doing so inverts the current architectural layering where `eth2_network` doesn't know/care about the DAG, etc. Because the relevant forkdigest aspects are tied up in `blockchain_dag` currently, this is in no way obviously better overall.

Probably, there are alternatives (split out the fork digest parts from the DAG enough to make it reasonable to access them from eth2_network; create intermediate layers to unify at least the RPC & validator duty parts, which genuinely are parallel to each other; etc), but each has their own tradeoffs. The RPC part, in particular, I'm wary of excessively unifying.

arnetheduck: > doesn't know/care about the DAG

it still wouldn't know about the dag - the only thing needed to form the topics is the `RuntimeConfig` and a `genesis_state_root` from which fork digests can be created for all forks - it already implicitly depends on this information through the `ForkDigestsRef` structure - I'm suggesting to make that dependency explicit instead. The additional "knowledge" we're adding is the mapping between topic and eth2 type which is currently spread out - this mapping is something that's set by the spec and therefor shouldn't live in `nimbus_beacon_node`.

I'm open to a new `eth2_gossip` module that contains only the mapping knowledge but that seems like overkill to me.

I've expanded a little bit on this architecture here: https://github.com/status-im/nimbus-eth2/pull/2676#discussion_r686566251

beacon_chain/validators/validator_duties.nim
```diff
@@ -741,7 +742,10 @@ proc handleValidatorDuties*(node: BeaconNode, lastSlot, slot: Slot) {.async.} =
         aggregateWaitTime = shortLog(aggregateWaitTime.offset)
       await sleepAsync(aggregateWaitTime.offset)
 
-    await broadcastAggregatedAttestations(node, head, slot)
+    let sendAggregatedAttestationsFut =
```
arnetheduck, line 724: what?

tersec: it's a fragment of a larger update which includes sync committee changes. Basically, creates two futures, awaits both.

this is a bit don't-need-it-yet here but it's also a harmless refactoring that reduces the `unstable`/`altair` diff. Am not especially attached to it specifically, however.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/2773/commits/f46c205601d875315f7937b2d799ef695e9353ad

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/2773/commits/4d56bf25402394516c6133ad1299ae3e1d92139a

While a new `eth2_gossip`-like module would still be useful to further isolate wall clock-dependent parts of eth2_network, that's mostly an issue around testability, and anything the main job of which is to send/receive gossip messages already has most of the testability downsides of wall clock-dependent code, so the marginal loss isn't much.

beacon_chain/networking/eth2_network.nim
```diff
@@ -1124,10 +1126,11 @@ proc onConnEvent(node: Eth2Node, peerId: PeerID, event: ConnEvent) {.async.} =
               peer = peerId, peer_state = peer.connectionState
       peer.connectionState = Disconnected
 
-proc new*(T: type Eth2Node, config: BeaconNodeConf,
+proc new*(T: type Eth2Node, config: BeaconNodeConf, runtimeCfg: RuntimeConfig,
           enrForkId: ENRForkID, forkDigests: ForkDigestsRef,
-          switch: Switch, pubsub: GossipSub, ip: Option[ValidIpAddress],
-          tcpPort, udpPort: Option[Port], privKey: keys.PrivateKey, discovery: bool,
+          beaconClock: BeaconClock, switch: Switch, pubsub: GossipSub,
```
arnetheduck: in keeping with comments elsewhere, this will need to transition back to a `GetWallTimeFn` callback so that it can be replaced with a different clock - in particular a testing clock that can be used to control time and therefore test the epoch cc @zah

beacon_chain/networking/eth2_network.nim
```diff
@@ -1892,3 +1898,38 @@ func getStabilitySubnetLength*(node: Eth2Node): uint64 =
 
 func getRandomSubnetId*(node: Eth2Node): SubnetId =
   node.rng[].rand(ATTESTATION_SUBNET_COUNT - 1).SubnetId
+
+func forkDigestAtEpoch(node: Eth2Node, epoch: Epoch): ForkDigest =
+  if epoch < node.cfg.ALTAIR_FORK_EPOCH:
+    node.forkDigests.phase0
```
arnetheduck, line 1901: since all of `clock`, `cfg` and `genesisValidatorsRoot` are part of constructing an `eth2_network`, we no longer need the ForkDigests ref here - we can use the above to generate the correct fork digest on the fly, just like we generate the initial `enrForkId` - which relies on the same data - we shouldn't be using two different methods for enr and gossip

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/2773/commits/582b2ddd76b05dea4780193fbdab48617eb9400f does this. It's not great, in that it also still needs genesis time. `MIN_GENESIS_TIME + GENESIS_DELAY` from the `RuntimeConfig` isn't enough, because it could have waited on `MIN_GENESIS_ACTIVE_VALIDATOR_COUNT` too.

Another approach would be to create another callback, call it `GetBeaconTimeFn` or `GetSlotFn` or similar, which would be either an object or closure environment with just the genesis time plus the callback. However, creating this closure would typically pull in a lot of unnecessary information, so this seemed overall less risky while achieving a similar goal.

This doesn't come up in the other `GetTimeFn` consumers because they already have access to `BeaconClock`.

beacon_chain/networking/eth2_network.nim
tersec: `forkDigests` was already there, used for:
https://github.com/status-im/nimbus-eth2/blob/bfe5e7460754b2ac0ca7a2a1e839676789c86208/beacon_chain/networking/eth2_network.nim#L1611-L1617

zah: The primary reason why `ForkDigests` was introduced in the first place was to avoid generating fork digests prefixes on-the-fly, all the time. There was an alternative idea to create a complete table of all topic names, computed once during initialization - something like this could work as an alternative to the `ForkDigests` structure.

beacon_chain/networking/eth2_network.nim
zah: I think the most consensus-like decision that came out of the `BeaconClock` discussion is that we can introduce a `GenesisTime` type which will replace many of the usages of the current `BeaconClock` type. You need a `GenesisTime` value, a wall time value and a `RuntimeConfig` to compute the current slot (this would be the case after the `SECONDS_PER_SLOT` runtime setting is implemented).

In other words, the `Eth2Node` will then be initialized with a `GenesisTime` instead of a `BeaconClock` as here.

beacon_chain/networking/eth2_network.nim
arnetheduck: >  something like this could work as an alternative to the ForkDigests structure. something like this could work as an alternative to the ForkDigests structure.

yes - this would be nice in general because all names ant topics are known ahead of time

tersec: The sync committee and (non-aggregated) attestation topics are generated by appending numbers to these generated strings and it's a bit pointless to store dozens/hundreds of topics when they're all basically the same.

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/2773/commits/a97f23b65450d32458d1ed431974dad96a221544 does this. It seems like a reasonable approach to at least start with. It ensures that the `now()`/`fromNow()`/etc functions of `beacon_clock` are disallowed, while providing access to the one function (`toBeaconTime()` that's a function of `genesis`.

It would make sense, once this is merged, to adopt `GenesisTime` elsewhere for similar reasons, but I regard that as beyond the scope of this PR.

arnetheduck: this looks weird to say the least - a `getBeaconTimeFn` callback would conceptually be a lot simpler I think - above all because system time never really interesting in a beacon node context.. I imean, it's not wrong, it's just .. cumbersome passing an extra constant around at all times: unix epoch _and_ genesis time - `BeaconTime` was kind of invented as a variation on `Time` but with a simplified starting point

tersec: https://github.com/status-im/nimbus-eth2/pull/2773/commits/a16b66add535484564341faa6920e63a05d3cfa9 replaces this `GenesisTime` and `GetTimeFn` setup with a simpler `GetBeaconTimeFn`.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -302,9 +305,9 @@ proc init*(T: type BeaconNode,
     netKeys = getPersistentNetKeys(rng[], config)
     nickname = if config.nodeName == "auto": shortForm(netKeys)
                else: config.nodeName
+    getBeaconTime = dag.beaconClock.getBeaconTimeFn()
```
arnetheduck, line 308: so.. is the beacon clock in the dag used for anything else after these changes?

arnetheduck (approved):

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/blob/a16b66add535484564341faa6920e63a05d3cfa9/beacon_chain/sync/sync_protocol.nim#L126-L133

arnetheduck: this uses a callback though, not the beaconClock - so in my eyes, the beaconClock can now be removed from chaindag completely

tersec: Ah, misread your question. Not sure, checking.

tersec: It looks like one can probably move `beaconClock` out of `dag`.

tersec: https://github.com/status-im/nimbus-eth2/pull/2796

PR #2777, opened 2021-08-11
Book: update to specific version page

unixpi:


arnetheduck:
the make update commands need to be followed by the normal `make nimbus_beacon_node` commands to build a new binary then restart the service - best cross-link this to the "update" page - or possibly combine (upgrading and downgrading is the same "procedure" just that you specifically pick a version)

unixpi:
close in favour of https://github.com/status-im/nimbus-eth2/pull/3105

PR #2694, opened 2021-07-02
add checkDir = true to walkDir() across tests

tersec:
Otherwise, tests can go missing across test fixture versions without explicit `make test` errors.

PR #2784, opened 2021-08-16
sync committee/aggregate signature signing and verification

tersec:
Copies of `beacon_chain/spec/signatures.nim` and `beacon_chain/validators/validator_pool.nim` from `altair`, modulo removal of a trailing newline. Catches the test result summary up.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -115,6 +115,18 @@ OK: 3/3 Fail: 0/3 Skip: 0/3
 + Mocked start private key                                                                   OK
 ```
 OK: 3/3 Fail: 0/3 Skip: 0/3
+## Message signatures
+```diff
++ Aggregate and proof signatures                                                             OK
++ Attestation signatures                                                                     OK
++ Deposit signatures                                                                         OK
++ Slot signatures                                                                            OK
++ Sync committee message signatures                                                          OK
++ Sync committee selection proof signatures                                                  OK
++ Sync committee signed contribution and proof signatures                                    OK
++ Voluntary exit signatures                                                                  OK
+```
+OK: 8/8 Fail: 0/8 Skip: 0/8
 ## Old database versions [Preset: mainnet]
 ```diff
 + pre-1.1.0                                                                                  OK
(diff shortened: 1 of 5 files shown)


arnetheduck:
what tests cover this?

tersec:
None, just like the rest of the signature code.

tersec:
One general style of sanity tests that'd work here is something round-trip -- at least for `spec/signatures.nim`, sign, then check successful signature verification, and also verify that expected failure in other cases should happen too.

arnetheduck:
> One general style of sanity tests that'd work here is something round-trip -- at least for spec/signatures.nim, sign, then check successful signature verification, and also verify that expected failure in other cases should happen too.

this would certainly help exclude a class of bugs when looking for issues - some of the signature stuff is indirectly tested through gossip validation tests but that's quite limited.

tersec:
> this would certainly help exclude a class of bugs when looking for issues - some of the signature stuff is indirectly tested through gossip validation tests but that's quite limited.

https://github.com/status-im/nimbus-eth2/pull/2784/commits/b043fe55a275ceb440b960d1107d59d1e8f1c729

arnetheduck (approved):

AllTests-mainnet.md
```diff
@@ -38,12 +38,13 @@ OK: 12/12 Fail: 0/12 Skip: 0/12
...
 ## Bit fields
 ```diff
++ isZeros                                                                                    OK
 + iterating words                                                                            OK
 + overlaps                                                                                   OK
 + roundtrips BitArray                                                                        OK
 + roundtrips BitSeq                                                                          OK
 ```
-OK: 4/4 Fail: 0/4 Skip: 0/4
+OK: 5/5 Fail: 0/5 Skip: 0/5
```
arnetheduck, line 47: now that we have the unit test reports in CI, shall we .. remove these markdown files?

tersec: If you're referring to
2021-08-17-091920_672x274_scrot
then I prefer having non-CI-dependent ways of seeing changes in the test suite, of ensuring tests aren't appearing/disappearing invisibly which I can see before I commit. Between some of the v1.1.0 alpha/betas, the test vectors would shift sometimes without always showing a net change in the number of tests, too.

tersec: https://github.com/status-im/nimbus-eth2/pull/2694 and https://github.com/status-im/nimbus-eth2/pull/2703 closed one loophole whereby whole test vector directories could disappear without any other obvious tell, but others might still exist.

tersec: This can be addressed (or not) in a future PR.

PR #2646, opened 2021-06-11
Move ssz code to nim-ssz-serialization

kdeme:
Finished up what I had laying around.

- `updateRoot` still feels quite the hack, as it still a parameter of the nim-ssz-serialization `readSszValue` proc.
- The whole `Digest` vs `Eth2Digest` also is a bit of a mess and prone to compile errors.
- `test_ssz_merkleization` is not added in nim-ssz-serialization repository as it would require more code moving. But, I actually just realized that it is not even run right now in the test suite.
- The fixture tests can probably stay in nimbus-eth2 (for now)

PR #2785, opened 2021-08-16
disentangle eth2 types  from the ssz library

arnetheduck:
This PR continues the work in
https://github.com/status-im/nimbus-eth2/pull/2646,
https://github.com/status-im/nimbus-eth2/pull/2779 as well as past
issues with serialization and type, to disentangle SSZ from eth2 and at
the same time simplify imports and exports with a structured approach.

The principal idea here is that when a library wants to introduce SSZ
support, they do so via 3 files:

* `ssz_codecs` which imports and reexports `codecs` - this covers the
basic byte conversions and ensures no overloads get lost
* `xxx_merkleization` imports and exports `merkleization` to specialize
and get access to `hash_tree_root` and friends
* `xxx_ssz_serialization` imports and exports `ssz_serialization` to
specialize ssz/nim-serialization for a specific library

Those that need to interact with SSZ always import the `xxx_` versions
of the modules and never `ssz` itself so as to keep imports simple and
safe.

This is similar to how the REST / JSON-RPC serializers are structured in
that someone wanting to serialize spec types to REST-JSON will import
`eth2_rest_serialization` and nothing else.

* split up ssz into a core library that is independendent of eth2 types
* rename `bytes_reader` to `codec` to highlight that it contains coding
and decoding of bytes and native ssz types
* remove tricky List init overload that causes compile issues
* get rid of top-level ssz import
* reenable merkleization tests
* move some "standard" json serializers to spec
* remove `ValidatorIndex` serialization for now

There are a few things this PR doesn't solve:

* like #2646 this PR is weak on how to handle root and other
dontSerialize fields that "sometimes" should be computed - the same
problem appears in REST / JSON-RPC etc


tests/ssz/test_ssz_merkleization.nim
```diff
@@ -249,21 +249,21 @@ func attachMerkleProofsReferenceImpl(deposits: var openArray[Deposit]) =
       DEPOSIT_CONTRACT_TREE_DEPTH + 1, val_idx.uint64,
       deposit_data_sums[val_idx])
 
-let
-  digests = mapIt(1..65, eth2digest toBytesLE(uint64 it))
+# let
```
tersec: either remove or uncomment this whole section?

arnetheduck: removed file, it's broken anyway

zah:
I don't think this continues the work in https://github.com/status-im/nimbus-eth2/pull/2646. Instead, it makes an unnecessary step in a different direction. The main novelty here seems to be the module import/re-export strategy which can be easily applied on top of the other PR in a much smaller patch.

beacon_chain/ssz/codec.nim
```diff
@@ -6,16 +6,20 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
```
zah, line 6: There is a lot of decoding here, but not much coding (that would be the `writer` module), so I find the name `codec` a bit misleading.

beacon_chain/ssz/codec.nim
```diff
@@ -256,9 +228,22 @@ func readSszValue*[T](input: openArray[byte],
...
+template toSszType*(v: HashArray): auto = v
+template toSszType*(v: BitList): auto = v
+template toSszType*(v: BitArray): auto = v
+template toSszType*(v: array): auto = v
+template toSszType*(v: seq[byte]): seq[byte] = v
+template toSszType*(x: Eth2Digest): auto = x.data
+template toSszType*(v: object): auto = v
+template toSszType*(v: tuple): auto = v
+
+template toSszType*(v: enum): auto =
```
zah: As discussed in https://github.com/status-im/nim-eth/issues/391, I don't think this leads to better error messages.

beacon_chain/ssz/ssz_serialization.nim
```diff
@@ -222,23 +222,20 @@ proc writeValue*[T](w: var SszWriter, x: SizePrefixed[T]) {.raises: [Defect, IOE
...
 
-proc readValue*[T](r: var SszReader, val: var T) {.raises: [Defect, MalformedSszError, SszSizeMismatchError, IOError].} =
+proc readValue*(r: var SszReader, val: var auto) {.raises: [Defect, MalformedSszError, SszSizeMismatchError, IOError].} =
+  mixin readSszBytes
+  type T = type val
   when isFixedSize(T):
     const minimalSize = fixedPortionSize(T)
     if r.stream.readable(minimalSize):
-      readSszValue(r.stream.read(minimalSize), val, r.updateRoot)
+      readSszBytes(r.stream.read(minimalSize), val, r.updateRoot)
```
zah, line 231: This change doesn't look right. What was the intention?

beacon_chain/ssz/ssz_serialization.nim
```diff
@@ -222,23 +222,20 @@ proc writeValue*[T](w: var SszWriter, x: SizePrefixed[T]) {.raises: [Defect, IOE
...
     else:
       raise newException(MalformedSszError, "SSZ input of insufficient size")
   else:
     # TODO(zah) Read the fixed portion first and precisely measure the
     # size of the dynamic portion to consume the right number of bytes.
-    readSszValue(r.stream.read(r.stream.len.get), val, r.updateRoot)
+    readSszBytes(r.stream.read(r.stream.len.get), val, r.updateRoot)
```
zah, line 237: Ditto

beacon_chain/ssz/types.nim
```diff
@@ -133,9 +134,6 @@ type
 
 template asSeq*(x: List): auto = distinctBase(x)
 
-template init*[T](L: type List, x: seq[T], N: static Limit): auto =
```
zah, line 136: Why is this removed?

tests/test_ssz_merkleization.nim
```diff
@@ -1,269 +0,0 @@
-import
```
zah, line 1: Was this file removed? Why?

beacon_chain/ssz/codec.nim
arnetheduck: well, `bytes_reader` doesn't fit `toSszType`, got any better ideas?

beacon_chain/ssz/ssz_serialization.nim
arnetheduck: `readSszBytes` becomes the `mixin` point for implementing `updateRoot` for custom types - this still needs a better solution

arnetheduck:
> I don't think this continues the work in #2646. Instead, it makes an unnecessary step in a different direction. The main novelty here seems to be the module import/re-export strategy which can be easily applied on top of the other PR in a much smaller patch.

it's easier to work on things in isolation without involving other repos - the move to a different repo is two steps: disentangle the code, _then_ move the code to a different repo - this PR focuses on the first step. After that, the second step is mechanical - when doing refactoring, I find it easier to avoid mixing different orthogonal concerns, just like "normal" PR's.

It's a continuation because the PR introduced many of the ideas in this branch, but did not provide integration with how imports and exports _now_ work. If you look at the other PR, a lot of the code has moved on since then, so the amount of work involved in rebasing that branch is about the same as copying the now-extracted library to a different repo.

The difference is that the steps can be taken freely at any time now - crucially, the work is exactly the same: update the other code to this structure - the only difference is whether we execute `git bla bla bla` or `cp bla bla bla`.

beacon_chain/ssz/types.nim
arnetheduck: its presence causes some modules to randomly fail to compile, presumably because it triggers some compiler bug:

```
/home/arnetheduck/status/nimbus-eth2/vendor/nim-serialization/serialization.nim(19, 31) Error: type mismatch: got <type SszWriter, OutputStreamHandle>
but expected one of: 
template init[T; ](L: type List; x: seq[T]; N: static Limit): auto
  required type for L: type List
  but expression 'WriterType`gensym69265608' is of type: type SszWriter

expression: init(WriterType`gensym69265608, s`gensym69265607)
```

since it's unused, there's little point in fighting for it.

tests/test_ssz_merkleization.nim
arnetheduck: it doesn't compile (it was never part of CI), the test in it that do compile don't pass and hopefully, it should already be covered by spec tests

arnetheduck:
> module import/re-export strategy

many compiler bugs in SSZ and serialization in particular are either fixed or caused by using the proper imports and exports - many of them are caused because nim sort of allows circular imports, but this quickly breaks down in unexpected ways with generic code and mixins - it's an important part of making this whole thing work reasonably and reliably.

beacon_chain/ssz/ssz_serialization.nim
arnetheduck: ie this is something that #2646 also doesn't solve, but solving it becomes difficult in a single PR because of the sheer amount of changes needed (the less bad solutions I can think of right now involve nim-ser, but that's not .. ideal either)

arnetheduck: it's also the case that for reading SSZ from an in-memory structure, it's actually easier to bypass nim-ser completely and call `readSszBytes` directly - this is done in a few places already because of exceptions and observable / destructive stores - having a single entry point into the library that works with `openArray[byte]` makes it easier to audit.

beacon_chain/ssz/codec.nim
arnetheduck:
```
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(389, 11) unsupported
/home/arnetheduck/status/nimbus-eth2/beacon_chain/spec/eth2_ssz_serialization.nim(13, 16) template/generic instantiation of `readSszValue` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/codec.nim(183, 58) template/generic instantiation of `fixedPortionSize` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(432, 28) template/generic instantiation of `enumAllSerializedFields` from here
/home/arnetheduck/status/nimbus-eth2/vendor/nim-serialization/serialization/object_serialization.nim(151, 32) template/generic instantiation of `enumAllSerializedFieldsImpl` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(433, 23) template/generic instantiation of `isFixedSize` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(416, 28) template/generic instantiation of `enumAllSerializedFields` from here
/home/arnetheduck/status/nimbus-eth2/vendor/nim-serialization/serialization/object_serialization.nim(151, 32) template/generic instantiation of `enumAllSerializedFieldsImpl` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(417, 27) template/generic instantiation of `isFixedSize` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(409, 26) template/generic instantiation of `toSszType` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/codec.nim(248, 17) template/generic instantiation of `unsupported` from here
/home/arnetheduck/status/nimbus-eth2/beacon_chain/ssz/types.nim(389, 11) Error: SSZ serialization of the type Slot = distinct uint64 is not supported, overload toSszType and fromSszBytes
```

best of both worlds - `codec.nim(248, 17)` now links to the right place in the code where the issue is documented - a bonus point is that even if something that calls `toSszType` forgets to use `unsupported`, the error message will still consistently be there, which was what originally bothered me.

d82121aa7823a64eade4d9ed6940dd213c136657

beacon_chain/ssz/codec.nim
arnetheduck: notably, there is no writer module - this is because reading uses generic `openArray[byte]` readers while writing is specialized/limited to work with nim-ser only, and therefore sits in the `ssz_serialization` module

beacon_chain/ssz/ssz_serialization.nim
zah: `readSszBytes` is a "leaf" function in the seriazation logic. It's responsible for handling a single field or an element of a container. It's not really a substitute for `readSszValue` which knows how to handle various composite SSZ objects.

The solution for the `updateRoot` problem is not quite final in Kim's branch, but it's the right way to solve this problem in my opinion. It will eventually work for all serialization formats and not just SSZ.

beacon_chain/ssz/codec.nim
zah: There will be a `writer` module eventually when the library is fully structured as a nim-serialization package.
For the other libraries, we have modules such as `eth2_json_serialization`, so having a `eth2_ssz_serizalition` module will be consistent with this.

beacon_chain/ssz/ssz_serialization.nim
arnetheduck: not quite, `fromSszBytes` is the leaf function you're referring to - `readSszBytes` is a new function that didn't exist before that was introduced to improve efficiency of the beacon chain db and bypass nim-ser completely - it simply does the same thing as `readValue`, but using an `openArray` parameter instead of an `SszReader` - this makes it slightly easier to reason about because there's less abstraction.

> right way to solve this problem in my opinion

the problem is actually the `updateRoot` parameter which really shouldn't be part of `SSZ` in general, because not all types have roots or require semantics that resemble the roots problem - for now, this is just an easy way of working around the issue pragmatically, I can switch it to kims solution but it's broken the exact same way (prone to overly generic matching, doesn't solve the actual problem) - my approach in this branch ensure that the "workaround" does not grow beyond the "main" type being deserialized - kims solution applies the postXxx function to all nested fields as well which is currently not a use case and ideally shouldn't become one.

tests/test_ssz_merkleization.nim
zah: I remember working with this file not so long ago, so I prefer if this is filed as an issue that should be fixed.

arnetheduck: if by "not so long ago" you mean almost a year, then sure :) 

https://github.com/status-im/nimbus-eth2/commit/b022dc4d1fd898c39140c7a866121e927efafbfe#diff-a54d69f285d23ebf24346e3eb7da9e4349002a66725563aed0ffebef067eb218

here's an issue: https://github.com/status-im/nimbus-eth2/issues/2789 - until then, I suggest it's removed since that represents the more honest state of the codebase and doesn't confuse refactoring

beacon_chain/ssz/codec.nim
arnetheduck: this PR indeed introduces a `eth2_ssz_serizalition` module that deals with "nim-serialization"-style reading _and_ writing. 

`codec` (and `bytes_reader`) are different in that they work with raw nim types vs bytes without involving `nim-serialization` (per wasm argument that you highlighted) - the two functions, `toSszType` and `fromSszBytes` mirror each other semantically (almost, since toSszType needs one more step to get to raw bytes), and are generally used for coding and decoding respectively.

I can imagine `toSszType` could be moved into `types` or a separate `bytes_writer` but that would be cumbersome - to extend the library, one would have to create an `eth2_bytes_reader` and a `eth2_ssz_types`/`eth2_ssz_writer` module to host the two functions

zah:
> it's easier to work on things in isolation without involving other repos - the move to a different repo is two steps: disentangle the code, then move the code to a different repo - this PR focuses on the first step. After that, the second step is mechanical - when doing refactoring, I find it easier to avoid mixing different orthogonal concerns, just like "normal" PR's.

Right now, Kim's work has the advantage that it can actually compile the project with a ssz package as a dependency. Here, we only have a speculation that this should be easy. The PR will compare much more favorably if it takes the job to completion.

beacon_chain/ssz/codec.nim
arnetheduck: > eth2_json_serialization

not quite - we put generic "logging" serialization based on json_serialization in the same module as the type, _mostly_, and we put more specific json serialization in more specific modules, such as `eth2_rest_serialization` -this latter module will likely include SSZ serialization also through an import/export because `REST` requires both SSZ and a specific json encoding that might differ from "logging" json.

we also have a `eth2_json_rpc_serialization` which is based on `std/json` - this serialization is different from the REST serialization and we cannot change it without breaking the JSON-RPC api for users (incidentally, this also means we cannot trivially change  nim-json-rpc to use nim-json-serialization because the two libraries do not necessarily encode things the same way)

beacon_chain/ssz/ssz_serialization.nim
zah: So, the way `updateRoot` is used requires only the top-level value to be computed right now?
Home come it was not implemented as a simple wrapper function in the past?

```nim
proc decodeAndUpdateRoot(T: type, ...) =
  result = SSZ.decode(T, ...)
  updateRoot(result)
```

arnetheduck: > So, the way updateRoot is used requires only the top-level value to be computed right now?

yes - it's mainly needed in eth2_network for blocks basically for the p2p macro and gossip - because these use `nim-serialization` and it was difficult to understand how to disentangle them _and_ keep using `nim-ser` in general, the `updateRoot` thing became a minimally invasive solution that didn't require touching the macro and breaking something else - the key idea behind the change was that `root` should be computed only once, somewhere early, before the object becomes immutable.

tests might also rely on it, to a lesser extent - don't remember. the db on the other hand must not update any roots - it already knows the root and assigns it manually.

arnetheduck:
> This is a macro, you should use the error function (just like the code below).

199be855f32bfa46c998e4718007887349d9736b

beacon_chain/ssz/ssz_serialization.nim
zah: It will be very easy to modify the networking layer to call the special function that computes the root. Looks like a good way to get rid of the `updateRoot` variable which complicates things quite a bit.

arnetheduck: yeah, I was looking into that - it's a bit risky because of generics and the expectation that `SSZ.decode` will give a valid instance which was why `updateRoot` was made to be true by default - that said, the absolutely easiest thing is to skip `SSZ.decode` entirely and just rely on `readSszBytes` in eth2 exclusively

PR #2790, opened 2021-08-17
Number of REST fixes for Altair.

cheatfate:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -58,115 +58,193 @@ type
...
+        stream.getOutput(seq[byte])
+      except SerializationError:
+        default
+      except IOError:
+        default
+  res
+
+proc jsonResponseWRoot*(t: typedesc[RestApiResponse], data: auto,
+                        dependent_root: Eth2Digest): RestApiResponse {.
+     raises: [Defect].} =
```
arnetheduck: why not a `push raises defect`?

beacon_chain/validator_client/fork_service.nim
```diff
@@ -1,23 +1,70 @@
...
+        await vc.getForkSchedule()
+      except ValidatorApiError as exc:
+        error "Unable to retrieve fork schedule", reason = exc.msg
+        return
+      except CatchableError as exc:
+        error "Unexpected error occured while getting fork information",
+              err_name = exc.name, err_msg = exc.msg
+        return
+
+    let fork =
```
arnetheduck, line 57: so.. if we get a fork here that hasn't happened yet, this code will keep polling?

cheatfate: Yes it will keep polling it every epoch.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
cheatfate: I'm just not sure that `{.async.}` procedures could be annotated with `Defect` only.

cheatfate: Fixed.

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -788,21 +765,71 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+      notice "Attestation sent", attestation = shortLog(a), delay = delayStr
+
+    # Since our validation logic supports batch processing, we will submit all
+    # attestations for validation.
+    let pending =
+      block:
+        var res: seq[Future[void]]
+        for attestation in attestations:
+          res.add(processAttestation(attestation))
```
arnetheduck, line 789: sendAttestation?

PR #2795, opened 2021-08-18
remove `updateRoot` param from SSZ

arnetheduck:
call `readSszBytes` directly to skip root updates

### DIFF
--- a/beacon_chain/spec/eth2_ssz_serialization.nim
+++ b/beacon_chain/spec/eth2_ssz_serialization.nim
@@ -7,8 +7,9 @@
 
 {.push raises: [Defect].}
 
-# Import this module to get access to `SSZ.encode` and `SSZ.decode` for spec types
-
+# This module exports SSZ.encode and SSZ.decode for spec types - don't use
+# ssz_serialization directly! To bypass root updates, use `readSszBytes`
+# without involving SSZ!
 import
   ./ssz_codec,
   ../ssz/ssz_serialization,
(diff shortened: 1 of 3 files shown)


beacon_chain/spec/eth2_ssz_serialization.nim
```diff
@@ -38,3 +39,13 @@ template readSszBytes*(
 template readSszBytes*(
     data: openArray[byte], val: var altair.TrustedSignedBeaconBlock, updateRoot = true) =
   readAndUpdateRoot(data, val, updateRoot)
+
+template readSszBytes*(
+    data: openArray[byte], val: var auto, updateRoot: bool) =
+  readSszValue(data, val)
+
+func readSszBytes(T: type, data: openArray[byte], updateRoot = true): T {.
```
zah, line 47: Isn't this naming a bit obscure? `SSZ.decode` doesn't update the root, but `readSszBytes` does. I can also write inconsistent expressions such as `readSszBytes(uint64, bytes, updateRoot = true)`. A more specific overload such as `readSszBytesAndDeriveRoot(x: BeaconBlock)` makes more sense IMO. The `updateRoot` value is never determined at run-time as far as I can tell, so the right overload can be selected statically.

arnetheduck: > `SSZ.decode`

`SSZ.decode` does update the root by virtue of calling the `readSszBytes` function with the defaulted `true` parameter - the objective is that the root update should be opt-out, not matter what method (ssz or direct) is used because it's an optimization (it's never wrong to update the root, just slow). As such, a function like `readSszBytesAndDontDeriveRoot`  could be introduced, but that seems over the top.

arnetheduck: >  The updateRoot value is never determined at run-time as far as I can tell, so the right overload can be selected statically.

It's determined by a jungle of macros, generics and templates, which is why it's true by default - this way the default is safe even when refactoring happens.

PR #2796, opened 2021-08-19
move BeaconClock from ChainDAG to BeaconNode

tersec:


arnetheduck (approved):

zah:
The notion that the genesis time doesn't belong in the ChainDAG is silly. What are the criteria for deciding what belongs there? Why are for example the list of beacon proposers for every epoch more reasonable to store there than the genesis time?

arnetheduck:
> The notion that the genesis time doesn't belong in the ChainDAG is silly. 

The genesis time is _still_ in chaindag - `dag.headState.data.data.genesis_time`.

> Why are for example the list of beacon proposers for every epoch more reasonable to store there than the genesis time?

This is quite simple really - the list of proposers is derived from the DAG - depending on where in the dag you are, the list of proposers is different so a the dag is necessary and directly related to the list of proposers. If you move in the dag, the list of proposers changes. However, if the wall time moves, _nothing_ in the dag changes - ergo, wall time is not a natural part of the dag - the easiest way to understand this point is that nothing in the dag uses or depends directly on "current time".

The `BeaconClock` is **not** about genesis time - it's about .. well, a clock like the name implies - typical functions involving the clock are `now`, `fromNow`, `durationToNextEpoch` etc. Perhaps this is what the confusion is about: a `BeaconClock` is not a source for the genesis timestamp. It's a collection of helpers, one derived from the other, that deal with the notion of "current time" in units that are convenient for the rest of the application to consume.

The `Beacon` part is that the clock returns `BeaconTime` instances in particular (as opposed to any other clock that might return `Time`, or seconds-since-computer-started, and so on), and to do that, it needs access to a genesis time because the _implementation_ of the system clock is not based on beacon time - it's based on unix time so a value is needed for the conversion - the "genesis time" member of the beacon clock is an implementation detail.

The beacon clock also has a function called `toBeaconTime` - this is the only (kind of) function that doesn't rely on the _current_ `Time` but rather .. any `Time` instance - it could conceivably be moved elsewhere, but doing so would complicate all other functions in the clock, and serves no real .. purpose? Because `Time` is also a concept related to the system clock, and a genesis time is an implementation detail to convert it to a `BeaconTime` upon which the rest of the beacon-chain part of application is based.

arnetheduck:
Now; one potential refactoring that _would_ make sense would be to move the `BeaconTime` type to `spec/` instead - for example `spec/beacon_time` or something like this - since `BeaconTime` is to `Slot` roughly what `second` is to a `minute`, we could collect all of `BeaconTime`, `Slot`, `Epcoh` and so on in a single spec module - this makes sense insofar as sub-slot time does sometimes appear in the spec, and having a type to represent an absolute time relative to the start of the chain, is somewhat reasonable - if I was writing the  time functions _today_ I'd probably do that - I'd also use `int64` for `BeaconTime` since `Duration` doesn't support negative values very well.

zah:
I was arguing specifically about the genesis time and perhaps the round-about way to reach it is good enough indeed. I still anticipate that we'll end up relying more on directly accessing it instead of pushing the current model all the way to `getSlotFn`.

PR #2798, opened 2021-08-20
Fix the Prater ENR fork id

zah:


Menduist:
Fixes the issue :+1:

Menduist:
Is the `beacon_chain/spec/debug_datatypes` file presence voluntary?

arnetheduck:
there we go, removed the extra file

PR #2805, opened 2021-08-22
update REST API spec/documentation URLs

tersec:
The previous/current URLs don't work. This is a mass search and replace from
> `# https://ethereum.github.io/eth2.0-APIs/#/`

to

> `# https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/`

In theory, it might be interesting to specifically target https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.0.0 or https://ethereum.github.io/beacon-APIs/?urls.primaryName=v1 but those both yield
> Failed to load API definition.


cheatfate:
This is part of my https://github.com/status-im/nimbus-eth2/pull/2790

PR #2807, opened 2021-08-23
eth2.0-spec-tests -> consensus-spec-tests repo rename

tersec:


arnetheduck:
as part of changes like this, it would be good to get rid of generic terms like "official" and instead use "EF" or whatever the source is

PR #2809, opened 2021-08-23
cleanups

arnetheduck:
* use ForkedTrustedSignedBeaconBlock.ionit where appropriate
* move `is_aggregator` to `spec/`
* use `errReject` in a few more places
* update enr fork id when time is auspicious
* use network broadcast functions


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -467,10 +463,10 @@ proc validateAggregate*(
     var x = await cryptoFuts.slotCheck
     case x
     of BatchResult.Invalid:
-      return errReject(cstring("validateAggregate: invalid slot signature"))
+      return errReject("validateAggregate: validateAggregate: invalid slot signature")
     of BatchResult.Timeout:
       beacon_aggregates_dropped_queue_full.inc()
-      return err((ValidationResult.Reject, cstring("validateAggregate: timeout checking slot signature")))
+      return errReject("validateAggregate: timeout checking slot signature")
```
tersec: If one wants a helper function called `errReject` which does responds to timeouts as a kind of refactoring, I'm fine with that. It shouldn't be the same one that hard-errors-out in debugging/CI mode.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -482,7 +478,7 @@ proc validateAggregate*(
       return errReject("validateAggregate: invalid aggregator signature")
     of BatchResult.Timeout:
       beacon_aggregates_dropped_queue_full.inc()
-      return err((ValidationResult.Reject, cstring("validateAggregate: timeout checking aggregator signature")))
+      return errReject("validateAggregate: timeout checking aggregator signature")
```
tersec: errReject/timeout

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -494,7 +490,7 @@ proc validateAggregate*(
       return errReject("validateAggregate: invalid aggregate signature")
     of BatchResult.Timeout:
       beacon_aggregates_dropped_queue_full.inc()
-      return err((ValidationResult.Reject, cstring("validateAggregate: timeout checking aggregate signature")))
+      return errReject("validateAggregate: timeout checking aggregate signature")
```
tersec: errReject/timeout

beacon_chain/gossip_processing/gossip_validation.nim
tersec: This is sort of one reason I initially named it something more aggressive in the PR, precisely that it wasn't meant as some anodyne typing shortcut for `ValidationResult.Reject`.

arnetheduck: mmyeah, it would probably make sense to introduce a "shortcut" `errReject` and rename the current one to something slightly more telling - `errIgnore` being added puts the two in an equivalence relationship.

anyway, pretty much for the same reason, we shouldn't be rejecting on timeout - attestations use ignore and so should aggregate - it's our fault, not theirs, basically, so ignore is appropriate.

tersec: Changing the timeout handling to ignore rather than reject would sidestep this at the moment, and I agree would make sense regardless.

Even so, it might make sense to clarify the name of `errReject`, along the lines of what
https://github.com/status-im/nimbus-eth2/blob/e9c742c9cc817ffa40de078a803231beb21355a1/beacon_chain/gossip_processing/gossip_validation.nim#L158-L160
describes. The "never has false positives" part is important, and including timeouts undermines that.

arnetheduck: renamed to `checkedReject` - `errReject` becomes the shortcut helper

arnetheduck: 3efdaca912360cc393c9523949856ceb2f332fd7

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: 3efdaca912360cc393c9523949856ceb2f332fd7

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: 3efdaca912360cc393c9523949856ceb2f332fd7

tersec (approved):

PR #2813, opened 2021-08-25
Validator index reform

arnetheduck:
This is a work-in-progress  from the altair branch to reform validator
indices - in review, the changes have been found wanting due to the
difficulty of using type to communicate lifetime when the lifetime is
tied to state and not a static property of the type.

A more accurate change would be to introduce a `CachedValidatorIndex`
that sits at a `dag` level, but that would require refactoring the
current changes to not touch the state, whose validator index validity
works in a different way than the cache preventing these changes from
beeing merged in their current state, so as not to introduce confusion
in the spec code.

The changes also need to be reviewed for tree hash cache performance
regressions (calling the right `[]` etc).

The other change discussed in this context has been to harden the code
against silly values of CommitteeIndex and ValidatorIndex using their
natural / static range limits from the spec constants - this would be a
separate change to the reform in this branch. In particular, the notion
of a "valid" `CommitteeIndex` in this branch needs to be revisited (as
the validity depends on the state, not the lifetime of a program).

See also
https://github.com/status-im/nimbus-eth2/pull/2709#discussion_r686246957

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -8,7 +8,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[typetraits, tables],
+  std/[typetraits, tables, hashes],
   stew/[arrayops, assign2, byteutils, endians2, io2, objects, results],
   serialization, chronicles, snappy,
   eth/db/[kvstore, kvstore_sqlite3],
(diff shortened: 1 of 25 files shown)


arnetheduck:
the latest version of the validator index changes, as done in the altair branch, are also available in https://github.com/status-im/nimbus-eth2/tree/altair-with-valindex

arnetheduck:
Superceded by #3651

PR #2821, opened 2021-08-25
all: remove pyrmont

tqpcharlie:
TODO:
- [x] Remove run-pyrmont-beacon-node.sh
- [x] Remove pyrmont in the beacon book
- [x] Remove any extra configs


arnetheduck:
Thanks! One thing we'll want to do when we  pull the trigger is to include a small help text in the book explaining that pyrmont has been deprecated and is no longer supported, so that users that see it on the internet know what's going on.

Also, the first step to deprecate it will be the user manual and help texts - we'll keep the support in the binary for a while since it aids testing - ergo, a first set of patches would focus on those parts.

arnetheduck:
Hey @Zachinquarantine - we're ready to make a move on this, if you're able to rebase the PR that would be appreciated.

arnetheduck:
#3568 updates this PR to the latest `unstable`, together with some other testnet-related changes

PR #2832, opened 2021-08-29
subscribe to/unsubscribe from sync committee subnets

tersec:
This doesn't individually track the 4 sync committee subnets, but subscribes and unsubscribes to them all. It's enough to run Pyrmont and Prater for now, pending a more complete implementation.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -793,8 +793,29 @@ proc removePhase0MessageHandlers(node: BeaconNode) =
 proc addAltairMessageHandlers(node: BeaconNode, slot: Slot) =
   node.addPhase0MessageHandlers(node.dag.forkDigests.altair)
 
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -793,9 +793,30 @@ proc removePhase0MessageHandlers(node: BeaconNode) =
 proc addAltairMessageHandlers(node: BeaconNode, slot: Slot) =
   node.addPhase0MessageHandlers(node.dag.forkDigests.altair)
```
tersec, line 794: To copy over a comment from `altair` PR from @arnetheduck:
> I imagine we should have a type Topics = object; blocks: string; aggregates: string etc or something like that in eth2_network that has an init that takes a forkdigest instead of calling the spec function every time with a digest and generating the string anew

tersec: And, yes, something along those lines would be reasonable, but isnt a priority in terms of merging at the moment.

PR #2836, opened 2021-08-30
Restore the sync committee pool pruning and add tests

zah:


beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
```diff
@@ -15,30 +15,60 @@ import
...
-  # clear pool.blockVotes
+func pruneData*(pool: var SyncCommitteeMsgPool, slot: Slot) =
+  ## This should be called at the end of slot.
+  clear pool.seenContributionByAuthor
+  clear pool.seenSyncMsgByAuthor
+
+  if slot < SyncCommitteeMsgsRetentionSlots:
+    return
+
+  let minSlotToRetain = slot - SyncCommitteeMsgsRetentionSlots
```
tersec: What changed to make this viable again?

tests/all_tests.nim
```diff
@@ -26,6 +26,7 @@ import # Unit test
   ./test_message_signatures,
   ./test_peer_pool,
   ./test_statediff,
+  ./test_sync_committee_pool,
```
tersec, line 29: Would be good to rerun `make test` locally so that the markdown files get updated in the PR at the same time

tests/test_sync_committee_pool.nim
```diff
@@ -0,0 +1,218 @@
...
+        outContribution)
+
+      check:
+        not success
+
+      let aggregate = pool.produceSyncAggregate(root2)
+      check:
+        aggregate.sync_committee_bits.isZeros
+        aggregate.sync_committee_signature == ValidatorSig.infinity
+
```
tersec: Trailing newline

beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
zah: `SyncCommitteeMsgsRetentionSlots` was effectively 1 before and we were pruning some of the message while they were still viable for inclusion in certain situations. Now, it's set to 3 slots which is more conservative.

PR #2837, opened 2021-08-30
discovery: don't filter peers during sync

stefantalpalaru:
We don't care about peer attestation subnet subscriptions during block sync.

The ugly bit here is informing that part of the code that calls `queryRandom()` of the block sync status.

TODO: keep a percentage of unfiltered peers in our peer pool so we can discover the mesh layout quicker and more reliably. What's useless for attestation propagation can be useful for discovery.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -901,7 +901,7 @@ proc toPeerAddr(node: Node): Result[PeerAddr, cstring] =
   ok(peerAddr)
 
 proc queryRandom*(d: Eth2DiscoveryProtocol, forkId: ENRForkID,
-    attnets: BitArray[ATTESTATION_SUBNET_COUNT], syncInProgress = false):
+    attnets: BitArray[ATTESTATION_SUBNET_COUNT]):
     Future[seq[Node]] {.async.} =
   ## Perform a discovery query for a random target
   ## (forkId) and matching at least one of the attestation subnets.
@@ -909,32 +909,29 @@ proc queryRandom*(d: Eth2DiscoveryProtocol, forkId: ENRForkID,
   let eth2Field = SSZ.encode(forkId)
   let nodes = await d.queryRandom((enrForkIdField, eth2Field))
 
-  if syncInProgress:
-    return nodes
-  else:
-    var filtered: seq[(int, Node)]
-    for n in nodes:
-      let res = n.record.tryGet(enrAttestationSubnetsField, seq[byte])
+  var filtered: seq[(int, Node)]
+  for n in nodes:
+    let res = n.record.tryGet(enrAttestationSubnetsField, seq[byte])
 
-      if res.isSome():
-        let attnetsNode =
-          try:
-            SSZ.decode(res.get(), BitArray[ATTESTATION_SUBNET_COUNT])
-          except SszError as e:
-            debug "Could not decode attestation subnet bitfield of peer",
-              peer = n.record.toURI(), exception = e.name, msg = e.msg
-            continue
+    if res.isSome():
+      let attnetsNode =
+        try:
+          SSZ.decode(res.get(), BitArray[ATTESTATION_SUBNET_COUNT])
+        except SszError as e:
+          debug "Could not decode attestation subnet bitfield of peer",
+            peer = n.record.toURI(), exception = e.name, msg = e.msg
+          continue
 
-        var score: int = 0
-        for i in 0..<ATTESTATION_SUBNET_COUNT:
-          if attnets[i] and attnetsNode[i]:
-            inc score
+      var score: int = 0
+      for i in 0..<ATTESTATION_SUBNET_COUNT:
+        if attnets[i] and attnetsNode[i]:
+          inc score
 
-        if score > 0:
-          filtered.add((score, n))
+      if score > 0:
+        filtered.add((score, n))
 
-    d.rng[].shuffle(filtered)
-    return filtered.sortedByIt(it[0]).mapIt(it[1])
+  d.rng[].shuffle(filtered)
+  return filtered.sortedByIt(it[0]).mapIt(it[1])
 
 proc trimConnections(node: Eth2Node, count: int) {.async.} =
   # Kill `count` peers, scoring them to remove the least useful ones
@@ -1041,7 +1038,11 @@ proc runDiscoveryLoop*(node: Eth2Node) {.async.} =
       wantedAttnetsCount = wantedAttnets.countOnes()
 
     if wantedAttnetsCount > 0:
-      let discoveredNodes = await node.discovery.queryRandom(node.forkId, wantedAttnets, node.syncInProgress)
+      let discoveredNodes = if node.syncInProgress:
+        # Don't filter peers by attestation subnet subscription.
+        await node.discovery.queryRandom((enrForkIdField, SSZ.encode(node.forkId)))
+      else:
+        await node.discovery.queryRandom(node.forkId, wantedAttnets)
 
       var newPeers = 0
       for discNode in discoveredNodes:
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -82,6 +82,7 @@ type
     peerPingerHeartbeatFut: Future[void]
     cfg: RuntimeConfig
     getBeaconTime: GetBeaconTimeFn
+    syncInProgress*: bool
 
   EthereumNode = Eth2Node # needed for the definitions in p2p_backends_helpers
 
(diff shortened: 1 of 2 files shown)


Menduist:
That should not be required.

Yes, the peer discovery is a _tad_ slower than before because of peer cycling, but even during sync, we'll connect to every peer that is subscribed to at least one stability subnet, (so almost every peer)

Also note that after peer cycling, we'll only connect to peers as long as we have to, so you may plateau after a certain number of outgoing connections (~30), that is normal 

Please let me know if you actually see better performance with this branch, but it should be the same

The only actual issue, as discussed on discord, is that as soon as nimbus is configured for Altair, it will only connect to Altair ready nodes, thus slowing down discovery **a lot**. You only rely on incoming at this point.

kdeme (commented):
> Yes, the peer discovery is a tad slower than before because of peer cycling, but even during sync, we'll connect to every peer that is subscribed to at least one stability subnet, (so almost every peer)

Yes, it will cause the peer connections to increase slower. But you might be right, it might be minimal (perhaps unmeasurable even in the margin?) and thus the question is whether it really matters for the speed of syncing, but that should be tested/measured.
A peer that is syncing also only needs the amount of peers (or a few more to be good) that it will in parallel download blocks from & process. 

However, I do have 2 two questions about the current implementation and where this PR might matter more:
- The `getLowAttnets` call does provide already the required attnets for the validators the node is running with when the node is not synced yet?
- When we have a node with no validators attached, and no `--subscribe-all-subnets` set, will it still find peers?

Menduist:
The condition of `getLowAttnets` which is used during sync (or with no validator attached) is:
`Have 0 subnet with < 'd' peers from topic subscription`

So it will try to dial enough peer to have at least `d` (8) peers subscribed in each subnet.
If we are very lucky, we'll only dial 8 peers, but in reality, we'll dial at least 20 to have enough coverage

EDIT: to run `unstable` without the altair switch issue, you can try:
```sh
cd vendor/eth2-networks/
git checkout 9eeb9e33e10e9deb1439dd6aa4563c1d73c85a63
```
This will remove the altair info from the prater network, and thus give the correct fork_id.
Just tried it, after 3 minutes I had 20 peers with 12 outgoing & 8 incoming. (during sync)

stefantalpalaru:
> we'll connect to every peer that is subscribed to at least one stability subnet, (so almost every peer)

Are we sure about that? Here's the peer count during and after initial block sync, when using this branch:

Screenshot 2021-08-31 at 13-32-28 NBC local testnet sim (all nodes) - Grafana

Menduist:
Recap:
All theses connections, including the big dip on sync are incoming connections.
Because of the fork id issue, we are currently only "compatible" with teku nodes, until altair arrives on prater.

It looks like teku nodes like to drop us once we are in sync for some reason, hence the big dip
At least 90% of peers are subscribed to at least one subnet, so worst case scenario, peer cycling will slow down discovery by 10%.
The fork id issue however, blocks us from connecting to anyone except teku when a fork is about to happen. So that probably slows discovery by 85%.

Now, should we keep the behaviour to only connect to peers with stability subnets during sync?
I would say yes, because even though is slows down discovery a tiny bit, it probably doesn't matter in the scale of things, and allows us to be sure to have a healthy mesh as soon as the sync is finished.

EDIT: for reference, a graph without the fork id issue (with eth2-networks at `9eeb9`):
image
We can see that the outgoing connections stop as soon as the mesh is healthy, which is normal.
Around 16:25, we are in sync, so teku nodes drop us, but since we are connected to other clients as well, the dip is not as bad

stefantalpalaru:
Syncing Prater from scratch for an hour, first with "unstable", then deleting the db and syncing again with this branch:

Screenshot 2021-08-31 at 19-20-04 NBC local testnet sim (all nodes) - Grafana

But the biggest difference is in average sync speed: 0.008 blocks/s with "unstable" and 4 blocks/s with this branch.

What I noticed in the first few minutes is that with this branch I connect to a very reliable peer that feeds me blocks at maximum speed and I stay connected to it while searching for other peers that can deal with my networking habits. That gives me a high syncing speed from the start.

The "open streams" panel, in case it matters:

Screenshot 2021-08-31 at 19-26-35 NBC local testnet sim (all nodes) - Grafana

stefantalpalaru:
Another "unstable" run, to confuse things:

Screenshot 2021-08-31 at 21-15-54 NBC local testnet sim (all nodes) - Grafana

Good average syncing speed too: 4.4 blocks/s.

Menduist:
My _guess_ is that prysm switched to altair (and thus the new fork id) around 6pm, and caused a bump on peer count. I can also see it on my server which was running unstable 

Here a three runs (unstable, this branch, unstable). At this point I feel like I'm watching the stock market. The dip in the last one is probably an other weird network condition
Screenshot_2021-08-31_22-12-17

Anyway, I don't want to hold up people switching to unstable, so if you still feel like merging this you can. But between our last runs, my gut feel, and the fact that the differences are only on the incoming connections and this PR shouldn't affect incoming connections, I still don't think it's a good idea.

I'll leave my review, in case you still want to merge it

beacon_chain/networking/eth2_network.nim
```diff
@@ -1037,7 +1041,7 @@ proc runDiscoveryLoop*(node: Eth2Node) {.async.} =
       wantedAttnetsCount = wantedAttnets.countOnes()
 
     if wantedAttnetsCount > 0:
-      let discoveredNodes = await node.discovery.queryRandom(node.forkId, wantedAttnets)
+      let discoveredNodes = await node.discovery.queryRandom(node.forkId, wantedAttnets, node.syncInProgress)
```
Menduist: You should be able to do something like
```nim
      let discoveredNodes = if node.syncInProgress:
          await node.discovery.queryRandom((enrForkIdField, SSZ.encode(forkId)))
        else:
          await node.discovery.queryRandom(forkId, wantedAttnets)
```
And leave the queryRandom above as is

stefantalpalaru: Done.

Menduist:
I'll explain peer cycling in a bit more detail and address a few comments

### Peer cycling
The primary goal of peer cycling is to have a good mesh, no matter what `--max-peers` is. Obviously, not an issue with `--max-peers=160`, but trickier with `--max-peers=20`.
We define a good mesh as:
- Every subscribed topic is healthy (at least `dLow` (6) peers, at least `dOut` (3) outgoing peers)
- We have at least `d` (8) connected peers in every topic (used for fanout etc)

`getLowAttnets` returns the topic we have to find peers for, in order to reach this good mesh.

If `getLowAttnets` returns some topic, we are then unhealthy, and should dial new peers having theses attnets, even if that means kicking existing peers to do so.
If we are healthy however, we have no reasons to continue dialing, since that will just consume more resources for no added benefit. (Obviously, currently, we'll get incoming connections instead, but in the future the goal would be to also limit incoming connections in that situation)

This strategy works well with as low as 20 peers, and is also good for the network as a whole (we have a lot of incoming slots available, and don't dial too much)

Now, should we also try to have a good mesh during sync? That's up to debate, but I would say yes, because this way, as soon as we're in sync, we can start attesting right away with a healthy mesh. And as said earlier, the filtering does slow down the discovery process by 10%ish. But if you have a few blocks to sync, it won't matter because that will be fast anyway, and if you have a lot of blocks to sync, it'll take hours no matter what

### Comments
> TODO: keep a percentage of unfiltered peers in our peer pool so we can discover the mesh layout quicker and more reliably. What's useless for attestation propagation can be useful for discovery.

As mentioned elsewhere, the discovery process is actually completely independent of nimbus and libp2p.
It's located here: https://github.com/status-im/nim-eth/tree/master/eth/p2p/discoveryv5
and is not at all influenced by the number of connected peers, since it's fully UDP. BTW, the DHT used is not even specific to ETH2, it's shared by multiple applications

> There's something rotten in Denmark if we discard peers we connect to, but accept peers that connect to us.

When we dial a peer, we'll potentially kick another peer to make room for the new one. That means that we have to be sure that the new peer is a good one, otherwise we might kick a peer for no reason.
This is not the case with incoming peers, if someone dials us and we're full, we'll simply ignore him.

And obviously, we can choose who we dial, but not who dials us. If there was a way to filter incoming peers, I would happily do it :D
(And in a way it happens when we kick a peer to make room for a new dial, we'll kick the lowest-scored peers)



Anyway, I'm not against this PR per-se. I just wanted to show that the slow peer discovery is not caused by this, and that something else was at play. I think we both agree at this point that this was a combo of altair switching and random network conditions, after all this tedious troubleshooting.

EDIT: We also talked about doing a completely different peer cycling technique during sync with Jacek, but left it for later

arnetheduck:
@kdeme discovered another limitation in our ability to make concurrent outgoing connections: we do 10 concurrent connections - doubling that significantly speeds up initial connection acquisition.

arnetheduck:
https://github.com/status-im/nimbus-eth2/pull/2966 speeds up connectivity more generically - there are also other ideas about improving peer search, such as not waiting for all answers when doing the query, but  instead letting each worker make one query then dial its results (cc @kdeme )

arnetheduck:
Connectivity and peer cycling has been addressed in a different way - this PR is obsolete as much of the supporting data comes from pre-altair times, and before the current peer cycling work.

PR #2851, opened 2021-09-08
SSZ encoded responses for REST API calls.

cheatfate:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -692,6 +709,85 @@ proc writeValue*(writer: var JsonWriter[RestJson],
...
+  case value.beaconStateFork
+  of BeaconStateFork.forkPhase0:
+    writer.writeField("version", "phase0")
+    writer.writeField("data", value.bsPhase0)
+  of BeaconStateFork.forkAltair:
+    writer.writeField("version", "altair")
+    writer.writeField("data", value.bsAltair)
+  writer.endRecord()
+
+template toSszType*(v: BeaconBlockFork): auto =
```
zah, line 781: Are these used anywhere? Perhaps they are leftovers from the earlier idea of including discriminators in the response?

Issue #2810, opened 2021-08-23
nimbus-eth2 with Pyrmont: Non-altair client on altair network with non-altair blocks saved in Database


hydrodan:
Accidentally pulled and rebuilt the main git repo instead of the altair-beta tagged branch last week, and now my altair client isn't able to run, nimbus-eth2 crashed with this error:

```
Aug 20 14:54:57 bash[22937]: ../csu/libc-start.c(308)
Aug 20 14:54:57 bash[22937]: Error: unhandled exception: /home/pi/git/nimbus-eth2/beacon_chain/consensus_object_pools/blockchain_dag.nim(898, 14) ok Blocks in database should never fail to apply.. [AssertionError]
```

Running "nimbus_beacon_node --version" reports:
```
Nimbus beacon node v1.4.1-07a1c5-stateofus
Copyright (c) 2019-2021 Status Research & Development GmbH

eth2 specification v1.0.1

Nim Compiler Version 1.2.12 [Linux: arm64]
Copyright (c) 2006-2020 by Andreas Rumpf

git hash: 121628357ec7fae91335bd392f03b0e6dd249319
active boot switches: -d:release

```

This is built and running on Raspberry Pi 4 in arm64 configuration. I've also tried with local GETH for the ETH1 node and Infura, and same reaction so it appears to be the Nimbus build having the issues. (also tried the pre-built arm64 binary after all this, but it's giving me the same error now)

tersec:
To add some detail, this occurs in Pyrmont when applying the first post-fork block to the pre-fork state:
```
state_transition: process_block failed     topics="chaindag" tid=2142181 file=state_transition.nim:252 blck="(slot: 1972807, proposer_index: 118850, parent_root: \"abdb57d8\", state_root: \"4005d6cf\", eth1data: (deposit_root: a6000815801fdf3741cda8f9344bd19365ef6f7d0885aab521b11b77279fdbd2, deposit_count: 129040, block_hash: 8e1c7687971f73d3e33af1f7d14bbeaec058a5ce7e2126db104cb75f6479553c), graffiti: \"Unobtanium\", proposer_slashings_len: 0, attester_slashings_len: 0, attestations_len: 128, deposits_len: 0, voluntary_exits_len: 0, sync_committee_participants: -1)" slot=1972807 eth1_deposit_index=129025 deposit_root=1de90f29 error="process_block: Altair state with Phase 0 block"
[...]

PR #2852, opened 2021-09-08
roll back putative blocks from invalid hardforks

tersec:
Addresses https://github.com/status-im/nimbus-eth2/issues/2810

It removes the invalid blocks up-front, before anything else in the processing pipeline knows they exist.

Once the next head update occurs, the transition period naturally ends.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -789,10 +789,15 @@ proc getEth2FinalizedTo*(db: BeaconChainDB): Opt[DepositContractSnapshot] =
 proc containsBlock*(db: BeaconChainDBV0, key: Eth2Digest): bool =
   db.backend.contains(subkey(phase0.SignedBeaconBlock, key)).expectDb()
 
-proc containsBlock*(db: BeaconChainDB, key: Eth2Digest): bool =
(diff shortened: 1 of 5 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -789,11 +789,16 @@ proc getEth2FinalizedTo*(db: BeaconChainDB): Opt[DepositContractSnapshot] =
...
-    db.blocks.contains(key.data).expectDb() or
+proc containsBlockPhase0*(db: BeaconChainDB, key: Eth2Digest): bool =
+  db.blocks.contains(key.data).expectDb() or
     db.v0.containsBlock(key)
 
+proc containsBlockAltair*(db: BeaconChainDB, key: Eth2Digest): bool =
+  db.altairBlocks.contains(key.data).expectDb()
+
+proc containsBlock*(db: BeaconChainDB, key: Eth2Digest): bool =
+  db.containsBlockPhase0(key) or db.containsBlockAltair(key)
```
arnetheduck, line 800: generally, I think we should be testing the "newest" fork first - recent blocks are more likely to be checked for, all other things being equal

tersec: https://github.com/status-im/nimbus-eth2/pull/2891

PR #2855, opened 2021-09-08
Light client sync spec functions

zah:


beacon_chain/spec/beaconstate.nim
```diff
@@ -859,3 +839,5 @@ proc upgrade_to_altair*(cfg: RuntimeConfig, pre: phase0.BeaconState): ref altair
   post[].next_sync_committee = get_next_sync_committee(post[])
 
   post
+
+
```
tersec: Trailing newlines

beacon_chain/spec/helpers.nim
```diff
@@ -213,3 +233,8 @@ func add_flag*(flags: ParticipationFlags, flag_index: int): ParticipationFlags =
 func has_flag*(flags: ParticipationFlags, flag_index: int): bool =
   let flag = ParticipationFlags(1'u8 shl flag_index)
   (flags and flag) == flag
+
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-beta.3/specs/altair/sync-protocol.md#get_subtree_index
+func get_subtree_index*(idx: GeneralizedIndex): uint64 =
+  uint64(idx mod (type(idx)(1) shl log2trunc(idx)))
+
```
tersec: Trailing newline

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -519,3 +525,7 @@ func shortLog*(v: SyncAggregate): auto =
   $(v.sync_committee_bits)
 
 chronicles.formatIt SyncCommitteeMessage: shortLog(it)
+
+func hash*(x: LightClientUpdate): Hash =
+  hash(x.header.state_root.data)
+
```
tersec: Trailing newline

beacon_chain/spec/helpers.nim
```diff
@@ -47,6 +47,26 @@ template epoch*(slot: Slot): Epoch =
 template isEpoch*(slot: Slot): bool =
   (slot mod SLOTS_PER_EPOCH) == 0
 
+# https://github.com/ethereum/consensus-specs/blob/v1.0.1/specs/phase0/beacon-chain.md#is_valid_merkle_branch
+func is_valid_merkle_branch*(leaf: Eth2Digest, branch: openArray[Eth2Digest],
+                             depth: int, index: uint64,
+                             root: Eth2Digest): bool =
```
etan-status, line 53: Intended that `{.nbench.}` got dropped here?

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -519,3 +525,7 @@ func shortLog*(v: SyncAggregate): auto =
   $(v.sync_committee_bits)
 
 chronicles.formatIt SyncCommitteeMessage: shortLog(it)
+
+func hash*(x: LightClientUpdate): Hash =
+  hash(x.header.state_root.data)
```
etan-status: For now, probably good enough, but maybe more future-proof & readable with additional reusable `hash` implementations for `Eth2Digest` and `BeaconBlockHeader` so that this just becomes `hash(x.header)`.

beacon_chain/spec/helpers.nim
```diff
@@ -213,3 +233,8 @@ func add_flag*(flags: ParticipationFlags, flag_index: int): ParticipationFlags =
 func has_flag*(flags: ParticipationFlags, flag_index: int): bool =
   let flag = ParticipationFlags(1'u8 shl flag_index)
   (flags and flag) == flag
+
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-beta.3/specs/altair/sync-protocol.md#get_subtree_index
+func get_subtree_index*(idx: GeneralizedIndex): uint64 =
+  uint64(idx mod (type(idx)(1) shl log2trunc(idx)))
```
etan-status, line 240: LGTM, also w.r.t. edge case `uint32.high`.

etan-status: To avoid the division, could optimize this to use bit-magic to clear the most significant bit.
(not sure if divisions are expensive on target platforms though).

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+  stew/bitops2,
+  datatypes/altair,
+  helpers
+
+func branchIsAllZeros(branch: openarray[Eth2Digest]): bool =
+  for node in branch:
+    if node != Eth2Digest():
+      return false
+
+  return true
```
etan-status:
```
import sequtils

func branchIsAllZeros(branch: openarray[Eth2Digest]): bool =
  const ZeroDigest = Eth2Digest()
  branch.allIt it == ZeroDigest
```

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.3/specs/altair/sync-protocol.md#validate_light_client_update
+proc validate_light_client_update*(snapshot: LightClientSnapshot,
+                                   update: LightClientUpdate,
+                                   genesis_validators_root: Eth2Digest): bool =
+  # Verify update slot is larger than snapshot slot
+  if update.header.slot <= snapshot.header.slot:
+    return false
```
etan-status, line 13: Note: Lodestar chose to skip this check.
https://github.com/ChainSafe/lodestar/blob/master/packages/light-client/src/client/validation.ts#L24

beacon_chain/spec/helpers.nim
```diff
@@ -213,3 +233,8 @@ func add_flag*(flags: ParticipationFlags, flag_index: int): ParticipationFlags =
 func has_flag*(flags: ParticipationFlags, flag_index: int): bool =
   let flag = ParticipationFlags(1'u8 shl flag_index)
   (flags and flag) == flag
+
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-beta.3/specs/altair/sync-protocol.md#get_subtree_index
+func get_subtree_index*(idx: GeneralizedIndex): uint64 =
```
etan-status, line 238: Maybe add an `assert idx > 0` here?

`log2trunc` does not match the EF `floorlog2` definition regarding 0 handling.

EF spec:
```
def floorlog2(x: int) -> uint64:
    if x < 1:
        raise ValueError(f"floorlog2 accepts only positive values, x={x}")
    return uint64(x.bit_length() - 1)
```

stew returns `-1` instead, leading to division by 0 here and crashing the program with `SIGFPE: Arithmetic error.`:
```
  if x == 0: -1
  else:
    when nimvm:
      log2truncNim(x)
    else:
      when declared(log2truncBuiltin):
        log2truncBuiltin(x)
      else:
        log2truncNim(x)
```

beacon_chain/spec/helpers.nim
```diff
@@ -47,6 +47,26 @@ template epoch*(slot: Slot): Epoch =
...
+
+  for i in 0 ..< depth:
+    if (index div (1'u64 shl i)) mod 2 != 0:
+      buf[0..31] = branch[i].data
+      buf[32..63] = value.data
+    else:
+      buf[0..31] = value.data
+      buf[32..63] = branch[i].data
+    value = eth2digest(buf)
+  value == root
```
etan-status, line 68: Returns `true` if `depth < 0`. Intended?

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+  # Verify update does not skip a sync committee period
+  var snapshot_period = compute_epoch_at_slot(snapshot.header.slot) div EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+  var update_period = compute_epoch_at_slot(update.header.slot) div EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+  if update_period notin [snapshot_period, snapshot_period + 1]:
+    return false
+
+  # Verify update header root is the finalized root of the finality header, if specified
+  # TODO: Use a view type instead of `unsafeAddr`
+  let signed_header = if update.finality_header == BeaconBlockHeader():
+    if not branchIsAllZeros(update.finality_branch):
```
etan-status: Difference to spec intentional?

The spec also verifies that the branch has expected length, but here we just check that the branch is all zeroes without validating its length.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+                                  log2trunc(FINALIZED_ROOT_INDEX),
+                                  get_subtree_index(FINALIZED_ROOT_INDEX),
+                                  update.finality_header.state_root):
+      return false
+    unsafeAddr update.finality_header
+
+  # Verify update next sync committee if the update period incremented
+  # TODO: Use a view type instead of `unsafeAddr`
+  let sync_committee = if update_period == snapshot_period:
+    if not branchIsAllZeros(update.next_sync_committee_branch):
```
etan-status: Difference to spec intentional?

The spec also verifies that the branch has expected length, but here we just check that the branch is all zeroes without validating its length.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+  # Verify sync committee has sufficient participants
+  if countOnes(update.sync_committee_bits) < MIN_SYNC_COMMITTEE_PARTICIPANTS:
+    return false
+
+  # Verify sync committee aggregate signature
+  # participant_pubkeys = [pubkey for (bit, pubkey) in zip(update.sync_committee_bits, sync_committee.pubkeys) if bit]
+  var participant_pubkeys: seq[ValidatorPubKey]
+  for idx, bit in update.sync_committee_bits:
+    if bit:
+      participant_pubkeys.add(sync_committee.pubkeys[idx])
```
etan-status: * Initialize seq with correct capacity
* Reduce visual noise and match spec by also using list comprehension (sugar)
* Remove commented spec code

```
import sugar

  # Verify sync committee has sufficient participants
  let sync_committee_participants_count = countOnes(update.sync_committee_bits)
  if sync_committee_participants_count < MIN_SYNC_COMMITTEE_PARTICIPANTS:
    return false

  # Verify sync committee aggregate signature
  let participant_pubkeys = collect(newSeqOfCap(sync_committee_participants_count)):
    for idx, bit in update.sync_committee_bits:
      if bit: sync_committee.pubkeys[idx]
```

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+  for node in branch:
+    if node != Eth2Digest():
+      return false
+
+  return true
+
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.3/specs/altair/sync-protocol.md#validate_light_client_update
+proc validate_light_client_update*(snapshot: LightClientSnapshot,
+                                   update: LightClientUpdate,
+                                   genesis_validators_root: Eth2Digest): bool =
```
etan-status, line 10: Would it be useful to add debug logs when an update is rejected?
Not for production (as it would enable others to spam the log) but for development.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+
+  blsFastAggregateVerify(participant_pubkeys, signing_root.data, update.sync_committee_signature)
+
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-beta.3/specs/altair/sync-protocol.md#apply_light_client_update
+proc apply_light_client_update(snapshot: var LightClientSnapshot, update: LightClientUpdate) =
+  let snapshot_period = compute_epoch_at_slot(snapshot.header.slot) div EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+  let update_period = compute_epoch_at_slot(update.header.slot) div EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+  if update_period == snapshot_period + 1:
+      snapshot.current_sync_committee = snapshot.next_sync_committee
+      snapshot.next_sync_committee = update.next_sync_committee
```
etan-status: NIT: Indentation here is 4 spaces but should be 2 spaces for consistency.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+                                 current_slot: Slot,
+                                 genesis_validators_root: Eth2Digest): bool =
+  if not validate_light_client_update(store.snapshot, update, genesis_validators_root):
+    return false
+  store.valid_updates.incl(update)
+
+  var update_timeout = SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+  let sync_committee_participants_count = countOnes(update.sync_committee_bits)
+  if sync_committee_participants_count * 3 >= update.sync_committee_bits.len * 2 and
+     update.finality_header != BeaconBlockHeader():
```
etan-status: Would it make sense to define a `const ZeroBeaconBlockHeader` globally instead of allocating a fresh one at multiple locations throughout the code? Not sure how well Nim optimizes this.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+    # Apply update if (1) 2/3 quorum is reached and (2) we have a finality proof.
+    # Note that (2) means that the current light client design needs finality.
+    # It may be changed to re-organizable light client design. See the on-going issue eth2.0-specs#2182.
```
etan-status, line 92: `eth2.0-specs` got renamed to `consensus-specs` upstream.
Note: This would also affect ALL of the spec reference links. Maybe better to keep consistent for now and do the rename in one go globally without other changes to reduce noise.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+    apply_light_client_update(store.snapshot, update)
+    store.valid_updates.clear()
+  elif current_slot > store.snapshot.header.slot + update_timeout:
+    var best_update_participants = 0
+    var best_update: LightClientUpdate
+    for update in store.valid_updates:
+      let update_participants = countOnes(update.sync_committee_bits)
+      if update_participants > best_update_participants:
+        best_update = update
+        best_update_participants = update_participants
```
etan-status: Maybe, just inside this file, could define `<` for `LightClientUpdate` (Nim does not seem to support `max` with lambda predicates).

```
func `<`(x, y: SyncAggregate): bool =
  countOnes(x.sync_committee_bits) < countOnes(y.sync_committee_bits)
```
(needs to be at top level, or `max` does not detect it)

and then, this entire loop here becomes:
```
      let best_update = store.valid_updates.max
```

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+  var update_timeout = SLOTS_PER_EPOCH * EPOCHS_PER_SYNC_COMMITTEE_PERIOD
+  let sync_committee_participants_count = countOnes(update.sync_committee_bits)
+  if sync_committee_participants_count * 3 >= update.sync_committee_bits.len * 2 and
+     update.finality_header != BeaconBlockHeader():
+    # Apply update if (1) 2/3 quorum is reached and (2) we have a finality proof.
+    # Note that (2) means that the current light client design needs finality.
+    # It may be changed to re-organizable light client design. See the on-going issue eth2.0-specs#2182.
+    apply_light_client_update(store.snapshot, update)
+    store.valid_updates.clear()
+  elif current_slot > store.snapshot.header.slot + update_timeout:
```
etan-status, line 95: Maybe an assert to verify that `store.valid_updates` contains at least one candidate update. Otherwise, this will try to apply an invalid update.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -0,0 +1,113 @@
...
+    for update in store.valid_updates:
+      let update_participants = countOnes(update.sync_committee_bits)
+      if update_participants > best_update_participants:
+        best_update = update
+        best_update_participants = update_participants
+
+    # Forced best update when the update timeout has elapsed
+    apply_light_client_update(store.snapshot, best_update)
+    store.valid_updates.clear()
+  return true
```
etan-status: NIT: Would just `true` without the explicit `return` be more consistent with the rest of this code?

beacon_chain/ssz/merkleization.nim
```diff
@@ -32,6 +32,8 @@ func binaryTreeHeight*(totalElements: Limit): int =
   bitWidth nextPow2(uint64 totalElements)
 
 type
+  GeneralizedIndex* = uint32
```
etan-status, line 37: In the spec's `setup.py`, this is implemented as a standard `int` with machine-dependent width (`GeneralizedIndex = NewType('GeneralizedIndex', int)`).

Given that `get_subtree_index` forms a `uint64` out of the `generalized_index`, would suggest going with `uint64` here as well (otherwise, there is no way how `generalized_index` could ever exceed `uint32.high`.

etan-status:
Done with my initial review.

beacon_chain/spec/light_client_sync.nim
tersec: I generally periodically clean these up, especially as spec versions change anyway. Nice to fix as parts of PRs, but reasonable to merge regardless.

beacon_chain/spec/helpers.nim
zah: The functionality was not considered important enough to warrant the extra import in the new home of this function.

beacon_chain/spec/datatypes/altair.nim
zah: Implemented

beacon_chain/spec/helpers.nim
zah: Nice catch, I suggest implementing this and adding tests for it in your branch after this PR is merged.

beacon_chain/spec/light_client_sync.nim
zah: I was a bit on the fence whether the extra import is worth it, but I've applied the suggestion.

beacon_chain/spec/helpers.nim
zah: `doAssert` added (this is what we use since we want to keep the assertions in release as well).

beacon_chain/spec/helpers.nim
zah: `depth` usually carries a positive compile-time value. We could add an assertion for this, but the behavior with negative `depth` is benign enough perhaps.

beacon_chain/spec/light_client_sync.nim
zah: The `finality_branch` is an `array`, so it has a fixed size in Nim.

beacon_chain/spec/light_client_sync.nim
zah: The `next_sync_committee_branch` is an array, so it has a fixed size in Nim.

beacon_chain/spec/light_client_sync.nim
zah: Code changed to use `newSeqOfCap`

beacon_chain/spec/light_client_sync.nim
zah: Sure, but this would be more appropriate in the higher layers in the code.

beacon_chain/spec/light_client_sync.nim
zah: This loop is a bit inefficient as it has multiple copy operations of the large `ClientUpdate` structure. I think it would be better if we highlight this with a TODO comment instead of further hiding it behind a syntactic short-cut.

The semantics of a general operator `<` are a bit ambiguous perhaps - a chronological ordering would be another way to compare light client updates and one can argue that the "voting weight" is a valid basis for comparison only when the two updates refer to the same "election" (i.e. they point to the same sync committee period, etc).

beacon_chain/spec/helpers.nim
etan-status: OK. Note taken.

beacon_chain/spec/helpers.nim
etan-status: OK, if this is never called with user provided data / only with compile-time values, this is fine.

beacon_chain/spec/light_client_sync.nim
etan-status: Thanks for the explanation!

beacon_chain/spec/light_client_sync.nim
etan-status: How would they determine the reason, though? Duplicating the checks in the higher layers?

beacon_chain/spec/light_client_sync.nim
etan-status: Yeah, the `<` I suggested here because I could not find a `max` flavor taking a predicate in Nim. If there is one, would avoid overloading `<` for that reason that you just mentioned. If going with `<`, would keep it local to just this file (preferrably, local to just the function, but that does not seem to work either).

beacon_chain/spec/light_client_sync.nim
zah: I'm reluctant to remove this until we gain more experience with the full implementation.

beacon_chain/spec/light_client_sync.nim
zah: I've added a new helper in stew (`isZeroMemory`) and used it throughout this module to eliminate the temporary zero-initialized values.

etan-status: Cool! That's even better 😊

beacon_chain/spec/light_client_sync.nim
zah: `valid_updates` cannot be empty here because the function starts by adding the input update in `store.valid_updates.incl(update)`.

beacon_chain/ssz/merkleization.nim
zah: I've added a TODO item for this type. In practice, I think even 16-bit value would be enough as this stores "field index within a consensus object" which has a maximum value that is a function of the current spec types definitions (so, it can be calculated at compile-time)

beacon_chain/spec/light_client_sync.nim
zah: Actually, it's a good point that the hypothetical `set.max` operation can return a `lent` value, thus addressing my concern over the inefficient copies, so I've left another TODO comment suggesting that.

beacon_chain/spec/helpers.nim
etan-status: I ran some experiments, and optimizing this does not seem to produce any relevant speedup. Considering how `get_subtree_index` is solely being used on compile time constants so far, I'd rather keep it as is, to match the spec more closely.

PR #2856, opened 2021-09-08
Sync committee related REST API implementation.

cheatfate:
#### Calls implemented.

* https://ethereum.github.io/beacon-APIs/#/Beacon/getEpochSyncCommittees
* https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolSyncCommitteeSignatures
* https://ethereum.github.io/beacon-APIs/#/Validator/prepareSyncCommitteeSubnets (just stub with arguments verification).
* https://ethereum.github.io/beacon-APIs/#/Validator/produceSyncCommitteeContribution
* https://ethereum.github.io/beacon-APIs/#/Validator/publishContributionAndProofs

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -505,7 +505,8 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
     return RestApiResponse.jsonError(Http500, InternalServerError)
 
   # https://ethereum.github.io/beacon-APIs/#/Beacon/getEpochSyncCommittees
-  router.api(MethodGet, "/eth/v1/beacon/states/{state_id}/sync_committees") do (
+  router.api(MethodGet,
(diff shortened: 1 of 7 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -504,6 +504,85 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        res
+      else:
+        # If ``epoch`` not present then the sync committees for the epoch of
+        # the state will be obtained.
+        bslot.slot.epoch()
+
+    node.withStateForBlockSlot(bslot):
+      let keys =
+        block:
+          let res = node.dag.syncCommitteeParticipants(qepoch)
```
zah: `syncCommitteeParticipants` relies on `headState` internally, so it will fail to return answers for older states. Supporting older states properly seems to be the goal of using `withStateForBlockSlot` and you can achieve that by reading the `current_sync_committee` and `next_sync_committee` fields from the `stateData` object that is injected in the `withState` scope.

PR #2865, opened 2021-09-13
support Altair-only fields in getStateField

etan-status:
The `getStateField` macro provides access to a `ForkedHashedBeaconState`
field value, but only supports fields that exist across all forks. When
called with an Altair-only field a compile time error is produced, like
`Error: undeclared field: '...' for type phase0.BeaconState`.
This patch adds support for fields not present across all forks.


etan-status:
(Not needed right now.)

PR #2867, opened 2021-09-14
block mocking support for Altair

etan-status:
The existing functions for mocking blocks only support phase0 blocks.
This extends block mocking support to also support Altair blocks.

### DIFF
--- a/tests/mocking/mock_attestations.nim
+++ b/tests/mocking/mock_attestations.nim
@@ -15,7 +15,7 @@ import
   chronicles,
   # Specs
   ../../beacon_chain/spec/datatypes/phase0,
-  ../../beacon_chain/spec/[beaconstate, helpers, validator, signatures],
+  ../../beacon_chain/spec/[beaconstate, forks, helpers, validator, signatures],
   # Mocking procs
   ./mock_blocks,
   ./mock_validator_keys
@@ -27,7 +27,9 @@ proc mockAttestationData(
   doAssert state.slot >= slot
 
   if slot == state.slot:
-    result.beacon_block_root = mockBlockForNextSlot(state).message.parent_root
+    let forkedState = (ref ForkedHashedBeaconState)(beaconStateFork: forkPhase0,
+      hbsPhase0: phase0.HashedBeaconState(root: hash_tree_root(state), data: state))[]
+    result.beacon_block_root = mockBlockForNextSlot(forkedState).phase0Block.message.parent_root
   else:
     result.beacon_block_root = get_block_root_at_slot(state, slot)
 
--- a/tests/mocking/mock_blocks.nim
+++ b/tests/mocking/mock_blocks.nim
@@ -6,54 +6,74 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 import
-  options,
-  # Specs
-  ../../beacon_chain/spec/datatypes/phase0,
-  ../../beacon_chain/spec/[helpers, signatures, validator],
+  # Beacon chain internals
+  ../../beacon_chain/spec/[forks, helpers, signatures, state_transition],
   # Mock helpers
   ./mock_validator_keys
 
 # Routines for mocking blocks
 # ---------------------------------------------------------------
 
-proc signMockBlockImpl(
-      state: phase0.BeaconState,
-      signedBlock: var phase0.SignedBeaconBlock
-    ) =
-  let block_slot = signedBlock.message.slot
-  doAssert state.slot <= block_slot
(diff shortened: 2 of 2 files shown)


tests/mocking/mock_blocks.nim
```diff
@@ -8,52 +8,50 @@
...
-  var previous_block_header = state.latest_block_header
-  if previous_block_header.state_root == ZERO_HASH:
-    previous_block_header.state_root = state.hash_tree_root()
-  result.message.parent_root = previous_block_header.hash_tree_root()
-
-  signMockBlock(state, result)
-
-proc mockBlockForNextSlot*(state: phase0.BeaconState): phase0.SignedBeaconBlock =
-  mockBlock(state, state.slot + 1)
+template mockImpl[S,B](): untyped =
```
zah: Relying on explicit generic parameters is a bad habit. You'll eventually run into situations where they don't work well. Consider using regular type parameters instead (aka typedesc). The support for explicit generic parameters in the compiler is very rudimentary and you lose features such as overloading, default arguments and others.

zah (approved):

tests/mocking/mock_blocks.nim
etan-status: Thanks for the insightful comment! I have updated this to use `typedesc` instead.

etan-status:
Investigating some unexpected computations when using this for work-in-progress tests.

etan-status:
This is not sufficient for Altair, as there are some extra steps required.

See https://github.com/ethereum/consensus-specs/blob/e558618ec5077787c87f115ed8e0da8f19c1ddf6/tests/core/pyspec/eth2spec/test/helpers/block.py#L75-L105

etan-status:
Adjusted the mocking code to more closely reflect the spec.

Leaving as draft for now as this conflicts with #2869 and until I verified that this now work properly w.r.t. new tests.

tests/mocking/mock_attestations.nim
```diff
@@ -27,7 +27,9 @@ proc mockAttestationData(
   doAssert state.slot >= slot
 
   if slot == state.slot:
-    result.beacon_block_root = mockBlockForNextSlot(state).message.parent_root
+    let forkedState = (ref ForkedHashedBeaconState)(beaconStateFork: forkPhase0,
+      hbsPhase0: phase0.HashedBeaconState(root: hash_tree_root(state), data: state))[]
+    result.beacon_block_root = mockBlockForNextSlot(forkedState).phase0Block.message.parent_root
```
etan-status, line 32: Attestations out of scope of this PR, so only doing the minimum here for compilation while retaining previous phase0-only functionality.

etan-status:
This will also conflict with #2873 - leaving as draft until that one is merged.

tersec:
> This will also conflict with #2873 - leaving as draft until that one is merged.

Well, #2873 merged.

tests/mocking/mock_attestations.nim
tersec: Agreed.

tersec:
> Adjusted the mocking code to more closely reflect the spec.
> 
> Leaving as draft for now as this conflicts with #2869 and until I verified that this now work properly w.r.t. new tests.

https://github.com/status-im/nimbus-eth2/pull/2869 merged.

tests/mocking/mock_blocks.nim
```diff
@@ -6,54 +6,70 @@
...
+                of forkAltair: BeaconBlockFork.Altair
+  withBlck(result):
+    blck.message.slot = slot
+    blck.message.proposer_index = get_beacon_proposer_index(state, cache, slot).get.uint64
+    blck.message.body.eth1_data.deposit_count = getStateField(state, eth1_deposit_index)
+    blck.message.parent_root = previous_block_header.hash_tree_root()
+
+  applyRandaoReveal(state, result)
+
+  if state.beaconStateFork >= forkAltair:
```
tersec: If `state` is pre-Altair (or pre-Merge) fork epoch, but `slot` is post-Altair/Merge epoch, this will produce incorrect/incompete results.

This seems to be a result of not having `RuntimeConfig` or similar information available here.

That being said, it's at least an unambiguous improvement over the status quo, and doesn't render any existing use case worse. `unstable` currently supports phase 0, and that's exactly as well-supported with this PR. When the block fork matches the state fork, it also works fine.

It might be worth putting a TODO in to ensure this edge case is documented though, for when it makes sense to address the block fork != state fork case.

tersec (approved):

tests/mocking/mock_blocks.nim
etan-status: Thanks for catching this! The other parts of the test framework typically assume `defaultRuntimeConfig`, so the fork information is available. Will revise.

etan-status:
This should now also work at the fork from phase0 to altair.

PR #2868, opened 2021-09-14
remove forked message accessor

etan-status:
The template `message` extracts fork specific data from a forked block.
However, its current implementation does not compile due to each fork
returning a different type, i.e., the generated `case` code attempts to
return different types from each branch. As the template is unused, it
is safely removed for now.

### DIFF
--- a/beacon_chain/spec/forks.nim
+++ b/beacon_chain/spec/forks.nim
@@ -322,9 +322,6 @@ template root*(x: ForkedSignedBeaconBlock | ForkedTrustedSignedBeaconBlock): Eth
 template slot*(x: ForkedSignedBeaconBlock | ForkedTrustedSignedBeaconBlock): Slot =
   getForkedBlockField(x, slot)
 
-template message*(x: ForkedSignedBeaconBlock | ForkedTrustedSignedBeaconBlock): auto =
-  withBlck(x): blck.message
-
 template shortLog*(x: ForkedBeaconBlock): auto =
   withBlck(x): shortLog(blck)
 


cheatfate:
Sorry, but if this template is not used right now, it doesn't mean it will be not used tomorrow. This is API helpers which supposed to make life easier when working with different forks.

etan-status:
As is, the API does not work. The generated `case` statement returns different types on each branch.

cheatfate:
What does it mean it not work? Depends on `fork` it will return right now `phase0.BeaconBlock` or `altair.BeaconBlock`.

etan-status:
```
template withBlck*(x: ForkedBeaconBlock | ForkedSignedBeaconBlock | ForkedTrustedSignedBeaconBlock, body: untyped): untyped =
  case x.kind
  of BeaconBlockFork.Phase0:
    template blck: untyped {.inject.} = x.phase0Block
    body
  of BeaconBlockFork.Altair:
    template blck: untyped {.inject.} = x.altairBlock
    body

template message*(x: ForkedSignedBeaconBlock | ForkedTrustedSignedBeaconBlock): auto =
  withBlck(x): blck.message
```

Trying to call this generates this expression:
```
let m = b.message
```
=>
```
let m = case b.kind
        of BeaconBlockFork.Phase0:
          b.phase0Block.message
        of BeaconBlockFork.Altair:
          b.altairBlock.message
```

What is the type of `m` in this case?
Compiler says `Error: type mismatch: got <BeaconBlock> but expected 'BeaconBlock = object'`


Note that for sub-fields, another helper already exists, e.g.,
```
let i = getForkedBlockField(b, proposer_index)
```

zah:
Yes, this cannot work for signed beacon blocks as written.

PR #875, opened 2020-04-08
reenable block bls verification

arnetheduck:
Replaces #797

PR #2869, opened 2021-09-14
use correct proposer when mocking blocks

etan-status:
When creating beacon blocks, the state needs to be advanced to the new
slot to determine the correct proposer index. In #875 a regression got
introduced that removed this logic when mocking blocks. Existing tests
failed to detect this, as they do not make use of the proposer index.
This patch re-introduces the state advancement, fixing the problem.

### DIFF
--- a/tests/mocking/mock_blocks.nim
+++ b/tests/mocking/mock_blocks.nim
@@ -43,7 +43,7 @@ proc mockBlock(
   ## Mock a BeaconBlock for the specific slot
 
   var emptyCache = StateCache()
-  let proposer_index = get_beacon_proposer_index(state, emptyCache)
+  let proposer_index = get_beacon_proposer_index(state, emptyCache, slot)
   result.message.slot = slot
   result.message.proposer_index = proposer_index.get.uint64
   result.message.body.eth1_data.deposit_count = state.eth1_deposit_index


tests/mocking/mock_blocks.nim
```diff
@@ -37,13 +37,27 @@ proc signMockBlock*(state: phase0.BeaconState, signedBlock: var phase0.SignedBea
...
+    else:
+      # Create state copy to determine proposer index of future slot
+      var
+        cache = StateCache()
+        rewards = RewardInfo()
+        tmpState = (ref ForkedHashedBeaconState)(beaconStateFork: forkPhase0)[]
+      tmpState.hbsPhase0.data = assignClone(state)[]
+      tmpState.hbsPhase0.root = hash_tree_root(tmpState.hbsPhase0.data)
+      doAssert process_slots(cfg, tmpState, slot, cache, rewards, flags = {})
+      get_beacon_proposer_index(tmpState, cache, slot)
```
etan-status: The code here is as close to the previous version (before regression) as possible, wondering if this could be simplified to just `get_beacon_proposer_index(state, emptyCache, slot)`.

etan-status: The short version seems to produce the same results, so switching to that one for now. Not sure how stable this is across forks (`cfg` is not passed in), but in case it is a problem that function should probably take care of it instead of the caller here.

tersec (approved):

PR #2634, opened 2021-06-06
use ForkedHashedBeaconState in StateData

tersec:
Main changes:
- Replaced `phase0.HashedBeaconState` with `ForkedHashedBeaconState`, so the DAG can follow forks such as Altair. This is where all the `getStateField(foo, bar)` to `getStateField(foo.data, bar)` (and similarly for some other `forkedbeaconstate_helpers` helper functions). It'd have been possible to avoid this by defining wrappers over wrappers, so `foo(StateData)` would forward each relevant call to `foo(ForkedHashedBeaconState)`, but that has its own design tradeoffs I don't find compelling.
- Replaced `StateData` usage aimed at allowing this flexibility, outside the DAG and DAG users/callers, with `ForkedHashedBeaconState`, which removes some workarounds related to constructing fake `blck` fields (e.g., in `state_sim`).
- Moved `StateData` and `BlockRef` (which `StateData` contains) back to `consensus_object_pools/block_pools_types`, fixing the architectural regression caused by putting it in `spec/datatypes/`.
- Moved/replaced `consensus_object_pools/statedata_helpers` with similar `spec/forkedbeaconstate_helpers`. This creates `import` change noise, but architecturally where it should have been before. It had been pulled there by `StateData` being a DAG creature, mostly.
- In some cases where it just wasn't appropriate within the scope of this PR to further fix actual portability of typing/functionality between phase 0 and Altair, one sees various introductions of `foo.data.hbsPhase0.data` in place of `foo.data.data`. These both access the phase 0 `BeaconState` object, which is all that worked before anyway. So while the limitations are more obvious, these aren't regressions for `StateData` objects in a `consensus_object_pools` context. The goal is to do those changes separately, where they're not overwhelmed by the type-refactoring noise in this PR.
- `compute_time_at_slot()` and `addLocalValidator()` switched from being passed an entire state (of any type) and just getting the specific information they need.
- `get_attesting_indices()` (both its `ForkedHashedBeaconState` wrapper and individual `BeaconState` forms) switched from `iterator` to `func` to work around a Nim segfault bug (reported upstream).
- add and use a trigger slot in `state_transition{,_slots,_block}()` for when to run the Altair upgrade. It's enough to relatively-speaking hardcode this for now. Discussions are ongoing for how to determine when to trigger it exactly, but for the first testnets, it'll be decided before genesis.
- the `process_slots()`/`state_transition_slots()` functions have been merged, and now understand how to operate on/across fork boundaries.
[...]

PR #2872, opened 2021-09-15
allow adding Altair test blocks

etan-status:
testblockutil currently fails to add test blocks to Altair state because
it assumes phase0 blocks in various places. This patch corrects this
limitation by using forked types internally. Note that existing clients
so far solely operate on phase0 blocks and should eventually be updated.

### DIFF
--- a/tests/testblockutil.nim
+++ b/tests/testblockutil.nim
@@ -10,7 +10,7 @@ import
   options, stew/endians2,
   ../beacon_chain/[beacon_node_types],
   ../beacon_chain/validators/validator_pool,
-  ../beacon_chain/spec/datatypes/[phase0, altair, merge],
+  ../beacon_chain/spec/datatypes/merge,
   ../beacon_chain/spec/[helpers, signatures, state_transition, forks],
   ../beacon_chain/consensus_object_pools/attestation_pool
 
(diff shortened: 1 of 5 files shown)


etan-status:
Follow-up from #2634 that already started migrating parts of the files over to `ForkedHashedBeaconState`.

tests/testblockutil.nim
```diff
@@ -104,34 +103,61 @@ proc addTestBlock*(
...
+        deposits,
+        @[],
+        @[],
+        @[],
+        default(ExecutionPayload),
+        noRollback,
+        cache)
+      doAssert res.isOk(), "Should have created a valid block!"
+      ForkedBeaconBlock.init(res.get())
+    of forkAltair:
```
tersec, line 131: This duplicative code, while it's used elsewhere in `state_transition` and `block_sim`, is one of the uglier aspects IMO of the current design. It's necessitated currently by the
```nim
SyncAggregate(
          sync_committee_signature: ValidatorSig.infinity),
```
parameter being new, but otherwise this is two near-but-not-exactly identical calls which need to be updated/maintained in near-but-not-exact synchrony, in a way currently not easily caught by the test suites which mostly focus on phase 0 with a bit of altair and no specific merge coverage.

This would benefit from either some local caller-side metaprogramming (e.g., a template or macro to generate the two calls from the all-but-one-overlapping parameter) or adjusting the `makeBeaconBlock` function in `state_transition` to take a dummy/ignored `SyncAggregate` parameter, so that the calls could be literally identical, then updating the other `makeBeaconBlock` calls (in `block_sim` and `state_transition`, I believe) to also consolidate them.

tersec: Minimally, it'd be useful to extract all this out to a single helper that this site and the other two call, akin to the helpers in `forks.nim`, to mitigate this issue short of said metaprogramming.

etan-status: I like the idea of extending `forks.nim`, I'll take it as a TODO for now to avoid going beyond the scope of this PR.

etan-status:
Rebasing once more to cover the latest changes in unstable.

tests/testblockutil.nim
tersec: Fair enough, sure, it doesn't need to be done in this PR.

tersec (approved):

tests/testblockutil.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/2879

arnetheduck: on the flip side I wouldn't go too far with fancy techniques that rely on there being only a single-argument difference - sometimes a bit of duplication is the lesser evil compared to a jungle of macros and templates, in spec code more so than elsewhere because it is well tested and mostly doesn't change

etan-status: Sure, but in this specific case the concept of unused placeholder arguments already exist in form of the `executionPayload`. Does not hurt to also add the `sync_aggregate` for consistency.

PR #2874, opened 2021-09-16
add function to build merkle proofs

etan-status:
Adds a function that constructs a Merkle proof for a generalized index.
This will be used during light client sync to update light clients with
a new state (see NEXT_SYNC_COMMITTEE_INDEX / FINALIZED_ROOT_INDEX).

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -67,6 +67,59 @@ func is_valid_merkle_branch*(leaf: Eth2Digest, branch: openArray[Eth2Digest],
     value = eth2digest(buf)
   value == root
 
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.4/tests/core/pyspec/eth2spec/test/helpers/merkle.py#L4-L21
+func build_proof_impl(anchor: object, leaf_index: uint64, 
+                      proof: var openArray[Eth2Digest]) =
+  let
+    bottom_length = nextPow2(typeof(anchor).totalSerializedFields.uint64)
+    tree_depth = log2trunc(bottom_length)
+    parent_index = 
+      if leaf_index < bottom_length shl 1:
+        0'u64
+      else:
+        var i = leaf_index
+        while i >= bottom_length shl 1:
+          i = i shr 1
+        i
+
+  var 
+    prefix_len = 0
+    proof_len = log2trunc(leaf_index)
+    cache = newSeq[Eth2Digest](bottom_length shl 1)
+  block:
+    var i = bottom_length
+    anchor.enumInstanceSerializedFields(fieldNameVar, fieldVar):
+      if i == parent_index:
+        when fieldVar is object:
+          prefix_len = log2trunc(leaf_index) - tree_depth
+          proof_len -= prefix_len
+          let 
+            bottom_bits = leaf_index and not (uint64.high shl prefix_len)
+            prefix_leaf_index = (1'u64 shl prefix_len) + bottom_bits
+          build_proof_impl(fieldVar, prefix_leaf_index, proof)
+        else: raiseAssert "Invalid leaf_index"
+      cache[i] = hash_tree_root(fieldVar)
+      i += 1
+    for i in countdown(bottom_length - 1, 1):
+      cache[i] = withEth2Hash:
+        h.update cache[i shl 1].data
+        h.update cache[i shl 1 + 1].data
+
+  var i = if parent_index != 0: parent_index
+          else: leaf_index
+  doAssert i > 0 and i < bottom_length shl 1
+  for proof_index in prefix_len ..< prefix_len + proof_len:
+    let b = (i and 1) != 0
+    i = i shr 1
+    proof[proof_index] = if b: cache[i shl 1]
+                         else: cache[i shl 1 + 1]
+
+func build_proof*(anchor: object, leaf_index: uint64, 
+                  proof: var openArray[Eth2Digest]) =
+  doAssert leaf_index > 0
+  doAssert proof.len == log2trunc(leaf_index)
+  build_proof_impl(anchor, leaf_index, proof)
+  
 const SLOTS_PER_SYNC_COMMITTEE_PERIOD* =
   EPOCHS_PER_SYNC_COMMITTEE_PERIOD * SLOTS_PER_EPOCH
 
--- a/tests/official/altair/test_fixture_merkle_single_proof.nim
+++ b/tests/official/altair/test_fixture_merkle_single_proof.nim
@@ -1,0 +1,68 @@
+# beacon_chain
+# Copyright (c) 2021 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.used.}
+
+import 
+  # Standard library
+  std/[os, sequtils, streams],
+  # Status libraries
+  stew/bitops2,
+  # Third-party
+  yaml,
+  # Beacon chain internals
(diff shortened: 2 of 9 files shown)


etan-status:
I'm a bit confused about all the different merkleization libraries that are already included in the tree:
* vendor/nim-eth/eth/ssz/merkleization
* vendor/nim-eth/eth/ssz/merkle_tree
* beacon_chain/ssz but those seem specific to deposits?

This problem of duplicate libraries (that are not exact duplicates) extends to more basic types. For example, `BitArray` is exported three times.
* beacon_chain/ssz/bitseqs.nim
* vendor/nim-eth/eth/ssz/bitseqs.nim
* vendor/nim-stew/stew/bitseqs.nim

And, then there is also `Digest` vs `Eth2Digest`.

This makes trying to use the existing libraries quite complicated to even get it to compile.

Furthermore, the existing merkleization libraries seem rather complex with also lots of copying going on.
For the light client sync only very simple merkle proofs are required.

As for the new implementation being proposed here, it is quite easy to use:
```
var next_sync_committee_branch {.noinit.}: array[log2trunc(NEXT_SYNC_COMMITTEE_INDEX), Eth2Digest]
build_proof(state, NEXT_SYNC_COMMITTEE_INDEX, next_sync_committee_branch)
```

If there is a simple way to build a simple Merkle proof in the existing code, please advise :-)

etan-status:
@kdeme I was told that you are migrating to a new SSZ package. If that is still pending, would you mind if we take this additional function as an intermediate solution to unblock light client sync tests?

arnetheduck:
> vendor/nim-eth

can ignore anything in here for now - at some point the shared code will be migrated to a shared repository, but for now, just add what you need in `beacon_chain` - including tests that verify that the generation is done correctly.

etan-status:
Added test for the new function (builds a proof for each field of `BeaconState` and verifies that it validates as expected).

kdeme:
> @kdeme I was told that you are migrating to a new SSZ package. If that is still pending, would you mind if we take this additional function as an intermediate solution to unblock light client sync tests?

Yeah, what @arnetheduck said for now.

Now that some of the code has been better separated, I'd like to pull out most of https://github.com/status-im/nimbus-eth2/tree/unstable/beacon_chain/ssz into another repository (and use that in Fluffy and remove the code from nim-eth). Main item still is probably how to best deal with `Eth2Digest`.

etan-status:
Thanks for the clarification!

tests/test_helpers.nim
```diff
@@ -20,3 +22,28 @@ suite "Spec helpers":
       integer_squareroot(3'u64) == 1'u64
       integer_squareroot(4'u64) == 2'u64
       integer_squareroot(5'u64) == 2'u64
+
+  test "build_proof":
+    var forked = newClone(initGenesisState())
+    var cache = StateCache()
+    var rewards = RewardInfo()
+    doAssert process_slots(defaultRuntimeConfig, forked[], Slot(100), cache, rewards, flags = {})
```
etan-status: More ideal would be a generator for random objects. I'm just using the `BeaconState` here for testing because that's the object that will ultimately also be used in practice. But just for making proofs / verifying they are correct there is no reason that this has to be done on a `BeaconState`.

etan-status:
Note: This simple implementation does not cover Lists yet (but Lists are also not required for the light client sync).
I assume Kim's new implementation will eventually also support Lists, or they could be added once required.

tests/test_helpers.nim
```diff
@@ -20,3 +22,34 @@ suite "Spec helpers":
       integer_squareroot(3'u64) == 1'u64
       integer_squareroot(4'u64) == 2'u64
       integer_squareroot(5'u64) == 2'u64
+
+  test "build_proof":
```
arnetheduck: for tests like this, it's usually good to also test against a reference implementation using a "known-good" proof

arnetheduck:
> Main item still is probably how to best deal with `Eth2Digest`.

lol, this is where I stopped, for some reason ;)

tests/test_helpers.nim
etan-status: Thanks, I fixed the reference implementation, then extended it locally to grab some test vectors, then added some tests for it. Not sure what the recommended pattern for these is.

etan-status: https://github.com/ethereum/consensus-specs/pull/2619

etan-status: @tersec suggested to move the test vectors into a new subdirectory of https://github.com/status-im/nim-eth2-scenarios , e.g., `tests-nimbus` next to `tests-v1.1.0-beta.4`

etan-status: https://github.com/status-im/nim-eth2-scenarios/pull/7

tersec: Once https://github.com/status-im/nimbus-eth2/pull/2909 is merged, to update to v1.1.0 with the upstreamed test vectors, the rest of this PR can be more readily continued.

etan-status: Updated for v1.1.0 tests which now include the reference test vectors officially.
Split up the test cases into the official reference one and the internal one that enumerates BeaconState members.

tests/official/altair/test_fixture_merkle_single_proof.nim
```diff
@@ -0,0 +1,69 @@
+# beacon_chain
+# Copyright (c) 2018 Status Research & Development GmbH
```
tersec: 2018?

etan-status: gotcha. fixed.

etan-status: Actually, "-Present" is not correct. Replaced it with 2021.

beacon_chain/spec/helpers.nim
```diff
@@ -67,6 +67,59 @@ func is_valid_merkle_branch*(leaf: Eth2Digest, branch: openArray[Eth2Digest],
...
+    anchor.enumInstanceSerializedFields(fieldNameVar, fieldVar):
+      if i == parent_index:
+        when fieldVar is object:
+          prefix_len = log2trunc(leaf_index) - tree_depth
+          proof_len -= prefix_len
+          let 
+            bottom_bits = leaf_index and not (uint64.high shl prefix_len)
+            prefix_leaf_index = (1'u64 shl prefix_len) + bottom_bits
+          build_proof_impl(fieldVar, prefix_leaf_index, proof)
+        else: doAssert false, "Invalid leaf_index"
```
arnetheduck:
```suggestion
        else: static: raiseAssert "Invalid leaf_index"
```

two things: `raiseAssert` instead of `doAssert fals` and `static` turns it into a compile-time error - there's some fancier way of doing compile-time errors as well (cc @zah) but this is easy to remember at least

tests/official/altair/test_fixture_merkle_single_proof.nim
```diff
@@ -0,0 +1,68 @@
+# beacon_chain
```
arnetheduck, line 1: at some point we should rename "official" to "ef" or something similar (what's official, anyway?)

arnetheduck (approved):
I suspect we'll end up wanting to integrate this with the hashlist cache at some point, but until then, lgtm

tests/official/altair/test_fixture_merkle_single_proof.nim
etan-status: yeah, something like that would be a descriptive.

beacon_chain/spec/helpers.nim
etan-status: `raiseAssert` works well, but the `static` does not compile for me:
```
`false` Invalid leaf_index")' has no type (or is ambiguous)
```

I think that compile-time error cannot work here because this is not a generic function / it extracts type-information at run-time.

Amending to include `raiseAssert` for now.

tests/official/altair/test_fixture_merkle_single_proof.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/2976

PR #2875, opened 2021-09-16
Fallback ping requests on metadata V2

Menduist:
Some clients (like Prysm) disable their metadata V1 after the Altair fork, so we are unable to ping them correctly.

This PR adds a fallback to try metadata V2 if V1 fails. AFAICT, there is no way to know from `eth2_network.nim` if the fork happened, so we can't skip V1 altogether.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1513,17 +1513,17 @@ proc updatePeerMetadata(node: Eth2Node, peerId: PeerID) {.async.} =
   trace "updating peer metadata", peerId
 
   var peer = node.getPeer(peerId)
-  let response = await peer.getMetaData()
 
+  #getMetaData can fail with an exception
   let newMetadata =
-    if response.isOk:
-      response.get()
-    else:
-      let responseV2 = await peer.getMetaDataV2()
-      if responseV2.isErr:
-        debug "Failed to retrieve metadata from peer!", peerId
-        return
-      let metadataV2 = responseV2.get()
+    try:
+      tryGet(await peer.getMetaData())
+    except CatchableError:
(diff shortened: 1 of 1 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -1513,13 +1513,21 @@ proc updatePeerMetadata(node: Eth2Node, peerId: PeerID) {.async.} =
...
   var peer = node.getPeer(peerId)
-  let response = await peer.getMetaData()
 
-  if response.isErr:
-    debug "Failed to retrieve metadata from peer!", peerId
-    return
+  #getMetaData can fail with an exception
+  let newMetadata =
+    try:
+      tryGet(await peer.getMetaData())
```
mratsim, line 1520: Why the switch to try/catch here instead of isErr like before?

mratsim (approved):
sgtm, nitpick on exception/error style

beacon_chain/networking/eth2_network.nim
Menduist: getMetaData can fail with a Result error _or_ an exception

Previously the exception made the entire procedure fail, this is the cleanest way I found to handle both errors & exceptions

arnetheduck: given that post-upgrade, everyone will be supporting v2, we should be trying v2 first so as to future-proof the implementation

Menduist: Yes, I did it this way because right now, everyone except Prysm post-merge support V1, and most client pre-merge don't handle V2, so there is more people handling V1 than V2 (if you count mainnet)
But once every network supports it, we can even remove V1 without issues

PR #2878, opened 2021-09-19
REST /eth/v1/events API call implementation.

cheatfate:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -127,12 +127,46 @@ proc init*(T: type BeaconNode,
...
   let
+    eventBus = newAsyncEventBus()
     db = BeaconChainDB.new(config.databaseDir, inMemory = false)
 
   var
     genesisState, checkpointState: ref phase0.BeaconState
     checkpointBlock: phase0.TrustedSignedBeaconBlock
 
+  proc onAttestationAdded(data: Attestation) {.
+       gcsafe, raises: [Defect].} =
```
arnetheduck: are these gcsafe/raises necessary? they should not be needed

beacon_chain/consensus_object_pools/block_pools_types.nim
```diff
@@ -48,6 +48,20 @@ type
     Old
     Duplicate
 
+  OnBlockCallback* =
+    proc(data: ForkedTrustedSignedBeaconBlock) {.gcsafe, raises: [Defect].}
+  OnHeadCallback* =
```
arnetheduck, line 53: why not pass `HeadChangeInfoObject` here? ditto for the other callbacks?

beacon_chain/consensus_object_pools/block_pools_types.nim
```diff
@@ -48,6 +48,20 @@ type
...
+    proc(slot: Slot, blockRoot: Eth2Digest, stateRoot: Eth2Digest,
+         epochTransition: bool, previousDutyDepRoot: Eth2Digest,
+         currentDutyDepRoot: Eth2Digest) {.gcsafe, raises: [Defect].}
+  OnReorgCallback* =
+    proc(slot: Slot, depth: uint64, oldHeadBlockRoot: Eth2Digest,
+         newHeadBlockRoot: Eth2Digest, oldHeadStateRoot: Eth2Digest,
+         newHeadStateRoot: Eth2Digest) {.gcsafe, raises: [Defect].}
+  OnFinalizedCallback* =
+    proc(blockRoot: Eth2Digest, stateRoot: Eth2Digest,
```
arnetheduck: use a `CheckPoint` here?

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -303,6 +306,10 @@ proc addAttestation*(pool: var AttestationPool,
         attestation, signature):
       return
 
+  # Send notification about new attestation via callback.
+  if not(isNil(pool.onAttestationAdded)):
+    pool.onAttestationAdded(attestation)
```
arnetheduck: my gut feeling is this should happen _after_ `addForkChoiceVotes` is done (as well as any other side effects of adding the attestation to the pool) - if someone inspects the pool in the callback, it should already reflect what was added

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -187,24 +187,29 @@ func link*(parent, child: BlockRef) =
 
   child.parent = parent
 
-func isAncestorOf*(a, b: BlockRef): bool =
+func getDepth*(a, b: BlockRef): tuple[ancestor: bool, depth: int] =
```
arnetheduck, line 190: hm. is the depth meaningful for non-ancestors? `Option[int]` would perhaps be the better return type here

cheatfate: According to ETH2 specification https://ethereum.github.io/beacon-APIs/#/Events/eventstream
`depth` is only used for `chain_reorg` event, so it used only for non-ancestors:
```
event: chain_reorg
data: {"slot":"200", "depth":"50",
"old_head_block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf",
"new_head_block":"0x76262e91970d375a19bfe8a867288d7b9cde43c8635f598d93d39d041706fc76",
"old_head_state":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf",
"new_head_state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9",
"epoch":"2"}
```

arnetheduck (approved):

arnetheduck (approved):

PR #2879, opened 2021-09-20
revise `makeBeaconBlock` overloads

etan-status:
The phase0 and altair overloads of `makeBeaconBlock` slightly differ in
their signatures which makes using them unnecessarily verbose.
- A placeholder `sync_aggregate` argument similar to `executionPayload`
  is added to the phase0 overload to match the altair signature.
- A wrapper operating on `ForkedHashedBeaconState` is introduced.

### DIFF
--- a/beacon_chain/spec/state_transition.nim
+++ b/beacon_chain/spec/state_transition.nim
@@ -362,7 +362,7 @@ proc state_transition*(
     cfg, state, signedBlock, cache, flags, rollback)
 
 # https://github.com/ethereum/eth2.0-specs/blob/v1.0.1/specs/phase0/validator.md#preparing-for-a-beaconblock
-proc makeBeaconBlock*(
+template partialBeaconBlock(
     cfg: RuntimeConfig,
     state: var phase0.HashedBeaconState,
     proposer_index: ValidatorIndex,
@@ -375,17 +375,9 @@ proc makeBeaconBlock*(
     proposerSlashings: seq[ProposerSlashing],
     attesterSlashings: seq[AttesterSlashing],
     voluntaryExits: seq[SignedVoluntaryExit],
-    executionPayload: ExecutionPayload,
-    rollback: RollbackHashedProc,
-    cache: var StateCache): Result[phase0.BeaconBlock, string] =
-  ## Create a block for the given state. The last block applied to it must be
-  ## the one identified by parent_root and process_slots must be called up to
-  ## the slot for which a block is to be created.
-
-  # To create a block, we'll first apply a partial block to the state, skipping
-  # some validations.
-
-  var blck = phase0.BeaconBlock(
+    sync_aggregate: SyncAggregate,
+    executionPayload: ExecutionPayload): phase0.BeaconBlock =
+  phase0.BeaconBlock(
     slot: state.data.slot,
     proposer_index: proposer_index.uint64,
     parent_root: parent_root,
@@ -401,6 +393,35 @@ proc makeBeaconBlock*(
       deposits: List[Deposit, Limit MAX_DEPOSITS](deposits),
       voluntary_exits:
         List[SignedVoluntaryExit, Limit MAX_VOLUNTARY_EXITS](voluntaryExits)))
+
+proc makeBeaconBlock*(
+    cfg: RuntimeConfig,
+    state: var phase0.HashedBeaconState,
+    proposer_index: ValidatorIndex,
+    parent_root: Eth2Digest,
+    randao_reveal: ValidatorSig,
+    eth1_data: Eth1Data,
+    graffiti: GraffitiBytes,
+    attestations: seq[Attestation],
+    deposits: seq[Deposit],
+    proposerSlashings: seq[ProposerSlashing],
+    attesterSlashings: seq[AttesterSlashing],
+    voluntaryExits: seq[SignedVoluntaryExit],
+    sync_aggregate: SyncAggregate,
+    executionPayload: ExecutionPayload,
+    rollback: RollbackHashedProc,
+    cache: var StateCache): Result[phase0.BeaconBlock, string] =
+  ## Create a block for the given state. The last block applied to it must be
+  ## the one identified by parent_root and process_slots must be called up to
+  ## the slot for which a block is to be created.
+
+  # To create a block, we'll first apply a partial block to the state, skipping
+  # some validations.
+
+  var blck = partialBeaconBlock(cfg, state, proposer_index, parent_root, 
+                                randao_reveal, eth1_data, graffiti, attestations, deposits, 
+                                proposerSlashings, attesterSlashings, voluntaryExits,
+                                sync_aggregate, executionPayload)
 
   let res = process_block(cfg, state.data, blck, {skipBlsValidation}, cache)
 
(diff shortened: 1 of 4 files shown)


etan-status:
This PR was motivated by the review comment here:
https://github.com/status-im/nimbus-eth2/pull/2872#discussion_r710244166

beacon_chain/spec/state_transition.nim
```diff
@@ -450,3 +491,56 @@ proc makeBeaconBlock*(
...
+                            {skipBlsValidation}, cache)
+
+    if res.isErr:
+      warn "Unable to apply new block to state",
+        blck = shortLog(blck),
+        slot = state.`hbs kind`.data.slot,
+        eth1_deposit_index = state.`hbs kind`.data.eth1_deposit_index,
+        deposit_root = shortLog(state.`hbs kind`.data.eth1_data.deposit_root),
+        error = res.error
+      rollback(state)
```
etan-status, line 563: Wrapping those `rollback` callbacks is difficult because the existing `makeBeaconBlock` overloads operate on a fork-specific `state`, hence have an incompatible signature. This means that I would have to wrap the inner `state` into a `ForkedHashedBeaconState`, creating a copy, then calling the `rollback`, then apply the potential changes back to the fork-specific `state`. Creating an extra assertion that `innerState.addr == addr state.hbsXxx` cannot avoid these copies because capturing a `var` in a closure is not supported.

Wrapping example:
```
proc rollbackImpl(var state: phase0.HashedBeaconState) =
  var forked = ForkedHashedBeaconState(beaconStateFork: forkPhase0, hbsPhase0: state)
  rollback(forked)
  state = forked.hbsPhase0
```

Ultimately, I decided to extract the block-creation logic into `partialBeaconBlock` templates and replicating the existing logging + rollback once more here.

beacon_chain/spec/state_transition.nim
```diff
@@ -450,3 +491,56 @@ proc makeBeaconBlock*(
...
+      return err("Unable to apply new block to state: " & $res.error())
+
+    state.`hbs kind`.root = hash_tree_root(state.`hbs kind`.data)
+    blck.`kind Block`.state_root = state.`hbs kind`.root
+    ok(blck)
```
etan-status, line 568: Besides the `rollback`, the rest here could technically be implemented using `getStateField` et al. The problem is that in that case, each individual call would generate the `case` statement, bloating up the generated code even further. As almost every line contains fork specific fields, I opted to put the entire logic under the template.

beacon_chain/validators/validator_duties.nim
```diff
@@ -333,58 +333,40 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
-        stateData.data.hbsPhase0,
+        stateData.data,
         validator_index,
         head.root,
         randao_reveal,
         eth1Proposal.vote,
         graffiti,
         node.attestationPool[].getAttestationsForBlock(
-          stateData.data.hbsPhase0, cache),
+          stateData.data.`hbs kind`, cache),
```
etan-status, line 353: There is a `getAttestationsForBlock` that operates on forked state type, but it is only defined in the test section. This could be further cleaned up if that function would be integrated.

etan-status, line 445: Not very nice, but out of scope of this PR and not worse than before.

cheatfate: I really do not like `<SPACE>` inside.

tersec: It does resolve an issue that there's lots of duplication of exactly structurally isomorphic situations, just across different case object types. I'm not sure whether this is the optimal approach, but in general it can help reduce explicit `case` expressions everywhere.

Applying this kind of pattern systematically through `forks.nim` would shorten it considerably, for example.

etan-status: I think the space is needed for token concatenation (kind is the variable going into the template, "hbs" is just the prefix). Once concatenated, this becomes `hbsPhase0` or `hbsAltair`.

etan-status: https://nim-lang.github.io/Nim/manual.html#templates-identifier-construction

beacon_chain/spec/state_transition.nim
tersec: The approach taken in a couple places is to rollback the `StateData` as a whole regardless and rely on the `=`/`assign` function (it's a bit inconsistent in usage as to which is used, but the performance improvement from `assign` hopefully isn't hit enough to be relevant either way) to handle the case object aspects.

beacon_chain/spec/state_transition.nim
tersec: Reasonable enough. Another approach, though one with broader ramifications across the codebase, is to take advantage of every `rollback` being on the entire `StateData` object, and not rollback just each hardfork's version separately.

The wrapping would create more GC garbage and memory pressure, since those are all copies, and states are large.

The real solution, though is to only use `RollbackForkedHashedProc` and drop the other more specific versions.

beacon_chain/spec/state_transition.nim
tersec: Yeah, agree, once one knows statically which type of object's in use, better to take advantage of that.

tersec (approved):

beacon_chain/spec/state_transition.nim
etan-status: It would be a bit weird though to have the rollback procs operating on the `forked` type, but the rest of the function not requiring it.

But, my comment was more to explain why I duplicated the code one more time instead of wrapping the rollbacks. To avoid all those extra copying.

etan-status:
Rebased once more to cover `forkMerge`. In `unstable` this was handled same as `forkAltair`, so I did the same here. There is no merge-specific `makeBeaconBlock` so far.

tersec:
> Rebased once more to cover `forkMerge`. In `unstable` this was handled same as `forkAltair`, so I did the same here. There is no merge-specific `makeBeaconBlock` so far.

Indeed, in pure consensus terms, the merge fork doesn't represent much change. It's almost all identical to Altair.

That being said, beacon blocks will change to include execution payload information, but that's not there quite yet -- that PR was just establishing a basic foundation.

beacon_chain/validators/validator_duties.nim
```diff
@@ -425,58 +425,40 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
-        node.exitPool[].getVoluntaryExitsForBlock(),
-        node.sync_committee_msg_pool[].produceSyncAggregate(head.root),
-        default(ExecutionPayload),
-        restore,
-        cache).map(proc (t: auto): auto = ForkedBeaconBlock.init(t))
+    let doPhase0 = slot.epoch < node.dag.cfg.ALTAIR_FORK_EPOCH
+    return if doPhase0:
+      let sync_aggregate = SyncAggregate(sync_committee_signature: ValidatorSig.infinity)
+      makeBeaconBlock(phase0)
+    else:
```
tersec, line 459: It's been useful here to check for all hardfork cases explicitly and `raiseAssert` otherwise, to catch everywhere which needs changing at compile-time. Probably the `sync_aggregate`'s going to in the sharding (or otherwise next) fork anyway, but hard to know for certain.

etan-status: Hmm, not sure that covering _all_ hardfork cases explicitly is possible, as there is always the "else" case of "after known forks", or am I missing something? There is also similar logic spread throughout the codebase (search for `ALTAIR_FORK_EPOCH`) which may call for a dedicated PR that cleans up all of those instances in one sweep.
Note that this PR keeps the logic the same as before, it is just getting rid of the duplicated code.

etan-status:
Here an official reference for the execution payload logic for post-merge blocks:

https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.4/tests/core/pyspec/eth2spec/test/helpers/block.py#L75-L105

```
    if is_post_merge(spec):
        randao_mix = spec.compute_randao_mix(state, empty_block.body.randao_reveal)
        empty_block.body.execution_payload = build_empty_execution_payload(spec, state, randao_mix)
```

beacon_chain/validators/validator_duties.nim
tersec: The idea is "after known forks" should explicitly, and loudly, fail/assert.

It's not a loss of functionality because by definition, a hardfork means that one needs updated code to run it anyway, and it ensures that it's not overlooked when updating to new forks.

Right now, that code assumes something about the next fork that's likely, but uncertain, and will have to be checked.

etan-status: How would the check for "after known forks" look?

Define an `AFTER_SHARDING_FORK_EPOCH` that gets renamed to the next fork when it is defined? Or would it be `epoch > MERGE_FORK_EPOCH`? I don't know if this code is already updated for merge / sharding.

tersec: Hm, you're right, this isn't really feasible based on epochs.

beacon_chain/validators/validator_duties.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/2937
`<SPACE>` is gone :-)

beacon_chain/validators/validator_duties.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/2939

PR #2881, opened 2021-09-21
The peer cycling takes into accounts the syncnets ENR fields

zah:


kdeme (commented):
I guess we also need to update our own ENR regarding subscribed syncnets?

As is done for attnets: https://github.com/status-im/nimbus-eth2/blob/44f652f704f14c67e648f2b14dee652d44b858c5/beacon_chain/networking/eth2_network.nim#L1797

zah:
Already done in `updateSyncnetsMetadata`:

https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/networking/eth2_network.nim#L2049-L2063

kdeme:
Ah, great. Sorry, missed that one, was looking at wrong revision somehow :-/

PR #2885, opened 2021-09-22
speed up `validatorDb`/`validatorPerf`

arnetheduck:
avoid needless hashing and use bigger txn:s

### DIFF
--- a/ncli/ncli_db.nim
+++ b/ncli/ncli_db.nim
@@ -523,9 +523,13 @@ proc cmdValidatorPerf(conf: DbConf, cfg: RuntimeConfig) =
   for bi in 0..<blockRefs.len:
     blck = db.getBlock(blockRefs[blockRefs.len - bi - 1].root).get()
     while getStateField(state[].data, slot) < blck.message.slot:
+      let
+        nextSlot = getStateField(state[].data, slot) + 1
+        flags =
+          if nextSlot == blck.message.slot: {skipLastStateRootCalculation}
+          else: {}
       let ok = process_slots(
-        dag.cfg, state[].data, getStateField(state[].data, slot) + 1, cache,
-        rewards, {})
+        dag.cfg, state[].data, nextSlot, cache, rewards, flags)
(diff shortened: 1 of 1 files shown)


ncli/ncli_db.nim
```diff
@@ -746,14 +753,20 @@ proc cmdValidatorDb(conf: DbConf, cfg: RuntimeConfig) =
           int64(target_attester), # Target delta
           int64(head_attester), # Head delta
           delay)).expect("DB")
-    outDb.exec("COMMIT;").expect("DB")
+    if getStateField(state[].data, slot).epoch.int64 mod 16 == 0:
```
tersec, line 756: Seems potentially more elegant to use something like 16'u64 or some other type-compatible constant than to convert the runtime-variable `getStateField(state[].data, slot).epoch` to a theoretically incompatible integer type.

ncli/ncli_db.nim
```diff
@@ -775,6 +788,10 @@ proc cmdValidatorDb(conf: DbConf, cfg: RuntimeConfig) =
     if getStateField(state[].data, slot).isEpoch():
       processEpoch()
 
+  if inTxn:
+    inTxn = false
+    outDb.exec("COMMIT;").expect("DB")
```
tersec, line 793: How much does this commit-every-16-epochs optimization help, quantitatively?

tersec (approved):

ncli/ncli_db.nim
arnetheduck: that'll be highly dependent on the drive, but about 2x:ish on my laptop - what happens on commit is more or less that the WAL gets moved to the real db and an fsync is made.

ncli/ncli_db.nim
arnetheduck: yeah I thought about that but we already have to convert to int64 to insert row into database (no uint64 in sqlite).. it's a bridge to cross if ever it crashes

PR #2668, opened 2021-06-22
Peer cycling

Menduist:
closes #2341 

EDIT:
This PR adds a three functionalities:
- A metadata pinger, asking peers for their metadata every 30 minutes, and kicking unresponsive peers
- A smart dialer, which will only dial peers when required, and choose which peers to dial
- A connection trimmer, which make room for new interesting peers by kicking useless one when needed

### Peer cycling
The primary goal of peer cycling is to have a good mesh, no matter what `--max-peers` is. Obviously, not an issue with `--max-peers=160`, but trickier with `--max-peers=20`.
We define a good mesh as:
- Every subscribed topic is healthy (at least `dLow` (6) peers, at least `dOut` (3) outgoing peers)
- We have at least `d` (8) connected peers in every topic (used for fanout etc)

`getLowAttnets` returns the topic we have to find peers for, in order to reach this good mesh.

If `getLowAttnets` returns some topic, we are then unhealthy, and should dial new peers having theses attnets, even if that means kicking existing peers to do so.
If we are healthy however, we have no reasons to continue dialing, since that will just consume more resources for no added benefit. (Obviously, currently, we'll get incoming connections instead, but in the future the goal would be to also limit incoming connections in that situation)

This strategy works well with as low as 20 peers, and is also good for the network as a whole (we have a lot of incoming slots available, and don't dial too much)

PR #2888, opened 2021-09-22
Improve peer cycling

Menduist:
Follow up #2668

Some peer cycling improvements:
- Add the number of unhealthy topics in the Slot start log
- Add a metrics for the number of kicked peers
- Sort discovered peers correctly
- Smarter peer kicker
- Slightly increase the maximum metadata request ping failure time (30s -> 45s)

The new peer kicker will kick enough peer to make room for every discovered peer (with an upper limit of `peer_count / 5`)
This gives better performance on big server, since they can kick bad peers quicker, while maintaining a reasonable number of peers on smaller servers

This also fix a dial issue when the semaphore count was <0 and dialing previously stopped. Though the semaphore count should not be <0, so there must be a bug in libp2p, at least the case is handled correctly here.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1005,7 +1005,7 @@ proc trimConnections(node: Eth2Node, count: int) {.async.} =
   for peerId in scores.keys:
     #TODO kill a single connection instead of the whole peer
     # Not possible with the current libp2p's conn management
-    debug "kicking peer", peerId
+    debug "kicking peer", peerId, score=scores[peerId]
     await node.switch.disconnect(peerId)
     dec toKick
     inc(nbc_cycling_kicked_peers)
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -82,7 +82,6 @@ type
     peerPingerHeartbeatFut: Future[void]
     cfg: RuntimeConfig
     getBeaconTime: GetBeaconTimeFn
-    unhealthySubnetsCount*: int
 
   EthereumNode = Eth2Node # needed for the definitions in p2p_backends_helpers
 
@@ -1065,9 +1064,6 @@ proc runDiscoveryLoop*(node: Eth2Node) {.async.} =
       (wantedAttnets, wantedSyncnets) = node.getLowSubnets()
       wantedAttnetsCount = wantedAttnets.countOnes()
       wantedSyncnetsCount = wantedSyncnets.countOnes()
-
-    # Just take attnets into account for UX simplicity
-    node.unhealthySubnetsCount = wantedAttnetsCount
 
     if wantedAttnetsCount > 0 or wantedSyncnetsCount > 0:
       let discoveredNodes = await node.discovery.queryRandom(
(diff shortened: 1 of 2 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1110,6 +1110,7 @@ proc onSlotStart(
     delay = shortLog(delay),
     peers = len(node.network.peerPool),
     head = shortLog(node.dag.head),
+    unhealthySubnetCount = node.network.unhealthySubnetsCount,
```
Menduist: This is a bit simplistic, and may frighten users.
Unhealthy can be either:
- Topic has peers < `dLow` (bad)
- Topic has peers < `d` (okay-ish)
- Topic has outgoing peers < `dOut` (okay-ish)

So you may have unhealthySubnetCount=47 just because every topic are below d, even though you have at least dLow peers everywhere, so in practice everything works correctly.

Not sure how to log this in simple terms, maybe have two "tiers":
- Bad topics (<dLow)
- Suboptimal topics (<d or <dOut)

arnetheduck: hm, that's a good point - it's even worse than that because for sending attestations, we need a subscribed peer per topic that we'll be attesting on, while for block production and attestation aggregation, we need a healthy _mesh_. You're right that expressing this poorly will cause more confusion than it's worth.

A more complex logic would perhaps work something like that: look at what attestations we'll be sending "soon" and make sure those topics have _subscribed_ peers, then look at topics _we're_ subscribed to - for these we need a healthy mesh. The end result would be a number like `A/B` where B is the total monitored subscriptions and A is the number that is healthy.

Another nuance _could_ be to distinguish topics with only one peer from those that are fully healthy (above dlow).

All in all, let's perhaps leave this for later, when the cycling is tuned to take into account all nuances we need it to, then make a health assessment based on that.

beacon_chain/networking/eth2_network.nim
```diff
@@ -1001,9 +1005,10 @@ proc trimConnections(node: Eth2Node, count: int) {.async.} =
   for peerId in scores.keys:
     #TODO kill a single connection instead of the whole peer
     # Not possible with the current libp2p's conn management
-    debug "kicking peer", peerId
+    debug "kicking peer", peerId, score=scores[peerId]
```
arnetheduck, line 1007: would be good to see some tests for the scoring logic - ie given a set of peers, we should test the outcome of sorting them such that the algorithm gives a reasonable order (no need to test the actual score, just that the right peers are kicked)

arnetheduck:
> This also fix a dial issue when the semaphore count was <0 and dialing previously stopped. 

how is this fixed, in the PR? by simply kicking more peers?

Menduist:
Previously the case with count < 0 was not handled at all (since I though it was impossible, rookie error):
`if node.peerPool.lenSpace({PeerType.Outgoing}) == 0`

Since the new behaviour is `roomRequired - roomCurrent`, it will give correct response even with negatives roomCurrent (ex, 1 discovered peer, roomCurrent = -5, kicking up to 6 peers)

beacon_chain/networking/eth2_network.nim
Menduist: Good point, I'll do that in a follow-up PR since this one is semi-urgent

beacon_chain/nimbus_beacon_node.nim
arnetheduck: ping @Menduist - let's remove for now until we have a better idea of what this should mean

Menduist: Done - https://github.com/status-im/nimbus-eth2/pull/2888/commits/67ed7ac1eee52a61b9ce3da107366af92392c2ca

PR #2889, opened 2021-09-22
improve getStateField compile checks

etan-status:
The current `getStateField` implementation fails at run-time when called
on a post-Altair state. This is improved by replacing the `if` structure
with a `case` expression, which is checked for exhaustive coverage at
compile time. Care is taken to preserve the `unsafeAddr` optimization.

### DIFF
--- a/beacon_chain/spec/forks.nim
+++ b/beacon_chain/spec/forks.nim
@@ -122,11 +122,15 @@ func assign*(tgt: var ForkedHashedBeaconState, src: ForkedHashedBeaconState) =
     # with nimOldCaseObjects. This is infrequent.
     tgt = src
 
-macro getStateField*(s, y: untyped): untyped =
-  result = quote do:
-    (if `s`.beaconStateFork == forkPhase0:
-       unsafeAddr (`s`.hbsPhase0.data.`y`) else:
-         unsafeAddr (`s`.hbsAltair.data.`y`))[]
+template getStateField*(x, y: untyped): untyped =
+  # The use of `unsafeAddr` avoids excessive copying in certain situations, e.g.,
+  # ```
+  #   for index, validator in getStateField(stateData.data, validators).pairs():
+  # ```
+  # Without `unsafeAddr`, the `validators` list would be copied to a temporary variable.
+  (case x.beaconStateFork
+   of forkPhase0: unsafeAddr (x.hbsPhase0.data.y)
+   of forkAltair: unsafeAddr (x.hbsAltair.data.y))[]
 
 template getStateRoot*(x: ForkedHashedBeaconState): Eth2Digest =
   case x.beaconStateFork:


etan-status:
History: https://github.com/status-im/nimbus-eth2/pull/2634/commits/7d206d95c38f20a26a84d54e1c21676892c33b3b

tersec:
For reference, the C code it generates for `slot` (a `uint64`):
```c
#line 131 "nimbus-eth2/beacon_chain/spec/forks.nim"
    T5_ = (NU64*)0; switch ((*state_0).data.beaconStateFork) {
    case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 0):
    {       T5_ = (&(*state_0).data.hbsPhase0.data.slot);   }
    break;
    case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 1):
    {       T5_ = (&(*state_0).data.hbsAltair.data.slot);   }
    break;
    }
```

It's fine. Exactly as before. It was/is a bit funny it basically unnecessarily creates this pointer indirection, but that's not a regression. Not doing so would hurt the `validators` case too badly.

For a 32-byte array, used as an input to another function:
```c
#line 620 "nimbus-eth2/beacon_chain/consensus_object_pools/blockchain_dag.nim"
            switch ((*state_0).data.beaconStateFork) {            
            case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 0):                     
            {               T42_ = (*state_0).data.hbsPhase0.root;          }
            break;
            case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 1):
            {               T42_ = (*state_0).data.hbsAltair.root;          }
            break;                                                                           
            }                                                                 
```

For the perf-determining `validators` case:
```c
#line 131 "nimbus-eth2/beacon_chain/spec/forks.nim"
    T321_ = (tyObject_HashList__M609cw2woH7OaDSK8uR8ieg*)0; switch ((*dag).headState.data.beaconStateFork) {
    case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 0):
    {       T321_ = (&(*dag).headState.data.hbsPhase0.data.validators); }
    break;
    case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 1):
    {       T321_ = (&(*dag).headState.data.hbsAltair.data.validators); }
    break;
    }
```
This is the correct result -- it copies pointers rather than the whole `validators` list.

It's still preserving the non-large-structure-copying, non-allocating aspect of this existing approach. In no case I've found does it copy more than a pointer-size worth of data.

tersec (approved):

etan-status:
The extra pointer indirection for the `slot` case disappears after compiling C to ASM.

If you do it without `unsafeAddr`, it will still assign to a temp variable first in C, before then using that temp variable for the actual computation, so in C you always end up with the double copy. It's just that for the big structures like the `validators` array that the extra copy won't be optimized away when compiling to ASM.

PR #2890, opened 2021-09-22
implement forked merge state/block support

tersec:
Passes all the v1.1.0-beta.5 EF merge test vectors except `fork_choice`, which aren't supported by Nimbus for any hardforks.

It sometimes refers to beta.4 rather than beta.5, but there's not much point in updating the URLs for that when the non-beta v1.1.0 is so close.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -14,7 +14,7 @@ import
   eth/db/[kvstore, kvstore_sqlite3],
   ./networking/network_metadata, ./beacon_chain_db_immutable,
   ./spec/[eth2_ssz_serialization, eth2_merkleization, state_transition],
-  ./spec/datatypes/[phase0, altair],
+  ./spec/datatypes/[phase0, altair, merge],
   ./filepath
 
 export
@@ -484,6 +484,9 @@ proc putBlock*(db: BeaconChainDB, value: phase0.TrustedSignedBeaconBlock) =
 proc putBlock*(db: BeaconChainDB, value: altair.TrustedSignedBeaconBlock) =
   db.altairBlocks.putSnappySSZ(value.root.data, value)
   db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
+
+proc putBlock*(db: BeaconChainDB, value: merge.TrustedSignedBeaconBlock) =
+  doAssert false # TODO Deliberate -- put impl in separately
 
 proc updateImmutableValidators*(
     db: BeaconChainDB, validators: openArray[Validator]) =
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -655,9 +655,19 @@ proc readValue*(reader: var JsonReader[RestJson],
     if res.isNone():
       reader.raiseUnexpectedValue("Incorrect altair block format")
     value = ForkedBeaconBlock.init(res.get())
+  of BeaconBlockFork.Merge:
+    let res =
+      try:
+        some(RestJson.decode(string(data.get()), merge.BeaconBlock,
+                             requireAllFields = true))
+      except SerializationError:
+        none[merge.BeaconBlock]()
+    if res.isNone():
+      reader.raiseUnexpectedValue("Incorrect merge block format")
+    value = ForkedBeaconBlock.init(res.get())
 
 proc writeValue*(writer: var JsonWriter[RestJson], value: ForkedBeaconBlock) {.
-     raises: [IOError, Defect].} =
+     raises: [IOError, SerializationError, Defect].} =
   writer.beginRecord()
   case value.kind
   of BeaconBlockFork.Phase0:
@@ -666,6 +676,9 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: ForkedBeaconBlock) {.
   of BeaconBlockFork.Altair:
     writer.writeField("version", "altair")
     writer.writeField("data", value.altairBlock)
+  of BeaconBlockFork.Merge:
+    writer.writeField("version", "merge")
+    writer.writeField("data", value.mergeBlock)
   writer.endRecord()
 
 ## ForkedSignedBeaconBlock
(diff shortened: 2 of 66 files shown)


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -626,9 +626,19 @@ proc readValue*(reader: var JsonReader[RestJson],
...
+                             requireAllFields = true))
+      except SerializationError:
+        none[merge.BeaconBlock]()
+    if res.isNone():
+      reader.raiseUnexpectedValue("Incorrect merge block format")
+    value = ForkedBeaconBlock.init(res.get())
 
 proc writeValue*(writer: var JsonWriter[RestJson], value: ForkedBeaconBlock) {.
-     raises: [IOError, Defect].} =
+     raises: [IOError, SerializationError, Defect].} =
```
tersec: It's unclear to me, and suspicious, why adding the merge support would require `SerializationError` to be declared, when neither phase0 nor altair did.

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -848,14 +848,14 @@ func process_participation_flag_updates*(state: var altair.BeaconState) =
...
-proc process_sync_committee_updates*(state: var altair.BeaconState) =
+proc process_sync_committee_updates*(state: var (altair.BeaconState | merge.BeaconState)) =
   let next_epoch = get_current_epoch(state) + 1
   if next_epoch mod EPOCHS_PER_SYNC_COMMITTEE_PERIOD == 0:
     state.current_sync_committee = state.next_sync_committee
     state.next_sync_committee = get_next_sync_committee(state)
 
 # https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.2/specs/altair/beacon-chain.md#inactivity-scores
-func process_inactivity_updates*(cfg: RuntimeConfig, state: var altair.BeaconState) =
+func process_inactivity_updates*(cfg: RuntimeConfig, state: var (altair.BeaconState | merge.BeaconState)) =
```
tersec, line 858: There's a general question of whether to declare a post-phase0 state typeclass. Could make sense, but it might not buy much since the Nim bug requiring typeclasses to be repeatedly re-declared across modules.

It's also a relatively nonstructural cleanup that can be done later without cost or significant risk, so not something I'm that concerned about either way here.

beacon_chain/beacon_chain_db.nim
```diff
@@ -483,6 +483,9 @@ proc putBlock*(db: BeaconChainDB, value: altair.TrustedSignedBeaconBlock) =
   db.altairBlocks.putSnappySSZ(value.root.data, value)
   db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
 
+proc putBlock*(db: BeaconChainDB, value: merge.TrustedSignedBeaconBlock) =
+  doAssert false # TODO Deliberate -- put impl in separately
```
tersec, line 489: There will be a separate database PR

beacon_chain/validator_client/api.nim
```diff
@@ -770,6 +770,10 @@ proc publishBlock*(vc: ValidatorClientRef,
       publishBlock(it, data.phase0Block)
     of BeaconBlockFork.Altair:
       publishBlock(it, data.altairBlock)
+    of BeaconBlockFork.Merge:
+      raiseAssert "trying to publish merge block"
+      # TODO this doesn't build due to some nim-presto error
```
tersec, line 775:
```
nimbus-eth2/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim(92, 6) template/generic instantiation of `rest` from here
nimbus-eth2/vendor/nim-presto/presto/client.nim(284, 36) template/generic instantiation of `async` from here
nimbus-eth2/vendor/nim-presto/presto/client.nim(771, 32) Error: type mismatch: got <SignedBeaconBlock, string>
but expected one of:
proc encodeBytes[T: EncodeArrays](value: T; contentType: string): RestResult[
    seq[byte]]
  required type for value: T: EncodeArrays
  but expression 'body' is of type: SignedBeaconBlock
proc encodeBytes[T: EncodeTypes](value: T; contentType: string): RestResult[
    seq[byte]]
  required type for value: T: EncodeTypes
  but expression 'body' is of type: SignedBeaconBlock

expression: encodeBytes(body, restContentType)
```
Didn't seem important enough to block this PR. If there's an easy fix, that's good, but if not, this PR should probably go in without waiting for it.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
etan-status: Maybe because `ExecutionPayload` is still defined empty (`discard`).

beacon_chain/spec/state_transition_block.nim
```diff
@@ -502,6 +565,13 @@ proc process_block*(
   # The transition-triggering block creates, not acts on, an Altair state
   err("process_block: Altair state with Phase 0 block")
 
+proc process_block*(
+    cfg: RuntimeConfig,
+    state: var merge.BeaconState, blck: SomePhase0Block, flags: UpdateFlags,
+    cache: var StateCache): Result[void, cstring] {.nbench.} =
+  # The transition-triggering block creates, not acts on, an Altair state
```
etan-status: Merge not Altair.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -523,8 +593,47 @@ proc process_block*(
 
   ok()
 
+# Identical contents as altair/altair, but for merge; enables consistency
```
etan-status: altair/altair?

tersec: state/block hardfork names: its contents are identical, but whereas the other version matches Altair blocks with Altair states, this matches Merge blocks with merge states.

tests/mocking/mock_blocks.nim
```diff
@@ -61,6 +61,7 @@ proc mockBlock*(state: ForkedHashedBeaconState, slot: Slot): ForkedSignedBeaconB
   result.kind = case tmpState[].beaconStateFork
                 of forkPhase0: BeaconBlockFork.Phase0
                 of forkAltair: BeaconBlockFork.Altair
+                of forkMerge: BeaconBlockFork.Merge
```
etan-status: NIT: In the other files you aligned the content of the cases.

tersec: https://github.com/status-im/nimbus-eth2/pull/2890/commits/54bf6f8bb242a6c9bdabbdd2beabfec2699c8c9c

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/2890/commits/54bf6f8bb242a6c9bdabbdd2beabfec2699c8c9c

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
tersec: https://github.com/status-im/nimbus-eth2/blob/54bf6f8bb242a6c9bdabbdd2beabfec2699c8c9c/beacon_chain/spec/datatypes/merge.nim#L39-L55

etan-status: Ah, sorry, was confusing this with `ExecutionEngine` (that one's still empty, but not included in the block).

tersec: It's quite likely because merge blocks have a field of type `ExecutionPayload`, which has a field, `transactions`, which it doesn't know how to serialize for REST. That's worth doing, but isn't necessary for this PR, which avoids being more sprawling than it already is.

It's less harmful to disabling attempting to serialize what will fail, and instead stub out that part for this PR such that changes in exception specifications won't affect phase0/altair, so this PR is mergeable.

tersec: https://github.com/status-im/nimbus-eth2/pull/2890/commits/e22e2b54b76d499131ffd12e5c16842d45a41731 implements this.

PR #2894, opened 2021-09-24
sync committees don't exist in phase 0

tersec:
maybe maybe maybe


arnetheduck (approved):

tersec:
This PR is both a correctness fix and partway to bisecting the necessary conditions for CI finalization failure here. Merging anyway, for the former, but it's exhibiting the same issue.

https://github.com/status-im/nimbus-eth2/pull/2896 and https://github.com/status-im/nimbus-eth2/pull/2895 are the PRs to watch apropos the rest of the bisection.

PR #2899, opened 2021-09-24
Fix sync_committee_bits serialization.

cheatfate:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -489,6 +489,18 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: BitList) {.
      raises: [IOError, Defect].} =
   writeValue(writer, BitSeq value)
 
+## BitArray
+proc readValue*(reader: var JsonReader[RestJson], value: var BitArray) {.
+     raises: [IOError, SerializationError, Defect].} =
+  try:
+    hexToByteArray(readValue(reader, string), value.bytes)
+  except ValueError:
+    raiseUnexpectedValue(reader, "A BitArray value should be valid hex string")
```
etan-status: A few formality NITs for consistency with the rest of the file (e.g., `Eth2Digest` below):

1. `be a valid` with `a`.
2. `reader.readValue(` instead of `readValue(reader`.

tersec:
Can rebase to allow Jenkins CI to finalize.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
zah: Unfortunately, Nim has some quirks in the scope lookup rules in generics that introduce some very subtle differences between `x.foo()` and `foo(x)`. Most of the time, `foo(x)` will work in they way that you expect and using `x.foo()` may result in a compilation failure.

For this reason, you'll find a lot of code in the project sticking to the "safer" syntax.

arnetheduck: > Unfortunately, Nim has some quirks in the scope lookup rules in generics that introduce some very subtle differences 

are you able to describe those quirks with sufficient clarity that we can make a style guide entry out of them?

cheatfate: @zah right, some times its impossible to use `<firstarg>.<procedure/function>()`. For example when `firstarg` is an object and has field with name equal to `procedure/function` name.

etan-status: Thanks for the explanations. Both syntaxes compile fine for me in this specific case.
I just put the comment for reasons of consistency throughout the file, as the rest of the file prefers `x.foo()` syntax, i.e., searches for `reader.readValue()` in this file have 26 results, and `readValue(reader` has 3 other results, of which one is on a generic, and the other two are just wrappers without additional logic.

PR #2902, opened 2021-09-27
handle duplicate pubkeys in sync committee

etan-status:
When sync committee message handling was introduced in #2830, the edge
case of the same validator being selected multiple times as part of a
sync subcommittee was not covered. Not handling that edge case makes
sync contributions have a lower-than-expected participation rate as each
sync validator is only counted up through once per subcommittee.
This patch ensures that this edge case is properly covered.

### DIFF
--- a/tests/test_gossip_validation.nim
+++ b/tests/test_gossip_validation.nim
@@ -8,6 +8,8 @@
 {.used.}
 
 import
+  # Standard library
+  std/sequtils,
   # Status lib
   unittest2,
   chronicles, chronos,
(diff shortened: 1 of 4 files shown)


etan-status:
The new test fails to run on Windows, investigating.

```
[Suite] Extra
all_tests --xml:build/all_tests.xml --console failed; Aborting.
mingw32-make: *** [Makefile:257: test] Error 1
```

etan-status:
Windows problem was due to a missing `newClone`. Fixed now.

tests/test_gossip_validation.nim
```diff
@@ -170,3 +174,68 @@ suite "Gossip validation " & preset():
...
+            int(SLOTS_PER_EPOCH), false, cfg = cfg):
+          let added = 
+            case blck.kind
+            of BeaconBlockFork.Phase0:
+              const nilCallback = OnPhase0BlockAdded(nil)
+              dag.addRawBlock(quarantine, blck.phase0Block, nilCallback)
+            of BeaconBlockFork.Altair:
+              const nilCallback = OnAltairBlockAdded(nil)
+              dag.addRawBlock(quarantine, blck.altairBlock, nilCallback)
+          doAssert added.isOk()
```
tersec: `check: added.isOk()` is better when running in a testing context

zah (approved):

tersec (approved):

tests/test_gossip_validation.nim
etan-status: OK, also changing the other instance of `doAssert` further down in the test.

PR #2904, opened 2021-09-27
use errReject template everywhere

etan-status:
There were still a few instances that used the expansion of `errReject`
instead of using the template itself. It seems that those cases were
forgotten as part of other cleanups in #2809. Done now for readability.

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -158,7 +158,7 @@ template checkedReject(msg: cstring): untyped =
     # an internal consistency/correctness check only, and effectively never has
     # false positives. These don't, for example, arise from timeouts.
     raiseAssert $msg
-  err((ValidationResult.Reject, cstring msg))
+  errReject(msg)
 
 template checkedReject(error: (ValidationResult, cstring)): untyped =
   doAssert error[0] == ValidationResult.Reject
@@ -859,8 +859,7 @@ proc validateSignedContributionAndProof*(
   # [REJECT] contribution_and_proof.selection_proof selects the validator as an aggregator for the slot
   # i.e. is_sync_committee_aggregator(contribution_and_proof.selection_proof) returns True.
   if not is_sync_committee_aggregator(msg.message.selection_proof):
-    return err((ValidationResult.Reject, cstring(
-      "SignedContributionAndProof: invalid selection_proof")))
+    return errReject("SignedContributionAndProof: invalid selection_proof")
 
   block:
     # [IGNORE] The sync committee contribution is the first valid contribution


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -859,8 +859,7 @@ proc validateSignedContributionAndProof*(
   # [REJECT] contribution_and_proof.selection_proof selects the validator as an aggregator for the slot
   # i.e. is_sync_committee_aggregator(contribution_and_proof.selection_proof) returns True.
   if not is_sync_committee_aggregator(msg.message.selection_proof):
-    return err((ValidationResult.Reject, cstring(
-      "SignedContributionAndProof: invalid selection_proof")))
+    return errReject("SignedContributionAndProof: invalid selection_proof")
```
tersec, line 862: Arguably, `checkedReject` makes more sense here. It's not an exact semantic equivalence to the existing code, of course, but it fits the `checkedReject` criteria.

etan-status: Probably best to look at separately to avoid mixing mechanical rewrite from semantic change.

Would you recommend `checkedReject` for the other errors in this function as well?

tersec: Fair enough, regarding separation of types of changes.

Looking at the rest of `validateSignedContributionAndProof`, yes, I'd recommend the other `errReject`s become `checkedReject`s. They're all (signature verification, aggregation bit count, et cetera) indicative of logical errors which should themselves be fixed and thus which the local testnets and such should die on immediately, rather than timing glitches or mismatches it should try to recover from.

tersec (approved):

beacon_chain/gossip_processing/gossip_validation.nim
etan-status: Using `checkedReject`s here would require some minor adjustments, because the existing `checkedReject` assumes presence of an `AttestationPool` which is not available in `validateSignedContributionAndProof`. Furthermore, is there any scenario where a regular `errReject` should NOT be a `checkedReject`, throughout the entire file?

tersec: Yeah, `checkedReject` is something initially applied opportunistically, not in a systematic way, resulting in the fundamentally false dependency you note. It's not crucial either way, more of a debugging aid: rather than require people looking at what went wrong in a (primarily) CI or other local testnet to trawl through logs, this isolates the root cause for more efficient analysis. If it requires too much contortion, it's fine to leave things as `errReject`.

The primary criterion for when one actively shouldn't use `checkedReject` is in its comment block, that it needs to have a negligible transitory/ephemeral false-positive rate, and not be triggerable by timing quirks or lost network messages. It needs to be kinds of an independent property of each message. For the most part, the spec takes care of this by marking those other ones as ignore, rather than reject, but at least for a while, some nondeterministic rejections remained in the gossip validation which shouldn't become `checkedReject`.

PR #2710, opened 2021-07-07
Implement split preset/config support

arnetheduck:
This is the initial bulk refactor to introduce runtime config values in
a number of places, somewhat replacing the existing mechanism of loading
network metadata.

It still needs more work, this is the initial refactor that introduces
runtime configuration in some of the places that need it.

The PR changes the way presets and constants work, to match the spec. In
particular, a "preset" now refers to the compile-time configuration
while a "cfg" or "RuntimeConfig" is the dynamic part.

A single binary can support either mainnet or minimal, but not both.
Support for other presets has been removed completely (can be readded,
in case there's need).

There's a number of outstanding tasks:

* `SECONDS_PER_SLOT` still needs fixing
* loading custom runtime configs needs redoing
* checking constants against YAML file

PR #2905, opened 2021-09-27
make preset header comments consistent

etan-status:
Currently there is a mix of comment formats in the various preset files.
Altair presets have a leading header line, Phase0 presets don't.
Furthermore, the Altair `minimal` header refers to `mainnet` (#2710).
Updated all of the header lines to match the spec.

### DIFF
--- a/beacon_chain/spec/presets/mainnet/altair_preset.nim
+++ b/beacon_chain/spec/presets/mainnet/altair_preset.nim
@@ -1,5 +1,5 @@
 # Mainnet preset - Altair
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/mainnet/altair.yaml
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/mainnet/altair.yaml
 const
   # Updated penalty values
   # ---------------------------------------------------------------
--- a/beacon_chain/spec/presets/mainnet/phase0_preset.nim
+++ b/beacon_chain/spec/presets/mainnet/phase0_preset.nim
@@ -1,4 +1,5 @@
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/mainnet/phase0.yaml
+# Mainnet preset - Phase0
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/mainnet/phase0.yaml
 
 const
   #
--- a/beacon_chain/spec/presets/minimal/altair_preset.nim
+++ b/beacon_chain/spec/presets/minimal/altair_preset.nim
@@ -1,5 +1,5 @@
-# Mainnet preset - Altair
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/minimal/altair.yaml
+# Minimal preset - Altair
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/minimal/altair.yaml
 const
   # Updated penalty values
   # ---------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


tersec (approved):

beacon_chain/spec/presets/mainnet/altair_preset.nim
```diff
@@ -1,5 +1,5 @@
 # Mainnet preset - Altair
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/mainnet/altair.yaml
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/mainnet/altair.yaml
```
arnetheduck, line 2: fwiw, these tags document when the file was last checked for accuracy - when updating, it's important to also verify that nothing of note has changed - I have some vague recollection that some of these constants were still in flux when writing this file but I might be wrong

tersec: Indeed, should be double-checked first. In theory, important changes will get picked up by other tests, but that's supposed to be a backstop, rather than the only verification mechanism.

Also, regardless of this PR, there's going to be a major opportunity coming up to check all of these by hand, after supposedly in the coming days, v1.1.0 is released.

Even so, the usual mechanism I use depends on the existing-last-checked version being accurate: as a quick initial check, load up the previous and following version URLs, and just compare those parts of the spec. If they're the same, then I typically assume that whatever contortions or apparent inaccuracies exist there are intentional.

This has the issue that things can drift without being documented as such, but also means that one isn't hand-checking the sometimes nontrivial equivalence of some spec Python code and what's in Nimbus. For just constants lists, though, it doesn't end up mattering much.

tersec: https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/mainnet/altair.yaml and https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/mainnet/altair.yaml are identical though.

beacon_chain/spec/presets/mainnet/phase0_preset.nim
```diff
@@ -1,4 +1,5 @@
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/mainnet/phase0.yaml
+# Mainnet preset - Phase0
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/mainnet/phase0.yaml
```
tersec, line 3: identical indeed

beacon_chain/spec/presets/minimal/altair_preset.nim
```diff
@@ -1,5 +1,5 @@
-# Mainnet preset - Altair
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/minimal/altair.yaml
+# Minimal preset - Altair
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/minimal/altair.yaml
 const
```
tersec, line 3: identical

beacon_chain/spec/presets/minimal/phase0_preset.nim
```diff
@@ -1,4 +1,5 @@
-# https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/presets/minimal/phase0.yaml
+# Minimal preset - Phase0
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0-beta.5/presets/minimal/phase0.yaml
```
tersec, line 2: identical

beacon_chain/spec/presets/mainnet/altair_preset.nim
etan-status: Yeah, I also loaded up the new version side by side and compared that the constants are all still the same.

PR #2906, opened 2021-09-27
Revert "fixes"

tersec:
Reverts status-im/nimbus-eth2#2901

don't merge, for now

Check for if it's breakage cause

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -17,9 +17,7 @@ import
   ./spec/datatypes/[phase0, altair],
   ./filepath
 
-export
-  phase0, altair, eth2_ssz_serialization, eth2_merkleization, kvstore,
-  kvstore_sqlite3
+export phase0, altair, eth2_ssz_serialization, eth2_merkleization
 
 logScope: topics = "bc_db"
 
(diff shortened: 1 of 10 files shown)


arnetheduck:
fwiw, the linux tests finalized, so this looks like an unlikely candidate (there are no material changes in it - the rest deserialization affects only beaconstate which is never sent across the wire in  the fin test

tersec:
Not for https://github.com/status-im/nimbus-eth2/commit/e47a8cbe4242a37e01b402ce5d77853394c6971e -- both Jenkins platforms failed.

tersec:
I want to avoid post hoc ergo propter hoc here though; e.g., Jenkins hosts sometimes fail for a while for other reasons, and some commit's the last working one/first non-working-looking one.

Watching Jenkins runs from https://github.com/status-im/nimbus-eth2/commit/5bfff4378574f6fc84e5508607c17b0a38160d28 and if they're fine, then it was very likely just a glitch. Also rebased https://github.com/status-im/nimbus-eth2/pull/2890 to resolve conflicts, so watching that for similar reasons.

tersec:
https://github.com/status-im/nimbus-eth2/commit/5bfff4378574f6fc84e5508607c17b0a38160d28 is fine, so probably it's just a glitch.

PR #2911, opened 2021-09-28
merge hardfork database support

tersec:


beacon_chain/spec/state_transition.nim
```diff
@@ -242,6 +254,7 @@ proc process_slots*(
         state.hbsMerge.root = hash_tree_root(state)
 
     maybeUpgradeStateToAltair(cfg, state)
+    maybeUpgradeStateToMerge(cfg, state)
```
etan-status: `teststateutil` and `mock_genesis` also need this.

etan-status: and `mock_blocks` needs the post-merge executionpayload initialization for empty blocks.

tersec: Might be better to encapsulate it there's basically no caller that wants only `maybeUpgradeStateToAltair()` or `maybeUpgradeStateToMerge()`, if both apply. The spec is clear that they can be in the same epoch/one-directly-after-another, too, so there's no need to prevent that.

I removed the `teststateutil` one because `process_slots()` handles it already, but `mock_genesis` needs adjusting indeed.

etan-status: true, a `maybeUpgradeState` wrapping both would be useful.

tersec: As far as post-merge `ExecutionPayload` initialization, these PRs are intentionally doing kind of minimal-reasonable-approaches to things. Otherwise, it'd become a larger, sprawling PR, the intent being to avoid.

There's room for middle grounds, of course, and addressing other `maybeUpgradeStateToAltair()` callers is within a reasonable scope, but the only reason `maybeUpgradeStateToMerge()` is included at all is to provide a convenient source of merge states. `ExecutionPayload` is irrelevant to the database side except insofar as it its SSZ serialization and deserializations need to function (tested by SSZ consensus tests).

tersec: The rebase addresses these points.

etan-status: Regarding `mock_blocks` needs the post-merge initialization for empty blocks:
I created a PR for that here: https://github.com/status-im/nimbus-eth2/pull/2963

PR #2912, opened 2021-09-28
simplify state fork access pattern

arnetheduck:


beacon_chain/spec/forks.nim
```diff
@@ -149,62 +161,31 @@ func assign*(tgt: var ForkedHashedBeaconState, src: ForkedHashedBeaconState) =
...
   # ```
   #   for index, validator in getStateField(stateData.data, validators).pairs():
   # ```
   # Without `unsafeAddr`, the `validators` list would be copied to a temporary variable.
-  (case x.beaconStateFork
-   of forkPhase0: unsafeAddr (x.hbsPhase0.data.y)
-   of forkAltair: unsafeAddr (x.hbsAltair.data.y)
-   of forkMerge:  unsafeAddr (x.hbsMerge.data.y))[]
+  withState(x):
+    (unsafeAddr state.data.y)[]
```
etan-status: This puts the `[]` inside the `case`, does it still work like this?

beacon_chain/spec/forks.nim
```diff
@@ -133,7 +133,19 @@ template init*(T: type ForkedTrustedSignedBeaconBlock, blck: altair.TrustedSigne
 template init*(T: type ForkedTrustedSignedBeaconBlock, blck: merge.TrustedSignedBeaconBlock): T =
   T(kind: BeaconBlockFork.Merge,  mergeBlock: blck)
 
-# State-related functionality based on ForkedHashedBeaconState instead of BeaconState
+# State-related functionality based on ForkedHashedBeaconState instead of HashedBeaconState
+
+template withState*(fhbs: ForkedHashedBeaconState, body: untyped): untyped =
```
etan-status: note that DAG also has a `withState` template.

beacon_chain/spec/forks.nim
arnetheduck: cc @tersec any idea? I'm not sure tbh

beacon_chain/spec/forks.nim
arnetheduck: that's generally fine, though if you can think of a better name..

etan-status: yeah, it compiles fine :) more mentioning it to highlight that there is now the possibility for `withState` inception :)

beacon_chain/spec/forks.nim
arnetheduck: good catch, just checked the generated code and with the deref inside the case, a copy is generated in at least one place it wasn't before

tersec: The invariant to maintain is that at no point except a function/template/macro exit does there exist a value of this sort that's not a pointer.

etan-status: maybe:

```
(
  withState(x):
    unsafeAddr (state.data.y)
)[]
```
?

tersec: For example, it compiles if expressions similarly to:
```nim
var res: SomeType
if cond:
  res = val1    # bad
else:
  res = val2    # also bad
```
and case expressions analogously.

etan-status: or even
```
(withState(x): unsafeAddr (state.data.y))[]
```
(have not tried it)

tersec: Sure, anything with the dereferencing operator as outermost as one can get it should be okay, pending C/assembly-generated-code verification.

tersec: The upside and downside of Nim's code generation here is that it's neither clever nor unpredictable.

tersec:
As a potential follow-up to this PR, it's possible that plain `withState` is a convenient enough pattern some of the items used just once or twice could/should be removed. Not that important either way, but it would encourage a style which doesn't insert the `getStateField`-type conditional macro every statement/expression, but only once per larger code blocks.

beacon_chain/spec/forks.nim
arnetheduck: >  nor unpredictable

I would contest this point

reverted to an extra case statement - too many nested templates give complex error messages I don't want to investigate right now

arnetheduck:
alright, nailed a few more cleanups, this branch should be gtg

beacon_chain/spec/forks.nim
tersec: Seems like it's working, e.g.,
```c
#line 484 "nimbus-eth2/beacon_chain/nimbus_beacon_node.nim"
N_LIB_PRIVATE N_NIMCALL(void, getAttachedValidators__0OByA5QvJyPTA7s08vxY3Q)(tyObject_BeaconNodecolonObjectType___oo06t121DPxvFpf8tLBYhQ* node, tyObject_Table__29b3vF3sUiGKM7GqZVRo4JQ* Result) {	chckNil((void*)Result);	unsureAsgnRef((void**)&(*Result).data, NIM_NIL);	(*Result).counter = 0;	{		NI validatorIndex;		NI colontmp_;		tyObject_HashList__M609cw2woH7OaDSK8uR8ieg* T2_;		NI T6_;		NI i;		validatorIndex = (NI)0;		colontmp_ = (NI)0;
#line 487 "nimbus-eth2/beacon_chain/nimbus_beacon_node.nim"

#line 487 "nimbus-eth2/beacon_chain/nimbus_beacon_node.nim"

#line 171 "nimbus-eth2/beacon_chain/spec/forks.nim"
		T2_ = (tyObject_HashList__M609cw2woH7OaDSK8uR8ieg*)0;		switch ((*(*node).dag).headState.data.beaconStateFork) {
		case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 0):
		{			if (!(((1 &((NU8)1<<((NU)((*(*node).dag).headState.data.beaconStateFork)&7U)))!=0))){ raiseFieldError(((NimStringDesc*) &TM__Kq8yS6n9bGw6oLFM14s9a89cw_132)); }			T2_ = (&(*(*node).dag).headState.data.hbsPhase0.data.validators);		}
		break;
		case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 1):
		{			if (!(((2 &((NU8)1<<((NU)((*(*node).dag).headState.data.beaconStateFork)&7U)))!=0))){ raiseFieldError(((NimStringDesc*) &TM__Kq8yS6n9bGw6oLFM14s9a89cw_133)); }			T2_ = (&(*(*node).dag).headState.data.hbsAltair.data.validators);		}
		break;
		case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw) 2):
		{			if (!(((4 &((NU8)1<<((NU)((*(*node).dag).headState.data.beaconStateFork)&7U)))!=0))){ raiseFieldError(((NimStringDesc*) &TM__Kq8yS6n9bGw6oLFM14s9a89cw_134)); }			T2_ = (&(*(*node).dag).headState.data.hbsMerge.data.validators);		}
		break;
		}
		T6_ = ((*T2_).data ? (*T2_).data->Sup.len : 0);		colontmp_ = T6_;
#line 106 "nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/iterators_1.nim"
		i = ((NI) 0);		{
#line 107 "nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/iterators_1.nim"
			while (1) {				NI TM__Kq8yS6n9bGw6oLFM14s9a89cw_356;
#line 107 "nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/iterators_1.nim"
				if (!(i < colontmp_)) goto LA8;
```

tersec (approved):

PR #2913, opened 2021-09-28
Improve discovery during fork

Menduist:
This PR allows nimbus to dial peers:
- Which are aware of a fork (when we are not)
- Which are not aware of a fork (when we are)

It does so by adding a `discoveryForkId`, which is the "minimum allowed fork id". Until 1 epoch from a fork, it will be set to the non-fork-aware value. At fork, it will be set to the new fork id.

Any peer discovered needs to have a forkId >= to the discoveryForkId, otherwise it won't be dialed.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -75,6 +75,7 @@ type
     connWorkers: seq[Future[void]]
     connTable: HashSet[PeerID]
     forkId*: ENRForkID
+    discoveryForkId*: ENRForkID
     forkDigests*: ForkDigestsRef
     rng*: ref BrHmacDrbgContext
     peers*: Table[PeerID, Peer]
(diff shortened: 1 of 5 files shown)


Menduist:
@arnetheduck could check if this make sense in fork terms

beacon_chain/networking/eth2_network.nim
```diff
@@ -899,6 +900,22 @@ proc toPeerAddr(node: Node): Result[PeerAddr, cstring] =
   let peerAddr = ? nodeRecord.toPeerAddr(tcpProtocol)
   ok(peerAddr)
 
+proc isCompatibleForkId*(discoveryForkId: ENRForkID, peerForkId: ENRForkID): bool =
```
kdeme, line 903: I just want to point out that I found this quite confusing and initially thought it was faulty, more specifically the part:
```
   else:
      # Our next fork version is bigger than the peer's one
      false
```
Only realized afterwards because of what is done at `updateForkId` that is seems to be correct.
I guess the confusing part is mostly in the naming considering `discoveryForkId` and `ENRForkId` or `peerForkId` sounds like you are comparing the same things, but you are in fact not, the first is older, not updated until fork epoch - 1.

When I was thinking about this functionality, I thought about comparing the `current_fork_version` with the `next_fork_version`, but annoyingly the `current_fork_version` is not directly available in the `ENRForkID`.

Menduist: Yeah, the nice thing about having a different `discoveryForkId` is that you can choose at which moment to block the dialing of non-compatible peers (though 1 epoch in advance is a bit useless, something like 5h in advance would be more interesting)

Though it can be confusing, and I took care in using `discoveryForkId` instead of `forkId` to be clear about the fact that it is not our forkId, but that may not be enough. If you have a better name (or other solution) in mind, I'm happy to hear it. Otherwise I'll just add some comments to highlight that

zah:
Merging this now, so we can continue making progress towards testing our Altair release. Otherwise, indeed, the naming of the variables can be improved in follow-up PRs to make it easier to understand the intention of the code.

PR #2917, opened 2021-09-28
use `allSyncCommittees` everywhere

etan-status:
There are still a few cases with manual loops through sync subcommittees
even though an `allSyncCommittees` iterator exists. Adjusted the few
remaining instances to also use the iterator instead.

### DIFF
--- a/beacon_chain/spec/datatypes/altair.nim
+++ b/beacon_chain/spec/datatypes/altair.nim
@@ -426,6 +426,8 @@ template asInt*(x: SyncCommitteeIndex): int = int(x)
 template asUInt8*(x: SyncCommitteeIndex): uint8 = uint8(x)
 template asUInt64*(x: SyncCommitteeIndex): uint64 = uint64(x)
 
+template `[]`*(a: auto; i: SyncCommitteeIndex): auto = a[i.asInt]
+
 template `==`*(x, y: SyncCommitteeIndex): bool =
   distinctBase(x) == distinctBase(y)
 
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/datatypes/altair.nim
```diff
@@ -426,14 +426,18 @@ template asInt*(x: SyncCommitteeIndex): int = int(x)
...
   distinctBase(x) == distinctBase(y)
 
 iterator allSyncCommittees*: SyncCommitteeIndex =
   for committeeIdx in 0 ..< SYNC_COMMITTEE_SUBNET_COUNT:
     yield SyncCommitteeIndex(committeeIdx)
 
-template validateSyncCommitteeIndexOr*(networkValParam: uint64, elseBody: untyped) =
+template validateSyncCommitteeIndexOr*(
+    networkValParam: uint64, 
+    elseBody: untyped): SyncCommitteeIndex =
```
etan-status, line 440: Not sure how this ever worked without the return type.

tersec (approved):

beacon_chain/spec/datatypes/altair.nim
tersec: Might be worth filing an upstream bug around this, if it's indeed a bug and not some more subtle-but-correct effect.

etan-status: Hmm, I'm not sure if it is a bug or if we just got lucky before. Without the return type this failed for the new location where this template is used, but for the existing instantiation from before, compilation still succeeded without the return type.

etan-status: Context: For the new instantiation I got error because the returned `SyncCommitteeIndex` needed to be discarded / that it was unused. I guess this was because templates by default are clean / don't inject any values, so by either making the return type explicit or by adding {.inject.} it suppresses the error. 

But it's weird because in both cases the usage was same. Maybe it's the chain from an Optional's get() directly into this template that led to the error message. I did not spend too much time on researching it.

PR #2919, opened 2021-09-29
Altair config for mainnet

zah:
This still doesn't work properly because it leads to a change in the "eth" ENR field of the client. Since the client uses the value of this field to look for other nodes on the network, the change effectively prevents us from finding peers.


zah:
Works when rebased on top of https://github.com/status-im/nimbus-eth2/pull/2913 (which is already merged)

PR #2920, opened 2021-09-29
fix multithreading memory leak

mratsim:
The std/tasks destructor wasn't enforced:

https://github.com/status-im/nim-taskpools/commit/e1fa0ec097ec3fc6fe7b5f6997377b84aea4a691


arnetheduck:
we shouldn't be using destructors at all, they're broken in nim 1.2

mratsim:
It's in std/tasks (and my shim of it for 1.2) so that we can migrate seamlessly to 1.6:

- Stdlib https://github.com/nim-lang/Nim/blob/devel/lib/std/tasks.nim
- 99% copy-paste: https://github.com/status-im/nim-taskpools/blob/master/taskpools/shims_pre_1_6/tasks.nim

PR #2921, opened 2021-09-29
add remaining REST spec references

etan-status:
This adds spec references for a few APIs that were missing them so far,
in line with how spec references already exist for the other REST APIs.


etan-status:
Those seem to have been forgotten in #2756

PR #2925, opened 2021-09-29
Validator monitoring

arnetheduck:
Validator monitoring based on and mostly compatible with the
implementation in Lighthouse - tracks additional logs and metrics for
specified validators so as to stay on top on performance.

The implementation works more or less the following way:
* Validator pubkeys are singled out for monitoring - these can be
running on the node or not
* For every action that the validator takes, we record steps in the
process such as messages being seen on the network or published in the
API
* When the dust settles at the end of an epoch, we report the
information from one epoch before that, which coincides with the
balances being updated - this is a tradeoff between being correct
(waiting for finalization) and providing relevant information in a
timely manner)

TODO:
- [x] completed sync committee monitoring
- [x] altair rewards
- [x] docs

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -112,6 +112,9 @@ type
     # ----------------------------------------------------------------
     blockProcessor: ref BlockProcessor
 
+    # Validator monitoring
+    validatorMonitor: ref ValidatorMonitor
+
     # Validated with no further verification required
     # ----------------------------------------------------------------
     exitPool: ref ExitPool
@@ -135,6 +138,7 @@ type
 proc new*(T: type Eth2Processor,
           doppelGangerDetectionEnabled: bool,
           blockProcessor: ref BlockProcessor,
+          validatorMonitor: ref ValidatorMonitor,
           dag: ChainDAGRef,
           attestationPool: ref AttestationPool,
           exitPool: ref ExitPool,
@@ -150,6 +154,7 @@ proc new*(T: type Eth2Processor,
     doppelgangerDetection: DoppelgangerProtection(
       nodeLaunchSlot: getBeaconTime().slotOrZero),
     blockProcessor: blockProcessor,
+    validatorMonitor: validatorMonitor,
     dag: dag,
     attestationPool: attestationPool,
     exitPool: exitPool,
@@ -171,9 +176,8 @@ proc new*(T: type Eth2Processor,
 # could be used to push out valid messages.
 
 proc blockValidator*(
-    self: var Eth2Processor,
-    signedBlock: phase0.SignedBeaconBlock | altair.SignedBeaconBlock |
-                 merge.SignedBeaconBlock): ValidationRes =
+    self: var Eth2Processor, src: MsgSource,
+    signedBlock: ForkySignedBeaconBlock): ValidationRes =
   let
     wallTime = self.getCurrentBeaconTime()
     (afterGenesis, wallSlot) = wallTime.toSlot()
@@ -207,7 +211,7 @@ proc blockValidator*(
     trace "Block validated"
 
     self.blockProcessor[].addBlock(
-      ForkedSignedBeaconBlock.init(signedBlock),
+      src, ForkedSignedBeaconBlock.init(signedBlock),
       validationDur = self.getCurrentBeaconTime() - wallTime)
 
     # Validator monitor registration for blocks is done by the processor
@@ -250,9 +254,8 @@ proc checkForPotentialDoppelganger(
         quit QuitFailure
 
 proc attestationValidator*(
-    self: ref Eth2Processor,
-    attestation: Attestation,
-    subnet_id: SubnetId,
+    self: ref Eth2Processor, src: MsgSource,
+    attestation: Attestation, subnet_id: SubnetId,
     checkSignature: bool = true): Future[ValidationRes] {.async.} =
   let wallTime = self.getCurrentBeaconTime()
   var (afterGenesis, wallSlot) = wallTime.toSlot()
@@ -286,8 +289,12 @@ proc attestationValidator*(
     self.attestationPool[].addAttestation(
       attestation, [attester_index], sig, wallSlot)
 
+    self.validatorMonitor[].registerAttestation(
+      src, wallTime, attestation, attester_index)
+
     beacon_attestations_received.inc()
     beacon_attestation_delay.observe(delay.toFloatSeconds())
+
     ok()
   else:
     debug "Dropping attestation", validationError = v.error
@@ -295,7 +302,7 @@ proc attestationValidator*(
     err(v.error())
 
 proc aggregateValidator*(
-    self: ref Eth2Processor,
+    self: ref Eth2Processor, src: MsgSource,
     signedAggregateAndProof: SignedAggregateAndProof): Future[ValidationRes] {.async.} =
   let wallTime = self.getCurrentBeaconTime()
   var (afterGenesis, wallSlot) = wallTime.toSlot()
@@ -334,6 +341,9 @@ proc aggregateValidator*(
     self.attestationPool[].addAttestation(
       signedAggregateAndProof.message.aggregate, attesting_indices, sig, wallSlot)
 
+    self.validatorMonitor[].registerAggregate(
+      src, wallTime, signedAggregateAndProof, attesting_indices)
+
     beacon_aggregates_received.inc()
     beacon_aggregate_delay.observe(delay.toFloatSeconds())
 
@@ -345,8 +355,8 @@ proc aggregateValidator*(
     err(v.error())
 
 proc attesterSlashingValidator*(
-    self: var Eth2Processor, attesterSlashing: AttesterSlashing):
-    ValidationRes =
+    self: var Eth2Processor, src: MsgSource,
+    attesterSlashing: AttesterSlashing): ValidationRes =
   logScope:
     attesterSlashing = shortLog(attesterSlashing)
 
@@ -359,6 +369,8 @@ proc attesterSlashingValidator*(
 
     self.exitPool[].addMessage(attesterSlashing)
 
+    self.validatorMonitor[].registerAttesterSlashing(src, attesterSlashing)
+
     beacon_attester_slashings_received.inc()
   else:
     debug "Dropping attester slashing", validationError = v.error
@@ -367,8 +379,8 @@ proc attesterSlashingValidator*(
   v
 
 proc proposerSlashingValidator*(
-    self: var Eth2Processor, proposerSlashing: ProposerSlashing):
-    Result[void, ValidationError] =
+    self: var Eth2Processor, src: MsgSource,
+    proposerSlashing: ProposerSlashing): Result[void, ValidationError] =
   logScope:
     proposerSlashing = shortLog(proposerSlashing)
 
@@ -377,18 +389,21 @@ proc proposerSlashingValidator*(
   let v = self.exitPool[].validateProposerSlashing(proposerSlashing)
   if v.isOk():
     trace "Proposer slashing validated"
-    beacon_proposer_slashings_received.inc()
-  else:
-    debug "Dropping proposer slashing", validationError = v.error
 
     self.exitPool[].addMessage(proposerSlashing)
 
+    self.validatorMonitor[].registerProposerSlashing(src, proposerSlashing)
+
+    beacon_proposer_slashings_received.inc()
+  else:
+    debug "Dropping proposer slashing", validationError = v.error
     beacon_proposer_slashings_dropped.inc(1, [$v.error[0]])
+
   v
 
 proc voluntaryExitValidator*(
-    self: var Eth2Processor, signedVoluntaryExit: SignedVoluntaryExit):
-    Result[void, ValidationError] =
+    self: var Eth2Processor, src: MsgSource,
+    signedVoluntaryExit: SignedVoluntaryExit): Result[void, ValidationError] =
   logScope:
     signedVoluntaryExit = shortLog(signedVoluntaryExit)
 
@@ -400,6 +415,9 @@ proc voluntaryExitValidator*(
 
     self.exitPool[].addMessage(signedVoluntaryExit)
 
+    self.validatorMonitor[].registerVoluntaryExit(
+      src, signedVoluntaryExit.message)
+
     beacon_voluntary_exits_received.inc()
   else:
     debug "Dropping voluntary exit", error = v.error
@@ -408,7 +426,7 @@ proc voluntaryExitValidator*(
   v
 
 proc syncCommitteeMessageValidator*(
-    self: ref Eth2Processor,
+    self: ref Eth2Processor, src: MsgSource,
     syncCommitteeMsg: SyncCommitteeMessage,
     subcommitteeIdx: SyncSubcommitteeIndex,
     checkSignature: bool = true): Future[Result[void, ValidationError]] {.async.} =
@@ -441,6 +459,9 @@ proc syncCommitteeMessageValidator*(
       subcommitteeIdx,
       positions)
 
+    self.validatorMonitor[].registerSyncCommitteeMessage(
+      src, wallTime, syncCommitteeMsg)
+
     beacon_sync_committee_messages_received.inc()
 
     ok()
@@ -450,7 +471,7 @@ proc syncCommitteeMessageValidator*(
     err(v.error())
 
 proc contributionValidator*(
-    self: ref Eth2Processor,
+    self: ref Eth2Processor, src: MsgSource,
     contributionAndProof: SignedContributionAndProof,
     checkSignature: bool = true): Future[Result[void, ValidationError]] {.async.} =
   let
@@ -475,7 +496,12 @@ proc contributionValidator*(
 
   return if v.isOk():
     trace "Contribution validated"
-    self.syncCommitteeMsgPool[].addContribution(contributionAndProof, v.get)
+    self.syncCommitteeMsgPool[].addContribution(
+      contributionAndProof, v.get()[0])
+
+    self.validatorMonitor[].registerSyncContribution(
+      src, wallTime, contributionAndProof, v.get()[1])
+
     beacon_sync_committee_contributions_received.inc()
 
     ok()
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -856,8 +856,7 @@ proc validateContribution*(
     msg: SignedContributionAndProof,
     wallTime: BeaconTime,
     checkSignature: bool):
-    Future[Result[CookedSig, ValidationError]] {.async.} =
-
+    Future[Result[(CookedSig, seq[ValidatorIndex]), ValidationError]] {.async.} =
   let
     syncCommitteeSlot = msg.message.contribution.slot
 
@@ -900,6 +899,12 @@ proc validateContribution*(
     # [REJECT] The contribution has participants
     # that is, any(contribution.aggregation_bits).
     return errReject("SignedContributionAndProof: aggregation bits empty")
+
+  # TODO we take a copy of the participants to avoid the data going stale
+  #      between validation and use - nonetheless, a design that avoids it and
+  #      stays safe would be nice
+  let participants = dag.syncCommitteeParticipants(
+    msg.message.contribution.slot, subcommitteeIdx)
 
   let sig = if checkSignature:
     let deferredCrypto = batchCrypto.scheduleContributionChecks(
@@ -951,4 +956,4 @@ proc validateContribution*(
       return errReject("SyncCommitteeMessage: unable to load signature")
     sig.get()
 
-  return ok(sig)
+  return ok((sig, participants))
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -543,6 +543,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
         request.slot, subnet_id, request.validator_index,
         request.is_aggregator)
 
+      let validator_pubkey = getStateField(
+        node.dag.headState.data, validators).asSeq()[request.validator_index].pubkey
+
+      node.validatorMonitor[].addAutoMonitor(
+        validator_pubkey, ValidatorIndex(request.validator_index))
+
     return RestApiResponse.jsonMsgResponse(BeaconCommitteeSubscriptionSuccess)
 
   # https://ethereum.github.io/beacon-APIs/#/Validator/prepareSyncCommitteeSubnets
@@ -569,6 +575,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
              lenu64(getStateField(node.dag.headState.data, validators)):
             return RestApiResponse.jsonError(Http400,
                                              InvalidValidatorIndexValueError)
+          let validator_pubkey = getStateField(
+            node.dag.headState.data, validators).asSeq()[item.validator_index].pubkey
+
+          node.validatorMonitor[].addAutoMonitor(
+            validator_pubkey, ValidatorIndex(item.validator_index))
+
         subs
 
     warn "Sync committee subscription request served, but not implemented"
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -218,7 +218,7 @@ proc sendAttestation*(
   # libp2p calls the data handler for any subscription on the subnet
   # topic, it does not perform validation.
   let res = await node.processor.attestationValidator(
-    attestation, subnet_id, checkSignature)
+    MsgSource.api, attestation, subnet_id, checkSignature)
 
   return
     if res.isGoodForSending:
@@ -241,8 +241,9 @@ proc sendSyncCommitteeMessage*(
   # validation will also register the message with the sync committee
   # message pool. Notably, although libp2p calls the data handler for
   # any subscription on the subnet topic, it does not perform validation.
-  let res = await node.processor.syncCommitteeMessageValidator(msg, subcommitteeIdx,
-                                                         checkSignature)
+  let res = await node.processor.syncCommitteeMessageValidator(
+    MsgSource.api, msg, subcommitteeIdx, checkSignature)
+
   return
     if res.isGoodForSending:
       node.network.broadcastSyncCommitteeMessage(msg, subcommitteeIdx)
@@ -340,7 +341,7 @@ proc sendSyncCommitteeContribution*(
     msg: SignedContributionAndProof,
     checkSignature: bool): Future[SendResult] {.async.} =
   let res = await node.processor.contributionValidator(
-    msg, checkSignature)
+    MsgSource.api, msg, checkSignature)
 
   return
     if res.isGoodForSending:
@@ -571,7 +572,7 @@ proc proposeBlock(node: BeaconNode,
       # storeBlock puts the block in the chaindag, and if accepted, takes care
       # of side effects such as event api notification
       newBlockRef = node.blockProcessor[].storeBlock(
-        signedBlock, wallTime.slotOrZero())
+        MsgSource.api, wallTime, signedBlock)
 
     if newBlockRef.isErr:
       warn "Unable to add proposed block to block pool",
@@ -1156,7 +1157,7 @@ proc sendAggregateAndProof*(node: BeaconNode,
                             proof: SignedAggregateAndProof): Future[SendResult] {.
      async.} =
   # REST/JSON-RPC API helper procedure.
-  let res = await node.processor.aggregateValidator(proof)
+  let res = await node.processor.aggregateValidator(MsgSource.api, proof)
   return
     if res.isGoodForSending:
       node.network.broadcastAggregateAndProof(proof)
@@ -1176,7 +1177,7 @@ proc sendAggregateAndProof*(node: BeaconNode,
 proc sendVoluntaryExit*(node: BeaconNode,
                         exit: SignedVoluntaryExit): SendResult =
   # REST/JSON-RPC API helper procedure.
-  let res = node.processor[].voluntaryExitValidator(exit)
+  let res = node.processor[].voluntaryExitValidator(MsgSource.api, exit)
   if res.isGoodForSending:
     node.network.broadcastVoluntaryExit(exit)
     ok()
@@ -1188,7 +1189,7 @@ proc sendVoluntaryExit*(node: BeaconNode,
 proc sendAttesterSlashing*(node: BeaconNode,
                            slashing: AttesterSlashing): SendResult =
   # REST/JSON-RPC API helper procedure.
-  let res = node.processor[].attesterSlashingValidator(slashing)
+  let res = node.processor[].attesterSlashingValidator(MsgSource.api, slashing)
   if res.isGoodForSending:
     node.network.broadcastAttesterSlashing(slashing)
     ok()
@@ -1200,7 +1201,7 @@ proc sendAttesterSlashing*(node: BeaconNode,
 proc sendProposerSlashing*(node: BeaconNode,
                            slashing: ProposerSlashing): SendResult =
   # REST/JSON-RPC API helper procedure.
-  let res = node.processor[].proposerSlashingValidator(slashing)
+  let res = node.processor[].proposerSlashingValidator(MsgSource.api, slashing)
   if res.isGoodForSending:
     node.network.broadcastProposerSlashing(slashing)
     ok()
@@ -1236,7 +1237,7 @@ proc sendBeaconBlock*(node: BeaconNode, forked: ForkedSignedBeaconBlock
     wallTime = node.beaconClock.now()
     accepted = withBlck(forked):
       let newBlockRef = node.blockProcessor[].storeBlock(
-        blck, wallTime.slotOrZero())
+        MsgSource.api, wallTime, blck)
 
       # The boolean we return tells the caller whether the block was integrated
       # into the chain
--- a/beacon_chain/validators/validator_monitor.nim
+++ b/beacon_chain/validators/validator_monitor.nim
@@ -1,0 +1,749 @@
+import
+  std/[options, tables],
+  metrics, chronicles,
+  ../spec/[crypto, beaconstate, forks, helpers, presets],
+  ../spec/datatypes/[phase0, altair],
+  ../beacon_clock
+
+logScope: topics = "val_mon"
+
+# Validator monitoring based on the same feature in Lighthouse - using the same
+# metrics allows users to more easily reuse monitoring setups
+
+declareGauge validator_monitor_balance_gwei,
+  "The validator's balance in gwei.", labels = ["validator"]
+declareGauge validator_monitor_effective_balance_gwei,
+  "The validator's effective balance in gwei.", labels = ["validator"]
+declareGauge validator_monitor_slashed,
+  "Set to 1 if the validator is slashed.", labels = ["validator"]
+declareGauge validator_monitor_active,
+  "Set to 1 if the validator is active.", labels = ["validator"]
+declareGauge validator_monitor_exited,
+  "Set to 1 if the validator is exited.", labels = ["validator"]
+declareGauge validator_monitor_withdrawable,
+  "Set to 1 if the validator is withdrawable.", labels = ["validator"]
+declareGauge validator_activation_eligibility_epoch,
+  "Set to the epoch where the validator will be eligible for activation.", labels = ["validator"]
+declareGauge validator_activation_epoch,
+  "Set to the epoch where the validator will activate.", labels = ["validator"]
+declareGauge validator_exit_epoch,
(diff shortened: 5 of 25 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -483,52 +483,27 @@ proc proposeSignedBlock*(node: BeaconNode,
...
-            newBlock = blck.message, root = blck.root
-    return head
+  let wallTime = node.beaconClock.now()
+
+  return withBlck(newBlock):
+    let newBlockRef = node.blockProcessor[].storeBlock(
+      MsgSource.api, wallTime, blck)
+
+    if newBlockRef.isErr:
+      withBlck(newBlock):
```
zah: You are already within a `withBlock` block here, so this extra one is unnecessary.

arnetheduck:
Altair accounting in https://github.com/status-im/nimbus-eth2/pull/2981

arnetheduck:
```
INF 2021-11-17 17:33:19.000+01:00 Slot start                                 topics="beacnde" slot=2622166 epoch=81942 sync=synced peers=32 head=8b9ca1d3:2622164 finalized=81940:d6ce48c2 delay=330us95ns
INF 2021-11-17 17:33:19.219+01:00 Attestation included in block              attestationData="(slot: 2622165, index: 14, beacon_block_root: \"8b9ca1d3\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\")" inclusion_lag_slots=0 validator=b44be2ad
INF 2021-11-17 17:33:20.240+01:00 Unaggregated attestation                   topics="val_mon" attestation="(aggregation_bits: 0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"b3b556f7\")" validator=b173ac90
NOT 2021-11-17 17:33:20.240+01:00 Attestation sent                           topics="beacval" attestation="(aggregation_bits: 0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"b3b556f7\")" validator=b173ac90 delay=-2s759ms297us722ns subnet_id=10
INF 2021-11-17 17:33:20.245+01:00 Unaggregated attestation                   topics="val_mon" attestation="(aggregation_bits: 0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"b874f3ab\")" validator=b0348547
NOT 2021-11-17 17:33:20.245+01:00 Attestation sent                           topics="beacval" attestation="(aggregation_bits: 0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"b874f3ab\")" validator=b0348547 delay=-2s754ms730us665ns subnet_id=18
INF 2021-11-17 17:33:27.189+01:00 Slot end                                   topics="beacnde" slot=2622166 nextActionWait=3s810ms854us776ns nextAttestationSlot=2622167 nextProposalSlot=-1 head=7553b654:2622166
INF 2021-11-17 17:33:27.258+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:27.354+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101011111101111111111110011111111111101111111111101111110111111101111001111110111111110111011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"924bda1e\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:27.385+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:27.401+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:27.454+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:27.512+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101011111101111111111110011111111111101111111111101111110111111101111001111110111111110111011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"924bda1e\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:27.682+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:27.798+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101011111101111111111110011111111111101111111111101111110111111101111001111110111111110111011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"924bda1e\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:27.833+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101011111101111111111110011111111111101111111111101111110111111101111001111110111111110111011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"924bda1e\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:27.957+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:28.004+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110111001111110110111111101111111010100011100110011011111011101111111011111001111101011111111111011111101111011001111111011, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"8c3231f8\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:28.064+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110000010111100101111110101000111100111111111110011111101111101111011111100100110101110101110001111110111101110101011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"89c7527c\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:28.135+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101011111101111111111110011111111111101111111111101111110111111101111001111110111111110111011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"924bda1e\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:28.135+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101010111001101011111110011111100111101111111111101110110111111001111001111100111111110111011011011111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"86263a93\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:28.191+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111011110011000011110010111111100111011010100011100010010011111011001010110001011001111101000111111111011101001110011000111111001, data: (slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"942acb8c\")" src=gossip validator=b173ac90
INF 2021-11-17 17:33:28.194+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1110100010111110101111110101011111101111111111110011111111111101111111111101111110111111101111001111110111111110111011011111111110, data: (slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\"), signature: \"924bda1e\")" src=gossip validator=b0348547
INF 2021-11-17 17:33:31.000+01:00 Slot start                                 topics="beacnde" slot=2622167 epoch=81942 sync=synced peers=32 head=7553b654:2622166 finalized=81940:d6ce48c2 delay=896us856ns
INF 2021-11-17 17:33:31.188+01:00 Attestation included in block              attestationData="(slot: 2622166, index: 14, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\")" inclusion_lag_slots=0 validator=b173ac90
INF 2021-11-17 17:33:31.188+01:00 Attestation included in block              attestationData="(slot: 2622166, index: 22, beacon_block_root: \"7553b654\", source: \"81941:9d0944e4\", target: \"81942:8ecfaf31\")" inclusion_lag_slots=0 validator=b0348547
﻿```

the aggregate logging is somewhat overwhelming, but it's also accurate, in that sense

**Comment by arnetheduck** [comment]
Epoch end messages:
```
INF 2021-11-17 20:53:42.108+01:00 Previous epoch attestation included        topics="chaindag" timely_source=true timely_target=true timely_head=true epoch=81972 validator=b04395ff
INF 2021-11-17 20:53:42.108+01:00 Previous epoch attestation included        topics="chaindag" timely_source=true timely_target=true timely_head=false epoch=81972 validator=b173ac90
NOT 2021-11-17 20:53:42.108+01:00 Attestation failed to match head           topics="chaindag" epoch=81972 validator=b173ac90
INF 2021-11-17 20:53:42.108+01:00 Previous epoch attestation included        topics="chaindag" timely_source=true timely_target=true timely_head=true epoch=81972 validator=b869a146
INF 2021-11-17 20:53:42.108+01:00 Previous epoch attestation included        topics="chaindag" timely_source=true timely_target=true timely_head=true epoch=81972 validator=b8a1fc69
INF 2021-11-17 20:53:42.108+01:00 Previous epoch attestation included        topics="chaindag" timely_source=true timely_target=true timely_head=true epoch=81972 validator=b1fa2598
```

File: beacon_chain/rpc/rest_validator_api.nim [hunk 2]
```diff
@@ -512,6 +518,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
              lenu64(getStateField(node.dag.headState.data, validators)):
             return RestApiResponse.jsonError(Http400,
                                              InvalidValidatorIndexValueError)
+          let validator_pubkey = getStateField(
+            node.dag.headState.data, validators)[item.validator_index].pubkey
```
tersec: This is presumably invalidating that `HashList` cache for no particular reason.

beacon_chain/validators/validator_monitor.nim
```diff
@@ -0,0 +1,746 @@
...
+    slot = attestation.data.slot
+    delay = delay(slot, seen_timestamp, attestationSlotOffset)
+
+  self.withMonitor(idx):
+    let id = monitor.id
+    validator_monitor_unaggregated_attestation_total.inc(1, [$src, metricId])
+    validator_monitor_unaggregated_attestation_delay_seconds.observe(
+      delay.toFloatSeconds(), [$src, metricId])
+
+    info "Attestation seen",
```
tersec, line 532: These `info`-level logs ensure that the validator monitoring, when enabled, creates nearly unavoidably verbose logs.

beacon_chain/rpc/rest_validator_api.nim
arnetheduck: 67fa1bec3ee7fae5132b089ed2b4c879eea916ad

beacon_chain/validators/validator_monitor.nim
arnetheduck: well, that's kind of .. the point :) One downside is that we double-log things effectively - both "sent" and "seen" - it may be that we can revisit some of this duplication should we enable the feature by default.

The way to avoid the verbose logs is to set the log level to `NOT`.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -473,9 +496,15 @@ proc syncCommitteeContributionValidator*(
...
-  return if v.isOk():
+  if v.isOk():
     trace "Contribution validated"
-    self.syncCommitteeMsgPool[].addSyncContribution(contributionAndProof, v.get)
+
+    self.syncCommitteeMsgPool[].addSyncContribution(
+      contributionAndProof, v.get()[0])
+
+    self.validatorMonitor[].registerSyncContribution(
+      src, wallTime, contributionAndProof, v.get()[1])
```
zah, line 503: How about using named fields here?

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -855,10 +856,10 @@ proc validateSignedContributionAndProof*(
     initialized = false
     syncCommitteeSlot = msg.message.contribution.slot + 1
 
-  for validatorPubKey in dag.syncCommitteeParticipants(
-      syncCommitteeSlot,
-      committeeIdx,
-      msg.message.contribution.aggregation_bits):
+  let participants = toSeq(dag.syncCommitteeParticipants(
```
zah: Both this copy and the returned copy can be avoided once we have access to view types in the future. Leaving a TODO comment in the code would be appropriate.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -486,6 +486,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
         request.slot, subnet_id, request.validator_index,
         request.is_aggregator)
 
+      let validator_pubkey = getStateField(
+        node.dag.headState.data, validators).asSeq()[request.validator_index].pubkey
```
zah, line 547: Expensive copy of the `validators` field. Use `withState(..): state.data.validators[request.validator_index]` instead.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -512,6 +518,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
              lenu64(getStateField(node.dag.headState.data, validators)):
             return RestApiResponse.jsonError(Http400,
                                              InvalidValidatorIndexValueError)
+          let validator_pubkey = getStateField(
+            node.dag.headState.data, validators).asSeq()[item.validator_index].pubkey
```
zah, line 579: Ditto

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -648,6 +648,15 @@ iterator get_flag_index_deltas*(
     yield
       if is_unslashed_participating_index(
           state, flag_index, previous_epoch, vidx):
+
+        let pflag = case flag_index
```
zah: This looks like a reusable function.

beacon_chain/validators/validator_monitor.nim
```diff
@@ -0,0 +1,746 @@
...
+
+proc init*(T: type ValidatorMonitor, autoRegister = false, totals = false): T =
+  T(autoRegister: autoRegister, totals: totals)
+
+template summaryIdx(epoch: Epoch): int = (epoch.uint64 mod 2).int
+
+template withEpochSummary(
+  self: var ValidatorMonitor, monitor: var MonitoredValidator,
+  epochParam: Epoch, body: untyped) =
+  mixin summaryIdx
```
zah: Why is this mixed in? You don't want the user to override it in the caller scope.

beacon_chain/validators/validator_monitor.nim
```diff
@@ -0,0 +1,746 @@
...
+
+    validator_monitor_sync_committee_messages_total.inc(1, [$src, metricId])
+    validator_monitor_sync_committee_messages_delay_seconds.observe(
+      delay.toFloatSeconds(), [$src, metricId])
+
+    info "Sync committee message seen",
+      syncCommitteeMessage = shortLog(sync_committee_message.beacon_block_root),
+      src, epoch = slot.epoch, validator = id
+
+    self.withEpochSummary(monitor, slot.epoch):
```
zah, line 641: The same sync committee message can reach more than once here. As far as I can tell, this would lead to over-counting in the metrics.

beacon_chain/validators/validator_monitor.nim
```diff
@@ -0,0 +1,746 @@
...
+    validator_monitor_sync_contributions_total.inc(1, [$src, metricId])
+    validator_monitor_sync_contributions_delay_seconds.observe(
+      delay.toFloatSeconds(), [$src, metricId])
+
+    info "Sync contribution seen",
+      contribution = shortLog(sync_contribution.message.contribution),
+      src, epoch = slot.epoch, validator = id
+
+    self.withEpochSummary(monitor, slot.epoch):
+      epochSummary.sync_contributions += 1
```
zah, line 668: Ditto

beacon_chain/validators/validator_monitor.nim
```diff
@@ -0,0 +1,746 @@
...
+    pubkey: ValidatorPubKey) =
+  self.withMonitor(pubkey):
+    let id = monitor.id
+    validator_monitor_sync_committee_message_in_block_total.inc(1, ["block", id])
+
+    info "Sync signature included in block",
+      head = beacon_block_root, slot = slot, validator = id
+
+    self.withEpochSummary(monitor, slot.epoch):
+      epochSummary.sync_signature_block_inclusions += 1
```
zah, line 694: What if the signature ends up in multiple orphaned blocks? The resulting metrics may be misleading to the user.

beacon_chain/gossip_processing/eth2_processor.nim
arnetheduck: I don't care/mind either way, but the error return is also an anonymous tuple (called `ValidationError`) so I simply stayed consistent with that - if we change, we should change that too (perhaps in a separate PR)

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: there are _a lot_ of places that would benefit from avoiding copies - I've considered adding todo:s everywhere but refrained from doing so because often, a view type is not needed (an iterator can be used instead - I've already replaced some of your "view todo":s in that way) and if ever we do get access to view types that wrok, we'll need a tool to discover where it is relevant to optimize - in the meantime, I feel they clutter the code unnecessarily, but if you still think they're useful, I can add them here.

beacon_chain/validators/validator_monitor.nim
arnetheduck: I had a random compile issue and hoped this would fix it, why else would one use mixin? :)

it turned out to be the following, very annoying behavior: if there's a symbol called `epochSummary` in another module - anywhere - it will be matched over the local `epochSummary` below, when expanding `body` - quite annoying, any ideas what to do about it? I'd like `body` to prefer the "local" template.

beacon_chain/rpc/rest_validator_api.nim
arnetheduck: as noted in the other PR, there's no expensive copy of `validators` here - there is a somewhat expensive copy of `Validator` however that would need `lent` to avoid.

Here's how the code expands:
```
									T78_ = (tyObject_HashList__EbZJBpzKVMIaHsYOwWyCGQ *)0;
									switch ((*(*(*(*(*colonenvP_).colonup_).colonup_).node1).dag).headState.data.kind)
									{
									case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw)2):
									{
										T78_ = (&(*(*(*(*(*colonenvP_).colonup_).colonup_).node1).dag).headState.data.mergeData.data.validators);
									}
									break;
									case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw)1):
									{
										T78_ = (&(*(*(*(*(*colonenvP_).colonup_).colonup_).node1).dag).headState.data.altairData.data.validators);
									}
									break;
									case ((tyEnum_BeaconStateFork__Dw9aV5NV0W0A7EnWLN5Ji9aw)0):
									{
										T78_ = (&(*(*(*(*(*colonenvP_).colonup_).colonup_).node1).dag).headState.data.phase0Data.data.validators);
									}
									break;
									}
									(*colonenvP_).validator_pubkey16 = (*T78_).data->data[(*colonenvP_).request7.validator_index].pubkey;
									addAutoMonitor__Wv11X89aUgvCIKIDyjDFXRA((*(*(*(*colonenvP_).colonup_).colonup_).node1).validatorMonitor, (&(*colonenvP_).validator_pubkey16), (*colonenvP_).request7.validator_index);
```

I'd generally say the biggest issue here is that `validator_pubkey` is included in the closure environment - it's not a problem here, but it is a big problem in `async` in general, that variables that "could" be local end up spuriously being part of the environment - this increases memory consumption significantly.

beacon_chain/rpc/rest_validator_api.nim
arnetheduck: Ditto

beacon_chain/spec/state_transition_epoch.nim
arnetheduck: it does indeed, but there's nowhere else to use it, so yagni.

a bigger refactoring would perhaps diverge from the spec in how `flag_index` is represented throughout, which perhaps would net a higher readability reward - the way it's done current is _quite_ convoluted.

beacon_chain/validators/validator_monitor.nim
arnetheduck: This is in line with the lighthouse implementation - I guess this is why we record a `minimum` per-epoch delay.

That said, I'm unsure how it would happen exactly - sync committee messages _should_ be deduplicated by gossipsub (this function is called from eth2_processor for valid messages only), unless a single validator has two spots in the committee in which case it should be recorded twice, no?

beacon_chain/validators/validator_monitor.nim
arnetheduck: sync contributions, like aggregate inclusions, are indeed double-counted, in line with the lighthouse implementation - this is a quality metric of sorts: how many others included my attestation / syncsig? if it's not "all aggregators", something is wrong.

beacon_chain/validators/validator_monitor.nim
arnetheduck: yes, this is misleading indeed, but also in line with the lighthouse implementation - the same thing can happen for all the metrics because of reorgs, but is perhaps more probable for sync committees - I'm not sure what a better metric would be, that is viable from a state transition perspective (ie that doesn't require examining a complex out-of-state history) - any good ideas?

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: basically, I've been using vtune to find the ones that matter - for the rest, well, whether it's a view type or not is secondary to how well the code reads and what compiler bugs and unfamiliar features one must risk to get there

zah: I think a TODO comment here would be more useful than the typical trivial opportunities in other places because the required refactoring a bit more involved. Instead of returning the sync sub-committee members, the validation function would return the sub-committee ID. The caller would use a view-types based API to get the list of the members without introducing allocations and/or copying.

beacon_chain/rpc/rest_validator_api.nim
zah: Okay, I hadn't noticed this improvement, sorry for the wasted time.

beacon_chain/validators/validator_monitor.nim
zah: There are multiple layers of deduplication based on time eviction caches (one in gossipsub and another that is reset once per slot in the sync committee msg pool). Nevertheless, we know that gossip messages "echo" in the network and they end up being processed more than once. The bug with invalid contributions was happening because some sync committee messages were being processed more than once by a similar function in the sync committee msg pool.

beacon_chain/validators/validator_monitor.nim
zah: Yes, this is a nasty effect of the Nim generics pre-pass (which I've tried to convince @Araq to remove). 

Nim wants to assign symbol candidates to all identifiers in the body of generics and templates before the said generics and templates are instantiated. If there is only a single candidate for a particular symbol, it's bound early before the code has a chance to consider such injected symbols. So, a work-around would be to introduce a dummy overload of `epochSummary` next to the top-level one that was inappropriately selected. This would force Nim to keep the set of pre-pass matches as an "open sym choice" and your injected definition should be discovered later during the actual instantiation.

The use of `mixin summaryIdx` should not help for this problem in any way, so I would still suggest removing it.

beacon_chain/validators/validator_monitor.nim
zah: So, I guess the idea of the metric is to use the fact that we are seeing a message more than once as a signal of some sort that the message wes broadcasted well in the network. Seems a bit dubious logic for the stated arguments, but alright.

beacon_chain/validators/validator_monitor.nim
arnetheduck: ec860cb1eb9728e614b6413df9c9962b7fbda9d1

beacon_chain/validators/validator_monitor.nim
zah: Well, why do we need to replicate the Lighthouse implemetnation if has flaws like these. Our users would probably lose trust in these metrics if they see that they've lost money at a particular epoch, but the metrics report everything to be fine.

We can perhaps have another set of metrics that report the precise performance as soon as the chain finalizes.

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: Right, I'm not sure that would fly, as an approach, specially as long as we're using the head committee trick - it's quite fragile (aka I've fixed many such bugs) in the sense that the underlying sync committee might change in an accidental async call somewhere between validation and expansion - I can see the case for returning some lent view that the compiler guarantees to stay fresh. Here, the idea is that we validate the data and use it directly to compute the list then we export the outcome of that computation.

This was one of the downsides of separating validation from updates - I've considered an alternative design where the pools do the validation as part of adding the message - this has other problems, because it doesn't separate presumably fast gossip validation from slow pool mutation. So many choices.

beacon_chain/validators/validator_monitor.nim
zah: Overcounting could still mislead your though. You may count the same contribution multiple times and then conclude that all aggregators have seen your sync committee messages which is not necessary the case. So, the usefulness is limited to "higher counts are better than lower counts".

beacon_chain/validators/validator_monitor.nim
arnetheduck: > replicate

The feature is here so that dashboards and alerts become portable, by and large, so I reasoned we want to keep it pretty close.

Also, if we wait until finalization, we need to consider the additional delay that happens in reporting, we need to keep a dynamic number of epochs "open", multiple branches of history as well as the stampede of reports on finalization - waiting is not without its own issues.

beacon_chain/validators/validator_monitor.nim
arnetheduck: > broadcasted well in the network

indeed - "it got this far"

arnetheduck: > Nevertheless, we know that gossip messages 

sync committee messages have a pretty short lifespan, no? should we tune the gossip cache timeout?

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: b0162f73a9035afe61b04038db89c942a88d4e20

beacon_chain/validators/validator_monitor.nim
arnetheduck: b3545ac202ee8c6ab2b4ba83570d21463d41c101

beacon_chain/validators/validator_monitor.nim
arnetheduck: b3545ac202ee8c6ab2b4ba83570d21463d41c101

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
+# Validator monitoring
+
+The validator monitoring feature allows tracking the life-cycle and performance of one or more validators in detail. Monitoring can be done for any validator with slightly more detail for validators that are running through the same beacon node.
```
unixpi:
```suggestion
The validator monitoring feature allows for tracking the life-cycle and performance of one or more validators in detail. Monitoring can be carried out for any validator, with slightly more detail for validators that are running through the same beacon node.
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+
+Every time the validator performs an duty, the duty is recorded and the monitor will keep track of events related to getting rewarded for having performed it:
```
unixpi:
```suggestion
Every time the validator performs a duty, the duty is recorded and the monitor keeps track of the reward-related events for having performed it. For example:
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+
+* When attesting, the attestation being added to an aggregate, then block, then being given a reward in the state
+* When performing sync committee duties, likewise
+
+Validator actions can be traced either through logging, or comprehensive metrics that allow creating alerts in monitoring tools. The metrics are based on the same feature in [Lighthouse](https://lighthouse-book.sigmaprime.io/validator-monitoring.html), thus dashboards and alerts can be used with either client.
```
unixpi:
```suggestion
Validator actions can be traced either through logging, or comprehensive metrics that allow for creating alerts in monitoring tools. The metrics are based on the same feature in [Lighthouse](https://lighthouse-book.sigmaprime.io/validator-monitoring.html), thus dashboards and alerts can be used with either client.
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
+# Validator monitoring
+
+The validator monitoring feature allows tracking the life-cycle and performance of one or more validators in detail. Monitoring can be done for any validator with slightly more detail for validators that are running through the same beacon node.
+
+Every time the validator performs an duty, the duty is recorded and the monitor will keep track of events related to getting rewarded for having performed it:
+
+* When attesting, the attestation being added to an aggregate, then block, then being given a reward in the state
```
unixpi:
```suggestion
* When attesting, the attestation is added to an aggregate, then a block, before a reward is applied to the state
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Publish metrics as totals for all monitored validators instead of each validator separately - used for limiting the load on metrics when monitoring many validators
+./run-mainnet-beacon-node.sh --validator-monitor-totals
+```
+
+## Understanding monitoring
+
+When a validator performs a duty, such as signing an attestation or a sync committee message, this is broadcast to the network. Other nodes pick it up and package the message into an aggregate and later a block. The block is included in the canonical chain and a reward is given two epochs (~13 minutes) later.
+
+The monitor tracks these actions and will log each step at the INF level. If any step is missed, a NOT log is shown instead.
```
unixpi:
```suggestion
The monitor tracks these actions and will log each step at the `INF` level. If any step is missed, a `NOT` log is shown instead.
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+
+A typical lifecycle of an attestation might look something like the following:
```
unixpi:
```suggestion
The typical lifecycle of an attestation might look something like the following:
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+INF 2021-11-22 11:32:51.293+01:00 Attestation included in aggregate          topics="val_mon" aggregate="(aggregation_bits: 0b1111111101011111001101111111101100111111110100111011111110110101110111111010111111011101111011101111111111101111100001111111100111, data: (slot: 2656363, index: 11, beacon_block_root: \"bbe7fc25\", source: \"83010:a8a1b125\", target: \"83011:6db281cd\"), signature: \"8576b3fc\")" src=gossip epoch=83011 validator=b93c290b
+INF 2021-11-22 11:33:07.193+01:00 Attestation included in block              attestation_data="(slot: 2656364, index: 9, beacon_block_root: \"c7761767\", source: \"83010:a8a1b125\", target: \"83011:6db281cd\")" block_slot=2656365 inclusion_lag_slots=0 epoch=83011 validator=b65b6e1b
+```
+
+The lifecycle of a particular message can be traced by following the `epoch=.... validator=...` fields in the message.
+
+Failures at any point are recorder at a higher logging level, such as `NOT`(ice):
```
unixpi:
```suggestion
Failures at any point are recorded at a higher logging level, such as `NOT`(ice):
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+NOT 2021-11-17 20:53:42.108+01:00 Attestation failed to match head           topics="chaindag" epoch=81972 validator=...
+```
+
+:warning: It should be noted that metrics are tracked for the current history - in the case of a reorg on the chain - in particular a deep reorg - no attempt is made to revisit previously reported values. In the case that finality is delayed, the risk of stale metrics increases.
```
unixpi:
```suggestion
> :warning: It should be noted that metrics are tracked for the current history - in the case of a reorg on the chain - in particular a deep reorg - no attempt is made to revisit previously reported values. In the case that finality is delayed, the risk of stale metrics increases.
```

docs/the_nimbus_book/src/validator-monitor.md
```diff
@@ -0,0 +1,61 @@
...
+
+Likewise, many metrics, such as aggregation includion, reflect conditions on the network - it may happen that the same message is counted more than once under certain conditions.
```
unixpi:
```suggestion
Likewise, many metrics, such as aggregation inclusion, reflect conditions on the network - it may happen that the same message is counted more than once under certain conditions.
```

unixpi (commented):
minor edits, mainly style

PR #326, opened 2019-07-18
v0.8.1 tests refactor

mratsim:
Opening a PR so it's easier to track the changes and potentiel conflicts:

### Updates to the core nim-beaconchain

- `compact_committees_roots` in `initialize_beacon_state_from_eth1` is now initialized
- export `update` for Eth2Digest
- fixes some symbol visibility issues with method call syntax in `withEth2Hash` template

### Desync noticed with latest specs (but not updated)

- `verify_merkle_branch` is now `is_valid_merkle_branch`
- `graffiti` in BeaconBlockBody should be a raw array[32, byte]
- `compute_domain` now has a default parameter for the `fork_version`.
  There is now a type called `Version` for fork versions

## Description

To be filled

PR #2928, opened 2021-09-29
always use static keys in tests

etan-status:
There are multiple ways how states can be processed during tests. One is
through tests/mocking, the other through tests/xxxutil. Both methods can
create genesis states with a set of deposits from mock validators. The
private keys for those validators are created differently, though. While
tests/mocking generates random keys, testblockutil generates fixed keys.
This is a problem when mixing both tests/mocking and testblockutil in a
single test case, as they assume that the deposits were generated using
the same method. This patch aligns the key generation across tests.


tests/testblockutil.nim
```diff
@@ -12,14 +12,11 @@ import
   ../beacon_chain/validators/validator_pool,
   ../beacon_chain/spec/datatypes/merge,
   ../beacon_chain/spec/[helpers, signatures, state_transition, forks],
-  ../beacon_chain/consensus_object_pools/attestation_pool
+  ../beacon_chain/consensus_object_pools/attestation_pool,
+  mocking/mock_validator_keys
 
 func makeFakeValidatorPrivKey*(i: int): ValidatorPrivKey =
-  # 0 is not a valid BLS private key - 1000 helps interop with rust BLS library,
-  # lighthouse.
```
etan-status: This comment was made ~2 years ago, not sure that it is still relevant.

tests/testblockutil.nim
```diff
@@ -12,14 +12,11 @@ import
...
-  # TODO: switch to https://github.com/ethereum/eth2.0-pm/issues/60
```
etan-status, line 20: This repo is also dead for several years.
The proposed keygen is here: https://github.com/ethereum/eth2.0-pm/tree/master/interop/mocked_start

etan-status: Another more-recently introduced keygen for testing is here: https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/eth2spec/test/helpers/keys.py

etan-status:
Can either use the tests/mocking keygen from testblockutil, or the other way around use the static keys from tests/mocking. The problem is only when mixing them. Both are only used in tests, and for now I opted to keep the random one to have the additional chance of maybe sometime randomly discovering an edge case where a random key breaks things.

etan-status:
In retrospec, I think doing this the other way around is smarter. The `testblockutil` approach has the advantage of being deterministic, and also of scaling arbitrarily. The mocking approach relies on not more validators being created than available keys, and the errors I'm getting on CI do not really reflect the problem easily, so it wastes quite some time to debug this.

Another consideration is that mocking previously generated random keys, so I will check all of the callers taking advantage of it to see whether the randomness aspect matters. If it doesn't, it is possible to switch to the same approach that the spec is choosing in their own tests (1 + id), currently we have (1000 + id)

arnetheduck:
> I think doing this the other way around is smarter

+1

etan-status:
* `test_message_signatures` verifies that the different signature generation and checking functions works when used on round trips. It makes sense to use randomized keys there to avoid only testing the 0 / 1 case. I changed it to randomize the key index instead.
* `mock_attestations` creates attestations for tests. It is used in `test_process_attestations`. This is a test with 256 validators, so it is unlikely that using hardcoded privkeys instead of random ones affects this test. Leaving as is.
* `mock_blocks` creates fake blocks according to spec and uses the mock key for block signing. This is used by `mock_attestations`, which is only used by a test with a large amount of validators, so leaving as is.
* `mock_deposits` simply creates deposit objects with the given privkeys. It is used by the `test_process_deposits` which does not rely on crypto at all but just verifies that the deposit is properly recorded. No need to use random keys here.

=> It is not needed to have a facility for creating randomized mock keys. If this ever becomes necessary, the functionality can be reintroduced.

tests/mocking/mock_validator_keys.nim
```diff
@@ -9,52 +9,26 @@
...
+type
+  MockPrivKeysT = object
+  MockPubKeysT = object
+const 
+  MockPrivKeys* = MockPrivKeysT()
+  MockPubKeys* = MockPubKeysT()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0/tests/core/pyspec/eth2spec/test/helpers/keys.py
+func `[]`*(_: MockPrivKeysT, index: ValidatorIndex): ValidatorPrivKey =
+  var bytes = (index.uint64 + 1000'u64).toBytesLE()
```
etan-status: For some reason, block_sim fails if this is done with +1 instead of +1000. Seems like a separate issue that should be investigated as a followup.

tests/testblockutil.nim
etan-status: The interop keygen from https://github.com/ethereum/eth2.0-pm/tree/master/interop/mocked_start actually made it into `beacon_chain/interop.nim`

tests/testblockutil.nim
```diff
@@ -12,14 +12,11 @@ import
...
-  var bytes = uint64(i + 1000).toBytesLE()
-  copyMem(addr result, addr bytes[0], sizeof(bytes))
+  MockPrivKeys[i.ValidatorIndex]
```
arnetheduck, line 36: seems easier to just remove mock_validator_keys and move the code in here?

ie there's not much code in `mock*` in general that isn't repeated somewhere else, so perhaps we can get rid of it?

etan-status: Yeah, sure, if the `when isMainModule` printout is not needed anywhere, that can be done. Not sure who depends on this binary being available.

arnetheduck: I suspect it was used only during manual testing long ago - perhaps @mratsim knows more

etan-status: Alright, I'll get rid of it then. Can easily re-introduce it if this makes trouble.

mratsim:
To give your more context, the `mockXXX` procedures where added about 2.5 years ago. We had a lot of finalization issues when testing with 4 nodes. We need more robust ways to ensure that our attestations were correct so that we knew if the issue was in our consensus implementation, our fork choice or within networking.

At the time there were already something to create fake blocks that @arnetheduck introduced but it actually pre-dated some important spec changes.

IIRC it was between v0.5.1 and v0.6.0. https://github.com/ethereum/consensus-specs/compare/v0.5.1...v0.6.0 with 650 commits in less than a month and fork choice wasn't even specified at the time so we were relying on something hacked together.

We wanted to make sure that using those fake blocks proc didn't lead to our issues that we were carrying for ~3 months and added as much tests as possible as well as ones that didn't rely on old proc that might have become outdated.

The mocks were introduced as part as that effort in https://github.com/status-im/nimbus-eth2/pull/326/files. It turned out iirc that the issue was elsewhere hence the old fake block stuff were also valid, we kept both in the codebase.

etan-status:
Thanks for the additional background!

arnetheduck (approved):

PR #2930, opened 2021-09-30
Merge Interop

tersec:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -430,15 +430,24 @@ proc init*(T: type BeaconNode,
             getProposerSlashingsTopic(network.forkDigests.altair),
             getVoluntaryExitsTopic(network.forkDigests.altair),
             getAggregateAndProofsTopic(network.forkDigests.altair),
+
+            getBeaconBlocksTopic(dag.forkDigests.merge),
```
arnetheduck, line 470: we should probably only be subscribing to these when the merge epoch in the runtime config is not a farfutureepoch -

tersec: well the whole thing is an expedient kludge to keep things quicker and more predictable at the cost of CPU and bandwidth. That said, yes, `MERGE_FORK_EPOCH` being `FAR_FUTURE_EPOCH` means one can disable all those.

arnetheduck: well, given this will land for mainnet users as well, it's the least we can do - it's actually not that far from a "final" working solution either

PR #2936, opened 2021-09-30
remove privkey from mock withdrawal credentials

etan-status:
In tests, the private key was put into the validator deposit's withdraw
credentials so that it can be recovered later. This leads to problems
when creating the validators through other means that do not put the key
there. In general, mock private keys only depend on the validator index,
though, and because it is clear what the index of a validator is, it is
not actually needed to put the key into the credentials.

### DIFF
--- a/research/block_sim.nim
+++ b/research/block_sim.nim
@@ -128,8 +128,7 @@ cli do(slots = SLOTS_PER_EPOCH * 6,
               sig =
                 get_attestation_signature(getStateField(stateData.data, fork),
                   getStateField(stateData.data, genesis_validators_root),
-                  data, hackPrivKey(
-                    getStateField(stateData.data, validators)[validatorIdx]))
+                  data, MockPrivKeys[validatorIdx])
             var aggregation_bits = CommitteeValidatorsBits.init(committee.len)
             aggregation_bits.setBit index_in_committee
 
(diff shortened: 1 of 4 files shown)


research/block_sim.nim
```diff
@@ -128,8 +128,7 @@ cli do(slots = SLOTS_PER_EPOCH * 6,
               sig =
                 get_attestation_signature(getStateField(stateData.data, fork),
                   getStateField(stateData.data, genesis_validators_root),
-                  data, hackPrivKey(
-                    getStateField(stateData.data, validators)[validatorIdx]))
+                  data, MockPrivKeys[validatorIdx])
```
zah, line 131: The Nim style guide (aka [NEP1](https://nim-lang.org/docs/nep1.html)) suggests that constants should use `camelCase`. `PascalCase` is suggested for type names and variables storing types (in macros).

zah (approved):

research/block_sim.nim
arnetheduck: strictly, it doesn't:

> Constants can start with either a lower case or upper case letter.

arnetheduck: that said, yeah, I've started liking lowercase for them

etan-status: If you are referring to `MockPrivKeys`, I did not pick that name. But if it should be changed, let me know and I'll put it in my backlog as a followup PR.

PR #2939, opened 2021-10-01
abort compile when fork epoch is forgotten

etan-status:
There are a few locations in the code that compare the current epoch to
the various FORK_EPOCH constants and branch off into fork-specific code.
When a new fork is introduced, it is sometimes forgotten to update all
of those branch locations. This patch introduces a compile-time check
that ensures that all branches need to be covered exhaustively. This is
done by replacing if-elif structures with case expressions.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -311,12 +311,10 @@ func contains*(dag: ChainDAGRef, root: Eth2Digest): bool =
 
 proc containsBlock(
     cfg: RuntimeConfig, db: BeaconChainDB, blck: BlockRef): bool =
-  if blck.slot.epoch < cfg.ALTAIR_FORK_EPOCH:
-    db.containsBlockPhase0(blck.root)
-  elif blck.slot.epoch < cfg.MERGE_FORK_EPOCH:
-    db.containsBlockAltair(blck.root)
-  else:
-    db.containsBlockMerge(blck.root)
+  case cfg.stateForkAtEpoch(blck.slot.epoch)
+  of forkMerge:  db.containsBlockMerge(blck.root)
+  of forkAltair: db.containsBlockAltair(blck.root)
+  of forkPhase0: db.containsBlockPhase0(blck.root)
 
 func isStateCheckpoint(bs: BlockSlot): bool =
   ## State checkpoints are the points in time for which we store full state
@@ -578,23 +576,24 @@ proc getState(
   func restore() =
     assign(v[], restoreAddr[].data)
 
-  if blck.slot.epoch < dag.cfg.ALTAIR_FORK_EPOCH:
-    if state.data.beaconStateFork != forkPhase0:
-      state.data = (ref ForkedHashedBeaconState)(beaconStateFork: forkPhase0)[]
+  case dag.cfg.stateForkAtEpoch(blck.slot.epoch)
+  of forkMerge:
+    if state.data.beaconStateFork != forkMerge:
+      state.data = (ref ForkedHashedBeaconState)(beaconStateFork: forkMerge)[]
 
-    if not dag.db.getState(stateRoot, state.data.hbsPhase0.data, restore):
+    if not dag.db.getMergeState(stateRoot, state.data.hbsMerge.data, restore):
       return false
-  elif blck.slot.epoch < dag.cfg.MERGE_FORK_EPOCH:
+  of forkAltair:
     if state.data.beaconStateFork != forkAltair:
       state.data = (ref ForkedHashedBeaconState)(beaconStateFork: forkAltair)[]
 
     if not dag.db.getAltairState(stateRoot, state.data.hbsAltair.data, restore):
       return false
-  else:
-    if state.data.beaconStateFork != forkMerge:
-      state.data = (ref ForkedHashedBeaconState)(beaconStateFork: forkMerge)[]
+  of forkPhase0:
+    if state.data.beaconStateFork != forkPhase0:
+      state.data = (ref ForkedHashedBeaconState)(beaconStateFork: forkPhase0)[]
 
-    if not dag.db.getMergeState(stateRoot, state.data.hbsMerge.data, restore):
+    if not dag.db.getState(stateRoot, state.data.hbsPhase0.data, restore):
       return false
 
   state.blck = blck
@@ -613,12 +612,10 @@ template forkAtEpoch*(dag: ChainDAGRef, epoch: Epoch): Fork =
   forkAtEpoch(dag.cfg, epoch)
 
 proc forkDigestAtEpoch*(dag: ChainDAGRef, epoch: Epoch): ForkDigest =
-  if epoch < dag.cfg.ALTAIR_FORK_EPOCH:
-    dag.forkDigests.phase0
-  elif epoch < dag.cfg.MERGE_FORK_EPOCH:
-    dag.forkDigests.altair
-  else:
-    dag.forkDigests.merge
+  case dag.cfg.stateForkAtEpoch(epoch)
+  of forkMerge:  dag.forkDigests.merge
+  of forkAltair: dag.forkDigests.altair
+  of forkPhase0: dag.forkDigests.phase0
 
 proc getState(dag: ChainDAGRef, state: var StateData, bs: BlockSlot): bool =
   ## Load a state from the database given a block and a slot - this will first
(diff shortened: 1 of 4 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -712,7 +712,7 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         let altairRes = decodeBody(altair.SignedBeaconBlock, body)
         if altairRes.isOk():
           var res = altairRes.get()
-          if res.message.slot.epoch < node.dag.cfg.ALTAIR_FORK_EPOCH:
+          if node.dag.cfg.forkState(res.message.slot.epoch) < forkAltair:
```
zah: I'm not quite convinced this is an improvement. The static asserts in `forkState` don't guarantee that the `_FORK_EPOCH` values will be ordered correctly (we can check that while we load the config in `presets.nim`).

In the end the assumptions of ordering are still present in the code and the direct value comparison of the previous code is cheaper than the `forkState` call.

beacon_chain/spec/forks.nim
```diff
@@ -278,6 +278,17 @@ proc check_voluntary_exit*(
 
 # Derived utilities
 
+func forkState*(cfg: RuntimeConfig, epoch: Epoch): BeaconStateFork =
```
zah: Could be called `forkAtEpoch`. `forkState` sounds quite ambiguous.

etan-status: `forkAtEpoch` already exists, it returns `Fork`.

etan-status: Maybe `stateForkAtEpoch`?

beacon_chain/rpc/rest_beacon_api.nim
etan-status: OK, will keep the old version for the `<` comparisons.

arnetheduck:
btw, @zah I remember you didn't want a "unified" fork enum  for some reason? ie that a type might not be touched by all forks - that particular point though shouldn't be a problem really since kind objects can use multiple enum values for the same fields .. anything else?

beacon_chain/rpc/rest_beacon_api.nim
etan-status: (BTW: The `_FORK_EPOCH` ordering is verified in blockchain DAG initialization.)

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -477,6 +475,7 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   when true:
```
tersec: This `when true:` wrapping over those assertions was a debugging artifact when it went in and not useful now.

beacon_chain/spec/forks.nim
```diff
@@ -278,6 +278,17 @@ proc check_voluntary_exit*(
...
 
+func stateForkAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): BeaconStateFork =
+  ## Return the current fork for the given epoch.
+  static:
+    doAssert forkMerge  > forkAltair
+    doAssert forkAltair > forkPhase0
+
+  if   epoch >= cfg.MERGE_FORK_EPOCH:  forkMerge
+  elif epoch >= cfg.ALTAIR_FORK_EPOCH: forkAltair
+  elif epoch >= GENESIS_EPOCH:         forkPhase0
```
tersec: `GENESIS_EPOCH` is always 0: https://github.com/ethereum/consensus-specs/blob/v1.1.0/specs/phase0/beacon-chain.md#misc and `Epoch` and `Slot` are both `uint64`. That last comparison's going to be always true, and is a bit wasted. The compiler will probably optimize it out, and there's some value in having parallel structure, but I'm not convinced it's worthwhile overall.

In theory, some future spec version could either allow negative epochs or increase `GENESIS_EPOCH`, but at least the latter can be checked with a static/compile-time assertion instead of a runtime check each time. If negative epochs are allowed, that's a large enough change I'd not be worried about this specific aspects.

Of course, these conditions could change at some point, but (a) it's unlikely, and (b) if it does, it still saves the comparisons that are going to result from

beacon_chain/spec/forks.nim
```diff
@@ -278,6 +278,17 @@ proc check_voluntary_exit*(
 
 # Derived utilities
 
+func stateForkAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): BeaconStateFork =
+  ## Return the current fork for the given epoch.
+  static:
+    doAssert forkMerge  > forkAltair
+    doAssert forkAltair > forkPhase0
+
+  if   epoch >= cfg.MERGE_FORK_EPOCH:  forkMerge
```
tersec, line 288: There had been a mix of
```nim
if epoch < ALTAIR:
  ...
elif epoch < MERGE:
  ...
else:
  ...
```

The version in this PR should be more correct for, say, `ALTAIR_FORK_EPOCH == MERGE_FORK_EPOCH`, where it will pass through `ALTAIR` within the state transition function without any Altair-but-not-Merge-nature ever being visible outside that, but due to that slight discrepancy it does need to be consistent across the codebase.

So I support this change, just that it would then have to be caught/audited for elsewhere, too, and/or in the meantime, prohibit any of `GENESIS_EPOCH`, `ALTAIR_FORK_EPOCH`, and `MERGE_FORK_EPOCH` from equating. The latter seems distasteful, but fine in the service of this change.

beacon_chain/spec/forks.nim
```diff
@@ -278,6 +278,17 @@ proc check_voluntary_exit*(
...
+  elif epoch >= cfg.ALTAIR_FORK_EPOCH: forkAltair
+  elif epoch >= GENESIS_EPOCH:         forkPhase0
+  else: raiseAssert "No fork defined for given epoch"
```
tersec: I don't think this `raiseAssert` is even theoretically possible to hit, because of `GENESIS_EPOCH` being 0.

beacon_chain/spec/forks.nim
etan-status: Thanks for the hint about GENESIS_EPOCH statically being 0!

I'll replace this with a static assert for GENESIS_EPOCH == 0, and remove the extra comparison. This way, this is still cought (even at compile time)

beacon_chain/spec/forks.nim
etan-status: Yeah, I was thinking about that as well. For the `MERGE_FORK_EPOCH > ALTAIR_FORK_EPOCH > GENESIS_EPOCH` case it is the same logic as before, and for the `==` case it is a bit more correct in general, but that case is still broken because the `Fork` structures end up being incorrect when multiple forks overlap, and also because a lot of the genesis logic assumes that the chain starts at phase0. Fixing those parts of the code to support the `==` case would be more correct though, I think. For the mainnet and the public testnets, it is not a problem. Will track this as a followup!

beacon_chain/spec/forks.nim
etan-status: You are right! Replaced with a static assert for `GENESIS_EPOCH` == 0.

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: It will be removed in https://github.com/status-im/nimbus-eth2/pull/2930

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -477,6 +475,7 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
     doAssert cfg.GENESIS_FORK_VERSION != cfg.MERGE_FORK_VERSION
     doAssert cfg.ALTAIR_FORK_VERSION != cfg.MERGE_FORK_VERSION
+  doAssert GENESIS_EPOCH <= cfg.ALTAIR_FORK_EPOCH
```
tersec: Trivially true, since as verified by the static assert you added, this is just `doAssert 0 <= someUint64`.

tersec (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: Thanks, I'll remove this one as well.

beacon_chain/spec/forks.nim
etan-status: I think that @arnetheduck meanwhile extended Nimbus for sensible behaviour in the `==` case (i.e., start chain from any fork).

PR #2943, opened 2021-10-01
allow `withXxx` to access fork-specific fields

etan-status:
So far, `withState` and `withBlck` templates could only be used to have
convenience access to fork-agnostic BeaconState and BeaconBlock fields.
This patch:
- injects an additional `stateFork` constant that allows to use
  `when` expressions to also access Altair and Merge-specific fields.
- introduces a `withStateAndBlck` template to support operating on both
  a `BeaconState` and `BeaconBlock` at a time.
- makes sync committee related functions Merge aware.
- changes a couple if-else trees for forks into case statements so that
  forgotten future forks are promoted to compile-time errors.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1052,21 +1052,22 @@ func syncCommitteeParticipants*(dagParam: ChainDAGRef,
     dag = dagParam
     slot = slotParam
 
-  if dag.headState.data.beaconStateFork == forkAltair:
-    let
-      headSlot = dag.headState.data.hbsAltair.data.slot
-      headCommitteePeriod = syncCommitteePeriod(headSlot)
-      periodStart = syncCommitteePeriodStartSlot(headCommitteePeriod)
-      nextPeriodStart = periodStart + SLOTS_PER_SYNC_COMMITTEE_PERIOD
+  withState(dag.headState.data):
+    when stateFork >= forkAltair:
+      let
+        headSlot = state.data.slot
+        headCommitteePeriod = syncCommitteePeriod(headSlot)
+        periodStart = syncCommitteePeriodStartSlot(headCommitteePeriod)
+        nextPeriodStart = periodStart + SLOTS_PER_SYNC_COMMITTEE_PERIOD
 
-    if slot >= nextPeriodStart:
-      @(dag.headState.data.hbsAltair.data.next_sync_committee.pubkeys.data)
-    elif slot >= periodStart:
-      @(dag.headState.data.hbsAltair.data.current_sync_committee.pubkeys.data)
+      if slot >= nextPeriodStart:
+        @(state.data.next_sync_committee.pubkeys.data)
+      elif slot >= periodStart:
+        @(state.data.current_sync_committee.pubkeys.data)
+      else:
+        @[]
     else:
       @[]
-  else:
-    @[]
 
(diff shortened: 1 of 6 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1052,21 +1052,22 @@ func syncCommitteeParticipants*(dagParam: ChainDAGRef,
...
-      headCommitteePeriod = syncCommitteePeriod(headSlot)
-      periodStart = syncCommitteePeriodStartSlot(headCommitteePeriod)
-      nextPeriodStart = periodStart + SLOTS_PER_SYNC_COMMITTEE_PERIOD
-
-    if slot >= nextPeriodStart:
-      @(dag.headState.data.hbsAltair.data.next_sync_committee.pubkeys.data)
-    elif slot >= periodStart:
-      @(dag.headState.data.hbsAltair.data.current_sync_committee.pubkeys.data)
+  withState(dag.headState.data):
+    when stateFork >= forkAltair:
```
arnetheduck, line 1056: alternative that catches unintended upgrades:
```
when stateFork in {a, b}:
elif in {the_others}:
else: error
```

etan-status: Hmm, not sure about this, for these reasons:

- The spirit of this specific change seems to be "hey, there is a new feature, sync_committees. it will apply from altair onwards", i.e., it is not expected to go away randomly. If it does go away, we still get a compilation error because the field no longer exists. If there is a more subtle change within the sync committee, the check is at the wrong level and would need to be further inside a sub-function.
- The structure is not checked by the compiler for being exhaustive; we would need a `when case` expression of some kind for that purpose. So, we would only see the problem at run time at some future time, and depending on the change, only under certain situations.
- With the `>=`, the compiler will try to generate the code for new forks immediately, revealing exactly which of the called functions will require a new overload to support blocks of the new fork. With the alternative, you would first require to run the code a couple time, until you hit all the `else` occurrences, then add the new fork, then find out about the missing functions.
- Just because this specific code block requires special handling for pre/post altair states, does not mean that future forks will require the special handling at the same place. Technically, we would have to wrap every single line in such a block in case that a future fork introduces differences there. It is basically a "this code was reviewed up through fork X" marker. There are probably better tools to track that.
- EF has a lot of code of the form `is_post_altair(spec)` / `is_post_merge(spec)`. The proposed `>=` code is in line with how EF tests their code.

etan-status: One more reason:
- It would add visual clutter, i.e., it would refer to `merge` / `sharding` / whatever, although this code has nothing to do with those forks. Probably fine for now, but these won't be the last forks.

etan-status: Condensing my previous comments here, as they got a bit lengthy:

While the alternative would indeed catch unintended upgrades (at runtime), it would not fix the problem of intended upgrades. Each fork targets specific aspects of the beacon chain, it is very unlikely that future forks will touch this specific line of the code. Instead, they will introduce new branches on other lines that are not currently guarded by a similar construct.

The general spirit of `>=` comparison is also further promoted by EF functions `is_post_altair(spec)` / `is_post_merge(spec)`.

Finally, there is a developer convenience with `>=`, because one can simply search for `altair` in the IDE and only see altair-specific changes.

Thanks for the suggestion, though.

Issue #2847, opened 2021-09-05
[Altair] Nimbus crashes with SEGV every few hours


jclapis:
**Describe the bug**
Nimbus is crashing with a segfault every 2 to 10 hours, somewhat randomly. I've been [working with @stefantalpalaru on it in the dev-helpdesk Discord channel](https://discord.com/channels/613988663034118151/616299964242460682/882862136357945357). Using gdb, we tracked it down to this:

```
Thread 3 "nimbus_beacon_n" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xfffff795db90 (LWP 21968)]
0x0000aaaaaab21d50 in copyStringRC1 ()
(gdb) bt full
#0  0x0000aaaaaab21d50 in copyStringRC1 ()
No symbol table info available.
#1  0x0000aaaaaabe5bf8 in cb__Lz4sYG9aigPLc8Kg9bvj68Cw ()
No symbol table info available.
#2  0x0000aaaaaabe5bf8 in cb__Lz4sYG9aigPLc8Kg9bvj68Cw ()
No symbol table info available.
#3  0x0000aaaaaabe5bf8 in cb__Lz4sYG9aigPLc8Kg9bvj68Cw ()
No symbol table info available.
```
The latter lines repeat for tens of thousands of rows, suggesting this is a stack overflow.

Thread info:
```
(gdb) info threads
  Id   Target Id                                           Frame 
  1    Thread 0xfffff7d62010 (LWP 21964) "nimbus_beacon_n" __mul_384 () at /home/user/nimbus-eth2/vendor/nim-blscurve/vendor/blst/build/elf/mul_mont_384-armv8.S:1457
  2    Thread 0xfffff7c5db90 (LWP 21967) "nimbus_beacon_n" futex_wait_cancelable (private=0, expected=0, futex_word=0xaaaaaee0d628 <exportChan__Y9bSsqXOvK4gavMDqtF9beUQ+136>)
    at ../sysdeps/nptl/futex-internal.h:183
* 3    Thread 0xfffff795db90 (LWP 21968) "nimbus_beacon_n" 0x0000aaaaaab21d50 in copyStringRC1 ()
```

The linked Discord thread has more context between @stefantalpalaru and @cheatfate which is beyond me, but is likely useful to solving the problem.

Note that Prometheus was active and collecting metrics from Nimbus at the time (**metrics were enabled**). 

**To Reproduce**
[...]

PR #2944, opened 2021-10-01
dist: cross-build ARM64 on Debian-10.10

stefantalpalaru:
instead of Ubuntu-18.04. This fixes a crash when stress-testing the metrics HTTP server. We think it was due to gcc-7.4.0 miscompiling something. The new distro/version gives us gcc-8.3.0, but with the same old glibc-2.28 we need on Raspbian.

Kudos to @jclapis for reporting and debugging the problem.


jclapis:
This was a *wild* ride to debug. Relevant Discord conversation starts here for future reference: https://discord.com/channels/613988663034118151/812268053718237196/893163883949330473

This closes #2847.

etan-status:
Wow

PR #2949, opened 2021-10-04
register vc duties with subnet tracker

arnetheduck:
* fix activation logging during startup
* cache slot signature to avoid duplicate signature work
* schedule aggregation duties one slot at a time to avoid CPU spike at
each epoch
* lower aggregation subnet pre-subscription time to 4 slots (lowers
bandwidth and CPU usage)

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -457,9 +457,18 @@ proc init*(T: type BeaconNode,
     except Exception as exc: raiseAssert exc.msg
     node.addRemoteValidators()
 
-  for validator in node.attachedValidators[].validators.values():
-    if validator.index.isSome():
-      node.actionTracker.knownValidators[validator.index.get()] = node.beaconClock.now().slotOrZero()
+  block:
+    # Add in-process validators to the list of "known" validators such that
+    # we start with a reasonable ENR
+    let wallSlot = node.beaconClock.now().slotOrZero()
+    for validator in node.attachedValidators[].validators.values():
+      if validator.index.isSome():
+        node.actionTracker.knownValidators[validator.index.get()] = wallSlot
+
+    let stabilitySubnets = node.actionTracker.stabilitySubnets(wallSlot)
+    # Here, we also set the correct ENR should we be in all subnets mode!
+    node.network.updateStabilitySubnetMetadata(stabilitySubnets)
+
   network.initBeaconSync(dag, getBeaconTime)
 
   node.updateValidatorMetrics()
@@ -499,8 +508,7 @@ proc updateAttestationSubnetHandlers(node: BeaconNode, slot: Slot) =
     stabilitySubnets = node.actionTracker.stabilitySubnets(slot)
     subnets = aggregateSubnets + stabilitySubnets
 
-  if not node.actionTracker.subscribeAllSubnets:
-    node.network.updateStabilitySubnetMetadata(stabilitySubnets)
+  node.network.updateStabilitySubnetMetadata(stabilitySubnets)
 
   # Now we know what we should be subscribed to - make it so
   let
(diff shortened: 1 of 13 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -892,7 +659,18 @@ proc trackSyncCommitteeTopics*(node: BeaconNode) =
...
+  node.actionTracker.updateSlot(slot)
+
+  # Let the tracker know what duties are approaching - this will tell us how
+  # many stability subnets we need to be subscribed to and what subnets we'll
+  # soon be aggregating - in addition to the in-beacon-node duties, there may
+  # also be duties coming from the validator client, but we don't control when
+  # these arrive
+  await node.registerDuties(slot)
+
   # Syncing tends to be ~1 block/s, and allow for an epoch of time for libp2p
```
tersec: Is this ~1 blocks/slot still true? It was an empirical measurement at the time.

beacon_chain/validators/action_tracker.nim
```diff
@@ -0,0 +1,198 @@
...
+func updateSlot*(tracker: var ActionTracker, wallSlot: Slot) =
+  # Prune duties from the past - this collection is kept small because there
+  # are only so many slot/subnet combos - prune both internal and API-supplied
+  # duties at the same time
+  tracker.duties.keepItIf(it.slot >= wallSlot)
+
+  # Keep stability subnets for as long as validators are validating
+  var toPrune: seq[ValidatorIndex]
+  for k, v in tracker.knownValidators:
+    if v + SLOTS_PER_EPOCH * 3 < wallSlot: toPrune.add k
```
tersec: This is a little arbitrary, though reasonable -- how expensive are ENR updates?

beacon_chain/validators/action_tracker.nim
```diff
@@ -0,0 +1,198 @@
...
+    knownValidators*: Table[ValidatorIndex, Slot]
+      ## Validators that we've recently seen - we'll subscribe to one stability
+      ## subnet for each such validator - the slot is used to expire validators
+      ## that no longer are posting duties
+
+    duties*: seq[AggregatorDuty] ##\
+      ## Known aggregation duties in the near future - before each such
+      ## duty, we'll subscribe to the corresponding subnet to collect
+
+# https://github.com/ethereum/consensus-specs/blob/v1.0.1/specs/phase0/validator.md#phase-0-attestation-subnet-stability
```
tersec: There's still utility in noting where, yes, this functionality has existed in this form since v1.0.1, but the immediately relevant thing is that it's still this way, i.e. currently in v1.1.2.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -481,185 +485,27 @@ func toBitArray(stabilitySubnets: auto): BitArray[ATTESTATION_SUBNET_COUNT] =
...
-        node.attestationSubnets.aggregateSubnets[i] = true
+    aggregateSubnets = node.actionTracker.aggregateSubnets(slot)
+    stabilitySubnets = node.actionTracker.stabilitySubnets(slot)
+    subnets = aggregateSubnets + stabilitySubnets
 
-  # Accounting specific to non-stability subnets
-  for i in (prevAggregateSubnets - node.attestationSubnets.aggregateSubnets).
-      oneIndices():
-    node.attestationSubnets.subscribeSlot[i] = FAR_FUTURE_SLOT
+  if not node.actionTracker.subscribeAllSubnets:
```
tersec: `node.network.updateStabilitySubnetMetadata` effectively already checks this condition and will also early-exit

beacon_chain/validators/action_tracker.nim
arnetheduck: the thinking here was that we in theory have a duty every epoch, but then we might not once we hit 260k, so we add an extra epoch just to be safe and another to avoid messy rounding

beacon_chain/nimbus_beacon_node.nim
arnetheduck: eh, more like 4-5 blocks per second..

beacon_chain/validators/action_tracker.nim
arnetheduck: fixed, copypaste thing

beacon_chain/validators/action_tracker.nim
arnetheduck: documented

beacon_chain/nimbus_beacon_node.nim
arnetheduck: removed the if

PR #2958, opened 2021-10-05
update defaultRuntimeConfig for v1.1.1

etan-status:
This synchronizes defaultRuntimeConfig with v1.1.1 of the spec.
- Introduces `TERMINAL_TOTAL_DIFFICULTY` and `TERMINAL_BLOCK_HASH`.
- Sets `ALTAIR_FORK_EPOCH` in `mainnet` preset.

### DIFF
--- a/beacon_chain/spec/presets.nim
+++ b/beacon_chain/spec/presets.nim
@@ -9,8 +9,7 @@
 
 import
   std/[macros, strutils, parseutils, tables],
-  stew/endians2,
-  web3/[ethtypes]
+  stew/endians2, stint, web3/[ethtypes]
 
 export
   toBytesBE
@@ -30,9 +29,13 @@ type
   Eth1Address* = ethtypes.Address
 
   RuntimeConfig* = object
-    ## https://github.com/ethereum/eth2.0-specs/tree/v1.1.0-alpha.8/configs
+    ## https://github.com/ethereum/consensus-specs/tree/v1.1.1/configs
 
     PRESET_BASE*: string
+
+    # Transition
+    TERMINAL_TOTAL_DIFFICULTY*: UInt256
+    TERMINAL_BLOCK_HASH*: BlockHash
 
     # Genesis
     MIN_GENESIS_ACTIVE_VALIDATOR_COUNT*: uint64
@@ -144,10 +147,6 @@ const
 
     "CONFIG_NAME",
 
-    # Merge-related settings that are already part of the mainnet config:
-    "TERMINAL_TOTAL_DIFFICULTY",
-    "TERMINAL_BLOCK_HASH",
-
     "TRANSITION_TOTAL_DIFFICULTY", # Name that appears in some altair alphas, obsolete, remove when no more testnets
   ]
 
@@ -158,10 +157,19 @@ when const_preset == "mainnet":
   # TODO Move this to RuntimeConfig
   const SECONDS_PER_SLOT* {.intdefine.}: uint64 = 12
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.1.0-alpha.8/configs/mainnet.yaml
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.1/configs/mainnet.yaml
   # TODO Read these from yaml file
   const defaultRuntimeConfig* = RuntimeConfig(
     PRESET_BASE: "mainnet",
+
+    # Transition
+    # ---------------------------------------------------------------
+    # TBD, 2**256-2**10 is a placeholder
+    TERMINAL_TOTAL_DIFFICULTY:
+      u256"115792089237316195423570985008687907853269984665640564039457584007913129638912",
+    # By default, don't use this param
+    TERMINAL_BLOCK_HASH: BlockHash.fromHex(
+      "0x0000000000000000000000000000000000000000000000000000000000000000"),
 
     # Genesis
     # ---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


arnetheduck:
There's a TODO here to not hardcode the `defaultRuntimePreset` constants but rather load them from the yaml file - should be _fairly_ simple to fix this properly, can you take a look?

etan-status:
Sure, I'm currently facing weird compile errors, only on `i386`, here, and meanwhile the CI somehow abandoned this PR completely. Will take a look at the YAML once that is resolved / as a followup.

```
/home/runner/work/nimbus-eth2/nimbus-eth2/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1265, 5) template/generic instantiation of `closureScope` from here
23
/home/runner/work/nimbus-eth2/nimbus-eth2/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1267, 19) template/generic instantiation of `addAsyncValidator` from here
24
/home/runner/work/nimbus-eth2/nimbus-eth2/nimbus-eth2/beacon_chain/networking/eth2_network.nim(2011, 26) template/generic instantiation of `decode` from here
25
/home/runner/work/nimbus-eth2/nimbus-eth2/nimbus-eth2/vendor/nim-serialization/serialization.nim(67, 31) Error: type mismatch: got <type SszReader, InputStreamHandle>
26
Makefile:290: recipe for target 'nimbus_beacon_node' failed
27
but expected one of: 
28
proc init[A](s: var HashSet[A]; initialSize = defaultInitialSize)
29
30
  required type for s: var HashSet[init.A]
31
  but expression 'ReaderType`gensym478035009' is of type: type SszReader
32

33
expression: init(ReaderType`gensym478035009, stream`gensym478035008)
34
```

arnetheduck:
any imports you changed, just try moving them around, for example after json_serialization

etan-status:
CI issues were just intermittent. It resumed to run :-)

etan-status:
The compilation problems were indeed due to some `import` / `export` shenanigans. It's weird that they only affect i386, but whatever :-)

Bundled the fix with https://github.com/status-im/nimbus-eth2/pull/2960, once that one is merged, this here can be rebased and should build successfully.

etan-status:
The YAML import I'd like to look into separately, having the `defaultRuntimeConfig` updated and helpers for the new types active is already good to have, especially with the general flakiness that this area seems to have.

beacon_chain/spec/presets.nim
```diff
@@ -237,13 +245,22 @@ elif const_preset == "minimal":
...
     # Extends the minimal preset
     PRESET_BASE: "minimal",
 
+    # Transition
+    # ---------------------------------------------------------------
+    # TBD, 2**256-2**10 is a placeholder
+    TERMINAL_TOTAL_DIFFICULTY:
+      u256"115792089237316195423570985008687907853269984665640564039457584007913129638912",
+    # By default, don't use this param
+    TERMINAL_BLOCK_HASH: BlockHash.fromHex(
```
zah, line 261: Let's keep the usage of the `BlockHash` type to the absolute minimum. The type exists only because of an inappropriate design of our package dependencies resulting in multiple competing definitions for the Ethereum types. We would like to migrate to a world where every type is defined once and used consistently everywhere. `BlockHash` won't exist in that world.

etan-status: What would be the appropriate type here? `Eth2Digest`?

The inspiration for this was your commit https://github.com/status-im/nimbus-eth2/commit/81dc6c85599fb11d7a5153f8bc62a29c37d6ed02

arnetheduck: > where every type is defined once and used consistently everywhere

hm, there are some problems with that approach - mainly that you end up with massive depedency trees but also that when you change a type slightly because one consumer of the type needs something (for example a formatting of json or whatever), _all_ dependent code has to change as well - putting up some barriers between types isn't a bad idea, often

zah: I would respectufully disagree that putting the said barrier is a good idea in our particular situation.

zah: > The inspiration for this was your commit 81dc6c8

My changes are in the `eth1_monitor` module that needs to talk to an execution engine over JSON-RPC. Right now, types like `BlockHash` are the native types of nim-web3 (our json-rpc based library for consuming the Eth1 APIs). There are no good reasons for code outside the `eth1_monitor` module to work with these types.

etan-status: https://github.com/status-im/nimbus-eth2/blob/amphora-merge-interop/beacon_chain/spec/presets.nim#L73

arnetheduck: well, the dep tree is a practical problem - right now, to compile some simple spec tests, we also pull in libp2p which pulls in chronos and a bunch of crypto libraries that have nothing to do with the spec, merely because somewhere there's some silly type reused.

the reverse dep problem is actually more of a concern - central points of types also become central points of failure meaning a small seemingly benign change chan break many downstreams.

Anyway, this wasn't about this change in particular, rather a problem that we've been running into more frequently as the dep trees get more complex because not all projects move at the same upgrade pace.

etan-status:
(trying whether this compiles fine without ssz change. SSZ should not be required because this field is not exchanged via RPC only via REST).

etan-status:
Compiles fine. The CI issues were just intermittent connection issues.

PR #2961, opened 2021-10-06
Number of REST fixes.

cheatfate:
1. Fix /eth/v1/config/spec response.
2. Cache /eth/v1/node/version response.
3. Workaround `{state_id}` issue.
4. Fix number of issues in `/eth/v1/events` responses.


etan-status:
I have a similar PR lined up (without the caching) that extends those fixes for the RPC api as well. Will submit once this one clears.

beacon_chain/rpc/rest_config_api.nim
```diff
@@ -22,6 +22,12 @@ proc installConfigApiHandlers*(router: var RestRouter, node: BeaconNode) =
         (
           CONFIG_NAME:
             const_preset,
+          PRESET_BASE:
```
etan-status, line 25: I have extended this similarly, also with the remaining constants, here: https://github.com/status-im/nimbus-eth2/pull/2962

arnetheduck (approved):

PR #2963, opened 2021-10-06
allow `mock_blocks` to create Merge blocks

etan-status:
This extends `mock_blocks` support to also allow producing Merge blocks.
The only step missing was initialization of the `ExecutionPayload` to an
empty value (as created by `build_empty_execution_payload`).

### DIFF
--- a/tests/mocking/mock_blocks.nim
+++ b/tests/mocking/mock_blocks.nim
@@ -7,79 +7,118 @@
 
 import
   # Beacon chain internals
-  ../../beacon_chain/spec/[forks, helpers, signatures, state_transition],
+  ../../beacon_chain/spec/
+    [forks, helpers, signatures, state_transition, validator],
+  ../../beacon_chain/spec/datatypes/[phase0, altair, merge],
   # Test utilities
   ../testblockutil
 
 # Routines for mocking blocks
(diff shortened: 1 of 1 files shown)


etan-status:
Depends on https://github.com/status-im/nimbus-eth2/pull/2943

etan-status:
This is infrastructure for future tests. Ideally, we would already have tests using it, but not there yet.

arnetheduck:
needs to be switched to `unstable`

tests/mocking/mock_blocks.nim
```diff
@@ -7,79 +7,118 @@
...
+
+  blck.root = blck.message.hash_tree_root()
+  blck.signature = get_block_signature(
+    state.fork,
+    state.genesis_validators_root,
+    blck.message.slot,
+    blck.root,
+    privkey).toValidatorSig()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.1.0/tests/core/pyspec/eth2spec/test/helpers/execution_payload.py#L1-L31
```
tersec: Are these also v1.1.1?

etan-status: Yep! Updated.

PR #2964, opened 2021-10-06
interop metrics

stefantalpalaru:
https://github.com/ethereum/beacon-metrics/blob/master/metrics.md#interop-metrics


tersec (approved):

zah:
How is this PR implementing the linked spec?

In the spec `beacon_reorgs_total` is cited as a counter and here, which was the state of the code before the changes here.
I should also mention that the linked spec is very old and limited in scope to the initial Eth2 clients interop event. There should be more up-to-date resources on recommended client metrics.

stefantalpalaru:
> In the spec beacon_reorgs_total is cited as a counter

I hope that's not relevant for metric parsers. Our counters append "_total" to the main metric's name. While that would have worked here, there are other metrics coming which don't happen to end in "_total", so I prefer using gauges everywhere.

>  the linked spec is very old

And yet some clients did not implement all of it:

compat_metrics

> There should be more up-to-date resources on recommended client metrics.

Discord discussions from yesterday are as up-to-date as it gets ;-)

PR #2965, opened 2021-10-07
harden sync sub committee selection

arnetheduck:
also turn it into an iterator

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1027,19 +1027,16 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
     prunedHeads = hlen - dag.heads.len,
     dagPruneDur = Moment.now() - startTick
 
-func syncSubcommittee*(syncCommittee: openarray[ValidatorPubKey],
-                       committeeIdx: SyncCommitteeIndex): seq[ValidatorPubKey] =
-  ## TODO Return a view type
-  ## Unfortunately, this doesn't work as a template right now.
-  if syncCommittee.len == 0:
-    return @[]
+iterator syncSubcommittee*(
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1081,7 +1078,7 @@ func getSubcommitteePositionsAux(
     return @[]
   let validatorPubKey = validatorKey.get().toPubKey
 
-  for pos, key in syncCommittee.syncSubcommittee(committeeIdx):
+  for pos, key in toSeq(syncCommittee.syncSubcommittee(committeeIdx)):
```
mratsim, line 1081: Why not use `enumerate` or make the iterator return a tuple (pos, key)?

That does allocate a large sequence given that each ValidatorPubKey is 48 bytes.

mratsim (commented):
What was the hardening?

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: `enumerate` is 1.4+ and `pairs` requires a separate copy of the iterator (most of the time it's called without pos) - it was a seq before this change, so this is not a pessimization at least

arnetheduck:
> What was the hardening?

out of bounds committee index leads to Defect which leads to crash

arnetheduck:
(it happened in some test code we had)

tersec (approved):

beacon_chain/spec/forks.nim
```diff
@@ -68,7 +68,7 @@ type
     of BeaconBlockFork.Merge:
       mergeBlock*:  merge.TrustedSignedBeaconBlock
 
-  ForkDigests* {.requiresInit.} = object
+  ForkDigests* = object
```
etan-status, line 71: Submarine fix.

PR #2969, opened 2021-10-08
bump `nim-stint`

etan-status:
Upstream changes:
- Parsing should raise CatchableError for user inputs
- RangeError are actually defect and not catchable.

### DIFF
--- a/beacon_chain/interop.nim
+++ b/beacon_chain/interop.nim
@@ -28,11 +28,10 @@ func makeInteropPrivKey*(i: int): ValidatorPrivKey =
   var bytes: array[32, byte]
   bytes[0..7] = uint64(i).toBytesLE()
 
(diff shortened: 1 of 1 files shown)


beacon_chain/interop.nim
```diff
@@ -28,11 +28,10 @@ func makeInteropPrivKey*(i: int): ValidatorPrivKey =
   var bytes: array[32, byte]
   bytes[0..7] = uint64(i).toBytesLE()
 
+  # BLS381-12 curve order - same as milagro but formatted differently
+  const curveOrder =
+    u256"52435875175126190479447740508185965837690552500527637822603658699938581184513"
```
etan-status, line 33: Using `const` to avoid having to catch the newly introduced exception.

mratsim (approved):
LGTM

PR #2974, opened 2021-10-11
replace obsolete Rayonism interop docs with Amphora interop docs

tersec:
Nothing outside `docs/`


arnetheduck (approved):

arnetheduck:
would be good to include docs to each of the execution clients that support running the merge - basically the links from the amphora green checkmark hackmd

tersec:
Probably, but those can be separate PRs.

PR #2975, opened 2021-10-11
use specified execution_payload when constructing blocks; update Geth launch script

tersec:
- 80 character line + `merge.Foo` for `is_merge_block()`
- pass through the specified `ExecutionPayload` for merge block construction
- update `scripts/run-catalyst.sh` from Rayonism to Amphora

### DIFF
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -392,6 +392,10 @@ type
     branch*: array[DEPOSIT_CONTRACT_TREE_DEPTH, Eth2Digest]
     deposit_count*: array[32, byte] # Uint256
 
+  DepositContractSnapshot* = object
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/datatypes/base.nim
```diff
@@ -392,6 +392,10 @@ type
     branch*: array[DEPOSIT_CONTRACT_TREE_DEPTH, Eth2Digest]
     deposit_count*: array[32, byte] # Uint256
 
+  DepositContractSnapshot* = object
```
arnetheduck: hm - this is a nimbus-specific thing though, ie not spec afair..

tersec: https://github.com/status-im/nimbus-eth2/pull/2975/commits/048e8b186c46358c9673a5ba1a086576f923d431

PR #2980, opened 2021-10-12
support merkle multiproof verification

etan-status:
This introduces spec-compatible support to verify merkle multiproofs for
use, e.g., in light client syncing.

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -11,9 +11,10 @@
 
 import
   # Standard lib
-  std/[math, tables],
-  # Third-party
+  std/[algorithm, intsets, math, sequtils, tables],
+  # Status libraries
   stew/[byteutils, endians2, bitops2],
+  chronicles,
   # Internal
   ./datatypes/[phase0, altair, merge],
   ./eth2_merkleization, ./ssz_codec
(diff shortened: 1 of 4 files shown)


beacon_chain/spec/helpers.nim
```diff
@@ -114,12 +115,233 @@ func build_proof_impl(anchor: object, leaf_index: uint64,
...
+  var all_path_indices = initIntSet()
+  for index in indices:
+    for idx in get_branch_indices(index):
+      all_helper_indices.incl idx.int
+    for idx in get_path_indices(index):
+      all_path_indices.incl idx.int
+  all_helper_indices.excl all_path_indices
+  result = newSeqOfCap[GeneralizedIndex](all_helper_indices.len)
+  for idx in all_helper_indices:
+    result.add idx.GeneralizedIndex
```
mratsim, line 107: I'm not sure if this is a bottleneck but indexing has less overhead than `add`

Hence

```Nim
  result = newSeq[GeneralizedIndex](all_helper_indices.len)
  for i,idx in all_helper_indices:
    result[i] = idx.GeneralizedIndex
```

Style nit: we avoid using the implicit result unless forced, see https://status-im.github.io/nim-style-guide/language.result.html

mratsim (approved):
LGTM, style nit on result accumulation.

Note if this is needed here are Depth-First and Breadth-first traversal of binary trees with no stack/queue allocations (for 0-based trees): https://github.com/mratsim/weave/blob/71dc2d7/weave/datatypes/binary_worker_trees.nim#L91-L281

beacon_chain/spec/helpers.nim
etan-status: Thanks, and good to know that indexing is performing better than adding in such cases.
Went with `newSeqUninitialized` as well to avoid the extra memset, so that this line is still just a pre-allocation.

Sadly, `intsets` does not support `pairs` (it is not ordered), so this got quite a bit more verbose.
```
  var
    res = newSeqUninitialized[GeneralizedIndex](all_helper_indices.len)
    i = 0
  for idx in all_helper_indices:
    res[i] = idx.GeneralizedIndex
    inc i
  res.sort(SortOrder.Descending)
  res
```

arnetheduck:
Can this replace https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/eth1/merkle_minimal.nim?

etan-status:
I think the code you linked is for the opposite direction (data to proof). This here is for proof verification.

beacon_chain/spec/helpers.nim
mratsim: Ah, well given that we don't have enumerate either, if this is not a perf critical part of the code the original one is probably best for maintenance/clarity.

etan-status: Agree, `newSeqUninitialized` is just a wrapper around `newSeqOfCap` with the exception that the `len` is set in one go. However, as we have the counter here in the client code, we still count up to `len` one by one.

And the `res` does not solve the problems that `result` has, in this specific instance. The function is still wrong if the final `res` line is forgotten, and the error is only noticed at run-time.

PR #2987, opened 2021-10-13
Number of REST fixes.

cheatfate:
1. Add https://ethereum.github.io/beacon-APIs/#/Validator/getSyncCommitteeDuties implementation.
2. Fix order issue for https://ethereum.github.io/beacon-APIs/#/Beacon/getEpochSyncCommittees
3. Add `validator pubkey -> validator index` caching mechanism.
4. Refactor validator API calls to use new caching mechanism.
5. Fix error code when state could not be found.

### DIFF
--- a/beacon_chain/beacon_node_common.nim
+++ b/beacon_chain/beacon_node_common.nim
@@ -21,7 +21,7 @@ import
   ./networking/eth2_network,
   ./eth1/eth1_monitor,
   ./consensus_object_pools/[blockchain_dag, block_quarantine, attestation_pool],
-  ./spec/datatypes/base, ./spec/eth2_apis/rest_types,
+  ./spec/datatypes/base,
   ./sync/[sync_manager, request_manager]
 
 export
@@ -70,7 +70,7 @@ type
     beaconClock*: BeaconClock
     taskpool*: TaskPoolPtr
     onAttestationSent*: OnAttestationCallback
-    restKeysCache*: Table[ValidatorPubKey, RestValidatorIndex]
+    restKeysCache*: Table[ValidatorPubKey, ValidatorIndex]
 
 const
   MaxEmptySlotCount* = uint64(10*60) div SECONDS_PER_SLOT
--- a/beacon_chain/beacon_node_common.nim
+++ b/beacon_chain/beacon_node_common.nim
@@ -21,7 +21,7 @@ import
   ./networking/eth2_network,
   ./eth1/eth1_monitor,
   ./consensus_object_pools/[blockchain_dag, block_quarantine, attestation_pool],
-  ./spec/datatypes/base,
+  ./spec/datatypes/base, ./spec/eth2_apis/rest_types,
   ./sync/[sync_manager, request_manager]
 
 export
@@ -70,6 +70,7 @@ type
     beaconClock*: BeaconClock
     taskpool*: TaskPoolPtr
     onAttestationSent*: OnAttestationCallback
+    restKeysCache*: Table[ValidatorPubKey, RestValidatorIndex]
 
 const
   MaxEmptySlotCount* = uint64(10*60) div SECONDS_PER_SLOT
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -411,8 +411,12 @@ proc init*(T: type BeaconNode,
     requestManager: RequestManager.init(network, blockProcessor),
     beaconClock: beaconClock,
     taskpool: taskpool,
-    onAttestationSent: onAttestationSent
+    onAttestationSent: onAttestationSent,
   )
+
+  # initialize REST server cache tables.
+  if config.restEnabled:
+    node.restKeysCache = initTable[ValidatorPubKey, RestValidatorIndex](256)
 
   # set topic validation routine
   network.setValidTopics(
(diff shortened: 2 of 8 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -411,9 +411,13 @@ proc init*(T: type BeaconNode,
     requestManager: RequestManager.init(network, blockProcessor),
     beaconClock: beaconClock,
     taskpool: taskpool,
-    onAttestationSent: onAttestationSent
+    onAttestationSent: onAttestationSent,
   )
 
+  # initialize REST server cache tables.
+  if config.restEnabled:
+    node.restKeysCache = initTable[ValidatorPubKey, RestValidatorIndex](256)
```
arnetheduck: the table will init itself, this looks redundant? Basically, we have no way of knowing how many validators will appear, so no point doing special magic here because the asymptotic performance is  the same

beacon_chain/beacon_node_common.nim
```diff
@@ -70,6 +70,7 @@ type
     beaconClock*: BeaconClock
     taskpool*: TaskPoolPtr
     onAttestationSent*: OnAttestationCallback
+    restKeysCache*: Table[ValidatorPubKey, RestValidatorIndex]
```
arnetheduck: why not use `ValidatorIndex` here?

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -259,59 +241,95 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+              if item.index in indexset:
+                return RestApiResponse.jsonError(Http400,
+                                                 UniqueValidatorIndexError)
+              indexset.incl(item.index)
+          if len(keyset) > 0:
+            let optIndices = keysToIndices(node.restKeysCache, stateData.data,
+                                           keyset.toSeq())
+            for item in optIndices:
+              if item.isNone():
+                return RestApiResponse.jsonError(Http400,
```
arnetheduck, line 287: so if I supply 10 validators that are found and one that isn't, this results in a complete failure? seems harsh / difficult to work with

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -259,59 +241,95 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+          var res: seq[RestValidator]
+          if len(indices) == 0:
+            # There is no indices, so we going to filter all the validators.
+            for index, validator in getStateField(stateData.data,
+                                                  validators).pairs():
+              let
+                balance = getStateField(stateData.data, balances)[index]
+                status =
+                  block:
+                    let sres = validator.getStatus(current_epoch)
```
arnetheduck, line 307: shouldn't this be the status at the slot of the state that's being asked for?

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -259,59 +241,95 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+    node.withStateForBlockSlot(bslot):
+      let
+        current_epoch = get_current_epoch(node.dag.headState.data)
+        validatorsCount = lenu64(getStateField(stateData.data, validators))
+
+      let indices =
+        block:
+          var indexList: seq[ValidatorIndex]
+          var keyset: HashSet[ValidatorPubKey]
+          var indexset: HashSet[RestValidatorIndex]
```
arnetheduck: this looks very complex for the value that it is providing - why not just use a single `HashSet[ValidatorIndex]` and iterate over validatorIds and validate them one by one, adding them to the hashset and return error if a duplicate is found? arguably it's more correct also, ie if a key duplicates an index and the index came "first", the key error should be returned and vice versa - now it groups keys and indices..

beacon_chain/rpc/rest_utils.nim
```diff
@@ -322,5 +326,53 @@ proc toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
...
+                    forkedState: ForkedHashedBeaconState,
+                    keys: openArray[ValidatorPubKey]
+                   ): seq[Option[RestValidatorIndex]] =
+  var indices = newSeq[Option[RestValidatorIndex]](len(keys))
+  let keyset =
+    block:
+      var res: Table[ValidatorPubKey, int]
+      for inputIndex, pubkey in keys.pairs():
+        # Try to search in cache first.
+        let validatorIndex = cacheTable.getOrDefault(pubkey,
```
arnetheduck:
```
catchTable.withValue(pubkey, index) do: # when found
  indices[inputIndex] = some(vidx)
do: # when not found
  res[...] = inputINdex
```

beacon_chain/rpc/rest_beacon_api.nim
cheatfate: But it how its works. Check `infura.io` with something like 
```
curl -X GET "https://<your API key>@eth2-beacon-prater.infura.io/eth/v1/beacon/states/head/validators?id=0x8107ff6a5cfd1993f0dc19a6a9ec7dc742a528dd6f2e3e10189a4a6fc489ae6c7ba9070ea4e2e328f0d20b91cc129733&id=0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```

PR #2994, opened 2021-10-18
load altair state from database on startup

arnetheduck:
* fixes replay from last phase0 block on startup
* better forked block loading

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -313,10 +313,10 @@ func contains*(dag: ChainDAGRef, root: Eth2Digest): bool =
 
 proc containsBlock(
     cfg: RuntimeConfig, db: BeaconChainDB, blck: BlockRef): bool =
-  case cfg.stateForkAtEpoch(blck.slot.epoch)
-  of forkMerge:  db.containsBlockMerge(blck.root)
-  of forkAltair: db.containsBlockAltair(blck.root)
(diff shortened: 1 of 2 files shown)


tersec (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -625,11 +613,21 @@ proc getState(dag: ChainDAGRef, state: var StateData, bs: BlockSlot): bool =
...
-      stateRoot.isSome()):
-    return dag.getState(state, stateRoot.get(), bs.blck)
+  let stateRoot = dag.db.getStateRoot(bs.blck.root, bs.slot)
+  if stateRoot.isNone(): return false
 
-  false
+  let restoreAddr =
+    # Any restore point will do as long as it's not the object being updated
+    if unsafeAddr(state) == unsafeAddr(dag.headState):
+      unsafeAddr dag.clearanceState
```
zah, line 622: What are we doing here exactly? If `getStateData` fails, we are restoring the `state` object to an arbitrary valid state? What's the point of doing this?

arnetheduck: `StateData` must be internally consistent because the instance is reused over and over (as opposted to being garbage collected) - the state data must match the (cached) root and the block reference, so if a partial state transition happens, we must revert to any "known-good" point - maintaining this invariant helps simplify other code which can always rely on these states being valid and therefore sources of truth for various functionality

PR #2998, opened 2021-10-19
refactor EF operations fixtures

tersec:
It still, for now, keeps phase0/altair/merge worlds distinct.


tests/consensus_spec/phase0/test_fixture_operations.nim
```diff
@@ -0,0 +1,141 @@
...
+  ../../helpers/debug_state
+
+const
+  OpDir                 = SszTestsDir/const_preset/"phase0"/"operations"
+  OpAttestationsDir     = OpDir/"attestation"/"pyspec_tests"
+  OpAttSlashingDir      = OpDir/"attester_slashing"/"pyspec_tests"
+  OpBlockHeaderDir      = OpDir/"block_header"/"pyspec_tests"
+  OpDepositsDir         = OpDir/"deposit"/"pyspec_tests"
+  OpProposerSlashingDir = OpDir/"proposer_slashing"/"pyspec_tests"
+  OpVoluntaryExitDir    = OpDir/"voluntary_exit"/"pyspec_tests"
```
arnetheduck: should really have a test `check toseq(walkDir(OpDir)) == initHashSet[OpAttestationDir, OpAttSlash...)` etc

tests/consensus_spec/altair/test_fixture_operations.nim
```diff
@@ -0,0 +1,162 @@
+# beacon_chain
```
arnetheduck, line 1: wdyt of moving rewards in here also? they're ops tests as well?

arnetheduck (approved):

tests/consensus_spec/altair/test_fixture_operations.nim
tersec: Not similar enough, for a first pass, and they're not actually ops tests according to the EF: `RewardsDirBase = SszTestsDir/const_preset/"phase0"/"rewards"`

The real issue though is that they're not just load state, apply single function call to state, measure difference at end.

tests/consensus_spec/phase0/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/2998/commits/5389e96db4a5eef174f2d5fbe22ed55daa8084e2 does this for phase0 and Altair. Merge is set up for it, but still does the execution payload checking separately, pending another PR, so it'd fail that check.

tersec: One next-most-automated approach is do something like what the SSZ generic type checker does -- iterate through the directory and dispatch from the directory name to the correct check.

arnetheduck: nice .. yeah, I thought about another way to do it as well: when the test runs, it registers itself with some "processed this folder" registry which is checked by `all_tests` - not exactly pretty but also does the job..

arnetheduck:
testnet actually finalized (but maybe crashed at the end?)

tests/consensus_spec/phase0/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3003 gets the merge tests

PR #3007, opened 2021-10-20
fix compile error due to cyrillic T

etan-status:
There was a cyrillic T in some big-endian specific code that broke the
compilation on such platforms. This replaces that T with an ASCII T to
fix the build.


stefantalpalaru:
There's a copy of this code in nim-eth: https://github.com/status-im/nim-eth/blob/9a1bb5e1257bfee1ca01218b69fcd6b93080b7d7/eth/ssz/merkleization.nim#L472-L474

etan-status:
Thanks. Will backport this fix and also the others.

stefantalpalaru:
If we need the same exact file/functionality in nim-eth, can we use it directly from there and delete the copies here?

etan-status:
It's sadly not an exact copy.

PR #3008, opened 2021-10-20
fix big-endian merkleization for UintN arrays

etan-status:
UintN arrays were incorrectly merkleized on big-endian. This was fixed
by making sure to use the correct buffer to store the final chunk.


etan-status:
Note: Other types are maybe also incorrectly serialized on big-endian, i.e.,
C bit fields are usually packed in big-endian order on those platforms,
e.g., bit 0 is the MSB instead of the LSB.

PR #3011, opened 2021-10-20
clean up sync subcommittee handling

arnetheduck:
* `SyncCommitteeIndex` -> `SyncSubcommitteeIndex`
* `syncCommitteePeriod` -> `sync_committee_period` (spec spelling)
* tighten period comparisons
* fix assert when validating committee message with non-altair state in
REST api

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -206,7 +206,7 @@ proc sendAttestation*(
 
 proc sendSyncCommitteeMessage*(
     node: BeaconNode, msg: SyncCommitteeMessage,
-    committeeIdx: SyncCommitteeIndex,
+    committeeIdx: SyncSubcommitteeIndex,
     checkSignature: bool): Future[SendResult] {.async.} =
   # Validate sync committee message before sending it via gossip
   # validation will also register the message with the sync committee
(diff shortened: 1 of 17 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -228,43 +228,36 @@ proc sendSyncCommitteeMessage*(
 proc sendSyncCommitteeMessages*(node: BeaconNode,
                                 msgs: seq[SyncCommitteeMessage]
                                ): Future[seq[SendResult]] {.async.} =
-  let validators = getStateField(node.dag.headState.data, validators)
-  var statuses = newSeq[Option[SendResult]](len(msgs))
+  return withState(node.dag.headState.data):
```
zah, line 231: Switching from `getStateField` to `withState` doesn't change the number of lines in the Nim code, but it changes significantly the size of the generated C code (due to the fact that all code within the block is duplicated once per fork). At some point, we'll probably start addressing this code bloat problem.

arnetheduck: sure: 
```
-rw-r--r--. 1 arnetheduck arnetheduck  185M 20 okt 18.59 @mnetworking@snetwork_metadata.nim.c
-rw-r--r--. 1 arnetheduck arnetheduck  5,5M 20 okt 18.59 @mnetworking@seth2_network.nim.c
-rw-r--r--. 1 arnetheduck arnetheduck  3,2M 20 okt 18.59 stdlib_tables.nim.c
-rw-r--r--. 1 arnetheduck arnetheduck  2,9M 20 okt 18.59 @m..@svendor@snim-json-serialization@sjson_serialization@swriter.nim.c
-rw-r--r--. 1 arnetheduck arnetheduck  2,4M 20 okt 18.59 @m..@svendor@snim-chronos@schronos@sasyncloop.nim.c
```

arnetheduck: (for the record, the withState change is there to make it reasonable to not copy the entire validator set to a local variable)

PR #3016, opened 2021-10-21
homogenize merkleization code

etan-status:
There were various implementation styles for merkleization of different
data types. This patch homogenizes them for improved maintainability.

### DIFF
--- a/beacon_chain/ssz/merkleization.nim
+++ b/beacon_chain/ssz/merkleization.nim
@@ -348,9 +348,11 @@ template createMerkleizer*(totalElements: static Limit): SszMerkleizerImpl =
   const treeHeight = binaryTreeHeight totalElements
   var combinedChunks {.noInit.}: array[treeHeight, Eth2Digest]
 
+  let topIndex = treeHeight - 1
+
   SszMerkleizerImpl(
     combinedChunks: cast[ptr UncheckedArray[Eth2Digest]](addr combinedChunks),
-    topIndex: treeHeight - 1,
+    topIndex: if (topIndex < 0): 0 else: topIndex,
     totalChunks: 0)
 
 func getFinalHash*(merkleizer: SszMerkleizerImpl): Eth2Digest =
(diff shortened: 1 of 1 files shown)


etan-status:
I think that there is still more homogenization potential inside `merkleization.nim`. Marking as draft for now so they can be merged in one go.

beacon_chain/ssz/merkleization.nim
```diff
@@ -397,83 +399,87 @@ func mixInLength*(root: Eth2Digest, length: int): Eth2Digest =
...
-template merkleizeFields(totalElements: static Limit, body: untyped): Eth2Digest =
-  var merkleizer {.inject.} = createMerkleizer(totalElements)
+template merkleizeFields(totalChunks: static Limit, body: untyped): Eth2Digest =
+  var merkleizer {.inject.} = createMerkleizer(totalChunks)
 
   template addField(field) =
     let hash = hash_tree_root(field)
     trs "MERKLEIZING FIELD ", astToStr(field), " = ", hash
-    addChunk(merkleizer, hash.data)
+    merkleizer.addChunk hash.data
```
zah, line 410: I tend to avoid using the dot syntax in heavily generic code on purpose. Due to some Nim internal implementation details, the dot syntax is not processed in the same way as the `foo(bar)` syntax (and the relevant compiler code paths are a bit more buggy).

zah (commented):
Besides the cosmetic changes, what are the important fixes in this PR?

The `topIndex: if (topIndex < 0): 0 else: topIndex` would be an example. Where is this case relevant?

etan-status:
The `topIndex: if topIndex < 0): 0 else: topIndex` I took from the `nim-eth` implementation (it was already there). In `nim-eth` the case is needed for one of the tests I think.

The goal of the PR is to align the two implementations, the one in `nimbus-eth2` and the one in `nim-eth`, and also to align the cosmetics of them, so that it is easier to maintain them.

beacon_chain/ssz/merkleization.nim
etan-status: Hmm, @arnetheduck also mentioned the same thing on an earlier iteration for the exact same location at this template. So this is not about getting it to compile, but instead even buggy compilation?! 

Reverting this one for the template. Further down, there are functions that mix the different styles within itself, I guess at least those can be homogenized.

etan-status:
As of https://github.com/status-im/nimbus-eth2/pull/3021, the relevant code has been moved to a different repo. Will create a followup there and break down the changes in a more detailed way.

kdeme:
> The `topIndex: if (topIndex < 0): 0 else: topIndex` would be an example. Where is this case relevant?

I've added this fix also in the nim-ssz-serialization repo when adding `test_merkleization.nim`: https://github.com/status-im/nim-ssz-serialization/commit/63ab1b26e818a4c1f37bdd70c6a9be5eab9c57ac 

It is an issue that hits this test: https://github.com/status-im/nim-ssz-serialization/commit/63ab1b26e818a4c1f37bdd70c6a9be5eab9c57ac#diff-20a6454734e6272cb9eaef3ce9f7727cd812132c90dce783dbbd89f6f41daa94R39

beacon_chain/ssz/merkleization.nim
zah: The most common bug manifests as "unknown identifier" compilation error which some people resolve inappropriately by adding a `mixin` statement for the said symbol/identifier (this works, but it is a bit wrong because it exposes the code to a risk of picking up a wrong definition from the module using the generic). And not using the dot syntax avoids the wasted developer time from encountering the bug in the first place.

PR #3021, opened 2021-10-21
Use nim-ssz-serialization module and rm local ssz code

kdeme:


tests/test_bitseqs.nim
```diff
@@ -3,7 +3,7 @@
 import
```
kdeme: Only realized the existence of this test now, I guess I can also move it to the module

PR #3023, opened 2021-10-22
Fix a Spec Inconsistency with the `getStateValidators` REST Route

jclapis:
This fixes a small bug with the `/eth/v1/beacon/states/head/validators?id=0x...` REST route, described [here](https://discord.com/channels/613988663034118151/812268053718237196/900853477167759400).

When you pass a validator pubkey in the `id` parameter, but that validator doesn't exist yet, `keysToIndices()` will ignore it and return an empty set. This results in Nimbus returning all of the validators as though you had called the route without any `id` parameter at all.

This is inconsistent with [the official spec](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidators), which states the following:
> If an index or public key does not match any known validator, no information will be returned but this will not cause an error.

Typically the returned JSON should look like this:

```
{"data":[]}
``` 

This fix simply introduces a check in the route handler to see if both the number of included validator IDs **and** the number of indices are both zero before returning all of the validators. If the number of indices is zero but the number of validators is not, then it knows it hit a missing validator and returns nothing as expected.

Please confirm that the included unit test for this works as expected; I believe it checks for the correct behavior, but I'm not very familiar with the REST testing system so I would like some extra eyes on it before merging this.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -287,22 +287,23 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         block:
           var res: seq[RestValidator]
           if len(indices) == 0:
-            # There is no indices, so we going to filter all the validators.
-            for index, validator in getStateField(stateData.data,
-                                                  validators).pairs():
-              let
-                balance = getStateField(stateData.data, balances)[index]
-                status =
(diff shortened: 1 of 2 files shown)


cheatfate:
@jclapis thank you for your work, but this PR is duplicate of https://github.com/status-im/nimbus-eth2/pull/3022

PR #3025, opened 2021-10-23
avoid too-fast topic attestation subnet unsubscription/subscription

tersec:


beacon_chain/validators/action_tracker.nim
```diff
@@ -61,7 +61,12 @@ type
       ## duty, we'll subscribe to the corresponding subnet to collect
       ## attestations for the aggregate
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.2/specs/phase0/validator.md#phase-0-attestation-subnet-stability
+    pruneBackoffSlots*: uint64 ##\
+      ## libp2p doesn't support unsubscribing from and resubscribing to topics
```
arnetheduck:
```suggestion
      ## when resubscribing too quickly, the gossipsub pruneBackoff prevents forming a mesh with peers that were unsubscribed from
```

arnetheduck:
@Menduist does the prune backoff also apply to peers that we don't prune? Basically, when we receive an unsubscribe message, we should be removing the peers from the mesh implicitly, no? Another way to put this: is it required to send a prune message before unsubscribing?

Menduist:
There is no implicit prune:

> The application can invoke LEAVE(topic) to unsubscribe from a topic. The router will inform the peers in mesh[topic] by sending them a PRUNE control message

arnetheduck:
And yet when they unsubscribe, we remove them from the mesh without thinking about backoff: https://github.com/status-im/nim-libp2p/blob/master/libp2p/protocols/pubsub/gossipsub.nim#L238

Menduist:
AFAICT, if they pruned us, we'll handle the backoff correctly, even if we receive the prune after the unsubscribe

So if they unsub without pruning (which is against the spec), we'll still clean them up, without backoff
If they unsub with pruning, we'll handle their backoff correctly no matter what

arnetheduck:
>  If they unsub with pruning, we'll handle their backoff correctly no matter what

well, this is a problem _somewhere_: either in the spec or in the implementation - it's basically a way to circumvent backoff which may be a good thing or a bad thing, depending on how you view it.. it's something to a) check in the major impls and b) raise in the libp2p spec repo, if it's not crystal clear from the spec

arnetheduck:
we have a similar thing in our code where we allow peers to enter our mesh without being subscribed - I think that's wrong, but likely a leftover from  some attempt to deal with out-of-order messages - it's something that should never happen regardless, ie the sequence should always be `subscribe -> graft -> prune -> unsbuscribe`, with the `prune` being debatable - `graft` before `subscribe` however sounds like something that could be exploited / lead to weird meshes

Menduist:
Checked quickly, we seem on par with go-libp2p

Opened https://github.com/status-im/nim-libp2p/issues/641 to follow this up more thoroughly

tersec:
It's looking like this isn't the best way of accomplishing this goal. https://github.com/status-im/nim-libp2p/pull/665 seems like the better approach.

PR #3028, opened 2021-10-26
fix CI for contributions from forked repos

etan-status:
This updates the publish-unit-test-result integration with support for
pull requests coming from forked repositories. The update changes the
behaviour so that (a) unit test results are displayed as a comment, and
(b) macOS and Windows builds no longer time out during the CI action.
The official guide for the publish-unit-test-result action has been
followed for this purpose. See the official readme at:
https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.23/README.md#support-fork-repositories-and-dependabot-branches

### DIFF
--- a/.github/workflows/unit-test-results.yml
+++ b/.github/workflows/unit-test-results.yml
@@ -1,0 +1,38 @@
+# https://github.com/EnricoMi/publish-unit-test-result-action
+name: Unit Test Results
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+
+jobs:
+  unit-test-results:
+    name: Unit Test Results
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion != 'skipped'
+
+    steps:
+      - name: Download and Extract Artifacts
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        run: |
+           mkdir -p artifacts && cd artifacts
+
+           artifacts_url=${{ github.event.workflow_run.artifacts_url }}
+
+           gh api "$artifacts_url" -q '.artifacts[] | [.name, .archive_download_url] | @tsv' | while read artifact
+           do
+             IFS=$'\t' read name url <<< "$artifact"
+             gh api $url > "$name.zip"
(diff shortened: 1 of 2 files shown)


.github/workflows/unit-test-results.yml
```diff
@@ -0,0 +1,38 @@
+# https://github.com/EnricoMi/publish-unit-test-result-action
+name: Unit Test Results
+
+on:
+  workflow_run:
```
etan-status, line 5: `workflow_run` triggers are only evaluated on the repo's default branch, meaning that this part is needed on `stable`.

See https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run

etan-status:
The problem was that the unit test results could not be posted into the original repo as the `pull_request` action lacks the required privileges for that. Instead, the results are now stored together with metadata in the context of the forked repo, and then a privileged workflow is run as a followup that collects the metadata and test results and posts the comment with the results.

The reason why the checks for macOS and Windows suddenly turned red is that the GH action itself had a new release posted, and that we do not version lock the version of the GH action (`@v1.22` has green checkmark, `@v1.23` started timing out during upload, but we use `@v1` which tracks the most recent one). The problem of results not being posted also existed before, but was hidden. This now properly fixes it, and once the new workflow is on the default branch the comment posting should work in a similar way as for non-forked PRs.

.github/workflows/unit-test-results.yml
stefantalpalaru: I guess we'll know it works when those robo-comments come back.

stefantalpalaru: Don't we need the first part in "unstable" for it to work in PRs that target unstable?

etan-status: Yes, `stable` will need to be merged into `unstable` after this PR for it to work. Existing PRs still using the old `ci.yml` will continue to use the previous behaviour (maybe triggering an extra update due to the new workflow running, but it should be compatible).

etan-status: Done the merges from `stable` to `testing` and from `stable` to `unstable`. This was the only change.

PR #3029, opened 2021-10-26
Gossip sub attestation subnet scoring

Menduist:
Completely reworked scoring in gossipsub

First goal was to also score peers in the attestation subnet.
Previously, we only scored peers in global topics which was an issue when running with many peers, since we would only have a few of them in global topics, and the majority of peers wouldn't have a score, which would break, amongst other things, the opportunistic grafting.

The scoring system in gossipsub 1.1 is heavily convoluted, I tried my best to explain it in layman terms, and add helpers functions which allows to input "understable" data to generate the params.
I've not tweaked this yet, so WIP


beacon_chain/networking/eth2_network.nim
```diff
@@ -1920,13 +1920,208 @@ proc announcedENR*(node: Eth2Node): enr.Record =
...
+        timeInMeshQuantum = chronos.seconds(15) # Stable topic
+      )
+    of SubnetTopic:
+      getTopicParams(
+        topicWeight = 0.8 / ATTESTATION_SUBNET_COUNT,
+        heartbeatPeriod = heartbeatPeriod,
+        period = slotPeriod,
+        averageOverNPeriods = ATTESTATION_SUBNET_COUNT, # Smooth out empty committees
+        peersPerTopic = peersPerTopic,
+        expectedMessagesPerPeriod = TARGET_COMMITTEE_SIZE.int, #TODO use current number
```
Menduist, line 2338: If someone has a ~1-liner to get the current average committee size (or number of active validators) from `Eth2Node`, I'll take it

arnetheduck: we don't have it in here (network doesn't know about dag), but it changes only once per epoch so `onSlotEnd` could update it by grabbing an `getEpochRef` for `(head, wallslot)` similar to how it sets up attestation subnets

beacon_chain/networking/eth2_network.nim
```diff
@@ -1920,13 +1920,208 @@ proc announcedENR*(node: Eth2Node): enr.Record =
...
+# using the topicWeight of each topic.
+#
+# Nimbus strategy:
+# Trying to get a 100 possible points in each topic before weighting
+# And then, weight each topic to have 100 points max total
+#
+# In order of priority:
+# - A badly behaving peer (eg, sending bad messages) will be heavily sanctionned
+# - A good peer will gain good scores
+# - Inactive/slow peers will be mildly sanctionned.
```
```suggestion
# - Inactive/slow peers will be mildly sanctioned.
```

arnetheduck:
Before we merge this, we should likely do 2 things:

* add an option in nimbus to enable/disable it
* add metrics in libp2p to monitor the outcome of the parameter settings

beacon_chain/networking/eth2_network.nim
arnetheduck: more to the point - we might need to update the scoring while nimbus is running - for example during sync, we'll go from 20k "known" validators to 250k

beacon_chain/networking/eth2_network.nim
```diff
@@ -1920,13 +1920,208 @@ proc announcedENR*(node: Eth2Node): enr.Record =
...
+      TopicParams.init()
+
+static:
+  for topicType in ord(low(TopicType))..ord(high(TopicType)):
+    getTopicParams(
+        heartbeatPeriod = chronos.milliseconds(700),
+        slotPeriod = chronos.seconds(12),
+        peersPerTopic = 8,
+        TopicType(topicType)
+      ).validateParameters().tryGet()
```
arnetheduck, line 2351: `validateParameters` should probably be replaced by a `TopicParameters.init` that returns a `Result`

beacon_chain/networking/eth2_network.nim
```diff
@@ -1920,13 +1920,208 @@ proc announcedENR*(node: Eth2Node): enr.Record =
...
+    # A peer being first in 1/d% messages will reach a score of
+    # `shouldSendOverNPeriod`
+    firstMessageDecay =
+      computeDecay(
+        startValue = shouldBeFirstOverNPeriod,
+        endValue = 0.1,
+        timeToEndValue = period * averageOverNPeriods.int * 2,
+        heartbeatPeriod)
+ 
+    # Start to remove up to 30 points when peer send less
```
arnetheduck: this can bring a peer into negative score territory?

beacon_chain/networking/eth2_network.nim
Menduist: Ah right, didn't think about sync.
But we only subscribe to gossip once synced, no?

My intuition was that the number grows/reduce slowly enough to avoid updating it OTF, but we can do it if needed

arnetheduck: we shouldn't assume really that users restart nimbus regularly (they shouldn't have to) - ie there are still users that haven't upgraded since genesis

beacon_chain/networking/eth2_network.nim
Menduist: I'm still tweaking this, but my intuition is that it should be
FirstMessageDeliveriesBonus > InactivePenalty > TimeInMeshBonus

Currently, it's 80 > -30 > 20
A perfect peer would have 100 per topic. If it suddenly becomes inactive, he would gradually (in averageOverNPeriods) drop to -10 (InactivePenalty + TimeInMeshBonus), so it would be negative, meaning you can't grind for points, then become inactive.

So yeah, being inactive could bring to negative score, if you don't do good things aside. Now, what we do when they are into negative score is a different question. As the moment, this PR is fairly conservative on actions (to reach -40 you would need to be inactive on a lot of topics, or more probably, send a lot of invalid messages), since the primary goal as of now is to restore the opportunistic grafting, which helps a lot with inactive peers.

ATM i'm not very happy on the topicWeight computation, need to rework that

Menduist:
I've been running this for a while, and it seem to give reasonable scores.

Only weird thing I saw, we often give invalid message penalties to peer on `beacon_aggregate_and_proof`, which could indicate a incompatibility somewhere (or malicious peer, maybe)

> add an option in nimbus to enable/disable it

I'm not sure we need it in this case, because the impact is very minimal (we almost never kick peers because of it, we just prune inactive peers for a while)

I'll continue to monitor it, add the dynamic validator count, and should be GTG.
Scoring metrics in libp2p indeed need to be improved, I've added it to https://github.com/status-im/nim-libp2p/issues/633

tersec:
Superceded by:
- https://github.com/status-im/nimbus-eth2/pull/6979

PR #2691, opened 2021-07-01
.deb + .rpm via FPM

bernardoaraujor:
# Proposal

By using [FPM](https://fpm.readthedocs.io/en/latest/intro.html) and a shell script (`create-packages.sh`), we propose the automated creation of `nimbus_X.Y.Z_amd64.deb` and `nimbus_X.Y.Z_amd64.rpm` during release.

The script can be triggered by the [`release`](https://github.com/status-im/nimbus-eth2/blob/stable/.github/workflows/release.yml)  Workflow (not yet implemented).

# Considerations

Given the fact that we don't assume any `Nim` instalation  on the target system as a runtime pre-requisite, the package footprint differs from the current Tarball distribution.

Therefore, the underlying logic of the `run*.sh` scripts had to be adapted.

Here's a layout of the package image:
```
├── etc
│   └── default
│       └── nimbus.env
├── usr
│   └── bin
│   │   ├── nimbus_beacon_node
│   │   ├── nimbus_cli
│   │   ├── nimbus_cli_db
│   │   ├── nimbus_inspector
│   │   ├── nimbus_launch
│   │   ├── nimbus_signing_process
│   │   └── nimbus_validator_client
│   └── lib
│       └── systemd
│           └── system
│               └── nimbus.service
└── var
    └── lib
        └── nimbus
```

where:
- `/usr/lib/systemd/system/nimbus.service` holds the SystemD Service Unit definition.
- `/etc/default/nimbus.env` holds environment variables for the SystemD Service.
- `/var/lib/nimbus` holds the runtime data of the SystemD Service (`WorkingDir`).
- `/usr/bin/nimbus_launch` holds logic for launching `Nimbus` with proper `NETWORK` and `WEB3_URL` environment variables (inspired by `scripts/run-beacon-node.sh`).

We should also avoid making assumptions in regards to executable naming under the target runtime.
Therefore, some executables from `build/` were renamed with a `nimbus_*` prefix during population of `/usr/bin`.

# Testing

- `.deb` was manually tested under Ubuntu 20.04
- `.rpm` was manually tested under Fedora 33

# Limitations

- `/usr/bin/nimbus_launch` does not check for the validity of `WEB3_URL`, which could potentially lead to crashes with the executable.

- Testing covered for runtime execution, both via SystemD and manual. However, no assertions over the Ethereum protocol were made.

- Github Actions trigger not yet implemented. In case we get a green light on the PR draft, I'll send another commit calling `create-packages.sh` from `.github/workflows/release.yml`

PR #3034, opened 2021-10-27
RPM/DEB for systemd-based systems

arnetheduck:
Similar to https://github.com/status-im/nimbus-eth2/pull/2691, this PR
introduces packaging for nimbus beacon node.

Instead of relying on interactive prompts, a default configuration is
included that connects to mainnet and expects there to be an execution
client running with websockets enabled on the same host, on the standard
websocket port.

Should the user need to configure their nimbus after installation (for
example with a different web3 url), the "standard" way of doing so via
`systemctl` edit is recommended.


scripts/make_packages.sh
```diff
@@ -0,0 +1,78 @@
...
+PKG_NAME=nimbus_beacon_node
+PKG_ARCH=amd64
+PKG_IMG_DIR=package_image
+
+NBC_BIN=../build/nimbus_beacon_node
+NSP_BIN=../build/nimbus_signing_process
+
+if [ ! -f $NBC_BIN -o ! -f $NSP_BIN ]; then
+  printf "Please build nimbus_beacon_node and nimbus_signing_process\n"
+  printf "This script needs to be run from the scripts folder\n"
```
stefantalpalaru: You already guarantee that on line 10.

scripts/make_packages.sh
```diff
@@ -0,0 +1,78 @@
...
+
+  exit 1
+fi
+
+PKG_VERSION=$(./$NBC_BIN --version | awk -F[-v] 'NR==1{print $2}')
```
stefantalpalaru: You probably want "gawk" here, to avoid dealing with weird macOS versions of "awk".

On the other hand, we'll probably only run this script from Linux.

stefantalpalaru:
I think that "prerm" and "postrm" also run during upgrades, which makes them too invasive.

When I upgrade a package, I don't want the corresponding service stopped. I only need a message reminding me to restart it by hand afterwards.

I also don't need the user deleted and recreated - possibly with a different UID - because I may have files owned by that user that I want to keep between upgrades (like the data dir).

arnetheduck:
> I think that "prerm" and "postrm" also run during upgrades, which makes them too invasive.

don't think so, given that there's a `--after-upgrade` option as well in fpm: https://fpm.readthedocs.io/en/latest/cli-reference.html

stefantalpalaru:
Upgrades are usually implemented as remove + install. No special upgrade logic needed.

See also https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html:

"When a package is upgraded a combination of the scripts from the old and new packages is called during the upgrade procedure. If your scripts are going to be at all complicated you need to be aware of this, and may need to check the arguments to your scripts.

Broadly speaking the preinst is called before (a particular version of) a package is unpacked, and the postinst afterwards; the prerm before (a version of) a package is removed and the postrm afterwards."

https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called is even clearer.

arnetheduck:
ok - what do you propose instead?

stefantalpalaru:
Look at how Debian developers do it, for some similar package. Could be something like:

- create user if it doesn't already exists (there might be distro-specific helper scripts for that)
- don't ever delete that user
- don't automatically start/stop services

arnetheduck:
> Upgrades are usually implemented as remove + install. No special upgrade logic needed.

https://github.com/jordansissel/fpm/blob/master/templates/deb/postrm_upgrade.sh.erb

fpm deals with the differences

arnetheduck:
nonetheless - removed data purging / user removal and moved it to a specific debian "purge" command

I suspect there's no great standard here, but regardless, removing the whole directory seems a bit aggressive, specially since there might be stuff in there.

scripts/make_packages.sh
arnetheduck: can't run on mac since the binaries will be wrong - this is a bridge to cross another day

scripts/package_image/after_install
```diff
@@ -0,0 +1,22 @@
...
+
+set -e
+
+DISTRO=$(lsb_release -si)
+if ! id -u nimbus > /dev/null 2>&1; then
+  case $DISTRO in
+    Ubuntu|Debian)
+      adduser --system --no-create-home --group nimbus
+      ;;
+    Fedora)
```
zah: We should also add at least CentOS here.

Here are some stats for the popularity of various Linux distros on servers:

https://w3techs.com/technologies/details/os-linux
https://app.vagrantup.com/boxes/search (sort by downloads)

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -0,0 +1,61 @@
...
+[Unit]
+Description=Nimbus Beacon Node (Ethereum consensus client)
+Wants=network-online.target
+After=network-online.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+User=nimbus
```
zah: I can see that the `Group` option is missing. Does `systemd` set it to match the `User` option by default?

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -0,0 +1,61 @@
...
+Environment=DATA_DIR_PREFIX=/var/lib/nimbus
+
+# Default ports
+Environment=TCP_PORT=9000
+Environment=UDP_PORT=9000
+Environment=REST_PORT=5052
+Environment=METRICS_PORT=8008
+
+# Interaction and monitoring
+Environment=REST_ENABLED=1
```
zah: REST_ENABLED=`yes` may be more accessible for the non-techies out there.

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -0,0 +1,61 @@
...
+[Service]
+User=nimbus
+WorkingDirectory=/var/lib/nimbus
+TimeoutSec=1200
+Restart=always
+
+ExecStart=/usr/bin/nimbus_beacon_node \
+  --network=${NETWORK} \
+  --data-dir="${DATA_DIR_PREFIX}/shared_${NETWORK}_${NODE_ID}" \
+  --tcp-port=${TCP_PORT} \
```
zah: `NODE_ID` has to be used here as well, otherwise you'll get port conflicts.

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -0,0 +1,61 @@
...
+
+# Network - use `prater` to connect to the prater testnet
+Environment=NETWORK=mainnet
+
+# You need to have have access to an execution client - by default, we assume
+# a compatible executionc client is running on the same machine on the default
+# websocket port.
+Environment=WEB3_URL=ws://127.0.0.1:8546
+
+# Support for running multiple nodes side-by-side
```
zah: We can clarify here that only numeric values are accepted (otherwise, the port arithmetic won't work).

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: hm, can we actually do arithmetic in the shell, here?

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: https://www.freedesktop.org/software/systemd/man/systemd.exec.html:

> If no group is set, the default group of the user is used.

scripts/package_image/after_install
arnetheduck: that needs more magic and testing than the current thing I pulled from the other PR - apparently `lsb_release` isn't available on centos by default: https://cloudcone.com/docs/article/how-to-fix-lsb_release-command-not-found-in-centos-8/ - cc @stefantalpalaru any ideas?

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: cc @stefantalpalaru

jakubgs: You can't unless you wrap the thing in a `bash -c`.

stefantalpalaru: Jakub is right: https://unix.stackexchange.com/questions/216780/why-does-bash-parameter-expansion-not-work-inside-systemd-service-files

And it's because Systemd does not use a POSIX shell to parse that service file. It has its own, crippled parser: https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines

arnetheduck: alright - all that escaping sounds way too clever/messy - removing `NODE_ID` from these files seems to be the way to go - much more clean

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: 33dbf698fed0c63ac7b099c44bfc027db7ee7977

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: 33dbf698fed0c63ac7b099c44bfc027db7ee7977

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: 33dbf698fed0c63ac7b099c44bfc027db7ee7977

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: No arithmetic available in systemd, removed in 33dbf698fed0c63ac7b099c44bfc027db7ee7977

tersec (approved):

scripts/package_image/usr/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -0,0 +1,60 @@
+# To configure the service, use `systemctl edit nimbus_beacon_node.service`
+# and override the environment variables in this file:
+# [Service]
+# Environment="WEB3_URL=wss://provider/"
+#
+# To completely override the start command (to add custom parameters such as
+# graffitit), override the `ExecStart` value instead:
```
arnetheduck:
```suggestion
# graffitti), override the `ExecStart` value instead:
```

scripts/package_image/after_install
stefantalpalaru: Solved it with a dependency on "redhat-lsb-core" and added more distro IDs in there.

scripts/make_packages.sh
```diff
@@ -7,73 +7,143 @@
...
+  *)
+    echo "unsupported tarball type"
+    exit 1
+    ;;
+esac
+
+SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
+PKG_NAME="nimbus_beacon_node"
+PKG_IMG_DIR="${SCRIPT_DIR}/package_image"
+BINARIES="nimbus_beacon_node nimbus_signing_process"
```
zah: Needs to be updated - `nimbus_signing_process` is no longer present

scripts/make_packages.sh
```diff
@@ -0,0 +1,149 @@
...
```
zah:
```suggestion
BINARIES="nimbus_beacon_node"
```

PR #3038, opened 2021-10-28
Logging and startup improvements

### LINKED ISSUE #1794: Disable stdout logging
When running the node as a systemd service with json logging enabled, the stdout logging uses valuable cycles and is unnecessary - add a way to disable it.

### LINKED ISSUE #2326: `--log-file` option and colors missing from binary builds
v1.0.7, linux

```
[arnetheduck@tempus build]$ ./nimbus_beacon_node --log-file=xx.log
WRN 2021-02-16 09:17:12.348+01:00 The --log-file option is not active in the current build tid=215907 file=nimbus_binary_common.nim:62
```

looks like these were built as a server version?

arnetheduck:
Color support for released binaries!

* startup scripts no longer log to file by default - this only affects
source builds - released binaries didn't support file logging
* add --log-stdout option to control logging to stdout (colors, json)
* detect tty:s vs redirected logs and log accordingly
* add option to disable log colors at runtime
* simplify several "common" logs, showing the most important information
earlier and more clearly
* remove line numbers / file information / tid - these take up space and
are of little use to end users
  * still enabled in debug builds and tools
* remove `testnet_servers_image` compile-time option
* server images, released binaries and compile-from-source now offer
the same behaviour and features
* fixes #2326
* fixes #1794
* remove instanteneous block speed from sync message, keeping only
average

before:

```
INF 2021-10-28 16:45:59.000+02:00 Slot start                                 topics="beacnde" tid=386429 file=nimbus_beacon_node.nim:884 lastSlot=2384027 wallSlot=2384028 delay=461us84ns peers=0 head=75a10ee5:3348 headEpoch=104 finalized=cd6804ba:3264 finalizedEpoch=102 sync="wwwwwwwwww:0:0.0000:0.0000:00h00m (3348)"
INF 2021-10-28 16:45:59.046+02:00 Slot end                                   topics="beacnde" tid=386429 file=nimbus_beacon_node.nim:821 slot=2384028 nextSlot=2384029 head=75a10ee5:3348 headEpoch=104 finalizedHead=cd6804ba:3264 finalizedEpoch=102 nextAttestationSlot=-1 nextProposalSlot=-1 nextActionWait=n/a
```

after:

```
INF 2021-10-28 22:43:23.033+02:00 Slot start                                 topics="beacnde" slot=2385815 epoch=74556 sync="DDPDDPUDDD:10:5.2258:01h19m (2361088)" peers=37 head=eacd2dae:2361096 finalized=73782:a4751487 delay=33ms687us715ns
INF 2021-10-28 22:43:23.291+02:00 Slot end                                   topics="beacnde" slot=2385815 nextActionWait=n/a nextAttestationSlot=-1 nextProposalSlot=-1 head=eacd2dae:2361096
```


stefantalpalaru:
> released binaries will no longer write log files by default

They weren't: https://github.com/status-im/nimbus-eth2/blob/fdbfe19b3ea44646363eff3ad98ad43baa35e8cb/docker/dist/entry_point.sh#L23

arnetheduck:
ok - that's good actually, less disruptive - what about docker users?

arnetheduck:
phrased another way: where was `testnet_servers_image` de facto used?

stefantalpalaru:
> what about docker users?

They use the same binaries produced by `make dist-*`.

> where was testnet_servers_image de facto used?

Only on the fleet and for local testnets, as far as I know.

BTW, this is where fleet binaries are being built: https://github.com/status-im/infra-role-beacon-node-linux/blob/b9d78e430585ebf8dd9bae9885975308c18e744a/templates/build.sh.j2#L24-L25

(plus the similar repos for macOS and Windows)

arnetheduck:
hm, that's odd because there's been several reports of spurious log files, hm

stefantalpalaru:
It depends how Docker is set up to store stdout output (somewhere on the filesystem, by default) or how the Systemd service deals with stdout (usually logs it). This is all documented in https://nimbus.guide/binaries.html and https://nimbus.guide/docker.html

But, since the beginning, I've made sure that official binaries only write their logging to stdout, to avoid having to deal with a duplicated log file that defaulted to the working directory.

arnetheduck:
> They weren't: 

ok - so now we can do the exact same binary for everyone - updated dist as well to include all options.

beacon_chain/nimbus_beacon_node.nim.cfg
```diff
@@ -1,10 +1,10 @@
--d:"chronicles_sinks=textlines,json[file]"
+-d:"chronicles_sinks=textlines[dynamic],json[dynamic]"
 -d:"chronicles_runtime_filtering=on"
--d:"chronicles_default_output_device=dynamic"
+-d:"chronicles_disable_thread_id"
+
+@if release:
+  -d:"chronicles_line_numbers:0"
```
stefantalpalaru, line 6: Can this be overridden by NIMFLAGS? Nobody actually needs a slow `-d:debug` build while developing on Linux.

arnetheduck: yes, command line has precedence over cfg

beacon_chain/nimbus_binary_common.nim
```diff
@@ -56,27 +82,90 @@ proc updateLogLevel*(logLevel: string) {.raises: [Defect, ValueError].} =
...
+    let tmp =
+      if stdoutKind == StdoutLogKind.Auto:
+        if logFile.isSome():
+          # No stdout logging by default, when file logging is enabled
+          StdoutLogKind.None
+        elif isatty(stdout):
+          # On a TTY, let's be fancy
+          StdoutLogKind.Colors
+        else:
+          # When there's no TTY, we output colors because this matches what
```
zah: Did you meant to write "we output no colors"? I assume that's what `StdoutLogKind.NoColors` does below

config.nims
```diff
@@ -71,7 +71,7 @@ switch("passL", "-fno-omit-frame-pointer")
 --excessiveStackTrace:on
 # enable metric collection
 --define:metrics
---define:chronicles_line_numbers
+--define:chronicles_line_numbers # These are disabled for release binaries
```
zah, line 74: Hrm, you are supposed to provides a boolean value here (`yes`, `no`, `on`, `off`, etc). The fact that this is accepted is probably a bug in the Chronicles options parser.

zah (approved):

arnetheduck:
Rebased this PR a bit to factor out the startup changes to a separate PR since that's a different UX change

beacon_chain/nimbus_binary_common.nim
arnetheduck: right, fixed

config.nims
arnetheduck: https://github.com/status-im/nim-chronicles/issues/105

PR #3040, opened 2021-10-29
introduce strong typing for sync committee periods

etan-status:
The sync committee period used to be a plain `uint64`. With the light
client sync relying more heavily on them, it makes sense to introduce
a proper type for them, similar to how they are already used for `Slot`
and `Epoch`. This introduces such a `SyncCommitteePeriod` type.
Furthermore, some usage code dealing with those periods is cleaned up.

### DIFF
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -721,6 +721,7 @@ func `as`*(d: DepositData, T: type DepositMessage): T =
 
 ethTimeUnit Slot
 ethTimeUnit Epoch
+ethTimeUnit SyncCommitteePeriod
 
 template newClone*[T: not ref](x: T): ref T =
   # TODO not nil in return type: https://github.com/nim-lang/Nim/issues/14146
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -9,9 +9,9 @@ import stew/[results, base10, byteutils, endians2], presto/common,
        libp2p/peerid, serialization,
        json_serialization, json_serialization/std/[options, net, sets],
        nimcrypto/utils as ncrutils
-import ".."/[eth2_ssz_serialization, forks],
-       ".."/datatypes/[phase0, altair, merge],
-       "."/rest_types
+import "."/rest_types,
+       ".."/[eth2_ssz_serialization, forks],
+       ".."/datatypes/[phase0, altair, merge]
 
 export
   results, peerid, common, serialization, json_serialization, options, net, sets,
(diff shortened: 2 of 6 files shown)


beacon_chain/spec/datatypes/base.nim
```diff
@@ -721,6 +721,7 @@ func `as`*(d: DepositData, T: type DepositMessage): T =
 
 ethTimeUnit Slot
 ethTimeUnit Epoch
+ethTimeUnit SyncCommitteePeriod
```
arnetheduck, line 724: a time unit is something that points to point in time - it's not a period of time

arnetheduck: or is this meant to be the point in time when the period starts, relative to the chain start?  if so, the naming is off

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -9,9 +9,9 @@ import stew/[results, base10, byteutils, endians2], presto/common,
        libp2p/peerid, serialization,
        json_serialization, json_serialization/std/[options, net, sets],
        nimcrypto/utils as ncrutils
-import ".."/[eth2_ssz_serialization, forks],
-       ".."/datatypes/[phase0, altair, merge],
-       "."/rest_types
+import "."/rest_types,
+       ".."/[eth2_ssz_serialization, forks],
+       ".."/datatypes/[phase0, altair, merge]
```
etan-status, line 14: Re-ordered because of random compilation issues on `i386` in the original import order.

beacon_chain/spec/datatypes/base.nim
etan-status: It is a period of 256 `Epoch`s during which the sync committee is stable, similar to how `Epoch` is a range of `Slot`.

The `ethTimeUnit` is the name of the template that borrows functions for adding / comparing two instances of them.

etan-status: The naming is consistent with the spec,

i.e.,

```EPOCHS_PER_SYNC_COMMITTEE_PERIOD```

and

```SLOTS_PER_EPOCH```

File: beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim (hunk 2 above)
Comment (arnetheduck) [reply beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim:14]: is this still needed when rebased on latest unstable?

**Review by arnetheduck — APPROVED** [review]

File: beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim (hunk 2 above)
Comment (etan-status) [reply beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim:14]: On 233d756518ea2433e1fbf94e12714eabcd9a94f3, this change is still needed.

```
/nimbus-eth2/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim(662, 22) template/generic instantiation of `decode` from here
/nimbus-eth2/vendor/nim-serialization/serialization.nim(47, 31) Error: type mismatch: got <type ReaderType`gensym110725036, InputStreamHandle, requireAllFields: bool>
but expected one of: 
proc init[A](s: var OrderedSet[A]; initialSize = defaultInitialSize)
  required type for s: var OrderedSet[init.A]
  but expression 'ReaderType`gensym110725036' is of type: type ReaderType`gensym110725036

expression: init(ReaderType`gensym110725036, stream`gensym110725035, requireAllFields = true)

```

Issue #2685, opened 2021-06-29
Option to force syncing Eth1


mratsim:
When migrating validators we prepare a node by syncing Eth2. However Nimbus only sync Eth1 when a validator is attached.

To avoid additional delay and miss a block proposal on migration we probably want a `--force-sync-eth1` flag to enforce Eth1 syncing even with no validators.

arnetheduck:
actually, the simpler option here might be to simply always sync eth1 when an eth1 address is specified, avoiding the extra option - it's cheap enough

zah:
This is already present with the parameter `--web3-mode=enabled` (the default value is `auto` which means "enable when validators are attached"). Right now, this is considered a hidden parameter, but exposing it would be a one-line change.

arnetheduck:
> This is already present with the parameter --web3-mode=enabled (the default value is auto which means "enable when validators are attached"). Right now, this is considered a hidden parameter, but exposing it would be a one-line change.

Do we really need an option for it though? We're already expressing the desire to use web3 by feeding in a web3-url or not, having one more option just creates more invalid states that the application can be in.

PR #3042, opened 2021-11-01
Remove web3-mode, always keep web3 monitor enabled when given

arnetheduck:
* Init some components fully before BeaconNode, per component dependency
graph
* remove `--web3-mode` option
* fixes https://github.com/status-im/nimbus-eth2/issues/2685
* reshuffle some beacon node init code


arnetheduck:
This splits out the web3 and startup parts from https://github.com/status-im/nimbus-eth2/pull/3038 so as to keep the other branch focused on logging and this branch focused on the web3 mode change

PR #3051, opened 2021-11-04
Bugfix: Avoid the aggregation of duplicate signatures when creating sync committee contributions

zah:
Also introduces more sanity checks that would prevent aggregating sync committee messages created against a different head state.

### DIFF
--- a/beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
+++ b/beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
@@ -131,13 +131,14 @@ func computeAggregateSig(votes: seq[TrustedSyncCommitteeMsg],
     if vote.subcommitteeIndex != subcommitteeIndex:
       continue
 
-    if not initialized:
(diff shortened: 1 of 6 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1094,6 +1094,20 @@ func getSubcommitteePositions*(dag: ChainDAGRef,
     else:
       @[]
 
+func getSyncCommitteeHash*(dag: ChainDAGRef, slot: Slot): Eth2Digest =
+  let period = sync_committee_period(slot)
+  withState(dag.headState.data):
+    when stateFork >= BeaconStateFork.Altair:
+      let curPeriod = sync_committee_period(state.data.slot)
+      if period == curPeriod:
+        hashTreeRoot(state.data.current_sync_committee.pubkeys)
```
arnetheduck: hash_tree_root

PR #935, opened 2020-04-26
Use a stack size slightly smaller than 1MB

zah:
This is motivated by discussions in https://github.com/status-im/nim-beacon-chain/pull/924

The stack size on Android and iOS is 1MB, so we want to detect
any possible stack overflow issues there while we are testing
our software on the desktop. The limit is 1 million bytes instead
of exactly 1MB, so we can be on the safe side.

PR #3052, opened 2021-11-04
build: show stderr by default

stefantalpalaru:
Also added: `-d:cwarnings` and `-d:limitStackUsage`

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -13,6 +13,10 @@ switch("nimcache", nimCachePath)
 # `-flto` gives a significant improvement in processing speed, specially hash tree and state transition (basically any CPU-bound code implemented in nim)
 # With LTO enabled, optimization flags should be passed to both compiler and linker!
 if defined(release) and not defined(disableLTO):
+  # "-w" is not passed to the compiler during linking, so we need to disable
+  # some warnings by hand.
+  switch("passL", "-Wno-stringop-overflow -Wno-stringop-overread")
+
   if defined(macosx): # Clang
     switch("passC", "-flto=thin")
     switch("passL", "-flto=thin -Wl,-object_path_lto," & nimCachePath & "/lto")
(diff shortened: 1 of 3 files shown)


stefantalpalaru:
Supersedes https://github.com/status-im/nimbus-eth2/pull/935

arnetheduck:
lgtm - the next step would be to enable it in CI like the other branch does - `unstable` is now fixed with regards to stack usage so it should pass even with the option enabled

config.nims
```diff
@@ -26,6 +30,23 @@ if defined(release) and not defined(disableLTO):
...
+    put("gcc.options.always", "-mno-ms-bitfields " & common_gcc_options)
+    put("clang.options.always", "-mno-ms-bitfields " & common_gcc_options)
+  else:
+    put("gcc.options.always", common_gcc_options)
+    put("clang.options.always", common_gcc_options)
+
+if defined(limitStackUsage):
+  # This limits stack usage of each individual function to 1MB - the option is
+  # available on some GCC versions but not all - run with `-d:limitStackUsage`
+  # and look for .su files in "./nimcache/" that list the stack size of each function
```
arnetheduck, line 46:
```suggestion
  # and look for .su files in "./nimcache/" or `$TMPDIR` that list the stack size of each function
```

stefantalpalaru: How would they get in "/tmp/"?

arnetheduck: depends on the compiler version - actually, it's `nimcache/`, `build/` or `$TMPDIR` - in gcc 11, LTO writes them to the directory of the binary (since that's when they're generated)

arnetheduck: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=98922

stefantalpalaru: OK

PR #3041, opened 2021-10-31
Fork choice EF consensus tests

mratsim:
Consensus spec v1.1.1 finally added fork choice tests, this integrates them.

PR #3054, opened 2021-11-05
Support starting from altair

arnetheduck:
* hide `finalized-checkpoint-` - they are incomplete and usage may cause
crashes
* ~~remove genesis detection code (broken, obsolete)~~
* enable starting ChainDAG from altair checkpoints - this is a
prerequisite for checkpoint sync (TODO: backfill)
* tighten checkpoint state conditions
* show error when starting from checkpoint with existing database (not
supported)
* print rest-compatible JSON in ncli/state_sim
* altair/merge support in ncli
* more altair/merge support in ncli_db


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -163,76 +153,6 @@ func depositCountU64(s: DepositContractState): uint64 =
 
   uint64.fromBytesBE s.deposit_count[24..31]
 
-when hasGenesisDetection:
```
zah: Er, why is this being removed? There have been users using this mode in custom networks.

research/simutils.nim
```diff
@@ -61,31 +62,33 @@ func verifyConsensus*(state: ForkedHashedBeaconState, attesterRatio: auto) =
...
-    res = (ref phase0.HashedBeaconState)()
+
 
   if fileExists(genesisFn) and fileExists(contractSnapshotFn):
-    res.data = SSZ.loadFile(genesisFn, phase0.BeaconState)
-    res.root = hash_tree_root(res.data)
-    if res.data.slot != GENESIS_SLOT:
-      echo "Can only start from genesis state"
-      quit 1
+    let res = newClone(readSszForkedHashedBeaconState(
```
zah, line 75: The `readSszForkedHashedBeaconState` function can be implemented much more efficiently by getting the state slot through a SSZ navigator, checking the expected state type through the `RuntimeConfig` and proceeding to deserialize it.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -174,15 +176,25 @@ proc init*(T: type BeaconNode,
...
+        if getStateField(checkpointState[], slot) > 0:
+          fatal "Specifying a non-genesis --finalized-checkpoint-state requires specifying --finalized-checkpoint-block as well"
+          quit 1
     else:
       let checkpointBlockPath = config.finalizedCheckpointBlock.get.string
       try:
-        # TODO Perform sanity checks like signature and slot verification at least
-        checkpointBlock = SSZ.loadFile(checkpointBlockPath, phase0.TrustedSignedBeaconBlock)
-      except SerializationError as err:
+        checkpointBlock = readSszForkedTrustedSignedBeaconBlock(
```
zah, line 188: Don't we know the expected block fork from the checkpoint state?

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: oh, is there? didn't know it compiled even and the startup jungle was hard enough to reason about without it .. one problem is that it's not well defined for Altair - all pure altair (and merge) testnets to date have been started through an externally provided genesis state (there's no spec-supported way of starting into altair that I know).

If this functionality is to be kept, it should be a separate binary and not part of this init flow - this way, you start the binary: when execution is done, it can turn itself off and a "regular" start with a genesis.ssz can be initiated.

Do we have any concrete use cases actually?

arnetheduck: indeed, it doesn't compile, so it seems pretty hard to use it :)

arnetheduck: (and it lives on in git for any history buffs)

research/simutils.nim
arnetheduck: fixed, seemed better to use a dedicated type instead of the wrong type and dark magic. that said, it doesn't practically matter, when the wrong beaconstate type is used it's discarded at once because the sizes are off.

beacon_chain/nimbus_beacon_node.nim
arnetheduck: no, a block may be from before the given state, and from a before the given epoch (because empty slots) - commented

ncli/ncli.nim
```diff
@@ -171,14 +172,22 @@ proc doSSZ(conf: NcliConf) =
...
+  of "phase0_block_body": printit(phase0.BeaconBlockBody)
+  of "altair_block_body": printit(altair.BeaconBlockBody)
+  of "merge_block_body": printit(merge.BeaconBlockBody)
   of "block_header": printit(BeaconBlockHeader)
   of "deposit": printit(Deposit)
   of "deposit_data": printit(DepositData)
   of "eth1_data": printit(Eth1Data)
-  of "state": printit(phase0.BeaconState)
+  of "phase0_state": printit(phase0.BeaconState)
+  of "altiar_state": printit(altair.BeaconState)
```
tersec, line 189:
```suggestion
  of "altair_state": printit(altair.BeaconState)
```

beacon_chain/spec/beaconstate.nim
```diff
@@ -294,6 +294,35 @@ func get_initial_beacon_block*(state: phase0.BeaconState):
   phase0.TrustedSignedBeaconBlock(
     message: message, root: hash_tree_root(message))
 
+# https://github.com/ethereum/consensus-specs/blob/v1.1.3/specs/altair/beacon-chain.md#initialize-state-for-pure-altair-testnets-and-test-vectors
```
tersec: v1.1.4 now (as with `get_initial_beacon_block`); neither substantively changed, so it's just updating the URLs.

arnetheduck: fixed in rebase

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: readded broken code - removal or update will be dealt with separately

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1313,65 +1324,93 @@ proc updateHead*(
...
+  if not (
+      db.containsState(genesisStateRoot) and db.containsState(tailStateRoot)):
     return false
 
   true
 
 proc preInit*(
     T: type ChainDAGRef, db: BeaconChainDB,
-    genesisState, tailState: var phase0.BeaconState, tailBlock: phase0.TrustedSignedBeaconBlock) =
+    genesisState, tailState: ForkedHashedBeaconState,
```
mratsim, line 1356: This will conflict with https://github.com/status-im/nimbus-eth2/pull/3041 (I just changed the `var` there).

As https://github.com/status-im/nimbus-eth2/pull/3041 adds more tests and little else, I think it should be merged earlier.

mratsim (approved):
LGTM

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: 2021-11-08-153112_518x297_scrot

beacon_chain/nimbus_beacon_node.nim
zah: If the block is older, then the field `blck.message.state_root` will point to a different state, no?
Just few lines below you are requiring that they match.

beacon_chain/spec/forks.nim
```diff
@@ -412,60 +412,57 @@ func getForkSchedule*(cfg: RuntimeConfig): array[2, Fork] =
...
-      try:
-        readSszBytes(data, state.data)
-        state.root = hash_tree_root(state.data)
-        return result
-      except SszError as exc:
-        tried.incl result.kind
-
-  readFork()
+  # careful - `result` is used, RVO didn't seem to work without
+  # TODO move time helpers somewhere to avoid circular imports
```
zah, line 434: Circular imports are not welcome even temporarily. Can we address that before merging the PR?

arnetheduck: there are no circular imports - to avoid them, this code uses `div SLOTS_PER_EPOCH` instead of just `.epoch` which is the helper being referred to

beacon_chain/nimbus_beacon_node.nim
arnetheduck: You're right - forgot about a limitation in the chaindag startup code which reads state root from block, and not from our "usual" BlockSlot -> state root table - I've moved this check into `ChainDag.preInit` where it belongs, such that it'll be easier to follow up with a fix that relaxes this constraint.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: +1 - #3041 is not in a mergable shape right now, it has other conflicts already - can remove the `var` change from there when resolving the existing conflicts and issues

mratsim: It's mergeable now

arnetheduck: not yet :) needs stack size fixing

PR #3057, opened 2021-11-07
Correct implementation of the /validator/duties/sync/{epoch} API

zah:
According to the spec, this call should return the positions of
the specified validators within the sync committee. The existing
code was instead returning the indices of the sync sub-committees
where the validator is a member.

### DIFF
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -19,9 +19,6 @@ import
 
 export extras, forks, validator
 
-type
-  SomeBeaconState* = phase0.BeaconState | altair.BeaconState | merge.BeaconState
-
 # https://github.com/ethereum/consensus-specs/blob/v1.1.4/specs/phase0/beacon-chain.md#increase_balance
 func increase_balance*(balance: var Gwei, delta: Gwei) =
   balance += delta
@@ -859,6 +856,6 @@ func upgrade_to_merge*(cfg: RuntimeConfig, pre: altair.BeaconState):
     latest_execution_payload_header: ExecutionPayloadHeader()
   )
 
-template isValidInState*(idx: ValidatorIndex, state: SomeBeaconState): bool =
-  idx.int < state.validators.len
+template isValidInState*(idx: ValidatorIndex, state: ForkyBeaconState): bool =
+  idx.uint64 < state.validators.lenu64
 
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -19,6 +19,9 @@ import
 
 export extras, forks, validator
 
+type
+  SomeBeaconState* = phase0.BeaconState | altair.BeaconState | merge.BeaconState
+
 # https://github.com/ethereum/consensus-specs/blob/v1.1.4/specs/phase0/beacon-chain.md#increase_balance
 func increase_balance*(balance: var Gwei, delta: Gwei) =
   balance += delta
@@ -855,3 +858,7 @@ func upgrade_to_merge*(cfg: RuntimeConfig, pre: altair.BeaconState):
     # Execution-layer
     latest_execution_payload_header: ExecutionPayloadHeader()
   )
+
+template isValidInState*(idx: ValidatorIndex, state: SomeBeaconState): bool =
+  idx.int < state.validators.len
+
--- a/beacon_chain/spec/datatypes/altair.nim
+++ b/beacon_chain/spec/datatypes/altair.nim
@@ -432,7 +432,7 @@ type
   SomeBeaconBlockBody* = BeaconBlockBody | SigVerifiedBeaconBlockBody | TrustedBeaconBlockBody
 
   SyncSubcommitteeIndex* = distinct uint8
-  ValidatorIndexInSyncCommittee* = distinct uint16
+  IndexInSyncCommittee* = distinct uint16
 
 chronicles.formatIt BeaconBlock: it.shortLog
 chronicles.formatIt SyncSubcommitteeIndex: uint8(it)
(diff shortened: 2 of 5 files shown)


beacon_chain/spec/beaconstate.nim
```diff
@@ -19,6 +19,9 @@ import
 
 export extras, forks, validator
 
+type
+  SomeBeaconState* = phase0.BeaconState | altair.BeaconState | merge.BeaconState
```
arnetheduck: already exists as `ForkyBeaconState` in `forks.nim`

beacon_chain/spec/beaconstate.nim
```diff
@@ -855,3 +858,7 @@ func upgrade_to_merge*(cfg: RuntimeConfig, pre: altair.BeaconState):
     # Execution-layer
     latest_execution_payload_header: ExecutionPayloadHeader()
   )
+
+template isValidInState*(idx: ValidatorIndex, state: SomeBeaconState): bool =
+  idx.int < state.validators.len
```
arnetheduck: this cast should be to `uint64` and compare with `lenu64`

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -440,7 +441,12 @@ template asInt*(x: SyncSubcommitteeIndex): int = int(x)
 template asUInt8*(x: SyncSubcommitteeIndex): uint8 = uint8(x)
 template asUInt64*(x: SyncSubcommitteeIndex): uint64 = uint64(x)
 
-template `[]`*(a: auto; i: SyncSubcommitteeIndex): auto = a[i.asInt]
+template `[]`*(a: auto; i: SyncSubcommitteeIndex): auto =
+  a[i.asInt]
```
arnetheduck, line 445: why `asInt` sometimes, and `.int` sometimes? the `asInt` helpers are pretty confusing to me, not sure what they're supposed to offer?

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -430,6 +430,24 @@ proc readValue*(reader: var JsonReader[RestJson], value: var ValidatorIndex) {.
...
+               {.raises: [IOError, SerializationError, Defect].} =
+  let svalue = reader.readValue(string)
+  let res = Base10.decode(uint64, svalue)
+  if res.isOk():
+    let v = res.get()
+    if v < SYNC_COMMITTEE_SIZE:
+      value = ValidatorIndexInSyncCommittee(v)
+    else:
+      reader.raiseUnexpectedValue(
+        "Validator index is bigger then SYNC_COMMITTEE_SIZE")
```
arnetheduck:
```suggestion
        "Index in committee is bigger than SYNC_COMMITTEE_SIZE")
```

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -432,6 +432,7 @@ type
   SomeBeaconBlockBody* = BeaconBlockBody | SigVerifiedBeaconBlockBody | TrustedBeaconBlockBody
 
   SyncSubcommitteeIndex* = distinct uint8
+  ValidatorIndexInSyncCommittee* = distinct uint16
```
arnetheduck: this is not a validator index, really - `IndexIn..` would be more appropriate

arnetheduck: that said, we've used `int` in other places for similar index-in-list indices

beacon_chain/spec/beaconstate.nim
zah: Actually, it already exists in `datatypes/merge.nim`. Importing `forks.nim` here would have been a cyclic dependency.

The naming `SomeFoo` is more inline with the practice of the Nim standard library, so I might remove the use of the `Forky` type. 

EDIT: Nevermind, apparently I was looking at a stale branch.

arnetheduck: https://github.com/status-im/nimbus-eth2/pull/3050

beacon_chain/spec/datatypes/altair.nim
zah: In general, having overloads such as `asInt` signals that somebody consciously approved the validity of this conversion.

When you use a direct coercion, the code will continue to compile even if you change the underlying base type of the distinct in a way that makes the coercion unsafe. For this particular distinct type, this is unlikely to happen, but in other situations it might make a difference.

arnetheduck: I guess I'm asking because 2 lines below, plain `int` is used in a similar situation - in theory, typing `int` is also a consciously approved validity claim - ie that's the motivating logic behind implicit vs explicit coercions in the language, so it's a bit hard to distinguish the two cases

zah: To clarify further, the `asInt` approach allows you outlaw the conversion in the future in a way that will automatically catch all call sites. The `int` coercion doesn't offer this. The single definition in the code can document why the conversion is safe and the assumptions can be challenged by any team member in the future stumbling upon it.

Regarding the use of `idx.int` few lines below... it's just that `validatorIndex.int` is already used in too many places, so I've left it for another day to propose a similar treatment for it as well.

zah: p.p.

We've previously considered using the `as` operator as a gate-keeper for such "safe" conversions. It will have the similar property that you can outlaw conversions at will.

arnetheduck:
@zah this is a compile failure due to the import issue

zah:
I know, but I'm not able to reproduce it locally

zah:
Merged manually the earlier green commit. The review comments will be addressed in https://github.com/status-im/nimbus-eth2/pull/3069

PR #3070, opened 2021-11-09
Reduce the logging-related breaking changes

zah:
* Don't disable the stdout log when the `--log-file` option is used
* Rename `--log-stdout` to `--log-format` and hide it
* Deprecate the `--log-file` option

### DIFF
--- a/docs/the_nimbus_book/src/logging.md
+++ b/docs/the_nimbus_book/src/logging.md
@@ -50,3 +50,4 @@ To send logs to a file, you can redirect the stdout logs:
 ```
 
 We recommend keeping an eye on the growth of this file with a [log rotator](./log-rotate.md). Logs are written in the "JSON Lines" format - one `json` entry per line.
+
--- a/docs/the_nimbus_book/src/options.md
+++ b/docs/the_nimbus_book/src/options.md
@@ -23,7 +23,7 @@ The following options are available:
 
      --log-level               Sets the log level for process and topics (e.g. "DEBUG;
                                TRACE:discv5,libp2p; REQUIRED:none; DISABLED:none") [=INFO].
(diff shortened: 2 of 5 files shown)


docs/the_nimbus_book/src/logging.md
```diff
@@ -55,4 +55,4 @@ Nimbus also supports writing logs to a log file using the `--log-file` option -
 ./run-mainnet-beacon-node.sh --log-file=filename.jsonl # write json logs to the given filename
 ```
 
-When the `--log-file` option is enabled, stdout logs will by default be disabled. You can enable them with the `--log-stdout` option (choose any format other than `auto`).
+When the `--log-file` option is enabled, stdout logs will by default be disabled. You can enable them with the `--log-format` option (choose any format other than `auto`).
```
arnetheduck: this line needs removing

docs/the_nimbus_book/src/options.md
```diff
@@ -23,7 +23,7 @@ The following options are available:
 
      --log-level               Sets the log level for process and topics (e.g. "DEBUG;
                                TRACE:discv5,libp2p; REQUIRED:none; DISABLED:none") [=INFO].
-     --log-stdout              Specifies what kind of logs should be written to stdout
+     --log-format              Specifies what kind of logs should be written to stdout
```
arnetheduck:
```suggestion
     --log-format              Specifies what kind of logs should be written to stdout (beta)
```

docs/the_nimbus_book/src/logging.md
zah: Removed by Sacha in another PR.

arnetheduck (approved):

PR #3075, opened 2021-11-09
remove intsets usage

tersec:
- Remove some `result` usage
- Avoid potentially unsafe-at-runtime `int` <-> `GeneralizedIndex`conversions by using `HashSet[GeneralizedIndex]` instead of `IntSet`
- Avoid creation of a temporary set

There's nothing magically efficient about `excl`ing a whole `HashSet` or `IntSet` -- Nim just iterates with the same `for` loop one might use here:
- https://github.com/nim-lang/Nim/blob/version-1-2/lib/pure/collections/sets.nim#L306
- https://github.com/nim-lang/Nim/blob/version-1-6/lib/pure/collections/sets.nim#L322
- https://github.com/nim-lang/Nim/blob/version-1-2/lib/pure/collections/intsets.nim#L327

There appears to be no particular reason `excl` requires another set (`IntSet`, `HashSet`, `PackedSet`, or otherwise) as its second argument, rather than any iterable in general, except for language design reasons.

When we switch to Nim 1.6, it will be worth benchmarking and potentially replacing usage of `HashSet` with https://nim-lang.org/docs/packedsets.html, and one can then get something like `IntSet` type-safely, here and elsewhere in `nimbus-eth2`; I believe that all the usages are ordinal. Indeed, https://github.com/nim-lang/Nim/blob/version-1-6/lib/pure/collections/intsets.nim shows that Nim 1.6 literally defines `IntSet` as `PackedSet[int]`.

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -11,7 +11,7 @@
 
 import
   # Standard lib
-  std/[algorithm, intsets, math, sequtils, tables],
+  std/[algorithm, math, sequtils, sets, tables],
   # Status libraries
   stew/[byteutils, endians2, bitops2],
   chronicles,
@@ -93,21 +93,21 @@ func get_helper_indices*(
   ## to prove the chunks with the given generalized indices. Note that the
   ## decreasing order is chosen deliberately to ensure equivalence to the order
   ## of hashes in a regular single-item Merkle proof in the single-item case.
-  var all_helper_indices = initIntSet()
-  var all_path_indices = initIntSet()
+  var all_helper_indices = initHashSet[GeneralizedIndex]()
   for index in indices:
     for idx in get_branch_indices(index):
-      all_helper_indices.incl idx.int
+      all_helper_indices.incl idx
+  for index in indices:
     for idx in get_path_indices(index):
-      all_path_indices.incl idx.int
-  all_helper_indices.excl all_path_indices
+      all_helper_indices.excl idx
 
-  result = newSeqOfCap[GeneralizedIndex](all_helper_indices.len)
+  var res = newSeqOfCap[GeneralizedIndex](all_helper_indices.len)
   for idx in all_helper_indices:
-    result.add idx.GeneralizedIndex
-  result.sort(SortOrder.Descending)
+    res.add idx
+  res.sort(SortOrder.Descending)
+  res
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.2/ssz/merkle-proofs.md#merkle-multiproofs
+# https://github.com/ethereum/consensus-specs/blob/v1.1.5/ssz/merkle-proofs.md#merkle-multiproofs
 func check_multiproof_acceptable*(
     indices: openArray[GeneralizedIndex]): Result[void, string] =
   # Check that proof verification won't allocate excessive amounts of memory.


arnetheduck (approved):

beacon_chain/spec/helpers.nim
```diff
@@ -93,21 +93,21 @@ func get_helper_indices*(
...
+  var all_helper_indices = initHashSet[GeneralizedIndex]()
   for index in indices:
     for idx in get_branch_indices(index):
-      all_helper_indices.incl idx.int
+      all_helper_indices.incl idx
+  for index in indices:
     for idx in get_path_indices(index):
-      all_path_indices.incl idx.int
-  all_helper_indices.excl all_path_indices
+      all_helper_indices.excl idx
```
etan-status, line 102: This will lead to more scans through the entirety of `all_helper_indices` as duplicates are no longer eliminated in advance. But for the few cases that we are using so far it should not matter. Has the potential to have different performance than the previous implementation, though.

beacon_chain/spec/helpers.nim
```diff
@@ -93,21 +93,21 @@ func get_helper_indices*(
   ## to prove the chunks with the given generalized indices. Note that the
   ## decreasing order is chosen deliberately to ensure equivalence to the order
   ## of hashes in a regular single-item Merkle proof in the single-item case.
-  var all_helper_indices = initIntSet()
-  var all_path_indices = initIntSet()
+  var all_helper_indices = initHashSet[GeneralizedIndex]()
```
etan-status, line 96: Thanks. This will be nice when `GeneralizedIndex` advances to `uint32` or `uint64` in the future (likely)!

beacon_chain/spec/helpers.nim
tersec: Howso? Comparing with

https://github.com/status-im/nimbus-eth2/blob/fdbfe19b3ea44646363eff3ad98ad43baa35e8cb/beacon_chain/spec/helpers.nim#L89-L108

The sequence of operations there is:
1. calculate (completely, not pruning along the way) both `all_helper_indices` and `all_path_indices`, both of which avoid duplication internally because that's what `incl` does with both `IntSet` and `HashSet`

2. in-place subtract `all_path_indices` from `all_helper_indices`

The new sequence of operations is:
1. calculate (completely, not pruning along the way) `all_helper_indices`, which avoids duplication internally because that's what `incl` does with both `IntSet` and `HashSet`

2. don't ever construct `all_path_indices`, which was only a kind of deduplicated cache of the union of `get_path_indices()` call results. instead, directly `excl` `get_path_indices()` results one-by-one.

In neither case are duplicates ever stored. The performance difference is that where a duplication-probe used to be in `incl` in `all_path_indices.incl idx.int`, now it's in `all_helper_indices.excl idx`, which depends on `excl` doing nothing when an element doesn't exist.

The status of duplicates as regards `all_helper_indices` is 100% identical between both flows.

etan-status: > "which depends on `excl` doing nothing when an element doesn't exist.

Multiple branches may share some of their `path_indices`, so we are now excluding the same index multiple times.
If looking up whether an item exists is truly 0 cost, then excluding the same index repeatedly is not changing things, sure. On the other hand, the de-duplication of the spec approach also had some run-time cost associated with it. I don't know which method is faster, just pointing out that performance characteristics may have changed. New implementation uses a bit less memory at least.

tersec: `incl` is implemented via https://github.com/nim-lang/Nim/blob/8ccde68f132be4dba330eb6ec50f4679e564efac/lib/pure/collections/setimpl.nim#L45-L55 and

`excl` via https://github.com/nim-lang/Nim/blob/8ccde68f132be4dba330eb6ec50f4679e564efac/lib/pure/collections/setimpl.nim#L76-L95

The relevant part is `var index = rawGet(s, key, hc)` / `var i = rawGet(s, key, hc)`, which detect whether the key's already in the table. It's an identical mechanism regardless.

etan-status: Yeah, it probably depends on the query, how big those sets are getting.
When `all_path_indices` is small, and there are a ton of duplicates, and `all_helper_indices` is big, the spec one may still be faster, simply because of the set size.
Anyhow, it is too early to know how the usage pattern will be, and if it becomes a problem / spec version ends up actually being faster, it can still be changed once it becomes a problem.

PR #3077, opened 2021-11-09
Remote signing client/server.

cheatfate:
This PR adds support for https://github.com/ConsenSys/web3signer remote signature server. From my point of view `web3signer` is the most full version of remote signer, because EIP-3030 (https://eips.ethereum.org/EIPS/eip-3030) do not support enough entities to run even on `phase0` network.

To use remote signature for one of the validator's public keys you need to create new keystore file inside `<data>/validators/<public_key>` directory.

This file should be placed in `validators/<public key>/remote_keystore.json` file.

While loading keys, `nimbus_beacon_node` will follow this rules.
First it will try to load local validator keystore `keystore.json`. If it not accessible
it will try to load `remote_keystore.json` file. Only if both attempts filed it will show an error.

#### Example

```json
{
  "version": "1",
  "description": "This is simple remote keystore file",
  "type": "web3signer",
  "pubkey": "0x8107ff6a5cfd1993f0dc19a6a9ec7dc742a528dd6f2e3e10189a4a6fc489ae6c7ba9070ea4e2e328f0d20b91cc129733",
  "remote": "http://127.0.0.1:15052",
  "ignore_ssl_verification": true
}
```

#### Fields

1. `version` - optional string encoded integer.
2. `description` - optional string encoded description.
3. `type` - optional (by default `web3signer`) string encoded type of remote
signer. Right now only `web3signer` value is supported, but maybe in future we are going to support `dirk`.
4. `pubkey` - required hexadecimal string of validator's public key.
5. `remote` - required URL of remote service.
6. `ignore_ssl_verification` - optional boolean flag which provides option to  ignore self-signed certificates.


beacon_chain/spec/keystore.nim
```diff
@@ -497,6 +512,75 @@ proc readValue*(r: var JsonReader, value: var Kdf)
     r.raiseUnexpectedValue(
       "The Kdf value should have sub-fields named 'function' and 'params'")
 
+proc readValue*(r: var JsonReader, value: var RemoteKeystore)
```
zah, line 515: Why do we need to implement a custom serialization for this type? Isn't the definition something that we control? Why is the default implementation not adequate?

beacon_chain/conf.nim
```diff
@@ -33,6 +33,7 @@ const
   # Maybe there should be a config option for this.
   defaultListenAddress* = (static ValidIpAddress.init("0.0.0.0"))
   defaultAdminListenAddress* = (static ValidIpAddress.init("127.0.0.1"))
+  DefaultSigningNodeRequestTimeout* = 60
```
zah: We tend to use `camelCase` for constants (just like the lines above). `PascalCase` is reserved for types or compile-time variables storing types.

beacon_chain/conf.nim
```diff
@@ -590,7 +594,84 @@ type
...
+      desc: "Server identifier which will be used in HTTP Host header"
+      name: "server-ident" }: Option[string]
+
+    requestTimeout* {.
+      desc: "Request timeout, maximum time that node will wait for remote " &
+            "client request (in seconds)"
+      defaultValue: DefaultSigningNodeRequestTimeout
+      name: "request-timeout" }: int
+
+    case cmd* {.
```
zah, line 662: Do you anticipate any commands to be added in the future?

beacon_chain/nimbus_signing_node.nim
```diff
@@ -0,0 +1,355 @@
...
+
+proc initValidators(sn: var SigningNode): bool =
+  info "Initializaing validators", path = sn.config.validatorsDir()
+  var duplicates: seq[ValidatorPubKey]
+  var publicKeyIdents: seq[string]
+  for item in sn.config.validatorItems():
+    case item.kind
+    of ValidatorKind.Local:
+      let pubkey = item.privateKey.toPubKey().toPubKey()
+      if pubkey in duplicates:
```
zah: You can check whether the validator was already added with the `getValidator` function of the `ValidartorPool` (`sn.attachedValidators`). This will result in a linear algorithms instead of the current quadratic one which is not that nice considering that we run servers with thousands of validators.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -319,6 +324,14 @@ proc sszResponse*(t: typedesc[RestApiResponse], data: auto): RestApiResponse =
 template hexOriginal(data: openarray[byte]): string =
   "0x" & ncrutils.toHex(data, true)
 
+proc decodeJsonString*[T](t: typedesc[T],
+                          data: JsonString,
+                          requireAllFields = true): Result[T, cstring] =
+  try:
+    ok(RestJson.decode(string(data), T, requireAllFields = requireAllFields))
+  except SerializationError:
```
zah, line 332: The `SerializationError` exception carries useful diagnostic information (i.e. line and column information for the unexpected syntax/values in the json string). Why is this helper considered an improvement if it destroys the diagnostic information?

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -945,6 +958,355 @@ proc readValue*(reader: var JsonReader[RestJson],
   else:
     reader.raiseUnexpectedValue($res.error())
 
+# Web3SignerRequest
+proc writeValue*(writer: var JsonWriter[RestJson],
```
zah, line 962: `nim-json-serialization` already supports case objects. Why is this definition necessary?

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -945,6 +958,355 @@ proc readValue*(reader: var JsonReader[RestJson],
...
+  var
+    requestKind: Option[Web3SignerRequestKind]
+    forkInfo: Option[Web3SignerForkInfo]
+    signingRoot: Option[Eth2Digest]
+    data: Option[JsonString]
+    dataName: string
+
+  for fieldName in readObjectFields(reader):
+    case fieldName
+    of "type":
```
zah, line 1066: Is any of the web3signer implementations not listing the `type` field as a first field in the `Web3SignerRequest`? nim-json-serialization already supports case objects and should be able to handle the decoding otherwise.

beacon_chain/spec/eth2_apis/rest_remote_signer_calls.nim
```diff
@@ -0,0 +1,184 @@
...
+    of 200:
+      inc(nbc_remote_signer_200_responses)
+      let res = decodeBytes(Web3SignerSignatureResponse, response.data,
+                            response.contentType)
+      if res.isErr():
+        let msg = "Unable to decode remote signer response [" &
+                  $res.error() & "]"
+        inc(nbc_remote_signer_failures)
+        return Web3SignerDataResponse.err(msg)
+      let sig = res.get().signature.load()
```
zah, line 112: I think it makes sense to verify that the signature is valid here (against the public key and the data) before we broadcast the message on the network. 

I can imagine some configuration issues that may result in invalid signatures. If this happens, we don't want to get de-scored and kicked out from the network if only some of the validators have invalid configuration. Furthermore, we need to inform the user for the problem instead of failing silently.

beacon_chain/validators/keystore_management.nim
```diff
@@ -312,8 +328,29 @@ proc loadKeystoreUnsafe*(validatorsDir, secretsDir,
...
+  let remoteKeystorePath = validatorsDir / keyName / RemoteKeystoreFileName
+  let privateItem =
+    block:
+      let keystore =
+        try:
+          Json.decode(remoteKeystorePath, RemoteKeystore)
+        except SerializationError as e:
+          error "Failed to read remote keystore file",
+                keystore_path = remoteKeystorePath,
+                err_msg = e.formatMsg("<remote_keystore>")
```
zah:
```suggestion
                err_msg = e.formatMsg(remoteKeystorePath)
```

The exception carries only line and column information (to reduce the size of intermediate objects and to avoid string allocations). The argument to `formatMsg` is the filename that will be printed together with the line and column error.

beacon_chain/validators/validator_duties.nim
```diff
@@ -86,37 +86,56 @@ proc findValidator(validators: auto, pubKey: ValidatorPubKey):
 proc addLocalValidator(node: BeaconNode,
                        validators: auto,
                        item: ValidatorPrivateItem) =
-  let pubKey = item.privateKey.toPubKey()
-  node.attachedValidators[].addLocalValidator(
-    item,
-    findValidator(validators, pubKey.toPubKey()))
+  let
+    pubKey = item.privateKey.toPubKey()
+    index = findValidator(validators, pubKey.toPubKey())
```
zah, line 91: It's worth checking whether these linear searches contribute to poor start-up times of our Prater fleet nodes. We can populate a temporary index with a hash table to get a linear algorithm.

beacon_chain/validator_client/duties_service.nim
```diff
@@ -152,20 +152,30 @@ proc pollForAttesterDuties*(vc: ValidatorClientRef,
       res
 
   if len(addOrReplaceItems) > 0:
-    var pending: seq[Future[ValidatorSig]]
+    var pending: seq[Future[SignatureResult]]
+    var validators: seq[AttachedValidator]
     for item in addOrReplaceItems:
       let validator = vc.attachedValidators.getValidator(item.duty.pubkey)
```
zah, line 158: Is there a guarantee that `getValidator` cannot return `nil` here?

beacon_chain/validators/validator_duties.nim
```diff
@@ -86,37 +86,56 @@ proc findValidator(validators: auto, pubKey: ValidatorPubKey):
...
+        res.incl({HttpClientFlag.NoVerifyHost,
+                  HttpClientFlag.NoVerifyServerName})
+      res
+  let prestoFlags = {RestClientFlag.CommaSeparatedArray}
+  let client = RestClientRef.new($item.remoteUrl, prestoFlags, httpFlags)
+  if client.isErr():
+    warn "Unable to resolve remote signer address",
+         remote_url = $item.remoteUrl, validator = item.publicKey
+    return
+  let index = findValidator(validators, item.publicKey.toPubKey())
```
zah, line 109: Ditto

beacon_chain/conf.nim
cheatfate: I'm not decided what tool should be responsible to help user with creation of `remote_keystore.json` files.

beacon_chain/validator_client/duties_service.nim
cheatfate: @zah if you check line 125 you will see

```nim
    relevantDuties = duties.filterIt(
      checkDuty(it) and (it.pubkey in vc.attachedValidators)
    )
```
It means that `relevantDuties` will hold only duties with validators we know.

Also you could check that `addOrReplaceItems` is just a subset of `relevantDuties`, so `getValidator` could not return `nil` here.

beacon_chain/validators/validator_duties.nim
cheatfate: Answer is above.

beacon_chain/validators/validator_duties.nim
cheatfate: @zah this is not a new code it was present there before my PR. Even if this linear search over 100k validators could take some time in milliseconds, the biggest problem is how keystores are encoded/decoded (number of iterations) seriously affects our performance.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
cheatfate: @zah This is current specification https://consensys.github.io/web3signer/web3signer-eth2.html and i dont see any requirements for field `type` to be first, This is JSON, so `type` field could first or could be last. That's actually why i need to implement this custom type deserialization.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
cheatfate: `nim-json-serialization` supports case objects only when `case` delimiter is specified before other fields, it means that JSON should be formed in some strict way before we will be able to use it.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
cheatfate: Its not an improvement its the way how to avoid exceptions and code duplication in object parser's code.

Issue #3073, opened 2021-11-09
Kintsugi🍵


tersec:
- [x] M0: Implementation started
- [x] M1: Spec Implemented (https://github.com/status-im/nimbus-eth2/pull/3065, https://github.com/status-im/nimbus-eth2/pull/3066, https://github.com/status-im/nimbus-eth2/pull/3072)
- [x] M2: 1-to-1 Interop – One EL and one CL team pair off for basic interop testing
  - geth: https://github.com/status-im/nimbus-eth2/blob/unstable/docs/interop_merge.md
  - nethermind: https://github.com/status-im/nimbus-eth2/blob/kintsugi/docs/neth-m2-nimbus.md
- [x] M3: Many-to-Many Interop – 2 or more EL and CL teams perform interop testing
- [x] M4: Weekly devnet – Weekly devnet build for more in depth testing
- [x] M5: All-to-All persistent testnet – All ready EL and CL teams create a long-lived devnet
- [ ] M6: Kintsugi🍵 code released – clients release Kintsugi🍵 code for public consumption

Main implementation branch: https://github.com/status-im/nimbus-eth2/pull/3093

https://hackmd.io/@n0ble/kintsugi-spec defines the interop meta-specs and https://notes.ethereum.org/@djrtwo/kintsugi-milestones defines milestones.

tersec:
In lieu of Kiln

PR #3093, opened 2021-11-12
Kintsugi

tersec:
Replaces https://github.com/status-im/nimbus-eth2/pull/2930
For https://github.com/status-im/nimbus-eth2/issues/3073

Don't merge.


beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -61,14 +63,25 @@ type
...
 template slot*(blck: BlockRef): Slot = blck.bid.slot
 
-func init*(T: type BlockRef, root: Eth2Digest, slot: Slot): BlockRef =
+func init*(
+    T: type BlockRef, root: Eth2Digest, executionPayloadRoot: Eth2Digest,
+    slot: Slot): BlockRef =
   BlockRef(
-    bid: BlockId(root: root, slot: slot)
+    bid: BlockId(root: root, slot: slot),
+    executionBlockRoot: executionPayloadRoot,
```
arnetheduck, line 70: this will be expensive until we get rid of history - ie we can't put this on unstable (`BlockRef` is already the largest mem consumer there)

tersec: Indeed. My current intention is to use a smallish fixed-size cache. The reason it's there to begin with is that in principle, the head can keep moving arbitrarily in tree re-orgs. Probably one of the more reasonable approaches is a fixed-size LRU-ish (it doesn't have to be exactly) cache of CL block hash -> CL execution block roots.

I'd have to look at it again, but some codepaths within which the head is selected know without reloading blocks the actual content of the block, including the `ExecutionPayload.block_hash`, so avoid the database reload.

However, part of the goal is not to over-optimize the good case at the expense of making pathological cases even more pathological, so there are certain approaches I'll probably not exploit.

But I'm not going to fix it, per se, in this branch, but in a PR to `unstable` which, after a rebase, this will replace all this code in this branch.

tersec: In particular, I want to avoid relatively more subtle and error-prone flushing/invalidation infrastructure that anything non-fixed-size or tied to anything in-memory which is already otherwise managed like this provide by default.

tersec:
Replaced by https://github.com/status-im/nimbus-eth2/pull/3368

PR #3096, opened 2021-11-12
ncli_db: add putState, putBlock

arnetheduck:
These tools allow modifying an existing nimbus database for the purpose
of recovery or reorg, moving the head, tail and genesis to arbitrary
points.

* remove potentially expensive `putState` in `BeaconStateDB`
* introduce `latest_block_root` which computes the root of the latest
applied block from the `latest_block_header` field (instead of passing
it in separately)
* avoid some unnecessary BeaconState copies during init
* discover nim-lang/Nim#19094
* prefer `HashedBeaconState` in a few places to avoid recomputing state
root
* fetch latest block root from state when creating blocks
* harden `get_beacon_proposer_index` against invalid slots and document
* move random spec function tests to `test_spec.nim`
* avoid unnecessary state root computation before block proposal


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1394,12 +1394,12 @@ proc preInit*(
           tail_genesis_validators_root = shortLog(tail_genesis_validators_root)
         quit 1
 
-      let blck = get_initial_beacon_block(state.data)
+      let blck = get_initial_beacon_block(state)
       db.putGenesisBlock(blck.root)
       db.putBlock(blck)
 
+      db.putStateRoot(state.latest_block_root(), state.data.slot, state.root)
```
tersec, line 1401: Here and elsewhere, the `putState()`/`putStateRoot()` order switches. One nice aspect of doing `putState()` first is that it ensures, at all times, an internal consistency in when being accessed via the usual state-root mechanism. There can be un-pointed-to-states, but no dangling pointers from state roots.

arnetheduck: yeah, I can't decide which one is worse :) when reading, we don't rely on "pointers" being valid, and we `delState` before `delStateRoot` with the idea that root pointers are cheap and there may be lots of them for various informational purposes, dangling or not, lying around in the database - we don't do transactions, so we also avoid hard cross-table assumptions

research/simutils.nim
```diff
@@ -105,15 +105,10 @@ proc loadGenesis*(validators: Natural, validate: bool):
...
-      Eth2Digest(),
-      0,
-      deposits,
-      flags)[]
-
-    res.phase0Data.root = hash_tree_root(res[].phase0Data.data)
+    let res = (ref ForkedHashedBeaconState)(
+      kind: BeaconStateFork.Phase0,
+      phase0Data: initialize_hashed_beacon_state_from_eth1(
+        cfg, Eth2Digest(), 0, deposits, flags))
```
tersec, line 111: `ForkedHashedBeaconState.init(...)`? It's all initialized at once this way regardless.

tersec (approved):

research/simutils.nim
arnetheduck: yes, that would work if it wasn't for https://github.com/nim-lang/Nim/issues/19094

PR #3106, opened 2021-11-15
book: add FRITZ!Box UPnP setup

stefantalpalaru:


docs/the_nimbus_book/src/networking.md
```diff
@@ -1,25 +1,30 @@
 # Network setup
 
-Nimbus will automatically connect to peers based on the health and quality of peers that it's already connected to. Depending on the network and the number of validators attached to the the node, Nimbus may need anywhere from 10 to 60 peers connected to operate well.
+Nimbus will automatically connect to peers based on the health and quality of peers that it's already connected to. Depending on the network and the number of validators attached to the node, Nimbus may need anywhere from 10 to 60 peers connected to operate well.
 
-In addition to making outgoing connections, the beacon node node works best when others can connect to the node - this speeds up the process of finding good peers.
+In addition to making outgoing connections, the beacon node node works best when others can connect to it - this speeds up the process of finding good peers.
 
-To allow incoming connections, the peer must be reachable via a public IP address.
+To allow incoming connections, the node must be reachable via a public IP address and it must be aware of it, so it can advertise it to its peers.
```
unixpi:
```suggestion
To allow incoming connections, the node must be reachable via a public IP address. It must also be aware of this address, so that it can advertise it to its peers.
```

docs/the_nimbus_book/src/networking.md
```diff
@@ -1,25 +1,30 @@
...
 
-By default, Nimbus uses UPnP to set up port forwarding and detect your external IP address. If you do not have UPnP enabled, you may need to pass additional command-line options to the node, as detailed below.
-A collection of tips and tricks to help improve your network connectivity.
+## UPnP
+
+By default, Nimbus uses [UPnP](https://en.wikipedia.org/wiki/Universal_Plug_and_Play) to set up port forwarding and detect your external IP address. If you do not have UPnP enabled, you may need to pass additional command-line options to the node, as explained in subsequent sections.
+
+To enable UPnP, it's usually as simple as checking a box in your router's configuration. Unless it's a FRITZ!Box router, that is.
```
unixpi:
```suggestion
Enabling UPnP is usually as simple as checking a box in your router's configuration. Unless it's a FRITZ!Box router, that is.
```

docs/the_nimbus_book/src/networking.md
```diff
@@ -1,25 +1,30 @@
...
+
+With this brand, you also need to edit individual connections - in "Home Network" -> "Network" -> edit icon -> "Permit independent port sharing for this device". You might also want to enable "Always assign this network device the same IPv4 address", in case the setting is associated with IPs instead of MACs.
```
unixpi:
```suggestion
With this brand, you will also need to edit individual connections - in "Home Network" -> "Network" -> edit icon -> "Permit independent port sharing for this device". You might also want to enable "Always assign this network device the same IPv4 address", in case the setting is associated with IPs instead of MACs.
```

unixpi (commented):
great

docs/the_nimbus_book/src/networking.md
```diff
@@ -1,25 +1,30 @@
...
-To allow incoming connections, the peer must be reachable via a public IP address.
+To allow incoming connections, the node must be reachable via a public IP address. It must also be aware of this address, so that it can advertise it to its peers.
 
-By default, Nimbus uses UPnP to set up port forwarding and detect your external IP address. If you do not have UPnP enabled, you may need to pass additional command-line options to the node, as detailed below.
-A collection of tips and tricks to help improve your network connectivity.
+## UPnP
+
+By default, Nimbus uses [UPnP](https://en.wikipedia.org/wiki/Universal_Plug_and_Play) to set up port forwarding and detect your external IP address. If you do not have UPnP enabled, you may need to pass additional command-line options to the node, as explained in subsequent sections.
+
+Enabling UPnP is usually as simple as checking a box in your router's configuration. Unless it's a FRITZ!Box router, that is.
```
arnetheduck, line 13: for device-specific config, perhaps better move to a footnote section - ie there are a lot of devices out there, so the inline style doesn't scale, and the information here is uninteresting to (presumably) the vast majority of users.

stefantalpalaru: I mostly use it as a showcase of non-trivial UPnP configuration, to let people know they might need to dig deeper if it doesn't work by default.

PR #3108, opened 2021-11-16
Nim GC metrics for the main thread

stefantalpalaru:
Screenshot 2021-11-16 at 15-26-45 NBC local testnet sim (all nodes) - Grafana

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -984,10 +984,6 @@ proc onSlotEnd(node: BeaconNode, slot: Slot) {.async.} =
   # This is the last pruning to do as it clears the "needPruning" condition.
   node.consensusManager[].pruneStateCachesAndForkChoice()
 
-  # Nim GC metrics (for the main thread)
-  {.gcsafe.}:
-    updateSystemMetrics()
-
   when declared(GC_fullCollect):
     # The slots in the beacon node work as frames in a game: we want to make
     # sure that we're ready for the next one and don't get stuck in lengthy
@@ -1129,6 +1125,10 @@ proc onSecond(node: BeaconNode) =
   ## This procedure will be called once per second.
   if not(node.syncManager.inProgress):
     node.handleMissingBlocks()
+
+  # Nim GC metrics (for the main thread)
+  {.gcsafe.}:
+    updateSystemMetrics()
 
 proc runOnSecondLoop(node: BeaconNode) {.async.} =
   let sleepTime = chronos.seconds(1)
@@ -1885,7 +1885,7 @@ proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref BrHmacDrbgContext) {.r
     except Exception as exc:
       raiseAssert exc.msg # TODO fix metrics
 
-  # Nim GC metrics (for the main thread) will be collected in onSlotEnd(), but
+  # Nim GC metrics (for the main thread) will be collected in onSecond(), but
   # we disable piggy-backing on other metrics here.
   setSystemMetricsAutomaticUpdate(false)
 
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -984,6 +984,10 @@ proc onSlotEnd(node: BeaconNode, slot: Slot) {.async.} =
   # This is the last pruning to do as it clears the "needPruning" condition.
   node.consensusManager[].pruneStateCachesAndForkChoice()
 
+  # Nim GC metrics (for the main thread)
+  {.gcsafe.}:
+    updateSystemMetrics()
+
   when declared(GC_fullCollect):
     # The slots in the beacon node work as frames in a game: we want to make
     # sure that we're ready for the next one and don't get stuck in lengthy
@@ -1859,13 +1863,6 @@ proc loadEth2Network(config: BeaconNodeConf): Eth2NetworkMetadata {.raises: [Def
       echo "Must specify network on non-mainnet node"
       quit 1
 
-proc update_system_metrics_task() {.async.} =
-  let sleepTime = chronos.seconds(10)
-  while true:
-    await chronos.sleepAsync(sleepTime)
-    {.gcsafe.}:
-      updateSystemMetrics()
-
 proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref BrHmacDrbgContext) {.raises: [Defect, CatchableError].} =
   info "Launching beacon node",
       version = fullVersionStr,
@@ -1887,9 +1884,10 @@ proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref BrHmacDrbgContext) {.r
       raise exc
     except Exception as exc:
       raiseAssert exc.msg # TODO fix metrics
-    # Nim GC metrics (for the main thread)
-    setSystemMetricsAutomaticUpdate(false)
-    asyncSpawn update_system_metrics_task()
+
+  # Nim GC metrics (for the main thread) will be collected in onSlotEnd(), but
+  # we disable piggy-backing on other metrics here.
+  setSystemMetricsAutomaticUpdate(false)
 
   # There are no managed event loops in here, to do a graceful shutdown, but
   # letting the default Ctrl+C handler exit is safe, since we only read from
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1859,6 +1859,13 @@ proc loadEth2Network(config: BeaconNodeConf): Eth2NetworkMetadata {.raises: [Def
       echo "Must specify network on non-mainnet node"
       quit 1
 
+proc update_system_metrics_task() {.async.} =
+  let sleepTime = chronos.seconds(10)
+  while true:
+    await chronos.sleepAsync(sleepTime)
+    {.gcsafe.}:
+      updateSystemMetrics()
+
 proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref BrHmacDrbgContext) {.raises: [Defect, CatchableError].} =
   info "Launching beacon node",
       version = fullVersionStr,
@@ -1876,8 +1883,13 @@ proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref BrHmacDrbgContext) {.r
       url = "http://" & $metricsAddress & ":" & $config.metricsPort & "/metrics"
     try:
       startMetricsHttpServer($metricsAddress, config.metricsPort)
-    except CatchableError as exc: raise exc
-    except Exception as exc: raiseAssert exc.msg # TODO fix metrics
+    except CatchableError as exc:
+      raise exc
+    except Exception as exc:
+      raiseAssert exc.msg # TODO fix metrics
+    # Nim GC metrics (for the main thread)
+    setSystemMetricsAutomaticUpdate(false)
+    asyncSpawn update_system_metrics_task()
 
   # There are no managed event loops in here, to do a graceful shutdown, but
   # letting the default Ctrl+C handler exit is safe, since we only read from
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1129,6 +1129,7 @@ proc onSecond(node: BeaconNode) =
   # Nim GC metrics (for the main thread)
   {.gcsafe.}:
     updateSystemMetrics()
+    updateThreadMetrics()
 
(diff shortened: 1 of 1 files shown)


zah:
What if the same type is allocated in both threads? Wouldn't this lead to random reporting of the value of one of the threads?

arnetheduck:
As per the comments in the [metrics PR](https://github.com/status-im/nim-metrics/pull/39), there are several issues with this approach:

* it's still wrong - ie the per-type metrics are global for all threads while the gc metric is for one main thread alone - having recently introduced threading, it's only going to become more wrong over time, and one value does not add up with the other
* the GC mem usage can potentially be computed by simply summing the per-type metrics
* in the slot loop, we run a full GC at the end of every slot - a metric that shows memory usage should _probably_ be coordinated with that - the regular "swing" we see in memory usage right now is largely due to this recycling and it helps give a hint about how much memory we allocate per slot (mostly because we send gossip, and libp2p / chronos is very memory hungry) - it's a useful way to diagnose it
* the metrics PR introduces a regression: previously, the per-type memory statistics would be captured the instant when the metrics were being loaded by the poller - this PR introduces an interaction between the async task and the poller, making this approach less accurate and useful.

all in all, one of the better things we can do probably is to remove the faulty metric completely - in the metrics library, it's a design issue - here, it's simply a minor broken feature for which we lack a good solution, as of yet. removing it is arguably a lesser evil, because the other metrics are generally more useful.

arnetheduck:
>  value of one of the threads

oddly, `-d:nimTypeNames` introduces a global per-type counter (with some caveats for aliases and similar types), not a per-thread one - if we were writing an application that relied heavily on multithreaded memory allocation, this would of course be a major performance concern, but we're not, so it's not.

stefantalpalaru:
> What if the same type is allocated in both threads? Wouldn't this lead to random reporting of the value of one of the threads?

The Nim compiler sums those up for us, regardless from which thread we call the undocumented iterator: https://github.com/nim-lang/Nim/blob/92d6fb86c66ceda154972314ba695cea4aac9301/lib/system/gc_common.nim#L40

Probably because type statistics (number of instances and sizes) are kept in a single linked-list for all threads - the one headed by `nimTypeRoot`.

>  the per-type metrics are global for all threads while the gc metric is for one main thread alone

Thus spoken the upstream API. It's better than before, so take the incremental improvement, instead of blocking it for five more months.

> the GC mem usage can potentially be computed by simply summing the per-type metrics

No, it cannot, because we do not collect metrics for all types, but only for the 10 heaviest types for each collection step.

> in the slot loop, we run a full GC at the end of every slot - a metric that shows memory usage should probably be coordinated with that

Sure. We're still catching some spikes, due to one interval being 10s and the other 12s, but getting these metrics before the GC collection makes sense.

> the metrics PR introduces a regression: previously, the per-type memory statistics would be captured the instant when the metrics were being loaded by the poller - this PR introduces an interaction between the async task and the poller, making this approach less accurate and useful.

A few seconds of delay are not that important, when looking at the data aggregated over much larger intervals, which we usually do.

> one of the better things we can do probably is to remove the faulty metric completely

Just like we removed the Chronos metrics, I assume. Somehow, replacements did not magically appear and we are now less informed than 20 months ago: https://github.com/status-im/nim-chronos/pull/85

"Perfect" is the enemy of "good" (or "better", in our case).

stefantalpalaru:
GC metrics update moved to `onSlotEnd()`, before the call to `GC_fullCollect()`.

Minimal difference in Grafana:
Screenshot 2021-11-16 at 21-57-26 NBC local testnet sim (all nodes) - Grafana

arnetheduck:
> No, it cannot, because we do not collect metrics for all types, but only for the 10 heaviest types for each collection step.

the idea would be to do it on the nim-metrics side, as the top 10 are being collected

> It's better than before,

I think this is the contention point - this PR replaces one incorrect value with another incorrect value - a few months back, this might have been ok, but now threads are becoming a critical part of the application - in testing, the threading already showed two leaks which means that showing a value that looks like it might be OK is worse than showing a completely bogus one - the latter can simply be discarded as broken - the former actively misleads investigations. If we go down the route where we continue producing a metric like this, it should be renamed, so that this point is made clear. Alternatively, the broken metric can simply be removed - metric readers generally missing values. In particular, this does not break anything that previously worked.

> A few seconds of delay are not that important, when looking at the data aggregated over much larger intervals, which we usually do.

this is not so much about a few seconds delay, but rather how the feature makes the metric less accurate  and potentially misleading - previously, you could run a poller at lower intervals and get more accurate data - now, it's capped at this internal rate.

stefantalpalaru:
> now threads are becoming a critical part of the application

The fix could be as trivial as adding a "thread_id" label to thread-specific metrics, but we can't iterate on stuck PRs and we cannot develop nim-metrics in isolation.

zah:
I must agree that the current approach introduced in nim-metrics is quite hacky. 

Let's discuss what an ideal solution would look like. If we want a single thread to be able to report an accurate an up-to-date information regarding the GC stats of all threads, we'll need a way to enumerate the currently active Nim threads and to access the addresses of their thread-local `gch.region.{currMem, maxMem, freeMem, occ}` values. This would be relatively easy to implement by patching Nim's system module and it would be possible to introduce without breaking compatibility with vanilla Nim, so it is something that we can consider. The solution can also be up-streamed if we frame it as a set of hooks around tread creation and destruction, so the bulk of the logic can be implemented in a user-space package.

If we don't have the above capability, we have no other option besides relying on stale values. Right now, the nim-metrics approach chooses a rather hacky approach to update the stale values. It may have been better to just place an `updateSystemMetrics` or a separate cheaper `updateGcMetrics` call in the `onSecond` function in NBC. To address @arnetheduck's concerns, the update interval should probably be significantly lower than the Prometheus polling interval.

It's also possible to produce an always-up-to-date cached value for the GC stats by relying on the Nim's "hooks" GC:

https://github.com/nim-lang/Nim/blob/devel/lib/system/gc_hooks.nim#L45-L47

The way this would work is that we'll install our own "custom" GC which just delegates the hooked calls to the standard Nim GC. At the end of each allocation, we will record the per-thread GC stats in global array with enough slots to cover all of our threads.

stefantalpalaru:
> It may have been better to just place an updateSystemMetrics or a separate cheaper updateGcMetrics call in the onSecond function in NBC.

Done.

arnetheduck:
> To address @arnetheduck's concerns, the update interval should probably be significantly lower than the Prometheus polling interval.

It's important to remember that this remains a compromise, though one that is slightly more palatable because it's less likely to give bogus values (at the expense of performance).

The fundamental issue here is that prometheus is a polling system - when we update in onsecond and prom polls at a minute interval, 59 updates are useless - this is the main difference between statsd and prometheus, or push and pull - when implementing both as `nim-metrics` does, you have to compromise on one side and it's odd to me to introduce the compromise in such a way that our _main_ "user" of `nim-metrics` gets the shorter stick - specially because the polling model was already implemented and working (for the per-type metrics) - we're essentially making an effort to work against the system and then introduce a work-around to bring back a minimal baseline.

arnetheduck:
> It's also possible to produce an always-up-to-date cached value

This sounds like unnecessary overhead - what we really want to do is to update it only when it's used - for most metrics, it doesn't really matter that much, but for some, like memory allocations, hooking into the gc sounds like a major intrusion and source of extreme inefficiency, for .. no reason in particular.

arnetheduck:
> GC metrics update moved to onSlotEnd(), before the call to GC_fullCollect().

this is a excellent example of how the metrics become misleading to the point of being useless, in this model - we know from experience that there are major memory swings (in the hundreds of MB) because of the full GC run due to how `async` introduces cyclic references that get collected by the full gc run. Looking at this graph, it looks like nothing is happening.

stefantalpalaru:
> this is a excellent example of how the metrics become misleading to the point of being useless

Before, livelock and after the move to `onSecond()` (same initial sync, Prater):

Screenshot 2021-11-25 at 19-11-41 NBC local testnet sim (all nodes) - Grafana

> we know from experience that there are major memory swings (in the hundreds of MB) because of the full GC run due to how async introduces cyclic references that get collected by the full gc run

But are we sure those appear in the GC stats we get from the Nim runtime? Here, I changed Prometheus' scraping interval from 12s to 1s, at 19:19:30 :

Screenshot 2021-11-25 at 19-22-14 NBC local testnet sim (all nodes) - Grafana

If there were major GC usage swings due to that collection triggered once every 12s slot, we would see it here.

stefantalpalaru:
Also, are you sure those major memory swings appear during the initial block sync?

stefantalpalaru:
Sorry, forgot to change the data source scraping interval in Grafana. Fixed:

Screenshot 2021-11-25 at 19-37-27 NBC local testnet sim (all nodes) - Grafana

arnetheduck:
Screenshot from 2021-11-25 21-04-56

Here's a typical full collection swing, on the order of several hundred MB, if you aggregate the types - it's because of gossip, and how async creates cycles and hogs memory that only the a cycle collector can capture - @Menduist is investigating the cycles, and the magnitude of the swing is correlated to the inefficiency of data duplication.

Having mechanical sympathy between the collector (prometheus) and the producer (nimbus) increases the accuracy of these metrics - prometheus uses pull metrics, and it stands to reason that the library uses that as well, when possible - in rust/lighthouse for example, many key metrics are pulled from the client on demand which gives them an edge. It matters greatly even if it's not precise, just like a speedometer on a car is useful only if it's reasonably accurate, even if it's not perfect.

arnetheduck:
Since the top 10 types are swinging, you can be pretty sure that the "GC used" mem is swinging too - as mentioned earlier, an easy way to get an accurate and total "GC used" for all threads together is to sum it in the metrics library - this doesn't need any extensions to the language, doesn't need plugging in to the gc, and doesn't require a periodic update from an event loop.

Menduist:
I actually do my tests on top of this branch to have some reasonable metric about ram usage

I do not care too much about nim_gc_mem_bytes, since it's more or less RSS, but nim_gc_mem_occupied_bytes is very interesting to have some idea of what's going on
(for more granularity, I would use my [heaptrack hack](https://github.com/status-im/Nim/tree/heaptrack_support))

If summing up types usage in nim-metrics is accurate enough, imo let's do that, and get rid of gc_mem_bytes

zah:
> This sounds like unnecessary overhead - what we really want to do is to update it only when it's used

Nim is already updating these counters on every allocation. What the hook needs to do is just copy 4 integer values to another location. I doubt there will be a noticeable impact. The accuracy of the "sum the top 10 types" idea obviously depends on the specific application, so it doesn't fit very well in a general purpose library. It may be described as a custom "computed" metric perhaps that is defined by Nimbus.

zah:
We are also conflating the user needs with the needs of the developers here. We, the developers, have vastly more options for tracking the memory usage. Looking at Grafana when you hunt down memory issues is just handicapping yourself unnecessarily.

arnetheduck:
> What the hook needs to do is just copy 4 integer values to another location. 

the hook is an indirect call - it would need to do more atomic work to count correctly, meaning more interthread interference

> The accuracy of the "sum the top 10 types" 

It's not "top 10" - it's all types - nim-metrics currently fetches all types, sorts them by size then reports top 10 - in that same loop it can compute a sum - it's trivial, and I remain curious about what drives the need to complicate matters with compiler changes, hooks, unnecessary polling and so on, ie in all this, there's no .. er .. benefit and several downsides?

arnetheduck:
> Looking at Grafana when you hunt down memory issues is just handicapping yourself unnecessarily.

It allows us developers to diagnose "normal" usage of nimbus as part of its standard feature set, such that when something goes wrong in a user system, we have a cheap and easy way to collect basic information that might give us leads in investigations.

stefantalpalaru:
> in that same loop it can compute a sum

I did that, summing up all instance types in a new metric, and it's coming up a bit short:

```text
$ curl -sS http://127.0.0.1:8008/metrics | grep nim_gc | grep -Ev 'created|type_name|^#'
nim_gc_mem_bytes{thread_id="15016"} 1582231552.0
nim_gc_mem_occupied_bytes{thread_id="15016"} 676859760.0
nim_gc_heap_instance_occupied_summed_bytes 634171659.0
```

Do we still want it, for multi-threaded programs where you might not be interested in setting up per-thread statistics?

nim-metrics changes here: https://github.com/status-im/nim-metrics/pull/41

arnetheduck:
> I did that, summing up all instance types in a new metric, and it's coming up a bit short:

Probably because it's not counting the GC overhead, but rather just what is used for types - which is fine

arnetheduck:
> Do we still want it

The thread id in `nim_gc_mem*` is a good improvement - it clearly spells out what the metric actually represents instead of misleading the user.

I've summarized the remaining regressions and ways that the newly pushed work breaks existing use cases in https://github.com/status-im/nim-metrics/issues/43

PR #3113, opened 2021-11-18
Make sync horizon configurable

arnetheduck:
Currently, we don't have a good answer to the question "are we synced
yet" - the sync manager syncs based on the peers it's connected to, but
just because some peer looks like it should be synced from doesn't mean
we're out of sync.

Instead, we use a very silly time-based heuristic - the problem with
that is that the network can go into a rut where nobody produces blocks
- better heuristics would be needed here, but in the meantime, a command
line option can get us out of a tight spot - this PR places such an
option in the client, in the unlikely event it should be needed (most
likely in a testnet).

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -348,6 +348,13 @@ type
         name: "doppelganger-detection"
       }: bool
 
+      syncHorizon* {.
+        hidden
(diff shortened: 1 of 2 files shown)


tersec (approved):
It's clearly a stopgap, but at least, it's a stopgap which imposes little technical debt and is easy to remove if/when appropriate.

PR #3115, opened 2021-11-22
Speed up altair block processing 2x

### LINKED ISSUE #2985: [Performance] Batch-verify and cache verification of SyncAggregate
The altair syncaggregate are costly and unnecesarily rechecked when a trusted block is replayed from the DB.

1. Introduce a TrustedSyncAggregate data structure
  https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/spec/datatypes/altair.nim#L81-L84

The `process_sync_aggregate` should use the `withTrusted` template

https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/spec/signatures.nim#L15-L19

This avoids recomputing BLS verification here:

https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/spec/state_transition_block.nim#L437-L440

2. Gossip validation should use batch verification:

In 2 locations https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/gossip_processing/gossip_validation.nim#L819-L825

https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/gossip_processing/gossip_validation.nim#L938-L949

https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/spec/signatures.nim#L245-L257

3. `verify_selection_proof_signature` probably needs the same treatment

https://github.com/status-im/nimbus-eth2/blob/592e4928a117eba6cdd76cac248be3009d3b41c3/beacon_chain/spec/signatures.nim#L278-L292

arnetheduck:
Like #3089, this PR drastially speeds up historical REST queries and
other long state replays.

* cache sync committee validator indices
* use ~80mb less memory for validator pubkey mappings
* batch-verify sync aggregate signature (fixes #2985)
* document sync committee hack with head block vs sync message block
* add batch signature verification failure tests

Before:

```
../env.sh nim c -d:release -r ncli_db --db:mainnet_0/db bench --start-slot:-1000
All time are ms
     Average,       StdDev,          Min,          Max,      Samples,         Test
Validation is turned off meaning that no BLS operations are performed
    5830.675,        0.000,     5830.675,     5830.675,            1, Initialize DB
       0.481,        1.878,        0.215,       59.167,          981, Load block from database
    8422.566,        0.000,     8422.566,     8422.566,            1, Load state from database
       6.996,        1.678,        0.042,       14.385,          969, Advance slot, non-epoch
      93.217,        8.318,       84.192,      122.209,           32, Advance slot, epoch
      20.513,       23.665,       11.510,      201.561,          981, Apply block, no slot processing
       0.000,        0.000,        0.000,        0.000,            0, Database load
       0.000,        0.000,        0.000,        0.000,            0, Database store
```

After:

```
    7081.422,        0.000,     7081.422,     7081.422,            1, Initialize DB
       0.553,        2.122,        0.175,       66.692,          981, Load block from database
    5439.446,        0.000,     5439.446,     5439.446,            1, Load state from database
       6.829,        1.575,        0.043,       12.156,          969, Advance slot, non-epoch
      94.716,        2.749,       88.395,      100.026,           32, Advance slot, epoch
      11.636,       23.766,        4.889,      205.250,          981, Apply block, no slot processing
       0.000,        0.000,        0.000,        0.000,            0, Database load
       0.000,        0.000,        0.000,        0.000,            0, Database store
```

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -785,7 +785,7 @@ proc validateSyncCommitteeMessage*(
 
   ok((positionsInSubcommittee, cookedSignature.get()))
 
-# https://github.com/ethereum/eth2.0-specs/blob/v1.1.0-alpha.8/specs/altair/p2p-interface.md#sync_committee_contribution_and_proof
+# https://github.com/ethereum/eth2.0-specs/blob/v1.1.5/specs/altair/p2p-interface.md#sync_committee_contribution_and_proof
 proc validateSignedContributionAndProof*(
     dag: ChainDAGRef,
     syncCommitteeMsgPool: var SyncCommitteeMsgPool,
@@ -855,16 +855,22 @@ proc validateSignedContributionAndProof*(
     initialized = false
     syncCommitteeSlot = msg.message.contribution.slot + 1
 
-  for validatorPubKey in dag.syncCommitteeParticipants(
+  for validatorIndex in dag.syncCommitteeParticipants(
       syncCommitteeSlot,
       committeeIdx,
       msg.message.contribution.aggregation_bits):
-    let validatorPubKey = validatorPubKey.loadWithCache.get
+    let validatorPubKey = dag.validatorKey(validatorIndex)
+    if not validatorPubKey.isSome():
+      # This should never happen (!)
+      warn "Invalid validator index in committee cache",
+        validatorIndex
+      return errIgnore("SignedContributionAndProof: Invalid committee cache")
+
     if not initialized:
       initialized = true
-      committeeAggKey.init(validatorPubKey)
+      committeeAggKey.init(validatorPubKey.get())
     else:
-      committeeAggKey.aggregate(validatorPubKey)
+      committeeAggKey.aggregate(validatorPubKey.get())
 
   if not initialized:
     # [REJECT] The contribution has participants
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -237,7 +237,7 @@ proc initialize_beacon_state_from_eth1*(
                                              deposits.len)
   state.eth1_deposit_index = deposits.lenu64
 
-  var pubkeyToIndex = initTable[ValidatorPubKey, int]()
+  var pubkeyToIndex = initTable[ValidatorPubKey, ValidatorIndex]()
   for idx, deposit in deposits:
     let
       pubkey = deposit.pubkey
@@ -249,7 +249,7 @@ proc initialize_beacon_state_from_eth1*(
     do:
       if skipBlsValidation in flags or
          verify_deposit_signature(cfg, deposit):
-        pubkeyToIndex[pubkey] = state.validators.len
+        pubkeyToIndex[pubkey] = ValidatorIndex(state.validators.len)
         if not state.validators.add(get_validator_from_deposit(deposit)):
           raiseAssert "too many validators"
         if not state.balances.add(amount):
@@ -707,6 +707,9 @@ func get_next_sync_committee_keys(state: altair.BeaconState | merge.BeaconState)
   ## Return the sequence of sync committee indices (which may include
   ## duplicate indices) for the next sync committee, given a ``state`` at a
   ## sync committee period boundary.
+  # The sync committe depends on seed and effective balance - it can
+  # thus only be computed for the current epoch of the state, after balance
+  # updates have been performed
 
   let epoch = get_current_epoch(state) + 1
 
@@ -744,9 +747,9 @@ proc get_next_sync_committee*(state: altair.BeaconState | merge.BeaconState):
   # see signatures_batch, TODO shouldn't be here
   # Deposit processing ensures all keys are valid
   var attestersAgg: AggregatePublicKey
-  attestersAgg.init(res.pubkeys.data[0].loadWithCache().get)
+  attestersAgg.init(res.pubkeys.data[0].load().get)
   for i in 1 ..< res.pubkeys.data.len:
-    attestersAgg.aggregate(res.pubkeys.data[i].loadWithCache().get)
+    attestersAgg.aggregate(res.pubkeys.data[i].load().get)
 
   res.aggregate_pubkey = finish(attestersAgg).toPubKey()
   res
@@ -922,3 +925,36 @@ func latest_block_root*(state: ForkyBeaconState, state_root: Eth2Digest): Eth2Di
 
 func latest_block_root*(state: ForkyHashedBeaconState): Eth2Digest =
   latest_block_root(state.data, state.root)
+
+func get_sync_committee_cache*(
+    state: altair.BeaconState | merge.BeaconState, cache: var StateCache):
+    SyncCommitteeCache =
+  let period = state.slot.sync_committee_period()
+
+  cache.sync_committees.withValue(period, v) do:
+    return v[]
+
+  var
+    s = toHashSet(state.current_sync_committee.pubkeys.data)
+
+  for pk in state.next_sync_committee.pubkeys.data:
+    s.incl(pk)
+
+  var pubkeyIndices: Table[ValidatorPubKey, ValidatorIndex]
+  for i, v in state.validators:
+    if v.pubkey in s:
+      pubkeyIndices[v.pubkey] = i.ValidatorIndex
+
+  var res: SyncCommitteeCache
+  try:
+    for i in 0..<res.current_sync_committee.len():
+      res.current_sync_committee[i] =
+        pubkeyIndices[state.current_sync_committee.pubkeys[i]]
+      res.next_sync_committee[i] =
+        pubkeyIndices[state.next_sync_committee.pubkeys[i]]
+  except KeyError:
+    raiseAssert "table constructed just above"
+
+  cache.sync_committees[period] = res
+
+  res
--- a/beacon_chain/spec/signatures_batch.nim
+++ b/beacon_chain/spec/signatures_batch.nim
@@ -415,6 +415,8 @@ proc collectSignatureSets*(
             DOMAIN_VOLUNTARY_EXIT)
 
   block:
+    # 7. SyncAggregate
+    # ----------------------------------------------------
     withState(state):
       when stateFork >= BeaconStateFork.Altair and
           (signed_block is altair.SignedBeaconBlock or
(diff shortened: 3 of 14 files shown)


beacon_chain/spec/beaconstate.nim
```diff
@@ -744,9 +747,9 @@ proc get_next_sync_committee*(state: altair.BeaconState | merge.BeaconState):
   # see signatures_batch, TODO shouldn't be here
   # Deposit processing ensures all keys are valid
   var attestersAgg: AggregatePublicKey
-  attestersAgg.init(res.pubkeys.data[0].loadWithCache().get)
```
zah, line 747: What's wrong with using the cache here?

beacon_chain/validators/validator_duties.nim
```diff
@@ -729,8 +730,9 @@ proc handleSyncCommitteeContributions(node: BeaconNode,
     for subcommitteeIdx in allSyncSubcommittees():
       # TODO Hoist outside of the loop with a view type
       #      to avoid the repeated offset calculations
-      for valKey in syncSubcommittee(syncCommittee, subcommitteeIdx):
-        let validator = node.getAttachedValidator(valKey)
+      for valIdx in syncSubcommittee(syncCommittee, subcommitteeIdx):
+        let validator = node.getAttachedValidator(
+          getStateField(node.dag.headState.data, validators), valIdx)
```
zah, line 735: Didn't we recently learn that `getStateField(node.dag.headState.data, validators)` is a very expensive copy operation? The better alternative is:

```nim
withState(node.dag.headState.data): state.data.validators[valIdx]
```

beacon_chain/spec/beaconstate.nim
arnetheduck: The cache would only be used once per day for the 512 lookups and uses quite a lot of memory (absolute minimum 144 bytes per validator + hashset and threadvar overhead which is significant) - thanks to the index cache (about 4kb fixed size), the cost no longer matches the benefit. There are no other uses of this cache outside of test cases (I'd generally remove it).

There's an additional optimization that can be done to pre-seed the statecache with the index cache as long as the states line up, but that's for a separate PR - the main benefit is during replays where many blocks need to be applied - the creation of the index mapping is the elephant in the room.

beacon_chain/validators/validator_duties.nim
arnetheduck: `getStateField` is quite efficient thanks to a (somewhat) ugly pointer trick. It is no more (or less) performant than `withState` afaik.

The problem with accessing `headState.data` in general is that it's mutable - that causes the mutable overload of `[](x: var HashList, ...)` to be chosen which invalidates the cache unnecessarily - Nim always prefers the mutable overload given a choice.

The short-term solution is to bypass the cache with `toSeq()` in relevant cases - here, it should not be relevant because it's passed to a function that accepts an immutable argument and does the right thing.

beacon_chain/spec/signatures_batch.nim
```diff
@@ -414,4 +414,46 @@ proc collectSignatureSets*(
             volex.message.epoch,
             DOMAIN_VOLUNTARY_EXIT)
 
+  block:
```
mratsim, line 417: For documentation it would be nice to have a comment section
`7. SyncAggregate`

beacon_chain/validators/validator_duties.nim
```diff
@@ -235,16 +235,16 @@ proc sendSyncCommitteeMessages*(node: BeaconNode,
 
       let (keysCur, keysNxt) =
         block:
-          var resCur: Table[ValidatorPubKey, int]
-          var resNxt: Table[ValidatorPubKey, int]
+          var resCur: Table[uint64, int]
+          var resNxt: Table[uint64, int]
```
mratsim, line 239: Some alias type that tells us what this magic uint64 corresponds to?

arnetheduck: they're `uint64` values from the message that are repeated in the response, and in the message, they don't have any alias type - they're kind of validator indices, except we don't want to use `ValidatorIndex` because that's messy from a type size perspective.

beacon_chain/spec/signatures_batch.nim
arnetheduck: 8e445dc156786e003852048ed69149a92b2dc949 - there's not much to write though, like proposer signature, it's just a single field with a signature in it

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -855,16 +855,22 @@ proc validateSignedContributionAndProof*(
     initialized = false
     syncCommitteeSlot = msg.message.contribution.slot + 1
 
-  for validatorPubKey in dag.syncCommitteeParticipants(
+  for validatorIndex in dag.syncCommitteeParticipants(
       syncCommitteeSlot,
       committeeIdx,
       msg.message.contribution.aggregation_bits):
-    let validatorPubKey = validatorPubKey.loadWithCache.get
+    let validatorPubKey = dag.validatorKey(validatorIndex)
```
zah, line 862: The previous use of the cache here was quite important for performance. @mratsim, is `dag.validatorKey(index)` considered fast enough? (it's based internally on `loadValid` which in turn is based on `fromBytesKnownOnCurve`).

beacon_chain/spec/beaconstate.nim
zah: Previously, the keys would end up in the cache anyway because the cache was also used in gossip validation. I guess it's now possible to introduce a more precise eviction of the cache around the update points of the `SyncCommitteeCache` (or the cooked keys could be stored in the `SyncCommitteeCache`, but this seems potentially sub-optimal if we end up needing the same validators keys in other caches).

beacon_chain/gossip_processing/gossip_validation.nim
zah: `block_sim` is a good benchmark for this, because it simulates the full network traffic around sync committees and their aggregation. The `ncli_db bench` just replays blocks, so it covers only the sync aggregate verification code.

arnetheduck: `ImmutableValidatorData2` stores uncompressed pubkeys that are quite fast to load - this was introduced a while ago, giving a decent performance boost: https://github.com/status-im/nimbus-eth2/commit/d859bc12f088546816929aa4db626a21e32eeadb

We're using this same approach for attestations and everything else, pretty much - if we're going to update it, we should update it across the board by changing the backend storage for `validatorKey` - sync message validation was the outlier here, and `loadWithCache` is in general a problematic construct.

beacon_chain/spec/beaconstate.nim
arnetheduck: we have two caches of the same data, where one would be enough: `loadWithCache` and `validatorKey` - the latter is tied to our database storage and should be preferred, also because _constructing_ it is faster: `loadWithCache` works from compressed keys. If anything, we could keep cooked keys in the in-memory representation that backs `validatorKey`

beacon_chain/gossip_processing/gossip_validation.nim
zah: Posting before/after `block_sim` results will be a convincing evidence that there is no performance regression.

mratsim: I've added a benchmark for pubkeys and signature deserialization.
On a 11th gen Intel it's 14 µs per key.

image

https://github.com/status-im/nim-blscurve/pull/126

mratsim: Looking at https://github.com/nim-lang/Nim/pull/13440, the table speed with hashing (we don't need to hash in our use-case), perf is reported in nanoseconds (using epochTime) so we have a factor 1000x.

arnetheduck: > sync aggregate verification code.

`bench` doesn't cover signature verification at all: it covers replays which which works with trusted data - we've been having issues with these taking up disproportionate amounts of time

`block_sim`, pre:
```
Validators: 12800, epoch length: 32
Validators per attestation (mean): 0.0
All time are ms
     Average,       StdDev,          Min,          Max,      Samples,         Test
      19.121,       18.328,        5.200,      172.740,          186, Process non-epoch slot with block
     157.283,      172.965,       25.783,      434.367,            6, Process epoch slot with block
       0.050,        0.009,        0.012,        0.069,          192, Tree-hash block
       0.466,        0.025,        0.411,        0.542,          192, Sign block
     156.436,        8.348,      140.617,      182.255,          192, Have committee attest to block
     415.829,      293.114,        0.003,      728.960,          192, Produce sync committee actions
      54.980,        0.000,       54.980,       54.980,            1, Replay all produced blocks
```

post:

```
Validators: 12800, epoch length: 32
Validators per attestation (mean): 0.0
All time are ms
     Average,       StdDev,          Min,          Max,      Samples,         Test
      19.051,       17.602,        4.826,      159.903,          186, Process non-epoch slot with block
     164.300,      203.499,       21.694,      527.819,            6, Process epoch slot with block
       0.048,        0.009,        0.011,        0.066,          192, Tree-hash block
       0.444,        0.021,        0.389,        0.484,          192, Sign block
     148.513,        6.841,      130.055,      165.031,          192, Have committee attest to block
     387.641,      272.799,        0.002,      691.795,          192, Produce sync committee actions
      55.288,        0.000,       55.288,       55.288,            1, Replay all produced blocks
```

the differences are pretty much within the error margin - ie I don't _think_ the faster sync committee processing is relevant but rather down to luck

arnetheduck: that said, it's likely a valid optimization point regardless, to load the data into cooked pubkeys on startup.

arnetheduck: I'm not sure 13440 is relevant here - different data set, multiple flaws in methodology

zah: Yes, if loading uncompressed keys is so fast, loading them on start-up shouldn't be a problem either.

arnetheduck: separate PR though, since it affects attestations significantly

arnetheduck:
@mratsim what are you measuring there btw? compressed or uncompressed deserialization?

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: #3122

PR #3118, opened 2021-11-24
implement terminal-total-difficulty-override; keep kintsugi TTDs consistent

tersec:


beacon_chain/conf.nim
```diff
@@ -355,6 +356,14 @@ type
         defaultValueDesc: "50"
         name: "sync-horizon" }: uint64
 
+      # https://github.com/ethereum/consensus-specs/blob/v1.1.5/specs/merge/client-settings.md#override-terminal-total-difficulty
+      terminalTotalDifficultyOverride* {.
+        hidden
+        desc: "Override pre-configured TERMINAL_TOTAL_DIFFICULTY parameter"
+        defaultValue: defaultTerminalTotalDifficultyOverride
+        name: "terminal-total-difficulty-override"
+      }: uint64
```
arnetheduck: can't this be an `Option[uint64]` instead of that default value trick?

tersec: https://github.com/status-im/nimbus-eth2/pull/3118/commits/c4b1dff118034afd9638b13f45b8df5e220143d5

arnetheduck (approved):

PR #3122, opened 2021-11-25
Keep cooked pubkeys in cache

arnetheduck:
Turning uncompressed pubkeys into cooked ones is fast, but unnecessary -
this should avoid a little work for every signature validation we do by
pre-loading them at startup.

### DIFF
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -282,8 +282,12 @@ type
 
   # Non-spec type that represents the immutable part of a validator - an
   # uncompressed key serialization is used to speed up loading from database
-  ImmutableValidatorData2* = object
+  ImmutableValidatorDataDb2* = object
     pubkey*: UncompressedPubKey
+    withdrawal_credentials*: Eth2Digest
+
+  ImmutableValidatorData2* = object
+    pubkey*: CookedPubKey
     withdrawal_credentials*: Eth2Digest
 
   # https://github.com/ethereum/consensus-specs/blob/v1.1.5/specs/phase0/beacon-chain.md#validator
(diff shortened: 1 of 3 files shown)


arnetheduck:
Running block_sim shows a small improvement (a few %) - this is expected because block_sim doesn't really benchmark key loading in any significant way - nonetheless, no loadValid work is faster than fast loadValid work.

beacon_chain/spec/datatypes/base.nim
```diff
@@ -282,10 +282,14 @@ type
...
   # Non-spec type that represents the immutable part of a validator - an
   # uncompressed key serialization is used to speed up loading from database
-  ImmutableValidatorData2* = object
+  ImmutableValidatorDataDb2* = object
     pubkey*: UncompressedPubKey
     withdrawal_credentials*: Eth2Digest
 
+  ImmutableValidatorData2* = object
+    pubkey*: CookedPubKey
+    withdrawal_credentials*: Eth2Digest
```
mratsim, line 291: The types are hard to distinguish in actual code. Maybe rename `ImmutableValidatorCookedData`?

mratsim (approved):
LGTM apart from the confusing name

beacon_chain/spec/datatypes/base.nim
arnetheduck: well, they follow one pattern (variable and type), this would simply be another, equally confusing name - ie either you focus on the cooked:ness, or the db:ness, don't see a clear advantage to either and this is the minimal patch

arnetheduck: I'm sure that @zah has some advanced framework-based feature up his sleeve to deal with this that would resolve this 2-line problem with a multilayered abstract solution :)

PR #3124, opened 2021-11-25
move quarantine outside of chaindag

arnetheduck:
The quarantine has been part of the ChainDAG for the longest time, but
this design has a few issues:

* the function in which blocks are verified and added to the dag becomes
reentrant and therefore difficult to reason about - we're currently
using a stateful flag to work around it
* quarantined blocks bypass the processing queue leading to a processing
stampede
* the quarantine flow is unsuitable for orphaned attestations - these
should also should be quarantined eventually
* ~~blocks coming from the sync manager take up quarantine space - this
shouldn't be necessary - it's not expected that sync manager blocks
arrive out-of-order and if they do, sync manager will retry the download~~ requests may actually be responded to out of order, so this change was removed from the pr

Instead of processing the quarantine inside ChainDAG, this PR moves
re-queueing to `block_processor` which already is responsible for
dealing with follow-up work when a block is added to the dag

This sets the stage for keeping attestations in the quarantine as well.


arnetheduck:
I just realized that topups via `RequestManager` also need to go to the quarantine - these arrive out-of-order

beacon_chain/consensus_object_pools/block_quarantine.nim
```diff
@@ -8,23 +8,47 @@
...
+  Quarantine* = object
+    ## Keeps track of unvalidated blocks coming from the network
+    ## and that cannot yet be added to the chain
+    ##
+    ## This only stores blocks that cannot be linked to the
+    ## ChainDAGRef DAG due to missing ancestor(s).
+    ##
+    ## Trivially invalid blocks may be dropped before reaching this stage.
+
+    orphans*: Table[(Eth2Digest, ValidatorSig), ForkedSignedBeaconBlock] ##\
```
tersec, line 35: Requires resolving the specific of beacon block, block by block, rather than statically know ahead of time.

tersec: Same comment applies to this change throughout the PR, roughly. My preference has been to statically encode as much as feasible, rather than jump straight to using `ForkedFoo`. I don't strongly object to change, but it's worth noting the tradeoffs.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -191,7 +215,11 @@ proc storeBlock*(
     blockSlot = blck.get().slot,
     validationDur, queueDur, storeBlockDur, updateHeadDur
 
-  ok(blck.get())
+  for quarantined in self.consensusManager.quarantine[].pop(blck.get().root):
```
tersec, line 218: There's no intrinsic reason for this to use the `ForkedFoo` variation either, but anything else wouldn't required using more than a single incoming queue or encoding that queue differently in a way that wasn't necessarily better than just using `ForkedFoo`.

beacon_chain/consensus_object_pools/block_quarantine.nim
arnetheduck: yeah, I thought about that but you kind of have to, anyway, because each block is from a new slot so you can't take a "group decision" really.. the other reason is that the block queue in blockprocessor is also forked-beacon-block based which worked out to be easier, in terms of establishing a single ordering.

That said, all these little decisions merit a review now that the dust has settled - the introduction of `Forky` for example was one step in that direction, to establish a simple language for keeping things on the static side.

beacon_chain/gossip_processing/block_processor.nim
arnetheduck: there "sort of" is: when a block is resolved, the blocks that have it as parent can be from different forks - so you either have to check multiple forks with a `ForkedQuarantine` of sorts, or you accept that each block comes with a separate kind - this affects the queue order as well, where it makes sense that all children of the "current" block are processed before the "grand-children" are.

tersec: Yes, this approach smooths over what would have been some corner cases, and one can view it as optimistic for being more easily verified for correctness rather than putting that kind of boundary-condition handling separately, anew, in each subsystem.

beacon_chain/consensus_object_pools/block_quarantine.nim
tersec: The block processor being forked-beacon-block based I commented on elsewhere, but the initial vision for that was also to have multiple statically-typed lanes for different forks. I tried both approaches, and using forked blocks was more expedient and easier to reason about by enough of a margin to go with it. That would've made for a consistent story in this regard.

That said, yes, given that the block processor in general operates on the basis of forked blocks, it's natural enough (and the resulting code's less duplicative) to use them here that I'm okay with this change.

Partly, I'm anticipating a version of the quarantine which (a) doesn't have the `SLOTS_PER_EPOCH` size limitation and (b) stores things in a basically-ephemeral database (i.e. to the extent it has a schema, it's one designed to be thrown away if/when necessary insofar as it's basically a cache, rather than that one carefully maintains and migrates) rather than an in-memory `Table`, and for that purpose, consolidating the functions the way the `Forked` (vs `Forky`) version allows probably simplifies the logic enough to be worthwhile.

tersec (approved):

PR #3125, opened 2021-11-25
Polling support for the Eth1Monitor

zah:


beacon_chain/conf.nim
```diff
@@ -128,6 +128,12 @@ type
       desc: "One or more Web3 provider URLs used for obtaining deposit contract data"
       name: "web3-url" }: seq[string]
 
+    web3ForcePolling* {.
+      hidden
+      defaultValue: false
+      desc: "Force the use of polling when determing the head block of Eth1"
+      name: "web3-force-polling" }: bool
```
arnetheduck, line 135: why "force"? does it automatically choose polling under any circumstances?

zah: Yes, it chooses polling when a HTTP/S provider is used (since these providers don't offer push model subscriptions for new blocks).

zah: And polling is not the default for WebSocket providers because it consumes more requests.

PR #3126, opened 2021-11-25
Fix REST API issues affecting Vouch validator client

zah:


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -180,16 +180,28 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+    # The requested epoch may still be too far in the future.
+    if earliestSlotInPrevSyncPeriod > node.dag.head.slot:
+      if not(node.isSynced(node.dag.head)):
+        return RestApiResponse.jsonError(Http503, BeaconNodeInSyncError)
+      else:
+        return RestApiResponse.jsonError(Http400, EpochFromFutureError)
 
-    node.withStateForBlockSlot(bslot):
+    node.withStateForBlockSlot(node.dag.head.atSlot(earliestSlotInPrevSyncPeriod)):
```
arnetheduck, line 204: if I understand this correctly, this might put us at a slot _before_ the finalized checkpoint - these are extremely slow to load because we have to replay up to 32 epochs of data - if so, this will not fly at all

arnetheduck: basically, the correct / performant way to handle future requests is to use the head state

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -202,21 +214,38 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
                                  "List of sync committee participants is empty")
           kres
 
-      var duties =
+      # We use a local proc in order to reduce the code bloat from `withState`
```
arnetheduck, line 217: as mentioned earlier, the code bloat of `withState` is nothing compared to nim-serialization, chronicles, generics like Table and so on, as can be seen by C file sizes - this is basically a premature optimization by all estimates - that said, it doesn't hurt much either, except that it breaks style and makes reading the code slightly backwards

arnetheduck: one thing though: closures don't work very well inside `{.async.}` - `{.nimcall.}` is here to prevent accidental captures?

zah: Yes (for `nimcall`).

I distinguish two types of code bloat - unavoidable ones due to the nature of the problem being solved (nim-serialization) and unnecessary ones such as the one here.

I claim that the bloat from nim-serialization is unavoidable because any serialization framework will need to rely on type descriptions that are encoded and stored in the program in one way ore another. Nim-serialization takes the design choice that the type description of a type comes in the form of a function. The generated functions are quite minimalistic which is both helpful for performance and code size. You can reduce the code size by relying more on type descriptions stored as data and traversed with general functions in the vein of `genericAssignAux`, but I think this would be a poor trade-off.

Now, Nim may have some generics cache bugs that lead to the production of more copies of the same generic definition. At some point, I'll spend some time looking at the generated code to see what's happening and to address the problems. The changes are likely to be localized and easy to fix for the project as a whole, while the code bloat from `withState` and friends is dispersed and it will require a lot of manual incremental work to resolve. 

I don't think it's practical for someone to fix all `withState` instances at once unless we dedicate a properly planned monthly task for it, so incremental improvements are welcome for me.

arnetheduck: The bloat of `withState` is comparable to that of any other generic code - ie same as `ForkyBeaconBlock`, `TrustedBeaconBlock` or indeed `Table` and so on - it just feels odd to single it out as a significant contributor, when it's just one more, fairly minor (due to us normally factoring out "useful" code into separate functions that get reused) contributor..

It's this form of "savings" that I'm slightly against - the kind that exist _only_ to deal with `withState` (or any other minor contributor) and that work against readability and don't contribute to reuse or testing inother ways - that's the "premature optimization" part - you wouldn't be able to tell the difference of this code with a proc vs the "natural", more readable code that doesn't have the extra proc - not in binary size, not in compile time nor any other way.

arnetheduck:
these have been merged separately

PR #3128, opened 2021-11-25
Jenkins: macOS ARM64 CI job

stefantalpalaru:


jakubgs (approved):
Looks good, but please don't merge for now because for now we have just one M1 host, and it will be stuck working an endless queue of jobs.

jakubgs:
We have a second host added to CI: https://github.com/status-im/infra-ci/commit/7e7d110f

So it should be fine to merge this when ready.

PR #3131, opened 2021-11-26
Backward sync support for SyncManager.

cheatfate:
1. Unbound SyncQueue data structure from SyncManager.
2. Move peer scores constants to `peer_scores.nim`
3. Add support for backward and forward syncing.
4. Add more tests.


arnetheduck:
> Unbound peer scores.

what does this point mean? ie can peer scores go up without bounds?

cheatfate:
> > Unbound peer scores.
> 
> what does this point mean? ie can peer scores go up without bounds?

Sorry i just moved constants to `peer_scores.nim`

arnetheduck:
Something I just thought of: when we integrate this, we must ensure that we run forward sync before we run backward sync (ie backward sync should only run when we're synced with head)

beacon_chain/sync/sync_queue.nim
```diff
@@ -0,0 +1,783 @@
...
+
+  SyncQueueKind* {.pure.} = enum
+    Forward, Backward
+
+  SyncRequest*[T] = object
+    kind: SyncQueueKind
+    index*: uint64
+    slot*: Slot
+    count*: uint64
+    step*: uint64
```
arnetheduck, line 38: the `step` support could probably be removed - it's unlikely we'll ever use it

cheatfate: It still present in specification, and it actually can be used in future.

arnetheduck: Yes, but it's kind of a useless feature of the spec and we can simplify our implementation accordingly - the idea was that one would interleave requests from different clients, so you'd get block 0, 2, 4 from one and 1, 3, 5.. from the other but this is quite complex and doesn't have much benefit. It should imo never have been added to the spec to begin with - it's also an optional feature on the requesting side - ie we don't have to use it.

cheatfate: Anyway its not related to the purpose of this PR, isn't it? Lets remove this "unspecified" things in different PR, to avoid any breaks.

cheatfate:
> Something I just thought of: when we integrate this, we must ensure that we run forward sync before we run backward sync (ie backward sync should only run when we're synced with head)

The idea of this PR is to have support of the backward sync via SyncManager. All other checks should be done in other parts.

beacon_chain/sync/sync_queue.nim
arnetheduck: ok

beacon_chain/sync/sync_manager.nim
```diff
@@ -7,45 +7,24 @@
...
 {.push raises: [Defect].}
 
-import std/[
-  options, heapqueue, tables, strutils, sequtils, math, algorithm]
+import std/[options, heapqueue, tables, strutils, sequtils, math, algorithm]
 import stew/results, chronos, chronicles
-
 import
-  ../spec/datatypes/[phase0, altair],
+  ../spec/datatypes/[base, phase0, altair, merge],
```
arnetheduck, line 13: technically, `base` is not needed when the others are imported

beacon_chain/sync/sync_queue.nim
```diff
@@ -0,0 +1,783 @@
...
+            request_slot = item.request.slot,
+            request_count = item.request.count,
+            request_step = item.request.step,
+            blocks_map = getShortMap(item.request, item.data),
+            blocks_count = len(item.data), errCode = res.error,
+            topics = "syncman"
+
+      var resetSlot: Option[Slot]
+
+      if res.error == BlockError.MissingParent:
```
arnetheduck, line 606: Better use `case` here - if we add a new `BlockError` it will cause this code to fail to compile

arnetheduck (approved):

PR #3135, opened 2021-11-29
fix `LightClientUpdate` documentation

etan-status:
Some `LightClientUpdate` documentation of the spec is meant to span
multiple fields but was attached to a single, incorrect field in Nimbus.
This patch moves affected documentation to correct locations.

### DIFF
--- a/beacon_chain/spec/datatypes/altair.nim
+++ b/beacon_chain/spec/datatypes/altair.nim
@@ -161,18 +161,16 @@ type
     header*: BeaconBlockHeader  ##\
     ## Update beacon block header
 
-    next_sync_committee*: SyncCommittee ##\
-    ## Next sync committee corresponding to the header
+    # Next sync committee corresponding to the header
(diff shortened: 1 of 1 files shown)


zah:
Older versions of Nim used to attach `nnkComment` nodes to the type AST if you use syntax like:

```nim
type
  Foo = object
     x: int
        ## Documentation for the `x` field
     y: int # \
     ## Documentation for the `y` field
```

This is the reason you'll find this style used across many Nim codebases.

Eventually, the feature was removed and nowadays it doesn't matter where you'll place the comment.

etan-status:
Yeah, but regardless of syntax, it seems incorrect to me before this patch.

e.g., this one does not have anything to do with finality.
```
next_sync_committee_branch*: array[log2trunc(NEXT_SYNC_COMMITTEE_INDEX), Eth2Digest] ##\
     ## Finality proof for the update header
```

PR #3138, opened 2021-11-30
use BeaconTime instead of Slot in fork choice

tersec:
The `Slot` to `BeaconTime` change is required due to https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/phase0/fork-choice.md#on_block:
```python
    # Add proposer score boost if the block is timely
    time_into_slot = (store.time - store.genesis_time) % SECONDS_PER_SLOT
    is_before_attesting_interval = time_into_slot < SECONDS_PER_SLOT // INTERVALS_PER_SLOT
    if get_current_slot(store) == block.slot and is_before_attesting_interval:
        store.proposer_boost_root = hash_tree_root(block)
```
such that fork choice now depends on finer granularity than slots.

This PR shouldn't introduce any regressions in terms of what already works, but also doesn't introduce the
```python
    proposer_score = Gwei(0)
    if store.proposer_boost_root != Root():
        block = store.blocks[root]
        if get_ancestor(store, root, block.slot) == store.proposer_boost_root:
            num_validators = len(get_active_validator_indices(state, get_current_epoch(state)))
            avg_balance = get_total_active_balance(state) // num_validators
            committee_size = num_validators // SLOTS_PER_EPOCH
            committee_weight = committee_size * avg_balance
            proposer_score = (committee_weight * PROPOSER_SCORE_BOOST) // 100
    return attestation_score + proposer_score
```
logic from https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/phase0/fork-choice.md#get_latest_attesting_balance, so skips the added `proposer_boost_correct_head` fork choice test. This allows it to be merged while remaining compatible with the non-proposer-boosted merge devnet.

The `proposer_boost_root` tests which run (phase 0 and mainnet only due to other pre-existing issues) do run and pass, however.

### DIFF
--- a/beacon_chain/fork_choice/fork_choice.nim
+++ b/beacon_chain/fork_choice/fork_choice.nim
@@ -133,8 +133,15 @@ func process_attestation_queue(self: var ForkChoice) {.gcsafe.}
 
 proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: BeaconTime):
     FcResult[void] =
+  const step_size = seconds(SECONDS_PER_SLOT.int)
   if time > self.checkpoints.time:
-    ? on_tick(self.checkpoints, dag, time)
+    # Call on_tick at least once per slot.
+    while time >= self.checkpoints.time + step_size:
+      ? on_tick(self.checkpoints, dag, self.checkpoints.time + step_size)
+
+    if self.checkpoints.time < time:
+      # Might create two ticks for the last slot.
+      ? on_tick(self.checkpoints, dag, time)
 
     self.process_attestation_queue() # Only run if time changed!
 
--- a/beacon_chain/fork_choice/fork_choice.nim
+++ b/beacon_chain/fork_choice/fork_choice.nim
@@ -13,6 +13,7 @@ import
   # Status libraries
   stew/results, chronicles,
   # Internal
+  ../beacon_clock,
   ../spec/[beaconstate, helpers],
   ../spec/datatypes/[phase0, altair, merge],
   # Fork choice
@@ -92,12 +93,25 @@ func extend[T](s: var seq[T], minLen: int) =
 func compute_slots_since_epoch_start(slot: Slot): uint64 =
     slot - slot.epoch().compute_start_slot_at_epoch()
 
-proc on_tick(self: var Checkpoints, dag: ChainDAGRef, time: Slot): FcResult[void] =
+func on_tick*(self: var Checkpoints, time: BeaconTime): FcResult[void] =
   if self.time > time:
     return err ForkChoiceError(kind: fcInconsistentTick)
 
-  let newEpoch = self.time.epoch() != time.epoch()
+  # Reset store.proposer_boost_root if this is a new slot
+  if time.slotOrZero > self.time.slotOrZero:
+    self.proposer_boost_root = default(Eth2Digest)
+
   self.time = time
+
+  ok()
+
+proc on_tick(self: var Checkpoints, dag: ChainDAGRef, time: BeaconTime):
+    FcResult[void] =
+  let prev_time = self.time
+
+  ? self.on_tick(time)
+
+  let newEpoch = prev_time.slotOrZero.epoch() != time.slotOrZero.epoch()
 
   if newEpoch and
       self.best_justified.epoch > self.justified.checkpoint.epoch:
@@ -117,10 +131,10 @@ proc on_tick(self: var Checkpoints, dag: ChainDAGRef, time: Slot): FcResult[void
 
 func process_attestation_queue(self: var ForkChoice) {.gcsafe.}
 
-proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: Slot): FcResult[void] =
+proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: BeaconTime):
+    FcResult[void] =
   if time > self.checkpoints.time:
-    while time > self.checkpoints.time:
-      ? on_tick(self.checkpoints, dag, self.checkpoints.time + 1)
+    ? on_tick(self.checkpoints, dag, time)
 
     self.process_attestation_queue() # Only run if time changed!
 
@@ -153,7 +167,7 @@ func process_attestation*(
 
 func process_attestation_queue(self: var ForkChoice) =
   self.queuedAttestations.keepItIf:
-    if it.slot < self.checkpoints.time:
+    if it.slot < self.checkpoints.time.slotOrZero:
       for validator_index in it.attesting_indices:
         self.backend.process_attestation(
           validator_index, it.block_root, it.slot.epoch())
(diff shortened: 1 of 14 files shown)


tersec:
It doesn't suffice to only render finer the granularity locally in block handling, rather than in the time/tick mechanism for fork choice as a whole. For example:
```
- {tick: 52}
- {block: block_0x2ea951a83b617769a5a4585d0deac3c9cff8b70d2b4458531dcfcb6c53e589bc}
- checks:
    time: 52
    head: {slot: 4, root: '0xfe06154921d61f8af5d3188100b78be2f51ebbe80353dafef3a3ee0c243340f9'}
    justified_checkpoint: {epoch: 0, root: '0x1fcb8c722539b44be817428e8b37e4aa5012ab79f1d1880c1e92fee3ec304ba3'}
    finalized_checkpoint: {epoch: 0, root: '0x1fcb8c722539b44be817428e8b37e4aa5012ab79f1d1880c1e92fee3ec304ba3'}
    best_justified_checkpoint: {epoch: 0, root: '0x1fcb8c722539b44be817428e8b37e4aa5012ab79f1d1880c1e92fee3ec304ba3'}
    proposer_boost_root: '0x0000000000000000000000000000000000000000000000000000000000000000'
- checks: {proposer_boost_root: '0x0000000000000000000000000000000000000000000000000000000000000000'}
```
(`mainnet/phase0/fork_choice/on_block/pyspec_tests/proposer_boost_root_same_slot_untimely_block/steps.yaml`) specifically sets the time, independently, to a non-slot-boundary then applies a block. The fork choice machinery as a whole needs to be able to track this.

It'd be possible not to use `BeaconTime` for `get_head` and `on_attestation`, but then there'd be an internal inconsistency within the fork choice API, which seems worse than using the `BeaconTime` that sometimes already existed at the call-site, but had been converted by `slotOrZero`.

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -117,10 +131,11 @@ proc on_tick(self: var Checkpoints, dag: ChainDAGRef, time: Slot): FcResult[void
 
 func process_attestation_queue(self: var ForkChoice) {.gcsafe.}
 
-proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: Slot): FcResult[void] =
+proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: BeaconTime):
+    FcResult[void] =
   if time > self.checkpoints.time:
     while time > self.checkpoints.time:
-      ? on_tick(self.checkpoints, dag, self.checkpoints.time + 1)
+      ? on_tick(self.checkpoints, dag, time)
```
tersec, line 144: The intent of the code that was there seemed to be to step ahead a slot at a time (when `self.checkpoints.time` was a `Slot`), and the current version doesn't do that.

I'm not sure what would break in `on_tick`, though, if one just skips straight from `self.checkpoints.time` to `time`. `self.proposer_boost_root = default(Eth2Digest)` is idempotent.

```nim
  let newEpoch = prev_time.epoch() != time.epoch()

  if newEpoch and                    
      self.best_justified.epoch > self.justified.checkpoint.epoch:
    let blck = dag.getRef(self.best_justified.root)
```
might behave differently, if fork choice has skipped multiple epoch boundaries, but it's not obvious to me that it's useful to compute new justified checkpoints which will be overwritten in another `SLOTS_PER_EPOCH` slots in the same loop.

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -117,10 +131,10 @@ proc on_tick(self: var Checkpoints, dag: ChainDAGRef, time: Slot): FcResult[void
 
 func process_attestation_queue(self: var ForkChoice) {.gcsafe.}
 
-proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: Slot): FcResult[void] =
+proc update_time(self: var ForkChoice, dag: ChainDAGRef, time: BeaconTime):
+    FcResult[void] =
   if time > self.checkpoints.time:
-    while time > self.checkpoints.time:
-      ? on_tick(self.checkpoints, dag, self.checkpoints.time + 1)
+    ? on_tick(self.checkpoints, dag, time)
```
arnetheduck: I don't quite remember the logic here, but the `time+1` trick ensures that every slot gets an `on_tick` call, rather than one big jump - this _may_ have effects on the kind of checkpoints that get processed

tersec: Alright, worth keeping that behavior, pending further information.

tersec: https://github.com/status-im/nimbus-eth2/pull/3138/commits/73f6c8b0c16084e679a243d2e93f0c4116a0f4c9

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -321,6 +343,16 @@ proc process_block*(self: var ForkChoice,
           attestation.data.beacon_block_root,
           attestation.data.target.epoch)
 
+  # Add proposer score boost if the block is timely
+  let
+    time_into_slot =
+      self.checkpoints.time - self.checkpoints.time.slotOrZero.toBeaconTime
+    is_before_attesting_interval =
+      time_into_slot < (SECONDS_PER_SLOT div INTERVALS_PER_SLOT).int64.seconds
```
arnetheduck: We've collected some related constants here: https://github.com/status-im/nimbus-eth2/blob/0d4e49f9460a91edfc151064aa889178ecab036f/beacon_chain/beacon_clock.nim#L43

might be an idea to reuse them although that's a bit of a deviation from the pure spec code -- constants should probably be defined in terms of `INTERVALS_PER_SLOT` as well

arnetheduck (approved):

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3138/commits/a8dde89b67a410e0962f1f272d7845675b2c2449

Doing this correctly in seconds is is close enough to spec Python code to just use that directly, but doing it in milliseconds or something similarly more granular, to support slots times which aren't integral multiples of `INTERVALS_PER_SLOT`, is verbose enough it's not something I'd want to gamble on being used everywhere otherwise.

The other advantage ensuring consistency in unusual/less-tested scenarios so that some parts of Nimbus don't think attestations happen at 1 second and others at 1.67 seconds.

The `attestationSlotOffset` and similar constants already achieve this by moving the division outside the `Duration` construction, so the division can be as fine-grained as it needs to be -- better not to introduce anything which is inconsistent with that.

I also cleaned up some of the other users of `div 3`, to either use `div INTERVALS_PER_SLOT`, or refer to some other source of said `Duration`/`BeaconTime`.

The validator client has its own set, which I didn't want to immediately collapse into the `beacon_clock` one in case there's some reason it's there separately, e.g., being able to exist in its own world distinct from the BN, but that probably should be consolidated too.

However this is achieved, internal consistency trumps slavishly following spec code peciularities here. It's a non-issue for mainnet and minimal, but we already have at least one request for a non-multiple-of-3 slot time.

arnetheduck: `seconds` converts the time to a `Duration` which internally does nanoseconds - as long as the division happens after conversion, we should get enough precision for the foreseeable future

arnetheduck: > following spec

having `INTERVALS_PER_SLOT` appear _somewhere_ in the code maintains a reasonable audit trail as well, so this LGTM

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -279,15 +279,15 @@ proc attestationValidator*(
       self.batchCrypto, attestation, wallTime, subnet_id, checkSignature)
   return if v.isOk():
     # Due to async validation the wallSlot here might have changed
-    wallSlot = self.getCurrentBeaconTime().slotOrZero()
+    wallTime = self.getCurrentBeaconTime()
```
arnetheduck, line 282: this needs to be updated for aggregates as well, or we get stale times in the pool

tersec: https://github.com/status-im/nimbus-eth2/pull/3230

PR #3144, opened 2021-11-30
Introduce slot->BlockRef mapping for finalized chain

arnetheduck:
The finalized chain is linear, thus we can use a seq to lookup blocks by
slot number.

Here, we introduce such a seq, even though in the future, it should
likely be backed by a database structure instead, or, more likely, a
flat era file with a flat lookup index.

This dramatically speeds up requests by slot, such as those coming from
the REST interface or GetBlocksByRange, as these are currently served by
a linear iteration from head.

* fix REST block requests to not return blocks from an earlier slot when
the given slot is empty
* fix StateId interpretation such that it doesn't treat state roots as
block roots
* don't load full block from database just to return its root

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1338,13 +1338,17 @@ proc updateHead*(
       finalized = shortLog(getStateField(
         dag.headState.data, finalized_checkpoint))
 
-    dag.finalizedBlocks.setLen(finalizedHead.slot.int + 1)
-    var tmp = finalizedHead.blck
-    while not isNil(tmp) and tmp.slot >= dag.finalizedHead.slot:
-      dag.finalizedBlocks[tmp.slot.int] = tmp
-      tmp = tmp.parent
+    block:
+      # Update `dag.finalizedBlocks` with all newly finalized blocks (those
+      # newer than the previous finalized head), then update `dag.finalizedHead`
 
-    dag.finalizedHead = finalizedHead
+      dag.finalizedBlocks.setLen(finalizedHead.slot.int + 1)
+      var tmp = finalizedHead.blck
+      while not isNil(tmp) and tmp.slot >= dag.finalizedHead.slot:
+        dag.finalizedBlocks[tmp.slot.int] = tmp
+        tmp = tmp.parent
+
+      dag.finalizedHead = finalizedHead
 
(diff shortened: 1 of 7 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1318,6 +1338,12 @@ proc updateHead*(
       finalized = shortLog(getStateField(
         dag.headState.data, finalized_checkpoint))
 
+    dag.finalizedBlocks.setLen(finalizedHead.slot.int + 1)
+    var tmp = finalizedHead.blck
+    while not isNil(tmp) and tmp.slot >= dag.finalizedHead.slot:
```
zah: What is the intention of the code here?

It tries to write slots **above** `dag.finalizedHead.slot` to `dag.finalizedBlocks`. Isn't this contradictory?

beacon_chain/rpc/rest_utils.nim
```diff
@@ -40,8 +40,22 @@ proc validate(key: string, value: string): int =
   else:
     1
 
-proc getCurrentHead*(node: BeaconNode,
-                     slot: Slot): Result[BlockRef, cstring] =
+func getCurrentSlot*(node: BeaconNode, slot: Slot):
```
zah, line 43: I find this name quite surprising/misleading.

Also, I would expect that only a very small subset of the API calls will require this precise condition (i.e. they would allow a slot that's 2 epochs in the future). This is usually related to various committee-specific look-head periods in the spec, so the name should reference the specific look-ahead period somehow. The current name implies a much more general purpose and the existing call sites already look suspicious to me.

arnetheduck: I kind of agree but did not want to touch this aspect in this PR - the original motivation was to provide some form of protection against evidently out-of-bounds requests but indeed, it needs per-request refinement because the requirements are not the same. This PR just follows along the status quo.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: these are the blocks that are newer than the _old_ finalized head (but older than the new finalized head) - the next line sets `dag.finalizedHead` to the new finalized head.

arnetheduck: fcbc00769ca06c9922e650fbc4635fd0cdef65b4

PR #3150, opened 2021-12-01
Allow a docker image to be used instead of building nimbus_beacon_node

zah:
This would be useful when testing nimbus-eth1 in merge-mode simulations

### DIFF
--- a/scripts/launch_local_testnet.sh
+++ b/scripts/launch_local_testnet.sh
@@ -21,7 +21,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")"/..
 GETOPT_BINARY="getopt"
 if uname | grep -qi darwin; then
   # macOS
-  GETOPT_BINARY="/usr/local/opt/gnu-getopt/bin/getopt"
+  GETOPT_BINARY=$(find /opt/homebrew/opt/gnu-getopt/bin/getopt /usr/local/opt/gnu-getopt/bin/getopt 2> /dev/null || true)
   [[ -f "$GETOPT_BINARY" ]] || { echo "GNU getopt not installed. Please run 'brew install gnu-getopt'. Aborting."; exit 1; }
 fi
 
@@ -32,7 +32,7 @@ if [ ${PIPESTATUS[0]} != 4 ]; then
 fi
 
 OPTS="ht:n:d:g"
-LONGOPTS="help,preset:,nodes:,data-dir:,with-ganache,stop-at-epoch:,disable-htop,disable-vc,enable-logtrace,log-level:,base-port:,base-rpc-port:,base-metrics-port:,reuse-existing-data-dir,timeout:,kill-old-processes"
+LONGOPTS="help,preset:,nodes:,data-dir:,with-ganache,stop-at-epoch:,disable-htop,disable-vc,enable-logtrace,log-level:,base-port:,base-rpc-port:,base-metrics-port:,reuse-existing-data-dir,timeout:,kill-old-processes,eth2-docker-image:"
 
 # default values
 NUM_NODES="10"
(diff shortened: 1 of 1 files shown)


scripts/launch_local_testnet.sh
```diff
@@ -151,6 +153,14 @@ while true; do
       KILL_OLD_PROCESSES="1"
       shift
       ;;
+    --eth2-docker-image)
+      ETH2_DOCKER_IMAGE="$2"
+      shift 2
+      # TODO The validator client is still not being shipped with
+      #      our docker images, so we must disable it:
```
stefantalpalaru, line 160: When you want to add it, here's where you do it: https://github.com/status-im/nimbus-eth2/blob/5e9625c1be57045f23c8739c727c6257d0ecddac/docker/dist/entry_point.sh#L20

scripts/launch_local_testnet.sh
```diff
@@ -223,11 +228,25 @@ if [[ "${HAVE_LSOF}" == "1" ]]; then
...
 if [[ "$ENABLE_LOGTRACE" == "1" ]]; then
   BINARIES="${BINARIES} logtrace"
 fi
 
+if [[ ! -z "$ETH2_DOCKER_IMAGE" ]]; then
+  DATA_DIR_FULL_PATH=$(cd "${DATA_DIR}"; pwd)
+  # CONTAINER_DATA_DIR must be used everywhere where paths are supplied to BEACON_NODE_COMMAND executions.
+  # We'll use the CONTAINER_ prefix throughout the file to indicate such paths.
+  CONTAINER_DATA_DIR=/home/user/nimbus-eth2/testnet
+  BEACON_NODE_COMMAND="docker run -v /etc/passwd:/etc/passwd -u $(id -u):$(id -g) --net=host -v ${DATA_DIR_FULL_PATH}:${CONTAINER_DATA_DIR}:rw $ETH2_DOCKER_IMAGE"
```
stefantalpalaru, line 242: I think ":rw" is already the default.

scripts/launch_local_testnet.sh
```diff
@@ -245,6 +264,10 @@ cleanup() {
   for BINARY in ${BINARIES}; do
     rm build/${BINARY}
   done
+
+  if [[ ! -z "$ETH2_DOCKER_IMAGE" ]]; then
+    docker rm $(docker stop $(docker ps -a -q --filter ancestor=$ETH2_DOCKER_IMAGE --format="{{.ID}}"))
```
stefantalpalaru, line 269: Can these commands fail and exit the script early?

scripts/launch_local_testnet.sh
```diff
@@ -438,16 +468,16 @@ for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
     done
   fi
 
-  ./build/nimbus_beacon_node \
+  $BEACON_NODE_COMMAND \
     --non-interactive \
     --nat:extip:127.0.0.1 \
-    --network="${DATA_DIR}" \
+    --network="${CONTAINER_DATA_DIR}" \
```
stefantalpalaru, line 474: We end up using "CONTAINER_" variables even when not using a container?

stefantalpalaru (approved):

scripts/launch_local_testnet.sh
zah: Yes, then `CONTAINER_DATA_DIR` is just an alias for `DATA_DIR`.

scripts/launch_local_testnet.sh
zah: These commands fail from time to time for me actually (when the containers were already stopped). They produce benign error messages though and I didn't try to search for alternative solutions.

scripts/launch_local_testnet.sh
zah: Doesn't hurt to be explicit

PR #3151, opened 2021-12-01
batch-verify sync messages for a small perf boost

arnetheduck:
Generally reuses the same structure as attestation and aggregate
verification

* normalize `signatures` and `signature_batch` to use the same pattern
of verification
* normalize parameter names, order etc for signature stuff in general
* avoid calling `blsSign` directly - instead, go through `signatures`
consistently

### DIFF
--- a/beacon_chain/gossip_processing/batch_validation.nim
+++ b/beacon_chain/gossip_processing/batch_validation.nim
@@ -205,9 +205,10 @@ proc scheduleBatch(batchCrypto: ref BatchCrypto, fresh: bool) =
 
 template orReturnErr(v: Option, error: cstring): untyped =
   ## Returns with given error string if the option does not have a value
-  if v.isNone:
+  let tmp = v
+  if tmp.isNone:
     return err(error) # this exits the calling scope, as templates are inlined.
-  v.unsafeGet()
+  tmp.unsafeGet()
 
 template withBatch(
     batchCrypto: ref BatchCrypto, name: cstring,
--- a/beacon_chain/gossip_processing/batch_validation.nim
+++ b/beacon_chain/gossip_processing/batch_validation.nim
@@ -8,6 +8,7 @@
 {.push raises: [Defect].}
 
 import
+  std/sequtils,
   # Status
   chronicles, chronos,
   ../spec/signatures_batch,
@@ -188,6 +189,7 @@ proc getBatch(batchCrypto: ref BatchCrypto): (ref Batch, bool) =
     # len will be 0 when the batch was created but nothing added to it
     # because of early failures
     (batch, batch[].len() == 0)
+
 proc scheduleBatch(batchCrypto: ref BatchCrypto, fresh: bool) =
   if fresh:
     # Every time we start a new round of batching, we need to launch a deferred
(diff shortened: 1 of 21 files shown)


arnetheduck:
i386 fail is spurios

beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -219,12 +221,35 @@ proc scheduleBatch(batchCrypto: ref BatchCrypto, fresh: bool) =
     # If there's a full batch, process it eagerly assuming the callback allows
     batchCrypto.processBatch()
 
+template orReturnErr(v: Option, error: cstring): untyped =
```
zah, line 206: This suffers from double evaluation of `v`

beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -219,12 +221,35 @@ proc scheduleBatch(batchCrypto: ref BatchCrypto, fresh: bool) =
...
+template withBatch(
+    batchCrypto: ref BatchCrypto, name: cstring,
+    body: untyped): Future[BatchResult] =
+  block:
+    let
+      (batch {.inject.}, fresh) = batchCrypto.getBatch()
+
+    body
+
+    let fut = newFuture[BatchResult](name)
```
zah, line 222: @cheatfate, how should cancellation support be implemented here?

beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -278,72 +289,111 @@ proc scheduleAggregateChecks*(
...
-              .pendingBuffer
-              .addAttestation(
-                fork, genesis_validators_root, epochRef,
-                aggregate
-              )
-  if sig.isErr():
-    batchCrypto.scheduleBatch(fresh)
-    return err(sig.error())
+  let
+    aggregatorFut = batchCrypto.withBatch("scheduleAggregateChecks.aggregator"):
```
zah, line 293: The previous code was reusing a single `batch` object created with `batchCrypto.getBatch()` while the new code creates a bunch of them. Isn't this change in behavior/semantics? cc @mratsim

tests/test_sync_committee_pool.nim
```diff
@@ -57,22 +57,19 @@ suite "Sync committee pool":
 
       root1Slot = Slot(100)
       root2Slot = Slot(101)
-      root3Slot = Slot(101)
+      root3Slot = Slot(102)
```
zah: This was using the same slot as `root2Slot` intentionally (this checks more precisely that the various lookups in the sync committee pool don't accidentally use the wrong key).

arnetheduck: hm, I thought that was covered by the sig4 test  - ok, will revert

beacon_chain/gossip_processing/batch_validation.nim
arnetheduck: good catch.. I'm not too happy about this construct in general tbh, might be a good time to start moving towards `results.Opt`, though that would need polishing the results lib first

arnetheduck: 523755d4639f7acade7294878010997bb1b5aa0f

beacon_chain/gossip_processing/batch_validation.nim
arnetheduck: hm, yeah - there's a small difference in that each batch "normally" should contain no more than 16 signatures but with aggregates, it could go over that - here we create separate batches instead with a "hard" cap on 16 signatures each.. there's not much practical difference I suspect: some aggregates will be split over two batches, with the potential for a small delay in return for a more "even" batch size.

tests/test_sync_committee_pool.nim
arnetheduck: 6597c0d1be381193f5f2228378f8e825e780b0b1

beacon_chain/gossip_processing/batch_validation.nim
arnetheduck: another way to put that: we'll actually use the same batch for most aggregates, except when the batch gets full in the middle of an aggregate - then a single aggregate can get split over two batches which I guess is bad for the aggregate and good for whatever was in the batch already

beacon_chain/gossip_processing/batch_validation.nim
arnetheduck: why do we need cancellation here?

ie there exists a timeout mechanism already that purges the queue in a controlled way

zah: @cheatfate generally tries to ensure that all Futures are configured for cancellation. It's a matter of adding a reasonable `cancelCallback` and it's one of the things you usually do after using the `newFuture` API, which is considered "low-level".

arnetheduck: this future is never cancelled, so it doesn't need a cancellation handler - it's manually managed from beginning to end - instead of using the built-in cancellation support, it completes the future with an appropriate error instead.

arnetheduck: it's a good point though - if the future was cancelled, it would crash the application with a "double-complete" defect (both before and after this PR) - I think I'll fix that regardless, so as to avoid surprises

beacon_chain/gossip_processing/batch_validation.nim
mratsim: Clarification, it's not a hard cap but a soft one. For example aggregate attestations checks are added 3 by 3

https://github.com/status-im/nimbus-eth2/blob/f0f9735955070d297b33fbec466537cb7029e0cc/beacon_chain/gossip_processing/gossip_validation.nim#L567-L583
https://github.com/status-im/nimbus-eth2/blob/9e8081e405b4801bd81514289c81c46aefceb1b4/beacon_chain/gossip_processing/batch_validation.nim#L258-L280

So if we have 15 checks in the batch and 3 aggregate attestation checks (for the same agregate) are added, 18 will be processed at once, they won't be split.

PR #3153, opened 2021-12-02
More efficient implementation of the 'POST beacon_committee_subscriptions' API

zah:


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -503,30 +503,38 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
     if not(node.isSynced(node.dag.head)):
       return RestApiResponse.jsonError(Http503, BeaconNodeInSyncError)
 
+    let
+      wallSlot = node.beaconClock.now.slotOrZero
```
zah, line 507: Variables were hoisted outside of the loop.

beacon_chain/validators/action_tracker.nim
```diff
@@ -84,12 +87,11 @@ proc registerDuty*(
   if isAggregator:
     let newDuty = AggregatorDuty(slot: slot, subnet_id: subnet_id)
 
-    for duty in tracker.duties.mitems():
-      if duty == newDuty:
-        return
+    if newDuty in tracker.duties:
```
zah, line 90: Quadratic complexity was eliminated through switching to a `HashSet`.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -503,30 +503,38 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+      wallEpoch = wallSlot.epoch
+      head = node.dag.head
+
+    var currentEpoch, nextEpoch: Option[EpochRef]
+    template getAndCacheEpochRef(epochRefVar: var Option[EpochRef],
+                                 epoch: Epoch): EpochRef =
+      if epochRefVar.isNone:
+        epochRefVar = some node.dag.getEpochRef(head, epoch)
```
arnetheduck, line 515: `getEpochRef` already does the appropriate caching and is fast enough that we call it for every attestation we validate - it's not a bottleneck by any means (I've verified this) - there's no need to repeat it here (in fact, it makes reasoning about their lifetime a lot harder - in particular, we strive not to keep finalized EpochRef instances around, they're quite expensive).

zah: These are local variables in the function. What is the precise scenario that concerns you?

beacon_chain/validators/action_tracker.nim
```diff
@@ -56,11 +56,14 @@ type
...
       ## that no longer are posting duties
 
-    duties*: seq[AggregatorDuty] ##\
+    duties*: HashSet[AggregatorDuty] ##\
       ## Known aggregation duties in the near future - before each such
       ## duty, we'll subscribe to the corresponding subnet to collect
       ## attestations for the aggregate
 
+func hash*(x: AggregatorDuty): Hash =
+  hashData(unsafeAddr x, sizeof(x))
```
arnetheduck: no way.

beacon_chain/rpc/rest_validator_api.nim
arnetheduck: ah ok - thought for a moment they sat in a global section and would stay between calls.

beacon_chain/validators/action_tracker.nim
zah: Replaced with `hashObjectBytes` which uses `typetraits.supportsCopyMem` to ensure that we are dealing with a POD type:
https://github.com/status-im/nim-stew/pull/94/commits/38bb2bd0fec1e5ca8b4b8c9387866ee03bb43be6

beacon_chain/validators/action_tracker.nim
arnetheduck: the hashset is certainly an improvement  - fwiw, there exists an alternative implementation that @tersec was employing before, that was lost in refactoring: instead of storing each duty in a collection one can assign the bit field of subnets to listen to with an `or` operation, so that each duty sets bits for the slot within the epoch and those before - it makes for some ugly code but turns the new duty assignment into a constant operation with bounded memory - in hindsight it probably would have made sense to port that implementation to here.

arnetheduck (approved):

PR #3156, opened 2021-12-03
Broadcast blocks before integrating in ChainDAG 

arnetheduck:
This PR fixes two issues with block publishing:

* Gossip-valid blocks are published before integrating them into the
chain, giving broadcasting a head start, both for rest block and
* Outright invalid blocks from the API that could lead to the descoring
of the node are no longer broadcast

Bonus:

* remove undocumented and duplicated `post_v1_validator_block` JSON-RPC
call

### DIFF
--- a/beacon_chain/rpc/rpc_beacon_api.nim
+++ b/beacon_chain/rpc/rpc_beacon_api.nim
@@ -401,28 +401,18 @@ proc installBeaconApiHandlers*(rpcServer: RpcServer, node: BeaconNode) {.
       )
 
   rpcServer.rpc("post_v1_beacon_blocks") do (blck: phase0.SignedBeaconBlock) -> int:
-    if not(node.syncManager.inProgress):
-      raise newException(CatchableError,
-                         "Beacon node is currently syncing, try again later.")
-    let head = node.dag.head
-    if head.slot >= blck.message.slot:
-      node.network.broadcastBeaconBlock(blck)
-      # The block failed validation, but was successfully broadcast anyway.
-      # It was not integrated into the beacon node's database.
-      return 202
+    let res = await sendBeaconBlock(node, ForkedSignedBeaconBlock.init(blck))
+    if res.isErr():
+      raise (ref CatchableError)(msg: $res.error())
+
+    if res.get():
+      # The block was validated successfully and has been broadcast.
+      # It has also been integrated into the beacon node's database.
+      return 200
     else:
-      let res = await proposeSignedBlock(
-        node, head, AttachedValidator(),
-        ForkedSignedBeaconBlock.init(blck))
-      if res == head:
-        node.network.broadcastBeaconBlock(blck)
-        # The block failed validation, but was successfully broadcast anyway.
-        # It was not integrated into the beacon node''s database.
-        return 202
-      else:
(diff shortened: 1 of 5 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -1246,20 +1246,31 @@ proc sendProposerSlashing*(node: BeaconNode,
...
+      return SendBlockResult.err(res.error()[1])
+
+  # The block passed basic gossip validation - we can "safely" broadcast it now.
+  # In fact, per the spec, we should broadcast it even if it later fails to
+  # apply to our state.
+  node.network.broadcastBeaconBlock(forked)
+
+  let
+    head = node.dag.head
+    res = await node.proposeSignedBlock(head, AttachedValidator(), forked)
```
zah: `proposeSignedBlock` is also broadcasting the block. What's the idea here?

zah: Also, `proposeSignedBlock` should really take a validator public key instead of `AttachedValidator`, so it can print correct data instead of the default all zeros for the validator ID. We know the validator from the `blck.message.proposer_index` field.

arnetheduck: > What's the idea here?

good question :)

arnetheduck: rewrote the flow and fixed a few other things as well - internal blocks also had to go through chaindag before broadcast which is unnecessary - they are now broadcast before being integrated which is a better flow in general

beacon_chain/rpc/rpc_beacon_api.nim
```diff
@@ -401,28 +401,18 @@ proc installBeaconApiHandlers*(rpcServer: RpcServer, node: BeaconNode) {.
...
+      raise (ref CatchableError)(msg: $res.error())
+
+    if res.get():
+      # The block was validated successfully and has been broadcast.
+      # It has also been integrated into the beacon node's database.
+      return 200
+    else:
       # The block failed validation, but was successfully broadcast anyway.
-      # It was not integrated into the beacon node's database.
+      # It was not integrated into the beacon node''s database.
```
arnetheduck, line 414:
```suggestion
      # It was not integrated into the beacon node's database.
```

beacon_chain/rpc/rpc_validator_api.nim
```diff
@@ -53,20 +53,6 @@ proc installValidatorApiHandlers*(rpcServer: RpcServer, node: BeaconNode) {.
     else:
       raiseNoAltairSupport()
 
-  rpcServer.rpc("post_v1_validator_block") do (body: phase0.SignedBeaconBlock) -> bool:
```
zah, line 56: Isn't this a breaking change? Admittedly, it's unlikely that somebody is using this particular API

zah (approved):

beacon_chain/rpc/rpc_validator_api.nim
arnetheduck: we have two implementations of this same call - one in `rpc_validator_api` and one in `rpc_beacon_api` - the latter is documented in the book, this one is not - neither of them _actually_ work on an `altair+` chain, but that's a different story.

PR #3161, opened 2021-12-04
cleanups, partly from kintsugi branch

tersec:
- don't name variables in exception handling if not referenced
- unexport some functions from `eth1_monitor`
- avoid false positive unused-hints by marking some `withFoo`-template-created variables as `used`
- remove obsolete docs section which aligned Geth/Nimbus ports

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -36,8 +36,8 @@ proc batchVerify(quarantine: QuarantineRef, sigs: openArray[SignatureSet]): bool
   quarantine.rng[].brHmacDrbgGenerate(secureRandomBytes)
   try:
     return quarantine.taskpool.batchVerify(quarantine.sigVerifCache, sigs, secureRandomBytes)
-  except Exception:
-    raise newException(Defect, "Unexpected exception in batchVerify.")
+  except Exception as exc:
+    raiseAssert exc.msg
 
 proc addRawBlock*(
       dag: ChainDAGRef, quarantine: QuarantineRef,
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -36,7 +36,7 @@ proc batchVerify(quarantine: QuarantineRef, sigs: openArray[SignatureSet]): bool
   quarantine.rng[].brHmacDrbgGenerate(secureRandomBytes)
   try:
     return quarantine.taskpool.batchVerify(quarantine.sigVerifCache, sigs, secureRandomBytes)
-  except Exception as exc:
+  except Exception:
     raise newException(Defect, "Unexpected exception in batchVerify.")
 
 proc addRawBlock*(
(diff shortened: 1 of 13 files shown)


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -36,7 +36,7 @@ proc batchVerify(quarantine: QuarantineRef, sigs: openArray[SignatureSet]): bool
   quarantine.rng[].brHmacDrbgGenerate(secureRandomBytes)
   try:
     return quarantine.taskpool.batchVerify(quarantine.sigVerifCache, sigs, secureRandomBytes)
-  except Exception as exc:
+  except Exception:
```
arnetheduck: for cases like this, I actually prefer the style `except Exception as exc: raiseAssert exc.msg` because it gives the original message

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -318,12 +318,12 @@ func asEth2Digest*(x: BlockHash): Eth2Digest =
 template asBlockHash(x: Eth2Digest): BlockHash =
   BlockHash(x.data)
 
-func shortLog*(b: Eth1Block): string =
+func shortLog(b: Eth1Block): string =
```
arnetheduck: since `Eth1Block` itself exported, I find that small helpers like `shortLog` or `hash` that exist to make the type palatable to use make more sense to export as well - otherwise you might run into issues with open generic matching where a different overload is chosen based on which modules have been imported.

arnetheduck (approved):

beacon_chain/eth1/eth1_monitor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3161/commits/30ac1379ecd4be5d2b9a21d2aaec71b02c9922b5

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3161/commits/30ac1379ecd4be5d2b9a21d2aaec71b02c9922b5

I left the ones which didn't re-throw exceptions at all (but returned, e.g., an err(foo)) alone.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -355,12 +355,9 @@ proc addBlock*(chain: var Eth1Chain, newBlock: Eth1Block) =
   chain.blocksByHash[newBlock.voteData.block_hash.asBlockHash] = newBlock
   eth1_chain_len.set chain.blocks.len.int64
 
-func hash*(x: Eth1Data): Hash =
+func hash(x: Eth1Data): Hash =
```
arnetheduck, line 358: `hash` is another "small" type-bound function that I think should always be exported, or we'll run into subtle bugs at some point

tersec: https://github.com/status-im/nimbus-eth2/pull/3165

PR #3164, opened 2021-12-06
DAG loading: don't preload summaries in memory

stefantalpalaru:
RSS before and after ditching the block summaries preloading (from 1.56 to 1.11 GiB):

Screenshot 2021-12-06 at 17-32-19 NBC local testnet sim (all nodes) - Grafana

I also measured the DAG loading time, as seen in the logs, in a worst-case scenario (old HDD, dropping kernel caches with `echo 3 > /proc/sys/vm/drop_caches` before each run) and the increase is negligible - from 6m45s to 7m0s.

This is with https://github.com/status-im/nim-chronos/pull/243, before and after ditching the block summaries preloading (from 1.54 GiB to 787 GiB, no kernel cache dropping this time):

Screenshot 2021-12-06 at 17-59-40 NBC local testnet sim (all nodes) - Grafana

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -847,21 +847,23 @@ iterator getAncestors*(db: BeaconChainDB, root: Eth2Digest):
     yield res
     root = res.message.parent_root
 
-proc loadSummaries(db: BeaconChainDB): Table[Eth2Digest, BeaconBlockSummary] =
-  # Load summaries into table - there's no telling what order they're in so we
-  # load them all - bugs in nim prevent this code from living in the iterator.
-  var summaries = initTable[Eth2Digest, BeaconBlockSummary](1024*1024)
+proc loadSummary(db: BeaconChainDB, k: openArray[byte]): Result[BeaconBlockSummary, void] =
+  # Load summary from the db. Nim bugs prevent this code from living in the
+  # iterator.
+  var
+    res: BeaconBlockSummary
+    success = false
 
-  discard db.summaries.find([], proc(k, v: openArray[byte]) =
-    var output: BeaconBlockSummary
-
-    if k.len() == sizeof(Eth2Digest) and decodeSSz(v, output):
-      summaries[Eth2Digest(data: toArray(sizeof(Eth2Digest), k))] = output
+  let dbRes = db.summaries.get(k, proc(v: openArray[byte]) {.nimcall.} =
+    if decodeSSz(v, res):
+      success = true
     else:
-      warn "Invalid summary in database", klen = k.len(), vlen = v.len()
+      warn "Invalid summary in database", vlen = v.len()
   )
-
-  summaries
+  if dbRes.get(false) and success:
+    ok(res)
+  else:
+    err()
 
 type RootedSummary = tuple[root: Eth2Digest, summary: BeaconBlockSummary]
 iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
@@ -874,11 +876,7 @@ iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
   # Summaries are loaded from the dedicated summaries table. For backwards
   # compatibility, we also load from `kvstore` and finally, if no summaries
   # can be found, by loading the blocks instead.
-
-  # First, load the full summary table into memory in one query - this makes
-  # initial startup very fast.
   var
-    summaries = db.loadSummaries()
     res: RootedSummary
     blck: phase0.TrustedSignedBeaconBlock
     newSummaries: seq[RootedSummary]
@@ -907,10 +905,10 @@ iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
   # If a summary is missing, try loading it from the older version or create one
   # from block data.
   while true:
-    summaries.withValue(res.root, summary) do:
-      res.summary = summary[]
+    if (let summaryRes = loadSummary(db, res.root.data); summaryRes.isOk()):
+      res.summary = summaryRes.value
       yield res
-    do: # Summary was not found in summary table, look elsewhere
+    else: # Summary was not found in summary table, look elsewhere
       if db.v0.backend.getSnappySSZ(subkey(BeaconBlockSummary, res.root), res.summary) == GetResult.found:
         yield res
       elif db.v0.backend.getSnappySSZ(


arnetheduck:
This needs testing on more realistic hardware and conditions (SSD) where "typical" dag load times as measured (including state replay, loading validator keys and other work unrelated to `loadSummaries`) hover around 10-20s.

In particular, a slot is12s - an increase of 15s is not "negligible" from that point of view. 

It's important to remember that the in-memory cache represents a deliberate tradeoff that was made at the time of introduction: use more memory (and release it afterwards) to make startup faster, such that a restart after upgrade can be done without losing attestation work. It is critical to keep these times low.

stefantalpalaru:
SSD over SATA:

```text
Before:

INF 2021-12-07 00:44:53.781+01:00 Loading block dag from database            topics="beacnde" path=build/data/shared_prater_0/db
INF 2021-12-07 00:45:03.647+01:00 State replayed                             topics="chaindag" startSlot=1860192 slots=10 replayDur=1s739ms804us652ns blocks=10 stateRoot=dcdd3f0f stateSlot=1860202 blck=c0f902ff:1860202 found=true assignDur=4us79ns startRoot=2f2feb58

After:

INF 2021-12-07 00:52:30.184+01:00 Loading block dag from database            topics="beacnde" path=build/data/shared_prater_0/db
INF 2021-12-07 00:52:53.521+01:00 State replayed                             topics="chaindag" startSlot=1860192 slots=10 replayDur=1s604ms325us824ns blocks=10 stateRoot=dcdd3f0f stateSlot=1860202 blck=c0f902ff:1860202 found=true assignDur=4us175ns startRoot=2f2feb58
```

So it goes from 10s to 23s - perfectly acceptable trade-off, specially on something like a Raspberry Pi 4.

arnetheduck:
The PR is great as a prototype showing what the memory usage can look like with a more efficient `loadSummaries` implementation, but at the same time it's important to remember that the memory actually isn't "lost" - the OS can reuse it under pressure by reclaiming it and thus decreasing the RSS.

I suspect that on a PI, the difference will be greater, given it's weak CPU - what is done here is turning a linear scan over the database into ~2 million individual lookups, and this change turns the whole thing into a CPU-bound operation.

With a 10s startup time, you can restart within a slot - with 23s, you no longer can - this time will also keep growing as the chain gets longer. The restart times are important so as to keep attestation loss at a minimum, specially when multiple validators are run on a single beacon node and you don't have large "quiet" gaps - we have users depending on this.

There are 3 ways to go from here:

* improve the implementation such that it has neither the memory nor the CPU usage cost - this is quite possible and can be achieved by loading the summaries into a `seq[BlockRef]` and a spill table for conflicts (conflicts happen when abandoned forks have been stored in the database - these are rare, therefore a spill table will not use as much memory - #3144 started putting in place the infrastructure needed for a solution like this
* investigate if it makes sense to add an index in the summaries table such that the summaries can be loaded already-sorted-by-slot instead of in some random hash order
* follow the development plan for era files which obsoletes `loadSummaries` altogether (basically, `loadSummaries` was developed as a temporary stopgap solution to enable fast restarts, until era files are done)

stefantalpalaru:
> the memory actually isn't "lost" - the OS can reuse it under pressure by reclaiming it and thus decreasing the RSS

No, it cannot reclaim it, because the Nim GC won't free it. At best it can swap it, but even that depends on whether it's sprinkled with active GC cells that get scanned regularly.

> With a 10s startup time, you can restart within a slot

We have a 2 epoch pause at startup, by default, as part of "Doppelganger detection". Few people know about it. Even fewer disable it.

> this time will also keep growing as the chain gets longer.

Better than having RAM usage double in a year, due to the natural increase in the number of block summaries. This memory preloading scales worse than the naive solution.

> There are 3 ways to go from here

Sure, but let's do that after we let people optimise RAM usage, by merging this PR.

zah:
> improve the implementation such that it has neither the memory nor the CPU usage cost 

If it's possible to further optimize this change before merging it, this should be explored. While shipping a release with the new memory improvements would certainly be impressive, it's fair to say that nobody has a pressing problem caused by the current overuse of memory, while the increased start-up time could be perceived as a regression. The `unstable` branch is supposed to be shippable at all times, so it would be better if regressions are not introduced even temporarily.

stefantalpalaru:
>  it's fair to say that nobody has a pressing problem caused by the current overuse of memory

On my VPS with 4 GiB of RAM, I need a daily cron job restarting the beacon node to limit memory usage. Even so, before compiling a new version of the software - with just one Make job - I need to manually restart the running node to reduce its memory usage and avoid the OOM killer shutting down my build process.

(I do a full build, from scratch, as part of a Gentoo package)

>  the increased start-up time could be perceived as a regression

Or as a trade-off: half the RAM usage for 13 more seconds of start-up delay.

tersec:
One can buy more RAM more easily than more IOPS, past some point.

arnetheduck:
> On my VPS with 4 GiB of RAM

Our hardware requirements list 4gb as a minimum requirement - this implies the beacon node having access to these 4gb under normal running conditions - it's hard to reason about requirements if you also have to consider random build activity on the server itself - for these kinds of unusual ways of running the node, one can always enable swapping (which indeed we do in our raspberry pi guide, for example) or use a binary build - indeed, swapping is how you let the system deal with the inefficiency of the Nim GC. 

It's a cherry on top that we consume less, and we should of course continue looking for ways to lower consumption, also because it gives more security margin.

As noted, the change of making `loadSummaries` more memory-efficient is a good change in general - before merging, we just need to address the CPU regression, and there are simple ways of doing that as highlighted above.

arnetheduck:
There is now one slightly easier way to implement this that aligns well with what era files do: in chaindag, we've started tracking `ChainDAGRef.finalizedBlocks` in a `seq` - we _could_ persist this seq  in the database across restarts - it's cheap, which would speed up startup even more than the current block summary approach, _and_ it would use less memory - era files carry this index natively, so implementing it in the database as well would give the two formats "feature parity" more or less making it easier to create an abstraction that works with either storage

arnetheduck:
```
with RECURSIVE
  next(k, v) as (
    select `key`, value from beacon_block_summaries 
    where `key` == x'2d6c85483bc6c330d6b9cd6faa8e26469308aa74200245c33c52737811719892'

  UNION ALL
     select `key` as k, value as v from beacon_block_summaries, next 
    where `key` == substr(v, 9, 32)
)
select hex(k), hex(v) from next;
```

another option: single query to load all summaries

arnetheduck:
> another option: single query to load all summaries

nope: 4x slower than table variant (7.5s vs 1.9s)

on the bright side: dbseq version much faster (0.3s vs 1.9s on mainnet) (!)

zah:
Since alternative solutions are in development, I'm closing this PR.

PR #3165, opened 2021-12-07
Merge BeaconBlock gossip validation

tersec:


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -298,6 +299,66 @@ proc validateBeaconBlock*(
...
+       wallTime: BeaconTime, flags: UpdateFlags):
+       Result[void, ValidationError] =
+  ? dag.validateBeaconBlockAux(quarantine, signed_beacon_block, wallTime, flags)
+
+  template blck: auto = signed_beacon_block.message
+
+  # If the execution is enabled for the block -- i.e.
+  # is_execution_enabled(state, block.body) then validate the following:
+  let parentRef = dag.getRef(blck.parent_root)
+  doAssert not parentRef.isNil  # already checked in validateBeaconBlockAux
```
arnetheduck: return the parentref from aux then?

tersec: Saves one `Table` lookup while making the non-merge case slightly more complicated. Reasonable enough, though not an obvious win.

tersec: Also, conceptually, to the extent it's consistent with reasonable performance, I'd like to keep all these gossip validation functions as close to (input) -> (basically a boolean-ish result) as feasible.

arnetheduck: I see it more as a guarantee that parent_root has been checked, than a performance fix - basically a way to remove the `doAssert` - to the point that perhaps it's better to remove the "aux" construct altogether and rely on local "checked" variables?

tersec: All but literally one of the checks for merge is identical to phase0/altair, so this factors things out so that merge noise doesn't come into the phase0/altair validation, but it doesn't repeat the latter codewise. The aux construct seems to be the cleanest way of achieving that.

arnetheduck: what about flipping it? ie make the merge check the "aux" function?

arnetheduck: with noop template overloads for earlier forks?

tersec: https://github.com/status-im/nimbus-eth2/pull/3165/commits/7c4dbfaf622e84b0cc059ec3ce34534a5ecb982e

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -153,7 +153,15 @@ func init*(
...
       finalized_checkpoint: getStateField(state.data, finalized_checkpoint),
       shuffled_active_validator_indices:
-        cache.get_shuffled_active_validator_indices(state.data, epoch)
+        cache.get_shuffled_active_validator_indices(state.data, epoch),
+      merge_transition_complete:
+        case state.data.kind:
+        of BeaconStateFork.Phase0: false
+        of BeaconStateFork.Altair: false
+        of BeaconStateFork.Merge:
+          # https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/beacon-chain.md#is_merge_transition_complete
```
arnetheduck, line 162: lol, wow that's ugly

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -171,10 +166,59 @@ template checkedReject(error: ValidationError): untyped =
...
+  # is_execution_enabled(state, block.body) then validate the following:
+  let executionEnabled =
+    if signed_beacon_block.message.body.execution_payload !=
+        default(ExecutionPayload):
+      true
+    elif dag.getEpochRef(parent_ref, parent_ref.slot.epoch).merge_transition_complete:
+      # Should usually be inexpensive, but could require cache refilling
+      true
+    else:
+      # Somewhat more expensive fallback, with database I/O, but should be
```
arnetheduck, line 191: comment obsolete based on what goes on below?

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -432,8 +475,7 @@ proc validateAttestation*(
         return checkedReject(deferredCrypto.error)
 
       # Await the crypto check
-      let
-        (cryptoFut, sig) = deferredCrypto.get()
+      let (cryptoFut, sig) = deferredCrypto.get()
```
arnetheduck: many of these little seemingly harmless edits conflict with https://github.com/status-im/nimbus-eth2/pull/3165 that needs to change this code because it deals with validation

arnetheduck (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: I considered a number of alternative options here, for example:

- always hit the previous block (too much database I/O);
- store a bool or similar in a `BlockRef` (memory issues, plus database schema effects because they're initialized both from disk and from incoming blocks); or
- use the head state in the DAG, which works well in the straight-line everything-working nicely, but this code has to check the incoming block's parents even if it might be a block on a branch other than the current head's branch. I'm wary of these works-in-good-cases-has-potentially-pathological-effects-in-suboptimal-cases optimizations though.

Another thing I chose not to do based on the last concern, but might make sense, and at least is ... less ugly .. is to have a single flag in the chaindag with this information about the finalized block. But that has similar, albeit not as bad, potentially pathological behavior in suboptimal cases as the use-headstate appraoch does.

This part of the PR was easily the most delicate and nuanced. The actual timestamp consistency check is trivial.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: Links to this PR itself?

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: But I found this approach to be the least bad, overall.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: No? There's a distinct hierarchy here.

Checking for whether the current block has non-default `ExecutionPayload` is trivial and deterministic.

The "usually inexpensive" version mostly will be not much slower (just a lookup in a different in-memory table), but can become reasonably expensive, just not with high probability. Worth making the code more complicated to avoid in the usual post-merge non-default-`ExecutionPayload` case.

The "Somewhat more expensive fallback" always hits the database. SQLite might have cached it, or it might be in some disk cache, but it's 100% a database query.

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: oops, https://github.com/status-im/nimbus-eth2/pull/3151

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: oh, I meant the spec is ugly.

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: ah, I missed the loading of the blockdata.. but .. `is_execution_enabled(state, block.body)`, is that in the context of the block being validated or the parent?

tersec: It's of the state to which the the block applies. Nimbus doesn't necessarily have that state here, and rewinding is expensive enough I never seriously considered using that approach.

However, `process_execution_payload` renders checking whether the state-to-which-`block` applies has a nondefault `ExecutionPayload` equivalent to checking whether the parent block has a nondefault execution payload.

tersec: One intuition: non-default `ExecutionPayload` can be turned on, but once it's on, it's never turned off again. So the game is just to find the non-default-`ExecutionPayload` horizon. And it must be block-precise.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3165/commits/df896f99f512a182204d8c69e2e066d06d29fc0b

PR #3167, opened 2021-12-07
cleanup sync subcommittee accessors

etan-status:
This removes some dead code from `getSubcommitteePositionsAux` which is
no longer needed since the introduction of `SyncCommitteeCache`.
This also cleans up some formatting, uses `let` instead of `var` where
possible, and uses implicit `pairs` in one case for consistency.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1064,9 +1064,8 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
 iterator syncSubcommittee*(
     syncCommittee: openArray[ValidatorIndex],
     subcommitteeIdx: SyncSubcommitteeIndex): ValidatorIndex =
-  var
-    i = subcommitteeIdx.asInt * SYNC_SUBCOMMITTEE_SIZE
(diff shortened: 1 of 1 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1105,19 +1103,15 @@ func getSubcommitteePositionsAux(
...
     subcommitteeIdx: SyncSubcommitteeIndex,
     validatorIdx: uint64): seq[uint64] =
-  let validatorKey = dag.validatorKey(validatorIdx)
-  if validatorKey.isNone():
-    return @[]
-  let validatorPubKey = validatorKey.get().toPubKey
-
-  for pos, key in toSeq(syncCommittee.syncSubcommittee(subcommitteeIdx)):
-    if validatorIdx == uint64(key):
+  for pos, valIdx in toSeq(syncCommittee.syncSubcommittee(subcommitteeIdx)):
```
zah: The `toSeq` call here looks unnecessary. In pointlessly copies the iterator results in a sequence, just so it can iterate over them again. `pos` can be maintained as a regular variable incremented within the loop.

zah (approved):

PR #3169, opened 2021-12-07
use SSE3 instead of SSSE3

tersec:
If it proves on balance beneficial due to a long tail of AMD CPUs which don't support SSSE3, and that it's challenging to communicate for end-user distribution channels the distinctions.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -69,16 +69,11 @@ if defined(windows):
 # engineering a more portable binary release, this should be tweaked but still
 # use at least -msse2 or -msse3.
 #
-# https://github.com/status-im/nimbus-eth2/blob/stable/docs/cpu_features.md#ssse3-supplemental-sse3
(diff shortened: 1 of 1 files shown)


zah:
Ideally, before merging such PRs we would have various benchmark results (e.g. before/after block_sim, ncli_db bench, sync speed, steady state CPU load, etc). The more we have, the merrier.

tersec:
I'd been relying on @mratsim's benchmarks at https://github.com/status-im/nimbus-eth2/blob/stable/docs/cpu_features.md for such purposes.

arnetheduck:
one option might be to enable SSSE3, but to test the CPU on startup and link to our build-yourself-guide if the extension is missing - this particular extension has been around for _quite_ long, after all.

arnetheduck:
On such hardware, one could say that `-march=native` is even more important.

tersec:
Evidently, SSSE3 had been enabled even in already-shipping releases due to BLST automatically doing so.

PR #3171, opened 2021-12-07
Backfill support for ChainDAG

arnetheduck:
In the ChainDAG, 3 block pointers are kept: genesis, tail and head. This
PR adds one more block pointer: the backfill block which represents the
block that has been backfilled so far.

When doing a checkpoint sync, a random block is given as starting point
- this is the tail block, and we require that the tail block has a
corresponding state.

When backfilling, we end up with blocks without corresponding states,
hence we cannot use `tail` as a backfill pointer - there is no state.

Nonetheless, we need to keep track of where we are in the backfill
process between restarts, such that we can answer GetBeaconBlocksByRange
requests.

This PR adds the basic support for backfill handling - it needs to be
integrated with backfill sync, and the REST API needs to be adjusted to
take advantage of the new backfilled blocks when responding to certain
requests.

Future work will also enable moving the tail in either direction:
* pruning means moving the tail forward in time and removing states
* backwards means recreating past states from genesis, such that
intermediate states are recreated step by step all the way to the tail -
at that point, tail, genesis and backfill will match up.
* backfilling is done when backfill != genesis - later, this will be the
WSS checkpoint instead

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -137,6 +137,9 @@ type
       ## only recent contract state data (i.e. only recent `deposit_roots`).
     kHashToStateDiff # Obsolete
     kHashToStateOnlyMutableValidators
+    kBackfillBlock
+      ## Pointer to the earliest block that we have backfilled - if this is not
+      ## set, backfill == tail
 
   BeaconBlockSummary* = object
     ## Cache of beacon block summaries - during startup when we construct the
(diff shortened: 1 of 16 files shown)


tersec (approved):

beacon_chain/beacon_chain_db.nim
```diff
@@ -588,6 +591,9 @@ proc putTailBlock*(db: BeaconChainDB, key: Eth2Digest) =
 proc putGenesisBlock*(db: BeaconChainDB, key: Eth2Digest) =
   db.keyValues.putRaw(subkey(kGenesisBlock), key)
 
+proc putBackfillBlock*(db: BeaconChainDB, key: Eth2Digest) =
```
tersec, line 594: At some point, useful to have tests for `putBackfillBlock`/`getBackfillBlock`.

arnetheduck: We test `put*Block` indirectly via the reload-from-db tests (`dag2 = ...`) like so: https://github.com/status-im/nimbus-eth2/pull/3171/files#diff-2682c6ec19ad99423c71dd419f7584d4cd76b64231a6bdadb969d524e4a1eeebR683

PR #780, opened 2020-03-04
Bls v0.10.1

mratsim:
This updates the BLS implementation to the Eth2 spec v0.10.1 and should unblock the remaining  work on naive aggregation #497 and honest validator #758.

At the moment:
- it updates crypto.nim and all call to bls.Sign, bls.Verify. In particular the API changed, we don't pass privkey + hash_tree_root(data) + domain, we compute the domain, then derive the signing root from data + domain and then sign with privkey+signing root.
  This was made mechanically for the block pool / validator pool and the mocking procedures in testing and checked against v0.10.1 for the spec parts (i.e. part of the validator_pool got specced out in the honest validator document, proc name to be updated)
- BLS test vectors against the EF are not present, it would just be copy-pasting this file (possibly with the crypto.nim wrappers): https://github.com/status-im/nim-blscurve/blob/master/tests/eth2_vectors.nim. This can be done at the same time as the v0.10.2 test vectors that will have fixes for https://github.com/status-im/nim-blscurve/pull/36#issuecomment-586913678
- The test_interop.nim files has been deactivated from testing since there is no updated interop state for v0.10 (either the mocked start https://github.com/ethereum/eth2.0-pm/tree/master/interop/mocked_start or Zcli/Zrnt) and the deposit signatures after this PR are incompatible with v0.9.x deposit signatures.
- `Domain` is now defined in nim-beacon-chain and not in blscurve repo
- The BLS bench update is left as a future TODO

TODO:
- understanding this stack smashing
  image
- fixing CI (it didn't run yet)
- Check if infinity points/zero PublicKeys are still false positive https://github.com/status-im/nim-beacon-chain/blob/4d6918941ba13cbb000cb9e3edf227e4b377bd7c/beacon_chain/spec/crypto.nim#L142-L148

PR #3175, opened 2021-12-09
remove disabled incorrect attestation test

etan-status:
In #780 a test was disabled that verified that an attestation with
empty `aggregation_bits` completes successfully. The test was never
re-introduced, and as of the current consensus spec v1.1.6, such
attestations are not considered valid, as they fail the check in
`is_valid_indexed_attestation`. This patch fully removes that outdated
test, and moves it to the list of pending invalid attestation tests.


arnetheduck:
I suspect there's not much value in these tests at all - ie the upstream test suites should cover it - cc @mratsim

PR #3176, opened 2021-12-09
BlockId reform

arnetheduck:
Introduce `BlockId` that helps track a root/slot pair - this prepares
the codebase for backfilling and handling out-of-dag blocks

* move block dag code to separate module
* fix finalised state root in REST event stream
* fix finalised head computation on head update, when starting from
checkpoint
* clean up chaindag init
* revert `epochAncestor` change in introduced in #3144 that would return
an epoch ancestor from the canoncial history instead of the given
history, causing `EpochRef` keys to point to the wrong block

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -191,7 +191,7 @@ func epochAncestor*(blck: BlockRef, epoch: Epoch): EpochKey =
   ## during the epoch. Thus, last block in the ancestor epochs is the block
   ## that has an impact on epoch currently considered.
   ##
-  ## This function returns an epoh key pointing to that epoch boundary, ie the
+  ## This function returns an epoch key pointing to that epoch boundary, i.e. the
   ## boundary where the last block has been applied to the state and epoch
   ## processing has been done.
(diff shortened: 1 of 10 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -275,26 +185,29 @@ func getBlockBySlot*(dag: ChainDAGRef, slot: Slot): BlockSlot =
...
 
-func epochAncestor*(dag: ChainDAGRef, blck: BlockRef, epoch: Epoch): EpochKey =
+func epochAncestor*(blck: BlockRef, epoch: Epoch): EpochKey =
   ## The state transition works by storing information from blocks in a
   ## "working" area until the epoch transition, then batching work collected
   ## during the epoch. Thus, last block in the ancestor epochs is the block
   ## that has an impact on epoch currently considered.
   ##
-  ## This function returns a BlockSlot pointing to that epoch boundary, ie the
+  ## This function returns an epoh key pointing to that epoch boundary, ie the
```
tersec:
```suggestion
  ## This function returns an epoch key pointing to that epoch boundary, i.e. the
```

PR #3182, opened 2021-12-10
import `is_valid_merkle_branch` test cases from `nim-eth`

etan-status:
As of https://github.com/status-im/nim-eth/pull/379 `nim-eth` defines a
couple static test cases for merkle proof verification.
Since the EF has defined a `is_valid_merkle_branch` function in the spec
we are no longer using the custom implementation from `nim-eth`, but the
tests were never ported to target the new implementation. This patch now
follows up on that and integrates those tests from `nim-eth`.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -290,9 +290,10 @@ OK: 1/1 Fail: 0/1 Skip: 0/1
 + get_helper_indices                                                                         OK
 + get_path_indices                                                                           OK
 + integer_squareroot                                                                         OK
++ is_valid_merkle_branch                                                                     OK
 + verify_merkle_multiproof                                                                   OK
 ```
-OK: 6/6 Fail: 0/6 Skip: 0/6
+OK: 7/7 Fail: 0/7 Skip: 0/7
 ## Specific field types
 ```diff
 + root update                                                                                OK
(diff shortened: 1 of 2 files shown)


kdeme (approved):
lgtm, it would be nice however if the non beacon chain specific part of this helper code and tests could live in nim-ssz-serialization repository. Basically resolving this issue: https://github.com/status-im/nim-ssz-serialization/issues/4 (It is the only thing that is keeping the ssz duplication in nim-eth still).

etan-status:
The `isValidProof` function is not being used outside `nim-eth`: https://github.com/search?q=extension%3Anim+isValidProof
Moving the test here will allow deletion of the SSZ code from `nim-eth`: https://github.com/status-im/nim-eth/pull/447

etan-status:
Agree though that having the `is_valid_merkle_proof` logic (as well as multi merkle proof logic) in `nim-ssz-serialization` would make sense. There are still a handful of open PRs regarding multiproofs, I'll revisit moving the verification logic over once those are complete.

etan-status:
Pulled this ahead, as reviews on https://github.com/status-im/nim-ssz-serialization/pull/16 seem to take longer than expected: https://github.com/status-im/nim-ssz-serialization/pull/19

etan-status:
The `is_valid_merkle_branch` implementation has been moved to `nim-ssz-serialization`: https://github.com/status-im/nim-ssz-serialization/pull/19

PR #3186, opened 2021-12-13
ShortAgent is set without libp2p_agents_metrics

Menduist:
libp2p_agents_metrics is only used for filtering at the
metrics level, but shortAgent is set even without
-d:libp2p_agents_metrics


Menduist:
Oh, actually I just realized that's not the case, but it should be.

I'll fix in libp2p

PR #3194, opened 2021-12-14
bump nim-ssz-serialization to `3cd8d2d6b80bde0ce7f25609cb5cb9fc37852fe2`

etan-status:
This updates `nim-ssz-serialization` to
`3cd8d2d6b80bde0ce7f25609cb5cb9fc37852fe2`.

Notable changes:
- Serialization of object variant (case object) to/from SSZ Union.
- int -> int64 fix in hashTreeRootCached


tersec:
Does this imply that `SingleMemberUnion` (https://github.com/status-im/nim-ssz-serialization/search?q=SingleMemberUnion&type=) is obsolete/fully replaced?

I quickly implemented it when the merge specs required it for a couple of version, but they dropped it, so it has no users anymore. But I didn't remove it from `nim-ssz-serialization` because it seems likely they'll require some kind of unions again.

etan-status:
I don't think it is fully replaced yet. Merkleization code does not support them yet, only SSZ code has it. 
The message I quoted is just the commit title of upstream repo.

kdeme:
> I don't think it is fully replaced yet.

Indeed it's not. And it still needs the `isUnion(T)` checker also. So consider it "experimental".
It might also deem a bit more clumsy in its usage, especially if it is for just a single type.

PR #3199, opened 2021-12-15
forkchoiceUpdate support

tersec:
https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.5/src/engine/specification.md#engine_forkchoiceupdatedv1

Part of the merge.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -18,6 +18,7 @@ import
   json_serialization/std/[options, sets, net], serialization/errors,
   eth/db/kvstore,
   eth/keys, eth/p2p/discoveryv5/[protocol, enr],
+  web3/ethtypes,
 
   # Local modules
   ../spec/datatypes/[phase0, altair, merge],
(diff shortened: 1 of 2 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -416,6 +417,24 @@ proc getBlockProposalEth1Data*(node: BeaconNode,
...
+                        head_block_hash: Eth2Digest,
+                        finalized_block_hash: Eth2Digest,
+                        fee_recipient: ethtypes.Address,
+                        execution_engine: Web3DataProviderRef):
+                        Future[Option[merge.PayloadId]] {.async.} =
+  let
+    timestamp = compute_timestamp_at_slot(state, state.slot)
+    random = get_randao_mix(state, get_current_epoch(state))
+    payloadId =
+      (await execution_engine.forkchoiceUpdated(
```
arnetheduck, line 430: where do we handle exceptions for these calls?

tersec: In their callers, e.g.
https://github.com/status-im/nimbus-eth2/blob/6893c6344b3437d515d0d8256e76f5955d10c29a/beacon_chain/gossip_processing/block_processor.nim#L272-L291
I've found it useful to be able to decide how to interpret error conditions at a relatively outer layer of processing, rather than hide them in the raw RPC calling functions, because then one can treat a timeout, or certain errors, say, as equivalent to `SYNCING`, while others might be `INVALID`. This kind of policy isn't always consistent across callers (e.g., fcU has two callers: the block processor one can fail or time out, and it's mostly fine, but the validator duties one cannot) to bake into the lower-level functions.

arnetheduck: hm, we should probably condense those errors to an enum that represents the meaningful / actionable options in the "outer" layers

arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
tersec: There's a distinction here to be made, but (potentially distinct sets of) enums might be appropriate for both:

1) the "VALID" / "SYNCING" / "INVALID" possibilities, specified by the engine API specification. This is defined in `nim-web3`. There was some issue with these, so in the interest of expediency, I reverted that field to a string. But one blocker for me in merging more of the `kintsugi` branch is figuring out what happened there, and properly using an enum again. This, however, shouldn't be adorned beyond what the specification prescribes, but rather mirror it.

2) Separately, there's a set of states that amounts to a superset of (1), plus some possible error conditions either contacting the EL or that the EL might return in addition to or instead of "VALID" / "SYNCING" / "INVALID". If one wants to move timeout, exception, and other error handling into the inner functions, these would need to be characterized.

PR #3205, opened 2021-12-17
Apple M1: proper "-march=native" substitute

stefantalpalaru:


cskiraly:
I had the same problem on MacBook Pro M1 (not here but in waku and dagger builds). Found the same solution and it worked also for the MacBook. I've seen someone mentioning also an "-mcpu=apple-m1" option, but it wasn't available on my version of clang (12.5). Maybe in newer versions it is there. Anyway, it was said a14 and m1 options are the same under the hood.

cskiraly:
support landed in LLVM Clang 15.0
see https://github.com/llvm/llvm-project/commit/fcca10c69aaab539962d10fcc59a5f074b73b0de

PR #3209, opened 2021-12-17
Trusted node sync

arnetheduck:
Trusted node sync, aka checkpoint sync, allows syncing the chain from a
trusted node instead of relying on a full sync from genesis.

Features include:

* sync from any slot, including the latest finalized slot
* backfill blocks either from the REST api (default)

Future improvements:

* top up blocks between head in database and some other node - this
makes for an efficient backup tool
* recreate historical state to enable historical queries

* ~~use common forked block/state reader in REST API~~ https://github.com/status-im/nimbus-eth2/pull/3232
* ~~disable JSON state readed to avoid the risk of stack overflows~~ https://github.com/status-im/nimbus-eth2/pull/3232
* Backfilling moved to  #3263

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -37,7 +37,7 @@ const
   defaultSigningNodeRequestTimeout* = 60
 
 type
-  BNStartUpCmd* = enum
+  BNStartUpCmd* {.pure.} = enum
     noCommand
     createTestnet
     deposits
@@ -45,6 +45,7 @@ type
     record
     web3
     slashingdb
+    trustedNodeSync
 
   WalletsCmd* {.pure.} = enum
     create  = "Creates a new EIP-2386 wallet"
@@ -177,9 +178,9 @@ type
 
     case cmd* {.
       command
-      defaultValue: noCommand }: BNStartUpCmd
+      defaultValue: BNStartUpCmd.noCommand }: BNStartUpCmd
 
-    of noCommand:
+    of BNStartUpCmd.noCommand:
       bootstrapNodes* {.
         desc: "Specifies one or more bootstrap nodes to use when connecting to the network"
         abbr: "b"
@@ -417,7 +418,7 @@ type
         defaultValue: false
         name: "validator-monitor-totals" }: bool
 
-    of createTestnet:
+    of BNStartUpCmd.createTestnet:
       testnetDepositsFile* {.
         desc: "A LaunchPad deposits file for the genesis state validators"
         name: "deposits-file" }: InputFile
@@ -451,7 +452,7 @@ type
         desc: "Output file with list of bootstrap nodes for the network"
         name: "output-bootstrap-file" }: OutFile
 
-    of wallets:
+    of BNStartUpCmd.wallets:
       case walletsCmd* {.command.}: WalletsCmd
       of WalletsCmd.create:
         nextAccount* {.
@@ -484,7 +485,7 @@ type
       of WalletsCmd.list:
         discard
 
-    of deposits:
+    of BNStartUpCmd.deposits:
       case depositsCmd* {.command.}: DepositsCmd
       of DepositsCmd.createTestnetDeposits:
         totalDeposits* {.
@@ -543,7 +544,7 @@ type
           name: "epoch"
           desc: "The desired exit epoch" }: Option[uint64]
 
-    of record:
+    of BNStartUpCmd.record:
       case recordCmd* {.command.}: RecordCmd
       of RecordCmd.create:
         ipExt* {.
@@ -573,7 +574,7 @@ type
           desc: "ENR URI of the record to print"
           name: "enr" .}: Record
 
-    of web3:
+    of BNStartUpCmd.web3:
       case web3Cmd* {.command.}: Web3Cmd
       of Web3Cmd.test:
         web3TestUrl* {.
@@ -581,7 +582,7 @@ type
           desc: "The web3 provider URL to test"
           name: "url" }: Uri
 
-    of slashingdb:
+    of BNStartUpCmd.slashingdb:
       case slashingdbCmd* {.command.}: SlashProtCmd
       of SlashProtCmd.`import`:
         importedInterchangeFile* {.
@@ -596,6 +597,23 @@ type
         exportedInterchangeFile* {.
           desc: "EIP-3076 slashing protection interchange file to export"
           argument }: OutFile
+
+    of BNStartUpCmd.trustedNodeSync:
+      trustedNodeUrl* {.
+        desc: "URL of the REST API to sync from"
+        defaultValue: "http://localhost:5052"
+        name: "trusted-node-url"
+      .}: string
+
+      blockId* {.
+        desc: "Block id to sync to - this can be a block root, slot number, \"finalized\" or \"head\""
+        defaultValue: "finalized"
+      .}: string
+
+      backfillBlocks* {.
+        desc: "Backfill blocks directly from REST server instead of fetching via API"
+        defaultValue: true
+        name: "backfill"}: bool
 
   ValidatorClientConf* = object
     logLevel* {.
--- a/beacon_chain/trusted_node_sync.nim
+++ b/beacon_chain/trusted_node_sync.nim
@@ -1,0 +1,359 @@
+import
+  std/[os],
+
+  stew/[assign2, base10],
+  chronicles, chronos,
+  ./sync/sync_manager,
+  ./consensus_object_pools/blockchain_dag,
+  ./spec/eth2_apis/rest_beacon_client,
+  ./spec/[beaconstate, eth2_merkleization, forks, presets, state_transition],
+  "."/[beacon_clock, beacon_chain_db]
+
+{.push raises: [Defect].}
+
+type
+  DbCache = object
+    summaries: Table[Eth2Digest, BeaconBlockSummary]
+    slots: seq[Option[Eth2Digest]]
+
+const
+  emptyHash = Eth2Digest()
+
+proc updateSlots(cache: var DbCache, root: Eth2Digest, slot: Slot) =
+  # The slots mapping stores one linear block history - we construct it by
+  # starting from a given root/slot and walking the known parents as far back
+  # as possible - this ensures that
+  if cache.slots.len() < slot.int + 1:
+    cache.slots.setLen(slot.int + 1)
+
+  var
+    root = root
+    lastSlot = slot
+
+  while true:
+    cache.summaries.withValue(root, v) do:
+      let slot = v[].slot
+
+      for i in slot.int + 1..<lastSlot.int: # Avoid re-querying known gaps
+        cache.slots[i] = some(emptyHash)
+
+      cache.slots[slot.int] = some(root)
+
+      if slot == 0:
+        return
+
+      root = v[].parent_root
+      lastSlot = slot
+    do:
+      return
+
+proc update(cache: var DbCache, blck: ForkySignedBeaconBlock) =
+  let
+    slot = blck.message.slot
+
+  if blck.root notin cache.summaries:
+    cache.summaries[blck.root] = blck.message.toBeaconBlockSummary()
+
+  cache.updateSlots(blck.root, blck.message.slot)
+
+proc isKnown(cache: DbCache, slot: Slot): bool =
+  slot.int < cache.slots.len and cache.slots[slot.int].isSome()
+
+proc doTrustedNodeSync*(
+    cfg: RuntimeConfig, databaseDir: string, restUrl: string,
+    blockId: string, backfill: bool,
+    genesisState: ref ForkedHashedBeaconState = nil) {.async.} =
+  notice "Starting trusted node sync",
+    databaseDir, restUrl, blockId
+
+  let
+    db = BeaconChainDB.new(databaseDir, inMemory = false)
+
+  var
+    dbCache = DbCache(summaries: db.loadSummaries())
+
+  let
+    dbHead = db.getHeadBlock()
+    headSlot = if dbHead.isSome():
+      if dbHead.get() notin dbCache.summaries:
+        # This can happen with pre-blocksummary database - it's better to start
+        # over in this case
+        error "Database missing head block summary - database too old or corrupt"
+        quit 1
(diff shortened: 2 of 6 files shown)


beacon_chain/sync/sync_queue.nim
```diff
@@ -533,7 +533,7 @@ proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
           some(sq.readyQueue.pop())
       of SyncQueueKind.Backward:
         let maxSlot = sq.readyQueue[0].request.slot +
-                      (sq.readyQueue[0].request.count - 1'u64)
+                      sq.readyQueue[0].request.count
```
cheatfate: Request slot points to first slot and request count points to number of slots, so `maxSlot` is definitely not `request.slot + request.count` because:

request.slot = 0
request.cout = 2

so `maxSlot` is not `2` because there will be [0, 1] downloaded, so `maxSlot` should be `1`.

`outSlot` points to the next slot which should be processed, and does not points to the last slot processed.

jclapis:
Hi guys, first of all *thank you for doing this*. I cannot tell you how requested checkpoint syncing is from people; especially if they had a catastrophic node failure and need to get their validators back up immediately.

I just gave this branch a quick test. Some initial feedback:

- My ideal mode of operation would be to add both `--trusted-node-url` and `--backfill` as command line arguments to the normal `nimbus_beacon_node` command (i.e. "no action"), and include them if the user specified that they would like to checkpoint sync. I would then set `--trusted-node-url` to whatever they specified, and `--backfill` to false so Nimbus pulls the historical blocks from gossip once checkpoint syncing has completed.
  - This is the behavior that both Lighthouse and Teku currently provide.
  - With the separate `trustedNodeSync` action, I have to manually determine whether or not Nimbus has sync'd before starting it up, and whether or not to start with the special action first and then restart into the conventional mode after. I'd strongly suggest removing the explicit action and building support for this into the "no action" mode so the client itself can make this determination and behave accordingly.
- I tried syncing from Infura, and it failed with the following: `Unable to download genesis state           error="Downloading states via JSON not supported" restUrl=https://...@eth2-beacon-prater.infura.io`. Support for checkpoint syncing via a provider like Infura is likely going to be a requirement in order to use it with Rocket Pool, so I encourage adding JSON support.
- I *was* able to successfully checkpoint sync by pointing it at my other Nimbus node's REST endpoint. I see that setting `--backfill=false` and then restarting Nimbus to normal operations works as intended; it pulls historical blocks from gossip (`sync="backfill: 14d01h34m (0.03%) 1.6831slots/s (QQDDQQQQDQ:2045409)"`) which is excellent.
- If I disable my `eth1` client, Nimbus will fail with the following after a few minutes:
```
eth2_1           | WRN 2022-01-01 18:33:17.143+00:00 Eth1 chain monitoring failure, restarting  topics="eth1" err="net_version(web3.provider) failed 3 times"
eth2_1           | Traceback (most recent call last, using override)
eth2_1           | /home/user/nimbus-eth2/vendor/nim-json-rpc/json_rpc/client.nim(372) main
eth2_1           | /home/user/nimbus-eth2/vendor/nim-json-rpc/json_rpc/client.nim(365) NimMain
eth2_1           | /home/user/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1913) main
eth2_1           | /home/user/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1570) doRunBeaconNode
eth2_1           | /home/user/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1232) start
eth2_1           | /home/user/nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1176) run
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncloop.nim(279) poll
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(1215) colonanonymous
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(1202) start
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(1074) startEth1Syncing
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(1076) startEth1Syncing
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(872) resetState
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(879) resetState
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(397) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
eth2_1           | /home/user/nimbus-eth2/beacon_chain/eth1/eth1_monitor.nim(404) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-web3/web3.nim(92) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-json-rpc/json_rpc/client.nim(34) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(135) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncmacro2.nim(113) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-json-rpc/json_rpc/clients/websocketclient.nim(136) close
eth2_1           | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/asyncfutures2.nim(839) cancelAndWait
eth2_1           | /home/user/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(610) signalHandler
eth2_1           | SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```
Looks like it's failing when restarting the eth1 monitor. I suspect that isn't related to checkpoint syncing, but it should be looked at anyway.

arnetheduck:
> Support for checkpoint syncing via a provider like Infura is likely going to be a requirement in order to use it with Rocket Pool, so I encourage adding JSON support.

https://github.com/status-im/nimbus-eth2/pull/3232

beacon_chain/trusted_node_sync.nim
```diff
@@ -0,0 +1,359 @@
...
+    slots: seq[Option[Eth2Digest]]
+
+const
+  emptyHash = Eth2Digest()
+
+proc updateSlots(cache: var DbCache, root: Eth2Digest, slot: Slot) =
+  # The slots mapping stores one linear block history - we construct it by
+  # starting from a given root/slot and walking the known parents as far back
+  # as possible - this ensures that
+  if cache.slots.len() < slot.int + 1:
```
tersec, line 32: `lenu64` would allow this to be somewhat type-safer

beacon_chain/trusted_node_sync.nim
```diff
@@ -0,0 +1,359 @@
...
+  let
+    slot = blck.message.slot
+
+  if blck.root notin cache.summaries:
+    cache.summaries[blck.root] = blck.message.toBeaconBlockSummary()
+
+  cache.updateSlots(blck.root, blck.message.slot)
+
+proc isKnown(cache: DbCache, slot: Slot): bool =
+  slot.int < cache.slots.len and cache.slots[slot.int].isSome()
```
tersec: Also `lenu64`

beacon_chain/trusted_node_sync.nim
```diff
@@ -0,0 +1,359 @@
...
+    # When the checkpoint is newer than the head, we run into trouble: the
+    # current backfill in ChainDAG does not support filling in arbitrary gaps.
+    # If we were to update the backfill pointer in this case, the ChainDAG
+    # backfiller would re-download the entire backfill history.
+    # For now, we'll abort and let the user choose what to do.
+    error "Checkpoint block is newer than head slot - start with a new database or use a checkpoint no more recent than the head",
+      checkpointSlot, checkpointRoot = shortLog(checkpointBlock.root), headSlot
+    quit 1
+
+  if checkpointSlot.uint64 mod SLOTS_PER_EPOCH != 0:
```
tersec: `isEpoch` handles/encapsulates the type conversion aspect

beacon_chain/conf.nim
```diff
@@ -37,14 +37,15 @@ const
   defaultSigningNodeRequestTimeout* = 60
 
 type
-  BNStartUpCmd* = enum
+  BNStartUpCmd* {.pure.} = enum
```
mratsim, line 40: isn't that placebo now?

beacon_chain/trusted_node_sync.nim
```diff
@@ -0,0 +1,359 @@
...
+    for i in 0..<checkpointSlot.int:
+      if dbCache.slots[i].isNone():
+        total += 1
+    total
+
+  if missingSlots == 0:
+    info "Database fully backfilled"
+  elif backfill:
+    notice "Backfilling historical blocks",
+      checkpointSlot, missingSlots
```
mratsim, line 307: Will this part be mutualized with deferred backfill in #3263

beacon_chain/trusted_node_sync.nim
```diff
@@ -0,0 +1,359 @@
...
+        # This can happen with pre-blocksummary database - it's better to start
+        # over in this case
+        error "Database missing head block summary - database too old or corrupt"
+        quit 1
+      let slot =  dbCache.summaries[dbHead.get()].slot
+      dbCache.updateSlots(dbHead.get(), slot)
+      slot
+    else:
+      # When we don't have a head, we'll use the given checkpoint as head
+      FAR_FUTURE_SLOT
```
mratsim, line 95: So let's say an user:

1. starts Nimbus
2. realizes sync will take some time and look for trusted node sync
3. stops their nodes after 30min
4. restart with trusted node sync.

The head would not be the chain head but the forward sync head, hence the user would have to delete the DB and restart over?

mratsim (commented):
Question if user starts forward syncing first, then stops then checkpoint syncs.

beacon_chain/trusted_node_sync.nim
arnetheduck: ...and crash on the next line where `setLen(slot.int + 1)` happens :/

tersec: Yeah, I don't see a great approach to this. Maybe better to at least be consistent within these two lines, leave it as is.

beacon_chain/trusted_node_sync.nim
tersec: ... and, this has the same also-accessing-integer-indexed-array issue. So probably as fine as it can easily get as-is.

beacon_chain/trusted_node_sync.nim
arnetheduck: Yes - this condition is detected further down, when we find out the checkpoint is newer: https://github.com/status-im/nimbus-eth2/pull/3209/files/6160c7a91eff88801a9a07dbf9a79b1feaa41b40#diff-415a96547dcab28dacb7bd1503b5c191600e71925f475a85c8bd616fb584a962R190

It's possible to handle this case as well by simply moving the head to the new position and backfill to the old head instead of all the way to genesis, but that's for a separate future PR I think (in particular, the backfiller would need more work to support this case cc @cheatfate )

beacon_chain/trusted_node_sync.nim
arnetheduck: The backfill of 3263 picks up wherever this back fill is interrupted

beacon_chain/conf.nim
arnetheduck: it works as far as I can tell - ie when there's a conflict, it complains appropriately

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -0,0 +1,75 @@
...
+
+wget -o state.32000.ssz http://localhost:5052/eth/v2/debug/beacon/states/32000
+wget -o block.32000.ssz http://localhost:5052/eth/v2/beacon/blocks/32000
+
+build/nimbus_beacon_node --data-dir:trusted --finalized-checkpoint-block=block.32000.ssz --finalized-checkpoint-state=state.32000.ssz
+```
+
+## Caveats
+
+A node synced using trusted node sync will not be able to serve historical requests from before the checkpoint. Future versions will resolve this issue.
```
mratsim, line 75: What's the impact on sync committee duties?

mratsim (approved):

docs/the_nimbus_book/src/trusted-node-sync.md
arnetheduck: None, really - the duties are read from the head state which is always available

beacon_chain/trusted_node_sync.nim
jclapis: > So let's say an user:
> 
> 1. starts Nimbus
> 2. realizes sync will take some time and look for trusted node sync
> 3. stops their nodes after 30min
> 4. restart with trusted node sync.
> 
> The head would not be the chain head but the forward sync head, hence the user would have to delete the DB and restart over?

Based on the fact that checkpoint syncing is a separate action right now, this is likely the workflow we will end up using in Rocket Pool. Check if the database file exists, if not then do a checkpoint sync first (if enabled), else start normally. If the database already exists and the user wants to checkpoint sync because it's taking too long, we will just have them run a command to delete the database and start over using this logic.

arnetheduck: Hey, I grabbed your feedback and put it in https://github.com/status-im/nimbus-eth2/issues/3285: with trusted node sync in place, it's easier to consider a checkpoint URL for the "main" command as well, but there are some things we want to consider a bit from a UX point of view before moving in that direction.

What you propose for rocketpool is a good starting point regardless, but let's continue the discussion in that issue - as usual, the detail of the user story is much appreciated.

PR #3213, opened 2021-12-20
Merge fork gossip support

tersec:
It still doesn't run the local testnet with the merge fork, because that doesn't work with the current validator client.

It can't publish merge blocks:
https://github.com/status-im/nimbus-eth2/blob/c7be88b432cb0612ede893ccbf2941af69882dd5/beacon_chain/validator_client/api.nim#L767-L779

While at some level, there's a function overload straightforwardly missing, even filling it in here:
https://github.com/status-im/nimbus-eth2/blob/c7be88b432cb0612ede893ccbf2941af69882dd5/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim#L83-L91
just reproduces the `nim-presto` issue the comment describes.

And even if it could, the REST API on the other side wouldn't know what to do with them:
https://github.com/status-im/nimbus-eth2/blob/c7be88b432cb0612ede893ccbf2941af69882dd5/beacon_chain/rpc/rest_beacon_api.nim#L752-L780

Which is all summarized by
https://github.com/status-im/nimbus-eth2/blob/6ef3834f4ac4fa3dd8c0191686ba07c0233e8bc7/beacon_chain/validators/validator_duties.nim#L1224-L1225

Addressing these is out of scope of this PR.

Adding a `MERGE_FORK_EPOCH` to `scripts/launch_local_testnet.sh` and disabling `USE_VC` works fine, though.

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -619,10 +619,6 @@ proc validateAggregate*(
     genesis_validators_root =
       getStateField(pool.dag.headState.data, genesis_validators_root)
 
-  info "FOO5",
-    fork,
-    aggregate_data_slot = aggregate.data.slot
-
   let attesting_indices = get_attesting_indices(
     epochRef, aggregate.data, aggregate.aggregation_bits)
 
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -619,6 +619,10 @@ proc validateAggregate*(
     genesis_validators_root =
       getStateField(pool.dag.headState.data, genesis_validators_root)
 
+  info "FOO5",
+    fork,
+    aggregate_data_slot = aggregate.data.slot
+
   let attesting_indices = get_attesting_indices(
     epochRef, aggregate.data, aggregate.aggregation_bits)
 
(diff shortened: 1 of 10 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -619,6 +619,10 @@ proc validateAggregate*(
     genesis_validators_root =
       getStateField(pool.dag.headState.data, genesis_validators_root)
 
+  info "FOO5",
```
arnetheduck: hello foo5 :)

beacon_chain/spec/forks.nim
```diff
@@ -36,9 +36,10 @@ export
 
 type
   BeaconStateFork* {.pure.} = enum
-    Phase0,
-    Altair,
-    Merge
+    # Used to index arrays, so ensure correct indices.
```
arnetheduck: `array[BeaconStateFork, T]` is indexed by the `enum` "naturally"

tersec: https://github.com/status-im/nimbus-eth2/pull/3213/commits/bd1a53b32aeb92d1dad63fbeac0502b65027a465

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3213/commits/bd1a53b32aeb92d1dad63fbeac0502b65027a465

PR #3216, opened 2021-12-20
sync status: better ETA calculation

stefantalpalaru:
...and stop the status bar from jumping left and right.

## long story short

A day ago I finished syncing mainnet from scratch, in 3d8h40m. The initial ETA was way, way off and it stayed unusable for most of the duration of that initial sync, so I did something about it.

### commands used

```text
./run-mainnet-beacon-node.sh --web3-url="wss://mainnet.infura.io/ws/v3/TOKEN_HERE" --metrics --metrics-port=8008 --enr-auto-update --doppelganger-detection=off --log-file=build/data/shared_mainnet_0/nimbus.log --num-threads=0
# in some other terminal:
grep 'Slot start' build/data/shared_mainnet_0/nimbus.log | jq '.sync'
```

### before

```text
"13h43m (0.05%) 55.9907slots/s (PDQDQDQDDD:1248)"
"17h30m (0.05%) 43.8854slots/s (DPQQDQDQDQ:1408)"
"22h48m (0.06%) 33.6848slots/s (DDPDQDQDQD:1600)"
"01d00h33m (0.08%) 31.2586slots/s (QDQDQDQPDQ:2144)"
"20h36m (0.10%) 37.2633slots/s (DDPDQDQDQD:2880)"
"19h00m (0.13%) 40.3822slots/s (DDDDPDQDQD:3552)"
"18h30m (0.15%) 41.4705slots/s (QQQQQQQDQP:4096)"
"17h43m (0.17%) 43.2838slots/s (DDDDPDQDQD:4832)"
"17h02m (0.20%) 44.9881slots/s (DQDDDDPDQD:5504)"
"16h44m (0.22%) 45.8222slots/s (QQQQQDQDQP:6016)"
"16h34m (0.24%) 46.2587slots/s (QQQQQQQDQP:6656)"
"16h26m (0.26%) 46.6243slots/s (DDQDQPDQDQ:7232)"
"16h32m (0.28%) 46.3174slots/s (DDQPQQDDDD:7840)"
"16h34m (0.30%) 46.2444slots/s (DPQQDDQDQD:8384)"
```

### after

```text
"03d01h21m (0.04%) 30.40slt/s (DDQQDDDDDP)"
"03d20h08m (0.05%) 40.45slt/s (QQDDDPQQQQ)"
"04d10h58m (0.05%) 33.54slt/s (QDQQQDDDPQ)"
"04d21h38m (0.06%) 30.01slt/s (QQDDDPQQQQ)"
"04d06h01m (0.08%) 27.67slt/s (QQPDDQQQQQ)"
"03d15h02m (0.13%) 36.25slt/s (QQQPQQQQDD)"
"03d12h45m (0.15%) 38.27slt/s (QQQQQQQDQQ)"
"03d10h46m (0.16%) 39.88slt/s (QQDDDPQQQQ)"
"03d08h37m (0.18%) 40.90slt/s (QQQPQQDDDD)"
"03d06h55m (0.21%) 42.03slt/s (DDQQQDDDPQ)"
"03d04h13m (0.25%) 44.12slt/s (PQQDDDQDQQ)"
"03d04h45m (0.26%) 45.23slt/s (QDQQDDDPQQ)"
"03d04h22m (0.28%) 44.80slt/s (QQDDDPQQQD)"
"03d03h46m (0.30%) 45.17slt/s (QDQPQQDDDQ)"
```

## the data

Screenshot 2021-12-19 at 05-11-32 NBC local testnet sim (all nodes) - Grafana

You can see how a naive ETA algorithm - one assuming a linear syncing speed - would be so bad as to be unusable. Any straight line from the origin to any point on that curve fails to predict the rest of it.

## curve fitting

However, we're lucky to know beforehand how the real syncing curve looks like. We can just do some curve fitting on it and use that polynomial as our ETA prediction function. I chose R for this task, because I find the language and ecosystem hilarious. "curve_fit.R":

```r
#!/usr/bin/env Rscript

# Model the initial block syncing speed of nimbus_beacon_node on mainnet, on a
# local desktop, with multi-threading enabled.

# Got the query params from Grafana and lowered the step:
# curl -s 'http://localhost:9090/api/v1/query_range?query=beacon_head_slot&start=1639667400&end=1639865880&step=20' | jq -r '.data.result[0].values | map([.[0], .[1]] | join(", ")) | join("\n")' > slots.csv
data <- read.csv("./slots.csv", header=FALSE, col.names=c("unix_timestamp", "slots"))

# It's hard to get a good fit with a single polynomial, but easy with two.
curve_split_after_seconds <- 40000 # corresponds to slot 751597
curve_split <- curve_split_after_seconds / 20 # in number of elements in the vector

# "lm()" won't work on Unix timestamps - maybe because they are too close in
# the floating point domain. We need to lower these values so they start from
# zero.
slots_processed_all <- data.frame(slots = data$slots, seconds = mapply(function(x) x - data$unix_timestamp[1], data$unix_timestamp))
str(slots_processed_all)

png(file="slots.png", width=1000, height=800)
plot(slots_processed_all$seconds, slots_processed_all$slots, main="processed slots (mainnet)", xlab="seconds from start", ylab="slots", xaxt="n", pch=16, cex=.1)
axis(side=1, tck=-0.015, at=c(seq(from=0,to=200000,by=1000)))
grid(nx = 220, ny = NULL, col = "lightgray", lty = "dotted", lwd = par("lwd"), equilogs = TRUE)

message("Polynomial model for seconds_from_start=f(slots_synced) for the first ", curve_split_after_seconds, "s")
slots_processed_initial <- data.frame(slots = slots_processed_all$slots[1:curve_split], seconds = slots_processed_all$seconds[1:curve_split])
initial_fit <- lm(seconds~poly(slots, 3, raw=TRUE), data=slots_processed_initial)
summary(initial_fit)
# y = -3.910e-15 x^3 + 4.547e-08 x^2 + 2.173e-02 x + 6.760e+01
lines(predict(initial_fit, data.frame(slots = slots_processed_initial$slots)), slots_processed_initial$slots, col='red')

message("Polynomial model for seconds_from_start=f(slots_synced), after the first ", curve_split_after_seconds, "s")
slots_processed_final <- data.frame(slots = tail(slots_processed_all$slots, -curve_split), seconds = tail(slots_processed_all$seconds, -curve_split))
str(slots_processed_final)
final_fit <- lm(seconds~poly(slots, 3, raw=TRUE), data=slots_processed_final)
summary(final_fit)
# y = 1.436e-15 x^3 + 2.510e-08 x^2 + 1.743e-02 x + 1.314e+04
lines(predict(final_fit, data.frame(slots = slots_processed_final$slots)), slots_processed_final$slots, col='blue')
```

Yeah, I ended up with two different curves, for a better fit. The resulting plot:

slots

Implementing those fitted curves in Nim was trivial.

The beauty of this scheme is that it's a net improvement for any network, not just mainnet, and it will continue to be useful - without more curve fitting - until the merge makes blocks significantly bigger.

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -7,7 +7,7 @@
 
 {.push raises: [Defect].}
 
-import std/[options, heapqueue, tables, strutils, sequtils, algorithm]
+import std/[options, heapqueue, tables, strutils, sequtils, algorithm, math]
 import stew/results, chronos, chronicles
 import
   ../spec/datatypes/[phase0, altair, merge],
@@ -494,9 +494,23 @@ proc toTimeLeftString(d: Duration): string =
       res = res & "00m"
     res
 
+# Curve(s) fitted to data collected while syncing mainnet on an old desktop -
+# i.e.: the "reference system".
+func referenceCurveSlotsToSeconds(slots: float): float =
+  if slots < 751597:
+    -3.910e-15 * slots^3 + 4.547e-08 * slots^2 + 2.173e-02 * slots + 6.760e+01
+  else:
+    1.436e-15 * slots^3 + 2.510e-08 * slots^2 + 1.743e-02 * slots + 1.314e+04
+
 proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
   mixin getKey, getScore
-  var pauseTime = 0
+
+  let
+    # Offset it a little, to improve the initial estimate.
+    startTime = chronos.Moment.now() + chronos.seconds(5)
+    initialProgress = man.getLocalHeadSlot().float
+    # We don't want the ETA or syncing speed to jump all over the place too often.
+    updateInterval = chronos.seconds(10)
 
   asyncSpawn man.guardTask()
 
@@ -510,15 +524,12 @@ proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
 
       await man.notInSyncEvent.wait()
 
-      # Give the node time to connect to peers and get the sync process started
-      await sleepAsync(seconds(SECONDS_PER_SLOT.int64))
-
       var
         stamp = SyncMoment.now(man.queue.progress())
         syncCount = 0
 
       while man.inProgress:
-        await sleepAsync(seconds(SECONDS_PER_SLOT.int64))
+        await sleepAsync(updateInterval)
 
         let
           newStamp = SyncMoment.now(man.queue.progress())
@@ -545,25 +556,46 @@ proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
           waiting_workers_count = waiting,
           pending_workers_count = pending,
           wall_head_slot = wallSlot, local_head_slot = headSlot,
-          pause_time = $chronos.seconds(pauseTime),
           avg_sync_speed = man.avgSyncSpeed, ins_sync_speed = man.insSyncSpeed,
           direction = man.direction, topics = "syncman"
 
+    # This ETA algorithm only works for the initial forward sync. When
+    # checkpoint sync is added, and if we still want to display the backfill
+    # sync status in the status bar, the "total" and "progress" variables need
+    # to be adjusted accordingly.
     let
-      progress = float(man.queue.progress())
-      total = float(man.queue.total())
-      remaining = total - progress
+      progress = headSlot.float
+      total = wallSlot.float
       done = if total > 0.0: progress / total else: 1.0
+      elapsedSeconds = (chronos.Moment.now() - startTime).seconds.float
+      # How long would it have taken the reference system to sync up to this
+      # point?
+      referenceDuration = referenceCurveSlotsToSeconds(progress) -
+        referenceCurveSlotsToSeconds(initialProgress)
+      # Assume a linear relation between CPU performance and syncing speed and
+      # compute a correction factor between this system and the reference one.
+      correctionFactor =
+        if referenceDuration > 0.0 and elapsedSeconds > 0.0:
+          elapsedSeconds / referenceDuration
+        else:
+          1.0
+      # Now it's easy to get a credible ETA.
+      estimatedTotalDuration = (referenceCurveSlotsToSeconds(total) * correctionFactor)
       timeleft =
-        if man.avgSyncSpeed >= 0.001:
-          Duration.fromFloatSeconds(remaining / man.avgSyncSpeed)
+        if man.avgSyncSpeed >= 0.001 and elapsedSeconds > 0.0:
+          # Complicated a little by the fact that a node can be restarted
+          # during initial sync. We need to also estimate the time needed to
+          # sync up to the number of slots this instance already had when
+          # it started.
+          Duration.fromFloatSeconds(estimatedTotalDuration - elapsedSeconds -
+            (referenceCurveSlotsToSeconds(initialProgress) * correctionFactor))
         else: InfiniteDuration
 
     # Update status string
     man.syncStatus = timeLeft.toTimeLeftString() & " (" &
                     (done * 100).formatBiggestFloat(ffDecimal, 2) & "%) " &
-                    man.avgSyncSpeed.formatBiggestFloat(ffDecimal, 4) &
-                    "slots/s (" & map & ":" & $man.queue.outSlot & ")"
+                    man.avgSyncSpeed.formatBiggestFloat(ffDecimal, 2) &
+                    "slt/s (" & map & ")"
 
     if headAge <= man.maxHeadAge:
       man.notInSyncEvent.clear()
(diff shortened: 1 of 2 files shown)


beacon_chain/sync/sync_manager.nim
```diff
@@ -494,9 +494,23 @@ proc toTimeLeftString(d: Duration): string =
       res = res & "00m"
     res
 
+# Curve(s) fitted to data collected while syncing mainnet on an old desktop -
```
tersec, line 497: How portable is this to different machines? I don't doubt it's helpful for your case specifically, because that's what you fitted your segmented cubic polynomials on, but is this just an arbitrary function of, e.g., which blocks happened to be large in mainnet? Does it apply to Prater? Cubic polynomials also start doing wacky things at x-axis extrema, being, well, cubic in nature. Could backfire.

tersec: Also, they don't even exhibit C0 continuity, never mind C1 or C2 continuity:
```
>>> polya = lambda slots: -3.910e-15 * slots**3 + 4.547e-08 * slots**2 + 2.173e-02 * slots + 6.760e+01
>>> polyb = lambda slots:  1.436e-15 * slots**3 + 2.510e-08 * slots**2 + 1.743e-02 * slots + 1.314e+04
>>> polya(751597)
40425.62625332361
>>> polyb(751597)
41028.967451736215
>>> polyb(751596)
41028.90985801435
>>> polya(751596)
40425.54279938322
```

stefantalpalaru: Try it for yourself. It's better than the linear speed we assume now, on any system, on any network.

Continuity is not that important for what we're doing here. It just needs to be close enough for jazz.

tersec: I don't object if someone else wants to merge this. It probably is a net improvement over the status quo.

arnetheduck: > Does it apply to Prater?

It will likely be significantly misleading for testnets - these have pretty-much constant validator counts (which by and large determine sync speed) - the code needs to take the network it runs on into account.

Also, for backfill, the performance curve is completely different (mostly linear) - this code needs to take the direction of the sync queue into account and fall back to a normal average if backwards queues (or use a separate curve) - the canonical source for progress information is the sync queue - the code should not be calling `getHeadSlot`/`getWallSlot` directly, these are intended only for verifying the status reported by other peers.

How to get progress: https://github.com/status-im/nimbus-eth2/blob/1021e3324e189a6653edc6b54871829f43b5e121/beacon_chain/sync/sync_manager.nim#L517

How does this curve behave 6 months from now with slots in the 3 million range?

At what slot does the cube overflow and what happens then?

stefantalpalaru: > It will likely be significantly misleading for testnets

Not as misleading as it is now.

> Also, for backfill, the performance curve is completely different

But do we want a sync status being displayed for backfill?

> the canonical source for progress information is the sync queue

Too bad it's reset with each restart. I don't want to see the sync progress go back to 0% after restarting halfway through.

> How does this curve behave 6 months from now with slots in the 3 million range?

Better than the line we're assuming now.

> At what slot does the cube overflow and what happens then?

Why would uint64 converted to float64 overflow?

arnetheduck: > Not as misleading as it is now.

Well, it will overestimate significantly at the start which is just as bad - this is usually where people start with Nimbus so we ideally don't want to give them a worse picture than reality.

One way to measure this would  be to analyze it with something like a mean-square error for a sync on the same machine

The mere fact that the formula contains two curves with a slot selection suggests that the curves are not as good as they sound and are strongly tied to the network and slot numbers that are being synced.

> But do we want a sync status being displayed for backfill?

Yes, this is already implemented (https://github.com/status-im/nimbus-eth2/pull/3209/files#diff-3e73e1b595d5405f6872e00dd9f4ee2b238648b617dd2d29a670a09767b5659dR908) - backfilling takes about a day because it competes with other work, but while it's ongoing, attestation performance may suffer due to the extra load.

Also, the same `SyncManager` code is used for both directions, so we can't introduce a curve here without taking direction into account.

> Too bad it's reset with each restart. 

This is a bit of an open UX point: if you lose sync or restart, is 0 or 99.99% more useful as a number when you start syncing?

stefantalpalaru: > an open UX point

Not really. Syncing and stopping is like downloading a file and pausing it.

arnetheduck: Files typically don't get larger while you're downloading - ie you'll potentially see a decrease in progress no matter how you do it - that said, this point has merit, but we still need to move away from `getHeadSlot` / `getWallSlot`, this is not their intended use

arnetheduck: > Too bad it's reset with each restart

https://github.com/status-im/nimbus-eth2/pull/3276

zah:
Since Eth1 syncing completes much before Eth2 syncing, I can imagine a solution that takes into account the shape of the deposit chain to construct a predictive function that would work for any network. Another advantage of such a solution is that it wouldn't require any maintenance in the future.

arnetheduck:
> predictive function that would work for any network.

clever - that theory can easily be tested by comparing validator-vs-sync-speed correlation on mainnet and prater - these are two extremes on the "validator count shape" metric

PR #3217, opened 2021-12-21
Handle access to historical data for which there is no state 

arnetheduck:
With checkpoint sync in particular, and state pruning in the future,
loading states or state-dependent data may fail. This PR adjusts the
code to allow this to be handled gracefully.

In particular, the new availability assumption is that states are always
available for the finalized checkpoint and newer, but may fail for
anything older.

The `tail` remains the point where state loading de-facto fails, meaning
that between the tail and the finalized checkpoint, we can still get
historical data (but code should be prepared to handle this as an
error).

However, to harden the code against long replays, several operations
which are assumed to work only with non-final data (such as gossip
verification and validator duties) now limit their search horizon to
post-finalized data.

* harden several state-dependent operations by logging an error instead
of introducing a panic when state loading fails
* `withState` -> `withUpdatedState` to differentiate from the other
`withState`
* `updateStateData` can now fail if no state is found in database - it
is also hardened against excessively long replays
* `getEpochRef` can now fail when replay fails
* reject blocks with invalid target root - they would be ignored
previously
* ~~fix recursion bug in `isProposed`~~ fix included in https://github.com/status-im/nimbus-eth2/pull/3225

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -134,7 +134,9 @@ proc init*(T: type AttestationPool, dag: ChainDAGRef,
             epochRef.current_justified_checkpoint,
             epochRef.finalized_checkpoint)
         else:
-          epochRef = dag.getEpochRef(blckRef, blckRef.slot.epoch)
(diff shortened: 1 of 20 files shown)


tersec (approved):
Looks like at least it'd prevent undefined behavior. A kind of minimal necessary follow-on to what's already in `unstable`.

Issue #3129, opened 2021-11-26
Implement optimized history traversal in the REST API


zah:
A common usage pattern of the REST API is to traverse the entire history of the beacon chain by requesting epoch or slot information one step at a time (i.e. using increasing consecutive numbers). At the moment, this would be very expensive as each requests will go through the motions of loading and replaying the same blocks over and over again.

A simple caching mechanism can make our API significantly faster:

1) Each time an old state is restored from the database, we can keep it around for a certain amount of time (e.g. 10 seconds, a configurable value).

2) When another requests comes it, we take a look whether some of our existing cached states can be advanced cheaply to satisfy the requests (i.e. we can define a maximum slot/epoch difference to be considered a cheap match).

3) We use the existing lower-level APIs of the DAG to advance the cached state (e.g. `updateStateData`) and serve the request.

Practical issues:

The `updateStateData` function needs some minimal modifications to make this work - it needs to consider its `state` parameter as input/output and enhance its "UpdateStateData cache miss" path to consider the input state as a possible starting replay point. This would require some care at the existing call-sites that reuse existing state objects such as `dag.epochRefState` and may assume that `state` is a pure output parameter. Alternatively, we can just extract the replay logic from `updateStateData` into another reusable function.

Security considerations:

If the cache in unbounded, a rogue user of the API may exhaust the system resources by requesting different states. We can mitigate this by simply bounding the size of the cache to a configurable value with a reasonable default (e.g. 10). It is unlikely that typical users will run that many concurrent history traversals.

arnetheduck:
It's actually more simple than that: on the first rest API request, we can allocate a beacon state that we keep reusing just for the rest api, just like we reuse the current states for specific purposes - this division of states is here because there's no way in nim to track ownership, and we've had numerous bugs because one use of the state overwrites another.

re `updateStateData`, it has some primitive logic to compute the "cost" of a replay, so it should automatically reuse the "closest" `BeaconState` it knows of (ie there's several costs it can roughly anticipate - state load, block load, epoch switch and so on) - this should be trivial to extend in any direction.

That said, it's also slightly questionable as far as features go - ie there are two kinds of requests: pre-finality and post-finality - the post-finality requests are pretty fast no matter what because we keep every epoch in the database and it's actually not that expensive to reload (of course, it's a cost).

Pre-finality requests should generally neither be used nor be necessary on a beacon node with live validators, and we should really block them by default, and require a flag to enable historicals as well.

[...]

PR #3219, opened 2021-12-22
Implement #3129 (Optimized history traversals in the REST API)

zah:
Up to 10x speed-up for historic traversals through the API:

```
# Before
➜  nimbus-eth2 git:(feature/rest_api_optimized_history_traversal) build/rest_api_benchmark --start-slot=9000 --count=250
INF 2021-12-22 09:24:55.962+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateRoot
INF 2021-12-22 09:26:08.637+02:00 getStateRoot                               file=rest_api_benchmark.nim:43 t=72.67443995799999s
INF 2021-12-22 09:26:08.637+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateFork
INF 2021-12-22 09:27:20.277+02:00 getStateFork                               file=rest_api_benchmark.nim:43 t=71.639665333s
INF 2021-12-22 09:27:20.277+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateFinalityCheckpoints
INF 2021-12-22 09:28:33.398+02:00 getStateFinalityCheckpoints                file=rest_api_benchmark.nim:43 t=73.12129654100001s
INF 2021-12-22 09:28:33.398+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateValidatorBalances
INF 2021-12-22 09:29:55.869+02:00 getStateValidatorBalances                  file=rest_api_benchmark.nim:43 t=82.470523375s

# After
➜  nimbus-eth2 git:(feature/rest_api_optimized_history_traversal) build/rest_api_benchmark --start-slot=9000 --count=250
INF 2021-12-22 09:24:55.962+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateRoot
INF 2021-12-22 09:26:08.637+02:00 getStateRoot                               file=rest_api_benchmark.nim:43 t=6.67443995799999s
INF 2021-12-22 09:21:32.563+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateFork
INF 2021-12-22 09:21:39.626+02:00 getStateFork                               file=rest_api_benchmark.nim:43 t=7.062567791s
INF 2021-12-22 09:21:39.626+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateFinalityCheckpoints
INF 2021-12-22 09:21:46.180+02:00 getStateFinalityCheckpoints                file=rest_api_benchmark.nim:43 t=6.55376425s
INF 2021-12-22 09:21:46.180+02:00 Benchmarking ...                           file=rest_api_benchmark.nim:42 apiName=getStateValidatorBalances
INF 2021-12-22 09:22:01.645+02:00 getStateValidatorBalances                  file=rest_api_benchmark.nim:43 t=15.465585083s
```

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -328,6 +328,18 @@ type
         defaultValueDesc: "127.0.0.1"
         name: "rest-address" }: ValidIpAddress
 
+      restCacheSize* {.
+        defaultValue: 3
+        desc: "The maximum number of recently accessed states that are kept in " &
+              "memory. Speeds up requests obtaining information for consecutive " &
+              "slots or epochs."
+        name: "rest-statecache-size" }: Natural
+
+      restCacheTtl* {.
+        defaultValue: 60
+        desc: "The number of seconds to keep recently accessed states in memory"
+        name: "rest-statecache-ttl" }: Natural
+
       keymanagerEnabled* {.
         desc: "Enable the REST keymanager API (BETA version)"
         defaultValue: false
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -609,6 +609,7 @@ proc getEpochRef*(dag: ChainDAGRef, blck: BlockRef, epoch: Epoch): EpochRef =
   let
     ancestor = epochAncestor(blck, epoch)
 
+  dag.epochRefState.blck = BlockRef()
   dag.withState(
       dag.epochRefState, ancestor.blck.atEpochStart(ancestor.epoch)):
     dag.getEpochRef(stateData, cache)
@@ -944,8 +945,10 @@ proc updateStateData*(
 
     # Look for a state in the database and load it - as long as it cannot be
     # found, keep track of the blocks that are needed to reach it from the
-    # state that eventually will be found
-    while not dag.getState(state, cur):
+    # state that eventually will be found. Besides finding the state in the
+    # database we may also reach the input state provided to the function.
+    # It can also act as a viable starting point for the block replay later.
+    while not canAdvance(state, cur) and not dag.getState(state, cur):
       # There's no state saved for this particular BlockSlot combination, keep
       # looking...
       if cur.slot == cur.blck.slot:
--- a/beacon_chain/rpc/rest_utils.nim
+++ b/beacon_chain/rpc/rest_utils.nim
@@ -1,4 +1,4 @@
-import std/options,
+import std/[options, macros],
        presto,
        nimcrypto/utils as ncrutils,
        ../spec/[forks],
@@ -116,8 +116,29 @@ proc getBlockDataFromBlockIdent*(node: BeaconNode,
                                  id: BlockIdent): Result[BlockData, cstring] =
   ok(node.dag.get(? node.getBlockRef(id)))
 
+proc disallowInterruptionsAux(body: NimNode) =
+  for n in body:
+    const because =
+      "because the `state` variable may be mutated (and thus invalidated) " &
+      "before the function resumes execution."
+
+    if n.kind == nnkYieldStmt:
+      macros.error "You cannot use yield in this block " & because, n
+
+    if (n.kind in {nnkCall, nnkCommand} and
+       n[0].kind in {nnkIdent, nnkSym} and
+       $n[0] == "await"):
+      macros.error "You cannot use await in this block " & because, n
+
+    disallowInterruptionsAux(n)
+
+macro disallowInterruptions(body: untyped) =
+  disallowInterruptionsAux(body)
+
 template withStateForBlockSlot*(nodeParam: BeaconNode,
-                                blockSlotParam: BlockSlot, body: untyped): untyped =
+                                blockSlotParam: BlockSlot,
+                                body: untyped): untyped =
+
   block:
     let
       node = nodeParam
@@ -127,6 +148,20 @@ template withStateForBlockSlot*(nodeParam: BeaconNode,
       state.blck.atSlot(getStateField(state.data, slot)) == blockSlot
 
     var cache {.inject, used.}: StateCache
+
+    # If we have a cache hit, there is a concern that the REST request
+    # handler may continue executing asynchronously while we hit the same
+    # advanced state is another request. We don't want the two requests
+    # to work over the same state object because mutations to it will be
+    # visible in both, so we must outlaw yielding within the `body` block.
+    # Please note that the problem is not limited to the situations where
+    # we have a cache hit. Working with the `headState` will result in the
+    # same problem as it may change while the request is executing.
+    #
+    # TODO
+    # The solution below is only partion, because it theory yields or awaits
+    # can still be hidden in the body through the use of helper templates
+    disallowInterruptions(body)
 
     # TODO view-types
     # Avoid the code bloat produced by the double `body` reference through a lent var
@@ -146,19 +181,11 @@ template withStateForBlockSlot*(nodeParam: BeaconNode,
 
       node.dag.updateStateData(stateToAdvance[], blockSlot, false, cache)
 
-      let providedState = if cachedState == nil and node.stateTtlCache != nil:
+      if cachedState == nil and node.stateTtlCache != nil:
         # This was not a cached state, we can cache it now
         node.stateTtlCache.add(stateToAdvance)
-        stateToAdvance
-      else:
-        # If we have a cache hit, there is a concern that the REST request
-        # handler may continue executing asynchronously while we hit the same
-        # advanced state is another request. We don't want the two requests
-        # to work over the same state object because mutations to it will be
-        # visible in both, so we must create a fresh clone for each handler:
-        assignClone(stateToAdvance[])
 
-      withStateVars(providedState[]):
+      withStateVars(stateToAdvance[]):
         body
 
 proc toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
--- a/beacon_chain/rpc/rest_utils.nim
+++ b/beacon_chain/rpc/rest_utils.nim
@@ -5,7 +5,7 @@ import std/options,
        ../spec/eth2_apis/[rest_types, eth2_rest_serialization],
        ../beacon_node,
        ../consensus_object_pools/blockchain_dag,
-       ./rest_constants
+       "."/[rest_constants, state_ttl_cache]
 
 export
   options, eth2_rest_serialization, blockchain_dag, presto, rest_types,
@@ -116,19 +116,50 @@ proc getBlockDataFromBlockIdent*(node: BeaconNode,
                                  id: BlockIdent): Result[BlockData, cstring] =
   ok(node.dag.get(? node.getBlockRef(id)))
 
-template withStateForBlockSlot*(node: BeaconNode,
-                                blockSlot: BlockSlot, body: untyped): untyped =
-  template isState(state: StateData): bool =
-    state.blck.atSlot(getStateField(state.data, slot)) == blockSlot
+template withStateForBlockSlot*(nodeParam: BeaconNode,
+                                blockSlotParam: BlockSlot, body: untyped): untyped =
+  block:
+    let
+      node = nodeParam
+      blockSlot = blockSlotParam
 
-  if isState(node.dag.headState):
-    withStateVars(node.dag.headState):
-      var cache {.inject, used.}: StateCache
-      body
-  else:
-    let rpcState = assignClone(node.dag.headState)
-    node.dag.withState(rpcState[], blockSlot):
-      body
+    template isState(state: StateData): bool =
+      state.blck.atSlot(getStateField(state.data, slot)) == blockSlot
+
+    var cache {.inject, used.}: StateCache
+
+    # TODO view-types
+    # Avoid the code bloat produced by the double `body` reference through a lent var
+    if isState(node.dag.headState):
+      withStateVars(node.dag.headState):
+        body
+    else:
+      let cachedState = if node.stateTtlCache != nil:
+        node.stateTtlCache.getClosestState(blockSlot)
+      else:
+        nil
+
+      let stateToAdvance = if cachedState != nil:
+        cachedState
+      else:
+        assignClone(node.dag.headState)
+
+      node.dag.updateStateData(stateToAdvance[], blockSlot, false, cache)
+
+      let providedState = if cachedState == nil and node.stateTtlCache != nil:
+        # This was not a cached state, we can cache it now
+        node.stateTtlCache.add(stateToAdvance)
+        stateToAdvance
+      else:
+        # If we have a cache hit, there is a concern that the REST request
+        # handler may continue executing asynchronously while we hit the same
+        # advanced state is another request. We don't want the two requests
+        # to work over the same state object because mutations to it will be
+        # visible in both, so we must create a fresh clone for each handler:
+        assignClone(stateToAdvance[])
+
+      withStateVars(providedState[]):
+        body
 
 proc toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
                                                           ValidatorIndexError] =
--- a/beacon_chain/spec/eth2_apis/rest_types.nim
+++ b/beacon_chain/spec/eth2_apis/rest_types.nim
@@ -219,6 +219,7 @@ type
 
   RestMetadata* = object
     seq_number*: string
+    syncnets*: string
     attnets*: string
 
   RestNetworkIdentity* = object
@@ -502,16 +503,19 @@ type
   GetPhase0BlockSszResponse* = phase0.SignedBeaconBlock
   GetAltairBlockSszResponse* = altair.SignedBeaconBlock
 
+  RestRoot* = object
+    root*: Eth2Digest
+
   # Types based on the OAPI yaml file - used in responses to requests
   GetAggregatedAttestationResponse* = DataEnclosedObject[Attestation]
   GetAttesterDutiesResponse* = DataRootEnclosedObject[seq[RestAttesterDuty]]
   GetBlockAttestationsResponse* = DataEnclosedObject[seq[Attestation]]
   GetBlockHeaderResponse* = DataEnclosedObject[RestBlockHeaderInfo]
   GetBlockHeadersResponse* = DataEnclosedObject[seq[RestBlockHeaderInfo]]
-  GetBlockRootResponse* = DataEnclosedObject[Eth2Digest]
+  GetBlockRootResponse* = DataEnclosedObject[RestRoot]
   GetDebugChainHeadsResponse* = DataEnclosedObject[seq[RestChainHead]]
   GetDepositContractResponse* = DataEnclosedObject[RestDepositContract]
-  GetEpochCommitteesResponse* = DataEnclosedObject[RestGenesis]
+  GetEpochCommitteesResponse* = DataEnclosedObject[seq[RestBeaconStatesCommittees]]
   GetForkScheduleResponse* = DataEnclosedObject[seq[Fork]]
   GetGenesisResponse* = DataEnclosedObject[RestGenesis]
   GetNetworkIdentityResponse* = DataEnclosedObject[RestNetworkIdentity]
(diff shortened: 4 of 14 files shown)


docs/the_nimbus_book/src/options.md
```diff
@@ -21,89 +21,75 @@ nimbus_beacon_node [OPTIONS]... command
 
 The following options are available:
 
-     --log-level               Sets the log level for process and topics (e.g. "DEBUG;
-                               TRACE:discv5,libp2p; REQUIRED:none; DISABLED:none") [=INFO].
-     --log-format              Specifies what kind of logs should be written to stdout (beta)
-                               (auto, colors, nocolors, json) [=auto].
-     --log-file                Specifies a path for the written Json log file.
+     --log-level               Sets the log level for process and topics (e.g. "DEBUG; TRACE:discv5,libp2p;
```
arnetheduck, line 24: the help text here depends on the width of the terminal you generate it on .. afair it's something in the 90-100 character range, to avoid these unnecessary changes

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -219,6 +219,7 @@ type
 
   RestMetadata* = object
     seq_number*: string
+    syncnets*: string
```
arnetheduck, line 222: afair not all servers serve this field - it would need to be read with lax rules

docs/the_nimbus_book/src/options.md
zah: I'm aware of this. There is a maximum width in Confutils and I think we should stick to using it. Otherwise, how is everyone supposed to find out the width used by the previous developer when generating the help text?

arnetheduck: well, write it down here, in a markdown comment? the point is to avoid a battle of the widths-scenario where every time each developer uses their favorite setting.. I don't _really_ care what the width is, except that it should fit on a reasonable browser.

wdyt of a `make update-help` command that is run as part of the build every time `conf.nim` changes?

beacon_chain/conf.nim
```diff
@@ -327,6 +327,16 @@ type
         defaultValueDesc: "127.0.0.1"
         name: "rest-address" }: ValidIpAddress
 
+      restTraversals* {.
```
```suggestion
      restCacheSize* {.
```

I think calling this "traversals" unnecessarily limits the feature and abstracts it from what actually is going on in a slightly misleading way - ie traversals is _one_ operation that gets faster, but so does re-querying the same info over and over - later, once can imagine expanding the use cases in other ways as the cache and db gets smarter

beacon_chain/rpc/rest_traversals_cache.nim
```diff
@@ -0,0 +1,99 @@
...
+
+logScope:
+  topics = "rest_traversals_cache"
+
+proc init*(T: type RestTraversalsCache,
+           cacheSize: Natural,
+           cacheTtl: Duration): T =
+  doAssert cacheSize > 0
+
+  RestTraversalsCache(
```
arnetheduck: ditto - `StateTtlCache` maybe?

beacon_chain/rpc/rest_utils.nim
```diff
@@ -116,19 +116,50 @@ proc getBlockDataFromBlockIdent*(node: BeaconNode,
...
+        node.restTraversalsCache.getClosestState(blockSlot.slot)
+      else:
+        nil
+
+      let stateToAdvance = if cachedTraversalState != nil:
+        cachedTraversalState
+      else:
+        assignClone(node.dag.headState)
+
+      node.dag.updateStateData(stateToAdvance[], blockSlot, false, cache)
```
arnetheduck, line 182: as of https://github.com/status-im/nimbus-eth2/pull/3217, this can fail - there are also some conflicting fixes to the rest api in that branch, as all state requests, as of the checkpoint feature, can fail

beacon_chain/rpc/rest_traversals_cache.nim
```diff
@@ -0,0 +1,99 @@
...
+  var
+    bestSlotDifference = Slot.high
+    index = -1
+
+  for i in 0 ..< cache.entries.len:
+    if cache.entries[i] == nil:
+      continue
+
+    let stateSlot = getStateField(cache.entries[i].state.data, slot)
+    if stateSlot > slot:
```
arnetheduck: see https://github.com/status-im/nimbus-eth2/blob/1a6a56bdb1aab0b1778e99cd1c5834180c4a52ca/beacon_chain/consensus_object_pools/blockchain_dag.nim#L867 - this condition is both not enough and too strong: an exact slot match is fine as long as the block matches (the block has not been applied at the given slot), but you also need to check that the state belongs to the same history in the dag - this is important in two cases: 
* a state id was used to point to an explicit "alternative" history
* a reorg happened between two requests and the "default" canonical history most often used in rest requests changed

the above is covered mostly by the `canAdvance` construct in `updateStateData`, which in the loop ensures that the correct history is being considered (in `ChainDAG`) - the loop in which it is used performs the same logic as is desired here.

arnetheduck:
I've keep wondering why `Implement #3129` as the main headline for a git commit makes sense - in nim-lang, this is also the culture, what's the .. idea? 

The log message is written _for_ the consumer of it, and when reading the history (`git log --oneline`), 3129 isn't meaningful / helpful - it seems to me that `Bla bla (implements #XXXX)` is more useful to readers?

beacon_chain/conf.nim
```diff
@@ -327,6 +327,16 @@ type
...
+        defaultValue: 10
```
arnetheduck: this is actually quite a heavy default - ie it's roughly 100mb per state afair - we might want to start with something like 1 or 2 that roughly corresponds to the current ceiling consumption of the REST component

beacon_chain/conf.nim
zah: Re-querying the same info doesn't get faster because we store the advanced state in the cache (traversals won't be really speed-up otherwise).

beacon_chain/rpc/rest_utils.nim
zah: This PR is simpler though, so it makes sense to merge it first perhaps. #3217 haven't been reviewed yet.

beacon_chain/conf.nim
zah: If you don't perform alternative traversals, you don't pay for anything. Also, the cache expires very quickly. The limit is put there only to ensure that this cannot be used as a DoS vector against the server (even though we know that the REST API is an effective DoS vector anyway). Putting the cap at 1GB doesn't look that high to me. Using lower values is certainly reasonable, but 1 or 2 is rather extreme IMO. Perhaps 3 to 5 is okay.

docs/the_nimbus_book/src/options.md
zah: This is a neat idea, but someone must spend the time implementing it.

beacon_chain/rpc/rest_traversals_cache.nim
zah: `canAdvance` doesn't check the right conditions. It checks that the state can be fast-forwarded by applying empty slots (it doesn't have intermediate blocks in between). 

The modification in chaindag made in this PR enables advancing even in the presence of intermediate blocks (they will be discovered in the loop gathering blocks that used to terminate at a database snapshot). 

You are right that checking that we are on the same branch in the history is necessary as well, but it's not necessary for correctness, only to ensure that the optimization actually brings benefits. Considering that this cache is intended for historic queries where alternative histories are not present, being overly strict in the conditions may be an overkill.

beacon_chain/conf.nim
zah: Actually, re-querying can be faster indeed. I'll make the suggested change.

docs/the_nimbus_book/src/options.md
arnetheduck: well, it's a matter of running the compiled binary, there's nothing to implement really - this problem of resizing the help text to different widths comes up pretty much every time I review help text updates - it's a bit of a churn - even without any automation, just using a simple redirect of `--help` should be enough - that'll automatically use whatever width confutils uses for non-terminals.

beacon_chain/rpc/rest_traversals_cache.nim
arnetheduck: > It checks that the state can be fast-forwarded by applying empty slots (it doesn't have intermediate blocks in between).

hm, taken together with the loop it runs in, it _should_ check for the right condition - the conditions are exactly the same for this cache and for the `updateStateData` - we should be checking that histories match and that slots are reasonable - in the case of `updateStateData` it's necessary for correctness - in the case of this cache, it's necessary to reach optimal efficiency - the logic is exactly the same.

arnetheduck: > Considering that this cache is intended for historic queries 

in the case of non-finality, historic queries quickly become "current queries" - ie there's no guarantee that all callers of this feature use it for historic queries - it should nonetheless reuse a reasonable cache entry if it has it - for example, it's likely calls to the rest API query the state at a given offset from head - say.. `head-6` slots - this gives callers the ability to reuse the state without loading from DB when following head at a "safe" distance that's likely to become final, similar to how exchanges today require N verifications.

beacon_chain/conf.nim
arnetheduck: Not only that: slot-by-slot queries that follow head at a fixed distance also become faster etc - there is nothing inherently limiting this feature to historical queries even if that is how it was conceived in this PR - a general cache like this has been part of the REST API plan for some time.

beacon_chain/conf.nim
```diff
@@ -327,6 +327,16 @@ type
...
+        desc: "The maximum number of expected concurrent historic traversals (API requests obtaining consecutive slot or epoch information)"
+        name: "rest-traversals" }: Natural
+
+      restTraversalsTtl* {.
+        defaultValue: 60
```
arnetheduck, line 339: this should probably be either the duration of "a few slots+some" or maybe an "epoch + some" so that queries at "regular" intervals tied to the constants in the spec "make sense"

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -936,8 +937,10 @@ proc updateStateData*(
 
     # Look for a state in the database and load it - as long as it cannot be
     # found, keep track of the blocks that are needed to reach it from the
-    # state that eventually will be found
-    while not dag.getState(state, cur):
+    # state that eventually will be found. Besides finding the state in the
+    # database we may also reach the input state provided to the function.
+    # It can also act as a viable starting point for the block replay later.
+    while cur.blck.root != state.blck.root and not dag.getState(state, cur):
```
arnetheduck: I don't quite follow this change - we've already verified above that the starting point is _not_ the right block - `getState` will only return true when the `cur.blck.root` == `state.blck.root` so when exactly is the extra check against `cur` needed?

arnetheduck: also, this will fail when state has been advanced _past_ the point of interest but matches block - ie if `bs.blck` is from slot 42 and we want to load `(blck, 100)`, we might have a `state` that's been advanced to `(blck, 105)` that falsely would match the new condition

arnetheduck: ok, I think I understand when this triggers - it's the case where the given state is older than whatever we look for above in the `canAdvance` loop but newer than whatever is in the database - just like everything else, this needs to use a `canAdvance` check to be correct

beacon_chain/rpc/rest_traversals_cache.nim
arnetheduck: > being overly strict

also, checking the history as well is not "overly strict" - it will never result in any false negatives - it will simply increase the efficiency of the cache in certain cases (because it'll select a better entry right when it matters the most: when the network is already under load)

beacon_chain/conf.nim
arnetheduck: well, you do pay: before this branch, querying non-historical, non-head states would have a cap of 1 BeaconState "extra memory used" no matter how many queries you made, in which order, and for what data - it could be very slow, but it would never consume more memory than that - now that cap goes up to whatever this setting is, for _general_ use of REST

beacon_chain/rpc/rest_traversals_cache.nim
```diff
@@ -0,0 +1,99 @@
...
+  chronos,
+  chronicles,
+  ../consensus_object_pools/block_pools_types
+
+type
+  CacheEntry = ref object
+    state: ref StateData
+    lastUsed: Moment
+
+  RestTraversalsCache* = ref object
```
arnetheduck: this looks like it doesn't really have to be `ref` - it can simply have 0 entries if the cache is disabled which simplifies the code quite a bit - it can simply be made to work correctly with 0 entries which will get rid of potential nils, doasserts and so on

docs/the_nimbus_book/src/options.md
zah: There will be some new Makefiles rules, there will be a solution for the problem of splicing the help output in the book, etc (is there something like `include` statement in mdBook?). I don't see a reason for postponing merging this PR until the suggestion is implemented.

beacon_chain/rpc/rest_utils.nim
arnetheduck: ok - in that case, let's remove the change in `blockchain_dag` from this PR which is incorrect afaict (and conflicting) - we can add back a correct version when both PR:s have made it through (or add it to the other PR)

docs/the_nimbus_book/src/options.md
arnetheduck: it's not a blocker indeed, but an obvious low-hanging fruit

beacon_chain/conf.nim
zah: How about 3 then?

arnetheduck: sure why not, it's not a bad number at all: https://en.wikipedia.org/wiki/3

beacon_chain/conf.nim
zah: An epoch is almost 7 minutes though, which is quite a significant deviation from this suggested default. If your requests are so rare, perhaps you won't benefit significantly from the cache. And indeed, using a larger TTL will put more weight on the suggestion for reducing the default cache size.

arnetheduck: yeah indeed - I'm thinking about the case where there's roughly one query per epoch (which feels kind of reasonable as far as use cases go, given this is how often the effective balance, proposers and so on change)

beacon_chain/rpc/rest_utils.nim
zah: I'm not sure I understand. We cannot remove the change in `blockchain_dag` as it is crucial for the optimization brought by this PR.

arnetheduck: well, it's wrong (it'll cause crashes) and conflicting (3217 adds error handling to this loop) so my idea was to remove it from this PR, then merge this PR, then introduce 3217 which should go in soon (it fixes a critical bug in unstable among other things), then we can make a separate change or change 3217 to include the (corrected) change. it's a matter of merge logistics, mostly.

zah: Well, it's also trivial to fix, but perhaps I'm misunderstanding. Are you referring to the new terminating condition in the block gathering loop in `updateStateData`? It's also wrong only in the sense that it can be misused by other API users in the future. Can you trigger a crash with the REST API as implemented in this PR?

arnetheduck: > new terminating condition

yes

arnetheduck: but whatever..

arnetheduck:
fwiw, most comments on the PR are of the "value-on-the-table" type, except the loop condition thing - with that fixed, we can merge it (but it would be a shame, the comments will end up forgotten)

beacon_chain/rpc/rest_utils.nim
arnetheduck: I put the correct change in 3217 as well so that it doesn't get lost in any merge order

beacon_chain/rpc/rest_traversals_cache.nim
zah: Needs to be a ref to facilitate capturing it in a closure later in this module. Comment added.

arnetheduck: for the other managers and pools, we've stopped using `ref object` and instead use `ref Xxx` where a reference is needed for async - doing so allows keeping the code tight with respect to `var` vs `non-var` access in functions - I'm planning on giving `ChainDAG` this treatment as well: this would prevent stuff like REST from mutating the caches when it shouldn't, for example.

I wish we had `ref-to-immutable-instance` - instead we got the really weird strict func which is a really odd and inflexible way of looking at the problem :/

beacon_chain/rpc/rest_utils.nim
```diff
@@ -116,19 +116,50 @@ proc getBlockDataFromBlockIdent*(node: BeaconNode,
...
-    let rpcState = assignClone(node.dag.headState)
-    node.dag.withState(rpcState[], blockSlot):
-      body
+    template isState(state: StateData): bool =
+      state.blck.atSlot(getStateField(state.data, slot)) == blockSlot
+
+    var cache {.inject, used.}: StateCache
+
+    # TODO view-types
+    # Avoid the code bloat produced by the double `body` reference through a lent var
```
arnetheduck, line 167: actually, a reasonable thing to do here might be to introduce a `proc runBody(state: StateData...)` - that would guarantee that the body doesn't mutate the state

zah: How would that work? `body` is a block in the caller scope. If you put it in a closure, you'll start getting all kinds of additional restrictions regarding what variables can be referenced within the body. The codegen will be different because some local variables will be moved to the heap, parameters may end up being copied, etc.

arnetheduck: yes, there are downsides :)

but yeah, view types (well.. immutable references with bounded lifetimes) would be the proper solution here - we should in fact also use clearancestate/epochrefstate, _and_ we should ensure that inside `body`, there are no await statements and other escapes.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -116,19 +116,50 @@ proc getBlockDataFromBlockIdent*(node: BeaconNode,
...
+
+      node.dag.updateStateData(stateToAdvance[], blockSlot, false, cache)
+
+      let providedState = if cachedState == nil and node.stateTtlCache != nil:
+        # This was not a cached state, we can cache it now
+        node.stateTtlCache.add(stateToAdvance)
+        stateToAdvance
+      else:
+        # If we have a cache hit, there is a concern that the REST request
+        # handler may continue executing asynchronously while we hit the same
```
arnetheduck: this must never happen - if we use `headState` with such a `body`, it will break because the head state can change while the body is running

arnetheduck (approved):

beacon_chain/rpc/rest_utils.nim
zah: It can be done with my `fauxClosure` magic, but I have some doubts that you would like this :P

A `fauxClosure` is something like a closure that must be executed immediately. What the magic does is to turn every captured variable into a regular parameter of a non-closure proc with the supplied body and then this proc is immediately called. I would say that there is a rather low risk of this producing invalid code - if there are bugs they are likely to result in compilation problems (either in Nim or in C). The mechanism has been used extensively by the status-desktop team.

Besides reducing the code bloat, the generated procedure will be effectively non-async, eliminating the yielding risk.

beacon_chain/rpc/rest_utils.nim
zah: Partial "no awaits" enforcement was implemented here:
https://github.com/status-im/nimbus-eth2/pull/3219/commits/524912fba58dff5020594666637dd961ee1b3f0f

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -609,6 +609,7 @@ proc getEpochRef*(dag: ChainDAGRef, blck: BlockRef, epoch: Epoch): EpochRef =
   let
     ancestor = epochAncestor(blck, epoch)
 
+  dag.epochRefState.blck = BlockRef()
```
arnetheduck, line 612: what's this?

zah: Since the state here is always an "output parameter" in intent, this nullifies the possibilities that it will be used as an input parameter due to the new change. At least some of the logic in the tests have to be changed otherwise.

arnetheduck: but .. it should be used as an input parameter? basically, if replaying from epochrefstate is the best thing we can do, why wouldn't we?

arnetheduck: not needed after all, will fix in error handling branch

PR #3223, opened 2021-12-29
Implementation of the Keymanager API (BETA)

zah:
https://github.com/ethereum/keymanager-APIs

### DIFF
--- a/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim
+++ b/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim
@@ -8,11 +8,16 @@
 
 import
   chronos, presto/client, chronicles,
-  ".."/[helpers, forks], ".."/datatypes/[phase0, altair, merge],
-  ".."/eth2_ssz_serialization,
-  "."/[rest_types, eth2_rest_serialization]
+  ".."/".."/validators/slashing_protection_common,
+  ".."/datatypes/[phase0, altair, merge],
+  ".."/[helpers, forks, keystore, eth2_ssz_serialization],
+  "."/[rest_types, rest_keymanager_types, eth2_rest_serialization]
 
 export chronos, client, rest_types, eth2_rest_serialization
+
+UUID.serializesAsBaseIn RestJson
+KeyPath.serializesAsBaseIn RestJson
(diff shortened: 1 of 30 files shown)


beacon_chain/spec/eth2_apis/rest_beacon_calls.nim
```diff
@@ -8,12 +8,17 @@
...
-  ".."/eth2_ssz_serialization,
-  "."/[rest_types, eth2_rest_serialization]
+  ".."/".."/validators/slashing_protection_common,
+  ".."/datatypes/[phase0, altair, merge],
+  ".."/[helpers, forks, keystore, eth2_ssz_serialization],
+  "."/[rest_types, rest_keymanager_types, eth2_rest_serialization]
 
 export chronos, client, rest_types, eth2_rest_serialization
 
+UUID.serializesAsBaseIn RestJson
```
arnetheduck, line 18: in general, this looks like a bad idea: it creates a false dependency between the formats where a change in the base type will break the REST api - the "base type" JSON serialization is not aware of `RestJson` existing creating a fragile setup - `RestJson` is very specific about using concrete and explicit serializers for all types so that you don't have to go to a different serializer to understand what the format will look like - it's an intentional separation of concerns which makes it easy to audit restjson for correctness: you can look at the spec and the serializer and compare without having to dig deeply or risk breakage during upgrades.

zah: `UUID` is a distinct string. It exists primarily in order to avoid potential mistakes such as passing an UUID in the position of another string parameter (e.g. the password or the description) in the various APIs for working with keystores.

The statement here means "The UUID type is serialized in the same way as its underlying string in the RestJson format".  You seem to be confusing it with a different statement such as "The UUID type is serialized in RestJson in the same way as in regular Json".

arnetheduck: ok - all good then - though yes, I did get confused, it's not quite clear which base is meant - the Json base of RestJson or the one of Uuid.. I find the connection between RestJson and Json hard to reason about in general I think, ie which parts of RestJson get infected with Json

PR #3226, opened 2021-12-30
`.era` file update

arnetheduck:
Overhaul of era files, including documentation and reference
implementations

* store blocks, then state, then slot indices for easy lookup at low
cost
* document era file rationale
* altair+ support in era writer


tersec (approved):
Adds documentation, tests, modifies `ncli`, and adds some functions which don't interfere with anything to some parts of the core application. Don't have strong opinions on the code, but given how harmless this is, better in than out on balance.

PR #3228, opened 2022-01-02
convert between engine and consensus ExecutionPayloads

tersec:
- general cleanup (e.g., template renaming to clarify in `eth1_monitor`)
- add support and tests for conversion between engine and consensus `ExecutionPayload`s; tests are verbosely written, but the tradeoff for avoiding what would be a much more compact metaprogramming approach, and even avoiding defining helper templates to create various `List`s, is that one can easily manually pick any subexpression and it'll work/be an example of how to construct said objects. This is a different tradeoff than might be worthwhile in non-test code.
- add kintsugi geth launching shell script
- 32-bit platform TTD override workaround from `kintsugi` branch; this should eventually be fixed in `nim-confutils`, but it's a pretty harmless workaround
- fix missing `extra_data` copy from `ExecutionPayload` to `ExecutionPayloadHeader` during block processing
- remove redundant merge spec constants from code datatypes module which exist in presets anyway


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -314,6 +323,60 @@ func asEth2Digest*(x: BlockHash): Eth2Digest =
 template asBlockHash(x: Eth2Digest): BlockHash =
   BlockHash(x.data)
 
+func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV1):
+    ref merge.ExecutionPayload =
```
arnetheduck: this should probably be determined on the calling side with `newClone`, akin to `BeaconState` - the CI now checks for excessive stacks on build, so we can safely rely on RVO

tersec: The CI will see the transactions as basically nested `seq`s. Will the CI static stack size checking pick that up? This isn't just a large static object (e.g., `BeaconState`, more or less).

arnetheduck: `seq` are heap-allocated

arnetheduck: the CI stack size check works at a C level, so it should be .. accurate enough - one caveat is that it's done per-function, so it doesn't detect when nesting adds up, but that's usually not a problem

tersec: https://github.com/status-im/nimbus-eth2/pull/3228/commits/48239a7f191be7c79fcc3c09fc4b5f7461c8fc92

PR #3230, opened 2022-01-03
support GOSSIP_MAX_SIZE_MERGE blocks; prevent fork choice stutter via aggregate attestations

tersec:
- the merge fork supports larger (10MB) blocks to accommodate transactions. It's not ideal that the message ID function doesn't tightly bound the message size, but that's baked into how libp2p works, conceptually and in terms of its current interfaces. It doesn't have any internal sense of the idea of an Ethereum fork, nor should it. It treats the topic, even though there is a syntax that it could conceivably be taught how to decode to determine the fork and therefore the max gossip size, as a black box. Meanwhile, this would somewhat pointlessly complicate its inner message-processing loops. While any of this could be changed, the design costs and layer mixing/leaky abstractions and excessive specialization for libp2p for Ethereum peculiarities resulting probably wouldn't be worthwhile; and

- the aggregated attestation validator wasn't properly using the post-async-timeskip walltime, so could set fork choice time backwards

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1844,7 +1844,7 @@ template gossipMaxSize(T: untyped): uint32 =
          T is altair.SignedBeaconBlock:
       GOSSIP_MAX_SIZE
     else:
-      static: raiseAssert false
+      static: raiseAssert "unknown type"
   static: doAssert maxSize <= maxGossipMaxSize()
   maxSize.uint32
 
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1822,18 +1822,31 @@ proc newBeaconSwitch*(config: BeaconNodeConf, seckey: PrivateKey,
     .withTcpTransport({ServerFlags.ReuseAddr})
     .build()
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/phase0/p2p-interface.md#configuration
-# https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/p2p-interface.md#configuration
-template getMaxGossipMaxSize(): auto =
+# https://github.com/ethereum/consensus-specs/blob/v1.1.7/specs/phase0/p2p-interface.md#configuration
+# https://github.com/ethereum/consensus-specs/blob/v1.1.7/specs/merge/p2p-interface.md#configuration
+func maxGossipMaxSize(): auto {.compileTime.} =
   max(GOSSIP_MAX_SIZE, GOSSIP_MAX_SIZE_MERGE)
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/phase0/p2p-interface.md#configuration
-# https://github.com/ethereum/consensus-specs/blob/v1.1.6/specs/merge/p2p-interface.md#configuration
-template getForkGossipMaxSize(msgFork: BeaconBlockFork): auto =
-  if msgFork >= BeaconBlockFork.Merge:
-    GOSSIP_MAX_SIZE_MERGE
-  else:
-    GOSSIP_MAX_SIZE
+# https://github.com/ethereum/consensus-specs/blob/v1.1.7/specs/phase0/p2p-interface.md#configuration
+# https://github.com/ethereum/consensus-specs/blob/v1.1.7/specs/merge/p2p-interface.md#configuration
+template gossipMaxSize(T: untyped): uint32 =
+  const maxSize = static:
+    when isFixedSize(T):
+      fixedPortionSize(T)
+    elif T is merge.SignedBeaconBlock:
+      GOSSIP_MAX_SIZE_MERGE
+    # TODO https://github.com/status-im/nim-ssz-serialization/issues/20 for
+    # Attestation, AttesterSlashing, and SignedAggregateAndProof, which all
+    # have lists bounded at MAX_VALIDATORS_PER_COMMITTEE (2048) items, thus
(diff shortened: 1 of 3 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -1969,7 +1985,8 @@ proc addValidator*[MsgType](node: Eth2Node,
     inc nbc_gossip_messages_received
     trace "Validating incoming gossip message", len = message.data.len, topic
 
-    var decompressed = snappy.decode(message.data, GOSSIP_MAX_SIZE)
+    var decompressed =
+      snappy.decode(message.data, getForkGossipMaxSize(msgFork))
```
arnetheduck: while we're here, it wouldn't be a bad idea to implement a hardened version of this: on the attestation and sync committee topics, message sizes are actually bounded to the static size of `MsgType` - anything else is junk / cannot be valid - what we can do is to have a function `gossipMaxSize(T: type Attestation): sszFixedSize(T)` and so on - for "dynamic" types like block, we'd return the `GOSSIP_MAX..` constant instead.

arnetheduck: also, the max size, we need to give it to libp2p as well, there's a toggle in the gossipsub parameter init, else it will use a default of 1mb

arnetheduck: the latter works for the _compressed_ size, while this check works for the uncompressed size - I suspect the spec refers to the former: this constant is needed because libp2p needs to know how many bytes to read from the stream before giving up - when it's doing that, it doesn't (yet) know the topic of the message.

That said, a similar limit for the uncompressed size makes sense - I don't know if there are blowups in snappy, but if there are, we get rid of them this way.

tersec: Attestations, both aggregated and unaggregated, contain:
https://github.com/status-im/nimbus-eth2/blob/d4680df8d2d147cac8f091e4898b89c29f1883ab/beacon_chain/spec/datatypes/base.nim#L204

which, properly, fail the `isFixedSize(T)` check and for which `fixedPortionSize(T)` doesn't accurately cover their size.

tersec: One can make manual adjustments, outside the realm of the existing counting mechanism, but at that point the premise seems dicey. Since SSZ was moved out of the `nimbus-eth2` repo, there's no "while we're here" anymore to modifying that at the same time.

arnetheduck: huh, right, forgot about that one.. bummer - that said, `MAX_VALIDATORS_PER_COMMITTEE` is known at compile time so that provides an effective upper limit to the size - not as easily accessible though - on the flip side, if we build a generic "max-possible-size" helper for ssz based on the `List` limits, we could use that for blocks as well perhaps.

tersec: Sure, `AttesterSlashing` contains another non-fixed-size but also quite bounded list, for example:
https://github.com/status-im/nimbus-eth2/blob/d4680df8d2d147cac8f091e4898b89c29f1883ab/beacon_chain/spec/datatypes/base.nim#L190-L194

And the only gossip message which actually needs the full `GOSSIP_MAX_SIZE_MERGE` size is `merge.SignedBeaconBlock`, so a reasonably safe thing is to keep the rest at `GOSSIP_MAX_SIZE`, to at least not regress elsewhere.

`ProposerSlashing`, `SignedContributionAndProof`, `SignedVoluntaryExit`, and `SyncCommitteeMessage` are all truly fixed-size, so they're worth starting with to bound tightly.

arnetheduck: https://github.com/status-im/nim-ssz-serialization/issues/20

tersec: https://github.com/status-im/nimbus-eth2/pull/3230/commits/918f1216b5b2f34aa6bda99bff3dd7fb0e1da955

beacon_chain/networking/eth2_network.nim
```diff
@@ -1822,6 +1822,32 @@ proc newBeaconSwitch*(config: BeaconNodeConf, seckey: PrivateKey,
...
+    # TODO https://github.com/status-im/nim-ssz-serialization/issues/20 for
+    # Attestation, AttesterSlashing, and SignedAggregateAndProof, which all
+    # have lists bounded at MAX_VALIDATORS_PER_COMMITTEE (2048) items, thus
+    # having max sizes significantly smaller than GOSSIP_MAX_SIZE.
+    elif T is Attestation or T is AttesterSlashing or
+         T is SignedAggregateAndProof or T is phase0.SignedBeaconBlock or
+         T is altair.SignedBeaconBlock:
+      GOSSIP_MAX_SIZE
+    else:
+      raiseAssert false
```
```suggestion
      static: raiseAssert false
```

arnetheduck (approved):

beacon_chain/networking/eth2_network.nim
```diff
@@ -1822,6 +1822,32 @@ proc newBeaconSwitch*(config: BeaconNodeConf, seckey: PrivateKey,
...
+    # TODO https://github.com/status-im/nim-ssz-serialization/issues/20 for
+    # Attestation, AttesterSlashing, and SignedAggregateAndProof, which all
+    # have lists bounded at MAX_VALIDATORS_PER_COMMITTEE (2048) items, thus
+    # having max sizes significantly smaller than GOSSIP_MAX_SIZE.
+    elif T is Attestation or T is AttesterSlashing or
+         T is SignedAggregateAndProof or T is phase0.SignedBeaconBlock or
+         T is altair.SignedBeaconBlock:
+      GOSSIP_MAX_SIZE
+    else:
+      static: raiseAssert false
```
```suggestion
      static: raiseAssert "unknown type"
```

I guess passing incorrect parameter types is good also as compile-time check :)

PR #3232, opened 2022-01-03
REST JSON support improvements

arnetheduck:
* support downloading blocks / states via JSON in addition to SSZ -
slow, but needed for infura support - SSZ is still used when server
supports it
* use common forked block/state reader in REST API
* fix stack overflows in REST JSON decoder
* fix invalid serialization of `justification_bits` in
`/eth/v1/debug/beacon/states` and `/eth/v2/debug/beacon/states`
* fix REST client to use `/eth/...` instead of `/api/eth/...`, update
"default" urls to expose REST api via `/eth` as well as this is what the
standard says - `/api` was added early on based on an example "base url"
in the spec that has been removed since
* expose Nimbus REST extensions via `/nimbus` in addition to
`/api/nimbus` to stay consistent with `/eth`
* fix invalid state root when reading states via REST
* fix recursive imports in `spec/ssz_codec`
* remove usages of `serialization.useCustomSerialization` - fickle

### DIFF
--- a/beacon_chain/spec/eth2_apis/rest_debug_calls.nim
+++ b/beacon_chain/spec/eth2_apis/rest_debug_calls.nim
@@ -78,8 +78,10 @@ proc getStateV2Plain*(state_id: StateIdent): RestPlainResponse {.
   ## https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2
 
 proc getStateV2*(client: RestClientRef, state_id: StateIdent,
-                 forks: array[2, Fork],
-                 restAccept = ""): Future[ForkedHashedBeaconState] {.async.} =
+                 cfg: RuntimeConfig,
+                 restAccept = ""): Future[ref ForkedHashedBeaconState] {.async.} =
+  # nil is returned if the state is not found due to a 404 - `ref` is needed
+  # to manage stack usage
   let resp =
     if len(restAccept) > 0:
       await client.getStateV2Plain(state_id, restAcceptType = restAccept)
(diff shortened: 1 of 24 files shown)


beacon_chain/spec/eth2_apis/rest_debug_calls.nim
```diff
@@ -92,42 +94,22 @@ proc getStateV2*(client: RestClientRef, state_id: StateIdent,
       of "application/json":
         let state =
           block:
-            let res = decodeBytes(GetStateV2Response, resp.data,
-                                  resp.contentType)
-            if res.isErr():
-              raise newException(RestError, $res.error())
-            res.get()
+            let res = newClone(decodeBytes(GetStateV2Response, resp.data,
```
zah, line 97: We desperately need a way to detect this at compile-time. A patch in the compiler would be quite simple to pull off.

zah (approved):

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
arnetheduck: you mean the stack size? we are detecting it now, in C (at compile time) - the code unused so you only get the error when it actually gets compiled, and that doesn't happen until the trusted node sync branch. Sure, it would  be nice to have it in nim as well, but `C` has the advantage that it's more accurate - Nim generates _a lot_ of unnecessary temporaries that it would have trouble counting (because they're generated in the C backend randomly and differently for c/c++).


there were 2 things causing stack overflows in this particular case: before https://github.com/nim-lang/Nim/commit/8a3f542532ff442096bf6172bea3e4abba5918fe it placed a lot of temporaries on the stack for `randao_mixes` because the json deserializer for HashArray didn't use RVO _and_ it didn't use RVO properly for the beacon state either.
Above all, because they are bugs, nim wouldn't have counted them properly either - it wouldn't notice.

The latter is something that still needs fixing - the RVO support in nim is quite lacking compared to that of rust/c++.
There are two more things that need to happen as well:
* `lent` support in `Result`
* lifetime tracking in nim: `let a = b; let b = a.x` should not allocate a new stack entry for b, but rather reuse the one in a as an alias - I'm not sure to which extent this is done already.

zah: Do we check this in CI now? I've pushed a test branch here:
https://github.com/status-im/nimbus-eth2/pull/3251

I was able to compile it locally without problems and it produced a crash as expected.

arnetheduck: Yes, see https://github.com/status-im/nimbus-eth2/blob/8242e57f41e40de89b6d2a68772b3bd7754866bf/config.nims#L43

zah: Seems to work :)

PR #3236, opened 2022-01-03
use v1.1.8 test vectors

tersec:
https://github.com/ethereum/consensus-specs/releases/tag/v1.1.8
https://github.com/ethereum/consensus-spec-tests/releases/tag/v1.1.8

No changes outside `tests/`, the resulting markdown reports, and `spec/datatypes/base` where it updates the target version.

There's a lot of churn here which amounts to replacing the name "Merge" with its new name, "Bellatrix", but that's it.

It has some workarounds for stray `DS_Store` files which interfered with its directory-walking, which will be removed when feasible.


arnetheduck:
at some point we should really cut down `Ethereum Foundation` to just `EF`

tersec:
> at some point we should really cut down `Ethereum Foundation` to just `EF`

https://github.com/status-im/nimbus-eth2/pull/3242

PR #3241, opened 2022-01-04
Tune getLowSubnets

Menduist:
cc @arnetheduck

Changes the getLowSubnets strategy to be more aggressive on peers subscriptions (vs mesh health)

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1035,7 +1035,7 @@ proc getLowSubnets(node: Eth2Node, epoch: Epoch): (AttnetBits, SyncnetBits) =
   # Returns the subnets required to have a healthy mesh
   # The subnets are computed, to, in order:
   # - Have 0 subnet with < `dLow` peers from topic subscription
-  # - Have 0 subscribed subnet below `d`
+  # - Have 0 subscribed subnet below `dLow`
   # - Have 0 subscribed subnet below `dOut` outgoing peers
   # - Have 0 subnet with < `dHigh` peers from topic subscription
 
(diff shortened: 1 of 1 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -1034,41 +1034,48 @@ proc trimConnections(node: Eth2Node, count: int) {.async.} =
...
 
+      if node.pubsub.gossipsub.peers(topic) < node.pubsub.parameters.dHigh:
+        notHighOutgoingSubnets.setBit(subNetId)
+
       # Not subscribed
       if topic notin node.pubsub.mesh: continue
 
-      if node.pubsub.mesh.peers(topic) < node.pubsub.parameters.dLow:
-        belowDlowSubnets.setBit(subNetId)
+      if node.pubsub.mesh.peers(topic) < node.pubsub.parameters.d:
```
```suggestion
      if node.pubsub.mesh.peers(topic) < node.pubsub.parameters.dLow:
```

`low` here is needed or we'll go into perpetual connect - if there are `low` peers already in the mesh, the mesh won't be grown but we'll keep connecting

beacon_chain/networking/eth2_network.nim
```diff
@@ -1034,41 +1034,48 @@ proc trimConnections(node: Eth2Node, count: int) {.async.} =
 proc getLowSubnets(node: Eth2Node, epoch: Epoch): (AttnetBits, SyncnetBits) =
   # Returns the subnets required to have a healthy mesh
   # The subnets are computed, to, in order:
-  # - Have 0 subscribed subnet below `dLow`
-  # - Have 0 subnet with < `d` peers from topic subscription
+  # - Have 0 subnet with < `dLow` peers from topic subscription
+  # - Have 0 subscribed subnet below `d`
```
Menduist:
```suggestion
  # - Have 0 subscribed subnet below `dLow`
```

PR #3248, opened 2022-01-05
Support more content-types when specifying the Graffiti string

zah:


beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -165,12 +165,32 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+                                         $graffitiBytes.error)
+      node.graffitiBytes = graffitiBytes.get
       return RestApiResponse.jsonResponse((result: true))
+    of "text/plain":
+      return node.setGraffitiAux contentBody.get.strData
+    of "application/x-www-form-urlencoded":
+      return node.setGraffitiAux decodeUrl(contentBody.get.strData)
     else:
-      return RestApiResponse.jsonError(Http400, InvalidGraffitiBytesValye)
+      return RestApiResponse.jsonError(Http400, $contentBody.get.contentType)
```
arnetheduck: contentType?

zah: Hadn't noticed that in my review, thanks.

PR #3253, opened 2022-01-06
spec URL updates

tersec:
Automated with https://github.com/status-im/nimbus-eth2/blob/unstable/scripts/find_unchanged_consensus_spec_files.sh

Each before/after URL points to the exact same content.

All comments. No non-comments. No semantic changes.


tersec:
Not actually all comments, because built from wrong base commit.

PR #3259, opened 2022-01-08
Harden CommitteeIndex, SubnetId, SyncSubcommitteeIndex

arnetheduck:
Harden the use of `CommitteeIndex` et al to prevent future issues by
using a distinct type, then validating before use in several cases -
datatypes in spec are kept simple though so that invalid data still can
be read.

* fix invalid epoch used in REST
`/eth/v1/beacon/states/{state_id}/committees` committee length (could
return invalid data)
* normalize some variable names
* normalize committee index loops
* fix `RestAttesterDuty` to use `uint64` for `validator_committee_index`
* validate `CommitteeIndex` on ingress in REST API

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -79,7 +79,7 @@ func check_attestation_block(
   ok()
 
 func check_propagation_slot_range(
-    msgSlot: Slot, wallTime: BeaconTime): Result[void, ValidationError] =
+    msgSlot: Slot, wallTime: BeaconTime): Result[Slot, ValidationError] =
   let
     futureSlot = (wallTime + MAXIMUM_GOSSIP_CLOCK_DISPARITY).toSlot()
 
@@ -99,7 +99,7 @@ func check_propagation_slot_range(
       msgSlot + ATTESTATION_PROPAGATION_SLOT_RANGE < pastSlot.slot:
     return errIgnore("Attestation slot in the past")
 
-  ok()
+  ok(msgSlot)
 
 func check_beacon_and_target_block(
     pool: var AttestationPool, data: AttestationData):
@@ -149,12 +149,11 @@ func check_aggregation_count(
   ok()
 
 func check_attestation_subnet(
-    epochRef: EpochRef, attestation: Attestation,
+    epochRef: EpochRef, slot: Slot, committee_index: CommitteeIndex,
     subnet_id: SubnetId): Result[void, ValidationError] =
   let
     expectedSubnet = compute_subnet_for_attestation(
-      get_committee_count_per_slot(epochRef),
-      attestation.data.slot, attestation.data.index.CommitteeIndex)
+      get_committee_count_per_slot(epochRef), slot, committee_index)
 
   if expectedSubnet != subnet_id:
     return errReject("Attestation not on the correct subnet")
@@ -380,10 +379,11 @@ proc validateAttestation*(
   # [REJECT] The attestation's epoch matches its target -- i.e.
   # attestation.data.target.epoch ==
   # compute_epoch_at_slot(attestation.data.slot)
-  block:
+  let slot = block:
     let v = check_attestation_slot_target(attestation.data)
     if v.isErr():
       return errReject(v.error())
+    v.get()
 
   # attestation.data.slot is within the last ATTESTATION_PROPAGATION_SLOT_RANGE
   # slots (within a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e.
@@ -391,7 +391,7 @@ proc validateAttestation*(
   # >= attestation.data.slot (a client MAY queue future attestations for
   # processing at the appropriate slot).
   block:
-    let v = check_propagation_slot_range(attestation.data.slot, wallTime) # [IGNORE]
+    let v = check_propagation_slot_range(slot, wallTime) # [IGNORE]
     if v.isErr():
       return err(v.error())
 
@@ -434,8 +434,11 @@ proc validateAttestation*(
 
   # [REJECT] The committee index is within the expected range -- i.e.
   # data.index < get_committee_count_per_slot(state, data.target.epoch).
-  if not (attestation.data.index < get_committee_count_per_slot(epochRef)):
-    return checkedReject("Attestation: committee index not within expected range")
+  let committee_index = block:
+    let idx = epochRef.get_committee_index(attestation.data.index)
+    if idx.isErr():
+      return checkedReject("Attestation: committee index not within expected range")
+    idx.get()
 
   # [REJECT] The attestation is for the correct subnet -- i.e.
   # compute_subnet_for_attestation(committees_per_slot,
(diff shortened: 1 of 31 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -890,8 +895,11 @@ proc validateContribution*(
 
   # [REJECT] The subcommittee index is in the allowed range
   # i.e. contribution.subcommittee_index < SYNC_COMMITTEE_SUBNET_COUNT.
-  let subcommitteeIdx = msg.message.contribution.subcommittee_index.validateSyncCommitteeIndexOr:
-    return errReject("SignedContributionAndProof: subcommittee index too high")
+  let subcommitteeIdx = block:
+    let v = SyncSubcommitteeIndex.init(msg.message.contribution.subcommittee_index)
```
zah, line 899: Meh, people can learn to read more elegant code. 

I guess we can restore the conciseness with the `either` helper form the error-handling proposal, but the `validateOr` pattern also produces shorter code.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -488,14 +488,18 @@ proc readValue*(reader: var JsonReader[RestJson],
 ## CommitteeIndex
 proc writeValue*(writer: var JsonWriter[RestJson], value: CommitteeIndex) {.
      raises: [IOError, Defect].} =
-  writeValue(writer, Base10.toString(uint64(value)))
+  writeValue(writer, Base10.toString(uint8(value)))
```
zah: I still believe that avoiding "unchecked" type coercions has value and should be pursued in general in the project. The definitions such as `asUInt8` (or the alternative proposed syntax `value as uint8`) act as approved coercion gates that are documented and can be challenged in the future if any of the involved types change. Once you are remove some of the approved coercions, the compiler will indicate all the places in the code that need fixing (something that is not true with the regular coercions).

beacon_chain/validators/validator_duties.nim
```diff
@@ -902,46 +901,45 @@ proc sendAggregatedAttestations(
...
         slotSigs.add getSlotSig(validator, fork,
           genesis_validators_root, slot)
-        slotSigsData.add (committee_index, validatorIdx, validator)
+        slotSigsData.add (committee_index, validator_index, validator)
 
   await allFutures(slotSigs)
 
-  for curr in zip(slotSigsData, slotSigs):
-    let slotSig = curr[1].read()
+  for data, fut in zip(slotSigsData, slotSigs).items():
```
zah: Why was `items` added here? And how does `zip` work? Does it create a temporary object by allocating memory?

zah (approved):

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: `validateSyncCommitteeIndexOr` doesn't generalise / compose, one has to come up with a new template for every type which is annoying (just like it's annoying that iterators don't compose in nim) - I tried it on CommitteeIndex but it was a pita to introduce consistently - in this code in particular, the problem is not `validateXxx` vs `block:` style, but rather that `?` does not (yet) work with `async` - if it did, this code would simply be `let subCommitteeIdx = ? SyncSubCommittee.init(...).orErr(Reject)` or some spiritual equivalent..

beacon_chain/validators/validator_duties.nim
arnetheduck: `zip` creates a temporary seq indeed, afair, as nim iterators compose poorly (or nobody figured out how to write `zip` generically without collecting the result) - `pairs` is needed or `for` will call `pairs` on the `seq` that `zip` returns (unpacking `data` into an `int`).

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
arnetheduck: Yeah, I thought about turning all these `distinct` types into `type Xxx = object\n  data: uint8` which would force the matter, but that would turn an already-massive PR into a monster - the test cases are all about coercions - I might attempt that as a follow-up PR.

beacon_chain/validators/validator_duties.nim
zah: I thought the fancy for-loop macros were introduced to solve this `zip` iterator problem, but considering that we are iterating over regular sequences here, using `zip` looks like completely unnecessary pessimisation. Would be better to just use a loop with an index.

arnetheduck: the std lib is rarely updated to actually use the features in the language - I wish that for every feature that's added, there would be a mandatory review of the std lib as well, the onus lying with the person that makes the language addition: this would be healthy for many reasons - the breakage cost would be borne together with the feature introduction, minimal usefulness testing on the feature would happen naturally and there would be fewer shit features that look like good ideas but nobody actually uses.


cc @Araq not a bad idea for the Nim PR guidelines :)

zah: It's also possible to create a local definition of an efficient `zip` iterator that takes `openArrays` instead of other iterators.

arnetheduck: I don't disagree, in theory, but... - it's a bit outside the scope of this PR though, also because creating the signature dwarfs whatever `zip` does by a factor of 1:10000..

zah: Using a loop with an index is a two lines change. Claiming it's out of scope would suggest that a separate PR should be filed for it which seems a bit silly.

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: https://github.com/status-im/nim-stew/pull/96/commits/9bcba3ade870c30d92e4614461380a56f94179a6 - adds a version of `valueOr` that generally can be used instead of creating a new template every time.

beacon_chain/validators/validator_duties.nim
arnetheduck: Yes, but it's two lines that use an inferior form of looping that (in theory) introduces the potential for a defect (zip checks the computes a minimum length of both seqs), _and_ it's performance noise, ie nobody's ever going to notice the difference except the reader of the code that will have more to parse .. but .. sure, I can change it.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
arnetheduck: in the meantime: 85ce9915da817b62b2dd52c6a9f1129991de6860

beacon_chain/validators/validator_duties.nim
arnetheduck: hahaha, you asked for it: https://github.com/status-im/nimbus-eth2/pull/3259/commits/65a3c68671fbd68a051c7545da6792dfa4dddb4d

Araq: > the breakage cost would be borne together with the feature introduction, minimal usefulness testing on the feature would happen naturally and there would be fewer shit features that look like good ideas but nobody actually uses.

Yeah well... I have no idea how to introduce a `zip` via a for loop macro that would **not** break the existing API. And I don't like the constant churn, leave the old stuff alone (and maybe deprecate it) and introduce new things via packages.

PR #3260, opened 2022-01-08
better web3monitor error message

arnetheduck:
```
WRN 2022-01-08 11:35:00.963+01:00 Eth1 chain monitoring failure,
restarting  topics="eth1" err="Failed to setup web3 connection: (111)
Connection refused"
```

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -864,7 +864,10 @@ proc new*(T: type Web3DataProvider,
   yield web3Fut or sleepAsync(chronos.seconds(10))
   if (not web3Fut.finished) or web3Fut.failed:
     await cancelAndWait(web3Fut)
-    return err "Failed to setup web3 connection"
+    if web3Fut.failed:
+      return err "Failed to setup web3 connection: " & web3Fut.readError.msg
+    else:
+      return err "Failed to setup web3 connection"
 
   let
     web3 = web3Fut.read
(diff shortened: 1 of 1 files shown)


tersec (approved):

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1141,11 +1144,12 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
   info "Starting Eth1 deposit contract monitoring",
     contract = $m.depositContractAddress, url = web3Url
 
-  let dataProviderRes = await Web3DataProvider.new(
-    m.depositContractAddress,
-    web3Url)
+  m.dataProvider = block:
+    let v = await Web3DataProvider.new(m.depositContractAddress, web3Url)
+    if v.isErr():
```
zah, line 1149: Perhaps this should be the default behavior of `tryGet`. 

I could argue that the information that `tryGet` currently inserts ("Trying to access value with err" ) is already encoded in the exception type and thus redundant. It also uses developer jargon, so it's clearly not intended or useful for end-users. If the message was preserved precisely as it's done here, the API would facilitate the transport of error messages intended for end users.

arnetheduck: Hmm, yeah - so what `Result` does is `when compiles $(...): "..." & $(error)` - I guess it's not unreasonable to use the string conversion directly, specially because the exception _also_ carries an instance of the error .. :+1:

arnetheduck: the alternative would be to only translate "known" types, or introduce a `Result`-specific extension point like `toException`, that by default would call `$` but would allow callers to customize the string conversion of the error .. hm, maybe that's overkill..

arnetheduck: https://github.com/status-im/nim-stew/pull/96/commits/9bcba3ade870c30d92e4614461380a56f94179a6

PR #3263, opened 2022-01-09
Backfiller

arnetheduck:
Backfilling is the process of downloading historical blocks via P2P that
are required to fulfill `GetBlocksByRange` duties - this happens during
both trusted node and finalized checkpoint syncs.

In particular, backfilling happens after syncing to head, such that
attestation work can start as soon as possible.

This PR is captures the part of https://github.com/status-im/nimbus-eth2/pull/3209 that still needs more work.

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -89,9 +89,24 @@ proc speed*(start, finish: SyncMoment): float {.inline.} =
     slots / dur
 
 proc initQueue[A, B](man: SyncManager[A, B]) =
-  man.queue = SyncQueue.init(A, man.direction, man.getFirstSlot(),
-                              man.getLastSlot(), man.chunkSize,
-                              man.getSafeSlot, man.blockVerifier, 1)
+  case man.direction
+  of SyncQueueKind.Forward:
+    man.queue = SyncQueue.init(A, man.direction, man.getFirstSlot(),
+                               man.getLastSlot(), man.chunkSize,
+                               man.getSafeSlot, man.blockVerifier, 1)
+  of SyncQueueKind.Backward:
+    let
+      firstSlot = man.getFirstSlot()
+      lastSlot = man.getLastSlot()
+      startSlot = if firstSlot == lastSlot:
+                    # This case should never be happened in real life because
+                    # there is present check `needsBackfill().
+                    firstSlot
+                  else:
+                    Slot(firstSlot - 1'u64)
+    man.queue = SyncQueue.init(A, man.direction, firstSlot, lastSlot,
+                               man.chunkSize, man.getSafeSlot,
+                               man.blockVerifier, 1)
 
 proc newSyncManager*[A, B](pool: PeerPool[A, B],
                            direction: SyncQueueKind,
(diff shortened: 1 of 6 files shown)


beacon_chain/sync/sync_manager.nim
```diff
@@ -390,26 +415,47 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
...
+        await man.notInSyncEvent.wait()
+        man.workers[index].status = SyncWorkerStatus.WaitingPeer
+        peer = await man.pool.acquire()
+        await man.syncStep(index, peer)
+        man.pool.release(peer)
+        false
+      except CancelledError:
+        if not(isNil(peer)):
+          man.pool.release(peer)
+        true
```
arnetheduck, line 449: why not simple.. break here?

beacon_chain/sync/sync_manager.nim
```diff
@@ -138,35 +155,56 @@ proc getBlocks*[A, B](man: SyncManager[A, B], peer: A,
...
+      debug "Error, while reading getBlocks response",
+            peer = peer, slot = req.slot, count = req.count,
+            step = req.step, peer_speed = peer.netKbps(),
+            direction = man.direction, error = $res.error(),
+            topics = "syncman"
+      return
+    let forked =
+      res.map() do (blcks: seq[phase0.SignedBeaconBlock]) -> auto:
+        blcks.mapIt(ForkedSignedBeaconBlock.init(it))
+    return forked
```
mratsim, line 210: I think a template is worth it, so many wasted lines/mindspace

```Nim
template debugLog(msg: string): untyped {.dirty.} =
  debug msg, 
            peer = peer, slot = req.slot, count = req.count,
            step = req.step, peer_speed = peer.netKbps(),
            direction = man.direction, error = $res.error(),
            topics = "syncman"
```

beacon_chain/sync/sync_manager.nim
```diff
@@ -303,10 +341,6 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
           peer_speed = peer.netKbps(), index = index,
           direction = man.direction, topics = "syncman"
     peer.updateScore(PeerScoreUseless)
-
-    # Give the peer time to do some syncing
-    await sleepAsync(StatusExpirationTime div 2)
```
mratsim: How do we avoid dos-ing peers?

beacon_chain/sync/sync_manager.nim
```diff
@@ -381,26 +415,47 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
...
+      except CancelledError:
+        if not(isNil(peer)):
+          man.pool.release(peer)
+        true
+      except CatchableError as exc:
+        debug "Unexpected exception in sync worker",
+              peer = peer, index = index,
+              peer_score = peer.getScore(), peer_speed = peer.netKbps(),
+              errName = exc.name, errMsg = exc.msg, direction = man.direction,
+              topics = "syncman"
```
mratsim, line 455: I think the sync_manager needs a debugLog template to avoid littering the codebase with those. (though that might cause issue with `filename:linenumber` logic from Chronicles?

beacon_chain/sync/sync_manager.nim
cheatfate: Why do you think we DOS-ing peers? We descored peer which is not useful for us and we return it back to the PeerPool. So it is not supposed to be returned for next step.

beacon_chain/sync/sync_manager.nim
cheatfate: Just because I do not want to investigate one more `for/while+exception+break` issue.

arnetheduck: it wouldn't be bad to add tests for this in chronos - basically, a few simple for/while/finally etc sanity tests

beacon_chain/sync/sync_manager.nim
arnetheduck: typically, this is solved with `logScope`, but logscope is unfortunately unreliable (cc @zah can we do anything about that?)

Issue #3264, opened 2022-01-10
Limit execution time in jenkins


arnetheduck:
https://github.com/status-im/nimbus-eth2/pull/3263 introduces a bug in one of the test cases that causes the test suite to go into an infinite loop - https://github.com/status-im/nimbus-eth2/runs/4755088898?check_suite_focus=true gets cancelled after 360 minutes which is reasonable (but could be made shorter, say .. 180 minutes?)

Our Jenkins CI however never times out and instead takes up a spot in the job queue until manually cancelled (like this build that I cancelled after 9 hours: https://ci.status.im/job/nimbus/job/nimbus-eth2/job/multibranch/job/chaindag-backfill/1/)

PR #1341, opened 2020-07-16
refactor Jenkinsfile to be a pipeline, add more stages

jakubgs:
The purpose of this is to:

- Separate builds by platform so logs from `macos` and `linux` hosts don't get mixed
- Divide shell commands into as many stages as possible to ease debugging
- Generalize running testnets via a `launchLocalTestnet()` groovy method
- Handle uploading of results of running testnets stage-by-stage basis

This is mostly done, I just have to do some additional testing with CI.

Builds: https://ci.status.im/job/nimbus/job/nim-beacon-chain/job/linux/

PR #3266, opened 2022-01-10
ci: add 3 hour timeout for jobs

jakubgs:
If we had a pipeline we could just specify this in options:
https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#timeout-enforce-time-limit
```Jenkinsfile
  options {
      timeout(time: 3, unit: 'HOURS') 
  }
```
But as is we need to do more levels of nesting.

Resolves: https://github.com/status-im/nimbus-eth2/issues/3264

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -94,18 +94,22 @@ def runStages() {
 parallel(
 	"Linux": {
 		throttle(['nimbus-eth2']) {
-			node("linux") {
-				withEnv(["NPROC=${sh(returnStdout: true, script: 'nproc').trim()}"]) {
-					runStages()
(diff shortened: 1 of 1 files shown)


arnetheduck:
> If we had a pipeline 

why don't we have a pipeline? (whatever that is...)

jakubgs:
A pipeline is a format for writing Jenkinsfiles: https://www.jenkins.io/doc/book/pipeline/#declarative-pipeline-fundamentals

My attempt to implement it was rejected.

tersec:
https://github.com/status-im/nimbus-eth2/pull/1341?

jakubgs:
Yes.

PR #3270, opened 2022-01-11
Don't use GC memory for the initial beacon block summaries loading

zah:
This is an alternative to https://github.com/status-im/nimbus-eth2/pull/3164
Uses `malloc` and `free` behind the scenes.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -8,7 +8,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[typetraits, tables],
+  std/[typetraits, sharedtables],
   stew/[arrayops, assign2, byteutils, endians2, io2, objects, results],
   serialization, chronicles, snappy,
   eth/db/[kvstore, kvstore_sqlite3],
@@ -882,21 +882,35 @@ iterator getAncestors*(db: BeaconChainDB, root: Eth2Digest):
     yield res
     root = res.message.parent_root
 
-proc loadSummaries*(db: BeaconChainDB): Table[Eth2Digest, BeaconBlockSummary] =
+# BEWARE
+# There is an opportunistic hack here. Since we are allocating a large table
+# that is used briefly during start-up, we would like to avoid involving the
+# Nim allocator because at the moment it doesn't ever return allocated pages
+# to the OS. Once you reach a certain "high watermark" of used memory, the
+# memory consumption will stay the same until the end of the program.
+# The allocation here is large enough to leave such a lasting effect on the
+# program as demonstrated in:
+# https://github.com/status-im/nimbus-eth2/pull/3164
+#
+# By using `SharedTable`, we avoid the problem because it uses `malloc` and
+# `free` behind the scenes.
+#
+# TODO The hack we'll be removed in the future after further reforms in the
+# DAG that will make it unnecessary to populate such a table altogether.
+proc loadSummaries(db: BeaconChainDB): SharedTable[Eth2Digest, BeaconBlockSummary] =
   # Load summaries into table - there's no telling what order they're in so we
   # load them all - bugs in nim prevent this code from living in the iterator.
-  var summaries = initTable[Eth2Digest, BeaconBlockSummary](1024*1024)
+  result.init 1024*1024
+  let resultAddr = addr result # used only locally in the closure
 
   discard db.summaries.find([], proc(k, v: openArray[byte]) =
     var output: BeaconBlockSummary
 
-    if k.len() == sizeof(Eth2Digest) and decodeSSz(v, output):
-      summaries[Eth2Digest(data: toArray(sizeof(Eth2Digest), k))] = output
+    if k.len() == sizeof(Eth2Digest) and decodeSSZ(v, output):
+      resultAddr[][Eth2Digest(data: toArray(sizeof(Eth2Digest), k))] = output
     else:
       warn "Invalid summary in database", klen = k.len(), vlen = v.len()
   )
-
-  summaries
 
 type RootedSummary = tuple[root: Eth2Digest, summary: BeaconBlockSummary]
 iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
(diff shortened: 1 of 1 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -8,7 +8,7 @@
 {.push raises: [Defect].}
 
 import
-  std/[typetraits, tables],
+  std/[typetraits, sharedtables],
```
arnetheduck, line 11: wait, why sharedtables? that's for threading.. aren't you thinking of `TableRef`?

arnetheduck: also the `shared*` collections are giant hacks, some of which are already deprecated (and I hope the rest get thrown out as well) - I'm not at all comfortable putting trust in them - they're not used anywhere and for good reason

zah: The `sharedtables` are just like the normal tables, but they use `malloc` and `free` to manage their memory. The actual hash table insertion and lookup algorithms are extracted in templates that are shared between the two table types, so I'd say the risk of issues is rather low.

Futhermore, this is a temporary solution until the feature is removed altogether with the introduction of ERA files. Since this is not likely to happen soon enough, I think it's beneficial to take this intermediate step in the interest of our users. [External quality](https://meekrosoft.wordpress.com/2010/10/31/internal-and-external-software-quality/) is more important than internal quality in my opinion.

arnetheduck: >  just like

there's also locking in there, which moves us into risky territory - also, I'm not _quite_ sure what the benefit here really is (that isn't solved by meeting our minimum requirements  _or_ using a swap file and letting the OS take care of it which is what it's there for). 

also, progress on this front isn't blocked by era files: it actually makes sense to build a `DbSeq` with a linear history in the database and use that instead of the summaries for loading the chaindag - that's more memory-efficient, faster than any of these approaches, useful for trusted node sync and the solution will remain in place even in a post-era world - the logic for the linear finalized block lookup has recently been implemented in ChainDAG for other reasons (`ChainDAG.finalisedBlocks`) - see https://github.com/status-im/nimbus-eth2/pull/3164#issuecomment-1003307601 - this is actually _really simple_.

zah: > benefit

The `malloc` and `free` calls are handled by the C run-time and they don't suffer from Nim's problem of not returning allocated memory to the OS.

> risky territory

There is a limit to the usefulness of these type of fear-of-the-unknown arguments. At some point one has to spend some time reading through the code to find out if the concerns are justified. The extra locking and unlocking is implemented with direct calls to system APIs such as `pthread_mutex_lock` and `EnterCriticalSection`. If you can describe a plausible scenario leading to problems, I'm all ears.

> really simple

There is still time to submit alternative PRs (or we can merge this as an incremental improvement of the status quo and then further improve it in a next PR that removes the "risky" code). It would be a shame though if our next release doesn't bring the full potential of memory savings brought by the Chronos reforms.

arnetheduck: > describe a plausible scenario leading to problems,

On windows, calling `EnterCriticalSection` freezes the application if the lock is not properly initialized (zero-init does not cut it), or when memory gets overwritten at the wrong point in time - we've been here with nimbus already.

> full potential of memory savings brought by the Chronos reforms.

The chronos savings are entirely orthogonal to the gains seen here - we would see the same benefits had we not made the chronos changes - conversely, the chronos changes bring about an improvement regardless if we fix this issue also.

If we're going to go ahead with this ugly hack, it's appropriate to test it more thoroughly through a long run on the fleet, I believe.

arnetheduck: btw, there's a lower-hanging and better tested fruit that gives you half the benefit - just use a `TableRef` (this avoids an unnecessary copy of the entire table caused by nim codegen bugs)

zah: > On windows, calling EnterCriticalSection freezes the application if the lock is not properly initialized (zero-init does not cut it), or when memory gets overwritten at the wrong point in time - we've been here with nimbus already.

Well, the lock is properly initialized in the `init` call on the first line of `loadSummaries`.

There is an interplay between the Chronos savings and the "high watermark" behavior of the Nim allocator and this fix eliminates the negative consequences. You can argue that the benefit is only in perceived memory usage due to the fact that the OS will swap cold memory pages, but sometimes perception can be more important than reality.

arnetheduck: each of the two improvements takes down the "high watermark" a notch independently

beacon_chain/beacon_chain_db.nim
```diff
@@ -882,22 +882,21 @@ iterator getAncestors*(db: BeaconChainDB, root: Eth2Digest):
     yield res
     root = res.message.parent_root
 
-proc loadSummaries*(db: BeaconChainDB): Table[Eth2Digest, BeaconBlockSummary] =
+proc loadSummaries(db: BeaconChainDB): SharedTable[Eth2Digest, BeaconBlockSummary] =
```
arnetheduck, line 900: this is quite the hack and should be documented as such - there's is no intent here to share this table across threads

zah: Hack documented above.

arnetheduck (approved):

beacon_chain/beacon_chain_db.nim
```diff
@@ -918,6 +932,9 @@ iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
     blck: phase0.TrustedSignedBeaconBlock
     newSummaries: seq[RootedSummary]
 
+  defer:
+    deinitSharedTable summaries
```
mratsim, line 936: What's the latest on defer?

It used to (or still have) many bugs, especially in an async context.

zah: The bugs that we knew about were fixed. They were affecting closure iterators.

PR #3277, opened 2022-01-12
Slashing prot interchange tests v5.2.1

mratsim:
### Support v5.1.0 (June), v5.2.0 (September) and v5.2.1 (December) test suite.

Similar to Teku and Lighthouse, this adds support for a form of "minification".

Lighthouse summarizes all imported and existing attestations in one giant synthetic attestation.
In our case we only create a synthetic attestation on conflicting surrounded/surrounding votes import.

Minification is important to make imports not order dependent, see https://github.com/eth-clients/slashing-protection-interchange-tests/releases/tag/v5.1.0-alpha.1

2 tests are skipped:

- The test https://github.com/eth-clients/slashing-protection-interchange-tests/blob/48141ad/tests/generated/multiple_interchanges_single_validator_multiple_blocks_out_of_order.json relies on a new assumption, that if a block checked for slashability is OK, we import it. It's the only test that relies on this behavior: https://github.com/eth-clients/slashing-protection-interchange-tests/pull/12#issuecomment-1011158701
- The test https://github.com/eth-clients/slashing-protection-interchange-tests/blob/48141ad/tests/generated/single_validator_source_greater_than_target_surrounding.json relies on the DB importing an invalid attestation with source < target, breaking invariants.
  There are 3 potential ways to solve this:
  - Test specs mention that if no minification is done, subsequent tests should be ignored if `contains_slashable_data` is set. However in our case, we do minification only for surround votes, i.e. when 2 attestations conflict, we don't do minification when an attestation is inconsistent with itself. Ignoring subsequent tests mean we don't test minification at all.
  - It seems like Lighthouse is creating a synthetic attestation that violates the `source < target` invariant.
  - Or we create synthetic attestation with target_epoch = max(source_epoch, target_epoch)

### Remove Slashing protection v1

Unless someone didn't validate since the Altair hardfork, they all migrated to v2.


mratsim:
unrelated 32-bit failure 

image

beacon_chain/validators/slashing_protection_v2.nim
```diff
@@ -1212,6 +1235,40 @@ proc pruneAfterFinalization*(
 # Interchange
 # --------------------------------------------
 
+proc registerSyntheticAttestation*(
+       db: SlashingProtectionDB_v2,
+       validator: ValidatorPubKey,
+       source, target: Epoch) =
+  ## Add a synthetic attestation to the slashing protection DB
+  doAssert source < target
```
arnetheduck, line 1243: this looks like it could crash the application with a malicious attestation - why not return a `Result[void, cstring]` instead? Mainly as a precaution should the function get used with less validated data

beacon_chain/validators/slashing_protection.nim
```diff
@@ -90,36 +89,13 @@ proc init*(
...
-      info "Slashing DB migration successful."
-    of siPartial:
-      warn "Slashing DB migration is a partial success."
-    of siFailure:
-      fatal "Slashing DB migration failure. Aborting to protect validators."
-      quit 1
-
-    db_v1.close()
+    fatal "The slashing database predates Altair hardfork from October 2021." &
+      " You can migrate to the new DB format using Nimbus 1.5.5" &
```
arnetheduck:
```suggestion
      " You can migrate to the new DB format using Nimbus 1.6.0" & 
```

beacon_chain/validators/slashing_protection.nim
```diff
@@ -90,36 +89,13 @@ proc init*(
...
-    of siPartial:
-      warn "Slashing DB migration is a partial success."
-    of siFailure:
-      fatal "Slashing DB migration failure. Aborting to protect validators."
-      quit 1
-
-    db_v1.close()
+    fatal "The slashing database predates Altair hardfork from October 2021." &
+      " You can migrate to the new DB format using Nimbus 1.5.5" & 
+      " for a few minutes at https://github.com/status-im/nimbus-eth2/releases/tag/v1.5.5" &
```
arnetheduck:
```suggestion
      " for a few minutes at https://github.com/status-im/nimbus-eth2/releases/tag/v1.6.0" &
```

beacon_chain/validators/slashing_protection_v2.nim
mratsim: That's not possible unless it's dev misuse, and there are such tests in the test suite.

When checking attestation for errors, you can get one of the following: https://github.com/status-im/nimbus-eth2/blob/8409b30/beacon_chain/validators/slashing_protection_common.nim#L100-L101

```Nim
  BadVoteKind* = enum
    ## Attestation bad vote kind
    # h: height (i.e. epoch for attestation, slot for blocks)
    # t: target
    # s: source
    # 1: existing attestations
    # 2: candidate attestation

    # Spec slashing condition
    DoubleVote   # h(t1) == h(t2)
    SurroundVote # h(s1) < h(s2) < h(t2) < h(t1) or h(s2) < h(s1) < h(t1) < h(t2)

    # Non-spec, should never happen in a well functioning client
    TargetPrecedesSource # h(t1) < h(s1) - current epoch precedes last justified epoch

    # EIP-3067 (https://eips.ethereum.org/EIPS/eip-3076)
    MinSourceViolation   # h(s2) < h(s1) - EIP3067 condition 4 (strict inequality)
    MinTargetViolation   # h(t2) <= h(t1) - EIP3067 condition 5
    DatabaseError          # Cannot read/write the slashing protection db
```

the synthetic attestation is created in this code path:

https://github.com/status-im/nimbus-eth2/blob/8409b30/beacon_chain/validators/slashing_protection_common.nim#L362-L417

```Nim
      let status = db.registerAttestation(
        parsedKey,
        A.source_epoch.Epoch,
        A.target_epoch.Epoch,
        A.signing_root.Eth2Digest
      )
      if status.isErr():
        # We might be importing a duplicate which EIP-3076 allows
        # there is no reason during normal operation to integrate
        # a duplicate so checkSlashableAttestation would have rejected it.
        # We special-case that for imports.
        if status.error.kind == DoubleVote and
            A.signing_root.Eth2Digest != ZeroDigest and
            status.error.existingAttestation == A.signing_root.Eth2Digest:
          warn "Attestation already exists in the DB",
            pubkey = spdir.data[v].pubkey.PubKeyBytes.toHex(),
            candidateAttestation = A,
            conflict = status.error()
          continue

        elif status.error.kind == SurroundVote:
          doAssert A.source_epoch.Epoch == status.error.sourceSlashable
          doAssert A.target_epoch.Epoch == status.error.targetSlashable

          # Formal proof of correctness: https://github.com/michaelsproul/slashing-proofs
          let synth = SPDIR_SignedAttestation(
            source_epoch: EpochString max(status.error.sourceSlashable, status.error.sourceExisting),
            target_epoch: EpochString max(status.error.targetSlashable, status.error.targetExisting)
          )

          warn "Slashable surround vote. Constructing a synthetic attestation to reconcile DB and import",
            pubkey = spdir.data[v].pubkey.PubKeyBytes.toHex(),
            candidateAttestation = A,
            conflict = status.error(),
            syntheticAttestation = synth
          
          db.registerSyntheticAttestation(
            parsedKey,
            synth.source_epoch.Epoch,
            synth.target_epoch.Epoch
          )

          if synth.source_epoch.int > maxValidSourceEpochSeen:
            maxValidSourceEpochSeen = synth.source_epoch.int
          if synth.target_epoch.int > maxValidTargetEpochSeen:
            maxValidTargetEpochSeen = synth.target_epoch.int

          result = siPartial
          continue
        else:
          error "Slashable vote. Skipping its import.",
            pubkey = spdir.data[v].pubkey.PubKeyBytes.toHex(),
            candidateAttestation = A,
            conflict = status.error()
          result = siPartial
          continue
```

It is not reachable if source > target, in the latter case the attestation will just be discarded.

Also DB import is done from cold boot using nbc as a CLI tool. A long-running process would not exercise this codepath.

arnetheduck (approved):

PR #3279, opened 2022-01-13
allow Eth1 monitor to run without genesis_deposit_contract_snapshot.ssz

tersec:
Allows Eth1 monitor not only to use a provided deposit contract, but fetch it dynamically from a deposit contract address. This is useful for test networks such as kintsugi which might not have a bundled `genesis_deposit_contract_snapshot.ssz` in their network definition directory.

For example:
```
eth2-networks/shared$ ls -1 {pyrmont,prater,mainnet}/{deposit_contract_block.txt,deposit_contract.txt,genesis_deposit_contract_snapshot.ssz}
mainnet/deposit_contract_block.txt
mainnet/deposit_contract.txt
mainnet/genesis_deposit_contract_snapshot.ssz
prater/deposit_contract_block.txt
prater/deposit_contract.txt
prater/genesis_deposit_contract_snapshot.ssz
pyrmont/deposit_contract_block.txt
pyrmont/deposit_contract.txt
pyrmont/genesis_deposit_contract_snapshot.ssz
```

https://github.com/eth-clients/merge-testnets/tree/main/kintsugi, however, does not have this `genesis_deposit_contract_snapshot.ssz` which Nimbus otherwise would read directly, and without which the Eth1 monitor won't ideally function.

Post-merge (not just Bellatrix, but actual-The-Merge) a Web3 connection is necessary for block processing and proposal. Part of this involves specifically monitoring the Eth1/EL chain for blocks satisfying certain criteria (such as total difficulty, or a special block hash) signifying The Merge, and part asking the EL for validity of embedded `ExecutionPayload`s in Eth2/CL blocks.

Furthermore, for purposes of having a single policy point in place for choosing, say, between round-robin and primary/backup approaches to multiple Web3 URLs and Web3 URL failovers, and to ensure the Eth1 monitor and block processing/proposal Web3 connections have the same view of the network, it's useful not to simply duplicate this connection elsewhere, but to centralize management in the component which already monitors the Eth1/EL chain.

Therefore, allow the Eth1/EL chain monitor to find the genesis deposit contract itself, if it's not provided in the network definition, so that it can run on test networks such as kintsugi in a way compatible with implementing other Web3 connections required for The Merge.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -868,11 +868,35 @@ proc init*(T: type Eth1Chain, cfg: RuntimeConfig, db: BeaconChainDB): T =
     finalizedBlockHash: finalizedDeposits.eth1Block,
     finalizedDepositsMerkleizer: finalizedDeposits.createMerkleizer)
 
+proc createInitialDepositSnapshot*(
+    depositContractAddress: Eth1Address,
+    depositContractDeployedAt: BlockHashOrNumber,
+    web3Url: string): Future[Result[DepositContractSnapshot, string]] {.async.} =
+
+  let dataProviderRes =
+    await Web3DataProvider.new(depositContractAddress, web3Url)
+  if dataProviderRes.isErr:
+    return err(dataProviderRes.error)
+  var dataProvider = dataProviderRes.get
+
+  let knownStartBlockHash =
+    if depositContractDeployedAt.isHash:
+      depositContractDeployedAt.hash
+    else:
+      try:
+        var blk = awaitWithRetries(
+          dataProvider.getBlockByNumber(depositContractDeployedAt.number))
+        blk.hash.asEth2Digest
+      except CatchableError as err:
+        return err(err.msg)
+
+  return ok DepositContractSnapshot(eth1Block: knownStartBlockHash)
+
 proc init*(T: type Eth1Monitor,
            cfg: RuntimeConfig,
            db: BeaconChainDB,
            web3Urls: seq[string],
-           depositContractSnapshot: DepositContractSnapshot,
+           depositContractSnapshot: Option[DepositContractSnapshot],
            eth1Network: Option[Eth1Network],
            forcePolling: bool): T =
   doAssert web3Urls.len > 0
@@ -881,7 +905,8 @@ proc init*(T: type Eth1Monitor,
   for url in mitems(web3Urls):
     fixupWeb3Urls url
 
-  putInitialDepositContractSnapshot(db, depositContractSnapshot)
+  if depositContractSnapshot.isSome:
+    putInitialDepositContractSnapshot(db, depositContractSnapshot.get)
 
   T(state: Initialized,
     depositsChain: Eth1Chain.init(cfg, db),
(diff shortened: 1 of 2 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -937,6 +962,19 @@ proc resetState(m: Eth1Monitor) {.async.} =
     await m.dataProvider.close()
     m.dataProvider = nil
 
+proc ensureDataProvider*(m: Eth1Monitor) {.async.} =
+  if not m.dataProvider.isNil:
+    return
```
zah, line 941: Is this attempting to fix something? Previously, the `m.dataProvider` was being reset on every restart of the monitor, but now you are keeping it around between the restarts. It's known that the web3 client can enter invalid states, so resetting it has been a tactic to deal with potential problems in the upstream libraries.

tersec: It's to accommodate
https://github.com/status-im/nimbus-eth2/blob/1db04636c19427c06af844ebfd459a672ad59312/beacon_chain/gossip_processing/block_processor.nim#L302-L308
and
https://github.com/status-im/nimbus-eth2/blob/1db04636c19427c06af844ebfd459a672ad59312/beacon_chain/gossip_processing/block_processor.nim#L355-L365
(in another sort of staging branch for what's coming; not everything there's fully polished yet, but it's closer than in `kintsugi`)

The Eth1 monitor resetting the connection causes problems elsewhere. The `kintsugi` branch works around this by just not reusing the same connection, but see the discussion in this PR for why it's probably worth having the Eth1 monitor actually manage the web3 connections. Without something like `ensureDataProvider` of a couple of things happen, a disproportionate number of engine API calls fail because behind their backs, the eth1 monitor has decided that `getBlockByHash` on the finalized eth1 chain block with hash presumably-uninitialized-0 failing is reason to recycle the connection.

They are not failing because, in some organic way, the web3 connection failed, and, well, it's the Internet, need to deal with it. They're failing because the eth1 monitor decided that, presumably as part of the kind of workaround for upstream libraries you describe, it's better to recycle the connection. That's fine when the eth1 monitor's the only consumer of that websocket, but that's not going to be true anymore.

I'm open to more systematic refactorings, but as a whole, having a quasi-idempotent `ensureDataConnection` seems like a decent approach. It should be cheap, that is, it is important that it be callable before every single engine API call without worrying that it'll induce websocket connection recycling.

One general technique might be to try to identify more distinctly the failure causes this is attempting to identify.

Or, I can just switch back to using a separate websocket system, and leave eth1 monitor's main syncing loop alone for the moment (it'll come back when supporting validating).

tersec: Another approach, I guess, is for the await-with-retries template to check on failure whether the first action it needs to take is reconnect the the websocket, but that has architectural tradeoffs; maybe worthwhile.

I should note that branch does check, ultimately, in the non-eth1 monitor functions whether at actual call time, to avoid TOCTOU issues which might crop up due to chronos scheduling, eth1 monitor still managed in that brief window to recycle the websocket connection. If it has, well, it just fails the call. But that's a lot rarer now than it used to be without `ensureDataProvider`, where it happened ever 10-30 engine API calls on average during syncing.

Or,
https://github.com/status-im/nimbus-eth2/blob/47f1f7ff1a12c951904c220ba6d92b0739d7d4bd/beacon_chain/eth1/eth1_monitor.nim#L1240-L1242
provides another option of where one might put the connection reset/`nil`ing: keep the defensive programming aspect, but do it not in what this PR would call `ensureDataProvider`.

zah: My main concern is that I'm not convinced the code would be able to recover from all possible error states as I described above, so it's not quite ready for merging in its current state. The last idea of nullifying the `dataProvider` right next to `m.state = Failed` could work perhaps.

tersec: https://github.com/status-im/nimbus-eth2/pull/3279/commits/923abf1e5f26868d4e15631d7f7fde17984a0dbf

PR #3282, opened 2022-01-13
bump nim-web3 to use enums for engine API types

tersec:
https://github.com/status-im/nim-web3/compare/2b9e8166d444441847de3e8b35d442e540d356db...0b5466f56ecc45a1e3c039abcccaf6a5e8250e18 is the main change of interest for the purposes of this PR.


arnetheduck:
I guess there isn't reasonable "default" behaviour in case of an unknown enum string?

tersec:
The closest is `"SYNCING"`/`syncing`, for both, which the Nimbus-side handlers mostly treat as maybe-and-retry-later, but it swallows information too soon. It's the kind of equivalency a higher-level handler can reasonably draw, but not a lower-level parser. Certainly not `valid`, `invalid`, or `success`.

PR #3284, opened 2022-01-14
move `state_transition` to `Result`

arnetheduck:
* better error messages in api
* avoid `BlockData` copies when replaying blocks

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -810,9 +810,8 @@ proc getForkedBlock*(dag: ChainDAGRef, id: BlockId): Opt[ForkedTrustedSignedBeac
       return ok ForkedTrustedSignedBeaconBlock.init(data.get)
 
 proc getForkedBlock*(dag: ChainDAGRef, blck: BlockRef): ForkedTrustedSignedBeaconBlock =
-  let blck = dag.getForkedBlock(blck.bid)
-  if blck.isSome():
-    return blck.get()
+  dag.getForkedBlock(blck.bid).expect(
+    "BlockRef block should always load, database corrupt?")
 
 proc get*(dag: ChainDAGRef, blck: BlockRef): BlockData =
   ## Retrieve the associated block body of a block reference
@@ -841,10 +840,9 @@ proc advanceSlots(
     let preEpoch = getStateField(state.data, slot).epoch
     loadStateCache(dag, cache, state.blck, getStateField(state.data, slot).epoch)
 
-    doAssert process_slots(
-        dag.cfg, state.data, getStateField(state.data, slot) + 1, cache, info,
-        dag.updateFlags),
-      "process_slots shouldn't fail when state slot is correct"
+    process_slots(
+      dag.cfg, state.data, getStateField(state.data, slot) + 1, cache, info,
+      dag.updateFlags).expect("process_slots shouldn't fail when state slot is correct")
     if save:
       dag.putState(state)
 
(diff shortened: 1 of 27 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -860,28 +858,36 @@ proc advanceSlots(
...
+  of BeaconBlockFork.Bellatrix:
+    let data = dag.db.getMergeBlock(blck.root).expect("block loaded")
     state_transition(
-      dag.cfg, state.data, blck, cache, info,
-      flags + dag.updateFlags + {slotProcessed}, restore)
-  if ok:
-    state.blck = blck.refs
+      dag.cfg, state.data, data, cache, info,
+      flags + dag.updateFlags + {slotProcessed}, noRollback).expect(
+        "Blocks from database must not fail to apply")
```
mratsim, line 888: Maybe we need a template `withForkedBlock`.

As we get more hard forks, it will be a pain to hunt all the dispatch locations (i.e. the expression problem of variants vs classes, https://eli.thegreenplace.net/2016/the-expression-problem-and-its-solutions/)

beacon_chain/spec/state_transition.nim
```diff
@@ -237,51 +205,36 @@ proc process_slots*(
...
+    cache: var StateCache, flags: UpdateFlags): Result[void, cstring] =
   # Block updates - these happen when there's a new block being suggested
   # by the block proposer. Every actor in the network will update its state
   # according to the contents of this block - but first they will validate
   # that the block is sane.
-  if not (skipBLSValidation in flags or
-      verify_block_signature(state.data, signedBlock)):
-    return false
+  if skipBLSValidation notin flags:
+    ? verify_block_signature(state.data, signedBlock)
```
mratsim, line 220: unrelated to the PR, it might not be necessary now for testing :eyes:

beacon_chain/spec/state_transition.nim
```diff
@@ -310,21 +259,20 @@ proc state_transition_block*(
   # Ensure state_transition_block()-only callers trigger this
   maybeUpgradeStateToAltair(cfg, state)
 
-  let success = withState(state):
+  let res = withState(state):
     state_transition_block_aux(cfg, state, signedBlock, cache, flags)
 
-  if not success:
+  if res.isErr():
```
mratsim, line 265: We need logging in case of failure.

The new changes allow reporting the deeper error, say a `process_block_header: slashing`.
But it is swallowed and never logged as there is no log anymore in:
- process_block_header
- process_block
- state_transition_block_aux
- state_transition_block
- state_transition
- applyBlock (in blockchain_dag)

This will make debugging failed state_transition really hard.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: yeah, it did cross my mind, but there's more refactoring coming up for these parts so I didn't want to pollute the codebase with yet another template while the dust is still settling

beacon_chain/spec/state_transition.nim
arnetheduck: In all callers of `state_transition_*` we either panic (via `expect` in `applyBlock`) or log at different levels - for example when producing a block, we log at `ERROR` level because this should never fail - when processing an incoming block, it's `INF` for now which is a bit of a concession to reality: it does open up some potential for `inf` spam, but on the other hand, by the block gets to the state transition point, its signature has already been verified so there are only so many blocks that can make it this far.

beacon_chain/spec/state_transition.nim
```diff
@@ -403,20 +351,15 @@ proc makeBeaconBlock*(
...
-    return err("Unable to apply new block to state: " & $res.error())
+    return err(res.error())
 
   state.root = hash_tree_root(state.data)
   blck.state_root = state.root
+
   ok(blck)
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#preparing-a-beaconblock
+# https://git3hub.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#preparing-a-beaconblock
```
arnetheduck:
```suggestion
# https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#preparing-a-beaconblock
```

PR #3293, opened 2022-01-17
limit by-root requests to non-finalized blocks

arnetheduck:
Presently, we keep a mapping from block root to `BlockRef` in memory -
this has simplified reasoning about the dag, but is not sustainable with
the chain growing.

We can distinguish between two cases where by-root access is useful:

* unfinalized blocks - this is where the beacon chain is operating
generally, by validating incoming data as interesting for future fork
choice decisions - bounded by the length of the unfinalized period
* finalized blocks - historical access in the REST API etc - no bounds,
really

In this PR, we limit the by-root block index to the first use case:
finalized chain data can more efficiently be addressed by slot number.

Future work includes:

* limiting the `BlockRef` horizon in general - each instance is 40
bytes+overhead which adds up - this needs further refactoring to deal
with the tail vs state problem
* persisting the finalized slot-to-hash index - this one also keeps
growing unbounded (albeit slowly)

Anyway, this PR easily shaves ~128mb of memory usage at the time of
writing.

* No longer honor `BeaconBlocksByRoot` requests outside of the
non-finalized period - previously, Nimbus would generously return any
block through this libp2p request - per the spec, finalized blocks
should be fetched via `BeaconBlocksByRange` instead.
* return `Opt[BlockRef]` instead of `nil` when blocks can't be found -
this becomes a lot more common now and thus deserves more attention
* `dag.blocks` -> `dag.forkBlocks` - this index only carries unfinalized
blocks from now - `finalizedBlocks` covers the other `BlockRef`
instances
* in backfill, verify that the last backfilled block leads back to
genesis, or panic
* add backfill timings to log
* fix missing check that `BlockRef` block can be fetched with
`getForkedBlock` reliably
* shortcut doppelganger check when feature is not enabled
* in REST/JSON-RPC, fetch blocks without involving `BlockRef`

Screenshot from 2022-01-17 14-04-38

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -8,17 +8,12 @@
 {.push raises: [Defect].}
 
 import
-  std/tables,
   chronicles,
   stew/[assign2, results],
-  eth/keys,
-  ../spec/[
-    eth2_merkleization, forks, helpers, signatures, signatures_batch,
-    state_transition],
-  ../spec/datatypes/[phase0, altair],
-  "."/[blockchain_dag]
+  ../spec/[forks, signatures, signatures_batch, state_transition],
+  "."/[block_dag, blockchain_dag]
 
-export results, signatures_batch
+export results, signatures_batch, block_dag, blockchain_dag
 
 # Clearance
 # ---------------------------------------------
@@ -50,7 +45,7 @@ proc addResolvedHeadBlock(
 
   link(parent, blockRef)
 
-  dag.blocks.incl(KeyedBlockRef.init(blockRef))
+  dag.forkBlocks.incl(KeyedBlockRef.init(blockRef))
 
   # Resolved blocks should be stored in database
   dag.putBlock(trustedBlock)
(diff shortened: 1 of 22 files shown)


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -160,59 +155,52 @@ proc addHeadBlock*(
...
 
-  if parent == nil:
-    debug "Block parent unknown"
+  let parent = dag.getBlockRef(blck.parent_root).valueOr:
+    # There are two cases where the parent won't be found: we don't have it or
+    # it has been finalized already, and as a result the branch the new block
+    # is on is no longer a viable fork candidate - we can't tell which is which
+    # at this stage, but we can check if we've seen the parent block previously
+    # and thus prevent requests for it to be downloaded again.
+    if dag.db.containsBlock(blck.parent_root):
```
zah, line 189: We should probably have a metric for the number of database queries we perform, so we can see when changes in the DAG logic have an effect on them.

zah (approved):

beacon_chain/consensus_object_pools/block_clearance.nim
arnetheduck: yeah, I'm not too happy about this check in general - for gossip, it's fine because it happens after signature verification, but I need to think a bit more about what happens during sync - sync is in fact the main issue: if we return `MissingParent` here, it will go look for it which would be bad.

PR #3300, opened 2022-01-19
Update start-syncing.md

lbly:
I found this documentation a little confusing because it says "you need to" leading me to think there may be some kind of dire consequences for NOT having it synced before starting the validator (in the context of importing existing keys from another setup, not starting a new validator with a new deposit).  Modifying the language to make the consequences of not having the syncing done clear.

### DIFF
--- a/docs/the_nimbus_book/src/start-syncing.md
+++ b/docs/the_nimbus_book/src/start-syncing.md
@@ -1,9 +1,9 @@
 
 # Start syncing
 
-If you're joining a network that has already launched, you need to ensure that your beacon node is [completely synced](./keep-an-eye.md#keep-track-of-your-syncing-progress) before submitting your deposit.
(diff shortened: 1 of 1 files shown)


docs/the_nimbus_book/src/start-syncing.md
```diff
@@ -1,9 +1,9 @@
 
 # Start syncing
 
-If you're joining a network that has already launched, you need to ensure that your beacon node is [completely synced](./keep-an-eye.md#keep-track-of-your-syncing-progress) before submitting your deposit.
+To minimize the amount of downtime, you should ensure that your beacon node is [completely synced](./keep-an-eye.md#keep-track-of-your-syncing-progress) before submitting your deposit.  If it's not fully synced you will miss attestations and proposals until it becomes synced.
 
-This is particularly important if you are joining a network that's been running for a while.
+This is particularly important if you are joining a network that's been running for a while since the sync could take some time.
```
mratsim, line 6: Actually now that checkpoint sync is in the `unstable` branch, we should change the text to explain how to use it to start validating in a couple minutes.

mratsim (approved):
LGTM

We have another section on checkpoint sync to add afterwards

PR #3301, opened 2022-01-19
Fix 408 Timeout error returned by REST server.

cheatfate:
This PR turns off timeouts for REST server and also adds some configuration options available through CLI.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -339,6 +339,23 @@ type
         desc: "The number of seconds to keep recently accessed states in memory"
         name: "rest-statecache-ttl" }: Natural
 
+      restRequestTimeout* {.
+        defaultValue: 0
+        desc: "The number of seconds to wait until complete REST request will be received"
+        name: "rest-request-timeout".}: Natural
+
+      restMaxRequestBodySize* {.
+        defaultValue: 16_384
+        desc: "Maximum size of REST's request body in kilobytes. Increase " &
+              "this value, if you going to want to manage thousands of " &
+              "validators using REST"
+        name: "rest-max-body-size" .}: Natural
+
+      restMaxRequestHeadersSize* {.
+        defaultValue: 64
+        desc: "Maximum size of REST's request headers in kilobytes"
+        name: "rest-max-headers-size" .}: Natural
+
       keymanagerEnabled* {.
         desc: "Enable the REST keymanager API (BETA version)"
         defaultValue: false
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -341,19 +341,18 @@ type
 
       restRequestTimeout* {.
         defaultValue: 0
-        desc: "The number of seconds to wait until complete REST request will be received"
+        desc: "The number of seconds to wait until complete REST request " &
+              "will be received"
         name: "rest-request-timeout".}: Natural
 
       restMaxRequestBodySize* {.
         defaultValue: 16_384
-        desc: "Maximum size of REST's request body in kilobytes. Increase " &
-              "this value, if you going to want to manage thousands of " &
-              "validators using REST"
+        desc: "Maximum size of REST request body (kilobytes)"
         name: "rest-max-body-size" .}: Natural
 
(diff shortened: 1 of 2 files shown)


beacon_chain/conf.nim
```diff
@@ -341,6 +341,23 @@ type
...
         name: "rest-statecache-ttl" }: Natural
 
+      restRequestTimeout* {.
+        defaultValue: 0
+        desc: "The number of seconds to wait until complete REST request will be received"
+        name: "rest-request-timeout".}: Natural
+
+      restMaxRequestBodySize* {.
+        defaultValue: 16_384
+        desc: "Maximum size of REST's request body in kilobytes. Increase " &
```
tersec:
```suggestion
        desc: "Maximum size of REST request body in kilobytes. Increase " &

```

beacon_chain/conf.nim
```diff
@@ -341,6 +341,23 @@ type
...
+              "this value, if you going to want to manage thousands of " &
```
tersec:
```suggestion
              "this value if you're going to manage thousands of " &
```

beacon_chain/conf.nim
```diff
@@ -341,6 +341,23 @@ type
...
+              "validators using REST"
+        name: "rest-max-body-size" .}: Natural
+
+      restMaxRequestHeadersSize* {.
+        defaultValue: 64
+        desc: "Maximum size of REST's request headers in kilobytes"
```
tersec:
```suggestion
        desc: "Maximum size of REST request headers in kilobytes"
```

beacon_chain/conf.nim
```diff
@@ -341,6 +341,23 @@ type
         desc: "The number of seconds to keep recently accessed states in memory"
         name: "rest-statecache-ttl" }: Natural
 
+      restRequestTimeout* {.
+        defaultValue: 0
+        desc: "The number of seconds to wait until complete REST request will be received"
+        name: "rest-request-timeout".}: Natural
+
+      restMaxRequestBodySize* {.
+        defaultValue: 16_384
```
arnetheduck, line 350: how many validators is this? if validators are json-encoded, this looks like it's on the low end

arnetheduck: oh. kilobytes - then it's probably not very low after all :) still.. how many approx?

arnetheduck: and since it's not low, maybe we should leave it `hidden` - it's a nice-to-have escape hatch, but I'm not sure we want to clutter the `--help` with it

tersec: I also view hidden options as somewhat akin to undocumented options.

Since this whole thing seems like a kludgy, albeit unfortunately necessary for the moment, workaround, and I'm not especially convinced that Nimbus was doing it wrong or other clients are doing it right, it'd be nice to be able to more readily remove these options should the overall context change.

cheatfate: I've decided to remove this documentation sentence. Because i have made it confusing: you can manage whatever validators you have with any reasonable request body size. The only difference is how many requests you will be forced to perform to get information for whole list of validators.

beacon_chain/conf.nim
```diff
@@ -341,6 +341,22 @@ type
         desc: "The number of seconds to keep recently accessed states in memory"
         name: "rest-statecache-ttl" }: Natural
 
+      restRequestTimeout* {.
+        defaultValue: 0
+        desc: "The number of seconds to wait until complete REST request " &
+              "will be received"
+        name: "rest-request-timeout".}: Natural
```
zah: This would look better in the help screen if you add the `defaultValueDesc: "infinite"` pragma. Also, in the rest of the file we don't close the pragma sections with `.}`, but only with `}`.

beacon_chain/conf.nim
arnetheduck: > undocumented options.

Yes, I believe this is more or less where an option like this belongs: we shouldn't have to ask the user to decide on this number, it's there as insurance against an unexpected use case

PR #3302, opened 2022-01-20
rename MERGE_FORK_VERSION to BELLATRIX_FORK_VERSION

tersec:
https://github.com/ethereum/consensus-specs/releases/tag/v1.1.8 specifies this.

This results in many harmless
```
@["Unknown constants in file: ", "MERGE_FORK_VERSION"]
```
messages while building, because https://github.com/eth-clients/eth2-networks/blob/e14cfc658fcde662d08ba5cf50f263fca7421eba/shared/pyrmont/config.yaml, https://github.com/eth-clients/eth2-networks/blob/e14cfc658fcde662d08ba5cf50f263fca7421eba/shared/prater/config.yaml, and https://github.com/eth-clients/eth2-networks/blob/e14cfc658fcde662d08ba5cf50f263fca7421eba/shared/mainnet/config.yaml all in the pinned `eth2-networks` commit provide a `MERGE_FORK_VERSION` (v1.1.7 and earlier) but not a `BELLATRIX_FORK_VERSION` (v1.1.8 and later). It will substitute some, generally incorrect, default BELLATRIX_FORK_VERSION in its place.

The Prater one could be resolved (if one were willing to trade off by having to update/rename `MERGE_FORK_EPOCH` references too) because https://github.com/eth-clients/eth2-networks/commit/cacadc372a65cb09cf45bc76d6cd08b2dc5196bf fixes it for Prater.

I've created a PR for doing the mainnet one: https://github.com/eth-clients/eth2-networks/pull/72

That would just leave Pyrmont creating that build warning.

I'd rather not remove/silence that build warning, since it's legitimate/genuine, just, in this case, harmless, because the exact value of the `MERGE_FORK_VERSION`/`BELLATRIX_FORK_VERSION` only matters if they'll ever run in the Bellatrix fork. All of Pyrmont, Prater, and Mainnet config YAMLs specify that the fork epoch (which is still read correctly from each) is `high(uint64)`, i.e. will not happen.

So, this PR is safe to merge, if one's okay with the additional build warnings.


tersec:
https://github.com/eth-clients/eth2-networks/pull/73, if/when merged with the mainnet change, will allow a clean update for the rest.

PR #3307, opened 2022-01-21
Put `application/json` as a higher preference than `application/octet…

EmilIvanichkovv:
…-stream`

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -806,18 +806,17 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
 
     let contentType =
       block:
-        let res = preferredContentType("application/octet-stream",
-                                       "application/json")
+        let res = preferredContentType(jsonMediaType,
+                                       sszMediaType)
         if res.isErr():
           return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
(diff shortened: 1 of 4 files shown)


zah (approved):

tersec:
Why does this appear to be pulling in different commits of `vendor/nimbus-build-system`, `vendor/nim-presto`, `vendor/nim-http-utils`, and `vendor/nim-chronos`? It's possible it needs maybe one of those, but presumably not all four? Or, it's some basically mirage, in which case that should still be cleaned up before merging.

tersec:
2022-01-30-151903_946x783_scrot
Now it just needs different versions of `nim-chronos`, `nim-http-utils`, and `nim-presto` -- is this intentional/actually required?

EmilIvanichkovv:
Well, `nimbus-build-system` was bumped by mistake previously. The other three are required because they contain important fixes needed for the code in `nimbus-eth2`

arnetheduck:
Did you test that trusted node sync doesn't break and/or regress to JSON? It calls the state download function and relies on SSZ for acceptable performance, both for blocks and states.

zah:
The trusted sync should be supplying the proper Accept header, but good point. A bit of work that remains here is to add some tests that will detect this default preference through the `restapi.sh` script. This will be delivered in a follow-up PR (@EmilIvanichkovv will focus on his university studies in the coming week).

PR #3308, opened 2022-01-21
dynamic sync committee subscriptions

tersec:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -586,6 +586,13 @@ func verifyFinalization(node: BeaconNode, slot: Slot) =
 func subnetLog(v: BitArray): string =
   $toSeq(v.oneIndices())
 
+template forkDigests(node: BeaconNode): auto =
```
zah: This looks like it would be better off as a regular function. The generated code is probably large enough, so not inlining will result in a smaller binary.

beacon_chain/spec/network.nim
```diff
@@ -166,3 +166,26 @@ func getTargetGossipState*(
...
+func nearSyncCommitteePeriod*(epoch: Epoch): Option[uint64] =
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#sync-committee-subnet-stability
+  for epochsBefore in 0'u64 .. SYNC_COMMITTEE_SUBNET_COUNT:
+    if (epoch + epochsBefore).is_sync_committee_period():
+      return some epochsBefore
+
+  none(uint64)
+
+func getSyncSubnets*(
+    nodeHasPubkey: proc(pubkey: ValidatorPubKey): bool {.noSideEffect.},
```
zah: You need `raises` annotation on this function to avoid the default `raises: [Exception]` interpretation.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -705,38 +708,140 @@ proc removePhase0MessageHandlers(node: BeaconNode, forkDigest: ForkDigest) =
...
+  debug "trackCurrentSyncCommitteeTopics: aligning with sync committee subnets",
+    currentSyncCommitteeSubnets,
+    metadata_syncnets = node.network.metadata.syncnets,
+    gossipState = node.gossipState
+
+  for subcommitteeIdx in SyncSubcommitteeIndex:
+    doAssert not (newSyncSubnets[subcommitteeIdx] and
+                  oldSyncSubnets[subcommitteeIdx])
+    for gossipFork in node.gossipState:
+      let topic =
```
zah: If you make this a template it won't be computed unnecessarily for the topics where there is no change. You can also use `elif` below for the second condition.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -705,38 +708,140 @@ proc removePhase0MessageHandlers(node: BeaconNode, forkDigest: ForkDigest) =
...
+        getSyncCommitteeTopic(forkDigests[gossipFork], subcommitteeIdx)
+      if oldSyncSubnets[subcommitteeIdx]:
+        node.network.unsubscribe(topic)
+      if newSyncSubnets[subcommitteeIdx]:
+        node.network.subscribe(topic, basicParams)
+
+  node.network.updateSyncnetsMetadata(currentSyncCommitteeSubnets)
+
+proc trackNextSyncCommitteeTopics(node: BeaconNode, slot: Slot) {.async.} =
+  if not slot.is_epoch:
```
zah: Calling async functions is relatively expensive. This check may be better at the call-site.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -705,38 +708,140 @@ proc removePhase0MessageHandlers(node: BeaconNode, forkDigest: ForkDigest) =
...
+    return
+
+  let
+    epoch = slot.epoch
+    epochToSyncPeriod = nearSyncCommitteePeriod(epoch)
+
+  if  epochToSyncPeriod.isNone or
+      forkVersionAtEpoch(node.dag.cfg, epoch + 1) ==
```
zah: Shouldn't the check here use `epoch + epochToSyncPeriod.get`? Otherwise, if we are in the epochs just before the Altair branch and `epochToSyncPeriod` was set to 3 for example, we would still not not subscribe (because `epoch + 1` is still not an Altair epoch).

beacon_chain/consensus_object_pools/sync_committee_msg_pool.nim
```diff
@@ -271,3 +277,14 @@ proc produceSyncAggregate*(
...
+proc isEpochLeadTime*(
+    pool: SyncCommitteeMsgPool, epochsToSyncPeriod: uint64): bool =
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#sync-committee-subnet-stability
+  # This ensures a uniform distribution without requiring additional state:
+  # (1/4)                         = 1/4, 4 slots out
+  # (3/4) * (1/3)                 = 1/4, 3 slots out
+  # (3/4) * (2/3) * (1/2)         = 1/4, 2 slots out
+  # (3/4) * (2/3) * (1/2) * (1/1) = 1/4, 1 slot out
+  doAssert epochsToSyncPeriod > 0
+  pool.rng[].rand(epochsToSyncPeriod - 1) == 0
```
zah:
```suggestion
  epochsToSyncPeriod == 1 or pool.rng[].rand(epochsToSyncPeriod - 1) == 0
```

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -705,38 +708,140 @@ proc removePhase0MessageHandlers(node: BeaconNode, forkDigest: ForkDigest) =
...
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#sync-committee-subnet-stability
+  for subcommitteeIdx in SyncSubcommitteeIndex:
+    if (not node.network.metadata.syncnets[subcommitteeIdx]) and
+        nextSyncCommitteeSubnets[subcommitteeIdx] and
+        node.syncCommitteeMsgPool[].isEpochLeadTime(epochToSyncPeriod.get):
+      for gossipFork in node.gossipState:
+        node.network.subscribe(getSyncCommitteeTopic(
+          forkDigests[gossipFork], subcommitteeIdx), basicParams)
+
+  node.network.updateSyncnetsMetadata(nextSyncCommitteeSubnets)
```
zah: We didn't necessarily connect to all of the `nextSyncCommitteeSubnets` here though, so this update seems incorrect (due to the randomness in `isEpochLeadTime`).

beacon_chain/spec/network.nim
```diff
@@ -166,3 +166,26 @@ func getTargetGossipState*(
     {BeaconStateFork.Phase0, BeaconStateFork.Altair}
   else:
     raiseAssert "Unknown target gossip state"
+
+func nearSyncCommitteePeriod*(epoch: Epoch): Option[uint64] =
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/validator.md#sync-committee-subnet-stability
+  for epochsBefore in 0'u64 .. SYNC_COMMITTEE_SUBNET_COUNT:
+    if (epoch + epochsBefore).is_sync_committee_period():
```
zah: Less looping and divisions:

```nim
let epochBefore = EPOCHS_PER_SYNC_COMMITTEE_PERIOD - epoch.since_sync_committee_period_start()
if epochsBefore <= SYNC_COMMITTEE_SUBNET_COUNT:
  return some epochsBefore
```

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/019f243e7eddb03026b2d3b68f72ba638836e974

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/019f243e7eddb03026b2d3b68f72ba638836e974

beacon_chain/spec/network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/8de94576669d198f7f844e3fdd5649d9e7b5313a

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/8de94576669d198f7f844e3fdd5649d9e7b5313a

beacon_chain/nimbus_beacon_node.nim
tersec: Indeed, https://github.com/status-im/nimbus-eth2/pull/3308/commits/8de94576669d198f7f844e3fdd5649d9e7b5313a

arnetheduck:
This code needs to be restructured and moved to `ActionTracker` where it should work similar to attestation duties - this is necessary to support duties coming from the REST API which otherwise will be broken by this approach - so far, the rest api discards sync duties knowing that all subnets are subscribed regardless: https://github.com/status-im/nimbus-eth2/blob/c7e92bfd842e96f1e9ed1fa997042f9d972bb910/beacon_chain/rpc/rest_validator_api.nim#L604

this is a blocker, or we break 3rd-party vc:s

tersec:
> This code needs to be restructured and moved to `ActionTracker` where it should work similar to attestation duties - this is necessary to support duties coming from the REST API which otherwise will be broken by this approach - so far, the rest api discards sync duties knowing that all subnets are subscribed regardless:
> 
> https://github.com/status-im/nimbus-eth2/blob/c7e92bfd842e96f1e9ed1fa997042f9d972bb910/beacon_chain/rpc/rest_validator_api.nim#L604
> 
> this is a blocker, or we break 3rd-party vc:s

It handles VCs fine:
https://github.com/status-im/nimbus-eth2/blob/8de94576669d198f7f844e3fdd5649d9e7b5313a/beacon_chain/rpc/rest_validator_api.nim#L597-L598

https://github.com/status-im/nimbus-eth2/blob/8de94576669d198f7f844e3fdd5649d9e7b5313a/beacon_chain/nimbus_beacon_node.nim#L711-L715

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/019f243e7eddb03026b2d3b68f72ba638836e974

beacon_chain/spec/network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/0c24bdb1db8a4647ac36f06dc9ff9b946889f57f

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/0c24bdb1db8a4647ac36f06dc9ff9b946889f57f

arnetheduck:
hm, is `subscribeAllSubnets` handled here? I imagine that we should also add `subscribeAllSyncnets`, `subscribeAllAttnets` and change the `subscribeAllSubnets` to enable both

tersec:
> hm, is `subscribeAllSubnets` handled here? I imagine that we should also add `subscribeAllSyncnets`, `subscribeAllAttnets` and change the `subscribeAllSubnets` to enable both

https://github.com/status-im/nimbus-eth2/pull/3308/commits/9983903a68f3ff2bd293a32c902ef106c4bd67dc

This sets enabling `--subscribe-all-subnets` to override the other two, which allows for an overall more reasonable tradeoff.

Having subscribe-or-not all `${FOO}` subnets be different than a third, use-default state creates an actually-ternary setting which would have unintuitive behavior.

As long as one treats not specifying the options at all the same as explicitly specifying their default options, then the only other vaguely reasonable approach here (sync/attnets options override all-subnet options) means that one'd have to specify both sync/attnet and subnet to get anything. These options/switches exist in software, and could be documented, but present complexity up front without increasing expressiveness.

Furthermore, this version allows for backwards compatibility -- anyone already specifying `--subscribe-all-subnets` gets the status quo behavior of that option.

So, this seems like the least worst version. Alternatively, one could just not allow separately setting sync/attnets, but there are use cases where they're differently appropriate; e.g., a boot node should specify both, while someone mostly looking to boost attestation-based profit slightly might only be interested in the attnets one.

Some software uses a syntax closer to `--select-suboptions:a,b` where something like `--select-subnets:all`, `--select-subnets:optionA`, and `--select-subnets:optionB` or similar is used. This shows the relationship more implicitly, but also creates an entirely separate parsing sublanguage (which Nimbus already has in some cases) for probably not much gain.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -705,38 +709,146 @@ proc removePhase0MessageHandlers(node: BeaconNode, forkDigest: ForkDigest) =
...
-      # TODO This should be done in dynamic way in trackSyncCommitteeTopics
       node.network.unsubscribe(getSyncCommitteeTopic(forkDigest, idx))
 
   node.network.unsubscribe(
     getSyncCommitteeContributionAndProofTopic(forkDigest))
 
-proc trackSyncCommitteeTopics*(node: BeaconNode) =
-  # TODO
-  discard
+proc trackCurrentSyncCommitteeTopics(node: BeaconNode, slot: Slot) {.async.} =
```
arnetheduck: why async?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -705,38 +709,146 @@ proc removePhase0MessageHandlers(node: BeaconNode, forkDigest: ForkDigest) =
...
+      template topic(): auto =
+        getSyncCommitteeTopic(forkDigests[gossipFork], subcommitteeIdx)
+      if oldSyncSubnets[subcommitteeIdx]:
+        node.network.unsubscribe(topic)
+      elif newSyncSubnets[subcommitteeIdx]:
+        node.network.subscribe(topic, basicParams)
+
+  node.network.updateSyncnetsMetadata(currentSyncCommitteeSubnets)
+
+proc trackNextSyncCommitteeTopics(node: BeaconNode, slot: Slot) {.async.} =
```
arnetheduck: likewise

arnetheduck (approved):

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/b0a2d23e517ee3ef12af5e414b70836fcc7b827f

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3308/commits/b0a2d23e517ee3ef12af5e414b70836fcc7b827f

PR #3309, opened 2022-01-21
wss_sim: state transition simulator

arnetheduck:
It's sometimes useful to simulate what happens when a chain runs from a
given state with a given set of private keys - `wss_sim` allows running
such a simulation.

One use of such a tool is to simulate a weak subjectivity attack,
creating alternative histories of the same chain:
https://notes.status.im/nimbus-insecura-network#

### DIFF
--- a/research/wss_sim.nim
+++ b/research/wss_sim.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2019-2021 Status Research & Development GmbH
+# Copyright (c) 2022 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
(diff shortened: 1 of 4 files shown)


research/wss_sim.nim
```diff
@@ -0,0 +1,259 @@
+# beacon_chain
+# Copyright (c) 2019-2021 Status Research & Development GmbH
```
tersec: 2021?

arnetheduck: eh right, that's when I had the sim ready :)

tersec (approved):

mratsim (approved):
Excellent!

research/wss_sim.nim
arnetheduck:
```suggestion
# Copyright (c) 2022 Status Research & Development GmbH
```

PR #3311, opened 2022-01-23
Bugfix: Take into account the finalization delay in the ncli_db rewards calculation

zah:
This fixes a reward calculation error affecting Prater's epoch 31256

### DIFF
--- a/ncli/ncli_common.nim
+++ b/ncli/ncli_common.nim
@@ -1,5 +1,6 @@
 import
-  re, strutils, os,
+  re, strutils, os, math,
+  stew/bitops2,
   ../beacon_chain/spec/[
     datatypes/base,
     datatypes/phase0,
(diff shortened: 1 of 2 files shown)


arnetheduck:
this looks like https://github.com/status-im/nimbus-eth2/blob/c7e92bfd842e96f1e9ed1fa997042f9d972bb910/beacon_chain/spec/datatypes/base.nim#L449, no?

zah:
No, this is about whether the whole network is failing to finalize. It's related to the "inactivity leak" mechanism that affects the rewards.

arnetheduck:
> No, this is about whether the whole network is failing to finalize. 

so.. why is the finalized epoch in the state wrong in this case - or rather.. why does it need recomputing?

arnetheduck:
the epoch transition updates the checkpoints before computing the rewards based on the    justification of the previous epoch - if you run this same function on the already-epoch-transitioned state, it will give different results than those that are used in the state transition, no?

zah:
This is a problem with the order of computations (you need a piece of information that is generated by the state transition before you actually run the state transition). You can avoid it by copying the state, but that would slow down the overall processing speed significantly.

arnetheduck:
Yes, but which piece? I'm trying to understand the flow.

zah:
`process_justification_and_finalization` is part of the state transition and it mutates the state by updating the `finalized_checkpoint` field. This field is later read in the formulas affecting the rewards (more specifically in the computation of the `finality_delay` which is replicated here).

Because the current architecture of `ncli_db` is such that we need to be able to collect the reward information before executing the state transition function (the state transition destroys some of the information), we need to compute all the inputs upfront by duplicating the logic in the spec.

PR #3312, opened 2022-01-24
Harden handling of unviable forks

arnetheduck:
In our current handling of unviable forks, we allow peers to send us
blocks that come from a different fork - this is not necessarily an
error as it can happen naturally, but it does open up the client to a
case where the same unviable fork keeps getting requested - rather than
allowing this to happen, we'll now give these peers a small negative
score - if it keeps happening, we'll disconnect them.

* keep track of unviable forks in quarantine, to avoid filling it with
known junk
* collect peer scores in single module
* descore peers when they send unviable blocks during sync
* don't give score for duplicate blocks
* increase quarantine size to a level that allows finality to happen
under optimal conditions - this helps avoid downloading the same blocks
over and over in case of an unviable fork
* increase initial score for new peers to make room for one more failure
before disconnection
* log and score invalid/unviable blocks in requestmanager too
* avoid ChainDAG dependency in quarantine
* reject gossip blocks with unviable parent
* continue processing unviable sync blocks in order to build unviable
dag

### DIFF
--- a/beacon_chain/consensus_object_pools/block_pools_types.nim
+++ b/beacon_chain/consensus_object_pools/block_pools_types.nim
@@ -38,8 +38,8 @@ type
       ## appears or be discarded if finality obsoletes it
 
     UnviableFork
-      ## Block is from a different history / fork than the one we're interested
-      ## in (based on our finalized checkpoint)
+      ## Block is from a history / fork that does not include our most current
+      ## finalied checkpoint
 
     Duplicate
       ## We've seen this block already, can't add again
--- a/beacon_chain/consensus_object_pools/block_pools_types.nim
+++ b/beacon_chain/consensus_object_pools/block_pools_types.nim
@@ -39,7 +39,7 @@ type
 
     UnviableFork
       ## Block is from a history / fork that does not include our most current
-      ## finalied checkpoint
+      ## finalized checkpoint
 
     Duplicate
       ## We've seen this block already, can't add again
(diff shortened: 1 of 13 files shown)


beacon_chain/consensus_object_pools/block_quarantine.nim
```diff
@@ -32,18 +38,21 @@ type
...
+      ## with a valid signature to be dropped
 
-    missing*: Table[Eth2Digest, MissingBlock] ##\
-    ## Roots of blocks that we would like to have (either parent_root of
-    ## unresolved blocks or block roots of attestations)
+    unviable*: OrderedTable[Eth2Digest, tuple[]]
+      ## Unviable blocks are those from a different fork - we keep their hash
+      ## around so that we can avoid cluttering the orphans table with their
+      ## descendants - an ordered table is used to keep them in FIFO order more
+      ## or less..
```
mratsim: This needs a description of what is an unviable block:
- different last_finalized_checkpoint?

And explanation of when and how it is pruned, apparently in FIFO.

arnetheduck: ec08554b05c501e86a386d10387d70ed3d6888d0

beacon_chain/consensus_object_pools/block_pools_types.nim
```diff
@@ -38,8 +38,8 @@ type
       ## appears or be discarded if finality obsoletes it
 
     UnviableFork
-      ## Block is from a different history / fork than the one we're interested
-      ## in (based on our finalized checkpoint)
+      ## Block is from a history / fork that does not include our most current
+      ## finalied checkpoint
```
mratsim: typo finalized

mratsim (approved):

beacon_chain/consensus_object_pools/block_pools_types.nim
arnetheduck:
```suggestion
      ## finalized checkpoint
```

PR #3313, opened 2022-01-24
ncli_db: import states and blocks from era file

arnetheduck:


beacon_chain/beacon_chain_db.nim
```diff
@@ -540,6 +567,10 @@ proc putState*(db: BeaconChainDB, key: Eth2Digest, value: bellatrix.BeaconState)
   db.mergeStatesNoVal.putSnappySSZ(
     key.data, toBeaconStateNoImmutableValidators(value))
 
+proc putState*(db: BeaconChainDB, state: ForkyHashedBeaconState) =
+  db.withManyWrites:
+    db.putStateRoot(state.latest_block_root(), state.data.slot, state.root)
+    db.putState(state.root, state.data)
 # For testing rollback
```
tersec, line 574: No empty line between `proc`s?

arnetheduck: will fix in next PR (have it locally already)

beacon_chain/beacon_chain_db.nim
```diff
@@ -266,6 +272,24 @@ proc loadImmutableValidators(vals: DbSeq[ImmutableValidatorDataDb2]): seq[Immuta
       pubkey: tmp.pubkey.loadValid(),
       withdrawal_credentials: tmp.withdrawal_credentials)
 
+template withManyWrites*(db: BeaconChainDB, body: untyped) =
+  # We don't enforce strong ordering or atomicity requirements in the beacon
+  # chain db in general, relying instead on readers to be able to deal with
+  # minor inconsistencies - however, putting writes in a transaction is orders
+  # of magnitude faster when doing many small writes, so we use this as an
+  # optimization technique and the templace is named accordingly.
+  mixin expectDb
```
zah, line 281: `expectDb` is defined in this module, so the mixin treatment is wrong.

I assume you had a compilation error. This happened because Nim has some issues with the dot operator handling in the compiler pre-pass applied to generic functions and templates. You can work-around the issue by not using the dot syntax:

```nim
expectDb db.db.exec("BEGIN TRANSACTION;")
```

tersec (approved):

zah (approved):

beacon_chain/beacon_chain_db.nim
arnetheduck: but that's ugly :/

arnetheduck: fixed locally, coming soon

PR #3314, opened 2022-01-24
add flashbots (milestone 1) consensus beacon block types

tersec:
These enforce that the CL validator can't influence the transaction selection.

### DIFF
--- a/beacon_chain/spec/datatypes/bellatrix.nim
+++ b/beacon_chain/spec/datatypes/bellatrix.nim
@@ -326,36 +326,6 @@ type
 
   SomeSomeSignedBeaconBlock* = SomeSignedBeaconBlock | altair.SomeSignedBeaconBlock | phase0.SomeSignedBeaconBlock
 
-  # https://github.com/flashbots/mev-boost/blob/thegostep/docs/docs/milestone-1.md#blindedbeaconblockbody
-  # This is forked from bellatrix.BeaconBlockBody with execution_payload
-  # replaced with execution_payload_header
-  BlindedBeaconBlockBody = object
-    randao_reveal*: ValidatorSig
-    eth1_data*: Eth1Data
-    graffiti*: GraffitiBytes
-    proposer_slashings*: List[ProposerSlashing, Limit MAX_PROPOSER_SLASHINGS]
-    attester_slashings*: List[AttesterSlashing, Limit MAX_ATTESTER_SLASHINGS]
-    attestations*: List[Attestation, Limit MAX_ATTESTATIONS]
-    deposits*: List[Deposit, Limit MAX_DEPOSITS]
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -326,14 +326,44 @@ type
 
   SomeSomeSignedBeaconBlock* = SomeSignedBeaconBlock | altair.SomeSignedBeaconBlock | phase0.SomeSignedBeaconBlock
 
-  BlockParams* = object
+  # https://github.com/flashbots/mev-boost/blob/thegostep/docs/docs/milestone-1.md#blindedbeaconblockbody
+  # This is forked from bellatrix.BeaconBlockBody with execution_payload
+  # replaced with execution_payload_header
+  BlindedBeaconBlockBody = object
```
arnetheduck: should we put these in a separate, non-spec module? these will never be seen by the state transition, right? (or at least a separate module)

tersec: https://github.com/status-im/nimbus-eth2/pull/3314/commits/7b26105338346aa5ccb7311d884dd451f803d230

Sure, it's not part of the state transition.

I agree that it shouldn't get mixed in with the EF consensus specs, but there's reason not to separate it wholly either:

- it is a spec (still tentatively, but I'd anticipate it becoming a more solidly salient one in the coming months, however it transforms). I think of the `spec/` directory partly as a part of the repository which more-or-less directly spells in Nim data types and algorithms externally specified, more than contains much original design of its own. That might not be the EF CL specs.

- However, whether for the Bellatrix or upcoming sharding forks, it's genuinely attached/a fork, as it states, of the EF CL specs in many ways, as well as of the engine specs. In particular as regards the former, this means that each new EF CL fork will require, with the current paradigm, a matching/parallel MEV spec which does more or less the same thing of extracting out the beacon block body (likely to change each fork), replacing a field or two, and propagating that change back up the dependency tree. Practically, it'll need parallel maintenance; it's too enmeshed.

So this seems like one approach to achieving that, in terms of module structure. One option is to keep `spec/datatypes/` as meaning EF CL spec datatypes; as long as the MEV modules are kept separate enough, that's easy to refactor later too.

arnetheduck: fair, as long as it's greppable and clearly marked, it's fine - also because it's not covered by spec tests etc

arnetheduck (approved):

beacon_chain/spec/datatypes/bellatrix.nim
mratsim: `spec/mev` or `spec/flashbots` folder?

tersec: https://github.com/status-im/nimbus-eth2/pull/3319

PR #3316, opened 2022-01-24
book: migration guide update

unixpi:
add missing migration instructions for teku and lighthouse; add context; rework structure


docs/the_nimbus_book/src/migration.md
```diff
@@ -1,23 +1,27 @@
 # Migrate from another client
 
 
-*See [here](./migration-options.md) for advanced options*
+This guide will take you through the basics of how to migrate to Nimbus from another client. *See [here](./migration-options.md) for advanced options*.
 
-## Step 1 - Sync the beacon node
+The main pain point involves the exporting and importing of the [slashing protection database](https://eips.ethereum.org/EIPS/eip-3076), since each client takes a slightly different approach here.
```
arnetheduck, line 6: "The most important point about migrations is that you ensure that at no time are the two clients both validating at the same time" or something like that

docs/the_nimbus_book/src/migration.md
```diff
@@ -1,23 +1,27 @@
...
+
+## Step 1 - Sync the Nimbus beacon node
 No matter which client you are migrating over from, the first step is to sync the Nimbus beacon node.
 
-The easiest way to do this is to follow the [beacon node quick start guide](./quick-start.md).  Syncing the beacon node might take up to 30h depending on your hardware - you should keep validating using your current setup until it completes.
+The easiest way to do this is to follow the [beacon node quick start guide](./quick-start.md).  Syncing the beacon node might take up to 30 hours depending on your hardware - you should keep validating using your current setup until it completes.
 
+Once your Nimbus beacon node has synced and you're satisfied that it's working, move to **Step 2**.
 
 ## Step 2 - Export your slashing protection history
```
arnetheduck: Let's make a "stop your existing client" step maybe - if it's still validating, the db export will not work / be  stale

arnetheduck: or retitle this "Stop your existing client and export ..." - the stop part is significantly important

docs/the_nimbus_book/src/migration.md
```diff
@@ -41,13 +43,73 @@ prysm.sh validator slashing-protection export \
...
+sudo systemctl disable lighthousebeacon
+```
+
+*It's important that you disable the service as well as stopping it, to prevent it from starting up again on reboot.*
+
+#### 2. Export slashing protection history
+
+You can export Lighthouse's database with this command:
+```
+lighthouse account validator slashing-protection export lighthouse_interchange.json
```
arnetheduck: let's use a single filename for all exports, so ideally there's a single import command

docs/the_nimbus_book/src/migration.md
```diff
@@ -1,23 +1,27 @@
...
+
+> **Please take your time to get this right.** Don't hesitate to reach out to us in the `#helpdesk` channel of [our discord](https://discord.gg/j3nYBUeEad) if you come across a stumbling block. We are more than happy to help guide you through the migration process. Given what's at stake, there is no such thing as a stupid question.
+
+## Step 1 - Sync the Nimbus beacon node
 No matter which client you are migrating over from, the first step is to sync the Nimbus beacon node.
 
-The easiest way to do this is to follow the [beacon node quick start guide](./quick-start.md).  Syncing the beacon node might take up to 30h depending on your hardware - you should keep validating using your current setup until it completes.
+The easiest way to do this is to follow the [beacon node quick start guide](./quick-start.md).  Syncing the beacon node might take up to 30 hours depending on your hardware - you should keep validating using your current setup until it completes.
 
+Once your Nimbus beacon node has synced and you're satisfied that it's working, move to **Step 2**.
```
liftlines, line 17: To make the guide feel comprehensive, instead of leaving it to the user to determine if the beacon node is syncing well - can we suggest a few checks (perhaps checklist) that they can use to verify and move to the next step more confidently? 

For example - this page links to the Beacon node sync page (https://nimbus.guide/start-syncing.html) - which links to 'keep track of your syncing progress' page (https://nimbus.guide/start-syncing.html) where 'time left to full sync' and 'most recent slot synced' is in the description. We could pull out a few key checks and create a checklist perhaps, which makes the user feel confident when moving to the next stage.

docs/the_nimbus_book/src/migration.md
liftlines: How can the user check and verify that the prysm validator client has stopped validating? Perhaps a few checks here would help.

docs/the_nimbus_book/src/migration.md
```diff
@@ -1,23 +1,27 @@
...
 
 ## Step 2 - Export your slashing protection history
-> Hat tip to Michael Sproul for his [wonderful guide](https://lighthouse.sigmaprime.io/switch-to-lighthouse.html) on how to migrate from Prysm to Lighthouse.
```
liftlines, line 13: While this is helpful - it is perhaps redirecting users away from the steps.

unixpi: the red means it's been removed :)

docs/the_nimbus_book/src/migration.md
unixpi: trouble is there is no short and conclusive checklist because the details do depend on the user's setup. 

middle ground
Screenshot 2022-01-25 at 12 21 17

docs/the_nimbus_book/src/migration.md
unixpi: added a tip

PR #3320, opened 2022-01-25
Store finalized block roots in database (3s startup)

arnetheduck:
When the chain has finalized a checkpoint, the history from that point
onwards becomes linear - this is exploited in `.era` files to allow
constant-time by-slot lookups.

In the database, we can do the same by storing finalized block roots in
a simple sparse table indexed by slot, bringing the two representations
closer to each other in terms of conceptual layout and performance.

Doing so has a number of interesting effects:

* mainnet startup time is improved 3-5x (3s on my laptop)
* the _first_ startup might take slightly longer as the new index is
being built - ~10s on the same laptop
* we no longer rely on the beacon block summaries to load the full dag -
this is a lot faster because we no longer have to look up each block by
parent root
* a collateral benefit is that we no longer need to load the full
summaries table into memory - we get the RSS benefits of #3164 without
the CPU hit.

Other random stuff:

* simplify forky block generics
* fix withManyWrites multiple evaluation
* fix validator key cache not being updated properly in chaindag
read-only mode
* drop pre-altair summaries from `kvstore`
* recreate missing summaries from altair+ blocks as well (in case
database has lost some to an involuntary restart)
* print database startup timings in chaindag load log
* avoid allocating superfluos state at startup
* use a recursive sql query to load the summaries of the unfinalized
blocks

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -306,6 +306,7 @@ proc init*(T: type FinalizedBlocks, db: SqStoreRef, name: string): KvResult[T] =
   var
     low, high: Opt[Slot]
     tmp: Option[int64]
+
   for rowRes in minIdStmt.exec(tmp):
     expectDb rowRes
     if tmp.isSome():
@@ -317,6 +318,7 @@ proc init*(T: type FinalizedBlocks, db: SqStoreRef, name: string): KvResult[T] =
       high.ok(Slot(tmp.get()))
 
   maxIdStmt.dispose()
+  minIdStmt.dispose()
 
   ok(T(insertStmt: insertStmt,
          selectStmt: selectStmt,
@@ -1093,19 +1095,20 @@ iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
   # Backwards compat for reading old databases, or those that for whatever
   # reason lost a summary along the way..
   while true:
-    if db.v0.backend.getSnappySSZ(subkey(BeaconBlockSummary, res.root), res.summary) == GetResult.found:
-      yield res
+    if db.v0.backend.getSnappySSZ(
+        subkey(BeaconBlockSummary, res.root), res.summary) == GetResult.found:
+      discard # Just yield below
     elif (let blck = db.getPhase0Block(res.root); blck.isSome()):
       res.summary = blck.get().message.toBeaconBlockSummary()
-      yield res
     elif (let blck = db.getAltairBlock(res.root); blck.isSome()):
       res.summary = blck.get().message.toBeaconBlockSummary()
-      yield res
     elif (let blck = db.getMergeBlock(res.root); blck.isSome()):
       res.summary = blck.get().message.toBeaconBlockSummary()
-      yield res
     else:
       break
+
+    yield res
+
     # Next time, load them from the right place
     newSummaries.add(res)
 
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -29,6 +29,15 @@ type
     selectStmt: SqliteStmt[int64, openArray[byte]]
     recordCount: int64
 
+  FinalizedBlocks* = object
+    # A sparse version of DbSeq - can have holes but not duplicate entries
+    insertStmt: SqliteStmt[(int64, array[32, byte]), void]
+    selectStmt: SqliteStmt[int64, array[32, byte]]
+    selectAllStmt: SqliteStmt[NoParams, (int64, array[32, byte])]
+
+    low*: Opt[Slot]
+    high*: Opt[Slot]
+
   DepositsSeq = DbSeq[DepositData]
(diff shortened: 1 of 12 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -934,60 +1035,79 @@ proc loadSummaries*(db: BeaconChainDB): Table[Eth2Digest, BeaconBlockSummary] =
...
-      else:
-        break
-      # Next time, load them from the right place
-      newSummaries.add(res)
+    elif (let blck = db.getAltairBlock(res.root); blck.isSome()):
+      res.summary = blck.get().message.toBeaconBlockSummary()
+      yield res
+    elif (let blck = db.getMergeBlock(res.root); blck.isSome()):
+      res.summary = blck.get().message.toBeaconBlockSummary()
+      yield res
```
zah: Looks like you can place a single `yield` after the if/elif/else section here. Reducing the number of yeilds in an inline iterator is beneficial for reducing the overall code size.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -433,19 +454,59 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
     withBlck(genesisBlock): BlockRef.init(genesisBlockRoot, blck.message)
 
   var
-    headRef: BlockRef
+    backfillBlocks = newSeq[Eth2Digest](tailRef.slot.int)
```
zah, line 457: Doesn't this have to be `tailRef.slot.int - backfill.slot` semantically?

arnetheduck: no: the seq is there to answer by-slot requests from the rest and libp2p api and needs to cover the entire genesis->tail range (for which we have no blockref) - `backfill.slot` identifies the "valid" part of this seq. 

in other words:  after backfilling is done, this seq holds an in-memory mapping of `slot->root` which we then use to load blocks from the database with.

beacon_chain/beacon_chain_db.nim
arnetheduck: 5df81cd72425f8674536ee674406bcbb174db9f8

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: btw, this is the seq we talked about that potentially could be removed in future releases and replaced with a database query, in a future release - it's still a bit early for that though

zah: Isn't the name misleading then? My first interpretation of `backfillBlocks` will be "the blocks that we backfilled"

zah: Isn't `finalizedBlocks` the sequence that starts from genesis?

arnetheduck: https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/consensus_object_pools/block_pools_types.nim#L91 explains the structure quite extensively, but in short, your intuition is correct: `backfillBlocks` are those that were or are being backfillled - they are separate from `finalizedBlocks` (which follow `backfillBlocks`) because we can't use them for rewinding (because we don't have a state to rewind from - `tail` is the checkpoint block, and first block we can rewind from

arnetheduck: one, perhaps confusing inconsistency is that `finalizedBlocks` in the database covers all finalized blocks, while the in-memory structure is split in two, at the tail. I don't have good ideas for how to make that more clear though, right now

zah: But why does `backfillBlocks` need to go to genesis? Shouldn't it cover only the weak subjectivity horizon?

arnetheduck: maybe in the future? right now, the de-facto standard is to backfill to genesis because that's the only point from which you reliably can regenerate states (there's no standard way to distribute states, except the debug REST api) - also, it's needed for clients such as .. say .. nimbus, that don't support checkpoint syncing: a hard-fork is needed before genesis support is dropped.

zah: Well, and this gets us to my original question. Isn't the size of `backfillBlocks` semantically `tailRef.slot.int - backfill.slot`? (this works OK now because `backfill.slot` is zero and it will work OK in the future).

arnetheduck: still no - `backfill.slot` is the "first" slot for which we have a block in the database. on checkpoint sync start, it will have the value of tail.slot, then it moves towards genesis. we don't want to reallocate the seq every time a block is backfilled, so we allocate the full seq at once.

PR #3329, opened 2022-01-27
Book: trusted sync (edits and clarifications)

unixpi:
TODO (in a separate PR): add trusted node sync instructions (with client specific endpoints) to migration guide(s)

### DIFF
--- a/docs/the_nimbus_book/src/SUMMARY.md
+++ b/docs/the_nimbus_book/src/SUMMARY.md
@@ -11,8 +11,8 @@
 # How-to (beacon node)
 - [Install dependencies](./install.md)
 - [Build the beacon node](./build.md)
-- [Sync the beacon node](./start-syncing.md)
-  - [Trusted node sync](./trusted-node-sync.md)
+- [Sync from scratch](./start-syncing.md)
+- [Sync from a trusted node](./trusted-node-sync.md)
 - [Add a backup web3 provider](./web3-backup.md)
 
 # How-to (validator)
--- a/docs/the_nimbus_book/src/SUMMARY.md
+++ b/docs/the_nimbus_book/src/SUMMARY.md
@@ -12,7 +12,7 @@
 - [Install dependencies](./install.md)
 - [Build the beacon node](./build.md)
 - [Sync from scratch](./start-syncing.md)
-- [Sync from a trusted node](./trusted-node-sync.md)
+- [Sync from a trusted node (checkpoint sync)](./trusted-node-sync.md)
 - [Add a backup web3 provider](./web3-backup.md)
 
 # How-to (validator)
--- a/docs/the_nimbus_book/src/start-syncing.md
+++ b/docs/the_nimbus_book/src/start-syncing.md
@@ -1,5 +1,4 @@
-
-# Start syncing
+# Sync from scratch
 
 To minimize the amount of downtime, you should ensure that your beacon node is [completely synced](./keep-an-eye.md#keep-track-of-your-syncing-progress) before submitting your deposit.  If it's not fully synced you will miss attestations and proposals until it has finished syncing.
 
--- a/docs/the_nimbus_book/src/trusted-node-sync.md
+++ b/docs/the_nimbus_book/src/trusted-node-sync.md
@@ -4,7 +4,7 @@ When you [start the beacon node](./quick-start.md) for the first time, it will c
 
 Trusted node sync allows you to get started more quickly with Nimbus by fetching a recent checkpoint from a trusted node (we will expect it will save you 1 to 2 days).
 
-To use trusted node sync, you must have access to a node that you trust that exposes the Ethereum [REST API](./rest-api.md) (for example a locally running backup node).
+To use trusted node sync, you must have access to a node that you trust that exposes the Ethereum [Beacon API](./rest-api.md) (for example a locally running backup node).
 
 Should this node, or your connection to it, be compromised, your node will not be able to detect whether or not it is being served false information.
 
@@ -16,6 +16,15 @@ It is possibly to use trusted node sync with a third-party API provider -- see [
 >
 > `--trusted-node-url=http://127.0.0.1:3500`
 
+**Mainnet**
+
+To sync Mainnet, from the `nimbus-eth2` directory run:
+
+```bash
+build/nimbus_beacon_node trustedNodeSync --network:mainnet \
+ --data-dir=build/data/shared_mainnet_0 \
+ --trusted-node-url=http://localhost:5052
+```
 
 **Prater (testnet)**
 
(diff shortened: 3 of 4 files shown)


docs/the_nimbus_book/src/SUMMARY.md
```diff
@@ -11,8 +11,8 @@
 # How-to (beacon node)
 - [Install dependencies](./install.md)
 - [Build the beacon node](./build.md)
-- [Sync the beacon node](./start-syncing.md)
-  - [Trusted node sync](./trusted-node-sync.md)
+- [Sync from scratch](./start-syncing.md)
+- [Sync from a trusted node](./trusted-node-sync.md)
```
mratsim: I would add "also called checkpoint-sync or snapshot-sync".

docs/the_nimbus_book/src/start-syncing.md
```diff
@@ -1,5 +1,4 @@
-
-# Start syncing
+# Sync from scratch
 
 To minimize the amount of downtime, you should ensure that your beacon node is [completely synced](./keep-an-eye.md#keep-track-of-your-syncing-progress) before submitting your deposit.  If it's not fully synced you will miss attestations and proposals until it has finished syncing.
```
mratsim, line 3: Currently (February 2022) it takes between 2~3 days to sync depending on your hardware. It is strongly recommended to use a Solid State Drive (SSD° or NVMe disk.

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -1,19 +1,26 @@
-# Trusted node sync
+# Sync from a trusted node
 
-When you start the beacon node for the first time, it will connect to the beacon chain network and start syncing automatically, a process that can take several days.
+When you [start the beacon node](./quick-start.md) for the first time, it will connect to the beacon chain network and start syncing automatically, a process that can take several days.
 
-Trusted node sync allows you to get started more quickly with Nimbus by fetching a recent checkpoint from a trusted node.
+Trusted node sync allows you to get started more quickly with Nimbus by fetching a recent checkpoint from a trusted node (we will expect it will save you 1 to 2 days).
```
mratsim, line 5: I think "in a minute".

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -1,19 +1,26 @@
...
+Should this node, or your connection to it, be compromised, your node will not be able to detect whether or not it is being served false information.
 
-It is possibly to use trusted node sync with a third-party API provider - follow the steps below to verify that the chain you were given corresponds to the canonical chain at the time.
+It is possibly to use trusted node sync with a third-party API provider -- see [here](trusted-node-sync.md#verify-you-synced-the-correct-chain) for how to verify that the chain you are given corresponds to the canonical chain at the time.
+
+## Perform a trusted node sync
+
+> **Tip:** Make sure to replace `http://localhost:5052` in the commands below with the appropriate endpoint for you. `http://localhost:5052` is the endpoint exposed by Nimbus but this is not consistent across all clients. For example, if your trusted node is a [Prysm node](https://docs.prylabs.network/docs/how-prysm-works/ethereum-public-api#performing-requests-against-a-local-prysm-node), it exposes `127.0.0.1:3500` by default. Which means you would run the commands below with 
+>
+> `--trusted-node-url=http://127.0.0.1:3500`
```
mratsim, line 17: We should probably list all clients endpoints with a link to their documentation in case it changes without us being notified.

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -22,48 +29,62 @@ build/nimbus_beacon_node trustedNodeSync --network:prater \
...
+
+
+
+## Advanced
+
+### Delay block history backfill
 
 By default, both the state and the full block history will be downloaded from the trusted node.
 
 It is possible to get started more quickly by delaying the backfill of the block history using the `--backfill=false` parameter. In this case, the beacon node will first sync to the current head so that it can start performing its duties, then backfill the blocks from the network.
```
mratsim, line 72: This section is needs to be reworked a bit.

Basically after checkpoint sync, you can do your validator duties, but you don't have the whole blockchain history, and you do both concurrently. after --backfill=false

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -22,48 +29,62 @@ build/nimbus_beacon_node trustedNodeSync --network:prater \
...
 
 While it's backfilling blocks from the network, the node will be violating the beacon chain protocol and may be disconnected or lose reputation with other nodes.
```
mratsim: The violation needs to be explained:

"While backfilling blocks, your node will not be able to answer historical requests or sync requests even though it pretends to be up-to-date with the blockchain latest block. This might get you de-scored and eventually kicked by peers."

docs/the_nimbus_book/src/start-syncing.md
stefantalpalaru: > SSD° or NVMe disk

They're both SSDs. What differs is the bus/interface they're connected to: SATA or NVMe. To complicate things, we have form factors like "M.2", but we can keep everything simple by sticking with "SSD" since the rest is irrelevant for our needs.

arnetheduck: > Currently (February 2022) it takes between 2~3 days to sync depending on your hardware. It is strongly recommended to use a Solid State Drive (SSD° or NVMe disk.

Technically, nimbus will run on pretty much any disk - it's eth1 that needs the fast one. Long past are the days that we needed a fast drive - this is relevant if you have a slow and a fast drive for example.

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -1,19 +1,26 @@
...
+# Sync from a trusted node
 
-When you start the beacon node for the first time, it will connect to the beacon chain network and start syncing automatically, a process that can take several days.
+When you [start the beacon node](./quick-start.md) for the first time, it will connect to the beacon chain network and start syncing automatically, a process that can take several days.
 
-Trusted node sync allows you to get started more quickly with Nimbus by fetching a recent checkpoint from a trusted node.
+Trusted node sync allows you to get started more quickly with Nimbus by fetching a recent checkpoint from a trusted node (we will expect it will save you 1 to 2 days).
 
-To use trusted node sync, you must have access to a node that you trust that exposes the REST HTTP API, for example a locally running backup node.
+To use trusted node sync, you must have access to a node that you trust that exposes the Ethereum [REST API](./rest-api.md) (for example a locally running backup node).
```
arnetheduck: the official name is `Beacon API` which perhaps we should be using consistently

docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -22,48 +29,62 @@ build/nimbus_beacon_node trustedNodeSync --network:prater \
 
 **Mainnet**
```
arnetheduck: let's make `mainnet` the "default" / first instruction throughout, except where it explicitly doesn't make sense

docs/the_nimbus_book/src/trusted-node-sync.md
arnetheduck: that's what it says: "In this case, the beacon node will first sync to the current head so that it can start performing its duties, then backfill the blocks from the network"

even with checkpoint sync, one needs to sync to head (the checkpoint is older than head typically)

PR #3332, opened 2022-01-28
Fix counters in validator monitor totals mode

arnetheduck:
The current counters set gauges etc to the value of the _last_ validator
to be processed - as the name of the feature implies, we should be using
sums instead.

* fix missing beacon state metrics on startup, pre-first-head-selection


beacon_chain/validators/validator_monitor.nim
```diff
@@ -435,19 +474,43 @@ proc registerEpochInfo*(
...
+          in_current_sync_committee += 1
+
+        else:
+          if not self.totals:
+            validator_monitor_validator_in_current_sync_committee.set(0, [metricId])
+            debug "Validator isn't part of the current sync committee",
+              epoch = current_epoch,
+              validator = id
+
+        if state.next_sync_committee.pubkeys.data.contains(pubkey):
```
zah, line 495: How often is this code executed? There are ways to avoid the repeated per-validator linear scans here.

arnetheduck: once per epoch, which is why I didn't bother for now - might do a followup but I want the metrics to be correct first, and this PR is messy enough

zah:
Needs rebasing

PR #3340, opened 2022-01-29
Delete Pyrmont docs

tqpcharlie:
I'm not sure what to do with https://github.com/Zachinquarantine/nimbus-eth2/blob/removepyrmontdocs/docs/the_nimbus_book/src/eth2-stats.md


unixpi:
nice! eth2-stats is an old page that is no longer displayed. so no need to modify it.

PR #3341, opened 2022-01-29
serve libp2p protocol for light client sync

etan-status:
This extends the `--serve-light-client-data` launch option to serve
locally collected light client data via libp2p.
Backfill of historic best `LightClientUpdate` is not yet implemented.

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -7,6 +7,10 @@
 
 {.push raises: [Defect].}
 
+# References to `vFuture` refer to the pre-release proposal of the libp2p based
+# light client sync protocol. Conflicting release versions are not in use.
+# https://github.com/ethereum/consensus-specs/pull/2802
+
 import
   options, tables, sets, macros,
   chronicles, chronos, stew/ranges/bitranges, libp2p/switch,
@@ -21,10 +25,16 @@ logScope:
 
 const
   MAX_REQUEST_BLOCKS = 1024
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#configuration
+  MAX_REQUEST_LIGHT_CLIENT_UPDATES = 128
 
   blockByRootLookupCost = allowedOpsPerSecondCost(50)
   blockResponseCost = allowedOpsPerSecondCost(100)
   blockByRangeLookupCost = allowedOpsPerSecondCost(20)
+  lightClientUpdateResponseCost = allowedOpsPerSecondCost(100)
+  lightClientUpdateByRangeLookupCost = allowedOpsPerSecondCost(20)
+  lightClientBootstrapLookupCost = allowedOpsPerSecondCost(5)
+  lightClientBootstrapResponseCost = allowedOpsPerSecondCost(100)
 
 type
   StatusMsg* = object
@@ -438,6 +448,114 @@ p2pProtocol BeaconSync(version = 1,
 
     debug "Block root request done",
       peer, roots = blockRoots.len, count, found
+
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#bestlightclientupdatesbyrange
+  proc bestLightClientUpdatesByRange(
+      peer: Peer,
+      startPeriod: SyncCommitteePeriod,
+      reqCount: uint64,
+      response: MultipleChunksResponse[altair.LightClientUpdate])
+      {.async, libp2pProtocol("best_light_client_updates_by_range", 0,
+                              isLightClientRequest = true).} =
+    trace "Received LC updates by range request", peer, startPeriod, reqCount
+    if reqCount == 0'u64:
+      raise newException(InvalidInputsError, "Empty range requested")
+    let dag = peer.networkState.dag
+    doAssert dag.serveLightClientData
+
+    let
+      headPeriod = dag.head.slot.sync_committee_period
+      # Limit number of updates in response
+      maxSupportedCount =
+        if startPeriod > headPeriod:
+          0'u64
+        else:
+          min(headPeriod + 1 - startPeriod, MAX_REQUEST_LIGHT_CLIENT_UPDATES)
+      count = min(reqCount, maxSupportedCount)
+      onePastPeriod = startPeriod + count
+
+    peer.updateRequestQuota(count.float * lightClientUpdateByRangeLookupCost)
+    peer.awaitNonNegativeRequestQuota()
+
+    var found = 0
+    for period in startPeriod..<onePastPeriod:
+      let update = dag.getBestLightClientUpdateForPeriod(period)
+      if update.isSome:
+        await response.write(update.get)
+        inc found
+
+    peer.updateRequestQuota(found.float * lightClientUpdateResponseCost)
+
+    debug "LC updates by range request done", peer, startPeriod, count, found
+
+    if found == 0 and count > 0:
+      raise newException(ResourceUnavailableError,
+                         "No light client update available")
+
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#getlatestlightclientupdate
+  proc latestLightClientUpdate(
+      peer: Peer,
+      response: SingleChunkResponse[altair.LightClientUpdate])
+      {.async, libp2pProtocol("latest_light_client_update", 0,
+                              isLightClientRequest = true).} =
+    trace "Received latest LC update request", peer
+    let dag = peer.networkState.dag
+    doAssert dag.serveLightClientData
+
+    peer.awaitNonNegativeRequestQuota()
+
+    let update = dag.getLatestLightClientUpdate
+    if update.isSome:
+      await response.send(update.get)
+    else:
+      raise newException(ResourceUnavailableError,
+                         "No light client update available")
+
+    peer.updateRequestQuota(lightClientUpdateResponseCost)
+
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#getoptimisticlightclientupdate
+  proc optimisticLightClientUpdate(
+      peer: Peer,
+      response: SingleChunkResponse[OptimisticLightClientUpdate])
+      {.async, libp2pProtocol("optimistic_light_client_update", 0,
+                              isLightClientRequest = true).} =
+    trace "Received optimistic LC update request", peer
+    let dag = peer.networkState.dag
+    doAssert dag.serveLightClientData
+
+    peer.awaitNonNegativeRequestQuota()
+
+    let optimistic_update = dag.getOptimisticLightClientUpdate
+    if optimistic_update.isSome:
+      await response.send(optimistic_update.get)
+    else:
+      raise newException(ResourceUnavailableError,
+                         "No optimistic light client update available")
+
+    peer.updateRequestQuota(lightClientUpdateResponseCost)
+
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#getlightclientbootstrap
+  proc lightClientBootstrap(
+      peer: Peer,
+      blockRoot: Eth2Digest,
+      response: SingleChunkResponse[altair.LightClientBootstrap])
+      {.async, libp2pProtocol("light_client_bootstrap", 0,
+                              isLightClientRequest = true).} =
+    trace "Received LC bootstrap request", peer, blockRoot
+    let dag = peer.networkState.dag
+    doAssert dag.serveLightClientData
+
+    peer.updateRequestQuota(lightClientBootstrapLookupCost)
+    peer.awaitNonNegativeRequestQuota()
+
+    let bootstrap = dag.getLightClientBootstrap(blockRoot)
+    if bootstrap.isOk:
+      await response.send(bootstrap.get)
+    else:
+      raise newException(ResourceUnavailableError,
+                         "No light client bootstrap available")
+
+    peer.updateRequestQuota(lightClientBootstrapResponseCost)
 
   proc goodbye(peer: Peer,
                reason: uint64)


etan-status:
Do not use `Squash and merge` on this. Instead, `Rebase` or `Merge`.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -13,129 +20,306 @@ func get_active_header(update: LightClientUpdate): BeaconBlockHeader =
   else:
     update.attested_header
 
+# https://github.com/ethereum/consensus-specs/blob/v1.1.8/specs/altair/sync-protocol.md#get_safety_threshold
```
tersec: v1.1.9 is current spec now. Not sure whether that means just changing these URLs, or something more substantive reflecting some of the light client spec changes you've been getting into the EF CL specs.

beacon_chain/conf.nim
```diff
@@ -387,6 +387,16 @@ type
         desc: "A file specifying the authorizition token required for accessing the keymanager API"
         name: "keymanager-token-file" }: Option[InputFile]
 
+      serveLightClientData* {.
```
tersec: So long as marking these as beta is enough not to create a long-term support obligation, the details of these options don't much matter initially, for the moment. An additional way of marking this, and I'm not sure if this is on net beneficial or not, is to mark them as hidden options which `--help` won't display.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,801 @@
...
+func fromBlock(
+    T: type BeaconBlockHeader,
+    blck: ForkedTrustedSignedBeaconBlock): T =
+  ## Reduce a given full block to just its `BeaconBlockHeader`.
+  withBlck(blck):
+    BeaconBlockHeader.fromBlock(blck)
+
+template epochBoundarySlot(slot: Slot): Slot =
+  ## Compute the first slot at which a block with given slot could finalize.
+  (slot + (SLOTS_PER_EPOCH - 1)).epoch.start_slot
```
tersec: This underestimates by one epoch: first the block has to become part of a chain that's justified and then finalized, in sequence, both of which occur at epoch boundaries (`weigh_justification_and_finalization` in https://github.com/ethereum/consensus-specs/blob/v1.1.9/specs/phase0/beacon-chain.md#justification-and-finalization).

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,801 @@
...
+      let blck = dag.getBlockAtSlot(boundarySlot).blck
+      if blck.slot >= lowSlot:
+        dag.cachedLightClientBootstrap[blck.slot] =
+          CachedLightClientBootstrap(
+            current_sync_committee_branch:
+              dag.getLightClientData(blck.bid).current_sync_committee_branch)
+      boundarySlot = blck.slot.epochBoundarySlot
+      if boundarySlot < SLOTS_PER_EPOCH:
+        break
+      boundarySlot -= SLOTS_PER_EPOCH
```
tersec: Will underflow when `blck.slot` is 0. Which, shouldn't happen, because block at slot 0 is special and not a real block the same way others are, but potentially an issue if Nimbus is going to be able to start Altair or later genesis states at some point.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,801 @@
...
+  for (period, sync_committee_root) in dag.pendingBestLightClientUpdates.keys:
+    if period <= finalizedPeriod:
+      keysToDelete.add (period, sync_committee_root)
+  for key in keysToDelete:
+    dag.pendingBestLightClientUpdates.del key
+
+proc initializeBestLightClientUpdateForPeriod*(
+    dag: ChainDAGRef, period: SyncCommitteePeriod) =
+  ## Compute and cache the `LightClientUpdate` with the most sync committee
+  ## signatures (i.e., participation) for a given sync committee period.
```
tersec: `i.e.` isn't followed by a comma (that's `e.g.`)

beacon_chain/spec/light_client_sync.nim
etan-status: Yeah, for `sync-protocol.md` it seems you have not updated those yet in the rest of the code base, so I put `v1.1.8` for consistency. 
These links do not exist yet, as the proposed libp2p spec has not yet been merged, but they will eventually be correct pending the version number.

etan-status: The spec is currently living at https://github.com/ethereum/consensus-specs/pull/2802

tersec: Fair enough, yeah. I tend to get the efficient bulk updates first, and those aren't they. They'll trickle in in the coming days.

etan-status: Ah, I just seen that you just merged the 1.1.9 update. Will rebase and change my links as well.

beacon_chain/conf.nim
etan-status: I'll also mark them as `hidden`. I think as long as there is no persistence of the caches in database, that the startup times are too slow to show it to non-developers.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Ah, yeah, I think my comment was not very clear here. Have updated to:
```
  ## Compute the first possible epoch boundary state slot of a `Checkpoint`
  ## referring to a block at given slot.
```

e.g., I'm on slot 5 into an epoch. If there were no more slots with blocks in this epoch, its post state advanced to slot 32 could potentially become a `finalized_checkpoint` in the future.

**Update**: Changed description once more, to mention `Checkpoint` instead of `finalized_checkpoint` specifically.

etan-status: Also renamed to `nextEpochBoundarySlot`.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Why would it underflow? The `break` one line above should specifically guard against the underflow.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: I think both are with comma, no? `i.e.` = `that is`, and `e.g.` = `for example`.

etan-status:
Rebased to `unstable`, updated spec URLs for `1.1.9`, and re-phrased `epochBoundarySlot` helper documentation.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
tersec: Oh, it does vary too much, indeed, for these to be a single specifically correct approach.

tests/consensus_spec/altair/test_fixture_sync_protocol_light_client_sync.nim
```diff
@@ -0,0 +1,133 @@
...
+  yaml,
+  # Beacon chain internals
+  ../../../beacon_chain/spec/light_client_sync,
+  ../../../beacon_chain/spec/datatypes/altair,
+  # Test utilities
+  ../../testutil,
+  ../fixtures_utils
+
+const TestsDir =
+  SszTestsDir/const_preset/"altair"/"sync_protocol"/"light_client_sync"/"pyspec_tests"
```
tersec: Neither https://github.com/ethereum/consensus-spec-tests/tree/v1.1.9/tests/mainnet/altair nor https://github.com/ethereum/consensus-spec-tests/tree/v1.1.9/tests/minimal/altair seem to have such tests, nor https://github.com/ethereum/consensus-spec-tests/tree/master/tests/mainnet/altair or https://github.com/ethereum/consensus-spec-tests/tree/master/tests/minimal/altair?

tersec: I guess it depends on https://github.com/ethereum/consensus-specs/pull/2802 being merged

etan-status: Yep, the entire patch series depends on that. I put some instructions on how to generate and integrate the test suite in the PR description all the way to the top.

etan-status: The spec itself is still being discussed.

These are the pending PRs @ consensus-specs:
https://github.com/ethereum/consensus-specs/pulls?q=is%3Apr+sort%3Aupdated-desc+author%3Aetan-status+is%3Aopen

etan-status:
Fixed time delay of broadcasting `OptimisticLightClientUpdate`, it was combined with attestations and sync committee messages, but those may fire earlier than 1/3 of slot. `OptimisticLightClientUpdate` is always sent at 1/3.

Extended `sendBeaconBlock` REST/JSON-RPC API to send the `OptimisticLightClientUpdate` as well.

Avoid "No optimistic light client update for proposed block" log when proposing pre-Altair block.

beacon_chain/consensus_object_pools/block_pools_types.nim
```diff
@@ -159,6 +183,50 @@ type
       ## value with other components which don't have access to the
       ## full ChainDAG.
 
+    # -----------------------------------
+    # Cached data to enable light clients to stay in sync with the network
+
+    cachedLightClientData*: Table[BlockId, CachedLightClientData]
```
arnetheduck: perhaps move all this light server stuff to a separate `LightSyncCache` type?

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -53,6 +53,7 @@ const
 
   # https://github.com/ethereum/consensus-specs/blob/v1.1.9/setup.py#L478-L479
   FINALIZED_ROOT_INDEX* = 105.GeneralizedIndex
+  CURRENT_SYNC_COMMITTEE_INDEX* = 54.GeneralizedIndex
```
etan-status: These should also link to the spec, and mention which field is being accessed (e.g., finalized root is the root inside finalized checkpoint)

beacon_chain/conf.nim
```diff
@@ -387,6 +387,18 @@ type
...
+        hidden
+        desc: "BETA: Serve data for enabling light clients to stay in sync with the network"
+        defaultValue: false
+        name: "serve-light-client-data"}: bool
+
```
etan-status: Introduce additional option for on-demand computation for testing.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -720,6 +736,94 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
     stateDur = stateTick - summariesTick,
     indexDur = Moment.now() - stateTick
 
+  # Initialize cached light client data
+  if dag.createLightClientData:
```
arnetheduck: not for this branch, but the other option here is to build the cache up in the "background" as far as our architecture permits - ie at the end of every slot, do a bit of processing - we will be doing something similar when backfilling and recreating historical states, and it's likely we can kill two birds with that stone in general

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -720,6 +736,94 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
+
+      # Process blocks
+      let lowSlot = max(altairStartSlot, dag.tail.slot)
+      var
+        tmpState = assignClone(dag.headState)
+        oldCheckpoint: Checkpoint
+        checkpointIndex = 0
+      for i in countdown(blocksBetween.high, blocksBetween.low):
+        blockRef = blocksBetween[i]
+        dag.withUpdatedState(dag.headState, blockRef.atSlot(blockRef.slot)) do:
```
arnetheduck: this will use a new StateCache for every block which contributes to the inefficiency.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -720,6 +736,94 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
+      var
+        blocksBetween = newSeqOfCap[BlockRef](headRef.slot - earliestSlot + 1)
+        blockRef = headRef
+      while blockRef.slot > earliestSlot:
+        blocksBetween.add blockRef
+        blockRef = blockRef.parent
+      blocksBetween.add blockRef
+
+      # Process blocks
+      let lowSlot = max(altairStartSlot, dag.tail.slot)
```
arnetheduck: why does this loop run for slots older than `computeEarliestLightClientSlot`? isn't `cacheLightClientData` pruned for those slots?

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,802 @@
...
+    blck: ForkyTrustedSignedBeaconBlock): T =
+  ## Reduce a given full block to just its `BeaconBlockHeader`.
+  BeaconBlockHeader(
+    slot: blck.message.slot,
+    proposer_index: blck.message.proposer_index,
+    parent_root: blck.message.parent_root,
+    state_root: blck.message.state_root,
+    body_root: blck.message.body.hash_tree_root())
+
+func fromBlock(
```
arnetheduck: this can move into `spec/` somewhere (`forks` maybe?), but we tend to express the converters in the other direction (`blck.toBeaconBlockHeader`) - it should probably operate on `SomeForkyBeaconBlock`

beacon_chain/sync/sync_protocol.nim
```diff
@@ -374,6 +379,118 @@ p2pProtocol BeaconSync(version = 1,
...
+      raise newException(ResourceUnavailableError,
+                         "No `OptimisticLightClientUpdate` available")
+
+    peer.updateRequestQuota(lightClientUpdateResponseCost)
+
+  proc lightClientBootstrap(
+      peer: Peer,
+      blockRoot: Eth2Digest,
+      response: SingleChunkResponse[altair.LightClientBootstrap])
+      {.async, libp2pProtocol("light_client_bootstrap", 1).} =
```
arnetheduck: maybe gate the protocol with a "_preview" name so as not to infect the namespace - or use version 0

beacon_chain/validators/validator_duties.nim
```diff
@@ -841,14 +853,14 @@ proc handleSyncCommitteeContributions(node: BeaconNode,
               slot, head, subnet_id = candidateAggregators[i].subcommitteeIdx
 
 proc handleProposal(node: BeaconNode, head: BlockRef, slot: Slot):
-    Future[BlockRef] {.async.} =
+    Future[tuple[head: BlockRef, didPropose: bool]] {.async.} =
```
arnetheduck: `didPropose` == returned head != passed-in head

beacon_chain/sync/sync_protocol.nim
```diff
@@ -374,6 +379,118 @@ p2pProtocol BeaconSync(version = 1,
     debug "Block root request done",
       peer, roots = blockRoots.len, count, found
 
+  proc bestLightClientUpdatesByRange(
+      peer: Peer,
+      startPeriod: SyncCommitteePeriod,
+      reqCount: uint64,
+      reqStep: uint64,
+      response: MultipleChunksResponse[altair.LightClientUpdate])
+      {.async, libp2pProtocol("best_light_client_updates_by_range", 1).} =
```
arnetheduck: actually - the protocol should not be registered with libp2p at all if the command line flags don't say so .. cc @zah how to do this with the macro?

etan-status: From @zah , via Discord 07 Feb:

> you just need to define the new protocol in a separate file where you'll say p2pProtocol LightClientSync
> The idea is that the Eth2Node object has a function that can be used to enable the individual protocols like this: node.addCapability(BeaconSync) 
> I'll have to take a look at these details, because some of this functionality may have bit rotted because it's not actively used

Not sure how to do that though, `addCapability` operates on a different type. But agree that this would be cleaner, especially when we want to merge this into `unstable` before the spec has been accepted upstream.

beacon_chain/validators/validator_duties.nim
etan-status: `head` also  updates when someone else proposed the block. We only want to submit `OptimisticLightClientUpdate` when one of our own validators did propose. Can rename the arg to highlight that, as I have meanwhile seen `proposed` terminology to just indicate any non-empty slot.

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: Yeah, it would be better (and necessary for the historic import unless we can access arbitrary states within a second). Once the caches are initialized, any incremental updates from future blocks happen in <100ms, and the caches are pruned on finality.

beacon_chain/sync/sync_protocol.nim
arnetheduck: indeed, I would see this as the main blocker for merging - this must not be visible for "normal" operation

etan-status: - [x] X

beacon_chain/validators/validator_duties.nim
arnetheduck: `handleProposal` returns the head passed in as parameter, when someone else is doing the proposing - that's what you would compare

that said, this will eventually need to be worked out a different way completely because blocks can be produced by validator clients and submitted via REST as well.

etan-status: Ahh, sorry, I had the other function in mind, that accepts `didPropose` as a parameter to decide whether an `OptimisticLightClientUpdate` should be sent. Yes, exactly, `handleProposal` returns `didPropose == returned head != passed-in head`, can also do the check on the call side (but it requires keeping track of the old head there to do the comparison).

For the REST call, an async wait is performed when a block was broadcast, then the optimistic light client update is sent using the same logic (`didPropose` forced to `true` in that case).

etan-status:
- Rebased to `unstable`.
- Update to latest proposed libp2p spec (`allowForceUpdate` parameter).
   https://github.com/ethereum/consensus-specs/pull/2802
- Verify update `fork_version` against local `RuntimeConfig` during light client sync.

etan-status:
- Updated `sync_protocol` for latest style (early return for `step == 0` / `count == 0` case, and trace before throwing exception).

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: You are right, there was a small amount of unnecessary data processed in some edge cases. Good catch.

etan-status:
Thanks to @arnetheduck for the detailed review!

- Document `GeneralizedIndex` constants in `altair.nim`.

- Add new option `--import-light-client-data=on-demand` to compute historic light client data on demand instead of at startup. May lead to missed validator duties.
- Replace `--import-historic-light-client-data=false` with `--import-light-client-data=only-new`.
- Replace `--import-historic-light-client-data=true` with `--import-light-client-data=full`.
- Harden `initBestLightClientUpdateForPeriod` / `initLightClientBootstrapForPeriod` for untrusted `period` input.
- Move light client database initialization from `blockchain_dag.nim` into `blockchain_dag_light_client.nim`.
- Move light client specifics from `block_pools_types.nim` into `block_pools_types_light_client.nim`.
- Wrap light client database objects into `LightClientDatabase` and renamed field names to be a bit shorter.
- Re-use `StateCache` when operating on multiple states in loops.
- Avoid caching of unnecessary data in edge cases during light client database initialization.

- Use version `0` for libp2p protocol IDs to signify this being for a pre-release specification.
- Move `handleProposal` changes to call side, and renamed `didPropose` to `didSubmitBlock`.

etan-status:
- Update test instructions in PR description for new command line options.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/3410

beacon_chain/sync/sync_protocol.nim
etan-status: I patched `Eth2Node.new` to not "mount" those new endpoints with the "switch", whatever that means, and they no longer appear in the `Started libp2p node` log. Whether this is enough, I don't know. It's a bit of a macro jungle 🙃

etan-status:
- Tag new libp2p req/rsp endpoints to be served only when the `--serve-light-client-data` is specified. See `eth2_network.nim` changes.

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
```diff
@@ -0,0 +1,85 @@
...
+    finality_branch*:
+      array[log2trunc(altair.FINALIZED_ROOT_INDEX), Eth2Digest]
+
+  CachedLightClientBootstrap* = object
+    ## Cached data from historical finalized epoch boundary blocks to improve
+    ## speed when creating future `LightClientBootstrap` instances.
+    current_sync_committee_branch*:
+      array[log2trunc(altair.CURRENT_SYNC_COMMITTEE_INDEX), Eth2Digest]
+
+  LightClientDatabase* = object
```
arnetheduck: `Cache` or `Pool` or something - `Database` is for persistent storage in the rest of the app

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,919 @@
...
+type
+  HashedBeaconStateWithSyncCommittee =
+    bellatrix.HashedBeaconState |
+    altair.HashedBeaconState
+
+  TrustedSignedBeaconBlockWithSyncAggregate =
+    bellatrix.TrustedSignedBeaconBlock |
+    altair.TrustedSignedBeaconBlock
+
+func fromBlock(
```
arnetheduck: `toBeaconBlockHeader`?

etan-status: Yes, have not rebased yet but this will use `toBeaconBlockHeader`.

beacon_chain/spec/datatypes/altair.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/3443

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
etan-status: Going with `LightClientCache`.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/3474

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/3474

etan-status:
- Rebase to `unstable` - this leaves only the Req/Resp changes in this PR :)

etan-status:
- Rebase to `unstable` (`Option` -> `Opt` for `LightClientBoostrap` endpoint)

beacon_chain/sync/sync_protocol.nim
```diff
@@ -439,6 +449,122 @@ p2pProtocol BeaconSync(version = 1,
     debug "Block root request done",
       peer, roots = blockRoots.len, count, found
 
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#bestlightclientupdatesbyrange
```
arnetheduck, line 452: we will likely be deprecating step support in range requests: https://github.com/ethereum/consensus-specs/pull/2856

etan-status: Thanks for the additional background. Removing here as well.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -439,6 +449,122 @@ p2pProtocol BeaconSync(version = 1,
...
+  # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#getlightclientbootstrap
+  proc lightClientBootstrap(
+      peer: Peer,
+      blockRoot: Eth2Digest,
+      response: SingleChunkResponse[altair.LightClientBootstrap])
+      {.async, libp2pProtocol("light_client_bootstrap", 0,
+                              isLightClientRequest = true).} =
+    trace "Received LC bootstrap request", peer, blockRoot
+    let dag = peer.networkState.dag
+    if not dag.serveLightClientData:
```
arnetheduck: this shouldn't happen, no? either the request is turned on, or the request gets rejected at the protocol negotiation level

etan-status: Yes, it should not be reachable. Replacing these with asserts.

arnetheduck (approved):

etan-status:
- Remove `step` parameter from `BestLightClientUpdatesByRange` message
- Replace unreachable `serveLightClientData` checks with asserts

PR #3343, opened 2022-01-31
Streamline the validator reward analysis

zah:
Notable improvements:

* A separate aggregation pass is no longer required.

* The user can opt to produce only aggregated data
  (resuing in a much smaller data set).

* Large portion of the number cruching in Jupyter is now done in C
  through the rich DataFrames API.

* Added support for comparisons against the "median" validator
  performance in the network.


arnetheduck:
How much complexity would it be to add back an option to work with counts instead of balances? Or is it better to simply revive the old tool from git? As a developer / debugging tool, I find it has become hard to reason about the output, compared to counting messages - the latter are independent, while balances introduce a dependency on past behavior of the validator, also relative other validators.

arnetheduck (approved):

zah:
Using counts is very easy. For each of the separate reward components, you can count the cases where `outcome == max_reward`. The aggregated files that are used to speed up the Jupyter reports adds some complexity to the picture, but you can easily produce another aggregated file with count data.

PR #3344, opened 2022-01-31
Refactor `handleValidatorExitCommand`

EmilIvanichkovv:
Make `validator exit command` work both with `JSON-RPC` and `REST` APIs
Fix problem with specifying rest-url using `localhost`
Change back exit error messages in `state_transition_block`

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1523,16 +1523,197 @@ proc initStatusBar(node: BeaconNode) {.raises: [Defect, ValueError].} =
 
   asyncSpawn statusBarUpdatesPollingLoop()
 
-proc handleValidatorExitCommand(config: BeaconNodeConf) {.async.} =
+proc getSignedExitMessage(config: BeaconNodeConf,
+                          validatorKeyAsStr: string,
+                          exitAtEpoch: Epoch,
+                          validatorIdx: uint64 ,
+                          fork: Fork,
+                          genesisValidatorsRoot: Eth2Digest): SignedVoluntaryExit =
   let
-    client = RestClientRef.new(
-      try:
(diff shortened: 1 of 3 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1523,16 +1523,194 @@ proc initStatusBar(node: BeaconNode) {.raises: [Defect, ValueError].} =
...
+        quit 1
+  except CatchableError as err:
+    fatal "Failed to send the signed exit message to the beacon node RPC",
+           err = err.msg
+    quit 1
+
+proc restValidatorExit(config: BeaconNodeConf) {.async.} =
+  let
+    address = if isNone(config.restUrlForExit):
+      resolveTAddress("127.0.0.1:47000")[0]
```
zah: This is not the right default port (5052). The code here should really use the `DefaultEth2RestPort` constant from the `spec/network` module.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1523,16 +1523,194 @@ proc initStatusBar(node: BeaconNode) {.raises: [Defect, ValueError].} =
...
+                                   signedExit.message, key).toValidatorSig()
+
+  signedExit
+
+type
+  ClientExitAction = enum
+    quiting = "q"
+    confirmation = "I understand the implications of submitting a voluntary exit"
+
+proc askForExitConfirmation(): string =
```
zah: Since you have the enum above, why don't you use it as the return value of this function?

zah (changes requested):

arnetheduck:
At some point, we should move this code to a separate module - trusted node sync started this transformation, and now that this command becomes long with multiple options etc, a new module that handles it would be appropriate. it's not a blocker for the PR, but certainly a nice-to-have

PR #3345, opened 2022-01-31
val-mon: remove redundant `_total` suffix from counters

arnetheduck:
It turns out nim-metrics adds this suffix on its own - it also turns out
some of the names are non-conventional and need follow-up.


stefantalpalaru:
You can get full naming control with gauges instead of counters, when that default suffix is not acceptable.

arnetheduck:
> You can get full naming control with gauges instead of counters, when that default suffix is not acceptable.

that'll change the metric TYPE which also ends up being off

PR #3348, opened 2022-02-01
Tune peering part 1

Menduist:
Extracted from https://github.com/status-im/nimbus-eth2/pull/3346

Includes:
- Request metadata_v2 (altair) by default instead of the v1
- Change the metadata pinger to a 3 failure-then-kick, instead of being time based
- Update kicker scorer to take into account topics which we're not subscribed to, to be sure that we will be able to publish correctly
- Add some metrics to give "fanout" health (in the same spirit of mesh health)

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -102,7 +102,8 @@ type
     lastReqTime*: Moment
     connections*: int
     enr*: Option[enr.Record]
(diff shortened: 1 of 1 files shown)


arnetheduck (approved):

arnetheduck:
looking forward to seeing these fanout metrics and how they stack up against the mesh, should prove valuable

PR #3354, opened 2022-02-02
Eth1 monitor fixes

zah:
* Fix a resource leak introduced in https://github.com/status-im/nimbus-eth2/pull/3279

* Don't restart the Eth1 syncing proggress from scratch in case of
  monitor failures during Eth2 syncing.

* Switch to the primary provider as soon as it is back online.

* Log the web3 credentials in fewer places

Other changes:

The 'web3 test' command has been enhanced to obtain and print more
data regarding the selected provider.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -169,8 +169,19 @@ proc init*(T: type BeaconNode,
     eventBus.emit("head-change", data)
   proc onChainReorg(data: ReorgInfoObject) =
     eventBus.emit("chain-reorg", data)
-  proc onFinalization(data: FinalizationInfoObject) =
-    eventBus.emit("finalization", data)
+  proc makeOnFinalizationCb(
+      # This `nimcall` functions helps for keeping track of what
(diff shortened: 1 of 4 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -169,8 +169,19 @@ proc init*(T: type BeaconNode,
...
   proc onChainReorg(data: ReorgInfoObject) =
     eventBus.emit("chain-reorg", data)
-  proc onFinalization(data: FinalizationInfoObject) =
-    eventBus.emit("finalization", data)
+  proc makeOnFinalizationCb(
+      # This `nimcall` functions helps for keeping track of what
+      # needs to be captured by the onFinalization closure.
+      eventBus: AsyncEventBus,
+      eth1Monitor: Eth1Monitor): OnFinalizedCallback {.nimcall.} =
+    static: doAssert (eventBus is ref) and (eth1Monitor is ref)
```
arnetheduck, line 177: time to switch to `ref Eth1Monitor`? I find that style so much more clear than `ref object`.

PR #3355, opened 2022-02-03
docs: don't rely on ncli_db pruneDatabase for reducing storage usage

tersec:
This is a more effective and reliable approach.

### DIFF
--- a/docs/the_nimbus_book/src/troubleshooting.md
+++ b/docs/the_nimbus_book/src/troubleshooting.md
@@ -65,7 +65,7 @@ If you find that `make update` causes the console to hang for too long, try runn
 The directory that stores the blockchain data of the testnet is `build/data/prater_shared_0` (if you're connecting to another testnet, replace `prater` with that testnet's name). If you've imported the wrong keys, and wish to start over, delete this repository.
 
 ### Sync problems
-If you’re experiencing sync problems, we recommend running `make clean-prater` to delete the database and restart your sync (make sure you’ve updated to the latest `master` first though).
+If you’re experiencing sync problems, we recommend running `make clean-prater` to delete the database and restart your sync (make sure you’ve updated to the latest `stable` first though).
 
 > **Warning**: `make clean-prater` will erase all of your syncing progress so far, so it should only be used as a last resort -- if your client gets stuck for a long time (because it's unable to find the right chain and/or stay with the same head value) and a normal restart doesn't improve things.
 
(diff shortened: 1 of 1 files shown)


docs/the_nimbus_book/src/troubleshooting.md
```diff
@@ -65,25 +65,12 @@ If you find that `make update` causes the console to hang for too long, try runn
...
-```
-
-This will create `nbc_pruned.sqlite3` files in `build/data/shared_prater_0/db`, which you can use in place of the orginal `nbc.sqlite3` files. We recommend you hold onto the originals until you've verified that your validator is behaving as expected with the pruned files.
-
-Options:
-- `--keepOldStates` (boolean):  Keep pre-finalisation states; defaults to `true`.
-- `--verbose` (boolean): Print a more verbose output to the console; defaults to `false`.
-
-
+If you're running out of storage, you can [prune](https://blog.ethereum.org/2015/06/26/state-tree-pruning/) the database of unnecessary blocks and states by [syncing from checkpoint files](./trusted-node-sync.html#sync-from-checkpoint-files) and using the result to replace the existing database.
```
arnetheduck: I'd generally  remove this section altogether - it hails from a distant past where state bloat was a real thing - nowadays, it's really not that big of a difference - also, the next feature scheduled for checkpoint sync is to recreate the historical states.

arnetheduck: ie the user experience should really not be "run this magic command and magically regain lots of storage that wasn't really needed..

rather, with the implementation of era files we'll be pruning storage as part of normal operation (optionally to an era file) and at that point, it should also clean up junk in the database.

tersec: https://github.com/status-im/nimbus-eth2/pull/3355/commits/c1149617d268a66492882c5f1cb20fc89eb13e13

arnetheduck (approved):

docs/the_nimbus_book/src/troubleshooting.md
```diff
@@ -65,26 +65,10 @@ If you find that `make update` causes the console to hang for too long, try runn
 The directory that stores the blockchain data of the testnet is `build/data/prater_shared_0` (if you're connecting to another testnet, replace `prater` with that testnet's name). If you've imported the wrong keys, and wish to start over, delete this repository.
 
 ### Sync problems
-If you’re experiencing sync problems,  we recommend running `make clean-prater` to delete the database and restart your sync (make sure you’ve updated to the latest `master` first though).
+If you’re experiencing sync problems, we recommend running `make clean-prater` to delete the database and restart your sync (make sure you’ve updated to the latest `master` first though).
```
unixpi:
```suggestion
If you’re experiencing sync problems, we recommend running `make clean-prater` to delete the database and restart your sync (make sure you’ve updated to the latest `stable` first though).
```

PR #3361, opened 2022-02-04
Jordan/explore catalyst test network

mjfh:


tersec:
Just to make it easier to see the changes per se, it'd probably be expedient to set the merge base to `unstable` rather than `stable`. It doesn't necessarily matter as such for a branch which isn't meant to be merged, but the vast majority of the displayed commits and changed files aren't relevant either way to catalyst/kintsugi/the merge.

PR #3366, opened 2022-02-09
bump submodules

stefantalpalaru:
and add Nim-1.6 compatibility

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -1333,19 +1333,15 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
     eth1SyncedTo = targetBlock
     eth1_synced_head.set eth1SyncedTo.toGaugeValue
 
-proc start(m: Eth1Monitor, delayBeforeStart: Duration) =
+proc start(m: Eth1Monitor, delayBeforeStart: Duration) {.gcsafe.} =
   if m.runFut.isNil:
     let runFut = m.startEth1Syncing(delayBeforeStart)
     m.runFut = runFut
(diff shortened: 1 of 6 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1333,19 +1333,15 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
...
     eth1_synced_head.set eth1SyncedTo.toGaugeValue
 
-proc start(m: Eth1Monitor, delayBeforeStart: Duration) =
+proc start(m: Eth1Monitor, delayBeforeStart: Duration) {.gcsafe.} =
   if m.runFut.isNil:
     let runFut = m.startEth1Syncing(delayBeforeStart)
     m.runFut = runFut
-    runFut.addCallback do (p: pointer):
+    runFut.addCallback do (p: pointer) {.gcsafe.}:
       if runFut.failed:
```
zah, line 1341: Why is the change below necessary?

stefantalpalaru: Nim-1.6 figured out that a future's "error" field is always a "CatchableError", and it's right about it: https://github.com/status-im/nim-chronos/blob/87197230779002a2bfa8642f0e2ae07e2349e304/chronos/asyncfutures2.nim#L40

arnetheduck:
There are some pretty significant module bumps here: these need to be vetted independently before merging this branch: many of the sub-project tend to not keep backwards compat between commits or can significantly / negatively impact important parts of the application (faststreams, libp2p, nim-eth etc require separate bumps that review the changes in those projects and how they impact eth2)

stefantalpalaru:
Rebased, rebumped, reloaded.

zah (approved):

PR #3367, opened 2022-02-09
Fix #3267

### LINKED ISSUE #3267: Support Bellatrix beacon blocks in REST validator client and server
https://github.com/status-im/nimbus-eth2/pull/3213 describes the current situation based on a brief exploration:

The validator client can't publish Bellatrix blocks:
https://github.com/status-im/nimbus-eth2/blob/c7be88b432cb0612ede893ccbf2941af69882dd5/beacon_chain/validator_client/api.nim#L767-L779

While at some level, there's a function overload straightforwardly missing, even filling it in here:
https://github.com/status-im/nimbus-eth2/blob/c7be88b432cb0612ede893ccbf2941af69882dd5/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim#L83-L91
just reproduces the `nim-presto` issue the comment describes.

And even if it could, the REST API on the other side wouldn't know what to do with them:
https://github.com/status-im/nimbus-eth2/blob/c7be88b432cb0612ede893ccbf2941af69882dd5/beacon_chain/rpc/rest_beacon_api.nim#L752-L780

Which is all summarized by
https://github.com/status-im/nimbus-eth2/blob/6ef3834f4ac4fa3dd8c0191686ba07c0233e8bc7/beacon_chain/validators/validator_duties.nim#L1224-L1225

cheatfate:
Enable bellatrix blocks support.
Refactor publishBlock algorithm for json block deserialization.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -756,7 +756,10 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         if res.isErr():
           return RestApiResponse.jsonError(Http400, InvalidBlockObjectError,
                                            $res.error())
-        ForkedSignedBeaconBlock(res.get())
+        var forked = ForkedSignedBeaconBlock(res.get())
+        withBlck(forked):
+          blck.root = hash_tree_root(blck.message)
+        forked
 
     let res = await node.sendBeaconBlock(forkedBlock)
     if res.isErr():
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -760,7 +760,7 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
 
     let res = await node.sendBeaconBlock(forkedBlock)
     if res.isErr():
-      return RestApiResponse.jsonError(Http503, BeaconNodeInSyncError)
+      return RestApiResponse.jsonError(Http400, $res.error())
     if not(res.get()):
       return RestApiResponse.jsonError(Http202, BlockValidationError)
 
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -747,45 +747,18 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
   # https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock
   router.api(MethodPost, "/eth/v1/beacon/blocks") do (
     contentBody: Option[ContentBody]) -> RestApiResponse:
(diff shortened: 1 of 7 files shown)


tersec:
For https://github.com/status-im/nimbus-eth2/issues/3267

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -738,12 +739,304 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: ForkedBeaconBlock) {.
...
-    when false:
-      # TODO SerializationError
-      writer.writeField("data", value.bellatrixData)
+    writer.writeField("version", "bellatrix")
+    writer.writeField("data", value.bellatrixData)
   writer.endRecord()
 
+## RestPublishedBeaconBlockBody
+proc readValue*(reader: var JsonReader[RestJson],
+                value: var RestPublishedBeaconBlockBody) {.
```
arnetheduck, line 748: why this new approach based on field members? the approach where we first read the header and use the slot number to enforce a "correct" body seems more accurate (and it matches the algorithm we use for loading SSZ)

cheatfate: `publishBlock` call do not accept any other encoding except `application/json`. So we can't reuse `slot number` approach here like we made for SSZ. Also it will decode json using single pass. Before this PR we have used N passes (depends on number of forks).

arnetheduck: alright - we'll see for how long we can sustain this strategy - a body is not _guaranteed_ to change between forks, so at some point it might lead to ambiguity - the slot is the only sure way of telling what type is expected

cheatfate: I know that i have removed `slot` check and i can actually restore it, but after deserialization. The specification is not very clear here.

arnetheduck: it's probably a good idea to have it here, yes - the block would sooner or later be thrown out anyway (because it will not match the fork of the state), but it'll be easier to diagnose errors if they're caught here

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -757,6 +757,23 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
           return RestApiResponse.jsonError(Http400, InvalidBlockObjectError,
                                            $res.error())
         var forked = ForkedSignedBeaconBlock(res.get())
+
+        case forked.kind
```
arnetheduck:
```suggestion
        if forked.kind != cfg.blockForkAtEpoch(getForkedBlockField(forked, slot).epoch)
```

arnetheduck (approved):

PR #3372, opened 2022-02-11
move wallet/deposit commands to separate files

arnetheduck:
These commands have little to do with the "normal" beacon node operation
- ergo, they deserve to live in their own module.

* clean up imports/exports


arnetheduck:
merging as this was discussed before, no functional changes

PR #3378, opened 2022-02-11
Add CORS support for the REST services

zah:
The added options work in opt-in fashion. If they are not specified,
the server will respond to all requests as if the CORS specification
doesn't exist. This will result in errors in CORS-enabled clients.

If the options are provided, the server will enforce the Origin by
returning 403 Forbidden when the requests headers don't match the
desired value.

Please note that future versions may support more than one allowed
origin. The option names will stay the same, but the user will be
able to repeat them on the command line (similar to other options
such as --web3-url).

To be documented in the guide in a separate PR.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -331,6 +331,11 @@ type
         defaultValueDesc: "127.0.0.1"
         name: "rest-address" }: ValidIpAddress
 
+      restAllowedOrigin* {.
+        desc: "Limit the access to the REST API to a particular hostname " &
+              "(for CORS-enabled clients such as browsers)"
+        name: "rest-allow-origin" }: Option[string]
+
       restCacheSize* {.
         defaultValue: 3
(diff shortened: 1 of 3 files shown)


zah:
The decision to disallow access from CORS-enabled clients by default was a security consideration. We don't want to make it possible for malicious web-sites to try connecting to locally running beacon node (on localhost) that can be attacked.

arnetheduck:
Geth appears to have separate option for options for client and server side enforcement - how are the two related?

```
  --http.corsdomain value             Comma separated list of domains from which to accept cross origin requests (browser enforced)
  --http.vhosts value                 Comma separated list of virtual hostnames from which to accept requests (server enforced). Accepts '*' wildcard. (default: "localhost")
```

zah:
The `vhosts` option sounds like something enforcing the `Host` header value on the requests. In other words, it would prevent requests to be accidentally routed to the API on a server with multiple DNS hostnames.

PR #3385, opened 2022-02-13
fix `firstSlot` computation for backfill sync

etan-status:
When initializing backfill sync, the implementation intends to start at
the first unknown slot (`1` before tail). However, an incorrect variable
is passed, and backfill sync actually starts at the tail slot instead.
This patch corrects this by passing the intended variable. The problem
was introduced with the original backfill implementation at #3263.


arnetheduck:
cc @cheatfate I think this might be a general problem for forward sync as well, no? we re-download the head block on startup

arnetheduck:
this needs a manual test, it's not covered by test suite

PR #3386, opened 2022-02-13
use isZeroMemory for Eth2Digest comparisons

tersec:
Each `foo == default(Eth2Digest)`, `foo != default(Eth2Digest)`, `foo == Eth2Digest()`, or `foo != Eth2Digest()` creates one additional 32-byte `nimZeroMem()` call every time the comparison's done and each reference to an anonymous r-value `Eth2Digest` used for the comparison allocates another 32 bytes on the local C stack up to stack alignment considerations.

For example:
```nim
var a: Eth2Digest
echo a == default(Eth2Digest)
```

Adds one additional local C array, for the `default(Eth2Digest)` in `== default(Eth2Digest)`, and calls `nimZeroMem()` on it before comparing `a` and the object it just constructed.

```nim
var a: Eth2Digest
for i in [0, 1]:
  echo a == default(Eth2Digest)
```
calls `nimZeroMem` 3 times in its main function (once for `a`, as above, but now once per loop iteration). However, scope-wise, there's still only one distinct Nim-anonymous `Eth2Digest`, so it uses the same stack space as the first example in terms of `Eth2Digest` objects (excluding the loop counter).

```nim
var a: Eth2Digest
echo a == default(Eth2Digest)   # this
echo a == default(Eth2Digest)   # and this Eth2Digest are distinct for Nim
```
does not combine the two `default(Eth2Digest)`, and treats them separately, for 3 `nimZeroMem`s including for `a`, and both anonymous/temporary `default(Eth2Digest)` values at least notionally getting a C local variable.

With decent lifetime analysis and optimization, a C compiler should notice that they never coexist and be able to reuse the same stack space, but that's not ideal to rely on.

There are a few other places which compare with `default(Foo)`, but those aren't necessarily appropriate to use `isZeroMemory` for. This PR targets only `Eth2Digest`, to which it is applicable.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -9,7 +9,6 @@
 
 import
   chronicles,
-  stew/objects,
 
   ../spec/datatypes/[phase0, altair],
   ../spec/[helpers]
@@ -121,7 +120,7 @@ func isAncestorOf*(a, b: BlockRef): bool =
   isAncestor
 
 func link*(parent, child: BlockRef) =
-  doAssert (not (parent.root.isZeroMemory or child.root.isZeroMemory)),
+  doAssert (not (parent.root.isZero or child.root.isZero)),
     "blocks missing root!"
   doAssert parent.root != child.root, "self-references not allowed"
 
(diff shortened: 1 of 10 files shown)


arnetheduck:
woth adding an `isZero` function for Eth2Digest in particular, given its special casing?

arnetheduck (approved):

tersec:
> woth adding an `isZero` function for Eth2Digest in particular, given its special casing?

https://github.com/status-im/nimbus-eth2/pull/3386/commits/fd01c282ee6c55744df6785a8f267a39e235601b

PR #3387, opened 2022-02-14
clean up / document init

arnetheduck:
* drop `immutable_validators` data (pre-altair)
* document versions where data is first added
* avoid needlessly loading genesis block data on startup
* add a few more internal database consistency checks
* remove duplicate state root lookup on state load

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -97,7 +97,7 @@ type
     genesisDeposits*: DepositsSeq
 
     # immutableValidatorsDb only stores the total count; it's a proxy for SQL
-    # queries.
+    # queries. (v1.4.0+)
     immutableValidatorsDb*: DbSeq[ImmutableValidatorDataDb2]
     immutableValidators*: seq[ImmutableValidatorData2]
 
@@ -123,6 +123,7 @@ type
     summaries: KvStoreRef # BlockRoot -> BeaconBlockSummary
 
     finalizedBlocks*: FinalizedBlocks
+      ## Blocks that are known to be finalized, per the latest head (v1.7.0+)
 
   DbKeyKind = enum
     kHashToState
@@ -428,10 +429,12 @@ proc new*(T: type BeaconChainDB,
     summaries = kvStore db.openKvStore("beacon_block_summaries", true).expectDb()
     finalizedBlocks = FinalizedBlocks.init(db, "finalized_blocks").expectDb()
 
-  # `immutable_validators` stores validator keys in compressed format - this is
+  # Versions prior to 1.4.0 (altair) stored validators in `immutable_validators`
+  # which stores validator keys in compressed format - this is
   # slow to load and has been superceded by `immutable_validators2` which uses
-  # uncompressed keys instead. The migration is lossless but the old table
-  # should not be removed until after altair, to permit downgrades.
+  # uncompressed keys instead. We still support upgrading a database from the
+  # old format, but don't need to support downgrading, and therefore safely can
+  # remove the keys
   let immutableValidatorsDb1 =
       DbSeq[ImmutableValidatorData].init(db, "immutable_validators").expectDb()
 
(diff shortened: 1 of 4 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -428,10 +429,12 @@ proc new*(T: type BeaconChainDB,
...
 
-  # `immutable_validators` stores validator keys in compressed format - this is
+  # Versions prior to 1.4.0 (altair) stored validators in `immutable_validators`
+  # which stores validator keys in compressed format - this is
   # slow to load and has been superceded by `immutable_validators2` which uses
-  # uncompressed keys instead. The migration is lossless but the old table
-  # should not be removed until after altair, to permit downgrades.
+  # uncompressed keys instead. We still support upgrading a database from the
+  # old format, but don't need to support downgrading, and therefore safely can
+  # remove the keys
```
mratsim, line 437: I don't remember having a bench for this.
I only bench compressed key deserialization: https://github.com/status-im/nim-blscurve/blob/0237e4e/benchmarks/bls_signature.nim#L37-L43

Though it makes sense, decompression involves modular square root, which is over 400x more expensive than other operations:
image

https://github.com/supranational/blst/blob/eb6151961c133a930420e844e1a84708fbb4f6a4/src/e1.c#L236-L259

image

The tradeoff is 2x the size.
Assuming 300k validators that's 48 bytes (381-bit.nextPowerOfTwo() / 8) x 2 x 300k = 28.8MB (instead of 14.4)

beacon_chain/beacon_chain_db.nim
```diff
@@ -446,6 +449,10 @@ proc new*(T: type BeaconChainDB,
       ))
   immutableValidatorsDb1.close()
 
+  # Safe because nobody will be downgrading to pre-altair versions
+  # TODO: drop table maybe? that would require not creating the table just above
+  discard db.exec("DELETE FROM immutable_validators;")
```
mratsim, line 454: I think we need to drop the table. And I should do the same with the old slashing DB.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -462,64 +472,70 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
+  # history
   for blck in db.getAncestorSummaries(headRoot):
-    if midRef != nil and blck.summary.slot == midRef.slot:
-      if midRef.root != blck.root:
+    if midRef != nil and blck.summary.slot <= midRef.slot:
+      if midRef.slot != blck.summary.slot or midRef.root != blck.root:
         fatal "Finalized block table does not match ancestor summaries, database corrupt?",
           head = shortLog(headRoot), cur = shortLog(curRef),
-          midref = shortLog(midRef), blck = shortLog(blck.root)
+          mid = shortLog(midRef), blck = shortLog(blck.root)
```
mratsim, line 539: What is `mid`?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -462,64 +472,70 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
-    withBlck(genesisBlock): BlockRef.init(genesisBlockRoot, blck.message)
 
+  # Backfills are blocks that we have in the database, but can't construct a
+  # state for without replaying from genesis
   var
     backfillBlocks = newSeq[Eth2Digest](tailRef.slot.int)
+    # This is where we'll start backfilling, worst case - we might refine this
+    # while loading blocks, in case backfilling has happened already
     backfill = withBlck(tailBlock): blck.message.toBeaconBlockSummary()
     midRef: BlockRef
```
mratsim, line 492: midRef needs a comment on what it holds

beacon_chain/beacon_chain_db.nim
arnetheduck: indeed, this was benchmarked at the time  it was introduced and the difference was significant - you can still see it today: the time to write a checkpoint state from SSZ to the database on first startup is dominated by the time it takes to convert the keys from compressed to uncompressed

beacon_chain/beacon_chain_db.nim
arnetheduck: as the comment says, the check above recreates the table - this would need more surgery on the codebase

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: 3d6d0b97a465f2c94b03e431b19898a189899da5

PR #3389, opened 2022-02-14
VC: Fix forks handling.

cheatfate:
This PR also had a fix for `bellatrix` serialization/deserialization procedures.

### DIFF
--- a/beacon_chain/validator_client/common.nim
+++ b/beacon_chain/validator_client/common.nim
@@ -99,7 +99,8 @@ type
     runSlotLoop*: Future[void]
     beaconClock*: BeaconClock
     attachedValidators*: ValidatorPool
-    fork*: Option[Fork]
+    forks*: Option[seq[Fork]]
+    forksAvailable*: AsyncEvent
     attesters*: AttesterMap
     proposers*: ProposerMap
     beaconGenesis*: RestGenesis
@@ -241,3 +242,16 @@ proc getValidator*(vc: ValidatorClientRef,
       none[AttachedValidator]()
     else:
       some(validator)
+
+proc forkAtEpoch*(vc: ValidatorClientRef, epoch: Epoch): Option[Fork] =
+  if vc.forks.isSome():
+    let schedule = vc.forks.get()
+    # If schedule is present, it MUST not be empty.
+    doAssert(len(schedule) > 0)
+    var res: Fork
+    for item in schedule:
+      if item.epoch <= epoch:
+        res = item
+      else:
+        break
+    return some(res)
--- a/beacon_chain/validator_client/common.nim
+++ b/beacon_chain/validator_client/common.nim
@@ -99,7 +99,7 @@ type
     runSlotLoop*: Future[void]
     beaconClock*: BeaconClock
     attachedValidators*: ValidatorPool
-    forks*: Option[seq[Fork]]
+    forks*: seq[Fork]
     forksAvailable*: AsyncEvent
     attesters*: AttesterMap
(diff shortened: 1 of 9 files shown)


beacon_chain/validator_client/duties_service.nim
```diff
@@ -299,14 +299,28 @@ proc waitForNextSlot(service: DutiesServiceRef,
...
+
+  debug "Attester duties loop waiting for fork schedule update"
+  await vc.forksAvailable.wait()
+
   while true:
-    await vc.pollForAttesterDuties()
+    if vc.forks.isSome():
+      await vc.pollForAttesterDuties()
+    else:
+      debug "Fork schedule is not available yet, skipping attester duties"
```
arnetheduck: sleep?

arnetheduck: or.. why would we enter this part of the loop if vc.forksAvailable triggered?

arnetheduck: another way to say the same thing: why read forks from `vc.forks` on every iteration?

beacon_chain/validator_client/common.nim
```diff
@@ -241,3 +242,16 @@ proc getValidator*(vc: ValidatorClientRef,
       none[AttachedValidator]()
     else:
       some(validator)
+
+proc forkAtEpoch*(vc: ValidatorClientRef, epoch: Epoch): Option[Fork] =
+  if vc.forks.isSome():
```
arnetheduck: it seems a lot easier to make `vc.forks` a seq: if it's empty, then it's the same as `none` now, and the logic will "naturally" not find a  fork

beacon_chain/validator_client/duties_service.nim
cheatfate: Because fork schedule can be updated, e.g. for long-running validator client it could see `[phase0, altair]` for months and after some time it will see `[phase0, altair, bellatrix]`. I understand that its possible that there will be some changes needed in `validator_client` to properly support new fork, but its also possible that there will be no changes needed from validator_client's point of view. For example even if `bellatrix` has `execution_payload` in `BeaconBlock`, `validator_client` still uses the same approach to get block data, sign it and publish. So there no changes needed for `validator_client` to support `bellatrix` fork.

arnetheduck: >  So there no changes needed for validator_client to support bellatrix fork.

well .. it needs to decode the block and hash it so it knows what it's signing - but more interesting is this: why would it _become_ empty after being set? if that's possible, `forksAvailable` looks like an incomplete change

cheatfate: Ok you right, but some of the forks could be made without significant changes.
Anyway i'm removing this checks.

arnetheduck (approved):

beacon_chain/validator_client/common.nim
```diff
@@ -241,3 +242,14 @@ proc getValidator*(vc: ValidatorClientRef,
       none[AttachedValidator]()
     else:
       some(validator)
+
+proc forkAtEpoch*(vc: ValidatorClientRef, epoch: Epoch): Fork =
+  # If schedule is present, it MUST not be empty.
+  doAssert(len(vc.forks) > 0)
+  var res: Fork
+  for item in vc.forks:
```
arnetheduck, line 250: technically, sorting the forks in the other direction and returning the "first" found  is simpler, but whatever - this is the kind of function I'd add a test for, regardless

PR #3393, opened 2022-02-14
Minify slashing protection before SQLite

### LINKED ISSUE #3390: Investigate slashing protection import performance with genesis validators
It was reported that loading the slashing protection DB from a genesis validators takes hours.

This is both surprising and not:
- surprising because we use raw SQLite now which should be pretty efficient, especially on SSD
- unsurprising because apparently Lighthouse had to solve that issue: https://github.com/sigp/lighthouse/pull/2347 (the report also came from a Prysm user)

mratsim:
Fix #3390

Seems like we're hit by the issue described when users migrated from Prysm to Lighthouse with issue dealing with validators from genesis:
- https://github.com/sigp/lighthouse/issues/2354
- https://github.com/prysmaticlabs/prysm/issues/8893

WIP, edge cases to fix

`siPartial` for partial import might not be useful anymore with minification by default.
See https://github.com/eth-clients/slashing-protection-interchange-tests/releases/tag/v5.1.0-alpha.1


mratsim:
Minimal changes to unblock users.

Left for the future:
1. The metadata table is a remnant of v1 and can be deleted from the slashing protection DB going forward. Then for the release after the next hardfork we can assume that everyone upgraded (or if they didn't, didn't attest for so long that there is no slashing risk unless they have a clock issue and attested in the far future) and so can delete all related code for good.
2. There is a sorting on import that is extra unneeded work. It might be a bottleneck but I would be surprised since everything is in memory it wouldn't take minutes to process even billions epochs to sort, not even talking about hours. Aka I'm betting it's Sqlite IO that is the issue.
3. It would be nice to have nim-eth/sqlite use Option/Opt/Result to handle NULL in `exec`

beacon_chain/validators/slashing_protection_common.nim
```diff
@@ -294,6 +294,7 @@ proc importInterchangeV5Impl*(
         continue
       key.get()
 
+    # TODO: with minification sorting is unnecessary, cleanup
```
tersec, line 297: Does this `TODO` apply already, or only once other precondition are met?

mratsim: It already applies. Basically I wanted to get that ASAP to unblock users.

tersec (approved):

PR #3394, opened 2022-02-15
era: load blocks and states

arnetheduck:
Era files contain finalized history and can be thought of as an
alternative source for block and state data that allows clients to avoid
syncing this information from the P2P network - the P2P network is then
used to "top up" the client with the most recent data. They can be
freely shared in the community via whatever means (http, torrent, etc)
and serve as a permanent cold store of consensus data (and, after the
merge, execution data) for history buffs and bean counters alike.

This PR gently introduces support for loading blocks and states in two
cases: block requests from rest/p2p and frontfilling when doing
checkpoint sync.

The era files are used as a secondary source if the information is not
found in the database - compared to the database, there are a few key
differences:

* the database stores the block indexed by block root while the era file
indexes by slot - the former is used only in rest, while the latter is
used both by p2p and rest.
* when loading blocks from era files, the root is no longer trivially
available - if it is needed, it must either be computed (slow) or cached
(messy) - the good news is that for p2p requests, it is not needed
* in era files, "framed" snappy encoding is used while in the database
we store unframed snappy - for p2p2 requests, the latter requires
recompression while the former could avoid it
* front-filling is the process of using era files to replace backfilling
- in theory this front-filling could happen from any block and
front-fills with gaps could also be entertained, but our backfilling
algorithm cannot take advantage of this because there's no (simple) way
to tell it to "skip" a range.
* front-filling, as implemented, is a bit slow (10s to load mainnet): we
load the full BeaconState for every era to grab the roots of the blocks
- it would be better to partially load the state - as such, it would
also be good to be able to partially decompress snappy blobs
* lookups from REST via root are served by first looking up a block
summary in the database, then using the slot to load the block data from
the era file - however, there needs to be an option to create the
summary table from era files to fully support historical queries

To test this, `ncli_db` has an era file exporter: the files it creates
should be placed in an `era` folder next to `db` in the data directory.
What's interesting in particular about this setup is that `db` remains
as the source of truth for security purposes - it stores the latest
synced head root which in turn determines where a node "starts" its
consensus participation - the era directory however can be freely shared
between nodes / people without any (significant) security implications,
assuming the era files are consistent / not broken.

There's lots of future improvements to be had:

* we can drop the in-memory `BlockRef` index almost entirely - at this
point, resident memory usage of Nimbus should drop to a cool 500-600 mb
* we could serve era files via REST trivially: this would drop backfill
times to whatever time it takes to download the files - unlike the
current implementation that downloads block by block, downloading an era
at a time almost entirely cuts out request overhead
* we can "reasonably" recreate detailed state history from almost any
point in time, turning an O(slot) process into O(1) effectively - we'll
still need caches and indices to do this with sufficient efficiency for
the rest api, but at least it cuts the whole process down to minutes
instead of hours, for arbitrary points in time

### DIFF
--- a/beacon_chain/consensus_object_pools/block_pools_types.nim
+++ b/beacon_chain/consensus_object_pools/block_pools_types.nim
@@ -15,12 +15,12 @@ import
   # Internals
   ../spec/[signatures_batch, forks, helpers],
   ../spec/datatypes/[phase0, altair, bellatrix],
-  ".."/beacon_chain_db,
+  ".."/[beacon_chain_db, era_db],
   ../validators/validator_monitor,
   ./block_dag, block_pools_types_light_client
 
 export
-  options, sets, tables, hashes, helpers, beacon_chain_db, block_dag,
+  options, sets, tables, hashes, helpers, beacon_chain_db, era_db, block_dag,
   block_pools_types_light_client, validator_monitor
 
 # ChainDAG and types related to forming a DAG of blocks, keeping track of their
@@ -119,6 +119,8 @@ type
       ## the DAG and the canonical head are stored here, as well as several
       ## caches.
 
+    era*: EraDB
+
     validatorMonitor*: ref ValidatorMonitor
 
     forkBlocks*: HashSet[KeyedBlockRef]
@@ -142,7 +144,11 @@ type
     backfill*: BeaconBlockSummary
       ## The backfill points to the oldest block with an unbroken ancestry from
       ## dag.tail - when backfilling, we'll move backwards in time starting
-      ## with the parent of this block until we reach `genesis`.
+      ## with the parent of this block until we reach `frontfill`.
+
+    frontfillBlocks*: seq[Eth2Digest]
+      ## A temporary cache of blocks that we could load from era files, once
+      ## backfilling reaches this point - empty when not backfilling.
 
     heads*: seq[BlockRef]
     ## Candidate heads of candidate chains
@@ -293,6 +299,14 @@ type
     epoch*: Epoch
 
 template head*(dag: ChainDAGRef): BlockRef = dag.headState.blck
+
+template frontfill*(dagParam: ChainDAGRef): BlockId =
+  let dag = dagParam
+  if dag.frontfillBlocks.lenu64 > 0:
+    BlockId(
+      slot: Slot(dag.frontfillBlocks.lenu64 - 1), root: dag.frontfillBlocks[^1])
+  else:
+    dag.genesis
 
 template epoch*(e: EpochRef): Epoch = e.key.epoch
 
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -13,8 +13,8 @@ import
   metrics, snappy, chronicles,
   ../spec/[beaconstate, eth2_merkleization, eth2_ssz_serialization, helpers,
     state_transition, validator],
-  ".."/beacon_chain_db,
   ../spec/datatypes/[phase0, altair, bellatrix],
+  ".."/[beacon_chain_db, era_db],
   "."/[block_pools_types, block_quarantine]
 
 export
@@ -112,6 +112,32 @@ proc updateFinalizedBlocks*(db: BeaconChainDB, newFinalized: openArray[BlockId])
   db.withManyWrites:
     for bid in newFinalized:
       db.finalizedBlocks.insert(bid.slot, bid.root)
+
+proc updateFrontfillBlocks*(dag: ChainDAGRef) =
+  # When backfilling is done and manages to reach the frontfill point, we can
+  # write the frontfill index knowing that the block information in the
+  # era files match the chain
+  if dag.db.db.readOnly: return # TODO abstraction leak - where to put this?
+
+  if dag.frontfillBlocks.len == 0 or dag.backfill.slot > 0:
+    return
+
+  info "Writing frontfill index", slots = dag.frontfillBlocks.len
+
+  dag.db.withManyWrites:
+    let low = dag.db.finalizedBlocks.low.expect(
+      "wrote at least tailRef during init")
+    let blocks = min(low.int, dag.frontfillBlocks.len - 1)
+    var parent: Eth2Digest
+    for i in 0..blocks:
+      let root = dag.frontfillBlocks[i]
+      if not isZero(root):
+        dag.db.finalizedBlocks.insert(Slot(i), root)
+        dag.db.putBeaconBlockSummary(
+          root, BeaconBlockSummary(slot: Slot(i), parent_root: parent))
+        parent = root
+
+    reset(dag.frontfillBlocks)
 
 func validatorKey*(
     dag: ChainDAGRef, index: ValidatorIndex or uint64): Option[CookedPubKey] =
@@ -444,7 +470,10 @@ proc getForkedBlock*(db: BeaconChainDB, root: Eth2Digest):
 proc getBlock*(
     dag: ChainDAGRef, bid: BlockId,
     T: type ForkyTrustedSignedBeaconBlock): Opt[T] =
-  dag.db.getBlock(bid.root, T)
+  dag.db.getBlock(bid.root, T) or
+    getBlock(
+      dag.era, getStateField(dag.headState, historical_roots).asSeq,
+      bid.slot, Opt[Eth2Digest].ok(bid.root), T)
 
 proc getBlockSSZ*(dag: ChainDAGRef, bid: BlockId, bytes: var seq[byte]): bool =
   # Load the SSZ-encoded data of a block into `bytes`, overwriting the existing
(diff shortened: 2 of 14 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -407,31 +407,71 @@ proc getForkedBlock*(db: BeaconChainDB, root: Eth2Digest):
   else:
     err()
 
-proc getForkedBlock*(
-    dag: ChainDAGRef, root: Eth2Digest): Opt[ForkedTrustedSignedBeaconBlock] =
-  dag.db.getForkedBlock(root)
+proc getBlock*[T: ForkyTrustedSignedBeaconBlock](
```
zah: It's wise to avoid APIs where a type must be supplied as a generic parameter, because these have a number of limitations in the compiler:

* No overloading/specialization
* No default values
* Worse error messages

If the user of the API needs to supply a type explicitly, the better alternative is to use a regular `type` parameter.

beacon_chain/era_db.nim
```diff
@@ -0,0 +1,286 @@
...
+    db = EraDB.new(
+      defaultRuntimeConfig, dbPath,
+      Eth2Digest(
+        data: array[32, byte].initCopyFrom([byte 0x4b, 0x36, 0x3d, 0xb9])))
+    historical_roots = [
+      Eth2Digest(
+        data: array[32, byte].initCopyFrom([byte 0x40, 0xcf, 0x2f, 0x3c]))]
+
+  var got8191 = false
+  for slot, root in db.getSummaries(historical_roots, Era(0)):
```
zah: There is no current definition of `getSummaries`

beacon_chain/era_db.nim
```diff
@@ -0,0 +1,286 @@
...
+
+  result.ok(default(T))
+  try:
+    readSszBytes(tmp, result.get(), updateRoot = root.isNone)
+    if root.isSome():
+      result.get().root = root.get()
+  except CatchableError as exc:
+    result.err()
+
+proc getStateSZ*(
```
zah, line 141: What the intended meaning of this name?

beacon_chain/era_db.nim
```diff
@@ -0,0 +1,286 @@
...
+  ? f[].setFilePos(0, SeekPosition.SeekEnd).mapErr(ioErrorMsg)
+
+  # Last in the file is the state index
+  let
+    stateIdxPos = ? f[].findIndexStartOffset()
+  ? f[].setFilePos(stateIdxPos, SeekPosition.SeekCurrent).mapErr(ioErrorMsg)
+
+  let
+    stateIdx = ? f[].readIndex()
+  if stateIdx.offsets.len() != 1:
```
zah, line 68: If the indices are required to have certain size, why do you bother with the more complicated dynamic representation in the file (i.e. reading a `count` value, calculating offsets, etc)? 

Also, the idea of having a reverse length-prefix (one written after the data) is perhaps more fragile in the face of possible incomplete writes to the file. With forward length prefixes you can more easily ignore incomplete records because they will be inconsistent with regard to the length of the file.

Storing the index in a separate file looks more straight-forward in general.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -755,6 +796,71 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
+    for i in 0'u64..<historical_roots.lenu64():
+      var
+        found = false
+        done = false
+
+      for summary in dag.era.getBlockIds(historical_roots, Era(i)):
+        if summary.slot >= dag.backfill.slot:
+          # If we end up in here, we failed the root comparison just below in
+          # an earlier iteration
+          fatal "Era summaries don't lead up to backfill, database or era files corrupt?",
```
zah, line 880: What is the trust model around Era files? If I'm going to download them from the internet, shouldn't I assume that the data inside them is bogus until proven otherwise?

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: I couldn't get it to work reliably with type classes (`Forky` and friends) and instead kept running into compiler bugs, ergo this was the easiest way to make it work at all, even with those downsides

beacon_chain/era_db.nim
arnetheduck: https://github.com/google/snappy/blob/8dd58a519f79f0742d4c68fbccb2aed2ddb651e8/framing_format.txt#L34

beacon_chain/era_db.nim
arnetheduck: right, this was an old name: 68c29faaeb875602a6a652edbdd3752af63ac75a

beacon_chain/era_db.nim
arnetheduck: 68c29faaeb875602a6a652edbdd3752af63ac75a

beacon_chain/era_db.nim
arnetheduck: > If the indices are required to have certain size, why do you bother with the more complicated dynamic representation in the file (i.e. reading a count value, calculating offsets, etc)?

The length is dependent on chain parameters, so it seemed somewhat easier to keep it dynamic: it allows traversing the file without having access to fork information (which determines which constant to use, in case of future changes).

> perhaps more fragile

the idea is to write the index only after the rest of the file is complete - basically, this allows the file to remain append-only while creating it - the index is kind of optional / can be recreated by simply iterating over the records as well, meaning corruption is trivially detectable - having the index in place makes for some very easy reading and given that reading really is the primary use case it seemed reasonable to include the index in the file. A separate file is certainly possible (that was the first design), but it means the index would have to be recreated every time (or live with a more complex download mechanism).

The reverse direction is used so that when you append several era files together with a simple `cat`, the index stays valid - see `group` definition here: https://github.com/status-im/nimbus-eth2/blob/unstable/docs/e2store.md#structure

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: In this PR, we're reading era files from the data directory of nimbus, so the trust model is similar:ish to the sqlite database: only "valid" stuff should go in there - or rather: era files come into play _after_ establishing a head root / state (and therefore history), so from that point of view, they're less trusted than the sqlite database file - we assume they're valid (as in .. not corrupt), but we don't make any significant decisions based on them (like what to vote for).

In general, the direction that the codebase is going in is to make fewer assumptions about data availability in general and degrade gracefully when data is missing or corrupt - for example, when a replay fails, we used to panic, now we deal with it gracefully in most cases.

One risk is that we start sending out invalid blocks to neighbours (future patches will use `getBlockSZ` to respond to block range requests which means we won't even check if the contents decompress correctly, must less if they're valid), but as far as nimbus itself goes, we should simply treat invalid data in them as if we didn't have the data at all.

zah: But would there be something like an import procedure for an Era file downloaded from the internet where we try to establish that the data inside the file agrees with a recent head obtained through light client sync for example (which is an example of a more reliable source of authentic information)?

arnetheduck: well, there are two ways we can go about it:

* we can be strict and verify everything up-front - importing an era file this way can be done at a pace of maybe a thousand blocks / sec, so a few seconds per era file, so maybe an hour per year, give or take - this is `hash_tree_root`, one sig verification and snappy decompression work per block, and one state to verify per era.

* we can be lazy and verify era files as they actually get used - this is the approach taken in this branch where we *know* there should be data at a particular spot (because we established the head in another way) in the file and we know what that data should be (from the historical roots), so we try to load it on the fly - I'm leaning towards this approach for historical data because it degrades "gracefully" just fine

I can also see us doing a mix: with light client sync we need a state, and for _that_ state in particular we would do a a full verification - the reindexing done in https://github.com/status-im/nimbus-eth2/pull/3452 can be seen as another form of verification - it's important to note though that all this is not really related to security any more: it's more of a data integrity check.

beacon_chain/consensus_object_pools/block_pools_types.nim
```diff
@@ -294,6 +300,13 @@ type
 
 template head*(dag: ChainDAGRef): BlockRef = dag.headState.blck
 
+template frontfill*(dag: ChainDAGRef): BlockId =
```
zah: Multiple evaluations of the `dag` parameter.

beacon_chain/consensus_object_pools/blockchain_dag.nim
zah: > it degrades "gracefully" just fine

Perhaps I'm misunderstanding the code, but this doesn't look like a graceful degradation. It's a fatal error right now?

beacon_chain/era_db.nim
zah: My point was that a corrupted file is not trivially detectable in the current code. It will happily try to read the last bytes of the file and will proceed to act on them if the index was not fully written. Not a blocker for this PR, but it's a discussion worth having before we ship this and commit to maintaining backwards compatibility for the format.

arnetheduck:
> Perhaps, I'm misunderstanding the code, but this doesn't look like a graceful degradation. It's a fatal error right now?

yeah, this branch is a step on the path - up to this week, chaindag did not have the error reporting infrastructure to gracefully handle errors, but it's getting there, piece by piece - the code changes are quite extensive, in the sense that there were many places where we assumed data was valid and available - that said, most of those have been fixed now but doing so in this branch would have diluted it with refactoring noise.

arnetheduck:
> My point was that a corrupted file is not trivially detectable in the current code. It will happily try to read the last bytes of the file and will proceed to act on them if the index was not fully written.

sort of - there are the usual sanity checks that the data makes sense - offsets point to the right places within the file, that the state decompresses as snappy, etc - I'd say these are sufficient to rule out erroneous files specially because they're written last - if a partial write happens, these will not make sense, the snappy crc32 will fail and we'll discard the data. 

In this branch, what can really fail is when deliberately invalid data is supplied: the reason the error above is fatal is that in order for that situation in particular to happen, the file must have been deliberately constructed to fail - the name was wrong, the state was invalid, etc - it's not a random error at that point, but rather something explicitly fishy going on.

When I talk about graceful degradation, I mean for example that if we load a block from an era file and we can't apply it to a state, we now have error handling in place that simple aborts that one operation instead of crashing the application. Where exactly to draw that line is a matter for debate.

beacon_chain/consensus_object_pools/block_pools_types.nim
arnetheduck: fixed

PR #3401, opened 2022-02-16
make `SyncQueue` / `SyncManager` generic

etan-status:
The libp2p light client sync protocol defines an endpoint for syncing
historic data that is structured similarly to `beaconBlocksByRange`,
i.e., it uses a start/count/step tuple to sync from finalized to head.
As preparation, this patch extends the `SyncQueue` and `SyncManager`
implementations to work with such new `ByRange` endpoints as well.

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -42,24 +42,24 @@ type
     future: Future[void]
     status: SyncWorkerStatus
 
-  SyncManager*[A, B] = ref object
+  SyncManager*[A, B; E: SyncEndpoint] = ref object
     pool: PeerPool[A, B]
     responseTimeout: chronos.Duration
     maxHeadAge: uint64
     toleranceValue: uint64
-    getLocalHeadSlot: GetSlotCallback
-    getLocalWallSlot: GetSlotCallback
-    getSafeSlot: GetSlotCallback
-    getFirstSlot: GetSlotCallback
-    getLastSlot: GetSlotCallback
-    progressPivot: Slot
+    getLocalHeadKey: GetSyncKeyCallback[E.K]
+    getLocalWallKey: GetSyncKeyCallback[E.K]
+    getSafeKey: GetSyncKeyCallback[E.K]
+    getFirstKey: GetSyncKeyCallback[E.K]
+    getLastKey: GetSyncKeyCallback[E.K]
+    progressPivot*: E.K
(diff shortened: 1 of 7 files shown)


etan-status:
- `T`: Peer type
- `E`: Endpoint
- `K`: Key
- `V`: Value
- `R`: Result (network)

beacon_chain/sync/sync_manager.nim
```diff
@@ -618,17 +695,21 @@ proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
         of SyncQueueKind.Forward:
           if man.inProgress:
             man.inProgress = false
-            debug "Forward synchronization process finished, sleeping",
-                  wall_head_slot = wallSlot, local_head_slot = headSlot,
-                  difference = (wallSlot - headSlot),
-                  max_head_age = man.maxHeadAge, direction = man.direction,
-                  topics = "syncman"
+            when E.K is Slot:
```
tersec, line 698: This kind of logic conflates what's kind of an implementation detail (the unit of measure of synchronization) with the type of synchronization (forward sync, not light client sync). It happens to work now -- is there reason to be concerned that this bijectivity might not always hold?

etan-status: The unit is mentioned in the log message ("wall_head_slot" / "local_head_slot"), the `when` here is just to support other units in this wrapped log statement as well.

beacon_chain/sync/sync_queue.nim
```diff
@@ -24,88 +24,132 @@ logScope:
   topics = "syncqueue"
 
 type
-  GetSlotCallback* = proc(): Slot {.gcsafe, raises: [Defect].}
-  ProcessingCallback* = proc() {.gcsafe, raises: [Defect].}
-  BlockVerifier* =
-    proc(signedBlock: ForkedSignedBeaconBlock):
-      Future[Result[void, BlockError]] {.gcsafe, raises: [Defect].}
+  SyncEndpoint*[K, V, R] = (K, V, R)
```
etan-status: Add comment with link to Nim issue https://github.com/nim-lang/Nim/issues/19531

cheatfate:
Personally i do not like all this renames, i think this PR should me much less intrusive.

Lets not mix renames and functionality changes in one PR, because with a lot of renames its very hard to see what was actually changed in terms of functionality.

etan-status:
This PR is no longer needed, because all needed light client data can be obtained in just 1-2 requests.

PR #3403, opened 2022-02-16
rename no{R=>r}eturn, no{I=>i}init, short{l=>L}og, E{T=>t}h2Node, Beacon{c=>C}hainDB

tersec:
Start preparing `nimbus-eth2` for `--styleCheck:usage --styleCheck:error`.

These are either where the language itself assigns a canonically correct version (`noinit` and `noreturn`) or where the issues are effectively typos, not equivalently good but by chance inconsistent capitalization patterns (one could argue for `ETH2Node` or `Eth2Node`, but not as readily for `ETh2Node`, for example).

https://nim-lang.org/docs/manual.html#pragmas-noreturn-pragma has a section header of "noReturn" but states the pragma name is `noreturn`. `styleCheck:usage` prefers the latter.

https://nim-lang.org/docs/manual.html#statements-and-expressions-var-statement shows the `noinit` version, and indeed: `Hint: 'noInit' should be: 'noinit' [Name]`

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -569,7 +569,7 @@ proc close*(db: BeaconChainDBV0) =
   discard db.stateStore.close()
   discard db.backend.close()
 
-proc close*(db: BeaconchainDB) =
+proc close*(db: BeaconChainDB) =
   if db.db == nil: return
 
   # Close things in reverse order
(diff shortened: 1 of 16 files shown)


tersec:
https://github.com/nim-lang/Nim/pull/19535 makes Nim's documentation consistent here.

PR #3406, opened 2022-02-17
Revert "bump submodules (#3366)"

arnetheduck:
This reverts commit 6e1ad080e87eb0793641be0a154449d0022b18f9.

The bumps cause a regression in network connectivity and needs further investigation to identify the specific part that breaks (pre-bump/bump/revert in graph - other metrics show similar regressions):

Screenshot from 2022-02-17 10-32-57

https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=metal-01.he-eu-hel1.nimbus.mainnet&var-container=beacon-node-mainnet-unstable-02&from=1644928308775&to=1645090406073

The 1.6 compiler also segfaults on code that works with 1.2: https://github.com/status-im/nimbus-eth2/runs/5221934580?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,48 +2,81 @@ name: CI
 on:
   push:
     paths-ignore: ['media/**', 'docs/**', '**/*.md']
-    branches:
-      - stable
-      - testing
-      - unstable
   pull_request:
     paths-ignore: ['media/**', 'docs/**', '**/*.md']
-  workflow_dispatch:
 
 jobs:
   build:
     strategy:
       fail-fast: false
+      max-parallel: 20
       matrix:
         target:
+          # Unit tests
           - os: linux
             cpu: amd64
+            TEST_KIND: unit-tests
           - os: linux
             cpu: i386
+            TEST_KIND: unit-tests
           - os: macos
             cpu: amd64
+            TEST_KIND: unit-tests
           - os: windows
             cpu: amd64
-        branch: [version-1-2, version-1-6]
+            TEST_KIND: unit-tests
+          # Devel cache corrupted for mingw? missing propidl.h on PR but not push CI
+          # - os: windows
+          #   cpu: i386
+          #   TEST_KIND: unit-tests
+
+          # # Minimal integration tests
+          # - os: linux
+          #   cpu: amd64
+          #   TEST_KIND: finalization-minimal
+          # - os: linux
+          #   cpu: i386
+          #   TEST_KIND: finalization-minimal
+          # # Requires GNU getopt
+          # - os: macos
+          #   cpu: amd64
+          #   TEST_KIND: finalization-minimal
+          # - os: windows
+          #   cpu: amd64
+          #   TEST_KIND: finalization-minimal
+          # # TODO - Bootstrap issue: https://github.com/status-im/nimbus-eth2/issues/1725
+          # # - os: windows
+          # #   cpu: i386
+          # #   TEST_KIND: finalization-minimal
+
+          # Mainnet integration tests
+          # - os: linux
+          #   cpu: amd64
+          #   TEST_KIND: finalization-mainnet
+          # # - os: linux
+          # #   cpu: i386
+          # #   TEST_KIND: finalization-mainnet
+          # # - os: macos
+          # #   cpu: amd64
+          # #   TEST_KIND: finalization-mainnet
+          # - os: windows
+          #   cpu: amd64
+          #   TEST_KIND: finalization-mainnet
+          # # TODO - Bootstrap issue: https://github.com/status-im/nimbus-eth2/issues/1725
+          # # - os: windows
+          # #   cpu: i386
+          # #   TEST_KIND: finalization-mainnet
         include:
           - target:
               os: linux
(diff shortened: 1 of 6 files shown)


tersec:
Most of those bumps are probably harmless/innocuous. It's typically the same few (`nim-libp2p`, for example) which carry such risks. It'd be nice to get some (just at a glance, those akin to `nim-zlib`, `NimYAML`, `nim-confutils`, `nim-libbacktrace`, `nim-rocksdb`, `nim-stew`, `nim-stint`, etc, which are typically lower-risk), in, rather than the whole thing flapping back and forth with literally most of the submodules Nimbus uses at risk of another revert.

tersec (approved):

arnetheduck:
>  It'd be nice to get some 

Indeed, this was the suggested approach the first time around when the wholesale bump PR was created: merge the low-risk bumps first so as to take them out of the change set, then do the more complex upgrades separately  - this regression is not unexpected, given what was known before the bump - might as well start over with that process and do it properly, since there are multiple breakages here: this would follow our usual tried and true process to not mix too many unrelated changes in a PR - many if them are indeed trivial but investigating which is which is part of the bumping process and owned by whoever makes the bumps.

stefantalpalaru:
Isn't this because we stopped subscribing to unnecessary topics? The "fanout health" metrics are computed in `getLowSubnets()` here: https://github.com/status-im/nimbus-eth2/blob/f64fc6dc3aede8dc1af13274720998afeee02730/beacon_chain/networking/eth2_network.nim#L1043-L1049

But that mainnet node doesn't have any validators attached, so why would its subnets have more than 6 peers from "topic subscription"?

If you look at this other mainnet node - also without any validators attached - you'll see that its fanout health is all over the place: https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=metal-02.he-eu-hel1.nimbus.mainnet&var-container=beacon-node-mainnet-unstable-01&from=now-7d&to=now&refresh=5m&viewPanel=106

This Prater node, with 2500 validators attached, shows no problems: https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=metal-01.he-eu-hel1.nimbus.prater&var-container=beacon-node-prater-unstable-01&from=now-7d&to=now&refresh=5m&viewPanel=106

So I think the regression is in how we still compute those metrics for beacon nodes with no validators attached.

arnetheduck:
> , so why would its subnets have more than 6 peers from "topic subscription"?

Most mainnet nodes run with `subscribe-all-subnets` to stress them and provide service to the network in general - they will subscribe to as many topics as the fanout / mesh health etc shows (notably, this did not change between one version and another) - the metrics are accurate.

>  If you look at this other mainnet node

This node is set up without incoming connections to measure the outgoing peer cycling efficiency - it's part of the testing regime  that we check when performing any kind of networking changes, bumps included.

You can tell this by looking at the composition of connections: https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=metal-02.he-eu-hel1.nimbus.mainnet&var-container=beacon-node-mainnet-unstable-01&from=now-7d&to=now&refresh=5m&viewPanel=16

> This Prater node, with 2500 validators attached

You can see a regression there also: it takes longer to establish a healthy mesh (more red just after start) - equally concerning, and another sign of networking issues is the aggregation / batching performance that drops significantly:

https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=metal-01.he-eu-hel1.nimbus.prater&var-container=beacon-node-prater-unstable-01&from=now-7d&to=now&refresh=5m&viewPanel=86

You can see this effect on the other nodes as well.

Prater has a different composition of nodes: many are run by client teams and the relative distribution of client types and incoming slot availability is different - therefore, mesh / fanout performance will be different as well.

arnetheduck:
This is interesting: we disconnected mainly other nimbuses during the bad bump period:

Screenshot from 2022-02-17 20-29-37

arnetheduck:
cc @Menduist ^

Menduist:
#3366 didn't bump libp2p but reverted it to an older version
Before 3366, nimbus was using libp2p unstable for some reason I can't remember

I'll merge unstable to master and open a proper bump

stefantalpalaru:
And that's why submodules should never point at a different branch from the one in ".gitmodules".

arnetheduck:
> And that's why submodules should never point at a different branch from the one in ".gitmodules".

regardless, submodule bumps are not be done without due diligence to investigate what changes, and make sure that those changes are desired in context, just like any other code - the author of a submodule does not necessarily have visibility into the specific needs of `nimbus-eth2` and the role of the bump is to get on top of these changes _before_ they hit the codebase and affect other users and developers. This is one additional safety net that allows us to be active instead of reactive - it's also based on the general understanding that the author of the changes in the submodule is the most competent person to make a bump here, when possible - "hoping" that someone with the right competence will be around at the right time to investigate issues like the above connectivity issues is fragile, as far as approaches go, specially when there's no extraordinary reason to not follow the ordinary process.

PR #3410, opened 2022-02-17
converter function for reducing blocks to headers

etan-status:
This introduces a function to convert `SignedBeaconBlock` to just their
`BeaconBlockHeader` and updates the usages for reduced code duplication.


arnetheduck (approved):

etan-status:
There are two more cases that are similar but use a different state root. I've kept them as is for now (as it's only two). If there are more of these over time, maybe an overload of the new function could be worthwhile.

PR #3411, opened 2022-02-17
ncli_db: more readonly support

arnetheduck:
Update several `ncli_db` commands to run in readOnly mode, allowing them
to be used with a running instance - in particular era export.

* export all eras by default
* skip already-exported eras

### DIFF
--- a/ncli/ncli_db.nim
+++ b/ncli/ncli_db.nim
@@ -132,9 +132,9 @@ type
         defaultValue: 0
         desc: "The era number to write".}: uint64
       eraCount* {.
-        defaultValue: 1
+        defaultValue: 0
         name: "count"
-        desc: "Number of eras to write".}: uint64
+        desc: "Number of eras to write (0=all)".}: uint64
 
     of DbCmd.importEra:
       eraFiles* {.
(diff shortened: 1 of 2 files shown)


tersec (approved):

ncli/ncli_db.nim
```diff
@@ -457,24 +458,29 @@ proc cmdExportEra(conf: DbConf, cfg: RuntimeConfig) =
...
+      let e2 = openFile(name, {OpenFlags.Write, OpenFlags.Create}).get()
+      defer: discard closeFile(e2)
+
+      var group = EraGroup.init(e2, firstSlot).get()
+      if firstSlot.isSome():
+        withTimer(timers[tBlocks]):
+          var blocks: array[SLOTS_PER_HISTORICAL_ROOT.int, BlockId]
+          for i in dag.getBlockRange(firstSlot.get(), 1, blocks)..<blocks.len:
+            if dag.getBlockSSZ(blocks[i], tmp):
+              group.update(e2, blocks[i].slot, tmp).get()
```
zah, line 475: What do we `get` here? Is this supposed to work similar to `expect()`?

arnetheduck: yes, they're synonyms modulo the error message - expect would be an improvement

Issue #3076, opened 2021-11-09
Jenkins CI macOS linking warnings


tersec:
These go back months, at least to https://github.com/status-im/nimbus-eth2/commit/2eb6246b90a3f90452a79f47e2a5881aaa6c6230 from 2021-08-10, and likely before; Jenkins records just disappear shortly before that.

![2021-11-09-183800_880x948_scrot](https://user-images.githubusercontent.com/11422416/140975765-62d32c87-7b9b-47b6-a728-31cbae8d6784.png)
shows how to locate them -- in Jenkins CI, in the Tools section (but elsewhere too), for macOS, they show up.

For the example commit:
```
\e[92mBuilding:\e[39m build/deposit_contract
\e[92mBuilding:\e[39m build/nimbus_beacon_node
\e[92mBuilding:\e[39m build/resttest
\e[92mBuilding:\e[39m build/logtrace
\e[92mBuilding:\e[39m build/nbench
\e[92mBuilding:\e[39m build/nbench_spec_scenarios
\e[92mBuilding:\e[39m build/ncli
\e[92mBuilding:\e[39m build/ncli_db
\e[92mBuilding:\e[39m build/process_dashboard
\e[92mBuilding:\e[39m build/stack_sizes
\e[92mBuilding:\e[39m build/nimbus_validator_client
\e[92mBuilding:\e[39m build/nimbus_signing_process
\e[92mBuild completed successfully:\e[39m build/process_dashboard
\e[92mBuilding:\e[39m build/libnfuzz.so
\e[92mBuild completed successfully:\e[39m build/nbench_spec_scenarios
\e[92mBuilding:\e[39m build/libnfuzz.a
\e[92mBuild completed successfully:\e[39m build/stack_sizes
/Users/jenkins/workspace/nimbus-eth2_multibranch_unstable/beacon_chain/consensus_object_pools/attestation_pool.nim(23, 7) Warning: imported and not used: 'extras' [UnusedImport]
/Users/jenkins/workspace/nimbus-eth2_multibranch_unstable/beacon_chain/consensus_object_pools/attestation_pool.nim(23, 7) Warning: imported and not used: 'extras' [UnusedImport]
\e[92mBuild completed successfully:\e[39m build/logtrace
/Users/jenkins/workspace/nimbus-eth2_multibranch_unstable/beacon_chain/consensus_object_pools/attestation_pool.nim(23, 7) Warning: imported and not used: 'extras' [UnusedImport]
[...]

PR #3412, opened 2022-02-17
Jenkins: get rid of macOS warnings

stefantalpalaru:


stefantalpalaru:
Fixes https://github.com/status-im/nimbus-eth2/issues/3076

Jenkinsfile
```diff
@@ -21,14 +21,15 @@ def runStages(nodeDir) {
 			}
 
 			cache(maxCacheSize: 250, caches: [
-				[$class: "ArbitraryFileCache", excludes: "", includes: "**/*", path: "${WORKSPACE}/${nodeDir}/vendor/nimbus-build-system/vendor/Nim/bin"],
```
jakubgs, line 24: Do we want to build the Nim compiler every time?

Makefile
```diff
@@ -31,9 +31,17 @@ CPU_LIMIT := 0
...
+  CPU_LIMIT_CMD :=
 else
-	CPU_LIMIT_CMD := cpulimit --limit=$(CPU_LIMIT) --foreground --
+  CPU_LIMIT_CMD := cpulimit --limit=$(CPU_LIMIT) --foreground --
+endif
+
+# TODO: move this to nimbus-build-system
+ifeq ($(shell uname), Darwin)
+  # Scary warnings in large volume: https://github.com/status-im/nimbus-eth2/issues/3076
+  SILENCE_WARNINGS := 2>&1 | grep -v "failed to insert symbol" | grep -v "could not find object file symbol for symbol" || true
```
jakubgs, line 42: You can avoid invoking two instances of `grep` by using `-e` twice for the same call.

jakubgs:
It would be good to explain in the commit(s) or the PR description why these "scary" warnings aren't actually scary and can be ignored.

Jenkinsfile
stefantalpalaru: Yes. It's 1-2 minutes of run time, in exchange for a cleaner setup that doesn't have to deal with old libraries and object files from previous runs.

Makefile
stefantalpalaru: Nice one, but not very readable. How many people know what `-e` does without reading the man page?

stefantalpalaru:
We're linking object files with debug info with ones without it. Perfectly justified, from our point of view, but `dsymutil` complains verbosely about it, with no switch to silence it.

Jenkinsfile
jakubgs: That's no that bad.

Makefile
jakubgs: Really? Your worry is if someone knows `-e` in this spaghetti of `Makefile`s?

jakubgs:
Okay, then lets include that explanation in the description.

stefantalpalaru:
Included in the commit desc: https://github.com/status-im/nimbus-eth2/commit/555ce310cc85aba794951a740665b1d0685abea1

Makefile
stefantalpalaru: :-)

PR #3413, opened 2022-02-18
db: make block loading generic

arnetheduck:
Streamline lookup with Forky and BeaconBlockFork (then we can do the
same for era)

We use type to avoid conditionals, as fork is often already known at a
"higher" level.

* load blockid before loading block by root - this is needed to map root
to slot and will eventually be done via block summary table for "old"
blocks

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -752,7 +752,7 @@ proc getBlockSSZ*(
   proc decode(data: openArray[byte]) =
     try: dataPtr[] = snappy.decode(data, maxDecompressedDbRecordSize)
     except CatchableError: success = false
-  db.blocks[BeaconBlockFork.PHase0].get(key.data, decode).expectDb() and success or
+  db.blocks[BeaconBlockFork.Phase0].get(key.data, decode).expectDb() and success or
     db.v0.getPhase0BlockSSZ(key, data)
 
 proc getBlockSSZ*[
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -104,13 +104,12 @@ type
     checkpoint*: proc() {.gcsafe, raises: [Defect].}
 
     keyValues: KvStoreRef # Random stuff using DbKeyKind - suitable for small values mainly!
-    blocks: KvStoreRef # BlockRoot -> phase0.TrustedBeaconBlock
-    altairBlocks: KvStoreRef # BlockRoot -> altair.TrustedBeaconBlock
-    mergeBlocks: KvStoreRef # BlockRoot -> bellatrix.TrustedBeaconBlock
+    blocks: array[BeaconBlockFork, KvStoreRef] # BlockRoot -> TrustedSignedBeaconBlock
+
     stateRoots: KvStoreRef # (Slot, BlockRoot) -> StateRoot
-    statesNoVal: KvStoreRef # StateRoot -> Phase0BeaconStateNoImmutableValidators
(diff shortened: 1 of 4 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -418,13 +417,18 @@ proc new*(T: type BeaconChainDB,
 
     # V1 - expected-to-be small rows get without rowid optimizations
     keyValues = kvStore db.openKvStore("key_values", true).expectDb()
-    blocks = kvStore db.openKvStore("blocks").expectDb()
-    altairBlocks = kvStore db.openKvStore("altair_blocks").expectDb()
-    mergeBlocks = kvStore db.openKvStore("merge_blocks").expectDb()
+    blocks = [
+      kvStore db.openKvStore("blocks").expectDb(),
+      kvStore db.openKvStore("altair_blocks").expectDb(),
+      kvStore db.openKvStore("merge_blocks").expectDb()]
```
arnetheduck, line 423: cc @tersec should we take the opportunity to rename the table as well? there's good time to do it, but there are certainly worse times

tersec: Yeah, should happen sometime. The backfill and checkpoint sync changes should rendered it more resilient to potentially missing phase0 blocks, too.

beacon_chain/beacon_chain_db.nim
```diff
@@ -738,31 +711,25 @@ proc getPhase0Block(db: BeaconChainDBV0, key: Eth2Digest): Opt[phase0.TrustedSig
...
-    Opt[phase0.TrustedSignedBeaconBlock] =
+proc getBlock*(
+    db: BeaconChainDB, key: Eth2Digest,
+    T: type phase0.TrustedSignedBeaconBlock): Opt[T] =
   # We only store blocks that we trust in the database
-  result.ok(default(phase0.TrustedSignedBeaconBlock))
-  if db.blocks.getSnappySSZ(key.data, result.get) != GetResult.found:
+  result.ok(default(T))
+  if db.blocks[T.toFork].getSnappySSZ(key.data, result.get) != GetResult.found:
+    # During the initial releases phase0, we stored blocks in a different table
```
tersec, line 720: This extends now-obsolete fallback logic to Altair and Bellatrix?

beacon_chain/beacon_chain_db.nim
```diff
@@ -776,31 +743,39 @@ proc getPhase0BlockSSZ(db: BeaconChainDBV0, key: Eth2Digest, data: var seq[byte]
...
+proc getBlockSSZ*(
+    db: BeaconChainDB, key: Eth2Digest, data: var seq[byte],
+    T: type phase0.TrustedSignedBeaconBlock): bool =
   let dataPtr = unsafeAddr data # Short-lived
   var success = true
   proc decode(data: openArray[byte]) =
     try: dataPtr[] = snappy.decode(data, maxDecompressedDbRecordSize)
     except CatchableError: success = false
-  db.blocks.get(key.data, decode).expectDb() and success or
+  db.blocks[BeaconBlockFork.PHase0].get(key.data, decode).expectDb() and success or
```
```suggestion
  db.blocks[BeaconBlockFork.Phase0].get(key.data, decode).expectDb() and success or
```

beacon_chain/beacon_chain_db.nim
arnetheduck: no, you can see how T limits this overload to phase0 - newer versions below. The logic is not obsolete: if a user has been running since 1.0, their database will contain phase0 blocks in the old store.

tersec (approved):

beacon_chain/beacon_chain_db.nim
arnetheduck: ok, when?

PR #3415, opened 2022-02-18
Support for Gnosis Chain

zah:
`make gnosis-chain-build` will build the Nimbus gnosis chain binary,
stored in `build/nimbus_beacon_node_for_gnosis_chain`.

`make gnosis-chain` will connect to the network.

Other changes:

* Restore compilation with `-d:has_genesis_detection`
* Removed Makefile target related to testnet0 and testnet1
* Added more debug logging for failed peer handshakes


media/gnosis-chain/config.yaml
```diff
@@ -0,0 +1,157 @@
+CONFIG_NAME: gnosis
```
arnetheduck: doesn't gnosis have a repo with genesis.ssz / etc? they had one for lighthouse I'm pretty sure..

media/gnosis-chain/boot_enr.yaml
```diff
@@ -0,0 +1,4 @@
+- enr:-IS4QGmLwm7gFd0L0CEisllrb1op3v-wAGSc7_pwSMGgN3bOS9Fz7m1dWbwuuPHKqeETz9MbhjVuoWk0ohkyRv98kVoBgmlkgnY0gmlwhGjtlgaJc2VjcDI1NmsxoQLMdh0It9fJbuiLydZ9fpF6MRzgNle0vODaDiMqhbC7WIN1ZHCCIyg
```
arnetheduck, line 1: this is generally in text format, in all other chain definitions - the yaml here is a lighthouse thing

arnetheduck:
re web3 changes, the right thing I think would be to make it runtime-detected / permissive,  which fields are there and not - it's quite likely this will change over time as the gnosis binaries get updated and json is _generally_ permissive - we've already messed this up in the rest client, being overly strict.

media/gnosis-chain/boot_enr.yaml
zah: We support both formats, so what's the practical difference?

media/gnosis-chain/config.yaml
zah: They have one repo per client and these are not quite clean because they contain a lot of additional files.
The metadata there also doesn't include either `genesis.ssz` or a `deposit_contract_block.txt` file, so I've decided to create this proper metadata folder here. Please note that I haven't included a genesis file, because the Gnosis network has started with just 4K validators and syncing the deposit history from scratch takes only few seconds.

zah:
> re web3 changes, the right thing I think would be to make it runtime-detected / permissive, which fields are there and not - it's quite likely this will change over time as the gnosis binaries get updated and json is generally permissive - we've already messed this up in the rest client, being overly strict.

This is not practical at the moment, because the web3 library is based on `std/json` which doesn't support optional fields.

media/gnosis-chain/boot_enr.yaml
arnetheduck: I wasn't aware tbh - good to know, been meaning to write a spec for that network config folder

media/gnosis-chain/config.yaml
arnetheduck: > They have one repo per client and these are not quite clean because they contain a lot of additional files.

shouldn't we have them maintain their chain in their repo just like for all other clients then? ie ideally, we should not have to keep chain-specific data in our repo but rather have tools for people to do this on their own - it's good that we develop infrastructure for it, but ..

arnetheduck: one more thing about `CONFIG_NAME`: how about using it for the generated binary? I've started using it for the `era` files to make them easily identifiable, and the binary here could simply be called `nimbus_beacon_node_$(CONFIG_NAME)` with exceptions for perfectly matching mainnet configs

arnetheduck:
> This is not practical at the moment, because the web3 library is based on std/json which doesn't support optional fields.

oof. shouldn't we be able  to work around that? ie `std/json` creates dynamic `JsonNode` instances - the mapping to object is doable in application, no? Automatic mapping frameworks inevitably have gaps like this and the key/basic feature in such frameworks is the escape hatch that allows a manual intervention..

media/gnosis-chain/boot_enr.yaml
zah: Well, you added the ENR support, so I was assuming you are aware :) The support is a bit broken btw - it can't handle "quoted" strings which are valid in YAML and were actually used in the original Gnosis repos.

media/gnosis-chain/config.yaml
zah: You've removed the support for reading preset files at compile-time some time ago, so I had to implement this build in a more manual fashion. I guess I can rename the binary now in anticipation of restoring the preset loading support in the future in a more general solution.

zah:
> Automatic mapping frameworks inevitably have gaps like this and the key/basic feature in such frameworks is the escape hatch that allows a manual intervention.

Well, it is possible, but the solution was considered more ugly. You would have to override the json loading of the large object holding the optional field (i.e. `BlockObject` and `BlockHeaderObject`), so you'll have start maintaining the correct field to type mappings for all other fields. The automated mapping has the benefit that it doesn't introduce a "maintenance trap". Solutions requiring a lot of manually written boilerplate code are frequently suffering from bugs where a developer changes one definition without also changing other definitions that are required to match.

Also, semantically, these are not optional fields in Ethereum. There are missing in the Gnosis network, because it's not a regular proof-of-work network, so a "proper" solution that would allow a client to speak with multiple networks at the same time would have to be more complex.

arnetheduck:
> a "maintenance trap"

well, these are standards, formal or less formal - the amount of maintenance tends to be quite limited, _specially_ if you write parsing in a lenient way that degrades gracefully, as json _generally _should: if the framework is unable to express what a standard mandates (or worse, it tries to be a "better" standard by imposing its own opinions), the workarounds and degradation in behaviour cause more maintenance, for everyone (users, helpdesk time, the programmer that has to defend the framework against rational arguments) - a tiny win in automation ends up having a significant cost elsewhere.

media/gnosis-chain/config.yaml
arnetheduck: right, well, `CONFIG_NAME` is a `config` variable, so it should be loaded at runtime - ergo, naming the binary after it is wrong - the closest we have is perhaps the name of the preset folder.

I recently added `CONFIG_NAME` to the spec, maybe we should propose a `PRESET_NAME` as well..

arnetheduck:
also, if "mainenance" is an objective, an undocumented and untested compile-time flag in nim-web3 hailing from a "user" of nim-web3 is difficult to defend: when maintaining nim-web3, it's hard to know that one has to look in nimbus-eth2 to understand where the flag is coming from.

zah:
My comment was specifically about the tradeoffs between the alternatives discussed in this PR. As I explained, I think the code will be less correct and less attractive if the serialization of `BlockObject` and `BlockHeader` is overridden to allow more lax parsing. A more proper implementation would be to introduce a separate type such as `GnosisChainBlockHeader` and then make it possible for the downstream users of the library to select the set of types they want to work with. This was deemed too much work at the moment and I took the short-cut in the PR in order to deliver a working binary for the Gnosis users sooner.

Your comments seem to be about your general distaste for automatic serialization. Since this argument has been repeated many times already, I see little value is bringing it in the discussion here. In any case, even after taking your arguments to heart, implementing serialization routines manually is not the right answer. Ideally, the serialization framework will provide sufficient options, so the user can just describe how their type is mapped to a particular standard by selecting one of the already implemented strategies (i.e. require all fields, allow unknown fields, etc). Something like this would be possible after nim-web3 is migrated to nim-json-serialization and after we add the "explicit mode" flavors there.

arnetheduck:
> A more proper implementation would be to introduce a separate type such as

I don't think so - a proper implementation would be "permissive in what it accepts" - this is the de-facto standard in JSON-RPC, to ignore missing fields unless you need them, and to allow extra fields - there's a simple reason: servers and clients on the internet get upgraded separately - orchestrating a simultaneous and complete upgrade is difficult - when you are permissive, meaning that if you don't use a particular set of fields, you allow your client to continue working when the server gets upgraded - this is a better user experience than having it break, then forcing an upgrade.

In this particular case, we clearly don't need / use these fields - for example, we don't validate their contents or ensure their integrity, meaning that whether they are there or not is *completely irrelevant* to our use of the service - the only difference they make is that sometimes, they break things without any benefit at all. This is a lose-lose situation: we get more maintenance and ugly compile-time spaghetti and users get a degraded experience.

FWIW, this isn't necessarily a blocker for this PR, but more an observation about how we're systemically approaching the problem with the wrong mindset - this is also in part why automatic deserializers, in order to be viable for use on the internet, must have certain features in order to be at all viable: the must have the option to be permissive, and they must absolutely allow an escape hatch for a manual override, because inevitably, they will be inadequate in some situations due to their nature: they try to force a structure and logic upon a problem, where often, none exists.

zah:
Ignoring missing fields is not always possible. What do you do with a block that has a missing `parentHash`? This is just not a valid block, so it better be reported as an error. My point was that the "validity" of blocks is different in the two networks because the networks use different consensus mechanism - the Gnosis chain is not a proof-of-work chain, so you cannot just impose its peculiarities on the way Ethereum mainnet blocks are validated. nim-web3 is a general purpose library after all, it shouldn't be affected too much by our own limited point of view.

arnetheduck:
> nim-web3 is a general purpose library after all, it shouldn't be affected too much by our own limited point of view.

well, this is my main point that you're de-facto agreeing with: nim-web3 should not make overly restrictive decisions about validity, and neither should the deserializer, beyond trivialities, because they depend on the context in which it is used - nim-web3 can't decide if a parentHash is valid and meaningful or not, this is something that the consumer does, optionally, depending on the use case.

There are some trivial validation decisions sometimes that can be dealt with at the serializer level which simplifies downstream code, assuming they are truly universal, but even a field like `parentHash` is borderline meaningless to validate its presence because the serializer doesn't have (and shouldn't have) access to the context to distinguish an all-zeroes parenthash from some random output, from a true parent - it's outside of its scope of responsibilities to reason about these things.

zah:
I think a response where the `parentHash` value is zero is a significantly different from a response where the field is missing altogether and I would classify the later as an invalid response. The specification of the Ethereum JSON-RPC API clearly indicates some of the fields as `required`:

https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=true&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false

Furthermore, I don't think the users of the web3 library will appreciate the alternative where all fields are defined as `Option[T]`, requiring checks before use.

PR #3420, opened 2022-02-19
NoMonitor and removed clock check for SyncManager.

cheatfate:
* Add `NoMonitor` flag to SyncManager, which helps to start custom SyncManager's which are not supposed to be running all the time.
* Add `sync_ident` to logs to help identify log lines for specific SyncManager.
* Remove `clock` check from SyncManager flow.
* Optimize logs with logScope.

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -49,7 +49,6 @@ type
     pool: PeerPool[A, B]
     responseTimeout: chronos.Duration
     maxHeadAge: uint64
-    toleranceValue: uint64
     getLocalHeadSlot: GetSlotCallback
     getLocalWallSlot: GetSlotCallback
     getSafeSlot: GetSlotCallback
@@ -122,7 +121,6 @@ proc newSyncManager*[A, B](pool: PeerPool[A, B],
                            blockVerifier: BlockVerifier,
                            maxHeadAge = uint64(SLOTS_PER_EPOCH * 1),
                            chunkSize = uint64(SLOTS_PER_EPOCH),
-                           toleranceValue = uint64(1),
                            flags: set[SyncManagerFlag] = {},
                            ident = "main"
                            ): SyncManager[A, B] =
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -38,6 +38,9 @@ type
     Sleeping, WaitingPeer, UpdatingStatus, Requesting, Downloading, Queueing,
     Processing
 
+  SyncManagerFlag* {.pure.} = enum
+    NoMonitor
+
   SyncWorker*[A, B] = object
     future: Future[void]
     status: SyncWorkerStatus
@@ -66,6 +69,7 @@ type
     syncStatus*: string
     direction: SyncQueueKind
     ident*: string
+    flags: set[SyncManagerFlag]
 
   SyncMoment* = object
     stamp*: chronos.Moment
@@ -119,6 +123,7 @@ proc newSyncManager*[A, B](pool: PeerPool[A, B],
                            maxHeadAge = uint64(SLOTS_PER_EPOCH * 1),
                            chunkSize = uint64(SLOTS_PER_EPOCH),
                            toleranceValue = uint64(1),
+                           flags: set[SyncManagerFlag] = {},
                            ident = "main"
                            ): SyncManager[A, B] =
   let (getFirstSlot, getLastSlot, getSafeSlot) = case direction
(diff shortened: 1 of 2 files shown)


beacon_chain/sync/sync_manager.nim
```diff
@@ -137,19 +144,28 @@ proc newSyncManager*[A, B](pool: PeerPool[A, B],
...
+    flags: flags
   )
   res.initQueue()
   res
 
 proc getBlocks*[A, B](man: SyncManager[A, B], peer: A,
                       req: SyncRequest): Future[BeaconBlocksRes] {.async.} =
   mixin beaconBlocksByRange, getScore, `==`
+
+  logScope:
```
arnetheduck: the alternative to logscope is to make a `shortLog(req: Request): auto = ...` which returns a condensed view of a request

beacon_chain/sync/sync_queue.nim
```diff
@@ -70,11 +70,23 @@ type
...
   SyncManagerError* = object of CatchableError
   BeaconBlocksRes* = NetRes[seq[ref ForkedSignedBeaconBlock]]
 
-chronicles.formatIt SyncQueueKind: $it
+chronicles.formatIt SyncQueueKind: toLowerAscii($it)
+
+template shortLog*[T](req: SyncRequest[T]): string =
+  Base10.toString(uint64(req.slot)) & ":" &
+  Base10.toString(req.count) & ":" &
+  Base10.toString(req.step) & "@" &
```
arnetheduck: maybe don't print step if it's 1: it's not used anywhere, and will likely never be used, even if it's part of the standard (I don't think any other client uses it either)

beacon_chain/sync/sync_manager.nim
```diff
@@ -617,43 +605,37 @@ proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
...
           if man.inProgress:
-            man.inProgress = false
-            debug "Forward synchronization process finished, sleeping",
-                  wall_head_slot = wallSlot, local_head_slot = headSlot,
-                  difference = (wallSlot - headSlot),
-                  max_head_age = man.maxHeadAge, direction = man.direction,
-                  topics = "syncman"
+            if SyncManagerFlag.NoMonitor in man.flags:
+              await man.syncClose(guardTaskFut, averageSpeedTaskFut)
+              man.inProgress = false
```
arnetheduck, line 608: even in nomonitor mode, shouldn't the `inProgress` flag be kept correct?

etan-status:
- Rebase to `unstable`.
  - `sync_ident` and `logScope` changes have already been integrated as part of https://github.com/status-im/nimbus-eth2/pull/3451
  - This leaves the `NoMonitor` and removed `clock` check in this PR.
  - Removed mentions of already integrated changes from commit messages (and reduced to just initial commit as 2/3 have already been integrated)
- Remove `toleranceValue` parameter and `PeerScoreHeadTooNew` constant that are no longer used (due to removed `clock` check)

beacon_chain/sync/sync_manager.nim
```diff
@@ -256,17 +260,6 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
     headSlot = man.getLocalHeadSlot()
     wallSlot = man.getLocalWallSlot()
 
-    if peerSlot > wallSlot + man.toleranceValue:
```
arnetheduck, line 259: for future reference, this check is already performed here: https://github.com/status-im/nimbus-eth2/blob/79bfe1b4bb95e320b9dc0bdd5a888503de52447e/beacon_chain/sync/sync_protocol.nim#L141:L141

arnetheduck (approved):

beacon_chain/sync/sync_manager.nim
etan-status: wallSlot in the sync manager may not always reflect the real wallSlot, depending on the callback. For bellatrix, the way how I understood the problem is that an older wallSlot may be given (like, 128 slots behind), and for some reason the toleranceValue could not simply be set to 128, making it so that this check sometimes led to unexpected descoring.

But yeah, the check you link seems to reflect the intention of this check more cleanly (minus the tolerance value). sync_protocol uses the real wallSlot.

tersec: Yes, https://github.com/ethereum/consensus-specs/blob/v1.1.10/sync/optimistic.md#constants notes that the `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` must be user-configurable.

The main initial user of this is
https://github.com/status-im/nimbus-eth2/blob/b4787dbce46694880bbe0efd44a73fd9b1fc6841/beacon_chain/nimbus_beacon_node.nim#L314-L318

in the `kiln-dev-auth` branch.

PR #3426, opened 2022-02-21
bump nim-json-ser

arnetheduck:
fixes handling of null cstrings / cleanups


tersec:
No objections to merging it, but seems like for reasons not obviously related to any specific commit in that repo, Windows, and only Windows, CI builds are consistently failing with:
```
Run mkdir -p external
curl: (3) URL using bad/illegal format or missing URL
Error: Process completed with exit code 3.
```

arnetheduck:
I believe the nim-1.6 testing branch resolved that as well - that, and the more recent commits are all fully green cc @stefantalpalaru

stefantalpalaru:
> curl: (3) URL using bad/illegal format or missing URL

This comes from a new version of Curl that stopped accepting straight spaces in URLs. The URLs in question are the ones for Mingw-w64 archives and I solved that by moving to MSYS2.

The failure may not be consistent, because different CI hosts can have different Curl versions installed by default.

If there's no plan to resurrect my PR, the fix is to replace spaces in URLs with "%20".

PR #3432, opened 2022-02-23
allow sync using light client protocol

etan-status:
Adds the `--trust-sync-committee` launch option to use the light client
sync protocol for obtaining the latest finalized head. Then, instead of
regularly syncing forward, sync backwards to skip most block validation.
`--trusted-block-root` can be used to initialize the light client; when
not specified, regular sync is used until the light client obtains info
about new sync committees from the network to jump to latest wall slot.


etan-status:
This will be split up into multiple smaller PRs.

PR #2225, opened 2021-01-11
allow always-on subscription to all attestation subnets when gossiping

tersec:

PR #3439, opened 2022-02-25
Ignore seen aggregates

arnetheduck:
https://github.com/ethereum/consensus-specs/pull/2225 removed an ignore
rule that would filter out duplicate aggregates from gossip publishing -
however, this causes increased bandwidth and CPU usage as discussed in
https://github.com/ethereum/consensus-specs/issues/2183 - the intent is
to revert the removal and reinstate the rule.

This PR implements ignore filtering which cuts down on CPU usage (fewer
aggregates to validate) and bandwidth usage (less fanout of duplicates)
- as #2225 points out, this may lead to a small increase in IHAVE
messages.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -280,7 +280,7 @@ func updateAggregates(entry: var AttestationEntry) =
             inc j
         inc i
 
-proc covers(entry: AttestationEntry, bits: CommitteeValidatorsBits): bool =
+func covers(entry: AttestationEntry, bits: CommitteeValidatorsBits): bool =
   for i in 0..<entry.aggregates.len():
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -280,6 +280,12 @@ func updateAggregates(entry: var AttestationEntry) =
             inc j
         inc i
 
+proc covers(entry: AttestationEntry, bits: CommitteeValidatorsBits): bool =
```
tersec: `func` should work here

tersec (approved):

beacon_chain/consensus_object_pools/attestation_pool.nim
arnetheduck: bde87288d55a757dc336cbeda2c410df5c4a3914

PR #3442, opened 2022-02-28
Add support for TOML config files (--config-file)

zah:


beacon_chain/conf.nim
```diff
@@ -93,6 +111,11 @@ type
     # migrate = "Export and remove specified validators from Nimbus."
 
   BeaconNodeConf* = object
+    configFile* {.
+      desc: "Loads the configuration from a TOML file"
+      defaultValueDesc: defaultBeacoNodeConfigFilePath()
```
arnetheduck: let's not add defaults to the config file option, this creates a support problem where there's no explicit indication of it being present or not in a user configuration - if users want to use a config file, they should be explicitly listed so that they don't end up using one by accident (and therefor including some configuration that they didn't intend to include)

beacon_chain/nimbus_binary_common.nim
```diff
@@ -14,7 +14,7 @@ import
   std/[os, tables, strutils, terminal, typetraits],
 
   # Nimble packages
-  chronos, confutils,
+  chronos, confutils, toml_serialization,
```
arnetheduck, line 17: why is toml_serialization here? shouldn't it be in `conf.nim`?

beacon_chain/conf.nim
mratsim: We should ship a default config file instead so that users can edit it accordingly.

arnetheduck: The gold standard is to print a config file to stdout with defaults from the application (for example when passing `--config-file` without a value), but commented out: this way you don't need to worry about stale config files and multiple sources of truth for defaults. It's important that the values in this generated config file are indeed commented, so that when we change defaults, "most" users get the new values unless they made an explicit choice.

zah: The default location have been removed for now. There is plenty of software with default config file location though.

beacon_chain/nimbus_binary_common.nim
zah: It's used here as well. We refer to the `Toml` format type when we decide which config path to load. I guess I could have re-exported it from `conf.nim`, but `conf.nim` is a popular file, so it would have resulted in more pollution.

arnetheduck: I find it's better to reexport regardless - we've had enough trouble with imports, the reexport makes for shorter import lists overall which helps find circular imports - it also makes for fewer compiler errors and weird import-order problems which lead to formatting issues in log files etc

arnetheduck (approved):
I'll assume command-line params take priority over config files? docs should be updated too

PR #3447, opened 2022-02-28
use next slot when signing for light client tests

etan-status:
In practice, the sync committee signs `LightClientUpdate` instances at
the next slot following the block. This is not correctly reflected in
the tests, where it is signed one slot early. This patch updates the
tests to use the correct slot for the computation.


tersec (approved):

tersec:
As a workaround, this is fine -- but it seems like something that should be resolved, ideally, in the specs themselves.

etan-status:
See https://github.com/ethereum/consensus-specs/pull/2805

PR #3451, opened 2022-03-01
Refactor and optimize sync logs.

etan-status:
Integrates @cheatfate's refactorings and optimizations of the sync manager / sync queue logs as well as the introduction of `shortLog(SyncRequest)` from #3420 

On top of these integrations, adds a few minor logging improvements:
- Fixes a typo (`was happened` -> `has happened`)
- Avoids passing `reset_slot` argument to log statement multiple times
- Uses same `rewind_to_slot` label when logging in both sync directions
- Consistent rewind point logging

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -65,6 +65,7 @@ type
     avgSyncSpeed*: float
     syncStatus*: string
     direction: SyncQueueKind
+    ident*: string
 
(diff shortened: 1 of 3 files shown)


arnetheduck:
#3420 already does lots of logging cleanup

etan-status:
As per @arnetheduck's request, also integrated the logging cleanups from #3420 into this PR.

PR #3452, opened 2022-03-01
Historical state reindex for trusted node sync

arnetheduck:
When performing trusted node sync, historical access is limited to
states after the checkpoint.

Reindexing restores full historical access by replaying historical
blocks against the state and storing snapshots in the database.

The process can be initiated or resumed at any point in time.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1888,10 +1888,10 @@ proc rebuildIndex*(dag: ChainDAGRef) =
 
     if slot == dag.tail.slot:
       link(midRef, dag.tail)
-      dag.tail = midRef
       break
 
     let next = BlockRef.init(root, slot)
+
     link(midRef, next)
     midRef = next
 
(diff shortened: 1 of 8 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1770,3 +1781,142 @@ proc aggregateAll*(
...
+    link(midRef, next)
+    midRef = next
+
+    finBlocks[midRef.slot.int] = midRef
+
+  dag.finalizedBlocks = finBlocks
+  dag.tail = dag.genesis
+
+  if junk.len > 0:
+    info "Dropping redundant states", junk
```
zah: Wouldn't this print too much data?

arnetheduck: typically no (junk gets created during a crash only, really - but ..)

arnetheduck: f3df03d138171135cbec0d14068268c7e819f87e

PR #3459, opened 2022-03-03
validate `fork_version` as light client

etan-status:
The spec does not provide code for validating the `fork_version` field
of `LightClientUpdate`. However, we can use our own logic for additional
validation of that field. The spec's python test suite sets up states
that do not follow the fork schedule (e.g., that use Altair fork version
before Altair fork epoch), which complicates upstreaming this as code.

### DIFF
--- a/beacon_chain/spec/light_client_sync.nim
+++ b/beacon_chain/spec/light_client_sync.nim
@@ -12,6 +12,32 @@ import
   datatypes/altair,
   helpers
 
+func period_contains_fork_version(
+    cfg: RuntimeConfig,
+    period: SyncCommitteePeriod,
+    fork_version: Version): bool =
+  ## Determine whether a given `fork_version` is used during a given `period`.
+  let
+    periodStartEpoch = period.start_epoch
+    periodEndEpoch = periodStartEpoch + EPOCHS_PER_SYNC_COMMITTEE_PERIOD - 1
+  return
+    if fork_version == cfg.SHARDING_FORK_VERSION:
+      periodEndEpoch >= cfg.SHARDING_FORK_EPOCH
+    elif fork_version == cfg.BELLATRIX_FORK_VERSION:
+      periodStartEpoch < cfg.SHARDING_FORK_EPOCH and
+      cfg.SHARDING_FORK_EPOCH != cfg.BELLATRIX_FORK_EPOCH and
+      periodEndEpoch >= cfg.BELLATRIX_FORK_EPOCH
+    elif fork_version == cfg.ALTAIR_FORK_VERSION:
+      periodStartEpoch < cfg.BELLATRIX_FORK_EPOCH and
+      cfg.BELLATRIX_FORK_EPOCH != cfg.ALTAIR_FORK_EPOCH and
+      periodEndEpoch >= cfg.ALTAIR_FORK_EPOCH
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/light_client_sync.nim
```diff
@@ -12,6 +12,25 @@ import
...
+    cfg: RuntimeConfig,
+    period: SyncCommitteePeriod,
+    fork_version: Version): bool =
+  ## Determine whether a given `fork_version` is used during a given `period`.
+  let
+    periodStartEpoch = period.start_epoch
+    periodEndEpoch = periodStartEpoch + EPOCHS_PER_SYNC_COMMITTEE_PERIOD - 1
+  return
+    if fork_version == cfg.BELLATRIX_FORK_VERSION:
+      periodEndEpoch >= cfg.BELLATRIX_FORK_EPOCH
```
tersec, line 29: A `case` statement would ensure this doesn't miss, e.g., a sharding fork without a compilation error.

beacon_chain/spec/light_client_sync.nim
```diff
@@ -53,6 +73,11 @@ proc validate_light_client_update*(
   if update_period notin [finalized_period, finalized_period + 1]:
     return false
 
+  # Verify fork version is acceptable
+  template fork_version(): auto = update.fork_version
```
tersec: https://github.com/status-im/nimbus-eth2/blob/368d9566a01917afdb9b7f391adbd2b6706b24ed/beacon_chain/spec/presets.nim#L28
`Version` just isn't that big a structure. This `template` trick seems a bit overkill for 4 bytes.

beacon_chain/spec/light_client_sync.nim
etan-status: `fork_version` can be anything, though. It is data coming in from the network.
So, the `case` statement would need a default `else` case, and you would not have a compilation error.

Also, I think `case` does not work on byte arrays (or even non-cost values in general)?

beacon_chain/spec/light_client_sync.nim
etan-status: Replaced with `let`.

etan-status:
- Replaced `template fork_version()` with `let` binding.
- Split `else` case into two cases, one for `GENESIS_FORK_VERSION`, one for unviable forks.
- Future-proof for `SHARDING_FORK_VERSION`.

beacon_chain/spec/light_client_sync.nim
tersec: Oh, indeed.

While it still seems useful to organize the logic such that it checks, as separate cases, if (a) it doesn't match any known fork at all, and (b), if it does, then do a strict case-match on the known forks, the `RuntimeConfig` object doesn't list has these versions in enumerable form, so it's not straightforward to automatically construct one from the other without, e.g., a macro which matches field names of `RuntimeConfig` or some compile-time `fieldPairs` iteration with similar string-matching on `FOO_FORK_VERSION`: there are better ways to go about this.

So, in this code, it's not yet possible to do completely straightforward compile-time checks for exhaustiveness, and probably refactoring `RuntimeConfig` would be better regardless.

tersec (approved):

PR #3461, opened 2022-03-03
avoid re-requesting finalized blocks during sync

etan-status:
When a `beaconBlocksByRange` response advances the `safeSlot`, but later
has errors, the sync queue keeps repeating that same request until it is
fulfilled without errors. Data up through `safeSlot` is considered to be
immutable, i.e., finalized, so re-requesting that data is not useful.
By advancing the sync progress in that scenario, those redundant query
portions can be avoided. Note, the finalized block _itself_ is always
requested, even in the initial request. This behaviour is kept same.

### DIFF
--- a/beacon_chain/sync/sync_queue.nim
+++ b/beacon_chain/sync/sync_queue.nim
@@ -511,9 +511,38 @@ proc advanceInput[T](sq: SyncQueue[T], number: uint64) =
 proc notInRange[T](sq: SyncQueue[T], sr: SyncRequest[T]): bool =
   case sq.kind
   of SyncQueueKind.Forward:
-    (sq.queueSize > 0) and (sr.slot != sq.outSlot)
+    (sq.queueSize > 0) and (sr.slot > sq.outSlot)
   of SyncQueueKind.Backward:
-    (sq.queueSize > 0) and (sr.slot + sr.count - 1'u64 != sq.outSlot)
+    (sq.queueSize > 0) and (sr.lastSlot < sq.outSlot)
+
+func numAlreadyKnownSlots[T](sq: SyncQueue[T], sr: SyncRequest[T]): uint64 =
+  ## Compute the number of slots covered by a given `SyncRequest` that are
+  ## already known and, hence, no longer relevant for sync progression.
+  let
+    outSlot = sq.outSlot
+    lowSlot = sr.slot
+    highSlot = sr.lastSlot
+  case sq.kind
+  of SyncQueueKind.Forward:
+    if outSlot > highSlot:
+      # Entire request is no longer relevant.
+      sr.count
+    elif outSlot > lowSlot:
+      # Request is only partially relevant.
+      outSlot - lowSlot
+    else:
+      # Entire request is still relevant.
+      0
+  of SyncQueueKind.Backward:
+    if lowSlot > outSlot:
+      # Entire request is no longer relevant.
+      sr.count
+    elif highSlot > outSlot:
+      # Request is only partially relevant.
+      highSlot - outSlot
+    else:
+      # Entire request is still relevant.
+      0
 
 proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
               data: seq[ref ForkedSignedBeaconBlock],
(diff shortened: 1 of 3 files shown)


beacon_chain/sync/sync_queue.nim
```diff
@@ -643,6 +643,32 @@ proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
...
+          of SyncQueueKind.Backward:
+            if sq.outSlot > safeSlot:
+              sq.outSlot - safeSlot
+            else:
+              0
+      var resetSlot: Option[Slot]
+      if numSlotsAdvanced != 0:
+        debug "Partially advancing sync progress",
+          slot_before = sq.outSlot, slot_after = safeSlot
+        sq.advanceOutput(numSlotsAdvanced)
```
arnetheduck: there's no need to advance output when reset it set: it will reset the output as well

beacon_chain/sync/sync_queue.nim
```diff
@@ -643,6 +643,32 @@ proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
...
+        resetSlot = some(safeSlot)
```
arnetheduck: instead of a full reset here, I think it would be better to push a smaller "debt" to the debt queue instead - ie get the safe slot and when pushing debt, modify it to take the safe slot into account - this will avoid resetting the queue completely (and has the advantage that the next request will be smaller: some clients throttle by making partial responses

beacon_chain/sync/sync_queue.nim
etan-status: No, output is only advanced forward when the entire request was successful. `resetWait` can only adjust backwards.

arnetheduck: right - I wonder why that is .. there's probably some good reason :)

beacon_chain/sync/sync_queue.nim
etan-status: Agree, modifying the `request` is more lightweight. Will adjust.

etan-status:
- Shrink `request` to slots at or past `sq.outSlot` when re-adding to debts queue.
- Apply same shrinking logic for retries of requests due to errors.
- Add test for backward sync direction.
- Rebase to `unstable`.

tests/test_sync_manager.nim
```diff
@@ -445,6 +546,13 @@ suite "SyncManager test suite":
       for item in UnorderedTests:
         unorderedAsyncTest(k, item)
 
+    test prefix & "Good response with missing values towards end":
```
arnetheduck, line 658: if a sync  request debt is fully consumed by a safe slot advance, will an already-queud item get processed as expected? ie we do 2 requests, the latter gets queued first , safe slot moves past first request, first request processing fails then safe slot call discards it instead of pushing to debt - at that point, the queud request should be processed

etan-status:
- Added `outOfBandAdvancementTests`.
- `wakeupWaiters` when out of band advancement is detected.
- Process responses also when `queue.outSlot` is already halfway passed through them.

tests/test_sync_manager.nim
etan-status: Added another test in that area (`outOfBandAdvancementTests`), does this cover that?

In general, requests get processed in-order if a `queueSize > 0` gets specified. With the default `-1`, out-of-order responses lead to a rewind of the sync queue.

etan-status:
- Fix `Backwards` direction test.

etan-status:
- Rebase to `unstable`.

etan-status:
- Move the request shortening logic from `toDebtsQueue` to `pop`.
- Fix response handling to take into account already registered out-of-band progress.

beacon_chain/sync/sync_queue.nim
```diff
@@ -625,14 +654,16 @@ proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
     let retryRequest =
       hasInvalidBlock or unviableBlock.isSome() or missingParentSlot.isSome()
     if not retryRequest:
-      sq.advanceOutput(item.request.count)
+      let numSlotsProgressed = item.request.count - sq.numAlreadyKnownSlots(sr)
+      sq.advanceOutput(numSlotsProgressed)
```
arnetheduck: with `pop` checking that requests aren't "already done", is the advanceoutput here still needed?

etan-status: The `advanceOutput` in `pop` only advances up through the `safeSlot`. For parts not yet considered final, there still needs to be a mechanism to advance progress while no errors are encountered.

etan-status:
- Rebase to `unstable` (only had to re-run tests, no code changes).
- Ran a complete sync on Prater from Genesis successfully.

etan-status:
- Rename `numSlotsProgressed` -> `numSlotsAdvanced`

arnetheduck (approved):

PR #3462, opened 2022-03-03
Work-around the sizeof change in behavior introduced in Nim 1.6

zah:


beacon_chain/networking/eth2_network.nim
```diff
@@ -733,7 +733,19 @@ proc handleIncomingStream(network: Eth2Node,
 
     let deadline = sleepAsync RESP_TIMEOUT
 
-    let msg = if sizeof(MsgRec) > 0:
+    const isEmptyMsg = when MsgRec is object:
```
arnetheduck, line 736: upstream bug report + TODO?

Menduist:
For reference, caused by https://github.com/nim-lang/Nim/pull/14751

PR #3464, opened 2022-03-04
--stop-at-synced-epoch

stefantalpalaru:
This allows benchmarking the initial sync (only forward sync, 1s error margin). Might be useful in CI, with a timeout, as a sanity check.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -261,11 +261,13 @@ type
         name: "verify-finalization" }: bool
 
       stopAtEpoch* {.
+        hidden
         desc: "The wall-time epoch at which to exit the program. (for testing purposes)"
         defaultValue: 0
         name: "stop-at-epoch" }: uint64
 
       stopAtSyncedEpoch* {.
+        hidden
         desc: "The synced epoch at which to exit the program. (for testing purposes)"
         defaultValue: 0
         name: "stop-at-synced-epoch" }: uint64
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -261,9 +261,14 @@ type
         name: "verify-finalization" }: bool
 
       stopAtEpoch* {.
-        desc: "A positive epoch selects the epoch at which to stop"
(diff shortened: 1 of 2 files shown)


beacon_chain/conf.nim
```diff
@@ -261,10 +261,15 @@ type
         name: "verify-finalization" }: bool
 
       stopAtEpoch* {.
-        desc: "A positive epoch selects the epoch at which to stop"
+        desc: "The wall-time epoch at which to exit the program. (for testing purposes)"
```
tersec, line 265: My increasing preference is that these truly testing-only options be `hidden`. The more that can be done to avoid them becoming required to be supported or show up in https://nimbus.guide/options.html even if/when they're not of use for Nimbus development, the better. This also applies to `stopAtSynced`, too, unless there are particularly good reasons for end-users to run it.

stefantalpalaru: Hiding makes sense for end-users, but we'll forget them and end up having to dig through "conf.nim" to find the exact option name.

Marking these in some way in the help output seems like a good compromise between user needs and developer needs.

tersec: My concern is more that the more visible they are, the more someone will use them, and breaking or changing them might be regarded as needing some deprecation period, or keeping the option around indefinitely as a no-op, etc, because someone, somewhere might have used it. That's true somewhat regardless (even hidden options can be used), but it's another signal they're not part of the end-user, supported interface. This is a developer concern, because backwards compatibility is a cost.

stefantalpalaru:
Writing it here, so I don't forget: we should do this sanity check with the distributable binaries we produce in CI.

beacon_chain/conf.nim
stefantalpalaru: OK, I'm hiding both options now.

tersec (approved):

arnetheduck:
FYI we resync two unstable mainnet nodes weekly as of a few days ago: https://github.com/status-im/infra-nimbus/commit/6488efd4e909cfbfaa46236c3f52fb7bc38cb81f

Issue #3453, opened 2022-03-02
Windows performance


stefantalpalaru:
Using the [stop-at-synced-epoch](https://github.com/status-im/nimbus-eth2/tree/stop-at-synced-epoch) branch that adds the option with the same name, I did a rough benchmark of initial syncing speed on mainnet for a regular build running on Linux native and Windows inside VirtualBox:

`rm -rf build/data/shared_mainnet_0; time ./run-mainnet-beacon-node.sh --web3-url="wss://mainnet.infura.io/ws/v3/API_TOKEN" --metrics --metrics-port=8008 --enr-auto-update --doppelganger-detection=off --num-threads=0 --stop-at-synced-epoch=100`

```text
# Linux:
real	1m31.121s
real	1m32.052s
real	1m38.146s

# Windows:
real    3m3.902s
real    3m2.620s
real    3m4.550s
```

I'm already excluding the Windows folder from Defender and VirtualBox is using hardware acceleration, KVM paravirtualisation and host I/O cache to minimise overhead.

The two main things we lack on Windows are lightweight backtraces and LTO.

stefantalpalaru:
The effect of disabling stack traces on Windows:

```text
real    1m51.754s
real    1m52.227s
real    1m53.606s
```

PR #3466, opened 2022-03-04
Windows: give up stack traces for 66% better performance

stefantalpalaru:
See: https://github.com/status-im/nimbus-eth2/issues/3453

TODO: make libbacktrace/libunwind work in MSYS2, to get those stack traces back.


stefantalpalaru:
Recomputed the performance gain. We go from syncing 0.54 epochs/s to 0.90 epochs/s.

PR #3467, opened 2022-03-07
allow using `BlockId` as key in tables

etan-status:
`BlockId` is a type that bundles a block root with its slot number.
The type can be useful as key in tables that deal with non-finalized
blocks (not uniquely identified by slot) and also support pruning
(drop data about older blocks by slot). Instead of creating a custom
type for those use cases, this patch suggests implementing `hash` for
`BlockId` to re-use the existing type.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -55,6 +55,9 @@ type
       ## Slot time for this BlockSlot which may differ from blck.slot when time
       ## has advanced without blocks
 
+func hash*(bid: BlockId): Hash =
+  hash(bid.root)
+
 template root*(blck: BlockRef): Eth2Digest = blck.bid.root
 template slot*(blck: BlockRef): Slot = blck.bid.slot
 


beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -55,6 +55,12 @@ type
       ## Slot time for this BlockSlot which may differ from blck.slot when time
       ## has advanced without blocks
 
+func hash*(bid: BlockId): Hash =
+  var h: Hash = 0
+  h = h !& hash(bid.root)
+  h = h !& hash(bid.slot)
```
arnetheduck: the `root` field already covers `slot` (because slot is part of `hash_tree_root`) so this is a bit pointless in valid blockid:s

etan-status: Good point, I'll remove the extra hashing of `slot`. Even for invalid blockid:s it won't break it, it would just be less efficient. And for block at a different slot, there is `BlockSlotId`.

arnetheduck (approved):

PR #3472, opened 2022-03-07
relax `BlockRef` database assumptions

arnetheduck:
* remove `getForkedBlock(BlockRef)` which assumes block data exists but
doesn't support archive/backfilled blocks
* fix REST `/eth/v1/beacon/headers` request not returning
archive/backfilled blocks
* avoid re-encoding in REST block SSZ requests (using `getBlockSSZ`)

### DIFF
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -198,7 +198,7 @@ func toBlockSlotId*(bs: BlockSlot): BlockSlotId =
 
 func isProposed*(bid: BlockId, slot: Slot): bool =
   ## Return true if `bid` was proposed in the given slot
-  bid.slot == slot
+  bid.slot == slot and not bid.root.isZero
 
 func isProposed*(blck: BlockRef, slot: Slot): bool =
   ## Return true if `blck` was proposed in the given slot
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -205,19 +205,24 @@ template validateBeaconBlockBellatrix(
       # mostly relevant around merge transition epochs. It's possible that
       # the previous block is phase 0 or Altair, if this is the transition
       # block itself.
-      let blockData = dag.getForkedBlock(parent)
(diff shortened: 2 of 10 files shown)


beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -198,7 +198,7 @@ func toBlockSlotId*(bs: BlockSlot): BlockSlotId =
 
 func isProposed*(bid: BlockId, slot: Slot): bool =
   ## Return true if `bid` was proposed in the given slot
-  bid.slot == slot
+  bid.slot == slot and not bid.root.isZero
```
mratsim, line 201: Do we have a test for that? Or in which situation do we get a 0 block id (backfill?)

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -294,11 +299,13 @@ proc validateBeaconBlock*(
 
   if slotBlock.isProposed() and
       slotBlock.blck.slot == signed_beacon_block.message.slot:
-    let data = dag.getForkedBlock(slotBlock.blck)
-    if getForkedBlockField(data, proposer_index) ==
-          signed_beacon_block.message.proposer_index and
-        data.signature.toRaw() != signed_beacon_block.signature.toRaw():
-      return errIgnore("BeaconBlock: already proposed in the same slot")
+    let curBlock = dag.getForkedBlock(slotBlock.blck.bid)
+    if curBlock.isOk():
```
mratsim, line 303: Nothing happens if the curBlock is an error?

beacon_chain/consensus_object_pools/block_dag.nim
arnetheduck: zero is returned if the block is not found (should probably change that to Option at some point but .. ) - there were tests but they didn't cover zero

beacon_chain/gossip_processing/gossip_validation.nim
arnetheduck: nothing - we don't have the block so there's nothing to compare with (should typically not happen, except in weird timing scenarios)

PR #3474, opened 2022-03-09
add option to collect light client data

etan-status:
Light clients require full nodes to serve additional data so that they
can stay in sync with the network. This patch adds a new launch option
`--import-light-client-data` to configure what data to make available.
For now, data is only kept in memory; it is not persisted at this time.
Note that data is only locally collected, a separate patch is needed to
actually make it availble over the network. `--serve-light-client-data`
will be used for serving data, but is not functional yet outside tests.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
@@ -1,0 +1,94 @@
+# beacon_chain
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+# This implements the pre-release proposal of the libp2p based light client sync
+# protocol. See https://github.com/ethereum/consensus-specs/pull/2802
+
+import
+  # Status libraries
+  stew/bitops2,
+  # Beacon chain internals
+  ../spec/datatypes/altair,
+  ./block_dag
+
+type
+  OnOptimisticLightClientUpdateCallback* =
+    proc(data: OptimisticLightClientUpdate) {.gcsafe, raises: [Defect].}
+
+  ImportLightClientData* {.pure.} = enum
+    ## Controls which classes of light client data are imported.
+    None = "none"
+      ## Import no light client data.
+    OnlyNew = "only-new"
+      ## Import only new light client data.
+    Full = "full"
+      ## Import light client data for entire weak subjectivity period.
+    OnDemand = "on-demand"
+      ## No precompute of historic data. Is slow and may miss validator duties.
+
+  CachedLightClientData* = object
+    ## Cached data from historical non-finalized states to improve speed when
+    ## creating future `LightClientUpdate` and `LightClientBootstrap` instances.
+    current_sync_committee_branch*:
+      array[log2trunc(altair.CURRENT_SYNC_COMMITTEE_INDEX), Eth2Digest]
+
+    next_sync_committee_branch*:
+      array[log2trunc(altair.NEXT_SYNC_COMMITTEE_INDEX), Eth2Digest]
+
+    finalized_bid*: BlockId
+    finality_branch*:
+      array[log2trunc(altair.FINALIZED_ROOT_INDEX), Eth2Digest]
+
+  CachedLightClientBootstrap* = object
+    ## Cached data from historical finalized epoch boundary blocks to improve
+    ## speed when creating future `LightClientBootstrap` instances.
+    current_sync_committee_branch*:
+      array[log2trunc(altair.CURRENT_SYNC_COMMITTEE_INDEX), Eth2Digest]
+
+  LightClientCache* = object
+    data*: Table[BlockId, CachedLightClientData]
+      ## Cached data for creating future `LightClientUpdate` instances.
+      ## Key is the block ID of which the post state was used to get the data.
+      ## Data is stored for the most recent 4 finalized checkpoints, as well as
+      ## for all non-finalized blocks.
+
+    bootstrap*: Table[Slot, CachedLightClientBootstrap]
+      ## Cached data for creating future `LightClientBootstrap` instances.
+      ## Key is the block slot of which the post state was used to get the data.
+      ## Data is stored for finalized epoch boundary blocks.
+
+    latestCheckpoints*: array[4, Checkpoint]
+      ## Keeps track of the latest four `finalized_checkpoint` references
+      ## leading to `finalizedHead`. Used to prune `data`.
+      ## Non-finalized states may only refer to these checkpoints.
+
+    lastCheckpointIndex*: int
+      ## Last index that was modified in `latestCheckpoints`.
+
+    bestUpdates*: Table[SyncCommitteePeriod, altair.LightClientUpdate]
+      ## Stores the `LightClientUpdate` with the most `sync_committee_bits` per
+      ## `SyncCommitteePeriod`. Updates with a finality proof have precedence.
+
+    pendingBestUpdates*:
+      Table[(SyncCommitteePeriod, Eth2Digest), altair.LightClientUpdate]
+      ## Same as `bestUpdates`, but for `SyncCommitteePeriod` with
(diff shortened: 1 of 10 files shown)


etan-status:
Should add another test case for checkpoint sync, and also allow collecting data when data of last 4 finalized checkpoints and non-finalized blocks are not all known.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,889 @@
...
+    minSlot = max(altairStartSlot, dag.tail.slot)
+  if currentSlot - minSlot < MIN_SLOTS_FOR_BLOCK_REQUESTS:
+    return minSlot
+
+  let earliestSlot = currentSlot - MIN_SLOTS_FOR_BLOCK_REQUESTS
+  max(earliestSlot.sync_committee_period.start_slot, minSlot)
+
+proc currentSyncCommitteeForPeriod(
+    dag: ChainDAGRef,
+    tmpState: ref StateData,
```
arnetheduck: generally we avoid `ref` as parameter if it's not necessary (except chaindagref which needs refactoring to remove the Ref as far as possible)

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,889 @@
...
+            dag.lightClientCache.lastCheckpointIndex = cpIndex
+            inc cpIndex
+
+            # Save new checkpoint block using `tmpState` (avoid replay after it)
+            if checkpoint.root != dag.finalizedHead.blck.root:
+              let cpRef =
+                dag.getBlockAtSlot(checkpoint.epoch.start_slot).blck
+              if cpRef != nil and cpRef.slot >= earliestSlot:
+                assert cpRef.bid.root == checkpoint.root
+                doAssert dag.updateStateData(
```
arnetheduck, line 789: so basically if I read this code right, it performs almost the same traversal twice: once for the blocks and once for the (lagging) finalized checkpoint - in the typical case, finalized checkpoints lag head by 2 epochs so whenever there's more than 2 epochs to process, this will do the exact same processing for the overlapping period - an alternative here would be to collect the checkpoint data for _every_ epoch into a cache, then apply it with a lag when finalized_checkpoint changes

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -0,0 +1,889 @@
...
+                  tmpState[], cpRef.atSlot, save = false, tmpCache)
```
arnetheduck, line 790: this looks off when slots are empty: `getBlockAtSlot` will return the most recent block in this case which might be from a different epoch than the checkpoint epoch

etan-status: Yes, which is the reason why I need to track latest 4 epochs. The sync committee refers to the checkpoint block at its originating slot, not at the epoch boundary slot.

etan-status: Reason being that light client is unable to advance slots. So, to validate the checkpoint's `BeaconBlockHeader`, via merkle proof, it needs to be at its original slot, which may be in a different epoch.

arnetheduck: ok this deserves mention in a comment

arnetheduck (approved):
In general, we can merge this and start running on a couple of `unstable` nodes just to see how it behaves - some tuning and optimization work will likely be needed here to avoid unnecessary work and degrade gracefully when blocks and states are missing etc, but in general looks good for a first cut.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Replaced with `var`.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: There are two states here, `dag.headState`, and `tmpState`.

`dag.headState` is used to replay all non-finalized blocks from `dag.finalizedHead` to `dag.head`, while `tmpState` is used to collect historic data about the various encountered `finalized_checkpoint.root` that the non-finalized blocks are referring to.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Added.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Note that, because light client finality proofs refer to checkpoints at their original slot, any individual non-finalized block may later become a checkpoint. We do not know in advance which ones will be orphaned, which ones will be the final slot of their epoch, etc. So finality proof data needs to be collected for all non-finalized slots, not just one per epoch.

etan-status:
- Add test for initialization from checkpoint sync.
- Do not rewind non-finalized blocks in `only-new` import mode. May lead to suboptimal (but still correct) data for current period, but speeds up startup time.

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
```diff
@@ -0,0 +1,94 @@
...
+      ## Tracks the `LightClientUpdate` for the latest slot. This may be older
+      ## than head for empty slots or if not signed by sync committee.
+
+    optimisticUpdate*: OptimisticLightClientUpdate
+      ## Tracks the `OptimisticLightClientUpdate` for the latest slot. This may
+      ## be older than head for empty slots or if not signed by sync committee.
+
+    importTailSlot*: Slot
+      ## The earliest slot for which light client data is collected.
+      ## Only relevant for `ImportLightClientData.OnlyNew`.
```
etan-status, line 94: Added `importTailSlot` for `OnlyNew` import mode. Data is only collected from this slot onward.

tests/test_light_client.nim
```diff
@@ -0,0 +1,189 @@
...
+      res = process_light_client_update(
+        store, latestUpdate.get, currentSlot, cfg, genesis_validators_root)
+    check:
+      latestUpdate.isSome
+      latestUpdate.get.attested_header.slot == dag.headState.blck.parent.slot
+      res
+      store.finalized_header == latestUpdate.get.finalized_header
+      store.optimistic_header == latestUpdate.get.attested_header
+
+  test "Init from checkpoint":
```
etan-status, line 166: This is the new checkpointing test. Simply checks whether there are any errors when initializing from a checkpoint.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Proof computation is very fast on a hot state (`hash_tree_root` are already cached), and proofs themselves are quite small. 11 roots for `LightClientUpdate` and 6 roots for `LightClientBootstrap`.

etan-status:
> In general, we can merge this and start running on a couple of `unstable` nodes just to see how it behaves

One remaining question would be what import mode to use.
- `OnlyNew` has no startup time increase but only includes data from startup onwards.
- `Full` is deterministic and includes the max weak subjectivity period, but has very slow startup time.
- `OnDemand` has ~10 seconds increased startup time, and has lag spikes when data is being requested, but otherwise behaves like `Full`.

arnetheduck:
Let's go with `OnlyNew` for now - a bit of chaos in the tests will stress the system as well

etan-status:
- Fix `backwards` direction test.

etan-status:
- Rebase to `unstable`, handling `getForkedBlock(BlockRef)` that got removed assuming blocks for existing `BlockRef` always succeed in loading.

etan-status:
> Let's go with `OnlyNew` for now - a bit of chaos in the tests will stress the system as well

https://github.com/status-im/nimbus-eth2/pull/3487

tests/test_light_client.nim
arnetheduck: this test ends up with bogus data (zero hashes) in some of its internal structures - I've disabled it in https://github.com/status-im/nimbus-eth2/pull/3478/commits/1249b8c4a56c13c8888f951745a07ebbd6fde807 because the errors surface with the new, more strict, error handling - this, and checkpoint sync in general, needs revisiting

etan-status: https://github.com/status-im/nimbus-eth2/pull/3498

PR #3475, opened 2022-03-09
Update the options page in the guide with explanation regarding the format of the config file

### LINKED ISSUE #1793: Config file
A user config file that supports:

- loglevel
- network
- infura endpoint
- ports

What else?

Ideally, would be automagically fetched when you run `make medalla`.

Otherwise

```
make medalla --config=myfile.config
```

Same for beacon node (needs to be done before we start recommending running the beacon node manually imo).

zah:


docs/the_nimbus_book/src/options.md
```diff
@@ -96,3 +105,20 @@ The following options are available:
...
+All command line options can also be provided in a [TOML](https://toml.io/en/) config file specified through the
+`--config-file` option. Within the config file, you need to use the long names of all options and options affecting
+Nimbus sub-commands should appear in a section of the file matching the sub-command name. Here is an example config
+file illustrating the above:
+
+```
+# nimbus-eth2-config.toml
+
+doppelganger-detection = on
+web3-url = "ws://192.168.1.10:8000"
```
arnetheduck: toml doesn't allow duplicate keys - how would one specify two web3 urls?

zah: Right, the example here is wrong, it should have been an array.

zah: More remarks added, please take another look.

docs/the_nimbus_book/src/options.md
```diff
@@ -96,3 +105,29 @@ The following options are available:
 
 ...
 ```
+
+All command line options can also be provided in a [TOML](https://toml.io/en/)
+config file specified through the `--config-file` flag. Within the config file,
+you need to use the long names of all options. Please note that certain options
```
arnetheduck, line 111: could lighten this up with some whitespace, it's quite dense (ie more whitespace and less words / prose) - or make a comment in the example toml file next to the relevant example..

what happens if you give a string and not an array?

zah: With the tweak added in https://github.com/status-im/nimbus-eth2/pull/3479, it now produces the following helpful error message:

> Failure while loading the configuration:
  /Users/zahary/status/nimbus-eth2/nimbus-eth2.toml(5, 12) expected '"["'
  Since the 'web3-url' option is allowed to have more than one value, please make sure to supply a properly formatted TOML array
  
(5, 12) are line and column in the configuration file.

arnetheduck: :+1: I was more thinking about the wall of text here in the manual, but that's a nice improvement as well

zah: I'll merge this now to not delay our release any further. We can revisit the wall of text in a follow up PR.

PR #3476, opened 2022-03-09
add support for `ResourceUnavailable` p2p error

etan-status:
The `p2p-interface.md` spec defines a `ResourceUnavailable` error to
return in situations where data that exists on the network is locally
unavailable, e.g., when a block within `MIN_EPOCHS_FOR_BLOCK_REQUESTS`
is requested by `BeaconBlocksByRange` but cannot be provided. This patch
adds support for that additional error code.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -158,6 +158,7 @@ type
     Success
     InvalidRequest
     ServerError
+    ResourceUnavailable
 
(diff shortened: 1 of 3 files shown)


etan-status:
- On server side, treats it as a server error, not triggering peer de-scoring in `eth2_network`
- On client side, treats it as `PeerScoreNoBlocks` in `sync_queue`

zah (approved):

PR #3479, opened 2022-03-09
Helpful error message when the user fails to use an array type in TOML

zah:


beacon_chain/nimbus_binary_common.nim
```diff
@@ -184,13 +184,30 @@ template makeBannerAndConfig*(clientId: string, ConfType: type): untyped =
...
+  except CatchableError as err:
+    # We need to log to stderr here, because logging hasn't been configured yet
+    stderr.write "Failure while loading the configuration:\n"
+    stderr.write err.msg
+    stderr.write "\n"
+
+    if err[] of ConfigurationError and
+       err.parent != nil and
+       err.parent[] of TomlFieldReadingError:
+      let fieldName = ((ref TomlFieldReadingError)(err.parent)).field
```
arnetheduck, line 204: the message is nice but .. this abstraction leakage, less so :/ if someone ends up working on the toml codebase it's pretty impossible to know that consumers of the code are doing ugly hacks like this - generally, when working with exceptions, each layer would ideally translate them into something appropriate for that level of abstraction

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-2022-03-07 v22.3.0
+2022-03-07 v22.3.1
```
arnetheduck: since the release hasn't been tagged, we can skip the point change

arnetheduck (approved):

PR #3480, opened 2022-03-09
Grafana: make local dashboard work with Windows metrics

stefantalpalaru:


grafana/beacon_nodes_Grafana_dashboard.json
```diff
@@ -3631,15 +3631,15 @@
...
+        "definition": "label_values(nim_gc_mem_bytes,instance)",
         "hide": 0,
         "includeAll": false,
         "multi": false,
         "name": "instance",
         "options": [],
         "query": {
-          "query": "label_values(process_virtual_memory_bytes,instance)",
-          "refId": "Prometheus-instance-Variable-Query"
+          "query": "label_values(nim_gc_mem_bytes,instance)",
```
arnetheduck, line 3641: but this is a different metric - it doesn't show what happens under memory pressure / swapping which is more interesting?

arnetheduck: ie the fix here would be on the nim-metrics side to produce such a metric for windows, no?

stefantalpalaru: It doesn't matter which metric it is, as long as it's present everywhere. We use it to read its labels and get a list of instances for the dashboard drop-down in the upper left corner.

PR #3484, opened 2022-03-11
Windows: disable status bar

stefantalpalaru:
It can randomly lock inside Windows terminal emulators, as I saw twice while running a mainnet node for 8 days.

Better play it safe and disable the status bar completely on Windows.

The nim-confutils bump is needed to disable two command line arguments with the same custom pragma: https://github.com/status-im/nim-confutils/pull/42

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -46,8 +46,10 @@ const
 
 when defined(windows):
   {.pragma: windowsOnly.}
+  {.pragma: posixOnly, hidden.}
 else:
   {.pragma: windowsOnly, hidden.}
+  {.pragma: posixOnly.}
 
 type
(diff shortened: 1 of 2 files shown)


tersec (approved):
Brute force, but I agree, better to do this and allow someone who wants to investigate root causes to do so without its being in the critical path of deployment, for a non-essential feature.

arnetheduck:
this should be accompanied by an issue in the eth2 repo and a `TODO` link in the code

stefantalpalaru:
> this should be accompanied by an issue in the eth2 repo and a TODO link in the code

I don't think it's feasible to use fancy ANSI control sequences on Windows terminal emulators, so I don't see this as something worth supporting in the future.

The status bar information we care about (sync status, no. of peers, etc.) is already present in the logs.

PR #3485, opened 2022-03-11
allow tagging light client specific libp2p messages

etan-status:
The pre-release light client sync protocol defines additional Req/Resp
messages to be made available when `--serve-light-client-data` is set.
This patch extends the `{.libp2pProtocol.}` pragma with an optional
parameter to tag such light client sync protocol specific messages.
The corresponding protocols are only selectively registered with libp2p.


arnetheduck (approved):
wfm unless @zah, our resident practitioner of the dark arts, disagrees ;)

PR #3486, opened 2022-03-11
libp2p light client gossip validation

etan-status:
When `--serve-light-client-data` is specified, provides stability on the
`optimistic_light_client_update` GossipSub topic.

### DIFF
--- a/beacon_chain/spec/network.nim
+++ b/beacon_chain/spec/network.nim
@@ -7,6 +7,10 @@
 
 {.push raises: [Defect].}
 
+# References to `vFuture` refer to the pre-release proposal of the libp2p based
+# light client sync protocol. Conflicting release versions are not in use.
+# https://github.com/ethereum/consensus-specs/pull/2802
+
 import
   "."/[helpers, forks],
   "."/datatypes/base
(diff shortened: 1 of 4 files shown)


beacon_chain/spec/network.nim
```diff
@@ -94,6 +98,11 @@ func getSyncCommitteeContributionAndProofTopic*(forkDigest: ForkDigest): string
   ## For subscribing and unsubscribing to/from a subnet.
   eth2Prefix(forkDigest) & "sync_committee_contribution_and_proof/ssz_snappy"
 
+# https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#optimistic_light_client_update
+func getOptimisticLightClientUpdateTopic*(forkDigest: ForkDigest): string =
+  ## For broadcasting or obtaining the latest `OptimisticLightClientUpdate`.
+  eth2Prefix(forkDigest) & "optimistic_light_client_update/ssz_snappy"
```
etan-status: Is there a versioning strategy for these as well, similar to the Req/Resp ones, or is this alright?

arnetheduck: the fork digest is the version for "normal" requests, but we might want to change this to `optimistic_light_client_update_v0` so as to avoid mixups

arnetheduck (approved):

beacon_chain/spec/network.nim
etan-status: OK, going with `_v0` then.

tersec (approved):

arnetheduck:
to make this work (even on testnets), we'll need to make sure that the mesh quality is taken into consideration when deciding to look for more peers, akin to attestation subnets etc

PR #3487, opened 2022-03-11
serve light client data in CI / dev builds

etan-status:
Adjust config for CI / dev builds to serve light client data by default:
`--serve-light-client-data=1 --import-light-client-data=only-new`

### DIFF
--- a/scripts/launch_local_testnet.sh
+++ b/scripts/launch_local_testnet.sh
@@ -597,6 +597,7 @@ for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
     ${STOP_AT_EPOCH_FLAG} \
     --rest-port="$(( BASE_REST_PORT + NUM_NODE ))" \
     --metrics-port="$(( BASE_METRICS_PORT + NUM_NODE ))" \
+    --serve-light-client-data=1 --import-light-client-data=only-new \
     ${EXTRA_ARGS} \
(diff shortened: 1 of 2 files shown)


scripts/launch_local_testnet.sh
```diff
@@ -187,6 +189,14 @@ while true; do
       LIGHTHOUSE_VC_NODES="$2"
       shift 2
       ;;
+    --serve-light-client-data)
+      SERVE_LIGHT_CLIENT_DATA="$2"
+      shift 2
+      ;;
+    --import-light-client-data)
```
stefantalpalaru: Also needs entries in the help section, a hundred lines above.

etan-status: In `conf.nim`, these are hidden for now. Should the docs already be exposed here?

stefantalpalaru: Yes, it's how we document this script. No concept of hidden argument here.

etan-status: Actually, seems like passing the same option multiple times with conflicting values is fine. As these options are pass-through, can just pass the defaults before the user options, and don't need the extra wrapping here. Also avoids the docs question.

stefantalpalaru (approved):

etan-status:
@tersec Would be nice if we could also have this enabled on Kiln.

PR #3489, opened 2022-03-12
Rpc pubkey

### LINKED ISSUE #1715: [SEC] Observations on Point Validation
## Description

This is a strictly informational observation intended to generate thought on the overall point validation strategy.

The BLS Signature specification places a high importance on point validation as evidenced by the following excerpt:

> [Section 5.1.  Validating public keys](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-5.1): All algorithms in Section 2 and Section 3 that operate on public keys require first validating those keys. ...

> [Section 5.2.  Skipping membership check](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-02#section-5.2): Some existing implementations skip the `signature_subgroup_check` invocation in CoreVerify (Section 2.7), whose purpose is ensuring that the signature is an element of a prime-order subgroup.  This check is REQUIRED of conforming implementations, for two reasons.

>   1.  For most pairing-friendly elliptic curves used in practice, the pairing operation e (Section 1.3) is undefined when its input points are not in the prime-order subgroups of E1 and E2.  The resulting behavior is unpredictable, and may enable forgeries.
>   2.  Even if the pairing operation behaves properly on inputs that are outside the correct subgroups, skipping the subgroup check breaks the strong unforgeability property.

The BLS Signature functionality implemented in `nim-blscurve` does in fact perform the validations described above. However, the `parsePubkey()` function in `beacon_chain/validator_api.nim` shown below simply deserializes a hex string and passes it along to downstream logic without further (subgroup) validation.

https://github.com/status-im/nim-beacon-chain/blob/4b38619c4f51847fc591e85327ed226ad46d24ac/beacon_chain/validator_api.nim#L36-L40

The above case itself is not a significant security issue, as the result is to be returned in responses for the `get_v1_beacon_states_stateId_validators` and `get_v1_beacon_states_stateId_validators_validatorId` RPC endpoints which should be validated by the request originator prior to usage. Arguably, point validation at usage and ingestion is far more important than an output path.

However, it is noted that the BLS signature functionality specification assumes data originates and terminates as byte strings (e.g., the `signature_to_point()` function is step 1 in the `CoreVerify`, `Aggregate`, and `CoreAggregateVerify` algorithms, while `point_to_signature()` is essentially the last step in the `CoreSign`, `Aggregate`, `PopProve` algorithms). In contrast, the `nim-crypto` and `nim-beacon-chain` implementation tends to deserialize points once, store them internally in point form, and pass these points to the individual algorithms. 

While there are a variety of places and paths in beacon chain to deserialize points, they all go through the lowest-level byte-oriented function (i.e., the `fromBytes()` function shown below). This function serves as a 'choke point' for G2 (E2) and there is a nearby sibling for G1 (E1). They currently do not perform subgroup validation.

https://github.com/status-im/nim-blscurve/blob/da9ae49dab4cbb95b2cdaa68ecc221ee15c67a9a/blscurve/miracl/common.nim#L649-L674

~~~nim
func fromBytes*(res: var ECP2_BLS12381, data: openarray[byte]): bool =
  ## Unserialize ECP2(G2) point from array of bytes ``data``.
  ##
  ## This procedure supports only compressed form of serialization.
  ##
  ## Returns ``true`` on success, ``false`` otherwise.
  if len(data) >= MODBYTES_384 * 2:
    if (data[0] and (1'u8 shl 7)) != 0:
      if (data[0] and (1'u8 shl 6)) != 0:
        # Infinity point
        res.inf()
        result = true
      else:
        var buffer: array[MODBYTES_384, byte]
        var x1, x0: BIG384
        let greatest = (data[0] and (1'u8 shl 5)) != 0'u8
        copyMem(addr buffer[0], unsafeAddr data[0], MODBYTES_384)
        buffer[0] = buffer[0] and 0x1F'u8
        if x1.fromBytes(buffer):
          copyMem(addr buffer[0], unsafeAddr data[MODBYTES_384], MODBYTES_384)
          if x0.fromBytes(buffer):
            var x: FP2_BLS12381
            x.fromBigs(x0, x1)
            if res.setx(x, greatest) == 1:
              result = true
~~~

In any event, ensuring all points are validated correctly prior to processing remains a constant challenge as the code evolves. Further, note that that this validation is expensive: point multiplication by the cofactor which is a 125-bit integer.

## Exploit Scenario

The original `parsePubkey()` function cited above demonstrates the potential for missing validation checks. As stated in the referenced standard, the pairing operation is undefined when its input points are not in the prime-order subgroups of E1 and E2.  The resulting behavior is unpredictable and may enable forgeries.

## Mitigation Recommendation

The recommendation is to consider placing the point (subgroup) validation check at the location of deserialization immediately after the code shown above. The two routines could perform point validation by default but skip validation as necessary when given an optional flag. The point structure could include a flag indicating the check has been performed, so that algorithms can confirm the check (rather than repeat it) or execute it in a lazy fashion (though this risks mutability issues). Alternatively, a validated point could be given a different type to cleanly separate/indicate its status, e.g. `unverifiedPubKey` and `verifiedPubKey`. Overall, these approaches would allow the removal of a variety of potentially redundant and expensive checks in the BLS signature functionality and increase performance as well as robustly ensuring all points are validated prior to processing. Finally, detecting errors early at ingestion improves debug and attribution capabilities.

mratsim:
This centralizes all crypto handling in crypto.nim instead of having it sprawled out.

In particular, for REST and RPC, pubkeys are supplied by users and due to lazy validation, those are unchecked (no check that they are infnity, on curve or in subgroup).
Fortunately:
1. We require the endpoint to be exposed only to trusted services (either local or in a safe network)
2. All codepaths in REST and RPC compares the pubkey with the known one from the Ethereum Beacon State, hence invalid keys cannot slip in.

Alternative: since RPC is scheduled for deprecation, we can also deleted the related files instead of merging this PR.

Regarding REST, validators pubkeys go through ValidatorIdent and/or `node.restKeysCache` and those are compared to the known validators key before use so no issue there: 

https://github.com/status-im/nimbus-eth2/blob/13b264d509d48909bb354f85b879f22641723509/beacon_chain/rpc/rest_beacon_api.nim#L339-L362

Closes #1715

### DIFF
--- a/beacon_chain/rpc/rpc_beacon_api.nim
+++ b/beacon_chain/rpc/rpc_beacon_api.nim
@@ -26,6 +26,14 @@ type
   RpcServer = RpcHttpServer
 
   ValidatorQuery = object
+    # Security note / threat model:
+    # - The validator pubkey are stored in their raw bytes representation
+    #   in the `keyset`.
+    # - While the input is from unknown source (as far as the beacon node is concerned),
+    #   users are asked to not expose their
+    #   RPC endpoints to untrusted network or use a reverse proxy in front.
+    # - At usage time, keys in the keyset are compared to keys registered
+    #   in the Ethereum BeaconState which are valid
     keyset: HashSet[ValidatorPubKey]
     ids: seq[uint64]
 
@@ -35,18 +43,7 @@ type
 template unimplemented() =
   raise (ref CatchableError)(msg: "Unimplemented")
 
-proc parsePubkey(str: string): ValidatorPubKey {.raises: [Defect, ValueError].} =
-  const expectedLen = RawPubKeySize * 2 + 2
-  if str.len != expectedLen: # +2 because of the `0x` prefix
-    raise newException(ValueError,
-      "A hex public key should be exactly " & $expectedLen & " characters. " &
-      $str.len & " provided")
-  let pubkeyRes = fromHex(ValidatorPubKey, str)
-  if pubkeyRes.isErr:
-    raise newException(ValueError, "Not a valid public key")
-  return pubkeyRes[]
-
-proc createIdQuery(ids: openArray[string]): Result[ValidatorQuery, string] =
+proc createIdQuery(ids: openArray[string]): Result[ValidatorQuery, cstring] =
   # validatorIds array should have maximum 30 items, and all items should be
   # unique.
   if len(ids) > 30:
@@ -63,19 +60,15 @@ proc createIdQuery(ids: openArray[string]): Result[ValidatorQuery, string] =
 
   for item in ids:
     if item.startsWith("0x"):
-      if len(item) != RawPubKeySize * 2 + 2:
-        return err("Incorrect hexadecimal key")
-      let pubkeyRes = ValidatorPubKey.fromHex(item)
-      if pubkeyRes.isErr:
-        return err("Incorrect public key")
-      res.keyset.incl(pubkeyRes.get())
+      let pubkey = ? ValidatorPubkey.fromHex(item)
+      res.keyset.incl(pubkey)
     else:
       var tmp: uint64
       try:
         if parseBiggestUInt(item, tmp) != len(item):
           return err("Incorrect index value")
       except ValueError:
-        return err("Cannot parse index value: " & item)
+        return err("Cannot parse index value")
       res.ids.add(tmp)
   ok(res)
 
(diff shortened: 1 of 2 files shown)


arnetheduck (approved):
lgtm - debt of this kind has a tendency to spread, even for deprecated stuff

PR #3490, opened 2022-03-12
error and progress codes for light client sync

etan-status:
When syncing as a light client, different behaviour is needed to handle
the various ways how errors may occur. The existing logic for blocks can
also be applied to light client objects:
- `Invalid`: Malformed object that is clearly an error by its producer.
- `MissingParent`: More data is needed to decide applicability.
- `UnviableFork`: Object may be valid but will never apply on this fork.
- `Duplicate`: No errors were encountered but the object was not useful.

### DIFF
--- a/beacon_chain/spec/light_client_sync.nim
+++ b/beacon_chain/spec/light_client_sync.nim
@@ -16,6 +16,8 @@ import
   datatypes/altair,
   helpers
 
+from ../consensus_object_pools/block_pools_types import BlockError
+
 func period_contains_fork_version(
     cfg: RuntimeConfig,
     period: SyncCommitteePeriod,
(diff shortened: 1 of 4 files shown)


beacon_chain/spec/light_client_sync.nim
```diff
@@ -90,37 +92,39 @@ proc validate_light_client_update*(
...
+        return err(BlockError.Duplicate)
 
   # Verify update does not skip a sync committee period
   let
     finalized_period = store.finalized_header.slot.sync_committee_period
     update_period = active_header.slot.sync_committee_period
-  if update_period notin [finalized_period, finalized_period + 1]:
-    return false
+  if update_period < finalized_period:
+    return err(BlockError.Duplicate)
```
tersec, line 125: Some of these are a little arbitrary --- e.g., as far as I understand this, this might not exactly be a duplicate, but cover what amounts to a duplicate time period. But they're all reasonable, and typically one of only maybe a couple of plausible choices in each case. Worth going with this, and if the code that ends up checking them wants to cluster error handling/recovery in slightly different ways, there's no real design-debt or difficulty in tweaking the exact type of block error.

tersec (approved):

beacon_chain/spec/light_client_sync.nim
arnetheduck: indeed, the principled approach here would be a `SyncError` in sync manager, then light client having its own errors, and then a translation function in the "block validation" callback which at that point is no longer a block validation, but a "data validation" as sync manager no longer operates on blocks only.. that said, it can be left for a future PR.

arnetheduck (approved):

PR #3492, opened 2022-03-14
bump nim-stew

tersec:
This is the first step to enabling JWT engine authentication support -- `kiln-dev-auth` currently has it, but nothing else does. This does so by adding support for passing a callback function which supplies HTTP request/WS connection headers.


arnetheduck:
copy-pasting is the decentralization of commit messages :)

arnetheduck (approved):

PR #3494, opened 2022-03-14
Kiln (JWT auth version)

tersec:
Implements/targets https://hackmd.io/@n0ble/kiln-spec

Replaces https://github.com/status-im/nimbus-eth2/pull/3093

Don't merge.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -109,23 +109,6 @@ OK: 2/2 Fail: 0/2 Skip: 0/2
 + parent sanity                                                                              OK
 ```
 OK: 2/2 Fail: 0/2 Skip: 0/2
-## DeleteKeys requests [Preset: mainnet]
-```diff
-+ Deleting not existing key [Preset: mainnet]                                                OK
-+ Invalid Authorization Header [Preset: mainnet]                                             OK
-+ Invalid Authorization Token [Preset: mainnet]                                              OK
-+ Missing Authorization header [Preset: mainnet]                                             OK
-```
-OK: 4/4 Fail: 0/4 Skip: 0/4
-## DeleteRemoteKeys requests [Preset: mainnet]
-```diff
-+ Deleting existing local key and remote key [Preset: mainnet]                               OK
-+ Deleting not existing key [Preset: mainnet]                                                OK
-+ Invalid Authorization Header [Preset: mainnet]                                             OK
-+ Invalid Authorization Token [Preset: mainnet]                                              OK
-+ Missing Authorization header [Preset: mainnet]                                             OK
-```
-OK: 5/5 Fail: 0/5 Skip: 0/5
 ## Diverging hardforks
 ```diff
 + Non-tail block in common                                                                   OK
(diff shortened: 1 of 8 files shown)


AllTests-mainnet.md
```diff
@@ -102,23 +102,6 @@ OK: 2/2 Fail: 0/2 Skip: 0/2
 + epochAncestor sanity [Preset: mainnet]                                                     OK
 ```
 OK: 1/1 Fail: 0/1 Skip: 0/1
-## DeleteKeys requests [Preset: mainnet]
-```diff
-+ Deleting not existing key [Preset: mainnet]                                                OK
-+ Invalid Authorization Header [Preset: mainnet]                                             OK
-+ Invalid Authorization Token [Preset: mainnet]                                              OK
-+ Missing Authorization header [Preset: mainnet]                                             OK
```
mratsim, line 117: What removed those? Was that missed from another PR? (So people might create conflicts when adding new tests based on unstable)

beacon_chain/conf.nim
```diff
@@ -95,7 +95,7 @@ type
   BeaconNodeConf* = object
     logLevel* {.
       desc: "Sets the log level for process and topics (e.g. \"DEBUG; TRACE:discv5,libp2p; REQUIRED:none; DISABLED:none\")"
-      defaultValue: "INFO"
+      defaultValue: "DEBUG"
```
mratsim: Debug leaking?

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -42,6 +43,8 @@ type
     bid*: BlockId ##\
       ## Root that can be used to retrieve block data from database
 
+    executionBlockRoot*: Eth2Digest
```
mratsim: Should we consider distinct types here? (We do have attestation_root, state_root and block_root already but now also ExecutionBlockRoot)

beacon_chain/conf.nim
tersec: Oh, indeed, but I'm keeping it. This is mostly to ensure it actually runs in CI. Happy to engage with code review here though. `INFO` logs aren't that useful for merge testnet debugging.

beacon_chain/consensus_object_pools/block_dag.nim
tersec: This could be a good idea, but part of what this relies on is serialization and merklization just working. Adding distinct types means that all has to be found and duplicated in some sense,. potentially across multiple git repositories/submodules.

AllTests-mainnet.md
tersec: It doesn't run the keymanager API tests.

beacon_chain/conf.nim
tersec: The other part is, large chunks of this will be merged soon, so it's still useful to review, but it will contain plenty of code that's just for merge testnet debugging.

tersec:
Review from previous issue that BlockRefs can't store `execution_payload` hash roots for memory reasons, at least in the near-term most relevant to merging that functionality into `unstable`: https://github.com/status-im/nimbus-eth2/pull/3093#discussion_r783192363

With dag finalization pruning, this is less of an issue than it used to be, though it's still worth consideration.

arnetheduck:
> With dag finalization pruning, this is less of an issue than it used to be, though it's still worth consideration.

merged, add away :)

PR #3498, opened 2022-03-14
fix light client data collection for checkpoint sync

etan-status:
When doing checkpoint sync, collecting light client data of known blocks
and states incorrectly assumes that `finalized_checkpoint` information
is also known. Hardens collection to only collect finalized checkpoint
data after `dag.computeEarliestLightClientSlot`.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -146,10 +146,14 @@ proc cacheLightClientData*(
 
   template finalized_checkpoint(): auto = state.data.finalized_checkpoint
   let
+    earliestSlot = dag.computeEarliestLightClientSlot
     bid =
       BlockId(root: blck.root, slot: blck.message.slot)
     finalized_bid =
-      dag.getBlockIdAtSlot(finalized_checkpoint.epoch.start_slot).expect("TODO").bid
+      if finalized_checkpoint.epoch.start_slot >= earliestSlot:
+        dag.getBlockIdAtSlot(finalized_checkpoint.epoch.start_slot).expect("TODO").bid
+      else:
+        default(BlockId)
   if dag.lightClientCache.data.hasKeyOrPut(
       bid,
       CachedLightClientData(
@@ -788,7 +792,8 @@ proc initLightClientCache*(dag: ChainDAGRef) =
             # Note that light client finality proofs refer to checkpoint blocks
             # at their original slot, without advancing to an epoch boundary.
             # This is because light clients are unable to advance slots.
-            if checkpoint.root != dag.finalizedHead.blck.root:
+            if checkpoint.root != dag.finalizedHead.blck.root and
+                checkpoint.epoch.start_slot >= earliestSlot:
               let cpRef =
                 dag.getBlockIdAtSlot(checkpoint.epoch.start_slot).expect("TODO").bid
               if cpRef.slot >= earliestSlot:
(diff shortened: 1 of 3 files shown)


etan-status:
- Rebase to `unstable`.
- Re-enable light client test that was marked to skip on `unstable` due to the bugs fixed in this PR.

etan-status:
- Rebase to `unstable` (re-run tests)

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -146,10 +146,14 @@ proc cacheLightClientData*(
 
   template finalized_checkpoint(): auto = state.data.finalized_checkpoint
   let
+    earliestSlot = dag.computeEarliestLightClientSlot
     bid =
       BlockId(root: blck.root, slot: blck.message.slot)
     finalized_bid =
-      dag.getBlockIdAtSlot(finalized_checkpoint.epoch.start_slot).expect("TODO").bid
+      if finalized_checkpoint.epoch.start_slot >= earliestSlot:
+        dag.getBlockIdAtSlot(finalized_checkpoint.epoch.start_slot).expect("TODO").bid
```
arnetheduck, line 154: this is probably a good time to fix the todo - if we can't get a block id, is it a good idea to cache bogus values? also, what should happen if the finalized checkpoint and the finalizedHead disagree? the latter is the former, except during checkpoint sync when the checkpoint implicitly is finalized.

etan-status: `finalizedHead` is not considered except for pruning purposes. If `finalized_checkpoint` < `tail`, it is also `< earliestSlot` and this will produce a default `BlockId` with 0 slot (which is also `< earliestSlot`). That one will then be ignored by the consumer because `finalized_bid.slot < earliestSlot`.

```
    let finalized_slot = attested_data.finalized_bid.slot
    if finalized_slot + UPDATE_TIMEOUT < attested_slot or
        finalized_slot < earliest_slot:
      latest_update.finalized_header = BeaconBlockHeader()
      latest_update.finality_branch.fill(Eth2Digest())
      if signature_period == attested_period + 1:
        latest_update.next_sync_committee = SyncCommittee()
        latest_update.next_sync_committee_branch.fill(Eth2Digest())
      else:
        latest_update.next_sync_committee =
          state.data.next_sync_committee
        latest_update.next_sync_committee_branch =
          attested_data.next_sync_committee_branch
```

LC code currently assumes that blocks / states >= `earliestSlot` (taking into account `tail`) always succeed to be retrieved. Allowing every access to fail even within that range is something that would need fixing through the entire file. If it is a priority I can try and address this separate problem as part of this PR as well, or can do a cleanup one later to tackle these (and also the other TODO that got introduced for similar reasons). Only fixing this specific instance of the TODO separately seems not that clean.

arnetheduck: the code needs a full review for missing blocks and failed rewinds, including fixing such that it doesn't crash on unprotected `Opt.get` calls and the like - this would be part of fixing these tests, the expects and so on - consider `dag.tail` advisory: things should be robust against it being wrong or blocks / states accidentally going missing

etan-status:
- Rebase to `unstable`

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Yes, but the `Opt` handling is a separate issue from the one here.

The fix in this PR corrects the interaction between light client data collection and checkpoint sync.
The `Opt` handling is a hardening for unexpected missing data in the database.

Would suggest splitting those two topics, and look into `Opt` handling once the DAG changes are complete, while moving forward with the checkpoint sync fix for now (which can be hit without database corruption).

etan-status:
- Rebase to `unstable` (re-run tests)

arnetheduck (approved):

etan-status:
- Rebase to `unstable` (re-run tests)

PR #3499, opened 2022-03-14
broadcast optimistic light client updates

etan-status:
After proposing a new block, broadcasts a `OptimisticLightClientUpdate`.
Works for both locally proposed blocks as well as VC submitted ones.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -7,6 +7,10 @@
 
 {.push raises: [Defect].}
 
+# References to `vFuture` refer to the pre-release proposal of the libp2p based
+# light client sync protocol. Conflicting release versions are not in use.
+# https://github.com/ethereum/consensus-specs/pull/2802
+
 import
   # Standard library
   std/[os, osproc, sequtils, streams, tables],
@@ -737,6 +741,19 @@ proc handleSyncCommitteeMessages(node: BeaconNode, head: BlockRef, slot: Slot) =
       asyncSpawn createAndSendSyncCommitteeMessage(node, slot, validator,
                                                    subcommitteeIdx, head)
 
+proc handleOptimisticLightClientUpdates(
+    node: BeaconNode, head: BlockRef, slot: Slot) =
+  if slot < node.dag.cfg.ALTAIR_FORK_EPOCH.start_slot():
+    return
+  doAssert head.parent != nil, "Newly proposed block lacks parent reference"
+  let msg = node.dag.lightClientCache.optimisticUpdate
+  if msg.attested_header.slot != head.parent.bid.slot:
+    notice "No optimistic light client update for proposed block",
+      slot = slot, block_root = shortLog(head.root)
+    return
+  node.network.broadcastOptimisticLightClientUpdate(msg)
+  notice "Sent optimistic light client update", message = shortLog(msg)
+
 proc signAndSendContribution(node: BeaconNode,
                              validator: AttachedValidator,
                              contribution: SyncCommitteeContribution,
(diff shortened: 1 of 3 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -1276,6 +1301,22 @@ proc sendBeaconBlock*(node: BeaconNode, forked: ForkedSignedBeaconBlock
         notice "Block published",
           blockRoot = shortLog(blck.root), blck = shortLog(blck.message),
           signature = shortLog(blck.signature)
+
+        if node.config.serveLightClientData:
+          proc publishOptimisticLightClientUpdate() {.async.} =
+            let cutoff = node.beaconClock.fromNow(
+              wallTime.slotOrZero.optimistic_light_client_update_deadline())
```
arnetheduck: why should the light client wait with publish here?

etan-status: The `optimistic_light_client_update` message only validates once the block has been processed by the network. The spec includes a delay of 1/3 of a slot (same delay as for attestations) to ensure that the block was received before starting propagation of the corresponding update message.

arnetheduck: this sounds odd - the deadlines in general are there so that production of messages happens even if the block doesn't appear, not the other way around - publishing is usually done "asap", though we do introduce a small propagation delay for blocks before sending attestations

etan-status: Yeah, it is a bit different because for attestations they can be produced as soon as the block arrives locally (but they still require the block to be present). Here, we are sending both the block and the object depending on the block, from the same entity, assuming the worst case in block propagation speed that the spec assumes for attestations (if it does not arrive within 1/3 of the slot, assume slot is empty and send attestation regardless of presence).

In the end, light client sync is always 1 slot behind, so being 4/3 slots behind is not a huge loss. But if there is a more ideomatic way of handling this, we can also send earlier. As long as the block arrives before the optimistic update, things go well.

arnetheduck: well, at the very least we need to keep the terminology apart since deadline in the codebase a timeout while waiting for an event - here instead we're delaying the send so call it .. delay or offset or something?

etan-status:
- Added comment with rationale for delaying `OptimisticLightClientUpdate` and spec reference.
- Rename `optimistic_light_client_update_deadline` > `optimistic_light_client_update_time` to denote the fixed time at which the `OptimisticLightClientUpdate` is sent.

beacon_chain/validators/validator_duties.nim
```diff
@@ -7,6 +7,10 @@
 
 {.push raises: [Defect].}
 
+# References to `vFuture` refer to the pre-release proposal of the libp2p based
```
arnetheduck, line 10: can also link to a specific hash in that repo..

etan-status: It's still an open PR that keeps getting updated, so the hash keeps changing. Deep-linking into "latest version on PR" does not seem to work. And referring to my temporary personal specs fork while the PR is open could work but seems weird as well.

beacon_chain/validators/validator_duties.nim
```diff
@@ -737,6 +741,18 @@ proc handleSyncCommitteeMessages(node: BeaconNode, head: BlockRef, slot: Slot) =
       asyncSpawn createAndSendSyncCommitteeMessage(node, slot, validator,
                                                    subcommitteeIdx, head)
 
+proc handleOptimisticLightClientUpdates(
+    node: BeaconNode, head: BlockRef, slot: Slot) =
+  if slot < node.dag.cfg.ALTAIR_FORK_EPOCH.start_slot():
+    return
+  let msg = node.dag.lightClientCache.optimisticUpdate
```
arnetheduck, line 749: this will crash when `head == tail` on a checkpoint sync - unlikely due to timing in other parts of the app, but still.. - in general, all `parent` accesses must be nil-checked (I have half of a plan / idea to make the field private)

arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
etan-status: This function is only called after we (or one of our VCs) submitted a new block, making that impossible to reach.
But sure, can add an assert for non-nil parent with a comment.

etan-status:
- Assert existing `parent` reference for newly produced block.

PR #3505, opened 2022-03-15
support for dumping light client data

etan-status:
Adds support for dumping light client data to files, similar to existing
dump functionality for other data received over the network.


arnetheduck (approved):
likewise here - we strive to intro functionality together with the code that actually uses it - it's hard to evaluate the total impact of .. dead code

etan-status:
Alright, will resubmit together with user, it seems small enough to not add too much noise.

PR #3506, opened 2022-03-15
define type for "any light client object"

etan-status:
Defines a type `SomeLightClientObject` to allow functions to accept any
light client related object.

### DIFF
--- a/beacon_chain/spec/datatypes/altair.nim
+++ b/beacon_chain/spec/datatypes/altair.nim
@@ -207,6 +207,11 @@ type
     is_signed_by_next_sync_committee*: bool ##\
     ## Whether the signature was produced by `attested_header`'s next sync committee
 
+  SomeLightClientObject* =
+    LightClientBootstrap |
+    LightClientUpdate |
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/datatypes/altair.nim
```diff
@@ -207,6 +207,11 @@ type
     is_signed_by_next_sync_committee*: bool ##\
     ## Whether the signature was produced by `attested_header`'s next sync committee
 
+  SomeLightClientObject* =
```
arnetheduck, line 210: eh, maybe as part of a pr with a motivating example?

tersec: Agree, this is a bit too free of motivating context. I don't especially doubt that a typeclass like this could be useful, but would at least one user would be nice.

etan-status: Alright, it's small enough to not add too much noise as part of a PR that also includes a user. Closing this one and will submit again together with a user.

PR #3507, opened 2022-03-15
remove StateData

arnetheduck:
One more step on the journey to reduce `BlockRef` usage across the
codebase - this one gets rid of `StateData` whose job was to keep track
of which block was last assigned to a state - these duties have now been
taken over by `latest_block_root`, a fairly recent addition that
computes this block root from state data (at a small cost that should be
insignificant)

99% mechanical change.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -10,7 +10,7 @@
 import
   chronicles,
   stew/[assign2, results],
-  ../spec/[forks, signatures, signatures_batch, state_transition],
+  ../spec/[beaconstate, forks, signatures, signatures_batch, state_transition],
   "."/[block_dag, blockchain_dag, blockchain_dag_light_client]
 
 export results, signatures_batch, block_dag, blockchain_dag
@@ -27,16 +27,15 @@ logScope:
 
 proc addResolvedHeadBlock(
        dag: ChainDAGRef,
-       state: var StateData,
+       state: var ForkedHashedBeaconState,
        trustedBlock: ForkyTrustedSignedBeaconBlock,
        parent: BlockRef, cache: var StateCache,
        onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded | OnBellatrixBlockAdded,
        stateDataDur, sigVerifyDur, stateVerifyDur: Duration
      ): BlockRef =
-  doAssert getStateField(state.data, slot) == trustedBlock.message.slot,
-    "state must match block"
-  doAssert state.blck.root == trustedBlock.message.parent_root,
-    "the StateData passed into the addResolved function not yet updated!"
+  doAssert state.matches_block_slot(
+    trustedBlock.root, trustedBlock.message.slot),
+    "Given state must have the new block applied"
 
   let
     blockRoot = trustedBlock.root
@@ -63,17 +62,16 @@ proc addResolvedHeadBlock(
 
   # Up to here, state.data was referring to the new state after the block had
   # been applied but the `blck` field was still set to the parent
-  state.blck = blockRef
+  dag.clearanceBlck = blockRef
 
   # Regardless of the chain we're on, the deposits come in the same order so
   # as soon as we import a block, we'll also update the shared public key
   # cache
-
-  dag.updateValidatorKeys(getStateField(state.data, validators).asSeq())
+  dag.updateValidatorKeys(getStateField(state, validators).asSeq())
 
   # Getting epochRef with the state will potentially create a new EpochRef
   let
-    epochRef = dag.getEpochRef(state, cache)
+    epochRef = dag.getEpochRef(state, blockRef, cache)
     epochRefTick = Moment.now()
 
   debug "Block resolved",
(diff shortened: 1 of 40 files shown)


tersec (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -690,22 +690,22 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
     # Now that we have a head block, we need to find the most recent state that
     # we have saved in the database
     while cur.blck != nil and
-        not getStateData(db, cfg, dag.headState, cur, noRollback):
+        not getState(db, cfg, dag.headState, cur, noRollback):
       cur = cur.parentOrSlot()
 
-    if dag.headState.blck == nil:
+    if cur.blck == nil:
```
etan-status, line 696: - `headRef`?

arnetheduck: no, we want to capture the condition where `getState` failed for the full history - previously that would be seen in `headState.blck` itself

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1057,27 +1057,25 @@ proc applyBlock(
...
 proc applyBlock(
     dag: ChainDAGRef,
-    state: var StateData, blck: BlockRef,
+    state: var ForkedHashedBeaconState, blck: BlockRef,
     cache: var StateCache, info: var ForkedEpochInfo) =
   # Apply a single block to the state - the state must be positioned at the
   # parent of the block with a slot lower than the one of the block being
   # applied
-  doAssert state.blck == blck.parent
+  doAssert state.matches_block(blck.parent.root)
```
etan-status, line 1065: and `blck.parent` is guaranteed to be `nil`, assuming the comment is right (otherwise potential nil access)

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1093,21 +1091,26 @@ proc updateStateData*(
 
   let
     startTick = Moment.now()
-    current {.used.} = state.blck.atSlot(getStateField(state.data, slot))
+    current {.used.} = withState(state):
+      BlockSlotId.init(
+        BlockId(
+          root: state.latest_block_root,
+          slot: state.data.latest_block_header.slot),
+        state.data.slot)
```
etan-status, line 1099: maybe worth to introduce a `toBlockSlotId(state)` helper

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -756,11 +756,11 @@ proc initLightClientCache*(dag: ChainDAGRef) =
     cpIndex = 0
   for i in countdown(blocksBetween.high, blocksBetween.low):
     blockRef = blocksBetween[i]
-    doAssert dag.updateStateData(
+    doAssert dag.updateState(
       dag.headState, blockRef.atSlot(blockRef.slot), save = false, cache)
     withStateVars(dag.headState):
-      let bdata = dag.getForkedBlock(blck.bid).get
-      withStateAndBlck(stateData.data, bdata):
+      let bdata = dag.getForkedBlock(dag.head.bid).get
```
etan-status: Hmm. `dag.headState` is only borrowed here to avoid an extra state copy. This worked because it ends up being restored to `head`. But `dag.head` is no longer a template into `dag.headState` so this will not be accurate anymore.  I guess `blockRef.atSlot(blockRef.slot)` needs to be assigned to a let and referred to here.
- [ ] X

tests/test_attestation_pool.nim
```diff
@@ -73,15 +73,16 @@ suite "Attestation pool processing" & preset():
...
 
   test "Can add and retrieve simple attestations" & preset():
     let
       # Create an attestation for slot 1!
       bc0 = get_beacon_committee(
-        state[].data, getStateField(state.data, slot), 0.CommitteeIndex, cache)
-      attestation = makeAttestation(state[].data, state.blck.root, bc0[0], cache)
+        state[], getStateField(state[], slot), 0.CommitteeIndex, cache)
+      attestation = makeAttestation(
+        state[], state[].latest_block_root(), bc0[0], cache)
```
etan-status: NIT: In the other test you use `latest_block_root` without `()` syntax.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -222,16 +216,17 @@ proc addHeadBlock*(
...
+  if not updateState(
       dag, dag.clearanceState, parent.atSlot(signedBlock.message.slot), true,
       cache):
     # We should never end up here - the parent must be a block no older than and
     # rooted in the finalized checkpoint, hence we should always be able to
     # load its corresponding state
     error "Unable to load clearance state for parent block, database corrupt?",
       parent = shortLog(parent.atSlot(signedBlock.message.slot)),
-      clearance = shortLog(dag.clearanceState.blck)
+      clearanceBlock = shortLog(dag.clearanceBlck)
```
etan-status, line 227: I guess the user-facing change to the log message is intended, just highlighting to ensure.

etan-status (approved):

beacon_chain/consensus_object_pools/block_clearance.nim
arnetheduck: yep :+1:

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: yeah, in a future pr that shuffles blockid around

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: yes - the previous assert would not have worked on a `nil` either - future PR:s that get rid of blockref will change this assert as well to something more wholesome

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
arnetheduck: 12ea008311122a28914f3c93ee8d72910007aea5

tests/test_attestation_pool.nim
arnetheduck: 12ea008311122a28914f3c93ee8d72910007aea5

mratsim (approved):

PR #3508, opened 2022-03-15
update to latest light client spec

etan-status:
Adds the additional check to ensure `optimistic_header` is always after
`finalized_header` in `LightClientStore`, as introduced to the spec in
https://github.com/ethereum/consensus-specs/pull/2814


tersec (approved):
Well it's a direct reflection of your approved and merged consensus specs PR, basically the same two lines of code in each, so, sure.

Not sure if it makes sense to wait for the StateData removal PR to go ini since there was indicated a preference for anything not time-critical not to go in before it does.

etan-status:
Sure, the only usage of this code is in testing. This does not touch `StateData` at all though, so both are fine. Let's err on waiting on the other PR as there is nothing to lose.

PR #3509, opened 2022-03-15
introduce light client processor

etan-status:
Adds `LightClientProcessor` as the pendant to `BlockProcessor` while
operating in light client mode. Note that a similar mechanism based on
async futures is used for interoperability with existing infrastructure,
despite light client object validation being done synchronously.

### DIFF
--- a/beacon_chain/gossip_processing/light_client_processor.nim
+++ b/beacon_chain/gossip_processing/light_client_processor.nim
@@ -1,0 +1,280 @@
+# beacon_chain
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  stew/objects,
+  chronos, metrics,
+  ../spec/datatypes/altair,
+  ../spec/light_client_sync,
+  ../consensus_object_pools/block_pools_types,
+  ".."/[beacon_clock],
+  ../sszdump
+
+export sszdump
+
+# Light Client Processor
+# ------------------------------------------------------------------------------
+# The light client processor handles received light client objects
+
+declareHistogram light_client_store_object_duration_seconds,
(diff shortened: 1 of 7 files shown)


etan-status:
- Rebase to `unstable` (re-run tests)

beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -0,0 +1,279 @@
...
+  ## all untrusted objects pass through here. When storing an object, we will
+  ## update the `LightClientStore` accordingly
+  let
+    startTick = Moment.now()
+    wallSlot = wallTime.slotOrZero()
+    store = self.store
+
+    res =
+      when obj is altair.LightClientBootstrap:
+        if store[].isSome:
```
arnetheduck, line 160: could junk bootstraps be used to spam the client? if so, shouldn't some sort of validation happen here with and invalid response so the peer can be descored?

beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -0,0 +1,279 @@
...
+    else:
+      obj.attested_header.slot
+  debug "Light client object processed", kind = typeof(obj).name,
+    localHeadSlot = store[].get.optimistic_header.slot,
+    objectSlot = objSlot,
+    storeObjectDur
+
+  when obj is altair.LightClientBootstrap:
+    if self.didInitializeStoreCallback != nil:
+      self.didInitializeStoreCallback()
```
arnetheduck, line 246: set to nil to release mem and prevent multiple calls?

beacon_chain/gossip_processing/light_client_processor.nim
etan-status: Bootstraps can only be obtained using a client-initiated request (Req/Resp). Because the request manager sends requests out concurrently to multiple peers, this is just to discard any bootstraps that arrived late.

beacon_chain/gossip_processing/light_client_processor.nim
etan-status: Multiple calls are not possible, as any sub-sequent bootstraps lead to `BlockError.Duplicate`.
The reset to `nil` is a good idea to clear potential references to captured variables by the callback.

etan-status:
- Clear callback after its last use.

arnetheduck (approved):

etan-status:
- Rebase to `unstable` (remove `StateData` in new tests, and re-run tests)

PR #3513, opened 2022-03-16
Prune `BlockRef` on finalization

arnetheduck:
Up til now, the block dag has been using `BlockRef`, a structure adapted
for a full DAG, to represent all of chain history. This is a correct and
simple design, but does not exploit the linearity of the chain once
parts of it finalize.

By pruning the in-memory `BlockRef` structure at finalization, we save,
at the time of writing, a cool ~250mb (or 25%:ish) chunk of memory
landing us at a steady state of ~750mb normal memory usage for a
validating node.

Above all though, we prevent memory usage from growing proportionally
with the length of the chain, something that would not be sustainable
over time -  instead, the steady state memory usage is roughly
determined by the validator set size which grows much more slowly. With
these changes, the core should remain sustainable memory-wise post-merge
all the way to withdrawals (when the validator set is expected to grow).

In-memory indices are still used for the "hot" unfinalized portion of
the chain - this ensure that consensus performance remains unchanged.

What changes is that for historical access, we use a db-based linear
slot index which is cache-and-disk-friendly, keeping the cost for
accessing historical data at a similar level as before, achieving the
savings at no percievable cost to functionality or performance.

A nice collateral benefit is the almost-instant startup since we no
longer load any large indicies at dag init.

The cost of this functionality instead can be found in the complexity of
having to deal with two ways of traversing the chain - by `BlockRef` and
by slot.

* use `BlockId` instead of `BlockRef` where finalized / historical data
may be required
* simplify clearance pre-advancement
* remove dag.finalizedBlocks (~50:ish mb)
* remove `getBlockAtSlot` - use `getBlockIdAtSlot` instead
* `parent` and `atSlot` for `BlockId` now require a `ChainDAGRef`
instance, unlike `BlockRef` traversal
* prune `BlockRef` parents on finality (~200:ish mb)
* speed up ChainDAG init by not loading finalized history index
* mess up light client server error handling - this need revisiting :)
* 
Screenshot from 2022-03-11 17-34-08

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -248,6 +248,7 @@ proc getBlockId*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockId] =
   err()
 
 func isCanonical*(dag: ChainDAGRef, bid: BlockId): bool =
+  ## Return true iff the given `bid` is part of the history selected by `head`
   let current = dag.getBlockIdAtSlot(bid.slot).valueOr:
     return false # We don't know, so ..
   return current.bid == bid
@@ -257,9 +258,8 @@ func parent*(dag: ChainDAGRef, bid: BlockId): Opt[BlockId] =
     return err()
 
   if bid.slot > dag.finalizedHead.slot:
-    # Have to take forks into consideration
-    let blck = dag.getBlockRef(bid.root).valueOr:
-      return err()
+    # Make sure we follow the correct history as there may be forks
+    let blck = ? dag.getBlockRef(bid.root)
 
     doAssert not isNil(blck.parent), "should reach finalized head"
     return ok blck.parent.bid
@@ -272,23 +272,21 @@ func parentOrSlot*(dag: ChainDAGRef, bsi: BlockSlotId): Opt[BlockSlotId] =
     return err()
 
   if bsi.isProposed:
-    let parent = dag.parent(bsi.bid).valueOr:
-      return err()
+    let parent = ? dag.parent(bsi.bid)
     ok BlockSlotId.init(parent, bsi.slot)
   else:
     ok BlockSlotId.init(bsi.bid, bsi.slot - 1)
 
 func atSlot*(dag: ChainDAGRef, bid: BlockId, slot: Slot): Opt[BlockSlotId] =
   if bid.slot > dag.finalizedHead.slot:
-    let blck = dag.getBlockRef(bid.root).valueOr:
-      return err() # Not part of known / relevant history
+    let blck = ? dag.getBlockRef(bid.root)
 
     if slot > dag.finalizedHead.slot:
       return blck.atSlot(slot).toBlockSlotId()
 
-  let existing = dag.getBlockIdAtSlot(bid.slot).valueOr:
-    return err() # Not part of known / relevant history
-
+  # Check if the given `bid` is still part of history - it might hail from an
+  # orphaned fork
+  let existing = ? dag.getBlockIdAtSlot(bid.slot)
   if existing.bid != bid:
     return err() # Not part of known / relevant history
 
@@ -307,7 +305,8 @@ func epochAncestor*(dag: ChainDAGRef, bid: BlockId, epoch: Epoch): EpochKey =
     return EpochKey(bid: dag.genesis, epoch: epoch)
 
   let bsi = dag.atSlot(bid, epoch.start_slot - 1).valueOr:
-    # Worst case, an ancestor is always its own key
+    # If we lack history for the given slot, we can use the given bid as epoch
+    # ancestor
     return EpochKey(epoch: epoch, bid: bid)
 
   EpochKey(epoch: epoch, bid: bsi.bid)
@@ -370,7 +369,7 @@ func isStateCheckpoint(dag: ChainDAGRef, bsi: BlockSlotId): bool =
   # there is only a checkpoint for the first epoch after a block.
 
   # The tail block also counts as a state checkpoint!
-  bsi.isProposed and bsi.bid == dag.tail or
+  (bsi.isProposed and bsi.bid == dag.tail) or
   (bsi.slot.is_epoch and bsi.slot.epoch == (bsi.bid.slot.epoch + 1))
 
 proc getState(
@@ -436,8 +435,7 @@ proc getForkedBlock*(db: BeaconChainDB, root: Eth2Digest):
 proc getBlock*(
     dag: ChainDAGRef, bid: BlockId,
     T: type ForkyTrustedSignedBeaconBlock): Opt[T] =
-  withState(dag.headState):
-    dag.db.getBlock(bid.root, T)
+  dag.db.getBlock(bid.root, T)
 
 proc getBlockSSZ*(dag: ChainDAGRef, bid: BlockId, bytes: var seq[byte]): bool =
   # Load the SSZ-encoded data of a block into `bytes`, overwriting the existing
@@ -554,13 +552,11 @@ proc advanceSlots*(
   # target
   doAssert getStateField(state, slot) <= slot
 
-  let stateBid = BlockId(
-    root: state.latest_block_root,
-    slot: getStateField(state, latest_block_header).slot)
-
+  let stateBid = state.latest_block_id
   while getStateField(state, slot) < slot:
     let
       preEpoch = getStateField(state, slot).epoch
+
     loadStateCache(dag, cache, stateBid, getStateField(state, slot).epoch)
 
     process_slots(
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -248,7 +248,7 @@ proc getBlockId*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockId] =
   err()
 
 func isCanonical*(dag: ChainDAGRef, bid: BlockId): bool =
-  ## Return true iff the given `bid` is part of the history selected by `head`
+  ## Return true iff the given `bid` is part of the history selected by `dag.head`
   let current = dag.getBlockIdAtSlot(bid.slot).valueOr:
     return false # We don't know, so ..
   return current.bid == bid
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -57,8 +57,8 @@ proc putBlock*(
   dag.db.putBlock(signedBlock)
 
 proc updateState*(
-  dag: ChainDAGRef, state: var ForkedHashedBeaconState, bs: BlockSlot, save: bool,
-  cache: var StateCache): bool {.gcsafe.}
+    dag: ChainDAGRef, state: var ForkedHashedBeaconState, bsi: BlockSlotId, save: bool,
+    cache: var StateCache): bool {.gcsafe.}
 
 template withStateVars*(
     stateInternal: var ForkedHashedBeaconState, body: untyped): untyped =
@@ -72,14 +72,15 @@ template withStateVars*(
 
 template withUpdatedState*(
     dag: ChainDAGRef, state: var ForkedHashedBeaconState,
-    blockSlot: BlockSlot, okBody: untyped, failureBody: untyped): untyped =
+    bsiParam: BlockSlotId, okBody: untyped, failureBody: untyped): untyped =
   ## Helper template that updates stateData to a particular BlockSlot - usage of
   ## stateData is unsafe outside of block, or across `await` boundaries
 
   block:
+    let bsi {.inject.} = bsiParam
     var cache {.inject.} = StateCache()
-    if updateState(dag, state, blockSlot, false, cache):
-      template blck(): BlockRef {.inject, used.} = blockSlot.blck
+    if updateState(dag, state, bsi, false, cache):
+      template bid(): BlockId {.inject, used.} = bsi.bid
 
       withStateVars(state):
         okBody
@@ -105,17 +106,12 @@ proc updateValidatorKeys*(dag: ChainDAGRef, validators: openArray[Validator]) =
   # data (but no earlier than that the whole block as been validated)
   dag.db.updateImmutableValidators(validators)
 
-proc updateFinalizedBlocks*(dag: ChainDAGRef) =
-  if dag.db.db.readOnly: return # TODO abstraction leak - where to put this?
+proc updateFinalizedBlocks*(db: BeaconChainDB, newFinalized: openArray[BlockId]) =
+  if db.db.readOnly: return # TODO abstraction leak - where to put this?
 
-  dag.db.withManyWrites:
-    let high = dag.db.finalizedBlocks.high.expect(
-      "wrote at least tailRef during init")
-
-    for s in high + 1 .. dag.finalizedHead.slot:
-      let tailIdx = int(s - dag.tail.slot)
-      if not isNil(dag.finalizedBlocks[tailIdx]):
-        dag.db.finalizedBlocks.insert(s, dag.finalizedBlocks[tailIdx].root)
+  db.withManyWrites:
+    for bid in newFinalized:
+      db.finalizedBlocks.insert(bid.slot, bid.root)
 
 func validatorKey*(
     dag: ChainDAGRef, index: ValidatorIndex or uint64): Option[CookedPubKey] =
@@ -135,14 +131,14 @@ func validatorKey*(
 
 func init*(
     T: type EpochRef, dag: ChainDAGRef, state: ForkedHashedBeaconState,
-    blck: BlockRef, cache: var StateCache): T =
+    cache: var StateCache): T =
   let
     epoch = state.get_current_epoch()
     proposer_dependent_root = withState(state): state.proposer_dependent_root
     attester_dependent_root = withState(state): state.attester_dependent_root
     epochRef = EpochRef(
       dag: dag, # This gives access to the validator pubkeys through an EpochRef
-      key: blck.epochAncestor(epoch),
+      key: dag.epochAncestor(state.latest_block_id, epoch),
       eth1_data: getStateField(state, eth1_data),
       eth1_deposit_index: getStateField(state, eth1_deposit_index),
       current_justified_checkpoint:
@@ -162,8 +158,6 @@ func init*(
             ExecutionPayloadHeader()
     )
     epochStart = epoch.start_slot()
-
-  doAssert epochRef.key.blck != nil, "epochAncestor should not fail for state block"
 
   for i in 0'u64..<SLOTS_PER_EPOCH:
     epochRef.beacon_proposers[i] = get_beacon_proposer_index(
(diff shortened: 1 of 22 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -363,41 +370,52 @@ func isStateCheckpoint(bs: BlockSlot): bool =
...
+  if not dag.isStateCheckpoint(bsi):
     return false
 
-  true
+  let rollbackAddr =
+    # Any restore point will do as long as it's not the object being updated
+    if unsafeAddr(state) == unsafeAddr(dag.headState):
+      unsafeAddr dag.clearanceState
+    else:
+      unsafeAddr dag.headState
```
mratsim, line 417: This may introduce weird bugs in a couple months if we ever introduce "ForkedBeaconState" because a hard fork requires us to.

Though at that point we will likely refactor all and see that unsafe use without going through heisenbug head-scratching hunting process.

arnetheduck: not sure what you mean - both of these are `ForkedHashedBeaconState` objects - all this function needs to do is return them to _any_ valid / consistent beacon state - on error, SSZ will leave a partially initialized state and we want to avoid letting those partials live in memory

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -278,11 +248,51 @@ proc getBlockId*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockId] =
   err()
 
 func isCanonical*(dag: ChainDAGRef, bid: BlockId): bool =
+  ## Return true iff the given `bid` is part of the history selected by `head`
```
etan-status: NIT: `dag.head` instead of `head`?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -206,43 +200,19 @@ func getBlockRef*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockRef] =
...
-
-      pos = pos - 1
-
-  err() # Not found
-
 func getBlockIdAtSlot*(dag: ChainDAGRef, slot: Slot): Opt[BlockSlotId] =
   ## Retrieve the canonical block at the given slot, or the last block that
   ## comes before - similar to atSlot, but without the linear scan - may hit
   ## the database to look up early indices.
+  if slot == dag.genesis.slot:
```
etan-status, line 207: maybe an early reject for `slot < dag.genesis.slot`. existing logic should handle it fine though (if it's even possible. lots of code just assumes `GENESIS_SLOT == 0` and compares to 0 instead of the constant anyway.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -278,11 +248,51 @@ proc getBlockId*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockId] =
...
+    let blck = ? dag.getBlockRef(bid.root)
+
+    if slot > dag.finalizedHead.slot:
+      return blck.atSlot(slot).toBlockSlotId()
+
+  # Check if the given `bid` is still part of history - it might hail from an
+  # orphaned fork
+  let existing = ? dag.getBlockIdAtSlot(bid.slot)
+  if existing.bid != bid:
+    return err() # Not part of known / relevant history
```
etan-status: If `bid.slot > dag.finalizedHead.slot`, I am not sure if this check should be enforced. When `bid` is from a non-finalized fork that is not currently selected by fork-choice, `dag.getBlockIdAtSlot` for `> dag.finalizedHead.slot` could return a different `bid`, even though the fork may become relevant again later?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -291,25 +301,21 @@ func epochAncestor*(blck: BlockRef, epoch: Epoch): EpochKey =
...
-  while blck.slot.epoch >= epoch and not blck.parent.isNil:
-    blck = blck.parent
+  if bid.slot.epoch == dag.genesis.slot.epoch:
+    return EpochKey(bid: dag.genesis, epoch: epoch)
 
-  if blck.slot.epoch > epoch:
-    EpochKey() # The searched-for epoch predates our tail block
-  else:
-    EpochKey(epoch: epoch, blck: blck)
+  let bsi = dag.atSlot(bid, epoch.start_slot - 1).valueOr:
```
etan-status, line 313: Potentially underflow when passing epoch 0 with a `bid` from a later epoch. Not sure if reachable by callers (function is exported)

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -490,302 +506,303 @@ export
...
 
+proc getViableHead(cfg: RuntimeConfig, db: BeaconChainDB): Opt[BlockId] =
+  # When the database has been written with a pre-fork version of the
+  # software, it may happen that blocks produced using an "unforked"
+  # chain get written to the database - we need to skip such blocks
+  # when loading the database with a fork-compatible version
+  let
+    headRoot = ? db.getHeadBlock()
+
+  for blck in db.getAncestorSummaries(headRoot):
```
etan-status, line 523: NIT: This returns `RootedSummary`, `blck` is a bit misleading varname here.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -490,302 +506,303 @@ export
...
+  withState(state):
+    dag.db.putState(state)
+
+  debug "Stored state", putStateDur = Moment.now() - startTick
+
+proc advanceSlots*(
+    dag: ChainDAGRef, state: var ForkedHashedBeaconState, slot: Slot, save: bool,
+    cache: var StateCache, info: var ForkedEpochInfo) =
+  # Given a state, advance it zero or more slots by applying empty slot
+  # processing - the state must be positions at a slot before or equal to the
```
etan-status: - `positions` -> `positioned`
- "at a slot before or equal to the target" -> "at or before \`slot\`"

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -490,302 +506,303 @@ export
...
+    state_transition(
+      dag.cfg, state, data, cache, info,
+      dag.updateFlags + {slotProcessed}, noRollback)
+  of BeaconBlockFork.Altair:
+    let data = getBlock(dag, bid, altair.TrustedSignedBeaconBlock).valueOr:
+      return err("Block load failed")
+    state_transition(
+      dag.cfg, state, data, cache, info,
+      dag.updateFlags + {slotProcessed}, noRollback)
+  of BeaconBlockFork.Bellatrix:
```
etan-status, line 602: NIT: forks in order from latest to older so that the if tree is slightly more efficient if one is generated.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -490,302 +506,303 @@ export
...
-      if midRef == nil:
-        fatal "First finalized block beyond tail, database corrupt?",
-          tail = shortLog(tailRef), slot, root = shortLog(root)
-        quit 1
+    headRef, curRef: BlockRef
+    slot = head.slot
+    # To know the finalized checkpoint of the head, we need to recreate its
+    # state - the tail is implicitly finalized, and if we have a finalized block
+    # table, that provides another hint
+    finalizedSlot = db.finalizedblocks.high.get(dag.tail.slot)
```
etan-status: NIT: capital `B` in `finalizedBlocks`

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1004,86 +987,21 @@ proc getBlockRange*(
...
-
 proc updateState*(
-    dag: ChainDAGRef, state: var ForkedHashedBeaconState, bs: BlockSlot,
+    dag: ChainDAGRef, state: var ForkedHashedBeaconState, bsi: BlockSlotId,
     save: bool, cache: var StateCache): bool =
   ## Rewind or advance state such that it matches the given block and slot -
   ## this may include replaying from an earlier snapshot if blck is on a
   ## different branch or has advanced to a higher slot number than slot
-  ## If `bs.slot` is higher than `bs.blck.slot`, `updateState` will fill in
+  ## If `bs.slot` is higher than `bs.blck.slot`, `updateStateData` will fill in
```
etan-status: It's still called `updateState`

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -864,35 +880,44 @@ proc getEpochRef*(
...
   ## The first ancestor BlockSlot that is a state checkpoint
-  var bs = bs
-  while not isStateCheckPoint(bs):
-    bs = bs.parentOrSlot
-  bs
+  var bsi = bsi
+  while not dag.isStateCheckPoint(bsi):
+    if bsi.isProposed:
+      bsi.bid = dag.parent(bsi.bid).valueOr:
+        break
```
etan-status, line 922: Is the end result here correct in case of early break?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1004,86 +987,21 @@ proc getBlockRange*(
...
+  ## If `bs.slot` is higher than `bs.blck.slot`, `updateStateData` will fill in
   ## with empty/non-block slots
 
   # First, see if we're already at the requested block. If we are, also check
   # that the state has not been advanced past the desired block - if it has,
   # an earlier state must be loaded since there's no way to undo the slot
   # transitions
 
-  if isNil(bs.blck):
+  if isZero(bsi.bid.root):
```
etan-status: Are there valid cases how a `zero` root may be produced? The new accessors seem to throw errors in cases where a block is unavailable.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1183,28 +1103,34 @@ proc updateState*(
...
       # the state we have can't trivially be advanced (in case it was older than
       # RewindBlockThreshold), keep looking..
       if cur.isProposed():
         # This is not an empty slot, so the block will need to be applied to
         # eventually reach bs
-        ancestors.add(cur.blck)
+        ancestors.add(cur.bid)
 
-      if cur.slot == dag.tail.slot or
+      if cur.slot == GENESIS_SLOT or
```
etan-status, line 1115: comment below refers to tail, code to genesis. intended change?

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -199,17 +199,13 @@ proc createLightClientUpdates(
...
   ## Create `LightClientUpdate` and `OptimisticLightClientUpdate` instances for
   ## a given block and its post-state, and keep track of best / latest ones.
   ## Data about the parent block's post-state and its `finalized_checkpoint`'s
   ## block's post-state needs to be cached (`cacheLightClientData`) before
   ## calling this function.
 
-  # Parent needs to be known to continue
-  if parent == nil:
-    return
-
```
etan-status, line 212: Hmm, which one is passed as parent when only a single block is in the DB?

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -540,37 +536,43 @@ proc initBestLightClientUpdateForPeriod(
...
     var
       maxParticipants = 0
-      maxBlockRef: BlockRef
+      maxBlockRef: Opt[BlockId]
       blockRef = highBlck
-    while blockRef.parent != nil and blockRef.parent.slot >= lowSlot:
+    while true:
+      let parent = dag.parent(blockRef).valueOr:
+        break
+      if parent.slot >= lowSlot:
```
etan-status: `<`, no?

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -540,37 +536,43 @@ proc initBestLightClientUpdateForPeriod(
...
-  proc maxParticipantsBlock(highBlck: BlockRef, lowSlot: Slot): BlockRef =
+  proc maxParticipantsBlock(
+      dag: ChainDAGRef, highBlck: BlockId, lowSlot: Slot): Opt[BlockId] =
     ## Determine the earliest block with most sync committee signatures among
     ## ancestors of `highBlck` with at least `lowSlot` as parent block slot.
     ## Return `nil` if no block with `MIN_SYNC_COMMITTEE_PARTICIPANTS` is found.
     var
       maxParticipants = 0
-      maxBlockRef: BlockRef
+      maxBlockRef: Opt[BlockId]
```
etan-status, line 546: Those are now `bid` no longer `blockRef` (varnames)

beacon_chain/rpc/rest_utils.nim
```diff
@@ -60,32 +60,32 @@ proc getCurrentHead*(node: BeaconNode,
...
   of StateQueryKind.Named:
     case stateIdent.value
     of StateIdentType.Head:
-      ok(node.dag.head.atSlot())
+      ok(node.dag.head.bid.atSlot())
     of StateIdentType.Genesis:
       ok(node.dag.genesis.atSlot())
     of StateIdentType.Finalized:
-      ok(node.dag.finalizedHead)
+      ok(node.dag.finalizedHead.toBlockSlotId().expect("non-nil"))
```
etan-status: NIT: Some of these use `non-nil`, others `not nil` in the `expect`.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -131,17 +131,17 @@ proc disallowInterruptionsAux(body: NimNode) =
 macro disallowInterruptions(body: untyped) =
   disallowInterruptionsAux(body)
 
-template withStateForBlockSlot*(nodeParam: BeaconNode,
-                                blockSlotParam: BlockSlot,
+template withStateForBlockSlotId*(nodeParam: BeaconNode,
+                                blockSlotIdParam: BlockSlotId,
```
etan-status: NIT: indentation needs re-alignment.

ncli/ncli_common.nim
```diff
@@ -110,18 +110,18 @@ func getFilePathForEpochs*(startEpoch, endEpoch: Epoch, dir: string): string =
...
-        result.add current
-    current = current.parent
+  result = newSeqOfCap[BlockId](ends - start)
+  var current = ends
+  while current > start:
+    current -= 1
+    let bsid = dag.getBlockIdAtSlot(current).valueOr:
+      continue
+
+    result.add(bsid.bid)
```
etan-status, line 126: Only if `bsid.bid.slot >= start`, no? If there are empty slot it could already be out of range.

etan-status (approved):
I did one review pass through the changes. Will take a more through look through the light client changes once this PR has been merged.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: Very early versions of the spec had GENESIS_SLOT > 0, so having these checks against non-zero is more of a path-dependent outcome than any explicit desire to support non-zero genesis - there's a mix of 0, GENESIS_SLOT and other ways of checking genesis indeed and maybe it would make sense to clean that up into something more coherent but that's a story for a different PR

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck:
```suggestion
  ## Return true iff the given `bid` is part of the history selected by `dag.head`
```

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: this is how we catch underflows

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: as correct as it gets - state checkpoints are needed to load history which requires an unbroken parent chain - the first parent to fail is the first place where a state is needed and therefore a good checkpoint candidate (ie we shouldn't remove it from storage)

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
arnetheduck: yeah, I left most renames for another day, there's enough noise in here already

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: good catch, 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: likewise, 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: not the right PR for it (this is a code move)

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/rpc/rest_utils.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

beacon_chain/rpc/rest_utils.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

ncli/ncli_common.nim
arnetheduck: 38cfc2ee383ea0d9ed52bc408539519597efdd94

PR #3514, opened 2022-03-17
update light client server for DAG failure modes

etan-status:
Gracefully handles the new failure modes recently introduced to the DAG
as part of https://github.com/status-im/nimbus-eth2/pull/3513
Data that is deemed to exist but fails to load leads to an error log to
avoid suppressing logic errors accidentally. In `verifyFinalization`
mode, the assertions remain active.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -79,7 +79,7 @@ proc addResolvedHeadBlock(
     epochRefDur = epochRefTick - putBlockTick
 
   # Update light client data
-  dag.processNewBlockForLightClient(state, trustedBlock, parent)
+  dag.processNewBlockForLightClient(state, trustedBlock, parent.bid)
 
   # Notify others of the new block before processing the quarantine, such that
   # notifications for parents happens before those of the children
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -58,10 +58,67 @@ func computeEarliestLightClientSlot*(dag: ChainDAGRef): Slot =
   let earliestSlot = currentSlot - MIN_SLOTS_FOR_BLOCK_REQUESTS
   max(earliestSlot.sync_committee_period.start_slot, minSupportedSlot)
 
+proc updateExistingState(
+    dag: ChainDAGRef, state: var ForkedHashedBeaconState, bsi: BlockSlotId,
+    save: bool, cache: var StateCache): bool =
+  ## Wrapper around `updateState` for states expected to exist.
+  let ok = dag.updateState(state, bsi, save, cache)
+  if not ok:
+    error "State failed to load unexpectedly", bsi, tail = dag.tail.slot
+    doAssert verifyFinalization notin dag.updateFlags
+  ok
+
+template withUpdatedExistingState(
+    dag: ChainDAGRef, state: var ForkedHashedBeaconState,
+    bsiParam: BlockSlotId, okBody: untyped, failureBody: untyped): untyped =
+  ## Wrapper around `withUpdatedState` for states expected to exist.
+  block:
+    let bsi = bsiParam
+    dag.withUpdatedState(state, bsiParam) do:
+      okBody
+    do:
+      error "State failed to load unexpectedly", bsi, tail = dag.tail.slot
+      doAssert verifyFinalization notin dag.updateFlags
+      failureBody
+
+proc getExistingBlockIdAtSlot(dag: ChainDAGRef, slot: Slot): Opt[BlockSlotId] =
+  ## Wrapper around `getBlockIdAtSlot` for blocks expected to exist.
+  let bsi = dag.getBlockIdAtSlot(slot)
+  if bsi.isErr:
+    error "Block failed to load unexpectedly", slot, tail = dag.tail.slot
+    doAssert verifyFinalization notin dag.updateFlags
+  bsi
+
+proc existingParent*(dag: ChainDAGRef, bid: BlockId): Opt[BlockId] =
+  ## Wrapper around `parent` for parents known to exist.
+  let parent = dag.parent(bid)
+  if parent.isErr:
+    error "Parent failed to load unexpectedly", bid, tail = dag.tail.slot
+    doAssert verifyFinalization notin dag.updateFlags
+  parent
+
+proc getExistingForkedBlock(dag: ChainDAGRef, root: Eth2Digest):
+    Opt[ForkedTrustedSignedBeaconBlock] =
+  ## Wrapper around `getForkedBlock` for blocks expected to exist.
+  let bdata = dag.getForkedBlock(root)
+  if bdata.isErr:
+    error "Block failed to load unexpectedly", root, tail = dag.tail.slot
+    doAssert verifyFinalization notin dag.updateFlags
+  bdata
+
+proc getExistingForkedBlock*(
+    dag: ChainDAGRef, bid: BlockId): Opt[ForkedTrustedSignedBeaconBlock] =
+  ## Wrapper around `getForkedBlock` for blocks expected to exist.
+  let bdata = dag.getForkedBlock(bid)
+  if bdata.isErr:
+    error "Block failed to load unexpectedly", bid, tail = dag.tail.slot
+    doAssert verifyFinalization notin dag.updateFlags
+  bdata
+
 proc currentSyncCommitteeForPeriod(
     dag: ChainDAGRef,
     tmpState: var ForkedHashedBeaconState,
-    period: SyncCommitteePeriod): SyncCommittee =
+    period: SyncCommitteePeriod): Opt[SyncCommittee] =
   ## Fetch a `SyncCommittee` for a given sync committee period.
   ## For non-finalized periods, follow the chain as selected by fork choice.
   let earliestSlot = dag.computeEarliestLightClientSlot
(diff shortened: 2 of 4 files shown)


etan-status:
Note: The errors handled here are all hypothetical and should not be hit unless there is a logic bug or a corrupted database.

This PR is separate from the logic fixes in https://github.com/status-im/nimbus-eth2/pull/3498

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -79,7 +79,7 @@ proc addResolvedHeadBlock(
     epochRefDur = epochRefTick - putBlockTick
 
   # Update light client data
-  dag.processNewBlockForLightClient(state, trustedBlock, parent)
+  dag.processNewBlockForLightClient(state, trustedBlock, parent.bid)
```
etan-status, line 82: Callee ultimately operates on bid, and here we know that parent is non-nil, hence moved the extraction out to here

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -144,12 +193,14 @@ proc cacheLightClientData*(
...
   let
     bid =
       BlockId(root: blck.root, slot: blck.message.slot)
-    finalized_bid =
-      dag.getBlockIdAtSlot(finalized_checkpoint.epoch.start_slot).expect("TODO").bid
+    finalized_slot =
+      state.data.finalized_checkpoint.epoch.start_slot
+    finalized_bsi =
+      dag.getExistingBlockIdAtSlot(finalized_slot).valueOr:
+        default(BlockSlotId)
```
etan-status: When `finalized_bsi` fails to load, it is initialized to `slot == 0` and will be ignored in `LightClientUpdate` production (as it is `< earliestSlot`).

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -175,23 +226,20 @@ template lazy_header(name: untyped): untyped {.dirty.} =
...
       target = `name ptr`[]
     else:
-      let bid =
-        when blockParam is BlockId:
-          blockParam
-        else:
-          blockParam.bid
-      target = toBeaconBlockHeader(dag.getForkedBlock(bid).get)
+      let bdata = dag.getExistingForkedBlock(bid).valueOr:
+        return
```
etan-status, line 238: A bit ugly to have the `return` as a side effect here, but this is just a helper for the next function, so should be alright.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -398,12 +446,12 @@ proc processNewBlockForLightClient*(
...
     dag.cacheLightClientData(state.bellatrixData, signedBlock)
-    dag.createLightClientUpdates(state.bellatrixData, signedBlock, parent.bid)
+    dag.createLightClientUpdates(state.bellatrixData, signedBlock, parentBid)
   elif signedBlock is altair.TrustedSignedBeaconBlock:
     dag.cacheLightClientData(state.altairData, signedBlock)
-    dag.createLightClientUpdates(state.altairData, signedBlock, parent.bid)
+    dag.createLightClientUpdates(state.altairData, signedBlock, parentBid)
   elif signedBlock is phase0.TrustedSignedBeaconBlock:
-    discard
+    raiseAssert "Unreachable"
```
etan-status: There is a check above this that verifies the block is relevant (including altair check)

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -438,11 +490,10 @@ proc processFinalizationForLightClient*(dag: ChainDAGRef) =
   if dag.importLightClientData == ImportLightClientData.None:
     return
   let
-    altairStartSlot = dag.cfg.ALTAIR_FORK_EPOCH.start_slot
+    earliestSlot = dag.computeEarliestLightClientSlot
     finalizedSlot = dag.finalizedHead.slot
-  if finalizedSlot < altairStartSlot:
+  if finalizedSlot < earliestSlot:
```
etan-status, line 500: Data before earliest slot would not be considered for light client data anyway, can discard early

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -537,42 +591,44 @@ proc initBestLightClientUpdateForPeriod(
       computeDur = endTick - startTick
 
   proc maxParticipantsBlock(
-      dag: ChainDAGRef, highBlck: BlockId, lowSlot: Slot): Opt[BlockId] =
+      dag: ChainDAGRef, highBid: BlockId, lowSlot: Slot): Option[BlockId] =
```
etan-status, line 602: It is not an error to return a `none` result here, hence `Option` not `Opt`. `none` indicates a successful optional response, not an error'ed non-nil response.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -582,55 +638,68 @@ proc initBestLightClientUpdateForPeriod(
...
-    finalizedBlck: Opt[BlockId]
-  while bestFinalizedRef.isSome:
+    bestFinalizedBid = bestNonFinalizedBid
+    finalizedBid {.noinit.}: BlockId
+    bestFinalizedHasCheckpointInPeriod = false
+  while bestFinalizedBid.isSome:
+    defer:
+      if not bestFinalizedHasCheckpointInPeriod:
+        let newLowSlot = bestFinalizedBid.get.slot + 1
+        bestFinalizedBid = dag.maxParticipantsBlock(highBid, newLowSlot)
```
etan-status, line 656: A bit ugly, but the `valueOr` sort of promotes early returns / continue etc

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -738,17 +814,25 @@ proc initLightClientCache*(dag: ChainDAGRef) =
   let lightClientStartTick = Moment.now()
   debug "Initializing cached light client data"
 
+  template handleUnexpectedError(buggedBid: BlockId) =
+    doAssert buggedBid.slot > dag.lightClientCache.importTailSlot
+    dag.lightClientCache.importTailSlot = buggedBid.slot + 1
+    earliestSlot = dag.computeEarliestLightClientSlot
```
etan-status: For new blocks / states, we get them directly so we don't need any historic lookups. Once the light client data collection is started it will never fail. There is only the finalized checkpoint lookup above that can fail (which falls back to just not including it in the update by setting a default bid into the cache).

For historic blocks, this just degrades to assuming that everything before the buggedBid is unreliable. This avoids having to deal with all sort of extra logic to handle hypothetical errors, e.g. expected data not being present in the caches. If there is a programming error, or a database corruption, those issues should be fixed, and the server would just continue make data available for the blocks / states after the bug.

If this template is hit after already some partial data has been stored, that will be cleaned up at the next finalization update (same as it gets cleaned up if the entire data loads successfully)

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -738,17 +814,25 @@ proc initLightClientCache*(dag: ChainDAGRef) =
...
-    blocksBetween.add blockRef
-    blockRef = dag.parent(blockRef).expect("TODO")
-  blocksBetween.add blockRef
+    bid = dag.head.bid
+  while bid.slot > lowSlot:
+    blocksBetween.add bid
+    bid = dag.existingParent(bid).valueOr:
+      handleUnexpectedError(bid)
+      break
+  if bid.slot >= earliestSlot:
```
etan-status, line 848: A bit tricky. `handleUnexpectedError` makes `earliestSlot >= bid.slot + 1` so this if will not be triggered in case of a bug. Otherwise, the last bid may be double added.

We want one extra bid before finalizedSlot in case the block at finalizedHead.slot was missed (in that case, we also need to collect data back to the actual finalized block).

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
arnetheduck: all of these comments, make them code comments instead :)

etan-status:
- Fix stack explosion on `linux-arm64`
- Extended code comments
- Few local variable renames

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -885,26 +993,30 @@ proc getLightClientBootstrap*(
...
           return none(altair.LightClientBootstrap)
 
+      let period = slot.sync_committee_period
       var tmpState = assignClone(dag.headState)
       var bootstrap {.noinit.}: altair.LightClientBootstrap
       bootstrap.header =
         blck.toBeaconBlockHeader
       bootstrap.current_sync_committee =
-        dag.currentSyncCommitteeForPeriod(tmpState[], period)
+        dag.currentSyncCommitteeForPeriod(tmpState[], period).valueOr:
```
arnetheduck:
```suggestion
        ? dag.currentSyncCommitteeForPeriod(tmpState[], period)
```

etan-status:
- Rebase to `unstable`
- Changed `getLightClientBootstrap` result from `Option` to `Opt`, and adopted `?`. Getting an empty bootstrap result for a valid input is an error condition (data is computable, but we cannot provide it). For the other functions, sticking with `Option` because of their best-effort nature, and an empty result for them can be a valid output (e.g. at start of periods). In practice, the distinction does not really matter that much for these functions, though.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
arnetheduck: `Opt == Option` though - if we wanted to talk about errors and  non-errors, a `Result[..., void]` would be used.. just like `getBlock(..): Opt[]` means "couldn't find block for _whatever_ reason". `err` is a bit unfortunate, I think we'll add a `none`-like helper to  `Opt` to make it read better.

arnetheduck (approved):

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: `Opt` is a typedef for `Result[..., void]`. `Option` is a successful optional result that can be `none`.
`Opt` actually has "convenience" methods to make it be compatible with `Option`.

Issue #2677, opened 2021-06-25
Improve cache efficiency of attestation checks


arnetheduck:
Attestation checks currently use an `EpochRef` to validate the committee, based on the `target` vote: https://github.com/status-im/nimbus-eth2/blob/44f652f704f14c67e648f2b14dee652d44b858c5/beacon_chain/gossip_processing/gossip_validation.nim#L220

During periods of instability, it is often the case that the target vote is off, specially close to epoch transitions - in particular, clients might miss the last few blocks in the epoch and vote for a target with skipped slots - when we use the target directly, it means we must replay the state with the skipped slots in order to recreate the shuffling needed to validate the attestation in the context it was created.

However, the shuffling is already fixed one epoch prior - it would be possible to instead rely on `(target.blck.atSlot(target.epoch-2).getLastSlotInEpoch())` - this is sometimes called the dependent root - and extract a lookahead shuffling from _that_ state instead.

Relying on this older epoch will filter out a lot of the noise because we'll be looking at older data - any off-by-one or off-by-two-blocks votes will no longer cause replays.

The lookahead (epoch + 1) shuffling is normally not computed - `EpochRef` stores the shuffling for `EpochRef.epoch` - notably, we can only precompute the shuffling this way - we can't fill in the other `EpochRef` fields.

This means we must either extend `EpochRef` to also hold the shuffling for the _next_ epoch or split out the shuffling into a separate cache.

An instance of this kind of cache inefficiency can be seen in https://github.com/status-im/nimbus-eth2/issues/2675 where a series of poorly formed attestations led to a period of heightened memory usage that could have been avoided with the above scheme.

arnetheduck:
One way I see this working is that we have a `seed -> shuffling` cache and store the seed in `EpochRef` - alternatively, we store the shuffling as a `ref` in EpochRef.

With increased load on the REST API for duty lookahead, this cache increasingly becomes important

PR #3517, opened 2022-03-18
fix finalized epoch ref loading on checkpoint start

arnetheduck:
regression from #3513 that did not take tail into consideration when
loading epoch ancestor

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -307,8 +307,11 @@ func epochAncestor*(dag: ChainDAGRef, bid: BlockId, epoch: Epoch): EpochKey =
   ## This function returns an epoch key pointing to that epoch boundary, i.e. the
   ## boundary where the last block has been applied to the state and epoch
   ## processing has been done.
-  if epoch == dag.genesis.slot.epoch:
-    return EpochKey(bid: dag.genesis, epoch: epoch)
+  if epoch < dag.tail.slot.epoch or bid.slot < dag.tail.slot:
+    return EpochKey() # We can't load these states
+
+  if epoch == dag.tail.slot.epoch:
+    return EpochKey(bid: dag.tail, epoch: epoch)
 
   let bsi = dag.atSlot(bid, epoch.start_slot - 1).valueOr:
     # If we lack history for the given slot, we can use the given bid as epoch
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -307,8 +307,11 @@ func epochAncestor*(dag: ChainDAGRef, bid: BlockId, epoch: Epoch): EpochKey =
   ## This function returns an epoch key pointing to that epoch boundary, i.e. the
   ## boundary where the last block has been applied to the state and epoch
   ## processing has been done.
-  if epoch == dag.genesis.slot.epoch:
-    return EpochKey(bid: dag.genesis, epoch: epoch)
+  if epoch < dag.tail.slot.epoch or bid.slot < dag.tail.slot:
+    return EpochKey() # We can't load these states
```
etan-status, line 311: Will the callers be able to handle this? This seems to be the first empty EpochKey. `Opt[EpochKey]` as return could be a clearer alternative.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -881,17 +884,24 @@ proc getEpochRef*(
   ## Requests for epochs < dag.finalizedHead.slot.epoch may fail, either because
   ## the search was limited by the `preFinalized` flag or because state history
   ## has been pruned - none will be returned in this case.
-
   if not preFinalized and epoch < dag.finalizedHead.slot.epoch:
     return err()
 
+  if bid.slot < dag.tail.slot or epoch < dag.tail.slot.epoch:
+    return err()
```
etan-status, line 891: Does this need special handling for the case of `tail.slot > tail.slot.epoch.start_slot` / `epoch == dag.tail.slot.epoch`?

etan-status (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: maybe, but there's a better approach overall to do this which is outlined in the TODO / #2677, so I don't expect this function to stay around for long - in fact, epoch ancestor == `proposer_dependent_root` now which is a first step and once those steps are complete, this whole mechanism will go away.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: we de-facto don't support non-epoch-start-slot tails, so no :)

arnetheduck:
jenkins crash is light-client related

PR #3520, opened 2022-03-18
Grafana: aggregats/slot in fleets like in local Grafana [skip ci]

mratsim:
Locally:

image

Fleet

image

In the fleet aggregates/slot are in the same graph as attestations/slot and have a completely different scale than attestations/slot (potentially because of the new changes to seen aggregates #3439) but they are the main source of CPU consumption and should be highlighted.

Warning: tested locally, not in the fleet, i'm not sure how to test Grafana updates for the fleet to avoid losing the whole monitoring (at the very least on type and version incompatibilities)

### DIFF
--- a/grafana/metrics.status.im.json
+++ b/grafana/metrics.status.im.json
@@ -1728,6 +1728,7 @@
       "pluginVersion": "8.0.4",
       "targets": [
         {
+          "exemplar": true,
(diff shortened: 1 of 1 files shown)


arnetheduck:
The alternative here is to move aggregates to a separate scale (each graph can have two independent y axis scales) - basically, attestations+aggregates and sync messages+contributions form the same "pattern" of messages with subnets and aggregate channels, so it would be good that the two are presented the same way as well.

PR #3523, opened 2022-03-18
fix light client data pruning

etan-status:
When eliminating orphaned forks, light client data about blocks was also
deleted when the orphaned fork was referring to a state several slots
after the block. Linking light client data pruning with block deletion
instead of state deletion fixes this problem. Light client data always
refers to blocks and their immediate post-state.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1247,13 +1247,14 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
 
     var cur = head.atSlot()
     while not cur.blck.isAncestorOf(dag.finalizedHead.blck):
-      # Update light client data
-      dag.deleteLightClientData(cur.blck.bid)
 
       # TODO: should we move that disk I/O to `onSlotEnd`
       dag.delState(cur.toBlockSlotId().expect("not nil"))
 
       if cur.isProposed():
+        # Update light client data
+        dag.deleteLightClientData(cur.blck.bid)
+
         dag.forkBlocks.excl(KeyedBlockRef.init(cur.blck))
         dag.db.delBlock(cur.blck.root)
 


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1247,13 +1247,14 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
 
     var cur = head.atSlot()
     while not cur.blck.isAncestorOf(dag.finalizedHead.blck):
-      # Update light client data
```
etan-status, line 1250: @arnetheduck is `isAncestorOf` still working fine after the DAG pruning?

And also, is the early `break` of this loop correct? Shouldn't intermediate states past `finalizedHead` still be pruned instead of exiting as soon as `parent` goes `nil`, i.e shouldn't the break wait until `cur.parentOrSlot` returns `nil` `blockRef`? Granted, I think it only matters if there are multiple epochs without blocks.

Sample output from logging `cur` in the loop on a local testnet:
```
- 4d4ed541:41
- 1d3b181d:39@41
- cd9d6b68:35
- 6297b53f:32@35
- 06198182:33
- 5f5954dd:31@33
```

etan-status:
Script that was used to locally reproduce the bug:

Create 6 checkouts of the code, then in 6 different terminals, run `source env.sh`, update `EXECUTOR` in this command, and run:
```
export EXECUTOR=3; while :; do ./scripts/launch_local_testnet.sh --preset minimal --nodes 4 --stop-at-epoch 5 --disable-htop --enable-logtrace \
                                        --data-dir local_testnet0_data --base-port $(( 9000 + EXECUTOR * 100 )) --base-rest-port \
                                        $(( 7000 + EXECUTOR * 100 )) --base-metrics-port $(( 8008 + EXECUTOR * 100 )) --timeout 600 \
                                        --kill-old-processes \
                                        -- --verify-finalization --discv5:no && ./scripts/launch_local_testnet.sh --nodes 4 --stop-at-epoch 5 --disable-htop --enable-logtrace \
                                        --data-dir local_testnet1_data --base-port $(( 9000 + EXECUTOR * 100 )) --base-rest-port \
                                        $(( 7000 + EXECUTOR * 100 )) --base-metrics-port $(( 8008 + EXECUTOR * 100 )) --timeout 2400 \
                                        --kill-old-processes \
                                        -- --verify-finalization --discv5:no || break; done
```
then go afk and check the terminals that have stopped for bugs.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: this code is about pruning non-canonical branches, ie those that no longer have (the new) finalizedhead as ancestor - the break is correct in that it stops when we hit any part of the canonical chain.. I *think* `isAncestorOf` works in this particular case because the non-canonical chain will not yet have its parent fields nilled..

arnetheduck: this logic can be simplified now actually because `parent == nil` happens always, and checking the reverse ancestor is no longer the way to go

arnetheduck (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: you're right about the bug - noted it here: #3528

etan-status: Yep, the comment there is what I meant :) Basically, when the first block is several epochs after finalized head, those intermediate states would not be cleaned up.

PR #3525, opened 2022-03-18
fix loading when finalized checkpoint slot is missing block

arnetheduck:
ref loop would stop one block early in this case - trying to load
everything in one loop ends up being pretty confusing..

* simplify finalizedBlocks topup by splitting it from the head loop /
query

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -765,13 +765,15 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
     "...but that's the last BlockRef with a parent"
 
   block: # Top up finalized blocks
-    let finHigh = db.finalizedBlocks.high.get(dag.genesis.slot)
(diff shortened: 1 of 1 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -776,6 +764,21 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   doAssert dag.finalizedHead.blck.parent == nil,
     "...but that's the last BlockRef with a parent"
 
+  block: # Top up finalized blocks
+    let finHigh = db.finalizedBlocks.high.get(dag.genesis.slot)
```
etan-status: dag.tail.slot or dag.genesis.slot?

etan-status (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: genesis, we want to fill the finalized blocks as far as possible so as to prime the backfill correctly

PR #3527, opened 2022-03-19
extract chain DAG loading to separate function

etan-status:
When transitioning from light client to full node the chain DAG will be
loaded separately from the rest of the beacon node initialization.
Extracting chain DAG loading to a separate function will allow reusing
a lot of the existing code. This code move doesn't change semantics.


etan-status:
- Also extract weak subjectivity checkpoint validation for reuse

arnetheduck (approved):

PR #3528, opened 2022-03-20
simplify unviable head pruning

arnetheduck:
Also note bug that exists that potentially prevents states from being
pruned correctly

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1257,6 +1257,9 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
     # transition was not on the canonical chain - these will not properly get
     # cleaned up by the current logic - but they should also be rare
     # TODO clean up the above as well
+    doAssert dag.finalizedHead.blck.parent == nil,
+      "finalizedHead parent should have been pruned from memory already"
+
     while cur.blck.parent != nil:
       # TODO: should we move that disk I/O to `onSlotEnd`
       dag.delState(cur.toBlockSlotId().expect("not nil"))
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1251,8 +1251,13 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
       continue
 
     var cur = head.atSlot()
-    while not cur.blck.isAncestorOf(dag.finalizedHead.blck):
-
+    # The block whose parent is nil is the `BlockRef` that's part of the
+    # canonical chain but has now been finalized - in theory there could be
+    # states at empty slot iff the fork had epoch-long gaps where the epoch
+    # transition was not on the canonical chain - these will not properly get
+    # cleaned up by the current logic - but they should also be rare
+    # TODO clean up the above as well
+    while cur.blck.parent != nil:
       # TODO: should we move that disk I/O to `onSlotEnd`
       dag.delState(cur.toBlockSlotId().expect("not nil"))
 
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1263,10 +1268,12 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
         dag.forkBlocks.excl(KeyedBlockRef.init(cur.blck))
         dag.db.delBlock(cur.blck.root)
 
-      if cur.blck.parent.isNil:
-        break
+      let tmp = cur
       cur = cur.parentOrSlot
 
+      if tmp.isProposed: # Release `BlockRef` memory
+        tmp.blck.parent = nil
```
etan-status: Hmm, not sure whether this optimization is worth it, considering the meaning of `parent == nil` that could introduce subtle bugs if this block is still referenced somewhere. After `dag.heads.del(n)`, the last reference from the DAG should be cleared anyway (and the `= nil` will only release the memory on the next GC run as well)

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1251,8 +1251,13 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
     var cur = head.atSlot()
-    while not cur.blck.isAncestorOf(dag.finalizedHead.blck):
-
+    # The block whose parent is nil is the `BlockRef` that's part of the
+    # canonical chain but has now been finalized - in theory there could be
+    # states at empty slot iff the fork had epoch-long gaps where the epoch
+    # transition was not on the canonical chain - these will not properly get
+    # cleaned up by the current logic - but they should also be rare
+    # TODO clean up the above as well
+    while cur.blck.parent != nil:
```
etan-status, line 1263: As long as the `isAncestorOf` was correct, this is the same logic (just, simpler). However, this introduces a stricter ordering, namely that the `BlockRef` pruning needs to have happened before. To avoid accidents from code reorders etc, would consider putting an assert before the loop: `doAssert dag.finalizedHead.blck.parent == nil`

tests/test_blockchain_dag.nim
```diff
@@ -474,6 +474,15 @@ suite "chain DAG finalization tests" & preset():
       dag2.finalizedHead.slot == dag.finalizedHead.slot
       getStateRoot(dag2.headState) == getStateRoot(dag.headState)
 
+    # No canonical block data should be pruned by the removal of the fork
+    for i in Slot(0)..dag2.head.slot:
+      let bids = dag.getBlockIdAtSlot(i).expect("found it")
```
etan-status, line 479: NIT: `bids` > `bsi`

etan-status (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: this change makes it consistent with finalized blocks - that parent == nil is the new norm and needs to be consistently protected against everywhere

etan-status: Yeah, but in this case here, `tmp` is an orphaned block, quite the opposite of a finalized one. This is just an optimization in case there is multiple forks happening from this orphaned block onward, so that the followup loop iterations do not redundantly attempt to delete this block and the ancestor states. In that regard, the optimization may be useful. My concern is just that if there is a bug somewhere that it may lead to higher chance of hard-to-debug things happening. But on the other hand, I am also not as familiar with the rest of the code touching this, so if you say it's fine it most likely is.

One thing that the new code makes worse is that now, instead of having the suboptimal state deletion between the finalized head and the first orphaned block, it also gets amplified to every single fork. Granted, very unlikely to have forks with multiple epochs being empty, but still a semantic difference.

So, if there is

- [A/B] finalized head
    - [A/B] multiple epochs empty   <<< this is the area where states may not get deleted with both old and new code
        - [A/B] orphaned block  <<< this block parent will be set to `nil` after pruning the A branch
           - [A] orphaned block  
           - [B] multiple epochs empty <<< this area becomes also not correctly pruned with the new code but was fine with the old code, because while pruning the B branch we assume that the [A/B] orphaned block is the finalized one
              - [B] orphaned block

arnetheduck: alright, will leave it non-nil for now to avoid those junk states, just in case.

PR #3530, opened 2022-03-21
extract DAG dependent init to own function

etan-status:
During operation as a light client, the chain DAG is not available.
As a preparation, the beacon node initialization logic is divided into
parts depending on the presence of the chain DAG, and parts that are
always available (including a future light client mode).
This is a pure code move without semantic changes.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -202,6 +202,121 @@ proc checkWeakSubjectivityCheckpoint(
           headStateSlot = getStateField(dag.headState, slot)
     quit 1
 
+proc initFullNode(
+    node: BeaconNode,
+    rng: ref BrHmacDrbgContext,
+    dag: ChainDAGRef,
+    taskpool: TaskpoolPtr,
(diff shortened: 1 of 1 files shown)


arnetheduck:
Something I've thought about recently is a hybrid mode where you sync "normal" up to altair, then launch into a light client sync from there onwards - likewise if you lose connectivity for a while.

also, an alternative to not creating a chaindag at all is to create a genesis chaindag - this chaindag will be correct, albeit not very useful beyond serving as a not-nil point :) however, it would fit in with the above hybrid mode.

arnetheduck (approved):

etan-status:
Yeah, a hybrid mode would be cool to improve sync speeds after being offline for a while, but I am not sure about the popularity of such a use case in practice. It would also lead to a fragmented DAG with multiple intermediate sections missing which we don't currently support. Regarding a genesis chain DAG, the DAG would need adjustments to allow changing the tail / finalized head after the fact. Some form of flag to remember that we are syncing light vs full would still be needed. Or, having a genesis dag with a memory DB and then replace it with a fresh one, but that's also kinda messy.

PR #3531, opened 2022-03-21
Polish, remove beta warning for validator monitor

arnetheduck:
Validator monitoring improves logging by giving more specific monitoring
information, and can now be seen as complete.

Previously, logging has focused on "Attestation sent" messages which
carry little informational value when things go wrong, and are overly
aggressive when everything works as expected (sending attestations is
the norm).

* lower "Attestation sent" log to `INFO`
* lower "Previous epoch attestation missing" to `NOTICE` for easier
filtering
* add delay logging to validator monitor logs
* simplify delay logging code post-`BeaconTime`
* mark 1.7.0 as the start of the validator monitor feature - previous
versions had significant bugs in totals mode

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -390,7 +390,7 @@ proc createAndSendAttestation(node: BeaconNode,
       wallTime = node.beaconClock.now()
       delay = wallTime - attestationData.slot.attestation_deadline()
 
-    notice "Attestation sent",
+    info "Attestation sent",
       attestation = shortLog(attestation), validator = shortLog(validator),
       delay, subnet_id
 
@@ -990,7 +990,7 @@ proc sendAggregatedAttestations(
     let
       subnet_id = compute_subnet_for_attestation(
         committees_per_slot, slot, data.committee_index)
-    notice "Aggregated attestation sent",
+    info "Aggregated attestation sent",
       aggregate = shortLog(signedAP.message.aggregate),
       aggregator_index = signedAP.message.aggregator_index,
       signature = shortLog(signedAP.signature),
@@ -1225,7 +1225,7 @@ proc sendAttestation*(node: BeaconNode,
     wallTime = node.processor.getCurrentBeaconTime()
     delay = wallTime - attestation.data.slot.attestation_deadline()
 
-  notice "Attestation sent",
+  info "Attestation sent",
     attestation = shortLog(attestation), delay, subnet_id
 
   beacon_attestation_sent_delay.observe(delay.toFloatSeconds())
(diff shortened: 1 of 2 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -973,9 +962,10 @@ proc sendAggregatedAttestations(
     node.network.broadcastAggregateAndProof(signedAP)
 
     # The subnet on which the attestations (should have) arrived
-    let subnet_id = compute_subnet_for_attestation(
-      committees_per_slot, slot, data.committee_index)
-    notice "Aggregated attestation sent",
+    let
+      subnet_id = compute_subnet_for_attestation(
+        committees_per_slot, slot, data.committee_index)
+    info "Aggregated attestation sent",
```
mratsim, line 993: AFAIK we had the convention that everything related to user duties where on notice.
I don't think aggregates are that often, atleast for invidividual stakers.

Alternatively, we can create summary of duties fulfilled at slot end (in a separate message as it already is long).

beacon_chain/validators/validator_duties.nim
```diff
@@ -1209,17 +1199,12 @@ proc sendAttestation*(node: BeaconNode,
...
-        ($(wallTime - deadline), toFloatSeconds(wallTime - deadline))
+    delay = wallTime - attestation.data.slot.attestation_deadline()
 
-  notice "Attestation sent",
-    attestation = shortLog(attestation), delay = delayStr, subnet_id
+  info "Attestation sent",
+    attestation = shortLog(attestation), delay, subnet_id
 
-  beacon_attestation_sent_delay.observe(delaySecs)
+  beacon_attestation_sent_delay.observe(delay.toFloatSeconds())
```
mratsim, line 1231: Same on convention about user duties

mratsim (commented):
So far we said to users, things related to your duties are on notice.

If we move attestation sent / aggregate sent to info, we might want an additional summary of what the node did at slot end (in a separate message from slot end as it is already busy).

beacon_chain/validators/validator_duties.nim
arnetheduck: yeah, I'm thinking that perhaps it's time to move away from that convention and instead focus one when things *didn't* go as expected when possible, which is how the vmon is structured: "normal stuff"/successes on `INF`, things to consider on `NOT` (such as missed head votes etc) - the vmon is superior in that it traces not only performing the action (constructing an attestation and attempting broadcast) but rather track the entire lifetime up to chain inclusion - this makes it more powerful at showing things the user needs to care about, and as such, my idea was that we slightly can mute the things they don't need to care about - after all, sending the attestation doesn't make a profit, it's just one of many steps in its lifecycle.

PR #3534, opened 2022-03-21
Flake-based build environment for Nix users

zah:


arnetheduck:
the original nix files remained mostly broken and unmaintained - what's makes this time around different? does anyone else use nix really?

zah:
I've switched to NixOS on a new computer and I'll be using this on a daily basis. It's also something that I could see us using in the CI down the road (it's a way to pin down all software that is being used during the build - e.g. compilers, system libraries, etc).

arnetheduck:
>  It's also something that I could see us using in the CI down the road

if anything we should be using the same docker images we use for creating binaries so we test with the same gcc/glibc/patches

PR #3538, opened 2022-03-22
harden REST API `atSlot` against non-finalized blocks

arnetheduck:
* harden validator API against pre-finalized slot requests
* check `syncHorizon` when responding to validator api requests too far
from `head`
* limit state-id based requests to one epoch ahead of `head`
* put historic data bounds on block/attestation/etc validator production API, preventing them from being used with already-finalized slots
* add validator block smoke tests
* make rest test create a new genesis with the tests running roughly in
the first epoch to allow testing a few more boundary conditions

### DIFF
--- a/beacon_chain/rpc/rest_utils.nim
+++ b/beacon_chain/rpc/rest_utils.nim
@@ -62,7 +62,7 @@ proc getBlockSlotId*(node: BeaconNode,
     # head to avoid long empty slot replays (in particular a second epoch
     # transition)
     if not (stateIdent.slot.epoch <= (node.dag.head.slot.epoch + 1)):
-      return err("Requesting state too far ahead of head current head")
+      return err("Requesting state too far ahead of current head")
 
     let bsi = node.dag.getBlockIdAtSlot(stateIdent.slot).valueOr:
       return err("State for given slot not found, history not available?")
--- a/beacon_chain/rpc/rest_utils.nim
+++ b/beacon_chain/rpc/rest_utils.nim
@@ -61,7 +61,7 @@ proc getBlockSlotId*(node: BeaconNode,
     # Limit requests by state id to the next epoch with respect to the current
     # head to avoid long empty slot replays (in particular a second epoch
     # transition)
-    if not (stateIdent.slot.epoch <= (node.dag.head.slot.epoch + 1)):
+    if stateIdent.slot.epoch > (node.dag.head.slot.epoch + 1):
       return err("Requesting state too far ahead of current head")
 
     let bsi = node.dag.getBlockIdAtSlot(stateIdent.slot).valueOr:
--- a/beacon_chain/rpc/rest_utils.nim
+++ b/beacon_chain/rpc/rest_utils.nim
@@ -3,6 +3,7 @@ import std/[options, macros],
        ../spec/[forks],
        ../spec/eth2_apis/[rest_types, eth2_rest_serialization],
        ../beacon_node,
+       ../validators/validator_duties,
        ../consensus_object_pools/blockchain_dag,
        "."/[rest_constants, state_ttl_cache]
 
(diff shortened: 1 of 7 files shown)


beacon_chain/rpc/rest_utils.nim
```diff
@@ -39,20 +40,13 @@ proc validate(key: string, value: string): int =
...
-    err("Requesting slot too far ahead of the current head")
-
 proc getCurrentHead*(node: BeaconNode, slot: Slot): Result[BlockRef, cstring] =
-  let res = node.dag.head
-  # if not(node.isSynced(res)):
-  #   return err("Cannot fulfill request until node is synced")
-  if res.slot + uint64(2 * SLOTS_PER_EPOCH) < slot:
+  let head = node.dag.head
+
+  if slot > head.slot and not node.isSynced(head):
```
etan-status, line 46: `node.isSynced` returns `false` when `head` is more than `syncHorizon` behind the wall clock, it does not take into account the requested `slot`.

The old logic allowed `slot` to be `syncHorizon` ahead of `head` even while not yet fully synced.
The new logic no longer allows even requesting a single slot ahead of `head` while sync has not completed.

Intended?

beacon_chain/rpc/rest_utils.nim
```diff
@@ -64,7 +58,13 @@ proc getBlockSlotId*(node: BeaconNode,
                      stateIdent: StateIdent): Result[BlockSlotId, cstring] =
   case stateIdent.kind
   of StateQueryKind.Slot:
-    let bsi = node.dag.getBlockIdAtSlot(? node.getCurrentSlot(stateIdent.slot)).valueOr:
+    # Limit requests by state id to the next epoch with respect to the current
+    # head to avoid long empty slot replays (in particular a second epoch
+    # transition)
+    if stateIdent.slot.epoch + 1 > node.dag.head.slot.epoch:
+      return err("Requesting state too far ahead of head current head")
```
etan-status: - There is an extra `head` in the error message.
- Logic looks alright, `slot.epoch + 1` does not seem to be overflowable.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -84,8 +84,12 @@ proc getBlockSlotId*(node: BeaconNode,
...
       ok(node.dag.finalizedHead.toBlockSlotId().expect("not nil"))
     of StateIdentType.Justified:
-      ok(node.dag.head.atEpochStart(getStateField(
-        node.dag.headState, current_justified_checkpoint).epoch).toBlockSlotId().expect("not nil"))
+      # Take checkpoint-synced nodes into account
+      let justifiedEpoch =
+        max(
+          getStateField(node.dag.headState, current_justified_checkpoint).epoch,
+          node.dag.finalizedHead.slot.epoch)
+      ok(node.dag.head.atEpochStart(justifiedEpoch).toBlockSlotId().expect("not nil"))
```
etan-status, line 92: Good catch!

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -606,19 +615,22 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-        let rslot = res.get()
-        if epoch(rslot) < node.dag.cfg.ALTAIR_FORK_EPOCH:
-          return RestApiResponse.jsonError(Http400,
-                                           SlotFromTheIncorrectForkError)
-        rslot
+
+      let res = slot.get()
+      if res.isErr():
+        return RestApiResponse.jsonError(Http400, InvalidSlotValueError,
+                                          $res.error())
```
etan-status, line 647: NIT: Indentation of the second line w.r.t. `(`, on all of these.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -400,6 +406,9 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
               return RestApiResponse.jsonError(Http400, InvalidSlotValueError,
                                                $res.error())
             res.get()
+        if qslot <= node.dag.finalizedHead.slot:
```
etan-status, line 420: The altair check of other endpoints is inside the block fetching `qslot` -- should this additional check be located similarly, as in, first loading into `rslot`, then validating against the allowed range, then assigning to `qslot`? It doesn't matter semantically, it's more of a style / consistency thing.

etan-status (approved):

beacon_chain/rpc/rest_utils.nim
arnetheduck: yes, while syncing we don't want to allow spurious requests for slots that will cause expensive empty-slots replays of the state

beacon_chain/rpc/rest_utils.nim
arnetheduck: actually it was overflowing because of the magic far_future_epoch handling (that at some point should be applied to `+` as well...)

beacon_chain/rpc/rest_validator_api.nim
arnetheduck: indeed, there's a myriad of these little improvements that one _could_ make with infinite time at hand..

beacon_chain/rpc/rest_validator_api.nim
arnetheduck: this is true, but also a bit of a chore / best-effort thing as long as we don't have an  autoformatter.. the best way to highlight  these is through the little suggestion button which gives the right perfomance/effort ratio for fixing them (and avoids typing NIT in ALL-CAPS :))

Screenshot from 2022-03-22 20-17-04

PR #3539, opened 2022-03-22
reformat long lines

etan-status:
Shortens some long lines by introducing temp variables and line breaks.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1174,9 +1174,12 @@ proc onSlotEnd(node: BeaconNode, slot: Slot) {.async.} =
   await node.updateGossipStatus(slot + 1)
 
 func syncStatus(node: BeaconNode): string =
-  if node.syncManager.inProgress: node.syncManager.syncStatus
-  elif node.backfiller.inProgress: "backfill: " & node.backfiller.syncStatus
-  else: "synced"
+  if node.syncManager.inProgress:
+    node.syncManager.syncStatus
+  elif node.backfiller.inProgress:
+    "backfill: " & node.backfiller.syncStatus
+  else:
+    "synced"
 
 proc onSlotStart(
     node: BeaconNode, wallTime: BeaconTime, lastSlot: Slot) {.async.} =
@@ -1241,7 +1244,8 @@ proc onSecond(node: BeaconNode) =
   # Nim GC metrics (for the main thread)
   updateThreadMetrics()
 
-  if node.config.stopAtSyncedEpoch != 0 and node.dag.head.slot.epoch >= node.config.stopAtSyncedEpoch:
+  if node.config.stopAtSyncedEpoch != 0 and
+      node.dag.head.slot.epoch >= node.config.stopAtSyncedEpoch:
     notice "Shutting down after having reached the target synced epoch"
     bnStatus = BeaconNodeStatus.Stopping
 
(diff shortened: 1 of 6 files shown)


beacon_chain/rpc/rest_config_api.nim
```diff
@@ -16,13 +16,14 @@ logScope: topics = "rest_config"
 
 proc installConfigApiHandlers*(router: var RestRouter, node: BeaconNode) =
   let
+    cfg = addr node.dag.cfg
```
arnetheduck: we avoid `addr` for frivolous cases like this - ie the typing savings are not worth the risks it introduces - as much as possible, we want addr (and `ref`)-free code as these come with fewer compiler-based guarantees and instead rely on the programmer

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -978,15 +978,16 @@ proc updateGossipStatus(node: BeaconNode, slot: Slot) {.async.} =
     HYSTERESIS_BUFFER = 16
 
   let
-    head = node.dag.head
+    dag = node.dag
```
arnetheduck: tendentially we've avoided extra locals like this because they (sometimes) introduce async lifetime issues when copy-pasted into the wrong place - `dag` is fine, but `head` is somewhat vulnerable as it can change - if dag-based state variables are taken out of their shell, it should be accompanied by a comment explaining the expected behavior under async interruption conditions

tersec:
`rpc_validator_api` and `rpc_beacon_api` were deprecated in https://github.com/status-im/nimbus-eth2/releases/tag/v22.3.0 and I'm not sure it's worth fixing much about them at this point. I expect them to disappear before long.

tersec:
https://status-im.github.io/nim-style-guide/formatting.style.html does cite https://nim-lang.org/docs/nep1.html:
> Lines should be no longer than 80 characters. Limiting the amount of information present on each line makes for more readable code - the reader has smaller chunks to process.

But in general, many of these changes shorten lines from, say, 30 to 20 characters.

beacon_chain/nimbus_beacon_node.nim
tersec: Well, this kind of thing is the usual use of `template dag(): auto = node.dag`. If the goal here is merely lexicographical, templates are the straightforward way to do that.

beacon_chain/rpc/rest_config_api.nim
tersec: To the extent this is an issue, and I do see the repetition it avoids, `template cfg...` is a direct stand-in, without these issues.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -64,7 +64,8 @@ proc getBlockSlotId*(node: BeaconNode,
                      stateIdent: StateIdent): Result[BlockSlotId, cstring] =
   case stateIdent.kind
   of StateQueryKind.Slot:
-    let bsi = node.dag.getBlockIdAtSlot(? node.getCurrentSlot(stateIdent.slot)).valueOr:
+    let bsi = node.dag.getBlockIdAtSlot(
```
tersec: This stuff is fine, and is what I was expecting in a PR named "reformat long lines".

beacon_chain/rpc/rpc_beacon_api.nim
```diff
@@ -150,25 +150,26 @@ proc getStatus(validator: Validator,
 proc getForkedBlockFromBlockId(
     node: BeaconNode, blockId: string): ForkedTrustedSignedBeaconBlock {.
     raises: [Defect, CatchableError].} =
+  let dag = node.dag
```
tersec: If one wants to do this, `template` is probably better than `let`.

But see separate comment, `rpc_beacon_api.nim` is deprecated and should go away soon regardless.

etan-status:
- Rebase to `unstable`
- Avoid `ref` -> use `template` instead
- Drop changes to deprecated RPC api
- Drop introduction of `dag` local

PR #3540, opened 2022-03-23
kiln beacon node

tersec:
Non-goals: block proposing or JWT support. Attesting should work, but I wouldn't consider it not working a blocker for this PR. Becaue of lack of JWT support, using the Geth configuration at https://notes.ethereum.org/@launchpad/kiln for example, one needs to use `--web3-url=ws://127.0.0.1:8546`. For example, I'm using a script:
```sh
#!/usr/bin/env sh
set -eu
nimbus-eth2/build/nimbus_beacon_node \
    --network=merge-testnets/kiln \
    --web3-url=ws://127.0.0.1:8546
    --rest \
    --metrics \
    --log-level=DEBUG \
    --terminal-total-difficulty-override=20000000000000 \
    --jwt-secret="/tmp/jwtsecret"
```
which is a copy/paste of the launchpad example, but `s/8551/8546/`.

It does function as a beacon node following Kiln and avoids nontrivial changes to the Eth1 monitor; in particular, it does not track do anything with transition total difficulty. It detects The Merge simply by when non-empty `ExecutionPayload`s appear, which is enough for its purpose.

Its initial form is a bit of a dump of a subset of the code in `kiln-dev-auth`, and needs cleanup, but since the goal is to merge this in a reasonably timely way, leaving it as non-draft.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -721,7 +721,8 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   # Load head -> finalized, or all summaries in case the finalized block table
   # hasn't been written yet
   for blck in db.getAncestorSummaries(head.root):
-    let newRef = BlockRef.init(blck.root, blck.summary.slot)
+    let newRef = BlockRef.init(
+      blck.root, default(Eth2Digest), blck.summary.slot)
     if headRef == nil:
       doAssert blck.root == head.root
       headRef = newRef
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -109,7 +109,7 @@ type
     forcePolling: bool
     jwtSecret: seq[byte]
 
-    dataProvider*: Web3DataProviderRef  # TODO evidently not meant for export
+    dataProvider: Web3DataProviderRef
     latestEth1Block: Option[FullBlockId]
 
     depositsChain: Eth1Chain
@@ -454,29 +454,29 @@ proc getPayload*(p: Web3DataProviderRef,
 
   p.web3.provider.engine_getPayloadV1(FixedBytes[8] payloadId)
 
-proc newPayload*(p: Web3DataProviderRef,
-                 payload: engine_api.ExecutionPayloadV1): Future[PayloadStatusV1] =
+proc newPayload*(p: Eth1Monitor, payload: engine_api.ExecutionPayloadV1):
+    Future[PayloadStatusV1] =
   # Eth1 monitor can recycle connections without (external) warning; at least,
   # don't crash.
-  if p.isNil:
+  if p.dataProvider.isNil:
     var epr: Future[PayloadStatusV1]
     epr.complete(PayloadStatusV1(status: PayloadExecutionStatus.syncing))
     return epr
 
-  p.web3.provider.engine_newPayloadV1(payload)
+  p.dataProvider.web3.provider.engine_newPayloadV1(payload)
 
-proc forkchoiceUpdated*(p: Web3DataProviderRef,
+proc forkchoiceUpdated*(p: Eth1Monitor,
                         headBlock, finalizedBlock: Eth2Digest):
                         Future[engine_api.ForkchoiceUpdatedResponse] =
   # Eth1 monitor can recycle connections without (external) warning; at least,
   # don't crash.
-  if p.isNil:
+  if p.dataProvider.isNil:
     var fcuR: Future[engine_api.ForkchoiceUpdatedResponse]
     fcuR.complete(engine_api.ForkchoiceUpdatedResponse(
       payloadStatus: PayloadStatusV1(status: PayloadExecutionStatus.syncing)))
     return fcuR
 
-  p.web3.provider.engine_forkchoiceUpdatedV1(
+  p.dataProvider.web3.provider.engine_forkchoiceUpdatedV1(
     ForkchoiceStateV1(
       headBlockHash: headBlock.asBlockHash,
 
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -14,13 +14,13 @@ import
   chronos, json, metrics, chronicles/timings, stint/endians2,
   web3, web3/ethtypes as web3Types, web3/ethhexstrings, web3/engine_api,
   eth/common/eth_types,
-  eth/async_utils, stew/[byteutils, shims/hashes],
+  eth/async_utils, stew/[byteutils, objects, shims/hashes],
   # Local modules:
   ../spec/[eth2_merkleization, forks, helpers],
   ../spec/datatypes/[base, phase0, bellatrix],
   ../networking/network_metadata,
   ../consensus_object_pools/block_pools_types,
-  ".."/[beacon_chain_db, beacon_node_status],
+  ".."/[beacon_chain_db, beacon_node_status, beacon_clock],
   ./merkle_minimal
 
 export
@@ -109,7 +109,7 @@ type
     forcePolling: bool
     jwtSecret: seq[byte]
 
-    dataProvider: Web3DataProviderRef
+    dataProvider*: Web3DataProviderRef  # TODO evidently not meant for export
     latestEth1Block: Option[FullBlockId]
 
     depositsChain: Eth1Chain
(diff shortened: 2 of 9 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1880,3 +1885,37 @@ proc rebuildIndex*(dag: ChainDAGRef) =
 
     for i in junk:
       dag.db.delState(i[1])
+
+proc newExecutionPayload*(
```
arnetheduck: it would be better to leave blockchain_dag async-free and put this somewhere else - ie the dag is a synchronous data store / adt mainly, it shouldn't have to deal with this stuff.. `validator_duties` maybe?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -324,6 +362,64 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
...
+        headBlockRoot = self.consensusManager.dag.head.executionBlockRoot
+
+        finalizedBlockRoot =
+          if not isZero(
+              self.consensusManager.dag.finalizedHead.blck.executionBlockRoot):
+            self.consensusManager.dag.finalizedHead.blck.executionBlockRoot
+          else:
+            default(Eth2Digest)
+
+      await self.runForkchoiceUpdated(headBlockRoot, finalizedBlockRoot)
```
arnetheduck, line 458: can this be an `asyncSpawn`  so that the next block can be processed while fcu is running? or can there be max 1 in-flight fcu at any time?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,15 +445,37 @@ proc getBlockByNumber*(p: Web3DataProviderRef,
 
 proc getPayload*(p: Web3DataProviderRef,
                  payloadId: bellatrix.PayloadID): Future[engine_api.ExecutionPayloadV1] =
+  # Eth1 monitor can recycle connections without (external) warning; at least,
+  # don't crash.
+  if p.isNil:
+    var epr: Future[engine_api.ExecutionPayloadV1]
+    epr.complete(default(engine_api.ExecutionPayloadV1))
+    return epr
```
mratsim, line 453: Seems like we need a some ConnectionPool type to make that behavior readily apparent

beacon_chain/spec/engine_authentication.nim
```diff
@@ -83,7 +83,7 @@ proc checkJwtSecret*(
 
   try:
     let lines = readLines(jwtSecret.get, 1)
-    if lines.len > 0 and lines[0].startswith("0x"):
+    if lines.len > 0:
```
mratsim, line 86: A comment should be added that mentions that the secret key is parsed in constant-time using https://github.com/cheatfate/nimcrypto/pull/44. In case people were tempted to refactor that away.

beacon_chain/eth1/eth1_monitor.nim
tersec: Probably, but I'd consider that out of scope of this PR.

beacon_chain/spec/engine_authentication.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3540/commits/161a7511cd45cb90dcae3dc2b54c12ebd5c3ce75

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.8/src/engine/specification.md#message-ordering specifies that
> Consensus Layer client software **MUST** respect the order of the corresponding fork choice update events when making calls to the `engine_forkchoiceUpdated` method.

As long as Chronos doesn't reorder them the sequence of `newPayload` and `forkchoiceUpdated` calls from this function, it shouldn't violate the spec. There was some discussion of stricter ordering requirements from that perspective, but I don't see them in the current engine API spec.

However, while this is true for `forkchoiceUpdated`, there's still a partial ordering requirement of `fcU(triggered by block)` happening after `newPayload(triggered by block)`, as well as `newPayload(triggered by block)` being something that one can't use `asyncSpawn` for. So long as both of these partial ordering requirements can be met by using `asyncSpawn`.

A minor point is that my experience is that asyncSpawn means that one has to be more confident about having handled all exception cases, which this introduces in quantity given that it's network communication that, while currently localhost, might not alway be.

Finally, it's in this form for simplicity for now. Part of the goal is to get something working and tested in that further PRs in this series (e.g., the JWT PR and block proposing PR) can build on, rather than try to 100% perfect this. So currently, the model for interaction with ELs has proven stable and compatible, and this would potentially destabilize that, or, potentially, require more testing to ferret out bugs.

I agree with this as a longer-term goal for sure, but am ambivalent about the sequencing: I think I prefer
1) one-fcU-in-flight
2) JWT and block proposal [neither depends on other]
3) somewhere in here, Eth1 TTD monitoring, but block proposal doesn't strictly depend on it (not having TTD monitoring just means that Nimbus can't be the specific client to execute The Merge, but it can follow CLs which do)
4) changes to core engine API interaction loop [not code refactorings, but semantic changes]

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: `block_processor` doesn't pick up things exported from `validator_duties` currently, and the latter is too all-encompassing in general to want to add an import for that purpose, but I do agree about `blockchain_dag` not being optimally appropriate.

https://github.com/status-im/nimbus-eth2/pull/3540/commits/ff66f50a704d44af69f003529d8f28b58ac693b9 moves it to `block_processor` itself, which at least is the same module in which it's used.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -721,7 +721,8 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   # Load head -> finalized, or all summaries in case the finalized block table
   # hasn't been written yet
   for blck in db.getAncestorSummaries(head.root):
-    let newRef = BlockRef.init(blck.root, blck.summary.slot)
+    let newRef = BlockRef.init(
+      blck.root, default(Eth2Digest), blck.summary.slot)
```
arnetheduck, line 725: hm, this won't work for bellatrix of course .. what's the plan, extend summaries?

arnetheduck: or, a better question: the new field, how far back is it needed, non-final only or further back than that?

beacon_chain/spec/engine_authentication.nim
arnetheduck: is it .. *that* secret really? I mean. . it's a plaintext file on disk, what exactly will a constant-time parsing change here?

arnetheduck (approved):

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/blob/061bc419acbf33771e89447cc54070eedd50fe34/beacon_chain/consensus_object_pools/block_dag.nim#L66-L70
means that when a `BlockRef` is created by
https://github.com/status-im/nimbus-eth2/blob/061bc419acbf33771e89447cc54070eedd50fe34/beacon_chain/consensus_object_pools/block_clearance.nim#L42

That is, when it has access to the entire contents of the block without further I/O or memory usage.

tersec: There is a hole introduced here though, which is that, yeah, the database schema should be adjusted at some point. It's not critical, because once the node is running for a while, it mostly matters in-memory.

This block hash matters when sending `forkchoiceUpdated`, because it has to send the execution payload hash of the CL head to the EL client, not the CL hash tree root (which the EL client doesn't know about, in general). `forkchoiceUpdated`, in turn, occurs in two places:

1) already present in this PR, in `block_processor`, as part of the `newPayload`/`forkchoiceUpdated` loop. Here, it's not catastrophic to, for an epoch or two, not be able to successfully update the EL's fork choice information, though of course it's not great; and more importantly,

2) when proposing a block. This is a genuine problem, and does need some database work, because that `forkchoiceUpdated` is a crucial part of the `getPayload` flow, without which Nimbus won't be able to propose correctly. This is something that needs to be addressed before considering Nimbus fully merge-ready, and takes care of (1) automatically.

Neither case needs pre-finalization execution payload hashes. It's probably fine to have a by-design-semi-ephemeral cache table, or similar. It'd need to handle somewhat unbounded amounts of data, insofar as nonfinality could continue for a while, but ultimately it's not much data.

beacon_chain/spec/engine_authentication.nim
tersec: Well it was enough of an issue to trigger specific notice in the other case during an audit, so I guess maybe worthwhile treating this similarly.

PR #3544, opened 2022-03-24
fix bellatrix table names

arnetheduck:
this should/will cause existing nimbus databases to revert to the altair
merge and resync with the new table name


tersec:
Using it on an existing database, it gets
> Could not load head state, database corrupt? head=97a6b585:4798 tail=d30e447c:0 genesis=d30e447c:0

tersec (approved):

arnetheduck:
> Using it on an existing database, it gets

this has been traced to a database that had finalized on the kiln branch - the code is not able to rewind past the finality point - given that kiln has not seen a public release yet, early testers will need to wipe their database and resync.

PR #3546, opened 2022-03-24
remove TaintedString

tersec:
It's deprecated in Nim 1.6 (https://nim-lang.org/docs/system.html#TaintedString), wasn't used to meaningfully distinguish between trusted and untrusted input regardless, and even if it were, those attempts didn't have type-checking force because `-d:taintMode` hadn't been enabled. It was just a kind of dead code which now makes Nim 1.6 builds noisy.

### DIFF
--- a/beacon_chain/eth1/deposit_contract.nim
+++ b/beacon_chain/eth1/deposit_contract.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2018-2021 Status Research & Development GmbH
+# Copyright (c) 2018-2022 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
@@ -231,7 +231,7 @@ proc main() {.async.} =
 
   if conf.askForKey:
     var
-      privateKey: TaintedString
+      privateKey: string
       reasonForKey = ""
 
     if conf.cmd == StartUpCommand.sendDeposits:
--- a/beacon_chain/eth1/deposit_contract.nim
+++ b/beacon_chain/eth1/deposit_contract.nim
@@ -231,7 +231,7 @@ proc main() {.async.} =
 
   if conf.askForKey:
     var
-      privateKey: string
+      privateKey: string  # TODO consider using a SecretString type
       reasonForKey = ""
 
     if conf.cmd == StartUpCommand.sendDeposits:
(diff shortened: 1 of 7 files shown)


stefantalpalaru (approved):

beacon_chain/eth1/deposit_contract.nim
```diff
@@ -231,7 +231,7 @@ proc main() {.async.} =
 
   if conf.askForKey:
     var
-      privateKey: TaintedString
+      privateKey: string
```
mratsim: Maybe we need a separate `SecretString` distinct type for this

beacon_chain/spec/keystore.nim
```diff
@@ -389,7 +389,7 @@ proc cmpIgnoreCase(lhs: cstring, rhs: string): int =
   # Nim should have a `cmp` function for C strings.
   cmpIgnoreCase($lhs, rhs)
 
-proc validateMnemonic*(inputWords: TaintedString,
+proc validateMnemonic*(inputWords: string,
```
mratsim, line 392: Maybe we need a separate `SecretString` distinct type for this

beacon_chain/validators/keystore_management.nim
```diff
@@ -1038,7 +1038,7 @@ proc resetAttributesNoError() =
 
 proc importKeystoresFromDir*(rng: var BrHmacDrbgContext,
                              importedDir, validatorsDir, secretsDir: string) =
-  var password: TaintedString
+  var password: string
```
mratsim: Maybe we need a separate `SecretString` distinct type for this

mratsim (approved):
OK with the PR.

While I don't see cause for concern for input strings which all come from config files and are validated on parsing. For the keystores/passwords/seed words I think we also need a SecretString distinct type (in another PR) to make it easier to audit handling on extra sensitive secrets.

beacon_chain/eth1/deposit_contract.nim
tersec: Well, and this kind of thing could be more focused on using `burnMem` or similar functionality systematically, in a way that wouldn't have made sense with `TaintedString`.

tersec:
`TaintedString` was overused for the help-people-audit-sensitive-datapaths purpose anyway.

beacon_chain/validators/keystore_management.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3546/commits/57aa145fc9ae80071227a2b91193e4a1476c56c1

beacon_chain/spec/keystore.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3546/commits/57aa145fc9ae80071227a2b91193e4a1476c56c1

beacon_chain/eth1/deposit_contract.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3546/commits/57aa145fc9ae80071227a2b91193e4a1476c56c1

PR #3550, opened 2022-03-28
update light client protocol version

etan-status:
So far we use `v0` versions for the light client sync libp2p protocol.
When the protocol should be adopted, will have to use v1 instead.
This patch tracks that version change.


beacon_chain/spec/network.nim
```diff
@@ -101,7 +101,7 @@ func getSyncCommitteeContributionAndProofTopic*(forkDigest: ForkDigest): string
 # https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#optimistic_light_client_update
 func getOptimisticLightClientUpdateTopic*(forkDigest: ForkDigest): string =
   ## For broadcasting or obtaining the latest `OptimisticLightClientUpdate`.
-  eth2Prefix(forkDigest) & "optimistic_light_client_update_v0/ssz_snappy"
+  eth2Prefix(forkDigest) & "optimistic_light_client_update/ssz_snappy"
```
tersec: Does this introduce namespace pollution? I had thought one reason for v0 was to avoid that, but this looks a bit more ... bold.

etan-status: Yes, this PR is blocked until https://github.com/ethereum/consensus-specs/pull/2802 is accepted. 
Still have to keep in mind that this rename is still open, hence the draft PR.

etan-status:
libp2p Spec has been merged: https://github.com/ethereum/consensus-specs/pull/2802

PR #3551, opened 2022-03-28
use snappy-framed format for compressing bellatrix+ database entries

arnetheduck:
`.era` files and Req/Resp protocols use framed formats - aligning the
database with these makes for less recompression work overall as gossip
is sent only once while req/resp repeats (potentially) - this also
allows efficient pruning-to-era where snappy-recompression is the major
cycle thief.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -584,9 +584,7 @@ type GetResult = enum
 proc getSSZ[T](db: KvStoreRef, key: openArray[byte], output: var T): GetResult =
   var status = GetResult.notFound
 
-  # TODO address is needed because there's no way to express lifetimes in nim
-  #      we'll use unsafeAddr to find the code later
-  var outputPtr = unsafeAddr output # callback is local, ptr wont escape
+  var outputPtr = addr output # callback is local, ptr wont escape
   proc decode(data: openArray[byte]) =
     status =
       if decodeSSZ(data, outputPtr[]): GetResult.found
@@ -602,9 +600,7 @@ proc putSSZ(db: KvStoreRef, key: openArray[byte], v: auto) =
 proc getSnappySSZ[T](db: KvStoreRef, key: openArray[byte], output: var T): GetResult =
   var status = GetResult.notFound
 
-  # TODO address is needed because there's no way to express lifetimes in nim
-  #      we'll use unsafeAddr to find the code later
-  var outputPtr = unsafeAddr output # callback is local, ptr wont escape
+  var outputPtr = addr output # callback is local, ptr wont escape
   proc decode(data: openArray[byte]) =
     status =
       if decodeSnappySSZ(data, outputPtr[]): GetResult.found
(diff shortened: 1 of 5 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -794,9 +871,53 @@ proc getBlockSSZ*(
...
+  let dataPtr = unsafeAddr data # Short-lived
+  var success = true
+  proc decode(data: openArray[byte]) =
+    try: dataPtr[] = framingFormatCompress(
+      snappy.decode(data, maxDecompressedDbRecordSize))
+    except CatchableError: success = false
+  db.blocks[BeaconBlockFork.Phase0].get(key.data, decode).expectDb() and success or
+    db.v0.getPhase0BlockSSZ(key, data)
+
+proc getBlockSZ*(
```
tersec, line 882: Why are getBlockSZ(phase0) and getBlockSZ(altair) separate functions? Bellatrix differs in that it does `assign` rather than `snappy.decode`.

beacon_chain/beacon_chain_db.nim
```diff
@@ -835,6 +956,47 @@ proc getStateOnlyMutableValidators(
...
+  ## Load state into `output` - BeaconState is large so we want to avoid
+  ## re-allocating it if possible
+  ## Return `true` iff the entry was found in the database and `output` was
+  ## overwritten.
+  ## Rollback will be called only if output was partially written - if it was
+  ## not found at all, rollback will not be called
+  # TODO rollback is needed to deal with bug - use `noRollback` to ignore:
+  #      https://github.com/nim-lang/Nim/issues/14126
+  # TODO RVO is inefficient for large objects:
+  #      https://github.com/nim-lang/Nim/issues/13879
```
tersec, line 968: Is RVO still inefficient for large objects? That Nim issue is closed.

beacon_chain/beacon_chain_db.nim
```diff
@@ -594,6 +617,24 @@ proc getSnappySSZ[T](db: KvStoreRef, key: openArray[byte], output: var T): GetRe
 proc putSnappySSZ(db: KvStoreRef, key: openArray[byte], v: auto) =
   db.put(key, encodeSnappySSZ(v)).expectDb()
 
+proc getSZSSZ[T](db: KvStoreRef, key: openArray[byte], output: var T): GetResult =
+  var status = GetResult.notFound
+
+  # TODO address is needed because there's no way to express lifetimes in nim
+  #      we'll use unsafeAddr to find the code later
```
tersec: Nim 1.6 deprecates `unsafeAddr` though, so by the time there's a later, `unsafeAddr` might be throwing deprecation warnings that would be unfortunate to keep around for the sake of being able to find code: https://github.com/nim-lang/Nim/pull/19373 https://github.com/nim-lang/Nim/pull/19408

beacon_chain/beacon_chain_db.nim
arnetheduck: `phase0` reads from an additional table.. yeah, there's   better ways to factor the code I'm sure but ..

beacon_chain/beacon_chain_db.nim
arnetheduck: ah, yes.. there's https://github.com/nim-lang/Nim/issues/15571 now instead

arnetheduck: longer story would be that to remove the TODO, one would need to examine the C code again and see what's going on - issue might be closed but that doesn't mean it's actually fixed for our use case

beacon_chain/beacon_chain_db.nim
arnetheduck: 19dddbc5113a57e25099814e4cebf56e648d9713

tersec (approved):

PR #3555, opened 2022-03-29
bump nim-stew to remove pragma disabling checks

tersec:
Using https://github.com/status-im/nim-stew/commit/fecc3cc5cbc1bb568dffc96b9bca46db724ce41c and https://github.com/status-im/nim-stew/commit/419903c9a31ab253cf5cf19f24d9a912dc4b5154

This should probably not go in until the beginning of the next release cycle, because it will unmask other possible `checks`-triggering code across the `nimbus-eth2` codebase.


Tuneabq (approved):

Tuneabq:
✅

PR #3557, opened 2022-03-30
integrate light client into beacon node

etan-status:
Adds a `LightClient` instance to the beacon node as preparation to
accelerate syncing in the future (optimistic sync).

- `--light-client-enable` turns on the feature
- `--light-client-trusted-block-root` configures block to start from

If no block root is configured, light client tracks DAG `finalizedHead`.

### DIFF
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -33,18 +33,18 @@ proc initLightClient*(
     node.network, rng, config, cfg,
     forkDigests, getBeaconTime, genesis_validators_root)
 
-  if config.lightClientEnable.get:
+  if config.lightClientEnable:
     lightClient.trustedBlockRoot = config.lightClientTrustedBlockRoot
 
   elif config.lightClientTrustedBlockRoot.isSome:
     warn "Ignoring `lightClientTrustedBlockRoot`, light client not enabled",
-      lightClientEnable = config.lightClientEnable.get,
+      lightClientEnable = config.lightClientEnable,
       lightClientTrustedBlockRoot = config.lightClientTrustedBlockRoot
 
   node.lightClient = lightClient
 
 proc startLightClient*(node: BeaconNode) =
-  if not node.config.lightClientEnable.get:
+  if not node.config.lightClientEnable:
     return
 
   node.lightClient.start()
@@ -54,7 +54,7 @@ proc installLightClientMessageValidators*(node: BeaconNode) =
     if node.config.serveLightClientData.get:
       # Process gossip using both full node and light client
       node.processor
-    elif node.config.lightClientEnable.get:
+    elif node.config.lightClientEnable:
       # Only process gossip using light client
       nil
     else:
@@ -76,7 +76,7 @@ proc updateLightClientGossipStatus*(
   node.lightClient.updateGossipStatus(slot, some isBehind)
 
 proc updateLightClientFromDag*(node: BeaconNode) =
-  if not node.config.lightClientEnable.get:
+  if not node.config.lightClientEnable:
     return
   if node.config.lightClientTrustedBlockRoot.isSome:
     return
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -80,7 +80,7 @@ proc updateLightClientFromDag*(node: BeaconNode) =
   let
     dagHead = node.dag.finalizedHead
     dagPeriod = dagHead.slot.sync_committee_period
-  if dagHead.slot.epoch < node.dag.cfg.ALTAIR_FORK_EPOCH:
+  if dagHead.slot < node.dag.cfg.ALTAIR_FORK_EPOCH.start_slot:
     return
 
   let lcHeader = node.lightClient.finalizedHeader
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -1,0 +1,99 @@
+# beacon_chain
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+# This implements the pre-release proposal of the libp2p based light client sync
+# protocol. See https://github.com/ethereum/consensus-specs/pull/2802
+
+import
+  chronicles,
+  ./beacon_node
+
+logScope: topics = "beacnde"
+
+proc initLightClient*(
+    node: BeaconNode,
+    rng: ref BrHmacDrbgContext,
+    cfg: RuntimeConfig,
+    forkDigests: ref ForkDigests,
+    getBeaconTime: GetBeaconTimeFn,
+    genesis_validators_root: Eth2Digest) =
+  template config(): auto = node.config
+
+  let lightClient = createLightClient(
+    node.network, rng, config, cfg,
+    forkDigests, getBeaconTime, genesis_validators_root)
+
+  if config.lightClientEnable.get:
+    lightClient.trustedBlockRoot = config.lightClientTrustedBlockRoot
+
+  elif config.lightClientTrustedBlockRoot.isSome:
+    warn "Ignoring `lightClientTrustedBlockRoot`, light client not enabled",
+      lightClientEnable = config.lightClientEnable.get,
+      lightClientTrustedBlockRoot = config.lightClientTrustedBlockRoot
+
+  node.lightClient = lightClient
+
+proc startLightClient*(node: BeaconNode) =
+  if not node.config.lightClientEnable.get:
+    return
+
+  node.lightClient.start()
+
+proc installLightClientMessageValidators*(node: BeaconNode) =
+  let eth2Processor =
+    if node.config.serveLightClientData.get:
+      # Process gossip using both full node and light client
+      node.processor
+    elif node.config.lightClientEnable.get:
+      # Only process gossip using light client
+      nil
+    else:
+      # Light client topics will never be subscribed to, no validators needed
+      return
+
+  node.lightClient.installMessageValidators(eth2Processor)
+
+proc updateLightClientGossipStatus*(
+    node: BeaconNode, slot: Slot, dagIsBehind: bool) =
+  let isBehind =
+    if node.config.serveLightClientData.get:
+      # Forward DAG's readiness to handle light client gossip
+      dagIsBehind
+    else:
+      # Full node is not interested in gossip
+      true
+
+  node.lightClient.updateGossipStatus(slot, some isBehind)
+
+proc updateLightClientFromDag*(node: BeaconNode) =
+  if not node.config.lightClientEnable.get:
+    return
+  if node.config.lightClientTrustedBlockRoot.isSome:
+    return
+
+  let
+    dagHead = node.dag.finalizedHead
+    dagPeriod = dagHead.slot.sync_committee_period
+  if dagHead.slot.epoch < node.dag.cfg.ALTAIR_FORK_EPOCH:
+    return
+
+  let lcHeader = node.lightClient.finalizedHeader
+  if lcHeader.isSome:
+    if dagPeriod <= lcHeader.get.slot.sync_committee_period:
+      return
+
+  let
+    bdata = node.dag.getForkedBlock(dagHead.blck.bid).valueOr:
+      return
+    header = bdata.toBeaconBlockHeader
+    current_sync_committee = block:
+      var tmpState = assignClone(node.dag.headState)
+      node.dag.currentSyncCommitteeForPeriod(tmpState[], dagPeriod).valueOr:
+        return
+  node.lightClient.resetToFinalizedHeader(header, current_sync_committee)
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -33,18 +33,18 @@ proc initLightClient*(
     node.network, rng, config, cfg,
     forkDigests, getBeaconTime, genesis_validators_root)
 
-  if config.lightClientEnable:
+  if config.lightClientEnable.get:
     lightClient.trustedBlockRoot = config.lightClientTrustedBlockRoot
 
   elif config.lightClientTrustedBlockRoot.isSome:
     warn "Ignoring `lightClientTrustedBlockRoot`, light client not enabled",
-      lightClientEnable = config.lightClientEnable,
+      lightClientEnable = config.lightClientEnable.get,
       lightClientTrustedBlockRoot = config.lightClientTrustedBlockRoot
 
   node.lightClient = lightClient
 
 proc startLightClient*(node: BeaconNode) =
-  if not node.config.lightClientEnable:
+  if not node.config.lightClientEnable.get:
     return
 
   node.lightClient.start()
@@ -54,7 +54,7 @@ proc installLightClientMessageValidators*(node: BeaconNode) =
     if node.config.serveLightClientData.get:
       # Process gossip using both full node and light client
       node.processor
-    elif node.config.lightClientEnable:
+    elif node.config.lightClientEnable.get:
       # Only process gossip using light client
       nil
     else:
@@ -76,7 +76,7 @@ proc updateLightClientGossipStatus*(
   node.lightClient.updateGossipStatus(slot, some isBehind)
 
 proc updateLightClientFromDag*(node: BeaconNode) =
-  if not node.config.lightClientEnable:
+  if not node.config.lightClientEnable.get:
     return
   if node.config.lightClientTrustedBlockRoot.isSome:
     return
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -25,6 +25,10 @@ proc initLightClient*(
     genesis_validators_root: Eth2Digest) =
   template config(): auto = node.config
 
+  # Creating a light client is not dependent on `lightClientEnable`
+  # because the light client module also handles gossip subscriptions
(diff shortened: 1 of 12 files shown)


beacon_chain/beacon_node_light_client.nim
```diff
@@ -0,0 +1,85 @@
...
+    node: BeaconNode,
+    cfg: RuntimeConfig,
+    rng: ref BrHmacDrbgContext,
+    genesisValidatorsRoot: Eth2Digest,
+    getBeaconTime: GetBeaconTimeFn) =
+  template config(): auto = node.config
+
+  let store = (ref Option[LightClientStore])()
+
+  proc getTrustedBlockRoot(): Option[Eth2Digest] =
```
tersec: Some of these (this, `isLightClientStoreInitialized`, `isNextSyncCommitteeKnown`, etc) can probably be `func` rather than `proc`.

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,369 @@
...
+          currentPeriod in [finalizedPeriod, finalizedPeriod + 1]
+        else:
+          currentPeriod == finalizedPeriod
+
+    # If not in sync, request light client data for older periods
+    if not isInSync:
+      doAssert currentPeriod > finalizedPeriod
+      let didProgress = await self.query(
+        BestLightClientUpdatesByRangeEndpoint,
+        finalizedPeriod .. (currentPeriod - 1))
```
tersec: Interesting, reason for not using `..<` ?

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,369 @@
...
+      var applyReward = false
+      for val in value.get.values:
+        let res = await self.valueVerifier(E)(val)
+        if res.isErr:
+          case res.error
+          of BlockError.MissingParent:
+            # Stop, requires different request to progress
+            return didProgress
+          of BlockError.Duplicate:
+            # Ignore, a concurrent request may have queryed this already
```
tersec:
```suggestion
            # Ignore, a concurrent request may have queried this already
```

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,369 @@
...
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+# This implements the pre-release proposal of the libp2p based light client sync
+# protocol. See https://github.com/ethereum/consensus-specs/pull/2802
+
+import chronos, chronicles, stew/base10
+import
+  eth/p2p/discoveryv5/random2,
```
tersec: Broader comment, and I know you're just following the usual idioms here, but it's absurd that the PRNG lives in `eth/p2p/discoveryv5/random2`.

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,369 @@
...
+  let
+    startPeriod = periods.a
+    lastPeriod = periods.b
+    reqCount = min(periods.len, MAX_REQUEST_LIGHT_CLIENT_UPDATES)
+  let response =
+    await peer.bestLightClientUpdatesByRange(startPeriod, reqCount.uint64)
+  if response.isOk:
+    if response.get.len > reqCount:
+      raise newException(ResponseError, "Too many values in response" &
+        " (" & Base10.toString(response.get.len.uint) &
```
tersec: There's a `lenu64` for this sort of thing.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -665,10 +665,8 @@ proc init*(T: type BeaconNode,
     validatorMonitor: validatorMonitor,
     stateTtlCache: stateTtlCache
   )
-
-  node.initFullNode(
-    rng, dag, taskpool, getBeaconTime)
-
+  node.initLightClient(cfg, rng, dag.genesisValidatorsRoot, getBeaconTime)
```
tersec: What's the effect of always doing these, even if the intention isn't to run a light client node?

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,369 @@
...
+  when E.V is altair.LightClientBootstrap:
+    self.bootstrapVerifier
+  elif E.V is altair.LightClientUpdate:
+    self.updateVerifier
+  elif E.V is OptimisticLightClientUpdate:
+    self.optimisticUpdateVerifier
+  else: static: doAssert false
+
+iterator values(v: auto): auto =
+  when v is seq:
```
tersec: Does `openArray` work here? In general, it's nicer to allow that, if it doesn't break things.

tersec: The version of Nim we're using, unfortunately, doesn't have a robust notion of concepts. Maybe after the Nim 1.6 transition, this kind of thing can stop checking for types and start using concepts.

beacon_chain/nimbus_beacon_node.nim
etan-status: There was pushback against introducing states where `node.dag` was uninitialized. This follows the same practice for `node.lightClient`. The light client facilities are initialized here, but not started. This means that objects obtained through gossip / rest / whatever could be processed without having to check for the current operation mode (light vs full). It will still error out, similar to asking for proposing a new block when the DAG is not sufficiently synced.

beacon_chain/sync/light_client_manager.nim
etan-status: No reason, just forgot about `..<`. Thanks for the reminder!

beacon_chain/sync/light_client_manager.nim
etan-status: Hmm, that will force it to use the 64-bit path though, even on 32-bit platform, and it will break in a 128-bit world. But granted, it's neither a hot path where it matters, nor a practical problem. Switching to `lenu64` for consistency.

tersec: Sure, so one possibility is that one could create and use a `lenu` or similar. Part of the intention there is to be more careful with integer type conversions, and `len` is a function with a range that's non-negative, while general `uint` to `int` conversions aren't safe and can trigger Nim `checks`.

I view it more as a compile-time type-safety thing, though, of course. it'd be implemented in the same way as you did here, just factored out. The goal though is create abstractions which don't require potentially runtime-unsafe integer conversions, and the 64-bit codepath is an acceptable tradeoff for that in a non-hot-path.

beacon_chain/sync/light_client_manager.nim
etan-status: Doesn't work for `openArray` - extended it with a comment to highlight the local purpose of this helper.

etan-status:
- add comment for `values` iterator purpose
- `proc` -> `func` for callback implementations
- `len.uint` -> `lenu64` for length logging
- `a .. (b - 1)` -> `a ..< b` for conciseness
- `queryed` -> `queried` typo

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,371 @@
...
+      let didProgress = cast[Future[bool]](future).read()
+      if didProgress:
+        anyDidProgress = true
+        cancelAll()
+    except CatchableError as exc:
+      cancelAll()
+
+  for i in 0 ..< workers.len:
+    workers[i].addCallback(handleFinishedWorker)
+  await allFutures(workers)
```
tersec: Is there any timeout in the req/resp protocol to prevent a peer from stalling this indefinitely?

beacon_chain/sync/light_client_manager.nim
```diff
@@ -0,0 +1,371 @@
...
+            return didProgress
+          of BlockError.Duplicate:
+            # Ignore, a concurrent request may have queried this already
+            when E.V is altair.LightClientBootstrap:
+              didProgress = true
+            else:
+              discard
+          of BlockError.UnviableFork:
+            # Descore, peer is on an incompatible fork version
+            notice "Received value from an unviable fork", value = val.shortLog,
```
tersec: This `notice` and the `Received invalid value` `warn` are both pretty visible logging for what shouldn't be a particularly catastrophic or unexpected networking glitch, I'd have thought?

beacon_chain/sync/light_client_manager.nim
tersec: Especially since it looks like, might well be missing something here, that
https://github.com/status-im/nimbus-eth2/blob/b290ddff13859df26b788808b529bd10df1ad07a/beacon_chain/sync/light_client_manager.nim#L305-L310
means that in the meantime, the peer will be effectively otherwise removed from the peer pool, in a sort of periodic, hopefully short-lived way, but still.

etan-status: Yeah, the generated `sync_protocol` functions have a default timeout of 10 seconds, for example:

```
proc latestLightClientUpdate*(peer: Peer;
                              timeout: Duration = milliseconds(10000'i64)): Future[
    NetRes[altair.LightClientUpdate]] {.gcsafe, libp2pProtocol(
    "latest_light_client_update", 0, isLightClientRequest = true).} =
  var msgBytes: seq[byte]
  makeEth2Request(peer, "/eth2/beacon_chain/req/latest_light_client_update/0/",
                  msgBytes, altair.LightClientUpdate, timeout)
```

If peer does not answer in time, this should lead to `PeerScoreNoBlocks` descore being applied to them.
Also, a new `peer = await self.network.peerPool.acquire()` is selected on every retry.

This will lead to gradual selection of peers that produce successful results.

Similar logic is implemented in the existing `request_manager.nim`.

beacon_chain/sync/light_client_manager.nim
etan-status: I used the same log levels as in `request_manager.nim`. Likewise, for logging before applying the descore vs after.

https://github.com/status-im/nimbus-eth2/blob/c7abc975450cd311f715540c957c277c5b258bb3/beacon_chain/sync/request_manager.nim#L122-L125

beacon_chain/sync/light_client_manager.nim
tersec: Alright.

beacon_chain/sync/light_client_manager.nim
tersec: Okay, I guess I'm not convinced about them either then, but that's outside the scope of this PR.

tersec (approved):

beacon_chain/sync/light_client_manager.nim
etan-status: Likewise 🙃

etan-status:
Note that this PR does not yet take into account GossipSub data. Will submit that separately as it's already quite noisy. GossipSub data can be used to keep `store.optimistic_header` updated.

beacon_chain/beacon_node_light_client.nim
```diff
@@ -0,0 +1,85 @@
...
+# This implements the pre-release proposal of the libp2p based light client sync
+# protocol. See https://github.com/ethereum/consensus-specs/pull/2802
+
+import
+  "."/beacon_node,
+  ./gossip_processing/light_client_processor,
+  ./spec/datatypes/altair,
+  ./sync/light_client_manager
+
+proc initLightClient*(
```
zah, line 19: Can we make a `LightClient` object that can be instantiated without a `BeaconNode`? Imagine trying to create the smallest possible binary that implement the light client sync algorithm.

etan-status:
- Rebase to `unstable` (`genesisValidatorsRoot` -> `genesis_validators_Root`).

beacon_chain/beacon_node_light_client.nim
etan-status: Yeah, the `LightClientManager` and `LightClientProcessor` do not know the `BeaconNode`, but we need it here to interface with the full sync, share the peer pool and the config, etc. `beacon_node_light_client.nim` is the only layer that knows about the `beacon_node`, it only contains the plumbing, not the actual logic of the "light client sync algorithm."

etan-status: https://github.com/status-im/nimbus-eth2/pull/3653

etan-status:
- Reworked to use light client library from https://github.com/status-im/nimbus-eth2/pull/3653
- Attach to gossip
- Sync from `dag.finalizedHead` if no `light-client-trusted-block-root` is configured

beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -322,92 +356,74 @@ proc addObject*(
...
+      # [IGNORE] The `finality_update` advances the `finalized_header` of the
+      # local `LightClientStore`.
+      errIgnore(typeof(obj).name & ": no significant progress")
   else:
-    case v.error
+    case r.error
     of BlockError.Invalid:
-      # [REJECT] The `finality_update` / `optimistic_update` is valid.
-      errReject($v.error)
+      # [REJECT] The `finality_update` is valid.
```
tersec, line 393: Is invalid (also for `optimistic_update` on the next line)?

beacon_chain/conf.nim
```diff
@@ -266,6 +266,16 @@ type
         desc: "Weak subjectivity checkpoint in the format block_root:epoch_number"
         name: "weak-subjectivity-checkpoint" }: Option[Checkpoint]
 
+      lightClientEnable* {.
+        hidden
+        desc: "BETA: Accelerate sync using light client."
+        name: "light-client-enable" .}: Option[bool]
```
tersec, line 272: How does this `Option`/`applyConfigDefault` help?

https://github.com/status-im/nimbus-eth2/blob/43876fc7eddb9a2e10060d822a59869c5aaa6a36/beacon_chain/nimbus_beacon_node.nim#L1699-L1709

show the mechanism, but it's not clear to me why this version (see also `lightClientTrustedBlockRoot`) is better than the more straightforward `Option`-free variation which sets some appropriate default and avoids the `Option`-`.get()` indirection.

beacon_chain/light_client.nim
```diff
@@ -167,36 +167,123 @@ proc start*(lightClient: LightClient) =
...
+    let res1 =
+      if eth2Processor != nil:
+        let
+          v = eth2Processor[].`validatorProcName`(MsgSource.gossip, msg)
+          res = v.toValidationResult()
+        if res == ValidationResult.Reject:
+          msg.logDropped(v.error)
+          return res
+        if res == ValidationResult.Ignore:
+          error = v.error
```
tersec: If `not eth2Processor.isNil and lightClient.manager.isGossipSupported(getLocalWallPeriod())`, it's possible for `error` to be set twice, thus overwriting whatever the Eth2Processor ignore reason was. `ValidationError` does need to be a `cstring`, but not sure if one's going to log this at all, whether it makes sense to log reasons from both.

There are definitely concerns about over-logging, but the choice here of which to prefer (the second ignore message, implicitly) is a bit arbitrary, and doesn't help the user form a completely predictive mental model of how LC gossip ignore processing works.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -682,9 +683,10 @@ proc init*(T: type BeaconNode,
     gossipState: {},
     beaconClock: beaconClock,
     validatorMonitor: validatorMonitor,
-    stateTtlCache: stateTtlCache
-  )
+    stateTtlCache: stateTtlCache)
 
+  node.initLightClient(
```
tersec, line 688: If `startLightClient()` rejects the light client starting due to settings, does it still make sense to `initLightClient`? My recollection is the additional resource usage is minimal, and it's possible that just always initializing the light client minimally avoids other special cases throughout the code, or will as it's more integrated.

beacon_chain/gossip_processing/light_client_processor.nim
etan-status: Those conditions, while visually appearing as "Reject if x", they actually all mean the opposite, as the spec introduces them with `the following validations MUST additionally pass before forwarding`. This is not a light client specific issue.

e.g., in `gossip_validation`:
```  if msg.message.contribution.aggregation_bits.countOnes() == 0:
    # [REJECT] The contribution has participants
    # that is, any(contribution.aggregation_bits).
    return errReject("SignedContributionAndProof: aggregation bits empty")
```

beacon_chain/conf.nim
etan-status: Had it at default-on for kiln/ropsten/prater, but you are right that now with the default off the straight-forward `bool` would also work. I guess when optimistic sync is fully implemented, that this would need to be changed to detect the case of "no user preference" vs "user selected off" and apply network specific default in first case. I can get rid of it once more and it can then be reintroduced later, or we can just keep it as is, up to you.

beacon_chain/light_client.nim
etan-status: Yeah, if all valide functions decide to IGNORE, it would be better to log all of the reasons. Will change it like that, agree that this makes it more straight forward when debugging.

beacon_chain/nimbus_beacon_node.nim
etan-status: Yeah, because `installMessageValidators` / `updateGossipStatus` also take over some full node duties in the light client specific topics, so having the object always initialized makes the logic easier there.

etan-status:
During the Ropsten change from Altair -> Bellatrix, noticed in the logs that the LC optimistic update for the first slot was not picked up on Gossip, due to the subscription on old net being dropped immediately after bellatrix. Light client uses `attested_header` to derive fork digest, but sends the message based on a future `signature_slot`, so the topics need to be available some longer. Will change it so that Altair light client subscriptions stay around for the first bellatrix epoch as well, to avoid missing the first slot.

beacon_chain/gossip_processing/light_client_processor.nim
tersec: Oh, indeed.

beacon_chain/conf.nim
tersec: Alright, keep as is.

beacon_chain/light_client.nim
tersec: Done via https://github.com/status-im/nimbus-eth2/pull/3557/commits/1eb2e18dcdfd1bcf5af6f3c6f8fe01a1b1de1672

PR #3558, opened 2022-03-30
handle dynamic IP changes

stefantalpalaru:
If the external IP was detected through UPnP/NAT-PMP, we check
periodically to see if it changes, then we inform the discovery subsystem
about it.

"nim-eth" is being bumped because it exports a new proc we need in here.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1410,8 +1410,7 @@ proc new*(T: type Eth2Node, config: BeaconNodeConf, runtimeCfg: RuntimeConfig,
           getBeaconTime: GetBeaconTimeFn, switch: Switch,
           pubsub: GossipSub, ip: Option[ValidIpAddress], tcpPort,
           udpPort: Option[Port], privKey: keys.PrivateKey, discovery: bool,
-          rng: ref BrHmacDrbgContext, externalIp: Option[ValidIpAddress],
-          externalTcpPort, externalUdpPort: Option[Port]): T {.raises: [Defect, CatchableError].} =
+          rng: ref BrHmacDrbgContext): T {.raises: [Defect, CatchableError].} =
   when not defined(local_testnet):
     let
       connectTimeout = 1.minutes
@@ -1453,9 +1452,9 @@ proc new*(T: type Eth2Node, config: BeaconNodeConf, runtimeCfg: RuntimeConfig,
     rng: rng,
     connectTimeout: connectTimeout,
     seenThreshold: seenThreshold,
-    externalIp: externalIp,
-    externalTcpPort: externalTcpPort,
-    externalUdpPort: externalUdpPort,
+    externalIp: ip,
+    externalTcpPort: tcpPort,
+    externalUdpPort: udpPort,
   )
 
   newSeq node.protocolStates, allProtocols.len
@@ -2101,9 +2100,7 @@ proc createEth2Node*(rng: ref BrHmacDrbgContext,
   let node = Eth2Node.new(
     config, cfg, enrForkId, discoveryForkId, forkDigests, getBeaconTime, switch, pubsub, extIp,
     extTcpPort, extUdpPort, netKeys.seckey.asEthKey,
-    discovery = config.discv5Enabled, rng = rng,
-    externalIp = extIp, externalTcpPort = extTcpPort,
-    externalUdpPort = extUdpPort)
+    discovery = config.discv5Enabled, rng = rng)
 
   node.pubsub.subscriptionValidator =
     proc(topic: string): bool {.gcsafe, raises: [Defect].} =
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -84,6 +84,10 @@ type
     peerTrimmerHeartbeatFut: Future[void]
     cfg: RuntimeConfig
     getBeaconTime: GetBeaconTimeFn
+    externalIp: Option[ValidIpAddress]
+    externalTcpPort: Option[Port]
+    externalUdpPort: Option[Port]
+    externalIpUpdateFut: Future[void]
 
   EthereumNode = Eth2Node # needed for the definitions in p2p_backends_helpers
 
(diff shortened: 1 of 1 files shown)


Menduist:
Just a note, we can also detect IP changes with the libp2p identify protocol (which is used everytime we dial someone/someone dials us)
Currently it's not hooked up, need to be a bit careful because peers can lie to us or other, but it's fully async so at least it won't block the main loop

LMK if you want me to work a bit on that

stefantalpalaru:
Isn't that the already existing ENR auto update?

Menduist:
I think the current auto update comes from the DHT, but don't quote me on that

stefantalpalaru:
Right, but it's the same level of peer trusting which can't be enabled by default. This can.

beacon_chain/networking/eth2_network.nim
```diff
@@ -1406,7 +1410,8 @@ proc new*(T: type Eth2Node, config: BeaconNodeConf, runtimeCfg: RuntimeConfig,
           getBeaconTime: GetBeaconTimeFn, switch: Switch,
           pubsub: GossipSub, ip: Option[ValidIpAddress], tcpPort,
           udpPort: Option[Port], privKey: keys.PrivateKey, discovery: bool,
-          rng: ref BrHmacDrbgContext): T {.raises: [Defect, CatchableError].} =
+          rng: ref BrHmacDrbgContext, externalIp: Option[ValidIpAddress],
+          externalTcpPort, externalUdpPort: Option[Port]): T {.raises: [Defect, CatchableError].} =
```
kdeme: The ip, udpPort and tcpPort parameters are already considered as external ip and ports. They are the ones that end up in the ENR that is used by discoveryv5. So I think you can just reuse those parameters to set to the object fields.

beacon_chain/networking/eth2_network.nim
```diff
@@ -2073,12 +2101,18 @@ proc createEth2Node*(rng: ref BrHmacDrbgContext,
...
-    discovery = config.discv5Enabled, rng = rng)
+    discovery = config.discv5Enabled, rng = rng,
+    externalIp = extIp, externalTcpPort = extTcpPort,
+    externalUdpPort = extUdpPort)
 
   node.pubsub.subscriptionValidator =
     proc(topic: string): bool {.gcsafe, raises: [Defect].} =
       topic in node.validTopics
 
+  if extIp.isSome() and not config.nat.hasExtIp and config.nat.nat in [NatAny, NatUpnp, NatPmp]:
```
kdeme, line 2107: The downside of this way of selecting it is that if `NatAny` is selected, and this is the default, but there is no actual UPnP or NAT-PMP available, this loop will keep trying to get the external IP every 10 mins.

beacon_chain/networking/eth2_network.nim
```diff
@@ -1967,6 +1978,25 @@ template gossipMaxSize(T: untyped): uint32 =
   static: doAssert maxSize <= maxGossipMaxSize()
   maxSize.uint32
 
+proc externalIpUpdateLoop(node: Eth2Node,
+                      config: BeaconNodeConf) {.async.} =
+  try:
+    while true:
+      # Trade-off between reacting quickly to dynamic IP changes (rather rare)
+      # and blocking the main event loop while we chat with the router.
```
kdeme, line 1985: How long does this block if the request times out?

beacon_chain/networking/eth2_network.nim
stefantalpalaru: Not an issue, because `extIp.isSome()` is only true if one of those mechanisms worked and gave us an external IP.

beacon_chain/networking/eth2_network.nim
stefantalpalaru: 3.755s: https://github.com/miniupnp/libnatpmp/pull/15
and https://github.com/status-im/nimbus-build-system/blob/a8ab2dc39aad4d69ba3be72868772d851b4b9741/makefiles/targets.mk#L140

beacon_chain/networking/eth2_network.nim
stefantalpalaru: Fixed.

beacon_chain/networking/eth2_network.nim
kdeme: That would be true for the `setupNat` proc, but with `setupAddress` the bindIp and the best route IP are checked whether they are a public IP address.

beacon_chain/networking/eth2_network.nim
```diff
@@ -1967,6 +1977,25 @@ template gossipMaxSize(T: untyped): uint32 =
...
+      await sleepAsync(10.minutes)
+
+      let extIpRes = getExternalIP(config.nat.nat, quiet = true)
```
arnetheduck: this is a synchronous operation? it freezes until router responds?

beacon_chain/networking/eth2_network.nim
arnetheduck: 3-4 s is not acceptable, that's enough to lose an attestation slot

beacon_chain/networking/eth2_network.nim
stefantalpalaru: Yes.

beacon_chain/networking/eth2_network.nim
stefantalpalaru: How about adding the same lookup for a public preferred routing source in `getExternalIP()`, since the "nat" module is no longer just about NAT?

stefantalpalaru: This is what I came up with: https://github.com/status-im/nim-eth/pull/496

stefantalpalaru:
Prevented delay when a public IP was detected from the routing table.

arnetheduck (changes requested):
This PR needs updating such that it doesn't block for seconds on the unhappy path (timeout) as this can disrupt attestation work due to how long the timeouts are

arnetheduck:
closing this since external dynamic ip changes are (mostly) handled by `enr-auto-update` - an async version of this PR would however be nice / accepted

PR #3559, opened 2022-03-30
serve light client data on `prater` by default

etan-status:
Applies a `prater` testnet specific config default to serve light client
data on that network.

### DIFF
--- a/beacon_chain/networking/network_metadata.nim
+++ b/beacon_chain/networking/network_metadata.nim
@@ -9,7 +9,7 @@
 
 import
   std/[sequtils, strutils, os],
-  stew/shims/macros, nimcrypto/hash,
+  stew/byteutils, stew/shims/macros, nimcrypto/hash,
   eth/common/eth_types as commonEthTypes,
   web3/[ethtypes, conversions],
   chronicles,
@@ -17,6 +17,9 @@ import
   ssz_serialization/navigator,
   ../spec/eth2_ssz_serialization,
   ../spec/datatypes/phase0
+
+from ../consensus_object_pools/block_pools_types_light_client
+  import ImportLightClientData
 
 # ATTENTION! This file will produce a large C file, because we are inlining
 # genesis states as C literals in the generated code (and blobs in the final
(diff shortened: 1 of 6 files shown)


etan-status:
Followup from https://github.com/status-im/infra-nimbus/pull/91

beacon_chain/networking/network_metadata.nim
```diff
@@ -143,14 +153,29 @@ proc loadEth2NetworkMetadata*(path: string, eth1Network = none(Eth1Network)): Et
       else:
         ""
 
+      configDefaults =
+        if path.endsWith("/prater"):
```
etan-status: @tersec : When merging this to `kiln-dev-auth`, please also add `/kiln` here.

etan-status:
- Fix test suite

beacon_chain/networking/network_metadata.nim
tersec: Sure

tersec (approved):

beacon_chain/networking/network_metadata.nim
zah: Er, this is not a good way to check which network you are connected to. The best way is to check the hash of the genesis block in the selected data dir. I've used this in a work-in-progress PR that I can submit today.

tersec:
A general note/caveat: while there are a number of locally-unprotected `get`s here, the BN-based codepaths look okay, and while there are design alternatives to this approach, they have their own tradeoffs.

beacon_chain/networking/network_metadata.nim
etan-status: Cool, if you have a better way to detect, let's use that one!

etan-status:
- Use first 40 bytes of `genesisData` (`genesis_time` and `genesis_validators_root`) to identify network instead of `--network` folder name.

beacon_chain/networking/network_metadata.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3494/commits/e88bc1b3500e921f9ddd7e66ce156dc13c6a1d6b

PR #3560, opened 2022-03-30
Add Nimble lock file

zah:


nimble.lock
```diff
@@ -0,0 +1,554 @@
+{
+  "version": 1,
+  "packages": {
+    "stew": {
+      "version": "0.1.0",
+      "vcsRevision": "bb705bf17b46d2c8f9bfb106d9cc7437009a2501",
```
stefantalpalaru, line 6: No branch specified?

Does this mean it fetches all branches, or that it only works if the target commit is on the default branch?

zah: When it works with a local check out, It uses an algorithm where it first fetches the current branch. If it doesn't contain the referenced revision, it will fetch other branches and other remotes until it finds the revision (or fails).

arnetheduck: What's the workflow now to get a fully "local checkout"? ie with `vendor` the project is fully self-contained with no annoying global caches etc polluting the machine or home folder - doing a checkout, building then removing the checkout folder does not leave any cruft.

nimble.lock
```diff
@@ -0,0 +1,554 @@
+{
```
arnetheduck, line 1: I don't see `nim` itself in the list - what am I missing? Nim and its std lib is a dependency like any other and needs to be locked as such.

beacon_chain.nimble
```diff
@@ -0,0 +1,84 @@
...
+  "taskpools",
+  "testutils",
+  "unicodedb >= 0.10",
+  "unittest2",
+  "web3",
+  "zlib",
+  "zxcvbn"
+)
+
+requires "https://gitlab.com/status-im/nimbus-security-resources.git"
```
arnetheduck, line 44: why is this separate?

beacon_chain.nimble
```diff
@@ -0,0 +1,84 @@
+mode = ScriptMode.Verbose
+
+version       = "1.4.2"
+author        = "Status Research & Development GmbH"
+description   = "The Nimbus beacon chain node is a highly efficient Ethereum 2.0 client"
+license       = "MIT or Apache License 2.0"
+
+requires(
```
arnetheduck, line 8: compared to the `vendor` folder, there are many deps that are not covered - what shall we do with those? ditto for test-only deps (really, per-named-binary deps).

zah: Transitive dependencies are discovered by Nimble. There are few ways to handle binary-specific dependencies that haven't been fully validated yet (e.g. having a binary-specific nimble file or a build-specific lock file).

beacon_chain.nimble
zah: This holds some large files that we don't want to store in the Nimbus repo. It uses an URL because it seemed inappropriate to pollute the official Nimble package namespace with such a package.

nimble.lock
zah: It's true that we can add Nim as a dependency here. Will do that in a follow up commit. The Nim binary is controller in the usual way through a script that builds and caches a specific revision.

arnetheduck: the `nim` used to build the software and to run the script does not have to be the same - ie one can imagine that you have a "global" nim in your installation that is of some random sufficient version - then, there are also side-by-side installations of other nim versions (just like any other library dependency) and these are used to actually build the stuff given in the lock file.

Probably it's appropriate to print a notice when nim-to-run-script != nim-that-builds-software because this might introduce subtle issues and security problems, but that seems like a reasonable way to "bootstrap" things from a nimble perspective. The alternative of course is to simply refuse to build with a different nim version unless a make-it-work flag is given.

beacon_chain.nimble
arnetheduck: Oh. Yaml also uses a URL - other than that, might be worth documenting that in the file because it looks odd.

nimble.lock
zah: I need to make one pass over the Nimble documentation to explain more clearly how commands such as `nimble lock`, `nimble sync` and `nimble develop` are supposed to be used. You can try to read the descriptions that @bobeff contributed here:

https://github.com/nim-lang/nimble

Users of Nimble will be working in a special "workspace" which is something like a shared `vendor` folder for all of your projects:

https://github.com/status-im/nim-workspace

You may find the following explanation helpful in order to understand the structure and the workflow in this repository:

https://forum.nim-lang.org/t/6738#41857

arnetheduck: Well, there's been a lot of back-and-forth regarding this both here and in the nimble forums, chats, repositories, documentations and so on so it's a bit of a mess to follow.

Notably though, such shared vendor folders an anti-feature when enabled by default because they prevent working on several projects in parallel, ie such sharing would ideally be explicitly enabled if desired. 

It's not reasonable or even feasible to expect that all projects at status for example will use the same versions of things (this would become a bottleneck for the organization as a whole), so a single workspace for _all_ teams doesn't make sense. It is a priority however want to make it easy to work on multiple projects at the same time (nimbus/waku etc), and these will each have their own nim-eth, nim-stew and so on, and the way to achieve that is through independent "workspaces" being the most easy and cheap thing to do rather than an exception that one must read long manuals to achieve. 

It's also problematic for users in the following way: they check out nimbus-eth2 to build from source and run into a problem - when the checkout is self-contained, one can wipe it, re-clone and it's almost guaranteed to work - with magic locations, this becomes more difficult because now they must learn a complex workflow to get rid of it.

zah: > Notably though, such shared vendor folders an anti-feature when enabled by default because they prevent working on several projects in parallel, ie such sharing would ideally be explicitly enabled if desired.

I disagree. The fact that libraries are not shared has been a source of significant pain when you need to verify that a change in a library works OK in multiple projects. With develop files you can still temporarily or permanently decide that a particular project will use a separate unique checkout of a library, but you get to decide when this is needed.

The precise revision control provided by lockfiles ensures that the other problems you are alluding to are non-problems.

zah: Also, if you prefer to work with multiple workspaces, this is extremely easy to do. Just clone the workspace repo multiple times. You should not impose your own preferences on others though.

arnetheduck: >  You should not impose your own preferences on others though.

This is not a personal preference but rather an objective statement - it's an acknowledgement of reality that all our major projects have reached the point where they use different versions of libraries, have different upgrade and release cycles - take a look at dagger, waku and desktop and you'll see that they are all on different versions of things like libp2p, chronos, stew and so on, and when these core libs get updated, they don't get tested with all this projects by whoever makes the change: rather, they get tested on a few, then it's the responsibility of each team to upgrade their dependencies when the time is right to do so. We don't want a dagger developer to upgrade libp2p in the feature freeze period of nimbus-eth2, and we don't want to impose the release cycle of nimbus-eth2 onto dagger and vice versa.

This is a good thing - it's a sign of maturity that we have decentralised development and it should be encouraged - this is indeed why we cut up teams once they reach a certain size: they _should_ have this autonomy and independence.

Nimbus is a special case: eth1 and eth2 _should_ really live in a single repo: that's why it feels like you want them in a single workspace: they are de-facto a single project more and more - but you'll notice that fluffy for example is _not_: their release and upgrade schedule is completely different from that of eth1 and it's mostly a historical accident that the two projects use the same repo and therefor suffer the same module churn.

More broadly, this applies to projects outside of status, but even more so: if they want to use our code (for example libp2p), it's not realistic to expect that _they_ will want to clone our master workspace and be beholden to our "central committee of library upgrades". Lock files exist to disconnect release and upgrade cycles from each other, not to bind them more tightly together.

arnetheduck: > change in a library works OK in multiple projects

Which project a library upgrade should be tested on is a personal preference based on familiarity with the different projects and codebases - this is why it should be a personal and explicit choice which libraries should be shared between projects - it'll be different for every developer depending on what they're working on, and we don't want a situation where "accidental" changes leak from one project to another.

zah: I'm not sure why you decided to write this long message. As I explained, nothing in the design of Nimble or the workspace repository prevents these decoupled release schedules of all projects. In other words, the workspace repository only enables use cases that were previously not possible, without impairing any existing workflows. There might be a misunderstanding somewhere - it's the lock file of each individual project that determines how each project is built and tested. There is no version pinning information in the workspace itself, so users of libp2p don't need to use the workspace mechanism at all and there is no "central committee of library upgrades".

zah: If I assume that you are just advocating for your old and well-known opinion that packages should be installed locally to the project, just take a look at the passion of people creating variations of the "what is the heaviest object in the universe?" meme :)

https://www.google.com/search?q=the+heaviest+object+in+the+universe+node+modules&client=firefox-b-d&sxsrf=APq-WBvOJkjH3-dKuJxNK2N4FoGV9YoGKg:1648771793778&source=lnms&tbm=isch&sa=X&ved=2ahUKEwiYt9XIyfH2AhWPyIsKHfMWCVAQ_AUoAXoECAEQAw&biw=1533&bih=894&dpr=2#imgrc=hl0q5RRyKKGRaM

You often cite that yarn was created with lessons learned from from using package management at scale and guess what - yarn has a similar mechanism for sharing packages between multiple projects called "yarn workspaces":

https://classic.yarnpkg.com/lang/en/docs/workspaces/

Furthermore, the vendor folder has one simple fundamental issue that is eliminated by the workspace layout - right now the config file of nimbus affects the compilations of the test suites of all libraries (because of the way config files are discovered and used by Nim). This alone is a good reason to move away from the layout based on nested directories.

arnetheduck: > , just take a look at the passion of people creating variations

doesn't seem to be a blocker for adoption really - I'd be glad if there existed enough Nim packages to make such a meme

> workspaces

I'm not arguing against workspaces - I'm arguing against a single master workspace for all status-nim projects - this is not sustainable. 

I'm also arguing that workspaces should be fully self-contained without any magic caches in hidden locations - in the best case, this is the default - in a slightly worse case, there exists a magic set of commands I can run to get to this point. 

I imagine that a "default" workspace for a project is that project and all its dependencies. How these get organised within the workspace, I don't care as long as `rm -rf nimbus-eth2-workspace` removes all traces of me ever working on `nimbus-eth2`. As far as I'm concerned, `nimble clone git:///.../nimbus-eth2` would create such a "workspace" with some pre-determined structure without a special git repo to manage it. 

This is my original question: using the new nimble, what do I need to do to get to the above workspace configuration, where `nimbus-eth2` and _all_ projects / libraries it depends on exist in a single folder (of any internal structure)?

Whether there exists a cache or not is completely uninteresting and orthogonal as far as core operating principles go: the point of a cache is that you don't notice it's there, except for some speedups here and there - getting the operating principles right is far more important than a few seconds or bytes saved here and there.

Workspaces are great btw - ideally coupled with a monorepo so that we don't have to commit to 20 repos - this remains a pita and will get worse with lockfiles due to the sub-par tooling (no more integration with github / editors / commit tools for example).

zah: The `nim-workspace` repo starts empty. It's up to you to decide what you want to include there (and how many workspaces you'll have).

You add your project by running the `add-project` command:
https://github.com/status-im/nim-workspace#add-project-project-name

arnetheduck: so .. why  is a git repo needed at all?

PR #3561, opened 2022-03-31
JWT support

tersec:
Merging the merge branch.

Some design points: while so far, the engine authentication is only meaningful purely spec-wise for Bellatrix, and at that, really, post-merge or later, a few reasons mitigate not to gate it on that basis:
- Eth1 monitor as an abstraction shouldn't, ideally, track CL hardforks per se;
- it would create an authentication failure mode that might only show up later, rather than in a more testable way on initial launch;
- WebSocket connections to an EL might be created before The Merge, or even Bellatrix, and have to exist through the transitions and after as well; it's not well-defined in the specs how that'd work, but a reasonable reading is that the authentication would have to happen even for, say, an Altair-era connection;
- it creates potential for odd transition errors in the time skew between CL and EL;
- it allows for testing JWT in isolation, rather than tying it when there's no intrinsic connection, to The Merge. The Engine API is related to, but distinct from, The Merge.

However, this means that by default, it'd be a sort of invasive change on all the Web3 connections, adding this header. In theory, it's harmless, but e.g., https://lighthouse-book.sigmaprime.io/api-vc-auth-header.html describes a non-JWT header effectively sharing this namespace, and it's possible it'd have negative side effects permanently enabled, so it's feature-gated.

It'd be nice not to have too many feature-gated settings like this, and proposer boosting is already there, but I anticipate just having that on, with no way to disable it, within a month or two.

There's some repetition in the construction of `optJwtSecret`, but creating a reusable abstraction proved more cost than benefit -- it needs to know a config option and the `jwtSecret`, and the `if` expression is probably more straightforward and barely longer than, e.g., a `template` or `func` covering both.


mratsim:
> However, this means that by default, it'd be a sort of invasive change on all the Web3 connections, adding this header. In theory, it's harmless, but e.g., https://lighthouse-book.sigmaprime.io/api-vc-auth-header.html describes a non-JWT header effectively sharing this namespace, and it's possible it'd have negative side effects permanently enabled, so it's feature-gated.

I assume that Lighthouse will deprecate their authentication method to align to the standard.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -905,10 +924,12 @@ proc init*(T: type Eth1Chain, cfg: RuntimeConfig, db: BeaconChainDB): T =
...
     depositContractAddress: Eth1Address,
     depositContractDeployedAt: BlockHashOrNumber,
-    web3Url: string): Future[Result[DepositContractSnapshot, string]] {.async.} =
+    web3Url: string,
+    jwtSecret: Option[seq[byte]]): Future[Result[DepositContractSnapshot, string]]
+    {.async.} =
 
   let dataProviderRes =
-    await Web3DataProvider.new(depositContractAddress, web3Url)
+    await Web3DataProvider.new(depositContractAddress, web3Url, jwtSecret)
```
mratsim, line 932: would it be useful to have a default param to `None[seq[byte]]`?

mratsim (approved):
LGTM

tersec:
> > However, this means that by default, it'd be a sort of invasive change on all the Web3 connections, adding this header. In theory, it's harmless, but e.g., https://lighthouse-book.sigmaprime.io/api-vc-auth-header.html describes a non-JWT header effectively sharing this namespace, and it's possible it'd have negative side effects permanently enabled, so it's feature-gated.
> 
> I assume that Lighthouse will deprecate their authentication method to align to the standard.

Presumably; until then, it's a bit of a risk, though. I do look forward to simply being able to have this on all the time, dropping another setting, and have it add one more random-maybe-not-that-useful Web3 request header, to gain a reduced testing matrix, less to document, and a simpler codebase, but that probably can't happen yet.

beacon_chain/eth1/eth1_monitor.nim
tersec: I'm not sure:
```nim
$ rg -C4 Web3DataProvider.new beacon_chain/
beacon_chain/eth1/eth1_monitor.nim
932-    web3Url: string,
933-    jwtSecret: seq[byte]): Future[Result[DepositContractSnapshot, string]] {.async.} =
934-
935-  let dataProviderRes =
936:    await Web3DataProvider.new(depositContractAddress, web3Url, jwtSecret)
937-  if dataProviderRes.isErr:
938-    return err(dataProviderRes.error)
939-  var dataProvider = dataProviderRes.get
940-
--
1003-
1004-  # This is a way to detect that the monitor was restarted. When this
1005-  # happens, this function will just return terminating the "async thread"
1006-  while m.runFut == initialRunFut:
1007:    let tempProviderRes = await Web3DataProvider.new(
1008-      m.depositContractAddress,
1009-      web3Url,
1010-      m.jwtSecret)
1011-
--
1040-  let web3Url = m.web3Urls[m.startIdx mod m.web3Urls.len]
1041-  inc m.startIdx
1042-
1043-  m.dataProvider = block:
1044:    let v = await Web3DataProvider.new(
1045-      m.depositContractAddress, web3Url, m.jwtSecret)
1046-    if v.isErr():
1047-      raise (ref CatchableError)(msg: v.error())
1048-    v.get()
```
Two of those (`ensureDataProvider` and `detectPrimaryProviderComingOnline`) wouldn't be able to use such a default option, leaving just the one callsite where, indeed, it's probably not crucial or all that helpful. If there were 3 or more such callsites that didn't need it, I'd be more inclined to have `Web3Provider.new` provide that default, but the tradeoff is that it allows for silently dropping/neglecting the JWT secret in new, changed, or refactored code even when it is necessary. Given these tradeoffs, I prefer it as a non-optional/defaulted parameter.

PR #3563, opened 2022-04-01
use Eth1Monitor as abstraction; increase timeouts; handle newPayload 'accepted'

tersec:
More merging of the merge branch.

Most of these are cleanups/bugfixes of already-existing functionality, and preparing for block proposal.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -32,7 +32,7 @@ export sszdump, signatures_batch
 declareHistogram beacon_store_block_duration_seconds,
   "storeBlock() duration", buckets = [0.25, 0.5, 1, 2, 4, 8, Inf]
 
-const web3Timeout = 650.milliseconds
+const web3Timeout = 4.seconds
 
 type
   BlockEntry* = object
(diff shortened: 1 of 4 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -340,14 +350,14 @@ proc runForkchoiceUpdated(
       forkchoiceUpdated(
         self.consensusManager.eth1Monitor, headBlockRoot, finalizedBlockRoot),
       web3Timeout):
-        info "runForkChoiceUpdated: forkchoiceUpdated timed out"
+        debug "runForkChoiceUpdated: forkchoiceUpdated timed out"
         default(ForkchoiceUpdatedResponse)
   except CatchableError as err:
-    info "runForkChoiceUpdated: forkchoiceUpdated failed",
+    debug "runForkChoiceUpdated: forkchoiceUpdated failed",
```
arnetheduck, line 356: wouldn't this indicate that eth1 is down and the user needs to take some action?

arnetheduck, line 356: wouldn't this indicate that eth1 is down and the user needs to take some action?

arnetheduck (approved):

beacon_chain/gossip_processing/block_processor.nim
tersec: Usually not -- the ones to ensure are at `INFO` (or maybe higher, but `INFO` anyway) would be the timeouts, which might well mean that. The kind of failure referred to here is one indicated by the EL, which means the EL was contactable.

The other failure mode that could use better diagnostics, but needs something more targeted, is JWT authentication misconfiguration. Right now, it's a bit tricky for end-users to figure out, because it manifests as basically some HTTP error, and typically doesn't show up in EL client logs.

tersec: See the other explanation, the timeouts and JWT failures are the ones which should be surfaced more -- fcU failed is interesting, and important to know for debugging purposes, but doesn't have a lot of obvious end-user action to take.

PR #3568, opened 2022-04-05
remove bundled pyrmont support

arnetheduck:
The `pyrmont` testnet has been discontinued.

For experiments, it's still possible to run pyrmont nodes by passing a
genesis/config, but this PR removes the bundled `--network:pyrmont`
option.

* update docs
* remove empty docs
* remove obsolete `eth2-stats` page


docs/the_nimbus_book/src/resources.md
```diff
@@ -4,7 +4,7 @@
 
  - [Validator launchpad](https://prater.launchpad.ethereum.org): to send Prater deposits
 
- - [Beacon chain explorer](https://pyrmont.beaconcha.in/) : to monitor testnet health
```
tersec, line 7: Maybe worth redirecting to https://prater.beaconcha.in/ unless that's already in the book somewhere? Generically, it's a useful resource -- Pyrmont's just not the relevant chain anymore.

tersec (approved):

docs/the_nimbus_book/src/resources.md
arnetheduck: right, I'm not sure why this points to testnets at all given it's a "general" resources page and not a testnet-specific one - our "default" is mainnet across the board (even when launching the app without arguments), so it's good if the manual consistently either is explicit or refers to mainnet-by-default

arnetheduck: fixed the other link as well

PR #3570, opened 2022-04-06
post-merge Bellatrix block proposals

tersec:
Maybe a little too risky to merge before the upcoming release, though it is factored out into what should be a no-op change for non-Bellatrix.

Following up on the previous discussion of logging levels, it increases timeout logging (which does suggest there might be some user intervention necessary)

`newPayload` aligns at least the user-visible logging with the engine API RPC name.

`newPayload failed` typically is going to be a result of invalid data from the network, which is interesting at a debug level but not anything an end-user can usefully do much about, and generally recorded already by clients dumping invalid blocks.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -417,21 +417,6 @@ proc init*(T: type BeaconNode,
     else:
       none(seq[byte])
 
-  # Better to find out at startup
-  if config.suggestedFeeRecipient.isSome:
-    try:
-      discard Eth1Address.fromHex(config.suggestedFeeRecipient.get)
-    except ValueError as exc:
-      fatal "Specified an invalid suggested fee recipient",
-        msg = exc.msg
-      quit 1
-  else:
-    # TODO should warn on this not being configured but not useful yet to do so
-    # and there will be other ways of specifying these so it's premature to use
-    # any particular design yet for this part.
-    #warn "No suggested fee recipient provided; use --suggested-fee-recipient"
-    discard
-
   template getDepositContractSnapshot: auto =
     if depositContractSnapshot.isSome:
       depositContractSnapshot
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -417,6 +417,21 @@ proc init*(T: type BeaconNode,
     else:
       none(seq[byte])
 
+  # Better to find out at startup
+  if config.suggestedFeeRecipient.isSome:
+    try:
+      discard Eth1Address.fromHex(config.suggestedFeeRecipient.get)
+    except ValueError as exc:
+      fatal "Specified an invalid suggested fee recipient",
+        msg = exc.msg
+      quit 1
+  else:
+    # TODO should warn on this not being configured but not useful yet to do so
+    # and there will be other ways of specifying these so it's premature to use
+    # any particular design yet for this part.
+    #warn "No suggested fee recipient provided; use --suggested-fee-recipient"
+    discard
+
   template getDepositContractSnapshot: auto =
     if depositContractSnapshot.isSome:
       depositContractSnapshot
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -519,8 +519,8 @@ func build_empty_execution_payload*(state: bellatrix.BeaconState): ExecutionPayl
   var payload = ExecutionPayload(
     parent_hash: latest.block_hash,
     state_root: latest.state_root, # no changes to the state
-    receipts_root: Eth2Digest(data: cast[array[32, uint8]](
-      "nimbus was here\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0")),
+    receipts_root: static(Eth2Digest.fromHex(
+      "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")),
     block_number: latest.block_number + 1,
     prev_randao: randao_mix,
     gas_limit: latest.gas_limit, # retain same limit
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -506,3 +506,30 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
   # Note: This function is unsafe with respect to overflows and underflows.
   let slots_since_genesis = slot - GENESIS_SLOT
   state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT
+
+# https://github.com/ethereum/consensus-specs/blob/v1.1.10/tests/core/pyspec/eth2spec/test/helpers/execution_payload.py#L1-L31
+func build_empty_execution_payload*(state: bellatrix.BeaconState): ExecutionPayload =
+  ## Assuming a pre-state of the same slot, build a valid ExecutionPayload
+  ## without any transactions.
+  let
+    latest = state.latest_execution_payload_header
+    timestamp = compute_timestamp_at_slot(state, state.slot)
+    randao_mix = get_randao_mix(state, get_current_epoch(state))
+
+  var payload = ExecutionPayload(
+    parent_hash: latest.block_hash,
+    state_root: latest.state_root, # no changes to the state
+    receipts_root: Eth2Digest(data: cast[array[32, uint8]](
+      "nimbus was here\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0")),
+    block_number: latest.block_number + 1,
+    prev_randao: randao_mix,
+    gas_limit: latest.gas_limit, # retain same limit
+    gas_used: 0, # empty block, 0 gas
+    timestamp: timestamp,
+    base_fee_per_gas: latest.base_fee_per_gas) # retain same base_fee
+
+  payload.block_hash = withEth2Hash:
+    h.update payload.hash_tree_root().data
+    h.update cast[array[13, uint8]]("FAKE RLP HASH")
+
+  payload
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -474,7 +474,7 @@ proc getExecutionPayload(node: BeaconNode, proposalState: auto):
     let
       feeRecipient =
         if node.config.suggestedFeeRecipient.isSome:
-          Eth1Address.fromHex(node.config.suggestedFeeRecipient.get)
+          node.config.suggestedFeeRecipient.get
         else:
           default(Eth1Address)
       latestHead =
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -459,7 +459,7 @@ proc getExecutionPayload(node: BeaconNode, proposalState: auto):
   # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
 
   # Only current hardfork with execution payloads is Bellatrix
-  static: doAssert high(BeaconBlockFork) == BeaconBlockFork.Bellatrix
+  static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
   template empty_execution_payload(): auto =
     build_empty_execution_payload(proposalState.bellatrixData.data)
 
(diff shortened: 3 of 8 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -474,23 +516,31 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
+        else:
+          node.syncCommitteeMsgPool[].produceSyncAggregate(head.root),
+        if slot.epoch < node.dag.cfg.BELLATRIX_FORK_EPOCH:
+          default(bellatrix.ExecutionPayload)
+        else:
+          (await getExecutionPayload(node, proposalState)),
+        noRollback, # Temporary state - no need for rollback
+        cache)
+    except CatchableError as err:
+      # Prefer not to create block at all if it can't get ExecutionPayload
```
arnetheduck: why not create with an empty EP? won't this net some small reward at least?

tersec: It can't be literally empty, once the first non-empty `ExecutionPayload` kicks in, because https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#is_merge_transition_complete determines whether the merge has completed literally by checking against an empty EP.

It is possible to do something conceptually similar, and that might be worthwhile -- build what the ELs seem to call something like an empty header EP, with no transactions.

tersec: https://github.com/ethereum/consensus-specs/blob/dev/tests/core/pyspec/eth2spec/test/helpers/execution_payload.py demonstrates how to build these.

tersec: And https://github.com/status-im/nimbus-eth2/blob/30eef0a3697cd02554332f9dc221a84c6f6b52a6/tests/mocking/mock_blocks.nim#L46-L71 implements it in Nim. So could just move that out of `tests/` and into `beacon_chain/spec/`.

beacon_chain/validators/validator_duties.nim
```diff
@@ -443,6 +443,48 @@ proc forkchoice_updated(state: bellatrix.BeaconState,
...
+    default(bellatrix.ExecutionPayload)
+  else:
+    asConsensusExecutionPayload(
+      await execution_engine.getPayload(payload_id.get))
+
+proc getExecutionPayload(node: BeaconNode, proposalState: auto):
+    Future[ExecutionPayload] {.async.} =
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
+  # TODO a more reasonable fallback
+  doAssert not node.eth1Monitor.isNil
```
tersec: The "more reasonable fallback" should be to build an (essentially) empty `ExecutionPayload` if required; it can't really be merged until it either special-cases the never-merged case, which is essentially a local testnet artifact, or actually ensures that this condition is handled reasonably, which is probably take the reduced profit, produce some block, and log an error.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3570/commits/3062fb5fbc969a58310485912da6bad7e31fc139

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3570/commits/3062fb5fbc969a58310485912da6bad7e31fc139

beacon_chain/spec/helpers.nim
```diff
@@ -506,3 +506,30 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
...
+  let
+    latest = state.latest_execution_payload_header
+    timestamp = compute_timestamp_at_slot(state, state.slot)
+    randao_mix = get_randao_mix(state, get_current_epoch(state))
+
+  var payload = ExecutionPayload(
+    parent_hash: latest.block_hash,
+    state_root: latest.state_root, # no changes to the state
+    receipts_root: Eth2Digest(data: cast[array[32, uint8]](
+      "no receipts here\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0")),
```
arnetheduck:
```suggestion
      "nimbus was here\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0")),
```

beacon_chain/validators/validator_duties.nim
```diff
@@ -443,6 +444,66 @@ proc forkchoice_updated(state: bellatrix.BeaconState,
...
+          payload)
+
+    if executionPayloadStatus != PayloadExecutionStatus.valid:
+      info "getExecutionPayload: newExecutionPayload not valid; using empty execution payload",
+        executionPayloadStatus
+      return empty_execution_payload
+
+    return payload
+  except CatchableError as err:
+    # Prefer not to create block at all if it can't get ExecutionPayload
```
arnetheduck: comment needs updating

arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
```diff
@@ -443,6 +444,66 @@ proc forkchoice_updated(state: bellatrix.BeaconState,
...
+    payload_id: Option[bellatrix.PayloadId], execution_engine: Eth1Monitor):
+    Future[bellatrix.ExecutionPayload] {.async.} =
+  return if payload_id.isNone():
+    # Pre-merge, empty payload
+    default(bellatrix.ExecutionPayload)
+  else:
+    asConsensusExecutionPayload(
+      await execution_engine.getPayload(payload_id.get))
+
+proc getExecutionPayload(node: BeaconNode, proposalState: auto):
```
arnetheduck, line 457: would it make sense to get execution payloads a few times in the slot leading up to the proposal slot? that would allow a design where we poll for payloads continuously, but then create a block at a specific cutoff

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -414,6 +414,21 @@ proc init*(T: type BeaconNode,
   # The JWT secret created always exists, it just might not always be used
   let optJwtSecret = if config.useJwt: some jwtSecret.get else: none(seq[byte])
 
+  # Better to find out at startup
+  if config.suggestedFeeRecipient.isSome:
```
zah: We are usually handing the invalid CLI parameters in the `parseCmdArg` overloads. This allows us to have a consistent way of failing on an invalid argument in general (even though the current handling is not particularly nice).

beacon_chain/spec/helpers.nim
```diff
@@ -506,3 +506,30 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
...
+  ## without any transactions.
+  let
+    latest = state.latest_execution_payload_header
+    timestamp = compute_timestamp_at_slot(state, state.slot)
+    randao_mix = get_randao_mix(state, get_current_epoch(state))
+
+  var payload = ExecutionPayload(
+    parent_hash: latest.block_hash,
+    state_root: latest.state_root, # no changes to the state
+    receipts_root: Eth2Digest(data: cast[array[32, uint8]](
```
zah: What is the purpose of this value? There is a specific correct `receipts_root` value for a block without transactions.

beacon_chain/validators/validator_duties.nim
```diff
@@ -443,6 +444,66 @@ proc forkchoice_updated(state: bellatrix.BeaconState,
...
+      latestHead =
+        if not node.dag.head.executionBlockRoot.isZero:
+          node.dag.head.executionBlockRoot
+        else:
+          default(Eth2Digest)
+      latestFinalized = node.dag.finalizedHead.blck.executionBlockRoot
+      payload_id = (await forkchoice_updated(
+        proposalState.bellatrixData.data, latestHead, latestFinalized,
+        feeRecipient, node.consensusManager.eth1Monitor))
+      payload = await get_execution_payload(
```
zah, line 489: Is there an intention in the spec to provide a little bit of time for the execution client to assemble the best block possible after calling `forkchoice_updated`? Here, we are requesting the block payload right away.

beacon_chain/validators/validator_duties.nim
```diff
@@ -443,6 +444,66 @@ proc forkchoice_updated(state: bellatrix.BeaconState,
...
+    asConsensusExecutionPayload(
+      await execution_engine.getPayload(payload_id.get))
+
+proc getExecutionPayload(node: BeaconNode, proposalState: auto):
+    Future[ExecutionPayload] {.async.} =
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
+
+  # Only current hardfork with execution payloads is Bellatrix
+  static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
+  template empty_execution_payload(): auto =
```
zah, line 463: As per my other comment in `build_empty_execution_payload`, I'm concerned that this is not a valid payload.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3570/commits/aa165b4d0b1b456b3c74c845018e07a2b1d007c4

beacon_chain/validators/validator_duties.nim
tersec: What this PR does is a simple flow which doesn't create long-lived or potentially inconsistent state anywhere else, but probably not good for profit maximization.

In theory, one wants the execution payload created as closely as possible to the transaction proposal itself -- this is a concern that comes up in MEV design as well, for example. There are costs here -- for example, this risks putting the `getPayload` in a timing-critical patch right next to the proposal.

I'm not sure I see the value in polling for execution payloads though -- just ask once, shortly before block proposal. What might make sense is to have a trigger late in the immediately pre-block-proposal-slot to call `getPayloadV1`.

There's still a problem here, though. The flow should ultimately be something like, whether via whatever mechanism the VC uses or in the block processing loop in `block_processor`, each time the fcU is called, if and only if the next slot will involve proposing a block, don't call the non-validating fcU, but call the validating (gets a payload ID; don't know if there's a standard terminology here) fcU, giving the EL client several seconds to prepare a payload and stuff it with the best transactions it can find.

The problem currently isn't really the lack of `getPayload` polling, it's the complete lack of waiting at all for the EL client after the `fcU` before calling `getPayload`.

But I'm not sure I want this PR to get into that sort of thing -- they're all, essentially, optimizations, and I'd like this PR to focus on correctness and establishing a feature set people can use on Kiln and similar testnets.

beacon_chain/validators/validator_duties.nim
tersec: Yes, see https://github.com/status-im/nimbus-eth2/pull/3570#discussion_r848463980

It's a simple, spec/protocol-correct flow, which is localized and avoids dealing with failure modes of global state. But it's definitely not optimal.

As far as how useful having that code at all is, it seems likely that due to the inherent issues with trusting that whatever payload ID some other fcU-calling code has stashed somewhere is real/available/correct, probably one would want a fallback in the block proposal to get some correct `ExecutionPayload` out of it anyway.

beacon_chain/spec/helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3570/commits/5600756fc2f74c40ab02c7c65984bbb7465e5a1b

https://github.com/ethereum/go-ethereum/blob/d4d288e3f1cebb183fce9137829a76ddf7c6d12a/core/types/block.go#L153-L156 states it's `EmptyRootHash`, which https://github.com/ethereum/go-ethereum/blob/d4d288e3f1cebb183fce9137829a76ddf7c6d12a/core/types/block.go#L35 defines as 0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421

The previous, less whimsical value of this also didn't match this hash, which is rather an Ethereum SHA3-based hash of:
```
>>> web3.sha3('0x80', encoding='hex')
'56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
```
https://pypi.org/project/web3/2.6.0/

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/5600756fc2f74c40ab02c7c65984bbb7465e5a1b

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3570/commits/3763efce7447b212fde99f4447cf80b7aa867c7b

beacon_chain/validators/validator_duties.nim
arnetheduck: > value

it would guard against the case where the EL gets blocked by something (loading caches / whatever) and is unable to produce a payload on time

tersec: Well, so does the empty execution payload fallback, but sure, certain execution layer clients have proven slower to respond to these queries than others.

tersec:
Will continue with other items in additional PRs, or obviously address other issues that come up. All of the specific issues mentioned within scope of this PR have been addressed.

Alternate timings of `fcU`/`getPayload` aren't for this PR, but certainly `fcU` and maybe `getPayload` should be tweaked.

PR #3577, opened 2022-04-08
`gnosis-chain` -> `gnosis`

### LINKED ISSUE #3504: Feature: Common naming convention for Gnosis network parameter

In order to make life easier for tooling like eth-docker, I'd love to see a common naming convention for specifying the Gnosis chain via parameter. Lighthouse uses `gnosis` and Nimbus uses `gnosis-chain`, and Prysm may be the third in the mix.

Please consult with the Lighthouse team and agree on which name should be used.

arnetheduck:
Use same name as LH/Teku throughout

* fixes #3504
* fixes large stack temporary that can cause crashes during genesis
detection


zah:
This is a breaking change though. We already have Gnosis users out there.

arnetheduck:
> This is a breaking change though. We already have Gnosis users out there.

added some deprecation warnings and compatibility names, should hopefully cut it

beacon_chain/spec/forks.nim
```diff
@@ -172,12 +172,13 @@ template toFork*[T: bellatrix.BeaconState | bellatrix.HashedBeaconState](
     t: type T): BeaconStateFork =
   BeaconStateFork.Bellatrix
 
-template init*(T: type ForkedHashedBeaconState, data: phase0.HashedBeaconState): T =
-  T(kind: BeaconStateFork.Phase0, phase0Data: data)
-template init*(T: type ForkedHashedBeaconState, data: altair.HashedBeaconState): T =
-  T(kind: BeaconStateFork.Altair, altairData: data)
-template init*(T: type ForkedHashedBeaconState, data: bellatrix.HashedBeaconState): T =
-  T(kind: BeaconStateFork.Bellatrix, bellatrixData: data)
+# TODO these cause stack overflows due to large temporaries getting allocated
```
zah, line 175: In the light of all discovered issues, do you still think that making `BeaconState` a ref type is a bad idea?

arnetheduck: yes - it allows the caller to make memory allocation decisions which is how it should be - the problem here is not `ref` or not but rather the lack of CI for both genesis detection and the gnosis build  - either would make this show in CI (I've configured my gcc to show these warnings always, which makes them trivial to find)

Issue #3495, opened 2022-03-14
/eth/v2/beacon/blocks returning invalid extra_data


mcdee:
**Describe the bug**

A request to the endpoint `/eth/v2/beacon/blocks` returns the `extra_data` field in an invalid format.

**To Reproduce**

Request a bellatrix block from `/eth/v2/beacon/blocks`, for example:

```
curl -s http://localhost:5052/eth/v2/beacon/blocks/21920 | jq '.data.message.body.execution_payload.extra_data'
[]
```

**Additional context**

This value should be returned as a hex string rather than a byte array.

tersec:
https://github.com/status-im/nimbus-eth2/pull/3581

PR #3579, opened 2022-04-08
hex-encode bellatrix extradata in JSON

tersec:
Fix https://github.com/status-im/nimbus-eth2/issues/3495

It returns `"0x"` for an empty `extra_data`, which on `kiln` is nearly all of them.

### DIFF
--- a/beacon_chain/spec/datatypes/bellatrix.nim
+++ b/beacon_chain/spec/datatypes/bellatrix.nim
@@ -19,6 +19,8 @@ import
   ../digest,
   "."/[base, phase0, altair]
 
+from std/typetraits import distinctBase
+
 export json_serialization, base
 
 const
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -347,6 +352,18 @@ proc readValue*(reader: var JsonReader, value: var ExecutionAddress) {.
     raiseUnexpectedValue(reader,
                          "ExecutionAddress value should be a valid hex string")
 
+proc writeValue*(writer: var JsonWriter, value: ExtraData) {.
```
zah, line 355: `ExtraData` is not a distinct type, so you are not supposed to create such overloads - they will affect other `List[byte, N]` instances as well. Admittedly, perhaps this treatment should be applied to all `List[byte]` instances, but then the overload should be fixed to reflect that.

tersec: https://github.com/status-im/nimbus-eth2/pull/3581

PR #3583, opened 2022-04-11
Implement all sync committee duties in the validator client

zah:
Other changes:

* logtrace can now verify sync committee messages and contributions
* Many unnecessary use of pairs() have been removed for consistency

### DIFF
--- a/beacon_chain/validator_client/api.nim
+++ b/beacon_chain/validator_client/api.nim
@@ -209,7 +209,7 @@ template onceToAll*(vc: ValidatorClientRef, responseType: typedesc,
       except CancelledError:
         ApiOperation.Interrupt
 
-    for idx, node {.inject.} in onlineNodes.pairs():
+    for idx, node {.inject.} in onlineNodes:
       it = node.client
       let apiResponse {.inject.} =
         block:
@@ -342,16 +342,17 @@ template firstSuccessTimeout*(vc: ValidatorClientRef, respType: typedesc,
     if exitNow:
       break
 
-    let offlineMask = {RestBeaconNodeStatus.Offline,
-                       RestBeaconNodeStatus.NotSynced,
-                       RestBeaconNodeStatus.Uninitalized}
-    let offlineNodes = vc.beaconNodes.filterIt(it.status in offlineMask)
-    let onlineNodesCount = len(vc.beaconNodes) - len(offlineNodes)
+    let unusableModeMask = {RestBeaconNodeStatus.Offline,
+                            RestBeaconNodeStatus.NotSynced,
+                            RestBeaconNodeStatus.Uninitalized,
+                            RestBeaconNodeStatus.Incompatible}
+    let unusableNodes = vc.beaconNodes.filterIt(it.status in unusableModeMask)
+    let onlineNodesCount = len(vc.beaconNodes) - len(unusableNodes)
 
     warn "No working beacon nodes available, refreshing nodes status",
-         online_nodes = onlineNodesCount, offline_nodes = len(offlineNodes)
+         online_nodes = onlineNodesCount, unusable_nodes = len(unusableNodes)
 
-    var checkFut = vc.checkNodes(offlineMask)
+    var checkFut = vc.checkNodes(unusableModeMask)
 
     let checkOp =
       block:
@@ -410,8 +411,7 @@ template firstSuccessTimeout*(vc: ValidatorClientRef, respType: typedesc,
       break
 
 proc getProposerDuties*(vc: ValidatorClientRef,
-                        epoch: Epoch): Future[GetProposerDutiesResponse] {.
-     async.} =
+                        epoch: Epoch): Future[GetProposerDutiesResponse] {.async.} =
   logScope: request = "getProposerDuties"
   vc.firstSuccessTimeout(RestResponse[GetProposerDutiesResponse], SlotDuration,
                          getProposerDuties(it, epoch)):
@@ -428,7 +428,7 @@ proc getProposerDuties*(vc: ValidatorClientRef,
       of 400:
         debug "Received invalid request response",
               response_code = response.status, endpoint = node
-        RestBeaconNodeStatus.Offline
+        RestBeaconNodeStatus.Incompatible
       of 500:
         debug "Received internal error response",
               response_code = response.status, endpoint = node
@@ -444,9 +444,10 @@ proc getProposerDuties*(vc: ValidatorClientRef,
 
   raise newException(ValidatorApiError, "Unable to retrieve proposer duties")
 
-proc getAttesterDuties*(vc: ValidatorClientRef, epoch: Epoch,
-                        validators: seq[ValidatorIndex]
-                       ): Future[GetAttesterDutiesResponse] {.async.} =
+proc getAttesterDuties*(
+    vc: ValidatorClientRef,
+    epoch: Epoch,
+    validators: seq[ValidatorIndex]): Future[GetAttesterDutiesResponse] {.async.} =
   logScope: request = "getAttesterDuties"
   vc.firstSuccessTimeout(RestResponse[GetAttesterDutiesResponse], SlotDuration,
                          getAttesterDuties(it, epoch, validators)):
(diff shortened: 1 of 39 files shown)


scripts/launch_local_testnet.sh
```diff
@@ -482,7 +482,7 @@ dump_logs() {
 
 dump_logtrace() {
   if [[ "$ENABLE_LOGTRACE" == "1" ]]; then
-    find "${DATA_DIR}" -maxdepth 1 -type f -regex '.*/log[0-9]+.txt' | sed -e"s/${DATA_DIR}\//--nodes=/" | sort | xargs ./build/logtrace aggasr --log-dir="${DATA_DIR}" || true
+    find "${DATA_DIR}" -maxdepth 1 -type f -regex '.*/log[0-9]+.txt' | sed -e"s/${DATA_DIR}\//--nodes=/" | sort | xargs ./build/logtrace traceAll --log-dir="${DATA_DIR}" || true
```
tersec: One reason specifically to look at aggregate attestations is that they're the ones that (maybe not anymore, even, but for a while) were almost always rebroadcast, because there was no question of subnets. The non-aggregated attestations, would routinely show up as being properly not-delivered, but, that was by design.

beacon_chain/validator_client/duties_service.nim
```diff
@@ -6,20 +6,24 @@ logScope: service = "duties_service"
...
+  of SyncCommitteeLoop: "sync_committee_loop"
 
 proc checkDuty(duty: RestAttesterDuty): bool =
   (duty.committee_length <= MAX_VALIDATORS_PER_COMMITTEE) and
   (uint64(duty.committee_index) <= MAX_COMMITTEES_PER_SLOT) and
   (uint64(duty.validator_committee_index) <= duty.committee_length) and
   (uint64(duty.validator_index) <= VALIDATOR_REGISTRY_LIMIT)
 
+proc checkSyncDuty(duty: RestSyncCommitteeDuty): bool =
+  uint64(duty.validator_index) <= VALIDATOR_REGISTRY_LIMIT
```
tersec, line 25: It seems easier to statically verify that type changing on the constant `VALIDATOR_REGISTRY_LIMIT` is safe and correct (also true of the other `uint64(duty.foo) <= CONSTANT` comparisons in `checkDuty` just before this code). It should also be more efficient, since in theory it can be done at compile-time.

beacon_chain/validator_client/api.nim
```diff
@@ -692,9 +796,48 @@ proc getAggregatedAttestation*(vc: ValidatorClientRef, slot: Slot,
...
+    else:
+      let response = apiResponse.get()
+      case response.status:
+      of 200:
+        debug "Received successful response", endpoint = node
+        return response.data.data
+      of 400:
+        debug "Received invalid request response",
+              response_code = response.status, endpoint = node
+        RestBeaconNodeStatus.Offline
```
tersec: Is `RestBeaconNodeStatus.Offline` the correct error code here (or for the internal error response, or unexpected error response)? It can't be offline if it received a response, even if it's an error response of some sort.

beacon_chain/validator_client/api.nim
```diff
@@ -727,10 +870,46 @@ proc publishAggregateAndProofs*(vc: ValidatorClientRef,
...
+      let response = apiResponse.get()
+      case response.status:
+      of 200:
+        debug "Contribution and proofs were successfully published", endpoint = node
+        return true
+      of 400:
+        debug "Received invalid request response",
+              response_code = response.status, endpoint = node,
+              response_error = response.getGenericErrorMessage()
+        RestBeaconNodeStatus.Offline
```
tersec: Same question: is RestBeaconNodeStatus.Offline the correct error code here (or for the internal error response, or unexpected error response)? It can't be offline if it received a response, even if it's an error response of some sort.

beacon_chain/validator_client/api.nim
```diff
@@ -851,3 +1030,44 @@ proc prepareBeaconCommitteeSubnet*(vc: ValidatorClientRef,
...
+      of 400:
+        debug "Received invalid request response",
+              response_code = response.status, endpoint = node,
+              response_error = response.getGenericErrorMessage()
+        return false
+      of 500:
+        debug "Received internal error response",
+              response_code = response.status, endpoint = node,
+              response_error = response.getGenericErrorMessage()
+        RestBeaconNodeStatus.Offline
```
tersec, line 1061: More seemingly-overbroad `RestBeaconNodeStatus.Offline`

beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -0,0 +1,356 @@
...
+    return
+  except CatchableError as exc:
+    error "Unexpected error while producing sync committee messages",
+          slot = slot,
+          duties_count = len(duties),
+          err_name = exc.name, err_msg = exc.msg
+    return
+
+  let contributionTime =
+    # chronos.Duration substraction could not return negative value, in such
```
tersec:
```suggestion
    # chronos.Duration subtraction cannot return a negative value; in such
```

beacon_chain/validator_client/api.nim
```diff
@@ -644,24 +710,62 @@ proc submitPoolAttestations*(vc: ValidatorClientRef,
...
+      let response = apiResponse.get()
+      case response.status
+      of 200:
+        debug "Sync committee message was successfully published", endpoint = node
+        return true
+      of 400:
+        debug "Received invalid request response",
+              response_code = response.status, endpoint = node,
+              response_error = response.getDutyErrorMessage()
+        RestBeaconNodeStatus.Offline
```
tersec: More `RestBeaconNodeStatus.Offline` when it's not actually offline

scripts/launch_local_testnet.sh
zah: I was aware that this makes the output a bit more noisy, but it didn't look that bad in my test runs. It was still possible to get an idea regarding the simulation results in a glance. Getting a cleaner output would require creating a new category like `traceAllUnexpecedNonDeliveried` perhaps.

tersec: Seems reasonable.

beacon_chain/validator_client/duties_service.nim
zah: `duty.validator_index` is a distinct type (`ValidatorIndex`) which lacks comparison operators. The main goal of the code here is to strip away the distinct type. It probably makes sense to add comparison operators directly to the distinct type which would make these coercions unnecessary, but as evident from the code above, this was a widespread practice in the validator client codebase, so fixing this everywhere could be done in a separate PR.

beacon_chain/validator_client/api.nim
zah: Fair enough. I'll change these to `RestBeaconNodeStatus.Incompatible`. This PR was merely copying the existing practice in the module. The 40x status codes were mapped to `Offline` in all existing requests.

beacon_chain/validator_client/api.nim
```diff
@@ -533,9 +570,39 @@ proc getHeadStateFork*(vc: ValidatorClientRef): Future[Fork] {.async.} =
 
   raise newException(ValidatorApiError, "Unable to retrieve head state's fork")
 
-proc getValidators*(vc: ValidatorClientRef,
-                    id: seq[ValidatorIdent]): Future[seq[RestValidator]] {.
-     async.} =
+proc getHeadStateRoot*(vc: ValidatorClientRef): Future[RestRoot] {.async.} =
```
arnetheduck: `getHeadStateRoot`? state or block?

beacon_chain/validator_client/common.nim
```diff
@@ -51,12 +52,28 @@ type
...
   DutyAndProof* = object
     epoch*: Epoch
     dependentRoot*: Eth2Digest
     data*: RestAttesterDuty
     slotSig*: Option[ValidatorSig]
 
+  SyncCommitteeDuty* = object
+    pubkey*: ValidatorPubKey
+    validator_index*: ValidatorIndex
+    validator_sync_committee_index*: uint16
```
arnetheduck: why not `IndexInSyncCommittee` here?

arnetheduck (approved):

PR #3593, opened 2022-04-14
stylecheck fixes

tersec:
- `skipBLSValidation` to `skipBlsValidation`
- `bls_sign` to `blsSign`
- add copyright header to `tests/test_keystore_management.nim`
- `KeyStorePass` to `KeystorePass`


tersec:
Jenkins failure is due to incompatibility of new version of Jenkins with caching plugin. This is a non-semantic change so should be safe in that regard.

PR #3598, opened 2022-04-15
avoid unnecessary recompression in block protocol

arnetheduck:
Blocks can be sent straight from compressed data sources

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -533,6 +533,25 @@ proc isLightClientRequestProto(fn: NimNode): NimNode =
 
   return newLit(false)
 
+proc writeChunkSZ*(
+    conn: Connection, responseCode: Option[ResponseCode],
+    uncompressedLen: uint64, payloadSZ: openArray[byte],
+    contextBytes: openArray[byte] = []): Future[void] =
+  var output = memoryOutput(payloadSZ.len + contextBytes.len + 11)
+  try:
+    if responseCode.isSome:
+      output.write byte(responseCode.get)
+
+    if contextBytes.len > 0:
+      output.write contextBytes
+
+    output.write toBytes(uncompressedLen, Leb128).toOpenArray()
+    output.write payloadSZ
+  except IOError as exc:
+    raiseAssert exc.msg # memoryOutput shouldn't raise
+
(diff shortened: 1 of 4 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -533,6 +533,25 @@ proc isLightClientRequestProto(fn: NimNode): NimNode =
 
   return newLit(false)
 
+proc writeChunkSZ*(
+    conn: Connection, responseCode: Option[ResponseCode],
+    uncompressedLen: uint64, payloadSZ: openArray[byte],
+    contextBytes: openArray[byte] = []): Future[void] =
+  var output = memoryOutput(payloadSZ.len + contextBytes.len + 11)
```
etan-status: Where is the 11 coming from?

beacon_chain/sync/sync_protocol.nim
```diff
@@ -334,11 +336,16 @@ p2pProtocol BeaconSync(version = 1,
...
+        let uncompressedLen = uncompressedLenFramed(bytes).valueOr:
+          warn "Cannot read block size, database corrupt?",
+            bytes = bytes.len(), blck = shortLog(blockRef)
+          continue
+
         peer.updateRequestQuota(blockResponseCost)
         peer.awaitNonNegativeRequestQuota()
 
-        await response.writeRawBytes(bytes, []) # phase0 bytes
+        await response.writeBytesSZ(uncompressedLen, bytes, []) # phase0
```
etan-status, line 348:
```suggestion
        await response.writeBytesSZ(uncompressedLen, bytes, []) # phase0 bytes
```

(above at line 284, the comment was not deleted. I guess it got removed here unintentionally)

etan-status (approved):

beacon_chain/networking/eth2_network.nim
arnetheduck: up to 10 bytes of leb128 and one byte of response code, worst case

etan-status:
```suggestion
  const numOverheadBytes = sizeof(byte) + Leb128.maxLen(typeof(uncompressedLen))
  var output = memoryOutput(payloadSZ.len + contextBytes.len + numOverheadBytes)
```

PR #3605, opened 2022-04-20
era file verification

arnetheduck:
Implement and document era file verification

* era file states now come with block applied for easier verification
* clarify conflicting version handling
* document verification requirements
* remove count from name, use start-era, end-root to discover range

### DIFF
--- a/beacon_chain/era_db.nim
+++ b/beacon_chain/era_db.nim
@@ -352,28 +352,29 @@ proc getPartialState(
     false
 
 iterator getBlockIds*(
-    db: EraDB, historical_roots: openArray[Eth2Digest], era: Era): BlockId =
-  # The state from which we load block roots is stored in the file corresponding
-  # to the "next" era
-  let fileEra = era + 1
-
+    db: EraDB, historical_roots: openArray[Eth2Digest], startSlot: Slot): BlockId =
   var
     state = (ref PartialBeaconState)() # avoid stack overflow
+    slot = startSlot
 
-  # `case` ensures we're on a fork for which the `PartialBeaconState`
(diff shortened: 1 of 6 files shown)


beacon_chain/era_db.nim
```diff
@@ -81,47 +65,215 @@ proc getEraFile(
...
+      return ok(f)
+
+  let
+    eraRoot = eraRoot(
+        db.genesis_validators_root, historical_roots, era).valueOr:
+      return err("Era outside of known history")
+    name = eraFileName(db.cfg, era, eraRoot)
+    f = ? EraFile.open(db.path / name)
+
+  if db.files.len > 16: # TODO LRU
```
zah, line 246: This should be configurable perhaps.

beacon_chain/era_db.nim
```diff
@@ -215,26 +352,29 @@ proc getPartialState(
...
-    for root in state[].block_roots:
-      yield BlockId(root: root, slot: slot)
+  while true:
+    # `case` ensures we're on a fork for which the `PartialBeaconState`
+    # definition is consistent
+    case db.cfg.stateForkAtEpoch(slot.epoch)
+    of BeaconStateFork.Phase0, BeaconStateFork.Altair, BeaconStateFork.Bellatrix:
+      let stateSlot = (slot.era() + 1).start_slot()
+      if not getPartialState(db, historical_roots, stateSlot, state[]):
+        state = nil # No `return` in iterators
```
zah, line 367: I don't quite understand the comment here. You should be able to say `break` here.

zah (approved):

PR #3607, opened 2022-04-23
Fix invalid merkle proof generation when there is a large number of unprocessed deposits

zah:


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -817,16 +817,17 @@ proc getBlockProposalData*(chain: var Eth1Chain,
             depositsRoot = vote.deposit_root,
             localDeposits = getStateField(state, eth1_data).deposit_count
 
+  let stateDepositIdx = getStateField(state, eth1_deposit_index)
   var pendingDepositsCount =
-    getStateField(state, eth1_data).deposit_count -
-      getStateField(state, eth1_deposit_index)
+    getStateField(state, eth1_data).deposit_count - stateDepositIdx
```
etan-status: sanity check for `stateDepositIdx > deposit_count` underflow?

zah: I'm pretty sure a state carrying such contradicting values would be illegal, but I added some sanity checks regardless.

PR #3611, opened 2022-04-28
don't always broadcast sync contributions and aggregates on Altair topics

tersec:


etan-status (approved):
Looks consistent with most of the other ones (`broadcastAttestation` has different visuals but is functionally the same). And also fixes the obvious bugs :)

One question is whether the message should instead go out based on the slot at which it originated, in case the wall slot advanced while processing. Granted, it's most likely the same slot (or not relevant re fork digests), and the messages are also of ephemeral validity, but still would be more deterministic, and in line with `broadcastBeaconBlock` and `broadcastOptimisticLightClientUpdate`.

The way how we process it on the receiving end, the logic does not care about the exact topic forkDigest that the message was received with, the validation itself only looks at the payload and does not cross-check it (for topics that share same content across multiple forkDigests). Not sure about other implementations.

tersec:
I've tended to view it as being broadcast at wall-slot-time when possible and conceptualize this as there being a moment in time where hardfork 1 ends and hardfork 2 begins and, ideally, there's a cutover of sent messages.

This can create issues verifying messages along that boundary, but because https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/altair/beacon-chain.md#syncaggregate shows that at least some sync committee messages have no internal notion of a slot or similar to guide this, it's as good as anything to fallback on wall time for this purpose for constructing the signature.

PR #3613, opened 2022-05-04
use MAX_CHUNK_SIZE_BELLATRIX for signed Bellatrix blocks

tersec:
The faststreams backend has relevant code, but it's a `TODO` because
```
$ make NIMFLAGS="-d:snappy_implementation=native" nimbus_beacon_node 
Building: build/nimbus_beacon_node
nimbus-eth2/beacon_chain/networking/faststreams_backend.nim(26, 46) template/generic instantiation of `async` from here
nimbus-eth2/beacon_chain/networking/faststreams_backend.nim(27, 21) template/generic instantiation of `fsTranslateErrors` from here
nimbus-eth2/beacon_chain/networking/faststreams_backend.nim(28, 11) Error: undeclared identifier: 'safeClose'
make: *** [Makefile:256: nimbus_beacon_node] Error 1
```

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -1971,8 +1971,7 @@ template gossipMaxSize(T: untyped): uint32 =
          T is altair.SignedBeaconBlock:
       GOSSIP_MAX_SIZE
     else:
-      echo T
-      {.fatal: "unknown type".}
+      {.fatal: "unknown type " & name(T).}
   static: doAssert maxSize <= maxGossipMaxSize()
   maxSize.uint32
 
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -614,6 +614,12 @@ when useNativeSnappy:
 else:
   include libp2p_streams_backend
 
+func maxChunkSize(t: typedesc[bellatrix.SignedBeaconBlock]): uint32 =
+  MAX_CHUNK_SIZE_BELLATRIX
+
+func maxChunkSize(t: typedesc): uint32 =
+  MAX_CHUNK_SIZE
+
 proc makeEth2Request(peer: Peer, protocolId: string, requestBytes: Bytes,
                      ResponseMsg: type,
                      timeout: Duration): Future[NetRes[ResponseMsg]]
@@ -635,7 +641,7 @@ proc makeEth2Request(peer: Peer, protocolId: string, requestBytes: Bytes,
     # Read the response
     return
       await readResponse(when useNativeSnappy: libp2pInput(stream) else: stream,
-                         peer, ResponseMsg, timeout)
+                         peer, maxChunkSize(ResponseMsg), ResponseMsg, timeout)
   finally:
     await stream.closeWithEOF()
 
@@ -787,8 +793,10 @@ proc handleIncomingStream(network: Eth2Node,
       NetRes[MsgRec].ok default(MsgRec)
     else:
       try:
-        awaitWithTimeout(readChunkPayload(s, peer, MsgRec), deadline):
-          returnInvalidRequest(errorMsgLit "Request full data not sent in time")
+        awaitWithTimeout(
+          readChunkPayload(s, peer, maxChunkSize(MsgRec), MsgRec), deadline):
+            returnInvalidRequest(
+              errorMsgLit "Request full data not sent in time")
 
       except SerializationError as err:
         returnInvalidRequest err.formatMsg("msg")
@@ -1963,6 +1971,7 @@ template gossipMaxSize(T: untyped): uint32 =
          T is altair.SignedBeaconBlock:
       GOSSIP_MAX_SIZE
     else:
+      echo T
       {.fatal: "unknown type".}
   static: doAssert maxSize <= maxGossipMaxSize()
   maxSize.uint32
--- a/beacon_chain/networking/faststreams_backend.nim
+++ b/beacon_chain/networking/faststreams_backend.nim
@@ -86,6 +86,8 @@ proc readSszValue(s: AsyncInputStream,
 proc readChunkPayload(s: AsyncInputStream,
                       noSnappy: bool,
                       MsgType: type): Future[NetRes[MsgType]] {.async.} =
+  # TODO this needs to sometimes be MAX_CHUNK_SIZE_BELLATRIX, for at least
+  # bellatrix.SignedBeaconBlock
   let prefix = await readSizePrefix(s, MAX_CHUNK_SIZE)
   let size = if prefix.isOk: prefix.value.int
              else: return err(prefix.error)
--- a/beacon_chain/networking/libp2p_streams_backend.nim
+++ b/beacon_chain/networking/libp2p_streams_backend.nim
@@ -99,6 +99,7 @@ proc uncompressFramedStream*(conn: Connection,
   return ok output
 
 proc readChunkPayload*(conn: Connection, peer: Peer,
+                       maxChunkSize: uint32,
                        MsgType: type): Future[NetRes[MsgType]] {.async.} =
   let sm = now(chronos.Moment)
   let size =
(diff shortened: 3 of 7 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -1963,6 +1971,7 @@ template gossipMaxSize(T: untyped): uint32 =
          T is altair.SignedBeaconBlock:
       GOSSIP_MAX_SIZE
     else:
+      echo T
       {.fatal: "unknown type".}
```
etan-status:
```suggestion
      {.fatal: "unknown type " & name(T).}
```

beacon_chain/networking/eth2_network.nim
```diff
@@ -614,6 +614,12 @@ when useNativeSnappy:
 else:
   include libp2p_streams_backend
 
+func maxChunkSize(t: typedesc[bellatrix.SignedBeaconBlock]): uint32 =
+  MAX_CHUNK_SIZE_BELLATRIX
```
etan-status, line 618: Does seem like a hidden bug source if not updated with each fork. Maybe a bit more future-proof:

```suggestion
func maxChunkSize(t: typedesc[ForkySignedBeaconBlock]): uint32 =
  when t.toFork >= BeaconBlockFork.Bellatrix:
    MAX_CHUNK_SIZE_BELLATRIX
  else:
    MAX_CHUNK_SIZE
```

beacon_chain/spec/keystore.nim
```diff
@@ -29,7 +29,7 @@ import nimcrypto/utils as ncrutils
 export
   results, burnMem, writeValue, readValue
 
-{.localPassC: "-fno-lto".} # no LTO for crypto
+{.localPassc: "-fno-lto".} # no LTO for crypto
```
etan-status, line 32: There is some more `passC` and `localPassC` in `config.nims`

beacon_chain/networking/libp2p_streams_backend.nim
```diff
@@ -99,6 +99,7 @@ proc uncompressFramedStream*(conn: Connection,
   return ok output
 
 proc readChunkPayload*(conn: Connection, peer: Peer,
+                       maxChunkSize: uint32,
```
etan-status, line 102: 1. There is a check in this function that errors on any `size > MAX_CHUNK_SIZE`.
2. Not sure if more ideomatic to do value before size, or size before value.

beacon_chain/networking/faststreams_backend.nim
```diff
@@ -86,6 +86,8 @@ proc readSszValue(s: AsyncInputStream,
 proc readChunkPayload(s: AsyncInputStream,
                       noSnappy: bool,
                       MsgType: type): Future[NetRes[MsgType]] {.async.} =
+  # TODO this needs to sometimes be MAX_CHUNK_SIZE_BELLATRIX, for at least
+  # bellatrix.SignedBeaconBlock
```
etan-status, line 90: Same for `readResponseChunk` / `readResponse`.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -70,28 +70,34 @@ type
...
+    MsgType: type (ref ForkedSignedBeaconBlock)):
     Future[NetRes[MsgType]] {.async.} =
   var contextBytes: ForkDigest
   try:
     await conn.readExactly(addr contextBytes, sizeof contextBytes)
   except CatchableError:
     return neterr UnexpectedEOF
 
+  # Ignores maxChunkSize; needs to be consistent formal parameters,
+  # but this function is where that's determined.
```
etan-status, line 83: Could maybe remove the need for `sync_protocol` to be aware of chunk sizes, with an exported wrapper in `eth2_network` that maps `readChunkPayload(conn, peer, MsgType)` to `readChunkPayload(conn, peer, maxChunkSize(MsgType), MsgType)`. The wrapper could then be used in `sync_protocol` to ensure `MsgType` and chunk size are consistent, in the `readChunkPayload(conn, peer, MAX_CHUNK_SIZE, altair.SignedBeaconBlock)`

etan-status (approved):

beacon_chain/networking/eth2_network.nim
tersec: Not sure this would work directly -- it's compile-time and the libp2p req/resp code at least in Nim terms doesn't guarantee these types at compile-time. That's why it's written differently than
https://github.com/status-im/nimbus-eth2/blob/039bece9175104b5c87a8c2ff6b1eafae731b05e/beacon_chain/networking/eth2_network.nim#L1948-L1968

I tried to reuse it directly, since it has some nice properties with regards to smaller messages too that shouldn't reach either `MAX_CHUNK_SIZE` or `GOSSIP_MAX_SIZE` (which happen to be identical, and which I checked in a static-assert way to be the case) or `MAX_CHUNK_SIZE_BELLATRIX` / `GOSSIP_MAX_SIZE_BELLATRIX` (ditto, regarding being identical numerically) and being a bit more robust to attacks from 1MB "attestations" and the like, but for its own purposes, it's better that be as static as possible, since the SSZ processing does in fact guarantee types at compile-time more often than not.

beacon_chain/networking/faststreams_backend.nim
tersec: Sure, the structure there is more or less parallel to the libp2p backend, so the same changes will have to occur, but it amounts to either
- the whole callstack having to care, and it seems a bit wasteful to annotate all of that with `TODOs` when it's broken enough currently I'm not doing any of it now; or
- architecting the check for which size to use in a sort of static-callback (mixin, etc) way so that lower-levels of the code which don't and shouldn't know anything in detail about Eth2 Req/Resp or SSZ types have to know what a fork is, or what Bellatrix or a block is. Possible, and slightly cleaner in some ways (because it obviates the need to pass this constraint down the call stack) but also fragile module-ordering-dependency magic that doesn't gain that much.

If one uses whole-callstack adjustments, compiler errors will naturally occur until the whole callstack is at least naïvely aligned in this regard, including `readResponseChunk` / `readResponse`, and if one adopts the latter approach, it's not actually necessary to annotate or change `readResponseChunk` / `readResponse`. Either way, I'm not sure how useful it is to do so here.

beacon_chain/spec/keystore.nim
tersec: Those `passC`s in `config.nims` are all `switch("passC", "...")`, which don't trip the pragma style check specifically, and which `nim --fullhelp` documents as
`-t, --passC:OPTION        pass an option to the C compiler`
as a command-line option.

There is an inconsistency here between the pragma and compiler switch cases, but the purpose of this change is to increase stylecheck-cleanliness, towards the goal of being able to use `stylecheck:error` at some point rather than `--stylecheck:hint`, so there's no immediate need to change/fix(?) all the `switch("passC", ...)` directives in `config.nims` for this purpose. I'd regard it as kind of a language/compiler design issue/glitch that survived because so little of the Nim ecosystem appears to use any kind of stylecheck, including until recently the compiler itself.

That said, did find and fix another `localPassC` in `validators/keystore_management.nim`: https://github.com/status-im/nimbus-eth2/pull/3613/commits/e7c7fc94efdcfe6bdb7722a4a17eeedba628bd56

beacon_chain/networking/libp2p_streams_backend.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3613/commits/fd54f7cdb2fb082d3e0973059bc9e10bedda2900

Not sure what you mean by "value" here in value before size or size before value -- the type-as-value, the parameter order, etc?

etan-status: Parameter order -- `maxChunkSize, MsgType`, `MsgType, maxChunkSize`
(btw, another thing I just noticed, here it is `MsgType: type`, but `readResponseChunk` is `MsgType: typedesc`, but I think that's just different ways to express the same in Nim)

tersec: I don't think there is a well-established idiom there in terms of those orderings. I chose it because to the extent there was one, it was adjacent/alongside the "use snappy" boolean, and the existing code treated the type as the last parameter consistently.

I don't have strong preferences here, but it was an attempt to retain a consistency with existing code.

beacon_chain/sync/sync_protocol.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3615

PR #3614, opened 2022-05-04
update presets / configs / REST config for v1.1.10

etan-status:
Updated outdated presets / configs / REST config to v1.1.10 specs.

- `TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH` and `PROPOSER_SCORE_BOOST` are
  not yet used in `eth2-networks`, added configurability as TODOs.

- `MIN_ANCHOR_POW_BLOCK_DIFFICULTY` is no longer needed, put on ignore
  list as some Altair devnets still reference it.


tersec (approved):

beacon_chain/spec/presets.nim
```diff
@@ -148,6 +147,7 @@ const
     "DOMAIN_CONTRIBUTION_AND_PROOF",
 
     "TRANSITION_TOTAL_DIFFICULTY", # Name that appears in some altair alphas, obsolete, remove when no more testnets
+    "MIN_ANCHOR_POW_BLOCK_DIFFICULTY", # Name that appears in some altair alphas, obsolete, remove when no more testnets
```
tersec, line 150: It's marked here so it can be found and removed if desired, but in general, I don't know how relevant it is to support such Altair devnets. There's not much cost this way, so seems reasonable to include it for now.

PR #3615, opened 2022-05-05
ensure MAX_CHUNK_SIZE usage consistent in sync_protocol

tersec:
Ideally, this wrapper function could/would be moved to `eth2_network`, and `maxChunkSize` not exported from there, but this at least ensures consistency within `sync_protocol` and use of `maxChunkSize` uniformly.

Attempts to move it to `eth2_network` thus far (using templates, the async machinery, not-the-async-machinery, procs/funcs, explicitly re-exporting from `sync_protocol` to make it indististinguishable despite that being in principle not necessary, `type` vs `typedesc`, using `ForkySignedBeaconBlock` vs some other more explicit representation, etc) all just result in one `sync_protocol` `readChunkPayload` not finding the `eth2_network` one, so this is a compromise.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -614,10 +614,10 @@ when useNativeSnappy:
 else:
   include libp2p_streams_backend
 
-func maxChunkSize(t: typedesc[bellatrix.SignedBeaconBlock]): uint32 =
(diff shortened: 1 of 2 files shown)


etan-status (approved):
Looks clean :) 

Weird that the `ForkySignedBeaconBlock`  works in `readChunkPayload` but not in `maxChunkSize` for the `when t.toFork >= BeaconBlockFork.Bellatrix`, but it is what it is.

PR #3623, opened 2022-05-09
update to latest light client libp2p protocol

etan-status:
Incorporates the latest changes to the light client sync protocol based
on Devconnect AMS feedback. Note that this breaks compatibility with the
previous prototype, due to changes to data structures and endpoints.

Due to the amount of changes, it is best to review the heavily edited
files from scratch (likewise, for the spec).

### DIFF
--- a/beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
@@ -13,13 +13,16 @@
 import
   # Status libraries
   stew/bitops2,
+  chronos,
   # Beacon chain internals
   ../spec/datatypes/altair,
   ./block_dag
 
 type
-  OnOptimisticLightClientUpdateCallback* =
-    proc(data: OptimisticLightClientUpdate) {.gcsafe, raises: [Defect].}
+  OnLightClientFinalityUpdateCallback* =
+    proc(data: altair.LightClientFinalityUpdate) {.gcsafe, raises: [Defect].}
+  OnLightClientOptimisticUpdateCallback* =
+    proc(data: altair.LightClientOptimisticUpdate) {.gcsafe, raises: [Defect].}
 
   ImportLightClientData* {.pure.} = enum
     ## Controls which classes of light client data are imported.
@@ -30,18 +33,17 @@ type
     Full = "full"
       ## Import light client data for entire weak subjectivity period.
     OnDemand = "on-demand"
-      ## No precompute of historic data. Is slow and may miss validator duties.
+      ## Don't precompute historic data. Slow, may miss validator duties.
 
   CachedLightClientData* = object
     ## Cached data from historical non-finalized states to improve speed when
     ## creating future `LightClientUpdate` and `LightClientBootstrap` instances.
     current_sync_committee_branch*:
       array[log2trunc(altair.CURRENT_SYNC_COMMITTEE_INDEX), Eth2Digest]
-
     next_sync_committee_branch*:
       array[log2trunc(altair.NEXT_SYNC_COMMITTEE_INDEX), Eth2Digest]
 
-    finalized_bid*: BlockId
+    finalized_slot*: Slot
     finality_branch*:
       array[log2trunc(altair.FINALIZED_ROOT_INDEX), Eth2Digest]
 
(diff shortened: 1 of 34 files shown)


etan-status:
Tested this against the test cases in the consensus-specs repo, and also manually checked logs of local testnet to ensure the gossip is properly sent.

tests/consensus_spec/altair/test_fixture_sync_protocol_update_ranking.nim
```diff
@@ -0,0 +1,87 @@
...
+          -1
+        elif b.is_better_update(a):
+          1
+        else:
+          0
+      check: updates.isSorted(cmp)
+
+  `testImpl _ sync_protocol_update_ranking _ identifier`()
+
+suite "EF - Altair - Sync protocol - Update ranking" & preset():
```
tersec, line 64: Is it useful to have Bellatrix versions of these, i.e. would they be different in any useful way, or just exact duplicates on exact duplicate inputs?

etan-status: Different fork-digest, but semantically same for now.

etan-status:
Ran CI tests 5 times, flakiness seems to be gone now. There was some randomness involved, in the way how we simulated sync committee signatures, where sometimes it would generate too many and would actually reach supermajority, breaking the test case assumption. This is now fixed by putting a cap on randomization to not exceed the supplied target participation percentage.

beacon_chain/spec/network.nim
```diff
@@ -98,10 +98,15 @@ func getSyncCommitteeContributionAndProofTopic*(forkDigest: ForkDigest): string
...
   eth2Prefix(forkDigest) & "sync_committee_contribution_and_proof/ssz_snappy"
 
-# https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#optimistic_light_client_update
-func getOptimisticLightClientUpdateTopic*(forkDigest: ForkDigest): string =
-  ## For broadcasting or obtaining the latest `OptimisticLightClientUpdate`.
-  eth2Prefix(forkDigest) & "optimistic_light_client_update_v0/ssz_snappy"
+# https://github.com/ethereum/consensus-specs/blob/vFuture/specs/altair/sync-protocol.md#light_client_finality_update
+func getLightClientFinalityUpdateTopic*(forkDigest: ForkDigest): string =
+  ## For broadcasting or obtaining the latest `LightClientFinalityUpdate`.
+  eth2Prefix(forkDigest) & "light_client_finality_update_v0/ssz_snappy"
```
tersec, line 104: Might be useful to have tests for this, along the lines of the tests for other functions in this module, specifically because of the `v0` aspect -- it will be good to make sure everything gets changed synchronously, and have the tests to demonstrate that.

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
```diff
@@ -13,13 +13,16 @@
 import
   # Status libraries
   stew/bitops2,
+  chronos,
```
tersec: What change prompts this to need `chronos` now? It's not exported to somewhere else and nothing obvious stands out.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -975,21 +758,20 @@ proc getLightClientBootstrap*(
     return err()
 
   let bdata = dag.getForkedBlock(blockRoot).valueOr:
-    debug "Bootstrap unavailable: Block not found", blockRoot
+    debug "LC bootstrap unavailable: Block not found", blockRoot
```
tersec, line 761: Is "LC" a otherwise used acronym for light client, and/or how valuable is it to keep the messages a bit shorter versus use a phrase people have a higher chance of understanding? "LC" probably isn't particularly Google-search-friendly in this regard either.

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
etan-status: `Future`

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Consistently using full `Light client` exceeds the 80-col mark in various places, just `light` is not greppable because you find all the Lighthouse things, no prefix makes logs more difficult to find, LC works well when filtering the logs and is in line with VC / BN. Agree though that it's not optimal.

beacon_chain/spec/network.nim
etan-status: Oh, I missed those. Extended test_honest_validator.nim accordingly, I assume you meant those.

tersec: Yeah.

tersec (approved):

beacon_chain/spec/helpers.nim
```diff
@@ -475,11 +479,119 @@ func has_flag*(flags: ParticipationFlags, flag_index: int): bool =
...
+
+func toMeta*(update: SomeLightClientUpdate): LightClientUpdateMetadata =
+  var meta {.noinit.}: LightClientUpdateMetadata
+  meta.attested_slot =
+    update.attested_header.slot
+  meta.finalized_slot =
+    when update is SomeLightClientUpdateWithFinality:
+      update.finalized_header.slot
+    else:
+      0.Slot
```
tersec: There's inconsistent usage of `GENESIS_SLOT` and `0.Slot` in places, including I think in the spec itself. Not sure that it matters here, as a result, but it'd be nice to have a consistent convention (see also `GENESIS_EPOCH`).

etan-status: Corrected it here for `GENESIS_SLOT`, thanks for pointing it out.

There does not seem to be any other correspondence in this PR, or in the matching spec PR, so I assume that your comment is more meant in a general way.

Note that there are also implicit assumptions of GENESIS_SLOT being 0. If that is not always true, e.g., for private networks, it could promote this cosmetic issue to a real problem.

tersec: Yes, it's a more general comment. I sometimes look at switching all `0.Slot` (and equivalent) to `GENESIS_SLOT` or vice versa, but both are referred to in the spec, so I refrain. Practically, I prefer `GENESIS_SLOT` slightly because Nim will put it in the constant pool as the correct type, rather than maybe runtime-construct it as it tends to do `default(SomeType)` even though that's really a compile-time constant.

tersec:
Linux i386 Nim 1.2:
```
[Suite] Light client [Preset: mainnet]
  [OK] Pre-Altair
    /home/runner/work/nimbus-eth2/nimbus-eth2/tests/test_light_client.nim(169, 29): Check failed: store.finalized_header == finalityUpdate.get.finalized_header
    store.finalized_header was (slot: 8192, proposer_index: 1, parent_root: fdedfe9932eb02cfb9380be5c86edfa03e54067726c084e6b29a440b16249233, state_root: f546a3f3e654a863614395ff7effdaf74850ddb3d23c6aa77ae54b7557efe74a, body_root: 2e0c9e75376f5dd11b4ba65c3688f265c84f7898fbc41908efc456856cdbfbff)
    finalityUpdate.get.finalized_header was (slot: 16384, proposer_index: 21, parent_root: 5d0d71a63d55c2ec04f148a2a57a9f9486b1906597b06fb01f9b0078f22bd869, state_root: 2d456114eb16f369a93b973d46ba2ff605f55dd45414afdc3c4f5375b98c398e, body_root: f8cdeb7bbb0e6061b17a21e7683ff124b13ab115265399c7ebe2c83a338b5376)
  [FAILED] Light client sync
  [OK] Init from checkpoint
```

tersec:
https://ci.status.im/blue/organizations/jenkins/nimbus%2Fnimbus-eth2%2Fmultibranch/detail/PR-3623/19/pipeline/87 shows the same thing, even after https://github.com/status-im/nimbus-eth2/pull/3623/commits/9bcc2b3bcae56a34bea2462ddc9dc4612c086548:
```
[Suite] Light client [Preset: mainnet]
  [OK] Pre-Altair
    /Users/jenkins/workspace/_nimbus-eth2_multibranch_PR-3623/macos_arm64/tests/test_light_client.nim(169, 29): Check failed: store.finalized_header == finalityUpdate.get.finalized_header
    store.finalized_header was (slot: 8192, proposer_index: 1, parent_root: fbbf0fc9bd94b04c596782cb49853466858a607c7cf3d67a0dc92bc34d8047c4, state_root: f9a3981b961bf781b820603304da3554f7c1fd79d07c1a432b5f0c02c60b25cf, body_root: 99731a01c1c648a16454e1d6dc79335c1804f2f647ec394a99f98f5088f29895)
    finalityUpdate.get.finalized_header was (slot: 16384, proposer_index: 21, parent_root: 5bbfc1240f2524cf9c817541ac3b4256065050fed506c0b9ff8f0a266911c288, state_root: c64332da8ad60c7868de5dc26c2507ec560b205853829779be72b25a4faab54e, body_root: 792ac60a443e2052cc6f52067d480d273f2ed493bbf0661a9e3935a68dd09b7e)
  [FAILED] Light client sync
  [OK] Init from checkpoint
```

beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
```diff
@@ -55,40 +57,44 @@ type
...
+
+    latestForwardedOptimisticSlot*: Slot
+      ## Latest optimistic update that was forwarded on libp2p gossip.
+      ## Tracks `optimistic_update.attested_header.slot`.
+
+    latestBroadcastedSlot*: Slot
+      ## Latest slot for which updates were broadcasted on libp2p gossip.
+      ## Tracks `update.signature_slot`.
+
+    broadcastGossipFut*: Future[void]
```
arnetheduck: this belongs in `BeaconNode` (which is where validator_duties keeps things for its own needs)

etan-status: Moved it into a separate pool type, as it is also used in `gossip_validation` (minus the `Future`). I think it's a nice compromise between having it in the `block_pools_types_light_client` where it clearly does not belong, and fragmenting these few vars across too many files.

etan-status:
Manually validated against light client poc.

PR #3628, opened 2022-05-12
Bump Chronos to fix HTTP client assertion failures

zah:
The other changes should be benign.


etan-status:
Note that the referenced Chronos is not part of `master`, but .gitmodules refers to `master`. Also, commit hash may change after squashing.

etan-status:
Alternative: https://github.com/status-im/nimbus-eth2/pull/3630

zah:
The two fixes have been combined here. The bumped revision now points to the latest Chronos master.

PR #3635, opened 2022-05-15
Document the Web3Signer setups (both regular and distributed)

zah:


docs/the_nimbus_book/src/web3-signer.md
```diff
@@ -0,0 +1,100 @@
+# Web3Signer
+
+[Web3Signer](https://docs.web3signer.consensys.net/en/latest/) is a remote signing server developed by Consensys. It offers a [standardized REST API](https://consensys.github.io/web3signer/web3signer-eth2.html) allowing the Nimbus beacon node or validator client to operate without storing any validator keys locally.
+
+Remote validators can be permanently added to a Nimbus installation (or more precisely to a particular data directory) either on-the-fly through the [`POST /eth/v1/remotekeys`](https://ethereum.github.io/keymanager-APIs/#/Remote%20Key%20Manager/ImportRemoteKeys) request when the [Keymanager API](./keymanager-api.md) is enabled or by manually creating a remote keystore file within the validators directory of the client which will be loaded upon the next restart. The validators directory can be specified through the `--validators-dir` command-line parameter. The default value is `$DATA_DIR/validators` where $DATA_DIR is the directory specified through the `--data-dir` parameter.
```
arnetheduck, line 5: manually creating? shouldn't this be a `deposit remote import` command to match the local keystore flow? it seems odd that we're suddenly asking users to edit files in the data folder of nimbus - this is a significant departure from the UX up to this stage

zah: This documentation reflects the current functionality. We will change the documentation once we have additional ways to import keys. I've been planning to add a CLI interface for interacting with the Keymanager API and the same interface can work against a data-dir perhaps.

zah: Please note that these file formats are going to remain a part of our stable interface that must be described in the documentatiion because projects such as DappNode are doing some automation based on them.

arnetheduck: > a part of our stable interface

hm, that's potentially an issue - why are they not using the web3 interface / rpc commands or the `deposit` set of commands? these files are not standardised across clients and I don't expect them to be - it would be better that dappnode didn't use them either / they were not part of the stable set _unless_ they become standard.. 

for example, right now we import keystores from the EF format to our own - this is kind of an unnecessary step to be taking, we could be using the keystore files from eth2-cli directly - maintaining stable on-disk config dir format is  a cost (just like we don't provide a stable db: we provide era files as a standard instead.

arnetheduck: also, keymanager API isn't ideal - `nbc deposit` is better in that it doesn't requiire a running instance

arnetheduck: > This documentation reflects the current functionality.

well, if we want users to use a command, we shouldn't document a different way of doing things, because these will end up in guides and manuals, and then we can't change them any more - has this UX been discussed somewhere?

zah: I've been helping DappNode migrate to the Keymanager API, so they might stop relying on this, but the format has been documented by @cheatfate in the original PR introducing the functionality, so others might be using it as well:

https://github.com/status-im/nimbus-eth2/pull/3077

Also, since these are text files, it's pretty inviting to look at the format and to target it directly. In the DevOps world, generating configuration files from templates is usually preferable to running a tool that mutates a configuration file. Since the format is versioned, there is just no good reason for us to break it anyway.

>  also, keymanager API isn't ideal - nbc deposit is better in that it doesn't requiire a running instance

Creating files comes with downsides as well as the client must be restarted to pick-up the changes, but I've mentioned that the CLI can target both scenarios potentially.

> import keystores from the EF format to our own

There is no "our format" when it comes to regular keystores btw. Nimbus ca load the original keystores from the eth2-deposit-cli tool just fine. The problem is that they are all encoded with the same password, so we prefer not to ask the user to generate N password files manually, all with the same content.

arnetheduck: > Also, since these are text files, it's pretty inviting to look at the format and to target it directly.

well, the point is more that when there are multiple ways to do things, there's extra maintenance and less testing in general - it also adds unnecessary confusion in the manual when we recommend different ways for similar operations ("for these keystores, use a command - for these others, quite arbitrarily, use text files") - it would be better to settle on one principal way of doing things and recommend it throughout the manual (ie that the UX is deliberate and not a hodge-podge of random ways of doing things)

arnetheduck: > There is no "our format" when it comes to regular keystores btw.

hm, they're named different (which presumably is part of the standard) - ie can I just drop one of the deposit cli json files in my validators dir (or better, point `--validators-dir` to wherever the files already are on my drive when creating them)?

arnetheduck: >  In the DevOps world, generating configuration files

basically, another way to look at it, if we really want to make the config files a part of the stable feature set, is that we have a separate devops guide: one work flow for users and one technical document describing the structure of the config directory for integrators (similar to how the e2store doc describes an era file) - regardless, "normal" users should not have to bother with file format descriptors *iff* our main UX is command-based, and therefore, it also doesn't belong in the manual like this.

zah: For me, it's reasonable for the manual to keep evolving and for us to introduce new and better ways to do things. But we already have users inquiring about the Web3Signer configuration, so describing the current functionality is an improvement from the status quo. 

We'll make the next revision when these new keystore management commands are introduced. For me, the long-term goal is that once the user installs Nimbus, the service is already running and executing a single command is enough to add your validators to it.

zah: > hm, they're named different (which presumably is part of the standard) - ie can I just drop one of the deposit cli json files in my validators dir (or better, point --validators-dir to wherever the files already are on my drive when creating them)?

The name of the keystore file is not standardized, but we can easily recognize different layouts. The problem is that the deposit-cli tool doesn't produce password files (which are not standardized either). The goal of the import procedure for us is to obtain the decryption password from the user.

arnetheduck: > For me, it's reasonable for the manual to keep evolving 

well, we can keep evolving it, but also take into account the user story while we're doing it - I often have to point users to the documentation, and this document is confusing when I look at it from the point of view of the user - one such way is to evolve it in a deliberate way (just like features should be developed with a deliberate UX in mind where consistency is kept in mind) - this means, for the purposes of this guide, that the file descriptions and other "structural" documentation entries are moved to a dedicated section - in the meantime, the keystore doc  for users can link to there

arnetheduck: > For me, the long-term goal is that once the user installs Nimbus, the service is already running and executing a single command is enough to add your validators to it.

I wouldn't want my nimbus to work this way, that's for sure - ie I don't want anything to start automatically on installation unless I explicitly enabled it - just like I don't want a second way of supervising things outside of systemd (ie I don't want any supervisor processes that launch processes or perform upgrade - systemd already performs this task and has had years of dedicated development behind it to do it well) - I also don't want to have to run a service to set up my keys and configuration - I don't want to keep that service running (with the accompanying security issues) with an open port when nimbus is ready and configured - this is an unnecessary risk - for many validators, configuration is something that's done once / rarely: no need to increase the security surface area this way by default except for complex setups that require dynamic configuration.

zah: I think you are not quite representative of a typical user because most services don't work in the way you would prefer. With control mechanisms typically exposed over Unix domain sockets by default, most software that is designed to work all the time is automatically configured to run at start-up after installation. Having auto-update is less of a typical choice, but I would say it's quite appropriate for our audience and the type of software that we are building (one that can get dysfunctional upon a hard-fork). 

It's perfectly fine to have preferences, but we must decide at some point how do we imagine the users we build the software for and what is the appropriate out-of-the-box behavior (I would claim our software should be designed for non-techies).

arnetheduck: Well, the point is more that there exist tools that do these jobs well, because that's what they have been doing for a long time - the package manager upgrades things and systemd makes sure they're running at the times that they are supposed to - adding a layer that does the same job but poorly adds complexity and surface area without any UX benefit (on the contrary) - spending as much time on upgrade and process management as the systemd and package manager people have done to handle corner cases will be difficult to motivate, where there are things that are much closer to our expertise remaining to be done.

zah:
This work has been revisited and merged with the following PR:
https://github.com/status-im/nimbus-eth2/pull/4912

PR #3636, opened 2022-05-16
Bugfix: Don't include invalid deposits in block proposals

zah:
This would result in our block being rejected (orphaned) in the network.
See the provided comments in `process_deposit` for more details.

### DIFF
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -85,7 +85,7 @@ proc process_randao(
   let
     epoch = state.get_current_epoch()
 
-  if skipBlsValidation notin flags:
+  if not flags.shouldSkipBls:
     let proposer_pubkey = state.validators.asSeq()[proposer_index.get].pubkey
 
     if not verify_epoch_signature(
@@ -154,7 +154,7 @@ proc check_proposer_slashing*(
     return err("check_proposer_slashing: slashed proposer")
 
   # Verify signatures
-  if skipBlsValidation notin flags:
+  if not flags.shouldSkipBls:
     for signed_header in [proposer_slashing.signed_header_1,
         proposer_slashing.signed_header_2]:
       if not verify_block_signature(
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/state_transition_block.nim
```diff
@@ -294,8 +294,22 @@ proc process_deposit*(cfg: RuntimeConfig,
     # Increase balance by deposit amount
     increase_balance(state, index.ValidatorIndex, amount)
   else:
-    # Verify the deposit signature (proof of possession) which is not checked
-    # by the deposit contract
+    # We must verify the deposit signature which is not checked by the
+    # deposit contract. This check can be omitted only when replying
+    # previously verified blocks.
```
arnetheduck, line 299: per https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits=, deposits wiht invalid signatures are also included in the deposit order - they just don't lead to a validator being added (see the early `return` on signature verification.

the code, as implemented, is correct: `TrustedBeaconBlock` skips (via type magic) the redundant signature checks but the deposit signature check must be redone every replay as well (this, together with the linear key scan, is why replays with deposits are much slower)

zah: That's a good point. Maybe `skipBlsValidation` just doesn't belong here, because when you use it you are effectively saying "I'm willing to skip some BLS checks even if this means changing the result of the state transition function" which sounds like a very niche use case and potentially a dangerous gotcha.

arnetheduck: yeah, the intent was for `skipBlsValidation` is to go away - effectively, it's used only in tests (to speed them up) and a few more places (but not block-related, afair) which ideally should be refactored to use types instead (ie `TrustedSig`) after having done the necessary sig checks

zah:
Closing in favor of https://github.com/status-im/nimbus-eth2/pull/3639

PR #3639, opened 2022-05-16
Bugfix: Invalid blocks were produced in the presence of invalid deposits

zah:
Since we were not verifying BLS signature in blocks that we produce,
we were failing to notice that some deposits need to be ignored (due
to having an invalid signature). Processing these deposits resulted
in a different ending state after the state transition which caused
our blocks to be rejected by the network.

### DIFF
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -296,7 +296,7 @@ proc process_deposit*(cfg: RuntimeConfig,
   else:
     # Verify the deposit signature (proof of possession) which is not checked
     # by the deposit contract
-    if skipBlsValidation in flags or verify_deposit_signature(cfg, deposit.data):
+    if verify_deposit_signature(cfg, deposit.data):
       # New validator! Add validator and balance entries
       if not state.validators.add(get_validator_from_deposit(deposit.data)):
         return err("process_deposit: too many validators")
--- a/tests/testblockutil.nim
+++ b/tests/testblockutil.nim
@@ -124,9 +124,11 @@ proc addTestBlock*(
       sync_aggregate,
       default(ExecutionPayload),
       noRollback,
-      cache)
+      cache,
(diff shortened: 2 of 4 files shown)


beacon_chain/spec/state_transition_block.nim
```diff
@@ -296,7 +296,7 @@ proc process_deposit*(cfg: RuntimeConfig,
   else:
     # Verify the deposit signature (proof of possession) which is not checked
     # by the deposit contract
-    if skipBlsValidation in flags or verify_deposit_signature(cfg, deposit.data):
+    if verify_deposit_signature(cfg, deposit.data):
```
arnetheduck, line 299: we have tests that don't create signatures for deposits but expect them to succeed

arnetheduck, line 299: we have tests that don't create signatures for deposits but expect them to succeed

arnetheduck (approved):

arnetheduck:
good find :tada:

tests/testblockutil.nim
```diff
@@ -124,9 +124,11 @@ proc addTestBlock*(
       sync_aggregate,
       default(ExecutionPayload),
       noRollback,
-      cache)
+      cache,
+      verificationFlags = {skipBlsValidation})
```
arnetheduck, line 128: ping @zah this should be `flags`

zah: I think this should be removed, but there wasn't enough time before the release to investigate how the tests can be fixed. Otherwise, feel free to rename it although the `flags` name would be a bit misleading in a function called `makeBeaconBlock` (the flags don't affect how the block is created, but rather how it's executed (or verified), hence my name).

arnetheduck: we use `flags` as a convention for this variable throughout the spec code which makes the code easier to read - it would be nice if this code adhered to it as well

PR #3642, opened 2022-05-18
enable hard failures for nim 1.6 in CI

arnetheduck:
makes errors more visible, now that it's sort of working again

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,8 +232,8 @@ jobs:
       - name: Run tests
         run: |
           if [[ "${{ matrix.branch }}" == "version-1-6" ]]; then
-            # hide the CI failure in GitHub's UI
-            ${make_cmd} -j ${ncpu} V=1 NIM_COMMIT=${{ matrix.branch }} DISABLE_TEST_FIXTURES_SCRIPT=1 test || true
+            # change to "|| true" to hide the CI failures in GitHub's UI (escape hatch if a blocker is detected in 1.6)
+            ${make_cmd} -j ${ncpu} V=1 NIM_COMMIT=${{ matrix.branch }} DISABLE_TEST_FIXTURES_SCRIPT=1 test || false
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -232,8 +232,8 @@ jobs:
       - name: Run tests
         run: |
           if [[ "${{ matrix.branch }}" == "version-1-6" ]]; then
-            # hide the CI failure in GitHub's UI
-            ${make_cmd} -j ${ncpu} V=1 NIM_COMMIT=${{ matrix.branch }} DISABLE_TEST_FIXTURES_SCRIPT=1 test || true
+            # change to "|| true" to hide the CI failures in GitHub's UI (escape hatch if a blocker is detected in 1.6)
+            ${make_cmd} -j ${ncpu} V=1 NIM_COMMIT=${{ matrix.branch }} DISABLE_TEST_FIXTURES_SCRIPT=1 test || false
```
tersec, line 236: Why not just remove `|| true` entirely instead of replacing it with `|| false`?

arnetheduck: can't say I trust the status quo - our 1.6 tests build with the 1.6 devel branch, not 1.6.6 specifically, so this way it stays within easy reach to disable once again

tersec (approved):

.github/workflows/ci.yml
tersec: That'd be a useful change, then, to prioritize testing with a 1.6 version one might ship with. Ideally, of course, testing with both, but I don't know how resource-intensive that might become. One option there is to test with 1.6.x (but some candidate shippable version) now, and swap the current 1.2 slot back to devel once we're satisfied enough to drop 1.2 in nimbus-eth2.

This creates a situation where at no point is there externally-imposed urgency from the Nim side, at the code of sometimes being a bit blind to upcoming changes in the 1.6.x series, but the latter should be settling down, and shipping 1.6 without testing on a non-devel version can't really happen.

PR #3643, opened 2022-05-19
only send JWT headers in Bellatrix-capable networks

tersec:


tersec:
https://github.com/status-im/nimbus-eth2/pull/3645

PR #3647, opened 2022-05-20
More spec-compliant handling of unknown fields in REST json

zah:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -694,7 +722,7 @@ proc readValue*[BlockType: Web3SignerForkedBeaconBlock|ForkedBeaconBlock](
                                     "ForkedBeaconBlock")
       data = some(reader.readValue(JsonString))
     else:
-      reader.raiseUnexpectedField(fieldName, "ForkedBeaconBlock")
+      unrecognizedFieldWarning()
```
arnetheduck, line 728: this will get noisy with lists - ie a list with 100k blocks during sync will result in 100k warning logs - the fact that an unrecognised field was hit should instead be passed up to the caller that can take appropriate action

arnetheduck (commented):
lgtm, except that serializers must never log (it's not the right layer for it)

zah:
I can imagine an API for notifying the caller, but implementing something like this would be a larger development effort. For this hotfix, we can either disable this warning (silently ignoring the fields) or we can assume that these warnings won't be encountered in practice (they can arise only when using the REST API, so the potential excessive output shouldn't be a security concern).

zah:
I've changed the warning to `debug` for now and added some TODO comments:
a7f37ce26275d2072a4507635ac14c56889618fe

arnetheduck:
> we can either disable this warning

I would tend towards complete disabling (or `trace`): we run debug on pretty much all servers and logging in serializers has a tendency to blow up - ie if an attestation gets logged, that's 128 logs per block etc - it's really bad if it happens.

arnetheduck (approved):

PR #3648, opened 2022-05-20
Add support for the Ropsten beacon chain

zah:


beacon_chain/networking/network_metadata.nim
```diff
@@ -217,11 +228,16 @@ template eth2Network(path: string, eth1Network: Eth1Network): Eth2NetworkMetadat
...
+template mergeTestnet(path: string, eth1Network: Eth1Network): Eth2NetworkMetadata =
+  loadCompileTimeNetworkMetadata(mergeTestnetsDir & "/" & path,
+                                 some eth1Network)
+
 when not defined(gnosisChainBinary):
   when const_preset == "mainnet":
     const
       mainnetMetadata* = eth2Network("shared/mainnet", mainnet)
       praterMetadata* = eth2Network("shared/prater", goerli)
+      ropstenMetadata = mergeTestnet("ropsten-beacon-chain", ropsten)
```
arnetheduck, line 240: just `ropsten`?

arnetheduck: ah, nvm, they renamed the config but not the folder

arnetheduck (approved):
cc @unixpi this will need a corresponding update in the book where we show users how to set up nimbus with an EL post-merge

beacon_chain/networking/network_metadata.nim
zah: This is a name of a metadata directory in a third-party repo.

PR #3649, opened 2022-05-21
fix handling of multiple fork transitions per epoch

tersec:
`validateForkSchedule` over-validates in the case where `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH`. https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/fork.md#upgrading-the-state specifies
```python
        fork=Fork(
            previous_version=pre.fork.current_version,
            current_version=BELLATRIX_FORK_VERSION,
            epoch=epoch,
        ),
```
and since one has to conceptually upgrade state from phase 0 to Altair to Bellatrix, and https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/fork.md#upgrading-the-state creates the Altair fork via
```python
        fork=Fork(
            previous_version=pre.fork.current_version,
            current_version=ALTAIR_FORK_VERSION,
            epoch=epoch,
        ),
```
this implies that `previous_version` in a Bellatrix state will either be (in a test scenario where it starts with Bellatrix) `BELLATRIX_FORK_VERSION`, or in a non-test scenario, `ALTAIR_FORK_VERSION`.

When `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH`, therefore, one runs `upgrade_to_bellatrix(upgrade_to_altair(preState))` conceptually, and the fork schedule will look something like:
1. (`previous_version`: `GENESIS_FORK_VERSION`, `current_version`: `GENESIS_FORK_VERSION`, `epoch`: `ALTAIR_FORK_EPOCH`)
2. (`previous_version`: `ALTAIR_FORK_VERSION`, `current_version`: `BELLATRIX_FORK_VERSION`, `epoch`: `BELLATRIX_FORK_EPOCH`)

The Bellatrix consensus spec fork tests confirm that this is the intended interpretation empirically and have a `postState` that looks like
> (`previous_version`: `ALTAIR_FORK_VERSION`, `current_version`: `BELLATRIX_FORK_VERSION`, `epoch`: `BELLATRIX_FORK_EPOCH`)

even when `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH == 0`, i.e. there was never any observable moment in beacon chain-time where the `previous_version` Altair fork was ever externally observable.

The fork schedule doesn't, therefore, one-to-one match the state fork linked list version chain the same and can and does skip fork versions entirely.

There's another odd side-effect of this: anything slightly in the past, such as incorporating attestations even a slot or two old, because problematic at just around this threshold, exactly because of the lack of overlap of these fork versions in a linked-list fashion. The failure mode, and it's unclear it's a bug in Nimbus per se, is that attestations near the end of a fork time period get signed with `GENESIS_FORK_VERSION` but then by the time they're incorporated into a block and verified, `GENESIS_FORK_VERSION` is neither `previous_version` nor `current_version`. Therefore, the `get_domain` (https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#get_domain) algorithm does not suffice:
```python
def get_domain(state: BeaconState, domain_type: DomainType, epoch: Epoch=None) -> Domain:
    """
    Return the signature domain (fork version concatenated with domain type) of a message.
    """
    epoch = get_current_epoch(state) if epoch is None else epoch
    fork_version = state.fork.previous_version if epoch < state.fork.epoch else state.fork.current_version
    return compute_domain(domain_type, fork_version, state.genesis_validators_root)
```
There's no way to access the correct `fork_version` to verify the message, because there's no overlap in `fork_version`s between `Fork`s, but rather this noncontiguous cutover. In Nimbus, https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#is_valid_indexed_attestation catches this first, because
```python
    domain = get_domain(state, DOMAIN_BEACON_ATTESTER, indexed_attestation.data.target.epoch)
```
only knows about `fork_version`s in `state`, but not the `fork_version` which created the attestation.

It would be possible for Nimbus to correctly handle these by refactoring things to use the `forkAtEpoch` helper rather than trust `BeaconState.fork`, but that would violate the spec as written for not much gain, currently, so leaving that alone.

### DIFF
--- a/beacon_chain/spec/forks.nim
+++ b/beacon_chain/spec/forks.nim
@@ -457,8 +457,9 @@ func altairFork*(cfg: RuntimeConfig): Fork =
     epoch: cfg.ALTAIR_FORK_EPOCH)
 
 func bellatrixFork*(cfg: RuntimeConfig): Fork =
-  # TODO in theory, the altair + merge forks could be in same epoch, so the
-  # previous fork version would be the GENESIS_FORK_VERSION
+  # TODO https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/beacon-chain.md#testing
+  # provides an example of `previous_version` also being BELLATRIX_FORK_VERSION
+  # in certain testing scenarios
   Fork(
     previous_version: cfg.ALTAIR_FORK_VERSION,
     current_version: cfg.BELLATRIX_FORK_VERSION,
@@ -482,13 +483,20 @@ func nextForkEpochAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): Epoch =
   of BeaconStateFork.Altair:    cfg.BELLATRIX_FORK_EPOCH
   of BeaconStateFork.Phase0:    cfg.ALTAIR_FORK_EPOCH
 
-func getForkSchedule*(cfg: RuntimeConfig): array[3, Fork] =
+func getForkSchedule*(cfg: RuntimeConfig): seq[Fork] =
   ## This procedure returns list of known and/or scheduled forks.
   ##
   ## This procedure is used by HTTP REST framework and validator client.
-  ##
-  ## NOTE: Update this procedure when new fork will be scheduled.
-  [cfg.genesisFork(), cfg.altairFork(), cfg.bellatrixFork()]
+
+  # NOTE: Update this function when new fork will be scheduled.
+  static: doAssert BeaconStateFork.Bellatrix == high(BeaconStateFork)
+
+  var forkSchedule = @[cfg.bellatrixFork()]
+  for forkInfo in [cfg.altairFork(), cfg.genesisFork()]:
+    # Detect multiple fork transitions on same epoch transition
+    if forkInfo.epoch != forkSchedule[0].epoch:
+      forkSchedule.insert(forkInfo, 0)
+  forkSchedule
 
 type
   # The first few fields of a state, shared across all forks
--- a/beacon_chain/validator_client/fork_service.nim
+++ b/beacon_chain/validator_client/fork_service.nim
@@ -4,19 +4,6 @@ import common, api
 
 logScope: service = "fork_service"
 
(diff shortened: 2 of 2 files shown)


cheatfate:
I dont think that elimination of validation logic is a good way of doing things. What we are trying to achieve is mostly needed for testing purposes but its not applicable to real world. There will not be present such scenario in mainnet. But as soon as you remove validation logic you going to trust any type of response from `beacon_node`. 
Why not upgrade validation logic?

tersec:
There are various heuristics one could use to detect this, with the level of information the VC is working with, none is ideal. However, https://github.com/status-im/nimbus-eth2/pull/3649/commits/728c1521eae1a0c731cf9069fc10199628cae125 should be reasonably solid.

tersec:
This isn't, in general, reliably possible: https://github.com/ethereum/consensus-specs/issues/2902

zah:
So, what's the verdict on this? Isn't this PR effectively creating a deviation from the spec?

tersec:
There are two concepts here from the spec that interact in ways I don't see defined:

- the consensus specs about the evolution of the `BeaconState` and how to resolve attestations, which I agree is fairly unambiguous, and if Nimbus were to try to be more enthusiastic about finding the correct fork versions, that would, indeed, represent a deviation from spec; and

- `getForkSchedule`, which according to https://github.com/ethereum/beacon-APIs/blob/master/apis/config/fork_schedule.yaml states that it should
> Retrieve all forks, past present and future, of which this node is aware.

One could argue this means that even forks that don't get scheduled as such should be included, but then, that's what this PR gets at -- is the validator client over-validating?

tersec:
Closing, since https://github.com/status-im/nimbus-eth2/pull/3691 disallows this situation from occurring except in the special case of starting from non-phase0, but that doesn't need the machinery here.

PR #3652, opened 2022-05-23
per-validator suggested fee recipients

tersec:
Inspired by https://lighthouse-book.sigmaprime.io/suggested-fee-recipient.html (`--suggested-fee-recipient-file` and the file format).


zah:
I was thinking that we can store the fee recipient metadata within the validator directories (one file per validator) or more generally to treat the configuration as a set of database entries instead of it being a configuration file. I have two rationales for this:

1) It should be easy to change the fee recipient using the REST API (this is standardized in the Keymanager API).
2) Providing a command-line interface for the feature would be more user-friendly than describing a file format (we'll also have  GUI for it in the future).

jclapis:
The Rocket Pool stack is currently designed around this Lighthouse model, where there is a file that specifies the default fee recipient for all validators and then has individual overrides per pubkey. Lighthouse also reloads this file and updates itself every time it proposes a block, which is extremely helpful and highly desirable for us. Teku and Prysm use a similar approach, though they both share the same JSON spec for this file rather than have their own custom formats like Lighthouse.

tersec:
> I was thinking that we can store the fee recipient metadata within the validator directories (one file per validator) or more generally to treat the configuration as a set of database entries instead of it being a configuration file.

https://github.com/status-im/nimbus-eth2/pull/3652/commits/b57eaa38b114df1ef52dd4ac43509aa9c3f247ed

arnetheduck:
shouldn't we have the same for graffiti?

PR #3653, opened 2022-05-23
add light client (standalone)

etan-status:
Introduces a new library for syncing using libp2p based light client
sync protocol, and adds a new `nimbus_light_client` executable that uses
this library for syncing. The new executable emits log messages when
new beacon block headers are received, and is integrated into testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,17 +35,17 @@ CPU_LIMIT := 0
 BUILD_END_MSG := "\\x1B[92mBuild completed successfully:\\x1B[39m"
 
 ifeq ($(CPU_LIMIT), 0)
-  CPU_LIMIT_CMD :=
+	CPU_LIMIT_CMD :=
 else
-  CPU_LIMIT_CMD := cpulimit --limit=$(CPU_LIMIT) --foreground --
+	CPU_LIMIT_CMD := cpulimit --limit=$(CPU_LIMIT) --foreground --
 endif
 
 # TODO: move this to nimbus-build-system
 ifeq ($(shell uname), Darwin)
-  # Scary warnings in large volume: https://github.com/status-im/nimbus-eth2/issues/3076
-  SILENCE_WARNINGS := 2>&1 | grep -v "failed to insert symbol" | grep -v "could not find object file symbol for symbol" || true
+	# Scary warnings in large volume: https://github.com/status-im/nimbus-eth2/issues/3076
+	SILENCE_WARNINGS := 2>&1 | grep -v "failed to insert symbol" | grep -v "could not find object file symbol for symbol" || true
 else
-  SILENCE_WARNINGS :=
+	SILENCE_WARNINGS :=
 endif
 
 # unconditionally built by the default Make target
@@ -60,6 +60,7 @@ TOOLS := \
 	ncli_split_keystore \
 	wss_sim \
 	stack_sizes \
+	nimbus_light_client \
 	nimbus_validator_client \
 	nimbus_signing_node \
 	validator_db_aggregator
@@ -267,12 +268,12 @@ clean_eth2_network_simulation_all:
 	rm -rf tests/simulation/{data,validators}
 
 GOERLI_TESTNETS_PARAMS := \
-  --tcp-port=$$(( $(BASE_PORT) + $(NODE_ID) )) \
-  --udp-port=$$(( $(BASE_PORT) + $(NODE_ID) )) \
-  --metrics \
-  --metrics-port=$$(( $(BASE_METRICS_PORT) + $(NODE_ID) )) \
-  --rest \
-  --rest-port=$$(( $(BASE_REST_PORT) +$(NODE_ID) ))
+	--tcp-port=$$(( $(BASE_PORT) + $(NODE_ID) )) \
+	--udp-port=$$(( $(BASE_PORT) + $(NODE_ID) )) \
+	--metrics \
+	--metrics-port=$$(( $(BASE_METRICS_PORT) + $(NODE_ID) )) \
+	--rest \
+	--rest-port=$$(( $(BASE_REST_PORT) +$(NODE_ID) ))
 
 eth2_network_simulation: | build deps clean_eth2_network_simulation_all
 	+ GIT_ROOT="$$PWD" NIMFLAGS="$(NIMFLAGS)" LOG_LEVEL="$(LOG_LEVEL)" tests/simulation/start-in-tmux.sh
@@ -281,7 +282,7 @@ eth2_network_simulation: | build deps clean_eth2_network_simulation_all
 #- https://www.gnu.org/software/make/manual/html_node/Multi_002dLine.html
 #- macOS doesn't support "=" at the end of "define FOO": https://stackoverflow.com/questions/13260396/gnu-make-3-81-eval-function-not-working
 define CONNECT_TO_NETWORK
-  scripts/makedir.sh build/data/shared_$(1)_$(NODE_ID)
+	scripts/makedir.sh build/data/shared_$(1)_$(NODE_ID)
 
 	scripts/make_prometheus_config.sh \
 		--nodes 1 \
@@ -297,7 +298,7 @@ define CONNECT_TO_NETWORK
 endef
 
 define CONNECT_TO_NETWORK_IN_DEV_MODE
-  scripts/makedir.sh build/data/shared_$(1)_$(NODE_ID)
+	scripts/makedir.sh build/data/shared_$(1)_$(NODE_ID)
 
 	scripts/make_prometheus_config.sh \
 		--nodes 1 \
@@ -338,6 +339,16 @@ define CONNECT_TO_NETWORK_WITH_VALIDATOR_CLIENT
 		--log-file=build/data/shared_$(1)_$(NODE_ID)/nbc_vc_$$(date +"%Y%m%d%H%M%S").log \
 		--data-dir=build/data/shared_$(1)_$(NODE_ID) \
 		--rest-port=$$(( $(BASE_REST_PORT) +$(NODE_ID) ))
+endef
+
+define CONNECT_TO_NETWORK_WITH_LIGHT_CLIENT
+	scripts/makedir.sh build/data/shared_$(1)_$(NODE_ID)
+
+	$(CPU_LIMIT_CMD) build/nimbus_light_client \
+		--network=$(1) \
+		--log-level="$(RUNTIME_LOG_LEVEL)" \
+		--log-file=build/data/shared_$(1)_$(NODE_ID)/nbc_lc_$$(date +"%Y%m%d%H%M%S").log \
+		--trusted-block-root="$(LC_TRUSTED_BLOCK_ROOT)"
 endef
 
 define MAKE_DEPOSIT_DATA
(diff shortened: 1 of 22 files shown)


scripts/launch_local_testnet.sh
```diff
@@ -696,14 +712,52 @@ for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
   fi
 done
 
+# light clients
+if [ "$LC_NODES" -ge "1" ]; then
+  echo "Waiting for Altair finalization"
+  ALTAIR_FORK_EPOCH="$(
+    curl -s "http://localhost:${BASE_REST_PORT}/eth/v1/config/fork_schedule" | \
+      jq -r '.data[1].epoch')"
```
tersec: This isn't necessarily a reliable way of extracting `ALTAIR_FORK_EPOCH`. See https://github.com/status-im/nimbus-eth2/pull/3649 for more discussion of this point.

Furthermore, even aside from that, there's no guarantee the BN sends the fork schedule in sorted order. The Nimbus VC sorts the results before using them.

Also, I think this adds a `jq` dependency that wasn't there, but that's fine, `jq` is something people should probably have installed anyway.

Makefile
```diff
@@ -412,6 +426,9 @@ ropsten: | ropsten-build
 ropsten-vc: | ropsten-build nimbus_validator_client
 	$(call CONNECT_TO_NETWORK_WITH_VALIDATOR_CLIENT,ropsten,nimbus_beacon_node,$(ROPSTEN_WEB3_URL))
 
+ropsten-lc: | nimbus_light_client
```
tersec, line 429: Probably makes sense to add the relevant `genesisData` to
https://github.com/status-im/nimbus-eth2/blob/f7eff8fcbe3c62047e71694508523dada3080184/beacon_chain/networking/network_metadata.nim#L167-L188
since it'll be another longer-lived testnet, it looks like, not as ephemeral as one of the Goerli or Mainnet shadow forks.

scripts/launch_local_testnet.sh
etan-status: How to know which fork version in the schedule belongs to what spec fork, if the result is not sorted? There could be multiple with same epoch, no? I guess could hardcode it to `1` as well, to match the config from above. For the `jq` dependency, will see how well the servers like it / if it is a practical problem.

etan-status: I guess one could download the full `config` to fetch `ALTAIR_FORK_EPOCH` / `ALTAIR_FORK_VERSION` and select based on that -- the schedule should only contain a single result with `current_version == ALTAIR_FORK_VERSION` I guess.

arnetheduck:
do we have preliminary measurements of the bandwidth requirements of this little gem?

beacon_chain/networking/eth2_network.nim
```diff
@@ -2143,16 +2192,16 @@ func gossipId(
...
+                      seckey: PrivateKey, address: MultiAddress,
                       rng: ref BrHmacDrbgContext): Switch {.raises: [Defect, CatchableError].} =
   SwitchBuilder
     .new()
     .withPrivateKey(seckey)
     .withAddress(address)
     .withRng(rng)
     .withNoise()
-    .withMplex(5.minutes, 5.minutes)
+    .withMplex(chronos.minutes(5), chronos.minutes(5))
```
tersec, line 2204: Why the changes from `n.minutes` and `n.seconds` to `chronos.minutes(n)` and `chronos.seconds(n)`?

Makefile
tersec: https://github.com/eth-clients/merge-testnets/tree/main/ropsten-beacon-chain has the relevant information.

scripts/launch_local_testnet.sh
tersec: Currently, the fork schedule is guaranteed to contain `ALTAIR_FORK_EPOCH`, as either `current_fork_version` or `previous_fork_version`, but that's a quirk of having only three forks deployed. Because `current_fork_version` is either going to correspond to genesis (pre-Altair, irrelevant), Altair (the goal), or Bellatrix.

It is possible to have a fork schedule doesn't have Altair as any `current_fork_version`, though -- per linked PR, `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH` produces this outcome.

One way to simplify this, maybe, is reframe it -- one cares really about the first fork transition of any type, because the only non-relevant fork is genesis/phase0. So maybe an alternate or more robust approach is to detect the first transition epoch away from that, rather than caring whether it's altair per se.

This still requires sorting, but fewer other assumptions.

etan-status: Changed to use `spec` endpoint instead.

beacon_chain/networking/eth2_network.nim
etan-status: The additional imports/exports led to ambiguity for the time units - this already happened a lot in other parts of the code where it was resolved similarly.

etan-status:
> do we have preliminary measurements of the bandwidth requirements of this little gem?

Snappy-encoded payloads:
- OptimisticUpdate ~224 bytes (on each slot)
- FinalityUpdate ~529 bytes (on finality change)
- Full update ~25K (at least once per sync committee period)

Libp2p overhead not measured though, scales with number of peers.
Also, depending on use case, can fetch latest update via Req/Resp as well instead of following Gossip.

Makefile
etan-status: https://github.com/status-im/nimbus-eth2/pull/3659

arnetheduck:
```
>>> (8192*224 + 8192/32*529 + 25*1024)/(8192*12)
20.3046875
```

~20 bytes / second, not bad.

arnetheduck:
ok, next step here: add web3 flag, so that if it's present, we call forkchoiceupdated (or whatever it is that drives the EL) - cc @tersec

tersec:
The EL wants a `newPayload(executionPayload)` / `forkchoiceUpdated(executionPayloadHash, ...)` sequence. One needs to send both, in general, for maximum compatibility. If one sends just one of those, what different ELs do will differ a bit more.

arnetheduck:
> If one sends just one of those, what different ELs do will differ a bit more.

is there any that can live on a fcu diet alone?

tersec:
> is there any that can live on a fcu diet alone?

I don't believe so, no. In principle, they mostly just want a payload/block header to sync to, from what I can tell, and `newPayload` gives them more than they strictly need, but there's no middle ground the engine API provides between just the hash (fcU) and a full `ExecutionPayload` (`newPayload`).

arnetheduck:
well, still doable given that we can download blocks by hash - a bit more messy but ..

Makefile
```diff
@@ -340,6 +341,16 @@ define CONNECT_TO_NETWORK_WITH_VALIDATOR_CLIENT
...
 endef
 
+define CONNECT_TO_NETWORK_WITH_LIGHT_CLIENT
+	scripts/makedir.sh build/data/shared_$(1)_$(NODE_ID)
+
+	$(CPU_LIMIT_CMD) build/nimbus_light_client \
+		--network=$(1) \
+		--log-level="$(RUNTIME_LOG_LEVEL)" \
+		--log-file=build/data/shared_$(1)_$(NODE_ID)/nbc_lc_$$(date +"%Y%m%d%H%M%S").log \
+		--trusted-block-root="$(LC_TRUSTED_BLOCK_ROOT)"
```
zah, line 351: How come the light client doesn't require a data-dir (even though you are creating one above)? Doesn't it persist its syncing progress upon restarts?

etan-status: No, there is no persistence (but after discovery, the sync is also done in seconds.
The created directory is for the log file.

beacon_chain/networking/eth2_network.nim
```diff
@@ -1669,12 +1708,15 @@ proc new*(T: type Eth2Node, config: BeaconNodeConf, runtimeCfg: RuntimeConfig,
       node.protocolStates[proto.index] = proto.networkStateInitializer(node)
 
     for msg in proto.messages:
-      if msg.isLightClientRequest and not config.serveLightClientData.get:
-        continue
+      when config is BeaconNodeConf:
+        if msg.isLightClientRequest and not config.serveLightClientData.get:
+          continue
+      when config is LightClientConf:
```
tersec: Can both of `config is BeaconNodeConf` and `config is LightClientConf` ever be true? If not, it might be useful to clarify that with `when: ... / else: ...`, and if so, document/comment as such.

scripts/launch_local_testnet.sh
```diff
@@ -696,14 +712,52 @@ for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
...
+      --log-level="${LOG_LEVEL}" \
+      --log-format="json" \
+      --network="${CONTAINER_DATA_DIR}" \
+      --bootstrap-node="${LC_BOOTSTRAP_NODE}" \
+      --tcp-port=$(( BASE_PORT + NUM_NODES + NUM_LC )) \
+      --udp-port=$(( BASE_PORT + NUM_NODES + NUM_LC )) \
+      --max-peers=$(( NUM_NODES + LC_NODES - 1 )) \
+      --nat="extip:127.0.0.1" \
+      --trusted-block-root="${LC_TRUSTED_BLOCK_ROOT}" \
+      ${STOP_AT_EPOCH_FLAG} \
```
tersec, line 752: Even though `STOP_AT_EPOCH_FLAG` can't contain spaces or similar characters, it's useful to quote this so one can statically verify that.

etan-status: That would make it a separate empty-string argument when it's absent.

tersec: Oh, indeed.

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3653/commits/63426cbb83c2696f7465e9a7d25bacd0701e1ea3

scripts/launch_local_testnet.sh
```diff
@@ -696,14 +712,52 @@ for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
   fi
 done
 
+# light clients
+if [ "$LC_NODES" -ge "1" ]; then
+  echo "Waiting for Altair finalization"
+  ALTAIR_FORK_EPOCH="$(
+    curl -s "http://localhost:${BASE_REST_PORT}/eth/v1/config/spec" | \
```
tersec: This introduces both `curl` and `jq` dependencies into this script (e.g., the CL test vector downloading uses it, but before this, I don't think running the local test network would invoke `curl`) and while if `jq` doesn't exist, it should signal en error that the `set -e` should pick up, the `curl` won't if it's missing, because it's `curl | jq` without `set -o pipefail` (or `set -eo pipefail`, since `set -e` already exists).

Other pipe usages in this script, but some of them are, e.g., `jobs`, `ls`, or `wc -l`, which are more reliably present and less likely to fail. I'm a bit warier of a silent, or quiet, failure mode here of especially `curl` being missing (or in theory, failing for some reason).

etan-status: `curl` was already used on line 311, but I've added some additional guards

arnetheduck:
btw, I talked about this with @karalabe in AMS - it's not impossible to make it work without blocks - though it might actually make sense that blocks are worked on the CL side given how this matches the "normal" flow - anyway, with fluffy, it becomes a different story.

tersec (approved):

PR #3656, opened 2022-05-23
remove deprecated JSON-RPC server

tersec:
https://github.com/status-im/nimbus-eth2/releases/tag/v22.3.0
> The [JSON-RPC](https://nimbus.guide/api.html) service (`--rpc` flag) option is now deprecated. It's scheduled for removal in version `v22.6` (i.e. June of this year). If you are currently relying on the JSON-RPC API, please consider switching to the official [REST API](https://nimbus.guide/rest-api.html).


beacon_chain/conf.nim
```diff
@@ -337,23 +337,6 @@ type
         defaultValueDesc: ""
         name: "status-bar-contents" }: string
 
-      rpcEnabled* {.
```
arnetheduck, line 340: probably better to keep the options (hidden) and log a warning - ie I've seen these options in copy-paste guides where it's doubtful the service is actually used, no point breaking those user setups just to make a point.

tersec: https://github.com/status-im/nimbus-eth2/pull/3656/commits/25013b1ff7b56b8b54172059da09002af3e952b1

arnetheduck: very nice, much better UX

arnetheduck (approved):

arnetheduck:
The client in https://github.com/status-im/nimbus-eth2/tree/mitem/beacon_chain/spec/eth2_apis can also be removed - likewise validator exit support etc - this would also need corresponding updates to the manual.

tersec:
https://github.com/status-im/nimbus-eth2/pull/3676

PR #3665, opened 2022-05-25
Switch back to nim-websock

Menduist:
The livelock should be fixed, might be worth another shot


arnetheduck:
how can this be tested properly? ie some form of stress testing regime?

Menduist:
I've been fuzzing it continuously for a week, so I'm pretty confident there are no deadlocks / defects left, at least
And it still passes Autobanh, which is a pretty big test suite

We could also let Waku battle test it until we're more confident in it :) But at this stage, we mostly have to jump the gap on one of the project and see how it goes

tersec:
For the merge branch, it'd be nice to wait a few days on this, because until https://github.com/status-im/nimbus-eth2/pull/3669 was merged, post-merge web3 URLs using `http` were basically unusable, which meant requiring websocket. I want to be sure that the is truly working (so far, so good, but...) before removing the fallback if this still doesn't work, because that'd leave the post-merge setup with potentially nothing.

zah:
It seems wiser to start using this on some of our fleet nodes without changing the default compilation mode.

tersec:
I'm fine with this now.

arnetheduck:
I think we can merge this now that we have some time to the next release

Menduist:
Is there something in specific we should monitor on the testfleet to check everything is going well after merging?

arnetheduck:
> Is there something in specific we should monitor on the testfleet to check everything is going well after merging?

the thing to verify (with @jakubgs) is that we have a healthy mix of nodes in our testnets that run both ws and http

jakubgs:
We currently don't use RPC/http at all:
https://github.com/status-im/infra-nimbus/blob/5aa3b64a/ansible/group_vars/nimbus.mainnet.yml#L29-L30
https://github.com/status-im/infra-nimbus/blob/5aa3b64a/ansible/group_vars/nimbus.prater.yml#L41-L42
https://github.com/status-im/infra-nimbus/blob/5aa3b64a/ansible/group_vars/nimbus.ropsten.yml#L11-L12
What would be the optimal setup? Random? Semi random? Some deterministic spread where it depends on branch?

arnetheduck:
>  Some deterministic spread where it depends on branch?

probably this - cc @tersec @zah - this becomes more important with the merge where the web3 connection becomes a lot more important

arnetheduck:
but since all are using WS, that's a good thing for now, means this will get a strong stress test

arnetheduck:
>  Some deterministic spread where it depends on branch?

basically, we're getting to the point where we have a whole bunch of settings that need to be tested in various constellations - I wonder how we can do it such that we have a good spread of nodes on each setting - http vs ws, subscribe-all vs not, stuff like this

tersec:
To summarize some relevant dimensions

1. `http://` / `ws://` URLs
2. within `ws://`, which websocket library is in use, `nim-websock` or otherwise [this should be temporary, so listing separately]
3. `--web3-force-polling` enabled or disabled [it's disabled by default, so that should be most such nodes]
4. Nim 1.2 / Nim 1.6 nodes [semi-temporary split, will probably resolve to pure-Nim 1.6 before too long]

Each of these, ideally, is independently choosable. The websocket library doesn't have any bearing for `http://` Web3 URLs, but that's fine in terms of parameterizing node and validator allocation, it's just a no-op.

So the general framework I'd prefer would be to independently pick ratios for each A/B pair in items (1) through (4), and then calculate from that which validators and which nodes get which combinations of those, in a basically just multiplicative manner.

Where there are irrelevant variables, one might end up, yes, with use `http://` and don't use `nim-websocket`, and that Nimbus can even be built to use or not use `nim-websocket` as specified, it just shouldn't matter.

jakubgs:
Please open an issue in `infra-nimbus` if you want changes in the config.

PR #3672, opened 2022-05-30
Bump chronos to fix httpclient assert issue (stable target branch).

tersec:
Exactly the same content as https://github.com/status-im/nimbus-eth2/pull/3669 except for `stable`, rather than `unstable`.


tersec:
Not going to use this.

PR #3706, opened 2022-06-05
Add the starting geth nodes as part of local testnet

hmel:
`Adds` --run-geth parameter to launch_local_testnet.sh script, which launches a number of geth nodes and connects them to each other before starting the nimbus-eth2 client.
Added missing --light-clients options to parsed parameters.


scripts/genesis.json
```diff
@@ -0,0 +1,44 @@
+{
```
tersec: While the https://github.com/eth-clients/merge-testnets/ repos do use a naming strategy of `genesis.json`, `besu_genesis.json`, and `nethermind_genesis.json` (e.g., https://github.com/eth-clients/merge-testnets/blob/main/mainnet-shadow-fork-6/genesis.json, https://github.com/eth-clients/merge-testnets/blob/main/mainnet-shadow-fork-6/besu_genesis.json, and https://github.com/eth-clients/merge-testnets/blob/main/mainnet-shadow-fork-6/nethermind_genesis.json), it seems potentially better not to privilege Geth implicitly this way, but to provide some sort of namespacing (within a filename, with directories, etc) which treats different EL clients structurally equivalently.

Similarly, there have been other uses for these EL genesis files in the `nimbus-eth2` repo, such as for joining some merge testnet such as Kiln or Ropsten. `genesis.json` is a quite broad descriptor.

scripts/launch_local_testnet.sh
```diff
@@ -272,6 +293,11 @@ if [[ "${OS}" != "windows" ]]; then
   which lsof &>/dev/null || \
     { echo "'lsof' not installed and we need it to check for ports already in use. Aborting."; exit 1; }
 
+  #TODO Stop geth nodes
+  #for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
+  #for PORT in $(( BASE_GETH_PORT + NUM_NODE )) $(( BASE_METRICS_PORT + NUM_NODE )) $(( BASE_REST_PORT + NUM_NODE )); do
+  #done
```
tersec: How does Geth respond to running to this script being run multiple times successively? One way I sometime test the stability of `nimbus-eth2`'s local testnet is by running some configuration in a loop, but that relies on its cleaning up after itself.

scripts/start_geth_nodes.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
tersec, line 1: For these test scripts which don't have to hide failure, my preference is that they shouldn't, along the lines of `set -eu` (or one of its variations (`set -euo pipefail`, etc). Sometimes this disallows genuinely useful tricks, and loses too much expressive power, but it's also good for catching, in this kind of setup, if there's some change which misaligns variable names or similar.

scripts/genesis.json
hmel: I was considering making the genesis file a command line argument and falling back on a generic name. Would that be ok?

tersec: Sure.

scripts/launch_local_testnet.sh
hmel: Geth gets killed on exit by cleanup() function like the other processes started from the script. I'm not sure how useful the --kill-old-processes argument is, which kills processes based on port number, so that's why I left that out. Do you know of any valid use cases for it? I'd actually suggest removing it altogether.

scripts/start_geth_nodes.sh
hmel: Added

scripts/genesis.json
hmel: Renamed to geth_genesis.json. Will consider adding command line param when I add nimbus-eth1 as alternate EL

scripts/geth_genesis.json
```diff
@@ -0,0 +1,44 @@
+{
+	  "config": {
+		    "chainId":1,
```
zah: `chainId == 1` is mainnet. Shouldn't we use a different value in a local simulation?

scripts/launch_local_testnet.sh
```diff
@@ -72,6 +72,9 @@ REMOTE_SIGNER_NODES=0
 REMOTE_SIGNER_THRESHOLD=1
 REMOTE_VALIDATORS_COUNT=0
 LC_NODES=1
+ACCOUNT_PASSWORD="nimbus"
+RUN_GETH="0"
+GETH_BINARY="${HOME}/go-ethereum/build/bin/geth"
```
zah: This scripts needs to be able to run in our CI. It needs to take care of downloading the Geth binary if it's missing. The current OS and CPU architectures must be taken into account. Using a well-known docker image is another option.

scripts/launch_local_testnet.sh
tersec: For a while, there were intermittent CI errors where some previously-running process caused port in use issues and this was a brute-force way to stop them. Because the Jenkins CIs don't necessarily run in separate, isolated Docker containers, it's possible for them to just leave behind processes this way. It's not critical but it's a nice safeguard to have in place.

scripts/geth_genesis.json
hmel: changed

scripts/launch_local_testnet.sh
hmel: Added --dl-geth option

scripts/launch_local_testnet.sh
hmel: Added

scripts/geth_genesis.json
```diff
@@ -0,0 +1,44 @@
...
+	  },
+	  "nonce":"0x42",
+	  "timestamp":"0x0",
+	  "extraData":"0x0000000000000000000000000000000000000000000000000000000000000000a94f5374fce5edbc8e2a8697c15331677e6ebf0b0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+	  "gasLimit":"0x1C9C380",
+	  "difficulty":"0x400000000",
+	  "mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000",
+	  "coinbase":"0x0000000000000000000000000000000000000000",
+	  "alloc":{
+		    "0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b":{"balance":"0x6d6172697573766477000000"},
```
zah, line 29: It would be nice if the private keys for these accounts are provided as well in the repo, so the developers can execute transactions with them.

scripts/nimbusel_genesis.json
```diff
@@ -0,0 +1,44 @@
+{
```
zah, line 1: The name of this file is somewhat inconsistent - shouldn't it be `nimbus_el_genesis.json`? Also, do we actually need this file, considering that it's exactly the same as the `geth_genesis.json` file?

scripts/launch_local_testnet.sh
```diff
@@ -267,18 +314,71 @@ else
...
+            exit 1
+          fi
+        done
+      done
+    done
+  fi
+
+  if [[ "${RUN_NIMBUS}" == "1" ]]; then
+    for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
+      for PORT in $(( NUM_NODE * NIMBUSEL_PORT_OFFSET + NIMBUSEL_NET_BASE_PORT )) $(( NUM_NODE * NIMBUSEL_PORT_OFFSET + NIMBUSEL_HTTP_BASE_PORT )) \
```
zah, line 359: The duplicated block of code here looks like a function that kills all processing listening on ports within a given interval. You can define and reuse such a function for both Geth and Nimbus

scripts/nimbusel_genesis.json
hmel: It was called genesis.json, but @tersec suggested it's too vague, so I changed it. This way it's easier to make changes to one without affecting the other.

scripts/launch_local_testnet.sh
hmel: Yeah, I did some copy-pasta there and forgot to change it. Will fix

scripts/geth_genesis.json
hmel: Yes, for next phase with accounts+transactions I will add them

zah:
Rebased on top of unstable here:

https://github.com/status-im/nimbus-eth2/pull/3790

arnetheduck:
so this can be closed?

PR #3715, opened 2022-06-07
keep track of latest blocks for optimistic sync

etan-status:
When launched with `--light-client-enable` the latest blocks are fetched
and optimistic candidate blocks are passed to a callback (log for now).
This helps accelerate syncing in the future (optimistic sync).

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
@@ -71,7 +71,7 @@ func getBackfillRoot*(lcBlocks: LCBlocks): Option[Eth2Digest] =
   else:
     some lcBlocks.backfill.parent_root
 
-func getCacheIndex(lcBlocks: LCBlocks, slot: Slot): Natural =
+func getCacheIndex(lcBlocks: LCBlocks, slot: Slot): uint64 =
   if slot < lcBlocks.headSlot and lcBlocks.headSlot != FAR_FUTURE_SLOT:
     lcBlocks.headSlot - slot
   else:
@@ -83,7 +83,7 @@ func getBlockAtSlot*(
     return err()
 
   let index = lcBlocks.getCacheIndex(slot)
-  if index >= lcBlocks.cache.len:
+  if index >= lcBlocks.cache.lenu64:
     return err()
   let existing = lcBlocks.cache[index]
   if existing == nil:
@@ -96,7 +96,7 @@ func getLatestBlockThroughSlot*(
     return err()
 
   let startIndex = lcBlocks.getCacheIndex(maxSlot)
-  for i in startIndex ..< lcBlocks.cache.len:
+  for i in startIndex ..< lcBlocks.cache.lenu64:
     let blck = lcBlocks.cache[i]
     if blck != nil:
       return ok blck[]
@@ -132,8 +132,8 @@ proc processBlock(
         debug "Head LC block from unviable fork"
       return err(BlockError.UnviableFork)
 
-    const index = 0 # Head block is always mapped to index 0, it cannot be empty
-    if index >= lcBlocks.cache.len:
+    const index = 0'u64 # Head block is always mapped to index 0 (never empty)
+    if index >= lcBlocks.cache.lenu64:
       lcBlocks.backfill.slot = blck.slot
       debug "Final head LC block"
       return ok()
@@ -155,7 +155,7 @@ proc processBlock(
   # Handle duplicate block
   if blck.slot >= lcBlocks.getBackfillSlot():
     let index = lcBlocks.getCacheIndex(blck.slot)
-    doAssert index < lcBlocks.cache.len
+    doAssert index < lcBlocks.cache.lenu64
     let existing = lcBlocks.cache[index]
     if existing == nil:
       debug "Duplicate LC block for empty slot"
@@ -189,13 +189,13 @@ proc processBlock(
   let
     previousIndex = lcBlocks.getCacheIndex(lcBlocks.backfill.slot)
     index = lcBlocks.getCacheIndex(blck.slot)
-  for i in previousIndex + 1 ..< min(index, lcBlocks.cache.len):
+  for i in previousIndex + 1 ..< min(index, lcBlocks.cache.lenu64):
     let existing = lcBlocks.cache[i]
     if existing != nil:
       warn "LC block reorg to empty", existing = existing[]
       lcBlocks.cache[i] = nil
 
-  if index >= lcBlocks.cache.len:
+  if index >= lcBlocks.cache.lenu64:
     lcBlocks.backfill.slot = blck.slot
     debug "Final LC block"
     return ok()
@@ -265,7 +265,7 @@ proc setFinalizedBid*(lcBlocks: var LCBlocks, finalizedBid: BlockId) =
   if finalizedBid.slot <= lcBlocks.headSlot and
       finalizedBid.slot >= lcBlocks.getBackfillSlot:
     let index = lcBlocks.getCacheIndex(finalizedBid.slot)
-    doAssert index < lcBlocks.cache.len
+    doAssert index < lcBlocks.cache.lenu64
     let existing = lcBlocks.cache[index]
     if existing == nil or finalizedBid.root != existing[].root:
       if existing != nil:
@@ -293,7 +293,8 @@ proc addBlock*(
       lcBlocks.setFinalizedBid(lcBlocks.finalizedBid)
       return err(BlockError.UnviableFork)
 
-  for i in lcBlocks.getCacheIndex(signedBlock.slot) + 1 ..< lcBlocks.cache.len:
+  let slot = signedBlock.slot
+  for i in lcBlocks.getCacheIndex(slot) + 1 ..< lcBlocks.cache.lenu64:
     let existing = lcBlocks.cache[i]
     if existing != nil:
       let res =
--- a/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
@@ -219,7 +219,9 @@ proc setHeadBid*(lcBlocks: var LCBlocks, headBid: BlockId) =
   debug "New LC head block", headBid
   if lcBlocks.maxSlots == 0:
     discard
-  elif headBid.slot >= lcBlocks.headSlot + lcBlocks.maxSlots.uint64 or (
+
+  elif lcBlocks.headSlot == FAR_FUTURE_SLOT or
+      headBid.slot >= lcBlocks.headSlot + lcBlocks.maxSlots.uint64 or (
       lcBlocks.headSlot - lcBlocks.cache.lenu64 != FAR_FUTURE_SLOT and
       headBid.slot <= lcBlocks.headSlot - lcBlocks.cache.lenu64):
     lcBlocks.cache.clear()
--- a/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
@@ -132,12 +132,11 @@ proc processBlock(
         debug "Head LC block from unviable fork"
       return err(BlockError.UnviableFork)
 
-    let index = lcBlocks.getCacheIndex(blck.slot)
+    let index = 0 # Head slot is always mapped to index 0, it cannot be empty
     if index >= lcBlocks.cache.len:
       lcBlocks.backfill.slot = blck.slot
       debug "Final head LC block"
       return ok()
-    doAssert index == 0
 
     lcBlocks.backfill = blck.toBeaconBlockSummary()
     let existing = lcBlocks.cache[index]
--- a/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance_light_client.nim
@@ -1,0 +1,304 @@
+# beacon_chain
+# Copyright (c) 2019-2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at http://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at http://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
+
+import
+  std/[deques, math],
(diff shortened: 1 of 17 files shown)


beacon_chain/sync/optimistic_sync_light_client.nim
```diff
@@ -0,0 +1,292 @@
...
+      syncDistance == 0:
+    return
+
+  # Cancel ongoing sync if sync target jumped
+  if headDiff >= SLOTS_PER_EPOCH and optSync.syncFut != nil:
+    if not optSync.syncFut.finished:
+      optSync.syncFut.cancel()
+    return
+
+  # When retargetting ongoing sync is not possible, cancel on finality change
```
tersec:
```suggestion
  # When retargeting ongoing sync is not possible, cancel on finality change
```

beacon_chain/spec/helpers.nim
```diff
@@ -597,14 +597,19 @@ func is_merge_transition_complete*(state: bellatrix.BeaconState): bool =
...
+
 # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#is_merge_transition_block
 func is_merge_transition_block(
     state: bellatrix.BeaconState,
     body: bellatrix.BeaconBlockBody | bellatrix.TrustedBeaconBlockBody |
           bellatrix.SigVerifiedBeaconBlockBody): bool =
-  const defaultBellatrixExecutionPayload = default(bellatrix.ExecutionPayload)
-  not is_merge_transition_complete(state) and
-    body.execution_payload != defaultBellatrixExecutionPayload
+  not is_merge_transition_complete(state) and body.is_execution_block()
```
tersec: In cases such as these where there's no algorithmic reason to rewrite spec functions, I'd prefer that they follow exactly what the spec states, with as little deviation as possible -- this makes updating them, should they change, easier. There's a bit of duplication (Though, it's less painful with `static(default(...))`), but aligns exactly with:
```python
def is_merge_transition_block(state: BeaconState, body: BeaconBlockBody) -> bool:
    return not is_merge_transition_complete(state) and body.execution_payload != ExecutionPayload()
```
from https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#is_merge_transition_block

I'd suggest that this is a useful revision/refactoring of the spec, and even a helpful one in terms of the spec being partly an explanatory document, partly a runnable implementation of a state transition function, but would prefer that until the spec has been thusly refactored, neither is this.

Of course, many parts of the spec are not quite so copied almost token-for-token, because the spec code is inefficient in some algorithmic compute-time or memory usage way, but where that's not the case, such as here, it's deliberately close to the spec-as-written.

This particular change I don't feel that strongly about, but there are endless such minor/trivial refactorings which would add up to the Nimbus code simply not aligning with the spec code, and I've chosen to be relatively purist about it for that reason.

beacon_chain/sync/sync_manager.nim
```diff
@@ -12,7 +12,7 @@ import stew/[results, base10], chronos, chronicles
 import
   ../spec/datatypes/[phase0, altair],
   ../spec/eth2_apis/rest_types,
-  ../spec/[helpers, forks],
+  ../spec/[helpers, forks, network],
```
tersec, line 15: No changes except an additional `import`?

beacon_chain/sync/optimistic_sync_light_client.nim
```diff
@@ -0,0 +1,292 @@
...
+  optSync
+
+proc start*(optSync: LCOptimisticSync) =
+  optSync.requestManager.start()
+
+func supportsRetarget(syncStrategy: SyncStrategy): bool =
+  case syncStrategy
+  of SyncStrategy.None:
+    true
+  of SyncStrategy.RequestManager:
```
tersec: The case expression is useful to preserve compile-time exhaustiveness checking, so the other obvious alternatives such as just `syncStrategy in [SyncStrategy.None, SyncStrategy.RequestManager]`, but this one can still replace the two identical `of` branches with `of SyncStrategy.None, SyncStrategy.RequestManager:` to reduce duplication somewhat.

beacon_chain/sync/optimistic_sync_light_client.nim
```diff
@@ -0,0 +1,292 @@
...
+
+  proc getLocalWallSlot(): Slot =
+    optSync.getBeaconTime().slotOrZero
+
+  var cancellationRequested = false
+  func getProgressSlot(): Slot =
+    if not cancellationRequested:
+      optSync.lcBlocks.getBackfillSlot()
+    else:
+      # Report out-of-bands completion of sync
```
tersec:
```suggestion
      # Report out-of-band completion of sync
```

beacon_chain/spec/helpers.nim
```diff
@@ -597,14 +597,19 @@ func is_merge_transition_complete*(state: bellatrix.BeaconState): bool =
   const defaultExecutionPayloadHeader = default(ExecutionPayloadHeader)
   state.latest_execution_payload_header != defaultExecutionPayloadHeader
 
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md
```
tersec:
```suggestion
# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#helpers
```

beacon_chain/sync/sync_manager.nim
etan-status: Yes, it is needed for `RESP_TIMEOUT`. So far it only worked by accident because all modules importing `sync_manager` also import `network` before this, but `optimistic_sync_light_client` does not, so the missing `network` import is now caught.

beacon_chain/sync/optimistic_sync_light_client.nim
tersec: resolved via https://github.com/status-im/nimbus-eth2/pull/3715/commits/a7eacef2a8a1b64f0723306ea5be93be505a40e9

beacon_chain/spec/helpers.nim
tersec: resolved via https://github.com/status-im/nimbus-eth2/pull/3715/commits/30651693f27f6aad7b90363cc11a7eb4b7478681

beacon_chain/sync/optimistic_sync_light_client.nim
```diff
@@ -0,0 +1,290 @@
...
+    network: Eth2Node,
+    getBeaconTime: GetBeaconTimeFn,
+    optimisticProcessor: MsgTrustedBlockProcessor,
+    safeSlotsToImportOptimistically: uint64): LCOptimisticSync =
+  const
+    numExtraSlots = 2 * SLOTS_PER_EPOCH + 1
+    supportedSafeSlots = (int.high - numExtraSlots.int).uint64
+  let
+    safeSlots = min(safeSlotsToImportOptimistically, supportedSafeSlots)
+    maxSlots = (safeSlots + numExtraSlots).int
```
tersec: Ideally, one could use something like `Natural` here to ensure that this `int` conversion is safe, but that has enough problems it's not a great idea.

The root cause of these type conversions (and the need elsewhere in this PR to track `safeSlotsToImportOptimistically`-related quantities as `uint64`) is
https://github.com/status-im/nimbus-eth2/blob/1dec3ff8b67af06456251c2782114fa31ec5a968/beacon_chain/conf.nim#L514-L520

Probably that's just excessive -- at most, that config option should be type `uint32`, or even better, `uint16` or `int16` (if someone complains later, well, increasing the range from 2^15-1 slots could be done later). That way it's guaranteed safe to do these other operations on it, rather than this quantity which is theoretically could be a trillion so other code has to anticipate.

The goal here is to remove as many potentially unsafe integer type conversions as feasible, and this one is definitely feasible to have 0 potentially unsafe integer runtime conversions.

tersec: resolved via https://github.com/status-im/nimbus-eth2/pull/3715/commits/15bc5858e92dbbcca0a5819a2ba4fee5c7e649cc

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
```diff
@@ -0,0 +1,306 @@
...
+      if isNewBlock:
+        debug "Head LC block from unviable fork"
+      return err(BlockError.UnviableFork)
+
+    let index = lcBlocks.getCacheIndex(blck.slot)
+    if index >= lcBlocks.cache.len:
+      lcBlocks.backfill.slot = blck.slot
+      debug "Final head LC block"
+      return ok()
+    doAssert index == 0
```
tersec: Either `index` is out of bounds of the cache or it's 0?

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
```diff
@@ -0,0 +1,306 @@
...
+
+    lcBlocks.backfill = blck.toBeaconBlockSummary()
+    let existing = lcBlocks.cache[index]
```
tersec, line 142: `doAssert index == 0`?

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
```diff
@@ -0,0 +1,306 @@
...
+    if existing != nil:
+      if blockRoot == existing[].root:
+        if isNewBlock:
+          debug "Head LC block already known"
+        return ok()
+      warn "Head LC block reorg", existing = existing[]
+    lcBlocks.cache[index] =
```
tersec, line 149: `doAssert index == 0`?

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
etan-status: For the head block, yes. The light client can only be targeted at slots with a block, it does not have a concept of empty slots. Hence, the head block will always end up at index 0.

In edge case when `cache.len` is 0, then `index >= cache.len`, and the head block is also the final block being accepted (one past frontfill).

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
etan-status: Yes, `blck.slot > headSlot` --> `Duplicate`.
`blck.slot < headSlot` --> `MissingParent`.
So, `blck.slot == headSlot` to end up here, which will be mapped to `index == 0`

etan-status: Or, is the question suggesting to directly do `let index = 0` instead of `lcBlocks.getCacheIndex`? Could do that, sure.

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
etan-status: made `index` a `const` for head block

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
tersec: Yeah, what you did was what I was getting it -- it's effectively constant, and it's both more efficient (maybe, if Nim doesn't infer its value from the `doAssert`) and clearer if it's treated that way.

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
```diff
@@ -0,0 +1,305 @@
...
+
+func getBackfillRoot*(lcBlocks: LCBlocks): Option[Eth2Digest] =
+  if lcBlocks.headSlot == FAR_FUTURE_SLOT:
+    none(Eth2Digest)
+  elif lcBlocks.backfill.slot < lcBlocks.getFrontfillSlot():
+    none(Eth2Digest)
+  else:
+    some lcBlocks.backfill.parent_root
+
+func getCacheIndex(lcBlocks: LCBlocks, slot: Slot): Natural =
```
tersec: Probably better to avoid `Natural`, since it has unfortunate behavior regarding `Defects` even as specified and also is a range type: https://github.com/status-im/nim-style-guide/pull/9 / https://github.com/status-im/nim-style-guide/pull/9/files#diff-441a1302931c43b60ccb9ec77aad891f04e49f10dac4b45f6b574396bff1f51e

beacon_chain/spec/datatypes/phase0.nim
```diff
@@ -218,15 +218,31 @@ type
...
   TrustedSignedBeaconBlock* = object
     message*: TrustedBeaconBlock
     signature*: TrustedSig
 
     root* {.dontSerialize.}: Eth2Digest # cached root of signed beacon block
 
-  SomeSignedBeaconBlock* = SignedBeaconBlock | SigVerifiedSignedBeaconBlock | TrustedSignedBeaconBlock
-  SomeBeaconBlock* = BeaconBlock | SigVerifiedBeaconBlock | TrustedBeaconBlock
-  SomeBeaconBlockBody* = BeaconBlockBody | SigVerifiedBeaconBlockBody | TrustedBeaconBlockBody
+  SomeSignedBeaconBlock* =
```
tersec, line 233: These `Trusted` types seem to be for LC sync only, and in particular for
```nim
    optimisticProcessor = proc(signedBlock: ForkedMsgTrustedSignedBeaconBlock):
        Future[void] {.async.} =
      debug "New LC optimistic block",
        opt = signedBlock.toBlockId(),
        dag = node.dag.head.bid,
        wallSlot = node.currentSlot
      return
    optSync = initLCOptimisticSync(
      node.network, getBeaconTime, optimisticProcessor,
      config.safeSlotsToImportOptimistically)
```
it seems like the light client can't even operate during phase 0, so I'm not sure why phase 0 would need these "trusted" block types defined.

etan-status: While the LC head must be post-Altair, syncing backwards from the LC head may end up in phase0 territory.
The outer block proposer signature requires a known beacon state that includes the corresponding validator pubkey, hence only "message" trust is inferred.

For regular checkpoint sync the backfill only touches blocks before the checkpoint state, so all validator pubkeys of earlier blocks are guaranteed to be known.

beacon_chain/consensus_object_pools/block_clearance_light_client.nim
tersec: resolved via https://github.com/status-im/nimbus-eth2/pull/3715/commits/02eb118996a64f11fc03c09fd277a45b8a942d7e

tersec (approved):

bhndt:
joss

PR #3719, opened 2022-06-08
preparations for persisting light client data

etan-status:
When switching the light client data in-memory cache to a database,
additional failure points are introduced due to possible inconsistencies
that are introduced into the DB out-of-band. These additional failure
points are handled as preparation to the move to persistence.
Also includes fixes related to theoretic edge cases in initial period,
and some minor cleanups, e.g., `asSigned` vs `isomorphicCast`.


etan-status:
Too many things mixed up here, will submit as separate PRs.

PR #3720, opened 2022-06-08
Fix /eth/v1/node/syncing

cheatfate:
Fix REST server `/eth/v1/node/syncing` call to return proper values even if SyncManager is not running.

### DIFF
--- a/beacon_chain/rpc/rest_node_api.nim
+++ b/beacon_chain/rpc/rest_node_api.nim
@@ -1,5 +1,4 @@
 import
-  std/[sequtils],
   stew/[byteutils, results],
   chronicles,
   eth/p2p/discoveryv5/enr,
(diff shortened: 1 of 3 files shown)


beacon_chain/rpc/rest_node_api.nim
```diff
@@ -254,7 +253,15 @@ proc installNodeApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
   # https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus
   router.api(MethodGet, "/eth/v1/node/syncing") do () -> RestApiResponse:
-    return RestApiResponse.jsonResponse(node.syncManager.getInfo())
+    let
+      wallSlot = node.beaconClock.now().slotOrZero()
+      headSlot = node.dag.head.slot
+      distance = wallSlot - headSlot
+      info = RestSyncInfo(
+        head_slot: headSlot, sync_distance: distance,
+        is_syncing: distance != 0'u64
```
arnetheduck: this is too strict, ie it'll show up as syncing at the start of every slot, in the period between the start of the slot and until we've received a block. why is syncmanager itself not giving a correct value here?

cheatfate: Because there so many different type of sync, that syncmanager is not always running.

arnetheduck: This call should be reporting the progress of the ordinary forward sync that we use to chase head after the node goes offline

cheatfate: > Requests the beacon node to describe if it's currently syncing or not, and if it is, what block it is up to.

There is no mention on type of sync.
Also people are using this function to check progress of checkpoint sync.

zah: I also think the expected definition of "synced" is "able to attest for the current head" (in other words, how far is our current head from the wall clock head). For the other types of syncing, we probably need separate API calls.

etan-status (approved):

PR #3727, opened 2022-06-09
Fix the Keymanager API's delete URLs

zah:


arnetheduck (approved):

Makefile
```diff
@@ -147,6 +148,15 @@ update: | update-common
 libbacktrace:
 	+ "$(MAKE)" -C vendor/nim-libbacktrace --no-print-directory BUILD_CXX_LIB=0
 
+restapi-test:
+	./tests/simulation/restapi.sh \
+		--data-dir resttest0_data \
+		--base-port $$(( 9100 + EXECUTOR_NUMBER * 100 )) \
```
etan-status, line 154: does `EXECUTOR_NUMBER` make sense in the context of `Makefile`?

zah: It has a default value, but if you supply it in a ENV variable, it will have an effect. Since the Makefile target is used in the Jenkins CI, the variable is automatically specified by Jenkins there.

Issue #3269, opened 2022-01-11
"No peer for topic skipping publish" / "Attestation sent" inconsistency


mratsim:
From discord:

```
eth2_1        | INF 2022-01-08 11:20:47.000+00:00 Slot start                                 topics="beacnde" slot=2901402 epoch=90668 sync=synced peers=122 head=0fa49ad1:2901401 finalized=90666:6d42cb5e delay=771us948ns
eth2_1        | NOT 2022-01-08 11:20:50.151+00:00 No peers for topic, skipping publish       topics="libp2p gossipsub" peersOnTopic=0 connectedPeers=0 topic=/eth2/afcaaba0/beacon_attestation_45/ssz_snappy
eth2_1        | NOT 2022-01-08 11:20:50.151+00:00 Attestation sent                           topics="beacval" attestation="(aggregation_bits: 0b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000, data: (slot: 2901402, index: 45, beacon_block_root: \"c50ceb66\", source: \"90667:27813a1d\", target: \"90668:3524a785\"), signature: \"<REDACTED>\")" validator=<REDACTED> delay=-848ms601us579ns subnet_id=45
eth2_1        | NOT 2022-01-08 11:20:50.155+00:00 Attestation sent                           topics="beacval" attestation="(aggregation_bits: 0b0000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 2901402, index: 48, beacon_block_root: \"c50ceb66\", source: \"90667:27813a1d\", target: \"90668:3524a785\"), signature: \"<REDACTED>\")" validator=<REDACTED> delay=-844ms273us234ns subnet_id=48
eth2_1        | NOT 2022-01-08 11:20:55.008+00:00 Aggregated attestation sent                topics="beacval" attestation="(aggregation_bits: 0b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000, data: (slot: 2901402, index: 45, beacon_block_root: \"c50ceb66\", source: \"90667:27813a1d\", target: \"90668:3524a785\"), signature: \"<REDACTED>\")" aggregator_index=<REDACTED> signature=<REDACTED> validator=<REDACTED> subnet_id=45
[...]

PR #3737, opened 2022-06-12
check for and log gossip broadcast failure

tersec:
Fix https://github.com/status-im/nimbus-eth2/issues/3269

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -2439,6 +2439,7 @@ proc broadcast(node: Eth2Node, topic: string, msg: auto):
       inc nbc_gossip_messages_sent
       return ok()
     else:
+      # Increments libp2p_gossipsub_failed_publish metric
       return err("No peers on libp2p topic")
   except IOError as exc:
     raiseAssert exc.msg # TODO in-memory compression shouldn't fail
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -46,6 +46,7 @@ type
 
   Bytes = seq[byte]
   ErrorMsg = List[byte, 256]
+  SendResult* = Result[void, cstring]
 
   # TODO: This is here only to eradicate a compiler
   # warning about unused import (rpc/messages).
@@ -216,7 +217,7 @@ func phase0metadata*(node: Eth2Node): phase0.MetaData =
     seq_number: node.metadata.seq_number,
     attnets: node.metadata.attnets)
 
-func toAltairMetadata*(phase0: phase0.MetaData): altair.MetaData =
+func toAltairMetadata(phase0: phase0.MetaData): altair.MetaData =
   altair.MetaData(
     seq_number: phase0.seq_number,
     attnets: phase0.attnets)
@@ -315,9 +316,6 @@ func shortLog*(peer: Peer): string = shortLog(peer.peerId)
 chronicles.formatIt(Peer): shortLog(it)
 chronicles.formatIt(PublicKey): byteutils.toHex(it.getBytes().tryGet())
 
-template remote*(peer: Peer): untyped =
-  peer.peerId
-
 proc openStream(node: Eth2Node,
                 peer: Peer,
                 protocolId: string): Future[Connection] {.async.} =
@@ -331,7 +329,7 @@ proc openStream(node: Eth2Node,
 
   return conn
 
-proc init*(T: type Peer, network: Eth2Node, peerId: PeerId): Peer {.gcsafe.}
+proc init(T: type Peer, network: Eth2Node, peerId: PeerId): Peer {.gcsafe.}
 
 func peerId*(node: Eth2Node): PeerId =
   node.switch.peerInfo.peerId
@@ -339,7 +337,7 @@ func peerId*(node: Eth2Node): PeerId =
 func enrRecord*(node: Eth2Node): Record =
   node.discovery.localNode.record
 
-proc getPeer*(node: Eth2Node, peerId: PeerId): Peer =
+proc getPeer(node: Eth2Node, peerId: PeerId): Peer =
   node.peers.withValue(peerId, peer) do:
     return peer[]
   do:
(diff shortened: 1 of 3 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -268,18 +272,27 @@ proc handleLightClientUpdates(node: BeaconNode, slot: Slot) {.async.} =
   let finalized_slot = latest.finalized_header.slot
   if finalized_slot > node.lightClientPool[].latestForwardedFinalitySlot:
     template msg(): auto = latest
-    node.network.broadcastLightClientFinalityUpdate(msg)
-    node.lightClientPool[].latestForwardedFinalitySlot = finalized_slot
-    beacon_light_client_finality_updates_sent.inc()
-    notice "LC finality update sent", message = shortLog(msg)
+    let sendResult = await node.network.broadcastLightClientFinalityUpdate(msg)
```
tersec, line 272: @etan-status is this independent failure-handling of LC finality and optimistic updates correct, both in that they're handled distinctly from each other (with, e.g., no early exit to avoid the other if one fails) correct?

Also, it only sets `node.lightClientPool[].latestForwardedFinalitySlot`/`node.lightClientPool[].latestForwardedOptimisticSlot` to new values if the send succeeds, consistently with the general approach here of if the gossip didn't happen, neither did the message in some sense, and for consensus with other nodes, the node creating those messages should act as if it didn't happen. Is this how the LC update logic should work?

etan-status: These variables are just there to avoid amplifying spam, only a single message is forwarded per slot. So, if it is cheap to try and send multiple times, it's fine as is. But, due to the ephemeral validity of these messages, it is also fine to just ignore the send result ("fire and forget") and to always set the variables on the first attempt to send.

Regarding early exits, yeah they are distinct, if one fails the other one should still be attempted. They are sent to different topics with potentially different subscribers.

In the error handling logic I'd expect `warn` level in the `failed to send` message to match the level of `Peer count low, no new peers discovered` in the logs. And, probably, also a metric, given we have one for successful sends. But if that's the same way as done in other code, best to just keep it consistent. There is nothing special about the light client update gossip, except that their fork digest is of an earlier slot (and hence, the subscription must stay an epoch longer).

etan-status: I guess, you could also start sending both of them concurrently, then await them after

tersec: > I guess, you couuld also start sending both of them concurrently, then use `await allFutures(finalityUpdateFut, optimisticUpdateFut)`

That gets at a larger issue -- before this, there was no `await` around any of this, so they all got fired off and if some broadcast failed, that would be detected later. Also, the logging of some-message-sent is done on function return, not on future completion. These two things combined cause the logging discrepancy.

This PR does the simpler, but in some sense less optimal thing, of using `await` so those two things coincide. What you suggest puts back some aspect of this to the previous/current, mostly better, system, which just doesn't log well.

But if it's a problem for LC messages, then it's probably a problem elsewhere, and one better approach in some sense is to correct the discrepancy by not logging on function return, but by adding a completion callback (well, building on `traceMessage` which this removes because `await` is used instead). Because logging needs `static[string]`s and there are other side effects, this would mean passing sort of bundles of pass/fail closures/continuations to every gossip broadcast.

This gets ugly quickly for certain of the side effects, where serial operation dependency is genuine (e.g., block proposal followed by attestation: it doesn't make sense not to wait to know whether the block proposal worked, because then one's attesting to the wrong block. And likewise for aggregated attestations and unaggregated attestations. It's not just logging; that's just the most visible manifestation of this problem.

So since enough operations in time-critical sections of the code do have such dependencies, I find `await` to be the overall better tradeoff, being as the more asynchronous callback-oriented approach doesn't buy much actual, in-practice asynchronicity. Here, it might be reasonable in this very local context to, yes, use the `await allFutures(...)` approach.

tersec: https://github.com/status-im/nimbus-eth2/pull/3737/commits/6bc01478e1cf4495e0d25c70bde212e188f23762 switches these to `warn`s.

As far as metrics, it's important to be reasonably sure about which specific metrics one wants, and it's possible to add them later, if they're desired, while once they're added, they should be supported unless there's a specific reason, so I'd prefer for them to wait for another PR, to see how useful those logs are.

beacon_chain/validators/validator_duties.nim
```diff
@@ -1386,67 +1428,82 @@ proc sendAttestation*(node: BeaconNode,
...
+    node: BeaconNode, slashing: ProposerSlashing): Future[SendResult]
+    {.async.} =
+  # REST helper procedure.
   let res =
     node.processor[].proposerSlashingValidator(MsgSource.api, slashing)
   if res.isGoodForSending:
-    node.network.broadcastProposerSlashing(slashing)
-    ok()
+    let sendResult = await node.network.broadcastProposerSlashing(slashing)
+    if sendResult.isErr:
```
arnetheduck: generally if we return a `Result`, we should not be logging (ie it's the responsibility of the calling code to log / perform appropriate UX stuff)

tersec: https://github.com/status-im/nimbus-eth2/pull/3737/commits/39137c0a2432cebba48a745fa37b2c2499c72315

I checked that the various callers do this -- added a couple in `validator_duties` and verified that the REST API server already did/does.

beacon_chain/networking/eth2_network.nim
```diff
@@ -2541,75 +2520,92 @@ func forkDigestAtEpoch(node: Eth2Node, epoch: Epoch): ForkDigest =
 proc getWallEpoch(node: Eth2Node): Epoch =
   node.getBeaconTime().slotOrZero.epoch
 
-proc broadcastAttestation*(node: Eth2Node, subnet_id: SubnetId,
-                           attestation: Attestation) =
+proc broadcastAttestation*(
+    node: Eth2Node, subnet_id: SubnetId, attestation: Attestation):
+    Future[SendResult] {.async.} =
```
arnetheduck: this doesn't need to be `async`: it should just call `node.broadcast` without `return await` - doing so avoids at least 2-3 copies of the attestation

beacon_chain/networking/eth2_network.nim
```diff
@@ -2455,10 +2432,12 @@ proc broadcast*(node: Eth2Node, topic: string, msg: auto) =
       except InputTooLarge:
         raiseAssert "More than 4gb? not likely.."
 
-    inc nbc_gossip_messages_sent
-
-    var futSnappy = node.pubsub.publish(topic, compressed)
-    traceMessage(futSnappy, topic)
+    let peers = await node.pubsub.publish(topic, compressed)
+    if peers > 0:
+      inc nbc_gossip_messages_sent
```
arnetheduck, line 2439: we should metric the failures here too somehow - ie we need to measure how many messages we've lost due to no peers so that it's possible to create alerts -

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3737/commits/51ee802acf9b213a41a59f4842a9d7e9265a0642

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3737/commits/51ee802acf9b213a41a59f4842a9d7e9265a0642

arnetheduck (approved):

beacon_chain/networking/eth2_network.nim
Menduist: `libp2p_gossipsub_failed_publish`

Is already incrementing everytime we can't publish because of 0 peer, fyi

arnetheduck: hm, ok - then maybe we can document here and drop the new metric - no point repeating

tersec: https://github.com/status-im/nimbus-eth2/pull/3737/commits/41b9dfc2f6c65b403ce2f67e19ffce2a6947c5ba

PR #3738, opened 2022-06-13
remove unused `withStateVars` template

etan-status:
Removes the `withStateVars` template that was not used meaningfully.


tersec:
Another i386 Linux Nim 1.2 issue. Has never shown up in Nim 1.6, so one thing to look forward to when dropping Nim 1.2 support, or at least shipping 1.6 by default.

etan-status:
Re-scheduled CI to check whether i386 1.2 is flaky or deterministically broken.

etan-status:
i386 1.2 seems good after retry. Retrying once more

PR #3740, opened 2022-06-13
rename light client config parameters

etan-status:
For consistency with other options, use a common prefix for light client
data configuration options.

* `--serve-light-client-data` --> `--light-client-data-serve`
* `--import-light-client-data` --> `--light-client-data-import-mode`

No deprecation of the old identifiers as they were only sparingly used
and all usage can be easily updated without interferance.


etan-status:
Can be merged once https://github.com/status-im/infra-nimbus/pull/107 is accepted

PR #3749, opened 2022-06-14
avoid small gaps in optimistic block stream

etan-status:
Ensures that all intermediate blocks are reported if a small gap is
encountered when downloading optimistic blocks. Gaps may occur when
a block is missed and still downloading, or when EL processing is slow.
If the gap exceeds 1 epoch, optimistic block stream jumps to latest.


etan-status:
Tested manually, and seems to work well.

Logs:
– `LC bootstrap` shows when nimbus peer with LC data was discovered but it started after trusted block root
– `LightClientBootstrap` shows when light client was initialized
– `LC object processed` shows when light client itself progresses
– `LC block cached` shows all optimistic blocks as they are downloaded
– `New LC optimistic block` shows the optimistic block stream, `opt` is the reported block

PR #3751, opened 2022-06-14
update CI timeouts

etan-status:
Assign a longer overall timeout that a CI job can spend in queue,
and define short timeouts for the individual stages to avoid stuck jobs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,58 +14,70 @@ def runStages(nodeDir) {
 	dir(nodeDir) {
 		try {
 			stage("Clone") {
-				/* source code checkout */
-				checkout scm
-				/* we need to update the submodules before caching kicks in */
-				sh "git submodule update --init --recursive"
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Good idea. I'm not 100% sure that this will work without a named argument, but other than that looks good.

Jenkinsfile
```diff
@@ -93,7 +93,7 @@ def runStages(nodeDir) {
 parallel(
 	"Linux": {
 		throttle(['nimbus-eth2']) {
-			timeout(time: 5, unit: 'HOURS') {
```
arnetheduck, line 96: 24? actions is fickle, of late - with each step being limited, a large value here shouldn't be a problem

etan-status:
> Good idea. I'm not 100% sure that this will work without a named argument, but other than that looks good.

Seems to work fine without the named arg, at least in the case where the timeout is not hit.

PR #3753, opened 2022-06-14
track LC data section that imported without errors

etan-status:
If database access errors are encountered while proccessing LC data,
track the section which was accessed without errors so that the rest
may be attempted to be re-indexed later.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -183,20 +183,31 @@ proc deleteLightClientData*(dag: ChainDAGRef, bid: BlockId) =
 
   dag.lightClientCache.data.del bid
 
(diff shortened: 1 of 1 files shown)


beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -585,29 +617,40 @@ proc initLightClientUpdateForPeriod(
   # period, despite it having a lower sync committee participation
   var
     tmpState = assignClone(dag.headState)
-    signatureBid = BlockId(slot: FAR_FUTURE_SLOT)
-    finalizedBid = BlockId(slot: FAR_FUTURE_SLOT)
+    signatureBid {.noinit.}, finalizedBid {.noinit.}: BlockId
```
zah, line 620: Why is this change considered an improvement?

etan-status: Avoids extra zeroing of `BlockId.root` that is never read back.

PR #3755, opened 2022-06-15
bearssl: split abi

arnetheduck:


beacon_chain/spec/keystore.nim
```diff
@@ -357,20 +352,19 @@ template add(m: var Mnemonic, s: cstring) =
...
-             entropyParam.len <= 256 and
-             entropyParam.len mod 32 == 0
-    entropy = @entropyParam
+  var entropy =
+    if entropyParam.len == 0:
+      rng.generateBytes(32)
+    else:
+      doAssert entropyParam.len >= 128 and
+              entropyParam.len <= 256 and
+              entropyParam.len mod 32 == 0
```
etan-status:
```suggestion
      doAssert entropyParam.len >= 128 and
               entropyParam.len <= 256 and
               entropyParam.len mod 32 == 0
```

arnetheduck: one day, I'm going to write an autoformatter just so you don't have to get worried about these spaces ;)

mratsim (approved):

PR #3759, opened 2022-06-16
rewrite merge sync

tersec:
Tested with Ropsten and Geth in various combinations of situations.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -23,8 +23,6 @@ import
   ../beacon_clock,
   ../sszdump
 
-from std/options import `$`
-
 export sszdump, signatures_batch
 
 # Block Processor
@@ -322,6 +320,8 @@ proc processBlock(self: var BlockProcessor, entry: BlockEntry) =
       if res.isOk(): Result[void, BlockError].ok()
       else: Result[void, BlockError].err(res.error()))
 
+func `$`(h: BlockHash): string = $h.asEth2Digest
+
 proc runForkchoiceUpdated(
     self: ref BlockProcessor, headBlockRoot, finalizedBlockRoot: Eth2Digest):
     Future[bool] {.async.} =
@@ -423,8 +423,6 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
       # larger network reads when under load.
       idleTimeout = 10.milliseconds
 
-      defaultBellatrixPayload = default(bellatrix.ExecutionPayload)
-
     discard await idleAsync().withTimeout(idleTimeout)
 
     let
@@ -432,8 +430,7 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
       hasExecutionPayload = blck.blck.kind >= BeaconBlockFork.Bellatrix
       executionPayloadStatus =
         if  hasExecutionPayload and
-            blck.blck.bellatrixData.message.body.execution_payload !=
-              defaultBellatrixPayload:
+            blck.blck.bellatrixData.message.body.is_execution_block:
           # Eth1 syncing is asynchronous from this
 
           # TODO self.consensusManager.eth1Monitor.terminalBlockHash.isSome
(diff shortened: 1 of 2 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -422,19 +430,26 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
       hasExecutionPayload = blck.blck.kind >= BeaconBlockFork.Bellatrix
       executionPayloadStatus =
         if  hasExecutionPayload and
-            # Allow local testnets to run without requiring an execution layer
             blck.blck.bellatrixData.message.body.execution_payload !=
               defaultBellatrixPayload:
```
```suggestion
            blck.blck.bellatrixData.message.body.is_execution_block:
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -401,6 +406,9 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
     optForkchoiceHeadSlot = GENESIS_SLOT # safe default
     optForkchoiceHeadRoot: Eth2Digest
 
+    # don't keep spamming same fcU to Geth; might be restarting sync each time
```
etan-status, line 411: Are these mentions Geth specifics? Maybe "EL client", also as behaviour may change with different versions.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -481,28 +490,44 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
...
+      #
+      # In particular, the constraints that hold here are that Geth expects a
+      # sequence of
+      # - newPayload(execution payload with block hash `h`) followed by
+      # - forkchoiceUpdated(head = `h`)
+      # This is intrinsically somewhat optimistic, because it's not possible a
+      # CL to even get Geth to state whether an execution payload is valid, in
+      # general, without setting the forkchoiceUpdated head to a block hash of
+      # some execution payload that it doesn't, in fact, know the validity of;
+      # otherwise it wouldn't be asking the EL.
```
```suggestion
      # This is intrinsically somewhat optimistic, because determining the 
      # validity of an execution payload requires the forkchoiceUpdated 
      # head to be set to a block hash of some execution payload with unknown
      # validity; otherwise it would not be necessary to ask the EL.
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -481,28 +490,44 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
...
+      #
+      # The main reason this isn't done more adjacently in this code flow is to
+      # outright invalid cases, where the EL can reject a payload, without even
+      # running forkchoiceUpdated on it.
```
```suggestion
      # The main reason this isn't done more adjacently in this code flow is to
      # catch outright invalid cases, where the EL can reject a payload, without
      # even running forkchoiceUpdated on it.
```

beacon_chain/gossip_processing/block_processor.nim
tersec: Yes, they're specific to Geth. Probably other ELs have some similar conditions, but this PR attempts to capture the union of Geth's constraints and the engine API specs. The items mentioned related to "Geth" are Geth-specific though and aren't per se part of the engine API.

The discussion of "ELs" in other comments refers to the construction/design of the spec itself.

tersec: For example, it'd be possible by the specs for an EL to allow `forkchoiceUpdated` on its own to trigger a sync, but Geth (sensibly enough, but not because the spec requires it) requires that `newPayload` / `forkchoiceUpdated` pair on the same block hash. It's the clear intent of the spec that they be paired, but it never per se requires it, unlike Geth. Other ELs might behave differently in this regard.

Geth also, and this is I believe widely shared behavior but not required by the spec, ignores certain backwards-moving `forkchoiceUpdated` calls.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3759/commits/9f02d1d5f318c2277a4d2d0e815ad98bd361ae4b

PR #3698, opened 2022-06-03
initial Capella support in RuntimeConfig

tersec:
It's part of CL specs version 1.2.0.

Also fixes
```
@["Unknown constants in file: ", "TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH", "PROPOSER_SCORE_BOOST"]
```
showing up during builds.

PR #3767, opened 2022-06-16
bump merge-testnets to use capella fork versions consistently

tersec:
This enables more rigorous fork consistency checking across all supported networks, in particular including mainnet shadow forks 2, 6, and 7 (the only active or soon-to-be-active ones).

The remaining config to check and perhaps fix is the Gnosis one.


etan-status:
One thing to verify that these are properly picked up could be the REST spec/config endpoint

tersec:
This PR doesn't per se affect that, or shouldn't. Adding the Capella runtime information was already done in https://github.com/status-im/nimbus-eth2/pull/3698 and that does add support to the REST endpoints.

PR #3772, opened 2022-06-17
include capella fork version in fork consistency check

tersec:
Followup to https://github.com/status-im/nimbus-eth2/pull/3767

Tested at build time with mainnet, Prater, Ropsten, and Sepolia, and Gnosis (the last using the define flag to trigger it):
https://github.com/status-im/nimbus-eth2/blob/c48ea3cdc8b3fec12a1256bc0c95a1ba1010d727/beacon_chain/networking/network_metadata.nim#L251-L253
https://github.com/status-im/nimbus-eth2/blob/c48ea3cdc8b3fec12a1256bc0c95a1ba1010d727/beacon_chain/networking/network_metadata.nim#L300

Tested with those at runtime, plus some not included in the compile-time set in this branch or `unstable`: kiln and mainnet shadow forks 2, 6, and 7.

### DIFF
--- a/tests/test_conf.nim
+++ b/tests/test_conf.nim
@@ -1,3 +1,11 @@
+# beacon_chain
+# Copyright (c) 2022 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [Defect].}
 {.used.}
 
 import
--- a/tests/test_conf.nim
+++ b/tests/test_conf.nim
@@ -1,3 +1,5 @@
+{.used.}
+
 import
   unittest2,
   ../beacon_chain/conf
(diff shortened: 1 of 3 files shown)


tests/test_conf.nim
```diff
@@ -1,3 +1,5 @@
+{.used.}
```
etan-status, line 9:
```suggestion
# beacon_chain
# Copyright (c) 2022 Status Research & Development GmbH
# Licensed and distributed under either of
#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
# at your option. This file may not be copied, modified, or distributed except according to those terms.

{.push raises: [Defect].}
{.used.}
```

etan-status (approved):

tests/test_conf.nim
tersec: Does `{.push raises: [Defect].}` work here? For reasons I'm not that clear on, it's not typically used in the tests. That might just be cargo-culting -- the existing tests don't, so neither do new tests people write -- but maybe it's because it genuinely can't be used.

etan-status: `test_block_clearance_light_client.nim` uses it, works fine. Not sure if it actually does anything, though.

The style guide does not list exceptions so maybe just missed in most tests so far (similar to copyright)
```
# Enable exception tracking for all functions in this module
`{.push raises: [Defect].}` # Always at start of module
```

https://status-im.github.io/nim-style-guide/errors.exceptions.html

arnetheduck: >  it's not typically used in the tests.

in tests, it's less useful because the `test` template has a generic exception-catcher - that said, it's certainly not wrong to add it, if only to provide a better copy-paste source

PR #3776, opened 2022-06-19
Fix #3650 (participation format in BeaconState result is out of spec)

### LINKED ISSUE #3650: Participation format in BeaconState result is out of spec
**Describe the bug**
During testing on hive with nimbus beacon client, hive uses https://github.com/protolambda/zrnt to parse the REST API results, and is currently resulting in error during a query to `eth/v2/debug/beacon/states/head`.

The reason is that nimbus currently returns two values which are out of spec on `previous_epoch_participation` and `current_epoch_participation`, which are specified as `List[ParticipationFlags, VALIDATOR_REGISTRY_LIMIT]` (effectively a list of `uint8`), both specified here https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#beaconstate

Current result from nimbus is:
```
"previous_epoch_participation": "0x0707070707070707...
```
instead of the expected
```
"previous_epoch_participation": ["7","7","7", ...
```

Hive simulator is aborting early due to this.

**To Reproduce**
On nimbus running altair fork or later:
`curl http://127.0.0.1:4000/eth/v2/debug/beacon/states/head`

zah:
Notable sub-module changes:

* Support trailing comma when loading sequence or inline tables in TOML:
  https://github.com/status-im/nim-toml-serialization/pull/47

### DIFF
--- a/beacon_chain/spec/datatypes/altair.nim
+++ b/beacon_chain/spec/datatypes/altair.nim
@@ -79,7 +79,7 @@ type
   ParticipationFlags* = uint8
 
   EpochParticipationFlags* =
-    HashList[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT]
+    distinct HashList[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT]
 
   # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/altair/beacon-chain.md#syncaggregate
   SyncAggregate* = object
@@ -579,6 +579,27 @@ template `[]`*(arr: array[SYNC_COMMITTEE_SIZE, auto] | seq;
 
 makeLimitedU8(SyncSubcommitteeIndex, SYNC_COMMITTEE_SUBNET_COUNT)
 makeLimitedU16(IndexInSyncCommittee, SYNC_COMMITTEE_SIZE)
+
+template asHashList*(epochFlags: EpochParticipationFlags): untyped =
+  HashList[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT] epochFlags
+
+template item*(epochFlags: EpochParticipationFlags, idx: ValidatorIndex): ParticipationFlags =
+  asHashList(epochFlags).item(idx)
+
+template `[]`*(epochFlags: EpochParticipationFlags, idx: ValidatorIndex|uint64): ParticipationFlags =
+  asHashList(epochFlags)[idx]
+
+template `[]=`*(epochFlags: EpochParticipationFlags, idx: ValidatorIndex, flags: ParticipationFlags) =
+  asHashList(epochFlags)[idx] = flags
+
+template add*(epochFlags: var EpochParticipationFlags, flags: ParticipationFlags): bool =
+  asHashList(epochFlags).add flags
+
+template len*(epochFlags: EpochParticipationFlags): int =
+  asHashList(epochFlags).len
+
+template data*(epochFlags: EpochParticipationFlags): untyped =
+  asHashList(epochFlags).data
 
 func shortLog*(v: SomeBeaconBlock): auto =
   (
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -133,15 +133,15 @@ type
 
 {.push raises: [Defect].}
 
-proc writePreviousEpochParticipation(writer: var JsonWriter[RestJson],
-                                     values: EpochParticipationFlags)
-                                    {.raises: [IOError, Defect].} =
-  for e in writer.stepwiseArrayCreation(values):
+proc writeValue*(writer: var JsonWriter[RestJson],
+                 epochFlags: EpochParticipationFlags)
+                {.raises: [IOError, Defect].} =
+  for e in writer.stepwiseArrayCreation(epochFlags.asHashList):
     writer.writeValue $e
 
-proc readPreviousEpochParticipation(
-    reader: var JsonReader[RestJson]): EpochParticipationFlags
-    {.raises: [SerializationError, IOError, Defect].} =
+proc readValue*(reader: var JsonReader[RestJson],
+                epochFlags: var EpochParticipationFlags)
+               {.raises: [SerializationError, IOError, Defect].} =
   # Please note that this function won't compute the cached hash tree roots
   # immediately. They will be computed on the first HTR attempt.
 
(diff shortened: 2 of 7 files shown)


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -128,8 +128,51 @@ type
...
+    except ValueError as err:
+      reader.raiseUnexpectedValue("A string-encoded 8-bit usigned integer value expected")
+
+    if parsed > uint8.high:
+      reader.raiseUnexpectedValue("The usigned integer value should fit in 8 bits")
+
+    if not result.data.add(uint8(parsed)):
+      reader.raiseUnexpectedValue("The participation flags list size exceeds limit")
+
+RestJson.useCustomSerialization(AltairBeaconState.current_epoch_participation):
```
arnetheduck: why not make ParticipationFlags a distinct type instead? this custom serialization stuff is quite fragile because the default is wrong: you don't get an error from the compiler if it's missing from the compile (same reason why Rest in general must never automatically serialize types)

arnetheduck: notably, this format for flags is also required when reading yaml files (in tests) - it's a distinct type in the REST spec, not a "custom field" - ie any time participation flags are serialized, they should be serialized as numbers - this is not specific to these fields

zah: This would require changing all usage sites, tweaks in SSZ, etc.

arnetheduck: well, yes - that's the point, a bit - it's a different type, per one of the specs for which it's used, so that should be reflected in our code as well.

arnetheduck: serializing by field maintains the existing semantic mismatch - sooner or later, the semantic mismatch will lead to a bug similar to the one that's being fixed here - custom field serialization is not the tool for the job here (it rarely is)

zah: The counter-argument is that it's not nice that all of the project code has to be modified just because one of the specs is slightly botched. If there is a consumer of the codebase that is not aware of the peculiarities of the REST API, they will be inconvenienced by the inconsistency. It's a trade-off and the answer depends on your values. Elsewhere, you've fought against changing the spec code at a great expense for the maintainer of the reward tracking code. Here, the effort for maintaining the REST implementation would be arguably much smaller.

But enough theorizing, I'll try the distinct type approach and if's not too invasive, I might submit an alternative PR. 

p.p.
We are not using the YAML test definitions. We used them briefly before the SSZ tests became available. Removing the support was advantageous because we had to patch the yaml library in the middle of the test run in order to support all tests (some tests required the original version, others required the patched version, it was quite the hack).

arnetheduck: > just because one of the specs is slightly botched

it's not botched really - it was an intentional (and documented) choice - if we weren't reusing spec types for serialization as well (ie overloading their responsibilities), this would be a non-problem - but as long as we're using type-based serializers, and the specs that we use them for are specified by type, well... here we are.

I can even predict the time of the next bug report: around the time of the next fork when we add a new beaconstate :)

arnetheduck: arguably, we should also remove the `list[byte-like]` overloads for serialization because they also are overly permissive, and if another field type is added that looks like a byte but is a number, it should not be picked up automatically either - instead, there should be a compile-time error so the right serialization can be chosen deliberately

zah: So, how would these non type-based serializers look like? If you have a spec where the same type is serialized in different ways in two different places, I would say "slightly botched" is an accurate description.

arnetheduck: Well .. it's not the same type, that's the point - it's a different type in one of the specs that is relevant to serialization and we've interpreted it wrong.

The pedantic solution here, if we were to not use the spec types for REST, would be a `RestBeaconState` and `RestParticipationFlags`. This would be "correct" but also truly a lot of work. The distinct type is not a lot of work, comparatively.

Using custom field serializers is ugly hack, and we can do it for this release perhaps, but it should be fixed immediately after.

Non-type-based serializers are simple functions that load data field by field without relying on type matching magic - typically, they'd be generated from the oapi spec.

zah: Continuing this discussion is a bit pointless, but why should the REST API use a different definition of the `BeaconState` type? Where is this "intention" documented? I would argue the inconsistency was an oversight and an unfortunate design flaw because it brings no benefits and increases the cost of the implementation significantly as you admit yourself.

arnetheduck: > Where is this "intention" documented? 

I asked the people more heavily heavily involved than us in beacon api spec writing. For them, this is not a flaw: it is the most natural way to represent the flags. We're looking at it from a SSZ-centric point of view (because our types were first developed for SSZ) - they're looking at it form a JSON-centric one, where encoding lists of numbers as .. lists of numbers is more natural. 

More formally, it's documented here: https://github.com/ethereum/beacon-APIs/blob/master/types/altair/epoch_participation.yaml#L6

>  but why should the REST API use a different definition of the BeaconState type

Well, we use Nim types to represent both REST types and SSZ types - the REST types are more information-rich than SSZ, and we must preserve that information: SSZ is a subset of REST, type-richness-wise, so we can save ourselves some work by reducing REST to SSZ and not the other way around. We're already reducing Nim types to REST and to SSZ: neither of these two standards match Nim exactly - this is important to remember, because the Nim types we use in the spec code will always be a compromise between the three use cases, as long as we reuse them for both Nim code, SSZ and REST.

Effectively, this becomes a constraint: it means we also can't use all features of Nim in our spec code - ie some Nim features don't map 1:1 to SSZ/REST and vice versa - we're left with the lowest common denominator between these use cases. It's a "cost" of serialization frameworks in general, reusing types that you use in code as "serialization specifications" which inevitably ends up having semantic differences.

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -79,7 +79,7 @@ type
   ParticipationFlags* = uint8
 
   EpochParticipationFlags* =
-    HashList[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT]
+    distinct HashList[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT]
```
arnetheduck, line 82: this is something of an odd choice - it's the `ParticipationFlags` type that has a different serialization than the other "byte-like" types (`UInt8` in the beacon API) - if later we have an `array[ParticipationFlags]` it will be serialized as a list of numbers. we can leave as-is for the release but this spec deviation should be noted in the code

see https://github.com/ethereum/beacon-APIs/blob/e3a3f6e1edb25371a0641d16097d13ee369589bc/types/primitive.yaml#L51 where `Bytes32` and `Root` is what makes a list be serialized as a string - ie our code is not representative of what goes on in the spec.

PR #3777, opened 2022-06-19
encapsulate LC data variables into single structure

etan-status:
Combines the LC data configuration options (serve / importMode), the
callbacks (finality / optimistic LC update) as well as the cache storing
light client data, into a new `LightClientDataStore` structure.
Also moves the structure into a light client specific file.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
+++ b/beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
@@ -18,11 +18,6 @@ import
   ./block_dag
 
 type
-  OnLightClientFinalityUpdateCallback* =
-    proc(data: altair.LightClientFinalityUpdate) {.gcsafe, raises: [Defect].}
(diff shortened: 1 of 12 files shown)


beacon_chain/consensus_object_pools/block_pools_types_light_client.nim
```diff
@@ -79,3 +79,26 @@ type
 
     importTailSlot*: Slot
       ## The earliest slot for which light client data is imported.
+
+  LightClientDataStore* = object
```
arnetheduck, line 83: technically, this is a "light server" store

etan-status: Yeah, haha, but that terminology is nowhere used yet, so sticking with "light client data" for now (technically, a client could also cache the obtained data and then provide it to others, or use it to roll-back in case it fails to advance when there is no supermajority, so use case of this data is not limited to just server)

arnetheduck (approved):

PR #3780, opened 2022-06-20
Add the `execution_optimistic` flag to REST API responses.

cheatfate:
* Add the `execution_optimistic` flag to REST API responses.
* Add `Eth-Consensus-Version` HTTP response header to `/eth/v2/beacon/blocks/{block_id}` and `/eth/v2/debug/beacon/states/{state_id}` endpoints.
* Add implementation of `/eth/v2/debug/beacon/heads` endpoint.
* Add `execution_optimistic` to `/eth/v1/events` outputs.
* Add `is_optimistic` to `/eth/v1/node/syncing` endpoint.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -163,7 +163,8 @@ proc prepareJsonStringResponse*(t: typedesc[RestApiResponse], d: auto): string =
   res
 
 proc jsonResponseWRoot*(t: typedesc[RestApiResponse], data: auto,
-                        dependent_root: Eth2Digest): RestApiResponse =
+                        dependent_root: Eth2Digest,
+                        execOpt: Option[bool]): RestApiResponse =
   let res =
     block:
       var default: seq[byte]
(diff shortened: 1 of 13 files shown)


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -198,6 +202,84 @@ proc jsonResponse*(t: typedesc[RestApiResponse], data: auto): RestApiResponse =
...
+        writer.beginRecord()
+        writer.writeField("version", data.kind.toString())
+        if execOpt.isSome():
+          writer.writeField("execution_optimistic", execOpt.get())
+        withBlck(data):
+          writer.writeField("data", blck)
+        writer.endRecord()
+        stream.getOutput(seq[byte])
+      except SerializationError:
+        default
```
zah, line 225: `SerializationError` doesn't seem possible here, but why should this code return an empty response in case of an error instead of returning an error response?

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -1988,6 +2073,68 @@ proc dump*(value: KeystoresAndSlashingProtection): string {.
...
+  writer.beginRecord()
+  writer.writeField("slot", value.slot)
+  writer.writeField("block", value.block_root)
+  writer.writeField("state", value.state_root)
+  writer.writeField("epoch_transition", value.epoch_transition)
+  writer.writeField("previous_duty_dependent_root",
+                    value.previous_duty_dependent_root)
+  writer.writeField("current_duty_dependent_root",
+                    value.current_duty_dependent_root)
+  if value.optimistic.isSome():
```
zah, line 2088: With the new `Option` support, you should be able to remove these procs, no?

cheatfate: I expect it.

PR #3788, opened 2022-06-22
[WIP] floating tip sync

tersec:


tersec:
Going with a true, spec optimistic sync instead in https://github.com/status-im/nimbus-eth2/pull/3793

PR #3789, opened 2022-06-22
bump blst - fix 32-bit builds on ADX CPUs

mratsim:
BLST could try to load inexisting assembly routines when build for 32-bit on CPUs with ADX (and -march=native)

image

Upstream has fixed that https://github.com/supranational/blst/pull/122


arnetheduck (approved):

arnetheduck:
quite a bit of changes in https://github.com/supranational/blst/compare/164ce6280bf9579086fe60c5cc2ca9c8bc1f0093...b38a52c86e4f5ae69ba6eca41d050707b6fcc72b - some of which look siginificant, ie extra zero checks - anything that affects us?

mratsim:
The zero check is relevant if you deserialize/serialize the result of a pairing operation. We don't do that. Also it's correct.

PR #3790, opened 2022-06-22
Add support for launching Geth and Nimbus-eth1 nodes in launch_local_testnet

zah:
Authored by @hmel


scripts/launch_local_testnet.sh
```diff
@@ -72,7 +82,20 @@ ETH2_DOCKER_IMAGE=""
 REMOTE_SIGNER_NODES=0
 REMOTE_SIGNER_THRESHOLD=1
 REMOTE_VALIDATORS_COUNT=0
-LC_NODES=1
+LC_NODES=0
```
etan-status: hm?

zah: I'm not sure how this came to be, but it looks like the network is failing to finalize when there are some light nodes under the new regime here. I've pushed `LC_NODES=1`, so you can take a look at the logs.

etan-status: Seems like it finalizes properly.

scripts/launch_local_testnet.sh
```diff
@@ -740,13 +928,15 @@ if [ "$LC_NODES" -ge "1" ]; then
...
-        "${JQ_BINARY}" -r '.data.epoch')"
+        tee -a curl_result.txt | "${JQ_BINARY}" -r '.data.epoch')"
     if [ "${CURRENT_FORK_EPOCH}" -ge "${ALTAIR_FORK_EPOCH}" ]; then
       break
     fi
     sleep 1
   done
 
+  log "After ALTAIR_FORK_EPOCH"
+
```
etan-status, line 939: There is already a log one line below this

scripts/launch_local_testnet.sh
zah: It fails locally, but alright, we can keep investigating this after merging.

PR #3792, opened 2022-06-23
fix action tracker validator decay

arnetheduck:
Validators were staying around 32x as long as they should - the effect
is minimal because it's rare validators disappear, but still..

### DIFF
--- a/beacon_chain/validators/action_tracker.nim
+++ b/beacon_chain/validators/action_tracker.nim
@@ -9,26 +9,24 @@ import
   std/[sequtils, tables],
   stew/shims/[sets, hashes], chronicles,
   eth/p2p/discoveryv5/random2,
-  ../spec/datatypes/base,
-  ../spec/[helpers, network],
-  ../consensus_object_pools/[blockchain_dag, spec_cache]
+  ../spec/forks,
+  ../consensus_object_pools/spec_cache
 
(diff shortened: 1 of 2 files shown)


etan-status (approved):

beacon_chain/validators/action_tracker.nim
```diff
@@ -139,7 +137,9 @@ func updateSlot*(tracker: var ActionTracker, wallSlot: Slot) =
   var toPrune: seq[ValidatorIndex]
   for k, v in tracker.knownValidators:
     if v + KNOWN_VALIDATOR_DECAY < wallSlot: toPrune.add k
```
Menduist, line 139: Shouldn't this be `>`?

Menduist (changes requested):

beacon_chain/validators/action_tracker.nim
etan-status: It seems correct to me.

if v is 5 and wallSlot is 1000, validator should be pruned, and 5 + 3\*32 < 1000.
if v is 999 and wallSlot is 1000, validator should be kept, and 5 + 3\*32 >= 1000

Menduist: Indeed, my bad

Menduist (approved):

PR #3793, opened 2022-06-23
optimistic sync

tersec:
https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -29,6 +29,7 @@ proc addResolvedHeadBlock(
        dag: ChainDAGRef,
        state: var ForkedHashedBeaconState,
        trustedBlock: ForkyTrustedSignedBeaconBlock,
+       blockVerified: bool,
        parent: BlockRef, cache: var StateCache,
        onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded | OnBellatrixBlockAdded,
        stateDataDur, sigVerifyDur, stateVerifyDur: Duration
@@ -73,6 +74,7 @@ proc addResolvedHeadBlock(
   debug "Block resolved",
     blockRoot = shortLog(blockRoot),
     blck = shortLog(trustedBlock.message),
+    blockVerified,
     heads = dag.heads.len(),
     stateDataDur, sigVerifyDur, stateVerifyDur,
     putBlockDur = putBlockTick - startTick,
@@ -81,8 +83,13 @@ proc addResolvedHeadBlock(
   # Update light client data
   dag.processNewBlockForLightClient(state, trustedBlock, parent.bid)
 
+  if not blockVerified:
+    dag.optimisticRoots.incl blockRoot
+
   # Notify others of the new block before processing the quarantine, such that
   # notifications for parents happens before those of the children
+  # TODO these might assume full blocks, but also, probably required for normal
+  # functioning
   if onBlockAdded != nil:
     onBlockAdded(blockRef, trustedBlock, epochRef)
   if not(isNil(dag.onBlockAdded)):
@@ -136,6 +143,7 @@ proc advanceClearanceState*(dag: ChainDAGRef) =
 proc addHeadBlock*(
     dag: ChainDAGRef, verifier: var BatchVerifier,
     signedBlock: ForkySignedBeaconBlock,
+    blockVerified: bool,
     onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded |
                   OnBellatrixBlockAdded
     ): Result[BlockRef, BlockError] =
@@ -256,11 +264,21 @@ proc addHeadBlock*(
   ok addResolvedHeadBlock(
     dag, dag.clearanceState,
     signedBlock.asTrusted(),
+    blockVerified = blockVerified,
     parent, cache,
     onBlockAdded,
     stateDataDur = stateDataTick - startTick,
     sigVerifyDur = sigVerifyTick - stateDataTick,
     stateVerifyDur = stateVerifyTick - sigVerifyTick)
+
+proc addHeadBlock*(
+    dag: ChainDAGRef, verifier: var BatchVerifier,
+    signedBlock: ForkySignedBeaconBlock,
+    onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded |
+                  OnBellatrixBlockAdded
+    ): Result[BlockRef, BlockError] =
+  addHeadBlock(
+    dag, verifier, signedBlock, blockVerified = true, onBlockAdded)
 
 proc addBackfillBlock*(
     dag: ChainDAGRef,
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -88,8 +88,6 @@ proc addResolvedHeadBlock(
 
   # Notify others of the new block before processing the quarantine, such that
   # notifications for parents happens before those of the children
-  # TODO these might assume full blocks, but also, probably required for normal
-  # functioning
   if onBlockAdded != nil:
     onBlockAdded(blockRef, trustedBlock, epochRef)
   if not(isNil(dag.onBlockAdded)):
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -746,7 +746,8 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   # Load head -> finalized, or all summaries in case the finalized block table
   # hasn't been written yet
   for blck in db.getAncestorSummaries(head.root):
-    let newRef = BlockRef.init(blck.root, ZERO_HASH, blck.summary.slot)
+    # The execution block root gets filled in as needed
+    let newRef = BlockRef.init(blck.root, none Eth2Digest, blck.summary.slot)
     if headRef == nil:
       doAssert blck.root == head.root
       headRef = newRef
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -165,12 +165,11 @@ func init*(
       attester_dependent_root: attester_dependent_root,
       merge_transition_complete:
         case state.kind:
-        of BeaconStateFork.Phase0: false
-        of BeaconStateFork.Altair: false
+        of BeaconStateFork.Phase0, BeaconStateFork.Altair: false
         of BeaconStateFork.Bellatrix:
           # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#is_merge_transition_complete
           state.bellatrixData.data.latest_execution_payload_header !=
-            ExecutionPayloadHeader()
+            (static(ExecutionPayloadHeader()))
     )
     epochStart = epoch.start_slot()
 
@@ -950,6 +949,17 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
 
   dag.initLightClientDataCache()
 
+  # If these aren't actually optimistic, the first fcU will resolve that
+  withState(dag.headState):
+    when stateFork >= BeaconStateFork.Bellatrix:
+      template executionPayloadHeader(): auto =
+        state().data.latest_execution_payload_header
+      const emptyExecutionPayloadHeader =
+        default(type(executionPayloadHeader))
+      if executionPayloadHeader != emptyExecutionPayloadHeader:
+        dag.optimisticRoots.incl dag.head.root
+        dag.optimisticRoots.incl dag.finalizedHead.blck.root
+
   dag
 
 template genesis_validators_root*(dag: ChainDAGRef): Eth2Digest =
@@ -1375,6 +1385,8 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
       "finalizedHead parent should have been pruned from memory already"
 
     while cur.blck.parent != nil:
+      dag.optimisticRoots.excl cur.blck.root
+
       # TODO: should we move that disk I/O to `onSlotEnd`
       dag.delState(cur.toBlockSlotId().expect("not nil"))
 
@@ -1393,6 +1405,75 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
     currentCandidateHeads = dag.heads.len,
     prunedHeads = hlen - dag.heads.len,
     dagPruneDur = Moment.now() - startTick
+
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#helpers
+template is_optimistic*(dag: ChainDAGRef, root: Eth2Digest): bool =
+  root in dag.optimisticRoots
+
+proc markBlockInvalid*(dag: ChainDAGRef, root: Eth2Digest) =
+  let blck = dag.getBlockRef(root).valueOr:
+    return
+  logScope: blck = shortLog(blck)
+
+  if not dag.is_optimistic(root):
+    # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#transitioning-from-valid---invalidated-or-invalidated---valid
+    # "It is outside of the scope of the specification since it's only possible
+    # with a faulty EE. Such a scenario requires manual intervention."
+    warn "markBlockInvalid: attempt to invalidate valid block"
+    doAssert verifyFinalization notin dag.updateFlags
+    return
+
+  if root == dag.finalizedHead.blck.root:
+    # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#re-orgs
+    # "If the justified checkpoint transitions from `NOT_VALIDATED` ->
+    # `INVALIDATED`, a consensus engine MAY choose to alert the user and force
+    # the application to exit."
+    #
+    # But be slightly less aggressive, and only check finalized.
+    warn "markBlockInvalid: finalized block invalidated"
+    doAssert verifyFinalization notin dag.updateFlags
+    return
+
+  dag.optimisticRoots.excl root
+
+  debug "markBlockInvalid"
+
+  # TODO copied from inner loop in pruneBlocksDAG, could refactor
+  dag.delState(blck.bid.atSlot())
+
+  # TODO These are gated by cur.isProposed in pruneBlocksDAG, but here should
+  # not be necessary or useful
+  dag.deleteLightClientData(blck.bid)
+  dag.forkBlocks.excl(KeyedBlockRef.init(blck))
+  dag.db.delBlock(root)
+
+proc markBlockVerified*(
+    dag: ChainDAGRef, quarantine: var Quarantine, root: Eth2Digest) =
+  # Might be called when block was never optimistic to begin with. In that
+  # circumstance, bail early.
+  if not dag.is_optimistic(root):
+    return
+
+  var cur = dag.getBlockRef(root).valueOr:
+    return
+  logScope: blck = shortLog(cur)
+
+  debug "markBlockVerified",
+    blck = shortLog(cur)
+
+  while true:
+    if not dag.is_optimistic(cur.bid.root):
+      return
+
+    dag.optimisticRoots.excl cur.bid.root
+
+    debug "markBlockVerified ancestor",
+      blck = shortLog(cur)
+
+    if cur.parent.isNil:
+      break
+
+    cur = cur.parent
 
 iterator syncSubcommittee*(
     syncCommittee: openArray[ValidatorIndex],
@@ -1533,6 +1614,27 @@ template getHeadStateMergeComplete*(dag: ChainDAGRef): bool =
     else:
       false
 
+proc loadExecutionBlockRoot*(dag: ChainDAGRef, blck: BlockRef): Eth2Digest =
+  if dag.cfg.blockForkAtEpoch(blck.bid.slot.epoch) < BeaconBlockFork.Bellatrix:
+    return ZERO_HASH
+
+  if blck.executionBlockRoot.isSome:
+    return blck.executionBlockRoot.get
+
+  let blockData = dag.getForkedBlock(blck.bid).valueOr:
+    blck.executionBlockRoot = some ZERO_HASH
+    return ZERO_HASH
+
+  let executionBlockRoot =
+    withBlck(blockData):
+      when stateFork >= BeaconStateFork.Bellatrix:
+        blck.message.body.execution_payload.block_hash
+      else:
+        ZERO_HASH
+  blck.executionBlockRoot = some executionBlockRoot
+
+  executionBlockRoot
+
 proc updateHead*(
     dag: ChainDAGRef,
     newHead: BlockRef,
@@ -1614,7 +1716,8 @@ proc updateHead*(
       stateRoot = shortLog(getStateRoot(dag.headState)),
       justified = shortLog(getStateField(
         dag.headState, current_justified_checkpoint)),
-      finalized = shortLog(getStateField(dag.headState, finalized_checkpoint))
+      finalized = shortLog(getStateField(dag.headState, finalized_checkpoint)),
+      isOptHead = dag.is_optimistic(newHead.root)
 
     if not(isNil(dag.onReorgHappened)):
       let
@@ -1635,7 +1738,8 @@ proc updateHead*(
       stateRoot = shortLog(getStateRoot(dag.headState)),
       justified = shortLog(getStateField(
         dag.headState, current_justified_checkpoint)),
-      finalized = shortLog(getStateField(dag.headState, finalized_checkpoint))
+      finalized = shortLog(getStateField(dag.headState, finalized_checkpoint)),
+      isOptHead = dag.is_optimistic(newHead.root)
 
     if not(isNil(dag.onHeadChanged)):
       let
@@ -1685,8 +1789,8 @@ proc updateHead*(
 
       dag.db.updateFinalizedBlocks(newFinalized)
 
-    if  oldFinalizedHead.blck.executionBlockRoot.isZero and
-        not dag.finalizedHead.blck.executionBlockRoot.isZero:
+    if  dag.loadExecutionBlockRoot(oldFinalizedHead.blck).isZero and
+        not dag.loadExecutionBlockRoot(dag.finalizedHead.blck).isZero:
       dag.vanityLogs.onFinalizedMergeTransitionBlock()
 
     # Pruning the block dag is required every time the finalized head changes
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1352,6 +1352,18 @@ proc delState(dag: ChainDAGRef, bsi: BlockSlotId) =
     dag.db.delState(root.get())
     dag.db.delStateRoot(bsi.bid.root, bsi.slot)
 
+proc pruneBlockSlot(dag: ChainDAGRef, bs: BlockSlot) =
+  # TODO: should we move that disk I/O to `onSlotEnd`
+  dag.delState(bs.toBlockSlotId().expect("not nil"))
+
+  if bs.isProposed():
+    # Update light client data
+    dag.deleteLightClientData(bs.blck.bid)
+
+    dag.optimisticRoots.excl bs.blck.root
+    dag.forkBlocks.excl(KeyedBlockRef.init(bs.blck))
+    dag.db.delBlock(bs.blck.root)
+
 proc pruneBlocksDAG(dag: ChainDAGRef) =
   ## This prunes the block DAG
   ## This does NOT prune the cached state checkpoints and EpochRef
@@ -1386,18 +1398,7 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
       "finalizedHead parent should have been pruned from memory already"
 
     while cur.blck.parent != nil:
-      dag.optimisticRoots.excl cur.blck.root
-
-      # TODO: should we move that disk I/O to `onSlotEnd`
-      dag.delState(cur.toBlockSlotId().expect("not nil"))
-
-      if cur.isProposed():
-        # Update light client data
-        dag.deleteLightClientData(cur.blck.bid)
-
-        dag.forkBlocks.excl(KeyedBlockRef.init(cur.blck))
-        dag.db.delBlock(cur.blck.root)
-
+      dag.pruneBlockSlot(cur)
       cur = cur.parentOrSlot
 
     dag.heads.del(n)
@@ -1435,18 +1436,8 @@ proc markBlockInvalid*(dag: ChainDAGRef, root: Eth2Digest) =
     doAssert verifyFinalization notin dag.updateFlags
     return
 
-  dag.optimisticRoots.excl root
-
   debug "markBlockInvalid"
-
-  # TODO copied from inner loop in pruneBlocksDAG, could refactor
-  dag.delState(blck.bid.atSlot())
-
-  # TODO These are gated by cur.isProposed in pruneBlocksDAG, but here should
-  # not be necessary or useful
-  dag.deleteLightClientData(blck.bid)
-  dag.forkBlocks.excl(KeyedBlockRef.init(blck))
-  dag.db.delBlock(root)
+  dag.pruneBlockSlot(blck.atSlot())
 
 proc markBlockVerified*(
     dag: ChainDAGRef, quarantine: var Quarantine, root: Eth2Digest) =
@@ -1459,8 +1450,7 @@ proc markBlockVerified*(
     return
   logScope: blck = shortLog(cur)
 
-  debug "markBlockVerified",
-    blck = shortLog(cur)
+  debug "markBlockVerified"
 
   while true:
     if not dag.is_optimistic(cur.bid.root):
@@ -1468,8 +1458,7 @@ proc markBlockVerified*(
 
     dag.optimisticRoots.excl cur.bid.root
 
-    debug "markBlockVerified ancestor",
-      blck = shortLog(cur)
+    debug "markBlockVerified ancestor"
 
     if cur.parent.isNil:
       break
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1441,8 +1441,8 @@ proc markBlockInvalid*(dag: ChainDAGRef, root: Eth2Digest) =
 
 proc markBlockVerified*(
     dag: ChainDAGRef, quarantine: var Quarantine, root: Eth2Digest) =
-  # Might be called when block was never optimistic to begin with. In that
-  # circumstance, bail early.
+  # Might be called when block was not optimistic to begin with, or had been
+  # but already had been marked verified.
   if not dag.is_optimistic(root):
     return
 
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -400,27 +400,19 @@ proc newExecutionPayload*(
     debug "newPayload failed", msg = err.msg
     return PayloadExecutionStatus.syncing
 
-from ../consensus_object_pools/blockchain_dag import
-  getBlockRef, loadExecutionBlockRoot, markBlockInvalid
+from ../consensus_object_pools/blockchain_dag import is_optimistic
 
 # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#helpers
 proc is_optimistic_candidate_block(
     self: BlockProcessor, blck: ForkedSignedBeaconBlock): bool =
-  # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#when-to-optimistically-import-blocks
-  # The current slot (as per the system clock) is at least
-  # `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` ahead of the slot of the block being
-  # imported.
-  if  blck.slot + self.safeSlotsToImportOptimistically <=
-      self.getBeaconTime().slotOrZero:
-    return true
+  let parent = withBlck(blck): blck.message.parent_root
 
-  let
-    parentRoot = withBlck(blck): blck.message.parent_root
-    parentBlck = self.consensusManager.dag.getBlockRef(parentRoot).valueOr:
-      return false
+  # Only blocks with nonzero execution payloads enter the opt store.
+  self.consensusManager.dag.is_optimistic(parent) or
+    blck.slot + self.safeSlotsToImportOptimistically <=
+      self.getBeaconTime().slotOrZero
 
-  # The parent of the block has execution enabled.
-  not self.consensusManager.dag.loadExecutionBlockRoot(parentBlck).isZero
+from ../consensus_object_pools/blockchain_dag import markBlockInvalid
 
 proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
   while true:
@@ -498,6 +490,5 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
       else:
         debug "runQueueProcessingLoop: block cannot be optimistically imported",
           blck = shortLog(blck.blck)
-        if not blck.resfut.isNil:
-          blck.resfut.complete(
-            Result[void, BlockError].err(BlockError.MissingParent))
+        blck.resfut.complete(
+          Result[void, BlockError].err(BlockError.MissingParent))
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -164,7 +164,6 @@ proc storeBackfillBlock(
 
   res
 
-from chronicles/chronos_tools import traceAsyncErrors
 from ../consensus_object_pools/attestation_pool import addForkChoice
 from ../consensus_object_pools/spec_cache import get_attesting_indices
 from ../spec/datatypes/phase0 import TrustedSignedBeaconBlock
@@ -173,7 +172,6 @@ proc storeBlock*(
     self: ref BlockProcessor,
     src: MsgSource, wallTime: BeaconTime,
     signedBlock: ForkySignedBeaconBlock, payloadValid: bool,
-    blockingHeadUpdate: bool,
     queueTick: Moment = Moment.now(),
     validationDur = Duration()):
     Future[Result[BlockRef, BlockError]] {.async.} =
@@ -256,16 +254,21 @@ proc storeBlock*(
 
   # Eagerly update head: the incoming block "should" get selected.
   #
-  # storeBlock gets called from validator_duties and from the block processing
-  # loop. The block processing loop can tolerate relatively longer delays. The
-  # validator_duties context is after a block being proposed; storeBlock is in
-  # a tail call position, so it's comparatively safe to wait here.
-  if blockingHeadUpdate:
-    await self.consensusManager.updateHeadWithExecution(
-      wallTime.slotOrZero)
+  # storeBlock gets called from validator_duties, which depends on its not
+  # blocking progress any longer than necessary, and processBlock here, in
+  # which case it's fine to await for a while on engine API results.
+  if not is_execution_block(signedBlock.message):
+    self.consensusManager[].updateHead(wallTime.slotOrZero)
   else:
-    traceAsyncErrors self.consensusManager.updateHeadWithExecution(
-      wallTime.slotOrZero)
+    # This primarily exists to ensure that by the time the DAG updateHead is
+    # called valid blocks have already been registered as verified. The head
+    # can lag a slot behind wall clock, complicating detecting synced status
+    # for validating, otherwise.
+    #
+    # TODO have a third version which is fire-and-forget for when it is merge
+    # but payloadValid is true, i.e. fcU is for EL's benefit, not CL. Current
+    # behavior adds unnecessary latency to CL event loop.
+    await self.consensusManager.updateHeadWithExecution(wallTime.slotOrZero)
 
   let
     updateHeadTick = Moment.now()
@@ -340,8 +343,8 @@ proc processBlock(
 
   let res = withBlck(entry.blck):
     await self.storeBlock(
-      entry.src, wallTime, blck, payloadValid = payloadValid,
-      blockingHeadUpdate = true, entry.queueTick, entry.validationDur)
+      entry.src, wallTime, blck, payloadValid, entry.queueTick,
+      entry.validationDur)
 
   if entry.resfut != nil:
     entry.resfut.complete(
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -164,6 +164,7 @@ proc storeBackfillBlock(
 
   res
 
+from chronicles/chronos_tools import traceAsyncErrors
 from ../consensus_object_pools/attestation_pool import addForkChoice
 from ../consensus_object_pools/spec_cache import get_attesting_indices
 from ../spec/datatypes/phase0 import TrustedSignedBeaconBlock
@@ -172,6 +173,7 @@ proc storeBlock*(
     self: ref BlockProcessor,
     src: MsgSource, wallTime: BeaconTime,
     signedBlock: ForkySignedBeaconBlock, payloadValid: bool,
+    blockingHeadUpdate: bool,
     queueTick: Moment = Moment.now(),
     validationDur = Duration()):
     Future[Result[BlockRef, BlockError]] {.async.} =
@@ -254,21 +256,16 @@ proc storeBlock*(
 
   # Eagerly update head: the incoming block "should" get selected.
   #
-  # storeBlock gets called from validator_duties, which depends on its not
-  # blocking progress any longer than necessary, and processBlock here, in
-  # which case it's fine to await for a while on engine API results.
-  if not is_execution_block(signedBlock.message):
-    self.consensusManager[].updateHead(wallTime.slotOrZero)
+  # storeBlock gets called from validator_duties and from the block processing
+  # loop. The block processing loop can tolerate relatively longer delays. The
+  # validator_duties context is after a block being proposed; storeBlock is in
+  # a tail call position, so it's comparatively safe to wait here.
+  if blockingHeadUpdate:
+    await self.consensusManager.updateHeadWithExecution(
+      wallTime.slotOrZero)
   else:
-    # This primarily exists to ensure that by the time the DAG updateHead is
-    # called valid blocks have already been registered as verified. The head
-    # can lag a slot behind wall clock, complicating detecting synced status
-    # for validating, otherwise.
-    #
-    # TODO have a third version which is fire-and-forget for when it is merge
-    # but payloadValid is true, i.e. fcU is for EL's benefit, not CL. Current
-    # behavior adds unnecessary latency to CL event loop.
-    await self.consensusManager.updateHeadWithExecution(wallTime.slotOrZero)
+    traceAsyncErrors self.consensusManager.updateHeadWithExecution(
+      wallTime.slotOrZero)
 
   let
     updateHeadTick = Moment.now()
@@ -343,8 +340,8 @@ proc processBlock(
 
   let res = withBlck(entry.blck):
     await self.storeBlock(
-      entry.src, wallTime, blck, payloadValid, entry.queueTick,
-      entry.validationDur)
+      entry.src, wallTime, blck, payloadValid = payloadValid,
+      blockingHeadUpdate = true, entry.queueTick, entry.validationDur)
 
   if entry.resfut != nil:
     entry.resfut.complete(
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -8,20 +8,21 @@
 {.push raises: [Defect].}
 
 import
-  std/math,
   stew/results,
   chronicles, chronos, metrics,
-  eth/async_utils,
-  web3/engine_api_types,
-  ../spec/datatypes/[phase0, altair, bellatrix],
-  ../spec/[forks, signatures_batch],
-  ../consensus_object_pools/[
-    attestation_pool, block_clearance, blockchain_dag, block_quarantine,
-    spec_cache],
-  ../eth1/eth1_monitor,
-  ./consensus_manager,
-  ../beacon_clock,
+  ../spec/signatures_batch,
   ../sszdump
+
+from ./consensus_manager import
+  ConsensusManager, updateHead, updateHeadWithExecution
+from ../beacon_clock import GetBeaconTimeFn, toFloatSeconds
+from ../consensus_object_pools/block_dag import BlockRef, root, slot
+from ../consensus_object_pools/block_pools_types import BlockError, EpochRef
+from ../consensus_object_pools/block_quarantine import
+  addOrphan, addUnviable, pop, removeOrphan
+from ../validators/validator_monitor import
+  MsgSource, ValidatorMonitor, registerAttestationInBlock, registerBeaconBlock,
+  registerSyncAggregateInBlock
 
 export sszdump, signatures_batch
 
@@ -63,6 +64,7 @@ type
     dumpEnabled: bool
     dumpDirInvalid: string
     dumpDirIncoming: string
+    safeSlotsToImportOptimistically: uint16
 
     # Producers
     # ----------------------------------------------------------------
@@ -92,7 +94,8 @@ proc new*(T: type BlockProcessor,
           rng: ref HmacDrbgContext, taskpool: TaskPoolPtr,
           consensusManager: ref ConsensusManager,
           validatorMonitor: ref ValidatorMonitor,
-          getBeaconTime: GetBeaconTimeFn): ref BlockProcessor =
+          getBeaconTime: GetBeaconTimeFn,
+          safeSlotsToImportOptimistically: uint16): ref BlockProcessor =
   (ref BlockProcessor)(
     dumpEnabled: dumpEnabled,
     dumpDirInvalid: dumpDirInvalid,
@@ -101,6 +104,7 @@ proc new*(T: type BlockProcessor,
     consensusManager: consensusManager,
     validatorMonitor: validatorMonitor,
     getBeaconTime: getBeaconTime,
+    safeSlotsToImportOptimistically: safeSlotsToImportOptimistically,
     verifier: BatchVerifier(rng: rng, taskpool: taskpool)
   )
 
@@ -131,6 +135,9 @@ proc dumpBlock[T](
     else:
       discard
 
+from ../consensus_object_pools/block_clearance import
+  addBackfillBlock, addHeadBlock
+
 proc storeBackfillBlock(
     self: var BlockProcessor,
     signedBlock: ForkySignedBeaconBlock): Result[void, BlockError] =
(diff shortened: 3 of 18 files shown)


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -81,8 +83,13 @@ proc addResolvedHeadBlock(
   # Update light client data
   dag.processNewBlockForLightClient(state, trustedBlock, parent.bid)
 
+  if not payloadValid:
+    dag.optimisticRoots.incl blockRoot
```
etan-status: `not payloadValid` could mean both "unverified" and "invalid". Would `isOptimistic` be a better name over `payloadValid`? The DAG doesn't seem to have a concept of "payload" in general, but it knows about the concept of "optimistic"

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -256,12 +264,21 @@ proc addHeadBlock*(
...
     sigVerifyDur = sigVerifyTick - stateDataTick,
     stateVerifyDur = stateVerifyTick - sigVerifyTick)
 
+proc addHeadBlock*(
+    dag: ChainDAGRef, verifier: var BatchVerifier,
+    signedBlock: ForkySignedBeaconBlock,
+    onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded |
+                  OnBellatrixBlockAdded
+    ): Result[BlockRef, BlockError] =
+  addMaybeOptHeadBlock(dag, verifier, signedBlock, true, onBlockAdded)
```
etan-status:
```suggestion
  addMaybeOptHeadBlock(
    dag, verifier, signedBlock, payloadValid = true, onBlockAdded)
```
Named parameter for `bool` improves readability.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+    blck = shortLog(blck.get)
+
+  # TODO copied from inner loop in pruneBlocksDAG, could refactor
+  dag.delState(blck.get.bid.atSlot())
+
+  # TODO These are gated by cur.isProposed in pruneBlocksDAG, but here should
+  # not be necessary or useful
+  dag.deleteLightClientData(blck.get.bid)
+  dag.forkBlocks.excl(KeyedBlockRef.init(blck.get))
+  dag.db.delBlock(root)
```
```suggestion
  let blck = dag.getBlockRef(root).valueOr:
    return
  logScope: blck = shortLog(blck)

  if root notin dag.optimisticRoots:
    warn "markPayloadInvalid: attempt to invalidate valid block"
    return

  dag.optimisticRoots.excl root

  debug "markPayloadInvalid"

  # TODO copied from inner loop in pruneBlocksDAG, could refactor
  dag.delState(blck.bid.atSlot())

  # TODO These are gated by cur.isProposed in pruneBlocksDAG, but here should
  # not be necessary or useful
  dag.deleteLightClientData(blck.bid)
  dag.forkBlocks.excl(KeyedBlockRef.init(blck))
  dag.db.delBlock(root)
```
Using `valueOr` and `logScope` makes this a bit more concise.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+    warn "markPayloadInvalid: attempt to invalidate valid block",
+      blck = shortLog(blck.get)
+    return
+
+  dag.optimisticRoots.excl root
+
+  debug "markPayloadInvalid",
+    blck = shortLog(blck.get)
+
+  # TODO copied from inner loop in pruneBlocksDAG, could refactor
```
etan-status: The `dag.optimisticRoots.excl root` above is "copied" as well.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
     dagPruneDur = Moment.now() - startTick
 
+proc markPayloadInvalid*(dag: ChainDAGRef, root: Eth2Digest) =
+  let blck = dag.getBlockRef(root)
+  if blck.isNone:
+    return
+
+  if root notin dag.optimisticRoots:
+    warn "markPayloadInvalid: attempt to invalidate valid block",
+      blck = shortLog(blck.get)
```
etan-status: During initialization, only `finalizedHead` and `head` are populated into `dag.optimisticRoots`. Could it happen that this function is called on one of the intermediate blocks? Is this a condition that is expected in edge cases, or is this a programming error? If the latter, a `doAssert verifyFinalization notin dag.updateFlags` could be useful to detect this being hit.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
     prunedHeads = hlen - dag.heads.len,
     dagPruneDur = Moment.now() - startTick
 
+proc markPayloadInvalid*(dag: ChainDAGRef, root: Eth2Digest) =
```
etan-status: `markBlockInvalid` may be more accurate

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+  # TODO copied from inner loop in pruneBlocksDAG, could refactor
+  dag.delState(blck.get.bid.atSlot())
+
+  # TODO These are gated by cur.isProposed in pruneBlocksDAG, but here should
+  # not be necessary or useful
+  dag.deleteLightClientData(blck.get.bid)
+  dag.forkBlocks.excl(KeyedBlockRef.init(blck.get))
+  dag.db.delBlock(root)
+
+proc markPayloadValid*(
```
etan-status: `markBlockValid` may be better. The spec wording talks about blocks, e.g.,
> When a block transitions from NOT_VALIDATED -> VALID, all ancestors of the block MUST also transition from NOT_VALIDATED -> VALID. Such a block and any previously NOT_VALIDATED ancestors are no longer considered "optimistically imported".

Or, alternatively, `markBlockVerified` to match the terminology of the `latest_verified_ancestor` helper function.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+    debug "markPayloadInvalid ancestor",
+      blck = shortLog(cur)
+
+    # This should never have been pruned, since finalization should never occur
+    # optimistically.
+    doAssert not cur.parent.isNil
+
+    cur = cur.parent
+
+  doAssert false
```
```suggestion
proc markPayloadValid*(
    dag: ChainDAGRef, quarantine: var Quarantine, root: Eth2Digest) =
  # Might be called when payload was never optimistic to begin with. In that
  # circumstance, bail early.
  if root notin dag.optimisticRoots:
    return

  var cur = dag.getBlockRef(root).valueOr:
    return
  logScope: blck = shortLog(cur)

  debug "markPayloadValid"

  while true:
    if cur.bid.root notin dag.optimisticRoots:
      return

    dag.optimisticRoots.excl cur.bid.root

    debug "markPayloadValid ancestor"

    # This should never have been pruned, since finalization should never occur
    # optimistically.
    doAssert not cur.parent.isNil

    cur = cur.parent

  doAssert false
```
`valueOr`, `logScope`, and fix `markPayloadInvalid` mention.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1400,6 +1408,63 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+      return
+
+    dag.optimisticRoots.excl cur.bid.root
+
+    debug "markPayloadInvalid ancestor",
+      blck = shortLog(cur)
+
+    # This should never have been pruned, since finalization should never occur
+    # optimistically.
+    doAssert not cur.parent.isNil
```
etan-status: On startup, `dag.finalizedHead` is marked as `optimistic`. This condition only usually holds because in most cases, the merge happens after genesis, so, there are multiple blocks between `dag.finalizedHead` and `dag.head`. Not sure if it is bulletproof though.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1539,9 +1604,42 @@ template getHeadStateMergeComplete*(dag: ChainDAGRef): bool =
     else:
       false
 
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#helpers
+func latest_verified_ancestor(
+    dag: ChainDAGRef, optHead: BlockRef): BlockRef =
+  doAssert not optHead.isNil
+  # TODO this, while probably not being a bottleneck, due to a tendency for a
+  # highly optimistic CL to be waiting for the EL, is still a de-facto O(n^2)
```
etan-status: or O(n*log(n)) if the `HashSet` lookup is `log(n)`.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1539,9 +1604,42 @@ template getHeadStateMergeComplete*(dag: ChainDAGRef): bool =
...
+    if cur.bid.root notin dag.optimisticRoots:
+      return cur
+
+    # This should never be pruned, because finalization should never occur
+    # optimistically.
+    doAssert not cur.parent.isNil
+
+    cur = cur.parent
+
+  doAssert false
```
etan-status: On testnet that has merge happening from beginning, this could go all the way back to Genesis, no?
Or at least, all the way back to finalized head.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -157,21 +164,27 @@ proc storeBackfillBlock(
...
   ## to know about it, such as the fork choice and the monitoring
+  template attestationPool(): auto = self.consensusManager.attestationPool
   let
-    attestationPool = self.consensusManager.attestationPool
     startTick = Moment.now()
     wallSlot = wallTime.slotOrZero()
-    vm = self.validatorMonitor
-    dag = self.consensusManager.dag
+  template vm(): auto = self.validatorMonitor
+  template dag(): auto = self.consensusManager.dag
```
etan-status: Would this lead to multiple pointer loads and going through nil checks every time `dag` is accessed, or is this cached? The variables are only pointer-sized.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -157,21 +164,27 @@ proc storeBackfillBlock(
...
-    validationDur = Duration()): Result[BlockRef, BlockError] =
+    signedBlock: ForkySignedBeaconBlock, payloadValid: bool,
+    queueTick: Moment = Moment.now(),
+    validationDur = Duration()):
+    Future[Result[BlockRef, BlockError]] {.async.} =
   ## storeBlock is the main entry point for unvalidated blocks - all untrusted
   ## blocks, regardless of origin, pass through here. When storing a block,
   ## we will add it to the dag and pass it to all block consumers that need
   ## to know about it, such as the fork choice and the monitoring
+  template attestationPool(): auto = self.consensusManager.attestationPool
```
```suggestion
```

beacon_chain/gossip_processing/block_processor.nim
```suggestion
  template vm(): auto = self[].validatorMonitor
  template dag(): auto = self[].consensusManager[].dag
  template attestationPool(): auto = self[].consensusManager[].attestationPool
```
Moving down the attestationPool groups those templates together.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -239,8 +252,23 @@ proc storeBlock*(
 
   let storeBlockTick = Moment.now()
 
-  # Eagerly update head: the incoming block "should" get selected
-  self.consensusManager[].updateHead(wallTime.slotOrZero)
+  # Eagerly update head: the incoming block "should" get selected.
+  #
+  # storeBlock gets called from validator_duties, which depends on its not
```
etan-status, line 257:
```suggestion
  # storeBlock gets called from validator_duties, which depends on it not
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -157,21 +164,27 @@ proc storeBackfillBlock(
 
   res
 
+from ../consensus_object_pools/attestation_pool import addForkChoice
+from ../consensus_object_pools/spec_cache import get_attesting_indices
+from ../spec/datatypes/phase0 import TrustedSignedBeaconBlock
+
 proc storeBlock*(
-    self: var BlockProcessor,
+    self: ref BlockProcessor,
```
etan-status, line 172: Not sure what triggers the need for `ref` here.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -133,9 +140,10 @@ proc advanceClearanceState*(dag: ChainDAGRef) =
     debug "Prepared clearance state for next block",
       next, updateStateDur = Moment.now() - startTick
 
-proc addHeadBlock*(
+proc addMaybeOptHeadBlock*(
```
etan-status: Should be fine to keep as `addHeadBlock`, it still has the extra argument to distinguish it from the non-optimistic overload.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -239,8 +252,23 @@ proc storeBlock*(
...
+  # Eagerly update head: the incoming block "should" get selected.
+  #
+  # storeBlock gets called from validator_duties, which depends on its not
+  # blocking progress any longer than necessary, and processBlock here, in
+  # which case it's fine to await for a while on engine API results.
+  if not is_execution_block(signedBlock.message):
+    self.consensusManager[].updateHead(wallTime.slotOrZero)
+  else:
+    # This primarily exists to ensure that by the time the DAG updateHead is
+    # called, known-valid it has already been registered. Otherwise the head
```
```suggestion
    # called, known-valid has already been registered. Otherwise the head
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -239,8 +252,23 @@ proc storeBlock*(
...
+    # can lag a slot behind reality, which creates problems in detecting the
```
etan-status: `reality` being wall clock?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -311,59 +341,29 @@ proc processBlock(self: var BlockProcessor, entry: BlockEntry) =
...
+from web3/engine_api_types import PayloadExecutionStatus, PayloadStatusV1
+from ../eth1/eth1_monitor import
+  Eth1Monitor, asEngineExecutionPayload, ensureDataProvider, newPayload
+from ../spec/datatypes/bellatrix import ExecutionPayload, SignedBeaconBlock
 
 proc newExecutionPayload*(
     eth1Monitor: Eth1Monitor, executionPayload: bellatrix.ExecutionPayload):
     Future[PayloadExecutionStatus] {.async.} =
+  if eth1Monitor.isNil:
+    info "newPayload: attempting to process execution payload without an Eth1Monitor. Ensure --web3-url setting is correct."
```
```suggestion
    warn "newPayload: attempting to process execution payload without an Eth1Monitor. Ensure --web3-url setting is correct."
```
`warn` as this is a misconfiguration

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -394,22 +389,39 @@ proc newExecutionPayload*(
...
 
-proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
-  # Don't want to vacillate between "optimistic" sync and non-optimistic
-  # sync heads. Relies on runQueueProcessingLoop() being the only place,
-  # in Nimbus, which does this.
-  var
-    optForkchoiceHeadSlot = GENESIS_SLOT # safe default
-    optForkchoiceHeadRoot: Eth2Digest
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#helpers
+proc is_optimistic_candidate_block(
```
etan-status, line 407:
```suggestion
func is_optimistic_candidate_block(
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -394,22 +389,39 @@ proc newExecutionPayload*(
...
+    self: BlockProcessor, blck: ForkedSignedBeaconBlock): bool =
```
etan-status, line 408: The function assumes `bellatrix`, which will make this crash if ever called on a pre-bellatrix block, and also in the post-capella future.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -394,22 +389,39 @@ proc newExecutionPayload*(
...
+  # Use presence of parent in opt sync roots as `is_execution_block` proxy;
```
etan-status: Implementation doesn't check the parent, but instead checks the block itself.
Also, doesn't this break because on init only the head and finalizedHead is marked optimistic but not its parent?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -424,112 +436,55 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
...
-            except CatchableError as err:
-              debug "runQueueProcessingLoop: newPayload failed",
-                err = err.msg
-              PayloadExecutionStatus.syncing
-          else:
-            debug "runQueueProcessingLoop: got execution payload before TTD"
-            PayloadExecutionStatus.syncing
+    let blck = await self[].blockQueue.popFirst()
+
+    static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
```
etan-status: Not sure if there is a way to realistically avoid the usual `withBlck(blck)`...`when stateFork >= BeaconStateFork.Bellatrix` for these in the future. Capella will likely change the memory layout of the block / state once more, so separate codegen is needed.

beacon_chain/spec/helpers.nim
```diff
@@ -599,10 +599,17 @@ func is_merge_transition_complete*(state: bellatrix.BeaconState): bool =
...
+          altair.BeaconBlock | altair.TrustedBeaconBlock |
+          altair.SigVerifiedBeaconBlock): bool =
+  false
+
+func is_execution_block*(
+    blck: bellatrix.BeaconBlock | bellatrix.TrustedBeaconBlock |
+          bellatrix.SigVerifiedBeaconBlock): bool =
   const defaultBellatrixExecutionPayload = default(bellatrix.ExecutionPayload)
-  body.execution_payload != defaultBellatrixExecutionPayload
+  blck.body.execution_payload != defaultBellatrixExecutionPayload
```
etan-status:
```
func is_execution_block*(blck: SomeForkyBeaconBlock): bool =
  when typeof(blck).toFork >= BeaconBlockFork.Bellatrix:
    not blck.body.execution_payload.isZeroMemory
  else:
    false
```

beacon_chain/validators/validator_duties.nim
```diff
@@ -207,10 +207,11 @@ proc isSynced*(node: BeaconNode, head: BlockRef): bool =
   # TODO if everyone follows this logic, the network will not recover from a
   #      halt: nobody will be producing blocks because everone expects someone
   #      else to do it
-  if wallSlot.afterGenesis and head.slot + node.config.syncHorizon < wallSlot.slot:
+  if  wallSlot.afterGenesis and
+      head.slot + node.config.syncHorizon < wallSlot.slot:
     false
   else:
-    true
+    head.root notin node.dag.optimisticRoots
```
etan-status: An `isOptimistic` helper template in `blockchain_dag.nim` would help avoid leaking the implementation detail of this `optimisticRoots` set. Other modules should not need to have to deal with the set directly.

beacon_chain/validators/validator_duties.nim
```diff
@@ -1510,8 +1512,8 @@ proc sendBeaconBlock*(node: BeaconNode, forked: ForkedSignedBeaconBlock
   let
     wallTime = node.beaconClock.now()
     accepted = withBlck(forked):
-      let newBlockRef = node.blockProcessor[].storeBlock(
-        MsgSource.api, wallTime, blck)
+      let newBlockRef = await node.blockProcessor.storeBlock(
+        MsgSource.api, wallTime, blck, true)
```
etan-status:
```suggestion
        MsgSource.api, wallTime, blck, payloadValid  = true)
```

tests/test_block_processor.nim
```diff
@@ -44,11 +44,11 @@ suite "Block processor" & preset():
       getTimeFn = proc(): BeaconTime = b2.message.slot.start_beacon_time()
       processor = BlockProcessor.new(
         false, "", "", keys.newRng(), taskpool, consensusManager,
-        validatorMonitor, getTimeFn)
+        validatorMonitor, getTimeFn, 128)
```
etan-status:
```suggestion
        validatorMonitor, getTimeFn, safeSlotsToImportOptimistically = 128)
```

tests/test_block_processor.nim
```diff
@@ -44,11 +44,11 @@ suite "Block processor" & preset():
...
 
   test "Reverse order block add & get" & preset():
-    let missing = processor[].storeBlock(
-      MsgSource.gossip, b2.message.slot.start_beacon_time(), b2)
+    let missing = waitFor processor.storeBlock(
+      MsgSource.gossip, b2.message.slot.start_beacon_time(), b2, true)
```
etan-status:
```suggestion
      MsgSource.gossip, b2.message.slot.start_beacon_time(), b2, payloadValid = true)
```

tests/test_block_processor.nim
```diff
@@ -57,8 +57,8 @@ suite "Block processor" & preset():
       FetchRecord(root: b1.root) in quarantine[].checkMissing()
 
     let
-      status = processor[].storeBlock(
-        MsgSource.gossip, b2.message.slot.start_beacon_time(), b1)
+      status = waitFor processor.storeBlock(
+        MsgSource.gossip, b2.message.slot.start_beacon_time(), b1, true)
```
etan-status:
```suggestion
        MsgSource.gossip, b2.message.slot.start_beacon_time(), b1, payloadValid = true)
```

etan-status (commented):
Still a few open `TODO` and `when true` that need to be fixed or logged as issues to avoid losing them.

Overall quite impressive how compact this change is implementation-wise.

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: It's derived from `NOT_VALIDATED` as defined in https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#helpers, i.e. `SYNCING` or `ACCEPTED`. Thus, it's deliberately not `payloadInvalid` being true or false, indeed, and it's intentional that `not payloadValid` could mean both "unverified" and "invalid".

I don't have strong views on exactly what it's called, but `payloadValid` is the direct translation or interpretation of the engine API call results. It's possible that for DAG purposes, it's better to use a different abstraction.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#transitioning-from-valid---invalidated-or-invalidated---valid describes, as far as this warning's topic:
>  These operations are purposefully omitted. It is outside of the scope of the specification since it's only possible with a faulty EE.
>
> Such a scenario requires manual intervention.

It's possible that it'd be better to be even more emphatic than a warning, because this is a failure on par with doppelganger detection. Something went very wrong. That said, while it's not necessarily a Nimbus programming error, it is a programming error creating potential consensus failure, and `verifyFinalization` mode making this more obvious would be useful.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/3c864f0ea41e633c497db0c0d7ea3e32f98d688b

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/3c864f0ea41e633c497db0c0d7ea3e32f98d688b adds `doAssert verifyFinalization notin dag.updateFlags`

beacon_chain/validators/validator_duties.nim
tersec: There's a quirk here that it didn't seem to make sense, at least initially, to have an overall check for whether a node is optimistic as a whole -- it's only optimistic or not with respect to a given block (or block root), and each part of Nimbus has its own view on what the most useful block roots to check with are.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -957,10 +957,19 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
 
   # If these aren't actually optimistic, the first fcU will resolve that
-  # TODO this sort of thing can be resolved via a database schema update
-  # but deliberately separate that
-  dag.optimisticRoots.incl dag.head.root
-  dag.optimisticRoots.incl dag.finalizedHead.blck.root
+  withState(dag.headState):
+    when stateFork >= BeaconStateFork.Bellatrix:
+      # TODO can't use this template due to Nim template symbol resolution but
+      # is fixed in Nim devel (not currently Nim 1.6.7).
```
narimiran: Do you maybe know which commit/PR fixed it, so I can see if it could be backported to 1.6.x?

tersec: Ah, this turned out not to be quite true -- the issue was a bit different than I'd gathered at the time, and also isn't affected by Nim 1.6 vs Nim devel regardless.

tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/e1eff5d8254215f9d5148cb4438b25f686ec86ca shows a working version

beacon_chain/gossip_processing/block_processor.nim
tersec: Using `self.foo` in `{.async.}` code. Will verify it's still necessary. I went through a couple designs, one more callback-heavy than this, before this one, and it might have been an artifact/fallout from that.

beacon_chain/gossip_processing/block_processor.nim
tersec: Yeah, I guess it's a bit imprecise.

beacon_chain/gossip_processing/block_processor.nim
tersec: Yeah, I can revert this, this was part of the `ref` change. But either `ref` isn't necessary anymore or the `ref` change supercedes these smaller tweaks.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: "Verified" vs "Valid" is basically an optimistic sync vs engine API terminology question. Opt sync code should probably use verified rather than valid, since that's the native terminology of the optimistic sync spec. That it's a result of `VALID` engine API return values is an implementation artifact.

beacon_chain/gossip_processing/block_processor.nim
tersec: Both grammatical, but each sentence parses and diagrams differently. Can probably guess why I added the extra letter to that line.

beacon_chain/spec/helpers.nim
tersec: Alright, maybe we don't need both of those, then, but either way, the function called `is_execution_block` saves so little duplication by deviating from the optimistic sync spec it's not worth doing so; https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/sync/optimistic.md#helpers shows
```python
def is_execution_block(block: BeaconBlock) -> bool:
    return block.body.execution_payload != ExecutionPayload()
```

tersec: Also, `isZeroMemory` doesn't necessarily work on composite objects where alignment between members/fields might not be exact. Empirically, I tried applying `isZeroMemory` to various objects before and it wasn't always reliable unless one was dealing with a simple array, primitive type, or equivalent; https://github.com/nim-lang/Nim/issues/19897 discusses this a bit more.

beacon_chain/gossip_processing/block_processor.nim
tersec: The various static asserts are there exactly to trigger before any runtime crashes, but, yes.

beacon_chain/gossip_processing/block_processor.nim
tersec:
```
/home/jenkins/workspace/_nimbus-eth2_multibranch_PR-3793/linux/beacon_chain/gossip_processing/block_processor.nim(404, 6) Error: 'is_optimistic_candidate_block' can have side effects
```
it's the `getBeaconTime()`.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: The O(n^2) is from a CL running this proportionately to once per slot. The O(n^2) isn't for one lookup, it's for multiple calls, and potentially worse, some of those traversals are long indeed.

Arguably, by that standard, any linear-time processing per slot does, and, indeed, mostly the consensus specs avoid those. Epoch state transitions conspicuously do have O(n) operations and they are comparatively slow.

This isn't as much an issue as before, and it might make sense to just not have `latest_verified_ancestor`. It's part of the optimistic sync specs, and it made sense when it was careful not to prune pre-verified finalization, but since it doesn't keep that in general, it can't assume that it can do anything useful with this function, and indeed,
```
    # This should never be pruned, because finalization should never occur                                                           
    # optimistically.
```
is now false. Will remove altogether.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: Yeah, that's the O(n^2) part -- going back to verified finalized checkpoint every time, slot after slot. O(n) slots * O(n) time per slot. Anyway, removing and if something like it is required, can investigate how to do this more efficiently, but it's pretty inconsistent with how this PR works now.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/610890b03c47048846a0de1d042f050746123897 removes `latest_verified_ancestor` entirely.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/610890b03c47048846a0de1d042f050746123897 removes `latest_verified_ancestor`, including this comment.

beacon_chain/spec/helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/610890b03c47048846a0de1d042f050746123897 works beyond Bellatrix, so long as `spec/forks` in general works past it.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/610890b03c47048846a0de1d042f050746123897 is more precise about this being the wall clock.

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/610890b03c47048846a0de1d042f050746123897 changes this back.

Sure, agree -- in the first versions of this PR, there was more use of the verified head, but the current version just has the one DAG head, and it's either optimistic or verified, but it doesn't hold two heads, so there's no point segregating them this way.

Still keeping the overload, as a convenience, for now.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/610890b03c47048846a0de1d042f050746123897 reverts this basically pointless change altogether.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/610890b03c47048846a0de1d042f050746123897 reverts this basically pointless change altogether.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/610890b03c47048846a0de1d042f050746123897 removes `latest_verified_ancestor`, so it's moot.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/610890b03c47048846a0de1d042f050746123897 reverts this.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/610890b03c47048846a0de1d042f050746123897 rephrases that whole comment block

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/8100623f4675283cc259e4238fd66192f9f8ad7c

A consistent nomenclature here is that the engine API deals with "payloads" and "valid", "not valid", and "invalid" statuses.

The optimistic sync side handles "blocks" and doesn't seem to have a special name for "invalid", so that can be reused. "not valid" maps to "optimistic" and "valid" map to verified".

But when possible, it's probably good to avoid mixing nomenclature from engine-API-world with optimistic-sync-world (e.g., "optimistic payload", etc).

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/8100623f4675283cc259e4238fd66192f9f8ad7c renames `markPayloadValid` to `markBlockVerified`, because it's part of the DAG and consistently uses optimistic-sync-language rather than engine-API-language.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/8100623f4675283cc259e4238fd66192f9f8ad7c essentially applies this, having adjusted it for renaming, using `valueOr`, fixing the incorrect log message, and removing the now-incorrect assumption [that was, in previous versions of this PR, true] that there would always be a verified block to traverse back to.

Now it's fine if there just isn't, that it runs out of parents before running out of optimistic blocks, and that's just not a special error condition.

beacon_chain/gossip_processing/block_processor.nim
tersec: Confirmed that this is still an issue:
```
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(183, 23) Error: 'self' is of type <var BlockProcessor> which cannot be captured as it would violate memory safety, declared here: /home/user/nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(172, 5)
```

given the following diff:
```diff
diff --git a/beacon_chain/gossip_processing/block_processor.nim b/beacon_chain/gossip_processing/block_processor.nim
index 8a275c4c..f00f8944 100644
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -169,7 +169,7 @@ from ../consensus_object_pools/spec_cache import get_attesting_indices
 from ../spec/datatypes/phase0 import TrustedSignedBeaconBlock
 
 proc storeBlock*(
-    self: ref BlockProcessor,
+    self: var BlockProcessor,
     src: MsgSource, wallTime: BeaconTime,
     signedBlock: ForkySignedBeaconBlock, payloadValid: bool,
     queueTick: Moment = Moment.now(),
@@ -221,7 +221,7 @@ proc storeBlock*(
             trustedBlock.message.slot, trustedBlock.root,
             state.data.current_sync_committee.pubkeys.data[i])
 
-  self[].dumpBlock(signedBlock, blck)
+  self.dumpBlock(signedBlock, blck)
 
   # There can be a scenario where we receive a block we already received.
   # However this block was before the last finalized epoch and so its parent
@@ -285,7 +285,7 @@ proc storeBlock*(
 
   for quarantined in self.consensusManager.quarantine[].pop(blck.get().root):
     # Process the blocks that had the newly accepted block as parent
-    self[].addBlock(MsgSource.gossip, quarantined)
+    self.addBlock(MsgSource.gossip, quarantined)
 
   return blck
 
@@ -342,7 +342,7 @@ proc processBlock(
     quit 1
 
   let res = withBlck(entry.blck):
-    await self.storeBlock(
+    await self[].storeBlock(
       entry.src, wallTime, blck, payloadValid, entry.queueTick,
       entry.validationDur)
 
diff --git a/beacon_chain/validators/validator_duties.nim b/beacon_chain/validators/validator_duties.nim
index 3c127fd0..accc4e5a 100644
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -807,7 +807,7 @@ proc proposeBlock(node: BeaconNode,
 
       # storeBlock puts the block in the chaindag, and if accepted, takes care
       # of side effects such as event api notification
-      newBlockRef = await node.blockProcessor.storeBlock(
+      newBlockRef = await node.blockProcessor[].storeBlock(
         MsgSource.api, wallTime, signedBlock, true)
 
     if newBlockRef.isErr:
@@ -1512,7 +1512,7 @@ proc sendBeaconBlock*(node: BeaconNode, forked: ForkedSignedBeaconBlock
   let
     wallTime = node.beaconClock.now()
     accepted = withBlck(forked):
-      let newBlockRef = await node.blockProcessor.storeBlock(
+      let newBlockRef = await node.blockProcessor[].storeBlock(
         MsgSource.api, wallTime, blck, payloadValid  = true)
 
       # The boolean we return tells the caller whether the block was integrated
```

tersec:
https://github.com/status-im/nimbus-eth2/runs/7121625887?check_suite_focus=true
```
2022-06-29T22:46:59.6510280Z Error: unhandled exception: over- or underflow [OverflowError]
2022-06-29T22:46:59.6830095Z make: *** [nimbus_beacon_node] Error 1
2022-06-29T22:46:59.6830830Z Makefile:270: recipe for target 'nimbus_beacon_node' failed
2022-06-29T22:46:59.6836911Z make: *** Waiting for unfinished jobs....
```

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/93c9b38ddc35e70122027f4fd280a4635e4cd686 removes specific the most fragile Bellatrix-dependence. `newPayload` still only is implemented for Bellatrix, but that's more reasonable, since it's more intrinsically fork-dependent.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/93c9b38ddc35e70122027f4fd280a4635e4cd686 removes all the explicit `bellatrixData` references from `block_processor`.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/93c9b38ddc35e70122027f4fd280a4635e4cd686 replaces this aspect of `is_optimistic_candidate_block` entirely:
```nim
proc is_optimistic_candidate_block(
    self: BlockProcessor, blck: ForkedSignedBeaconBlock): bool =
  let parent = withBlck(blck): blck.message.parent_root

  # Only blocks with nonzero execution payloads enter the opt store.
  self.consensusManager.dag.is_optimistic(parent) or
    blck.slot + self.safeSlotsToImportOptimistically <=
      self.getBeaconTime().slotOrZero
```

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: Switched this to `blockVerified`, because the DAG should deal in blocks and verified vs optimistic vs invalid, not valid vs invalidated vs not_validated.

tersec:
> Still a few open `TODO` and `when true` that need to be fixed or logged as issues to avoid losing them.
> 
> Overall quite impressive how compact this change is implementation-wise.

There's only one `when true`, and that it shows up in this PR artifact of refactoring the block processing loop. It was already there, and this PR isn't going to fix it. It's related to TTD detection functionality.

This PR now removes significantly more `TODO`s than it adds:
```
$ git diff -U0 f08f9e2bd7e1ba9fe55b02272189134871769031 | grep TODO
+  # TODO rename this to get_optimistic_head
+    # TODO have a third version which is fire-and-forget for when it is merge
+         # TODO self.consensusManager.eth1Monitor.terminalBlockHash.isSome
+         # TODO detect have-TTD-but-not-is_execution_block case, and where
-          # TODO self.consensusManager.eth1Monitor.terminalBlockHash.isSome
-    # TODO (cheatfate): Proper implementation required
-    # TODO (cheatfate): Proper implementation required
-    # TODO (cheatfate): Proper implementation required
-      # TODO (cheatfate): Proper implementation required
-          # TODO (cheatfate): Proper implementation required
-      # TODO (cheatfate): Proper implementation required.
-      # TODO (cheatfate): Proper implementation required.
-    # TODO (cheatfate): Proper implementation required.
-    # TODO (cheatfate): Proper implementation required
-    # TODO (cheatfate): Proper implementation required
```

These rest are pre-existing and/or out of scope/deliberately not part of this PR.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3793/commits/56d8bb18f60d29b5aa7b5a79b9fc10039e8db1cc refactors this

tersec:
Tested successfully with syncs from scratch on Ropsten with each of Erigon, Besu, Geth, and Nethermind.

beacon_chain/gossip_processing/consensus_manager.nim
```diff
@@ -80,24 +80,124 @@ proc expectBlock*(self: var ConsensusManager, expectedSlot: Slot): Future[bool]
 
   return fut
 
+from eth/async_utils import awaitWithTimeout
```
zah, line 83: I recently had a discussion with a D programmer who promoted this style of function-local imports. I guess your rationale is the same - to make it easy to copy/paste functions to a different module by keeping their list of imports attached to the function itself.

One potential risk of this approach is that it also makes it much easier to accidentally introduce an import cycle while moving the function which will result in the typical hard-to-decipher error messages brought by cyclic imports.

tersec: Sure, it helps for that purpose.

Other advantages:
- Nim 1.2 unused import detection is essentially broken, and Nim 1.6 is only somewhat better. This results in import lists simply accumulating because there's a 1,000-line module with a bunch of imports at the top, with no hint of what they're importing or why. Enough modules have some side-effect during import to disable the unused import warning, too. This makes it clearer that for modules imported not for their side effects (such as tests) if the imported symbols don't match, then the module import isn't necessary. It's not 100% automated, but it's better than just letting imports accumulate, as well as in theory providing Nim with more information for detecting unused imports. The more local and specific the imports, both in the symbols they name and where they're used, the better for this purpose.

It's related to one of the reasons I tend to unexport symbols when I notice them: it's not because of some OOP-thing about data hiding, it's about allowing Nim to detect dead code and, when reading or refactoring code, being confident that (unless there's an `export` statement with said symbol) changes in that code will only directly affect that one module, and that one doesn't have to scan the entire project. The deeper in the libraries one is working, the more this applies.

- Various surprising overloads can bubble up from imported modules and cause surprising errors of their own (e.g., where
```nim
withState(someState):
  template f(): state.foo
  f()
```
worked right until some irrelevant-to-purpose symbol from deep inside Chronos's HTTP server was imported transitively, and meant that produced a confusing error message, requiring `template f(): state().foo`. Easy fix, a while to find the exact culprit.

tersec: If this interferes with Nim's cyclic import detection, IMO that's a Nim issue.

zah: Cyclic imports are legal in Nim, they just produce hard to diagnose error messages at times.
While I have no strong objections against this style, my comment meant to serve as a warning that one has to be careful with the notion that individual functions are now easy to move between modules if their imports are structured in this way.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -157,11 +164,17 @@ proc storeBackfillBlock(
...
 proc storeBlock*(
-    self: var BlockProcessor,
+    self: ref BlockProcessor,
     src: MsgSource, wallTime: BeaconTime,
-    signedBlock: ForkySignedBeaconBlock, queueTick: Moment = Moment.now(),
-    validationDur = Duration()): Result[BlockRef, BlockError] =
+    signedBlock: ForkySignedBeaconBlock, payloadValid: bool,
+    queueTick: Moment = Moment.now(),
+    validationDur = Duration()):
+    Future[Result[BlockRef, BlockError]] {.async.} =
```
arnetheduck, line 177: this design in which storeblock becomes async is a bit unfortunate - ie the processes that call it are not necessarily well prepared to lose their current in-memory state for long periods of time, specially not while waiting for a (potentially slow) EL response (since they're part of validator duties)

beacon_chain/gossip_processing/block_processor.nim
arnetheduck: in `async`, `var` is not possible and non-var parameters are _copied_ silently - if you passed a `processor: BlockProcessor` in here, it would happily copy the whole thing, queues and all. I *think* in 1.6 you can do a trick with `copy=` to prevent this but I'm not sure .. cc @zah

arnetheduck: said copying is something we might have to address for blocks, turning them into `ref` potentially - I really don't like that because they become mutable, but nim offers no good solutions here, and the performance cost might outweigh the downsides at this point

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -394,22 +389,40 @@ proc newExecutionPayload*(
...
     return PayloadExecutionStatus.syncing
 
-proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
-  # Don't want to vacillate between "optimistic" sync and non-optimistic
-  # sync heads. Relies on runQueueProcessingLoop() being the only place,
-  # in Nimbus, which does this.
-  var
-    optForkchoiceHeadSlot = GENESIS_SLOT # safe default
-    optForkchoiceHeadRoot: Eth2Digest
+from ../consensus_object_pools/blockchain_dag import
```
arnetheduck, line 403: funny because in refactoring, I've been going in the opposite direction of exporting stuff much more consistently to _avoid_ long import lists - this helps keep symbols that should be visible, visible - the template problem with the wrong symbol being matched is annoying but can often be solved by turning the template into a `proc` or generally giving things more descriptive names.. it's a mess, but with exports, it's actually easier to stay away from redundant imports (because there's simply fewer of them that can go stale)

beacon_chain/gossip_processing/consensus_manager.nim
arnetheduck: > Cyclic imports are legal in Nim, they just produce hard to diagnose error messages at times.

they also produce invalid code with overly broad arguments, where the exact function that actually matches ends up depending on frivolous things like import order - there was no end to ssz import issues until all cyclical imports were removed - this is one more nim feature that sounds good on paper but is worthless practically because it's broken and untested.

beacon_chain/validators/validator_duties.nim
```diff
@@ -1512,8 +1515,8 @@ proc sendBeaconBlock*(node: BeaconNode, forked: ForkedSignedBeaconBlock
   let
     wallTime = node.beaconClock.now()
     accepted = withBlck(forked):
-      let newBlockRef = node.blockProcessor[].storeBlock(
-        MsgSource.api, wallTime, blck)
+      let newBlockRef = await node.blockProcessor.storeBlock(
+        MsgSource.api, wallTime, blck, payloadValid  = true)
```
arnetheduck, line 1519: how come payload is considered valid when block is coming from REST? ie rest is more or less gossip, from a trust-point-of-view

beacon_chain/gossip_processing/block_processor.nim
tersec: It could do something like `traceAsyncErrors`, or set up yet another queue, but there's fundamentally an asynchronous aspect to this, and it can't avoid but to change the `storeBlock` contract in some way -- either it can return, without any particular guarantee of "full" success (even detecting blatantly invalid blocks) or they need to wait. One minimal change is always having
`await self.consensusManager.updateHeadWithExecution(wallTime.slotOrZero)` there be either a queue or (implicitly a queue) `traceAsyncErrors` that would detach the two, and allow `storeBlock` itself to be synchronous again.

Another point I played with was within the `updateHead` functions -- the serial dependence here is `forkchoiceUpdated` (intrinsically a long-lasting call) having to be complete before one can, in general, tell, the DAG about blocks being optimistic or verified. It's possible to just have it fire off the fcU, then immediately call `dag.updateHead`, but that's generally not great. One could arrange a callback or similar, but that's the kind of pattern async/await more-or-less linerarizes control-flow-wise. So probably within `updateHeadWithExecution`, the `async` is warranted.

beacon_chain/gossip_processing/block_processor.nim
tersec: It also creates these environments have vastly more symbols than they need imported, with no automatic way to distinguish whether they're doing so by Nim, because those imports are indeed necessary, its just, only a small part of them was. When isolating reproducible test cases for Nim issue filing I run across these repeatedly -- some module needs three symbols from some relatively small modules, but for some slight convenience, it does
```nim
import large_umbrella_module
```

which in turn is something like
```nim
import a, b, c, d, e, f
export a, b, c, d, e, f
```

and then some of `a`, `b`, etc continue this fractally. It works, yes, but it also inhibits analysis of what imports are necessary. If only one module imported by `large_umbrella_module` turns out to be necessary after some refactoring, that will not be automatically detected in a visible way.

tersec: Another aspect is that compile times are already problematic, and these single monolithic imports hinder attempts at improving that. Some of this is that there are individual modules which take a long time to compile, and no amount dependency management fixes that, but in other cases, the times of lots of small-to-medium modules being imported literally adds up. Even if/when Nim within a single "project" (`nim c foo`) doesn't keep recompiling modules, which, hopefully, it gets repeated across the `nimbus_beacon_chain`, `nimbus_validator_client`, et cetera.

One first step towards addressing this for any given module is clarifying what, exactly, it needs, and in various cases (e.g., the sqlite C code being compiled) that's meant splitting out `testdbutils` from `testutils`, explicitly avoiding this pile of `import`/`export` modules.

beacon_chain/gossip_processing/block_processor.nim
tersec: A general theme here is how much of the serial dependency is real vs how much is an artifact of a specific way of writing these functions (with async, with callbacks, with queues, with anything), but, in principle, to do the tasks they need to do to participate in the chain, how intrinsic vs how incidental are these delays.

One approach is to care even less about validity than it does now, and just ... not wait. Treat it as fire-and-semi-forget. Just import it if it's a candidate for optimistic sync, move on, etc. There might be some weird display quirks (e.g., the status bar showing wall slot != wall head briefly each slot even when everything goes well, or repeatedly showing `opt sync` briefly if it tries to be truly accurate for fractions of a second), but probably more important that the intrinsic dependencies exist: the aggregated attestations depend on the unaggregated attestations which depend on the correct head being known.

The challenge here is that https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#validator-assignments notes:
> 
> ## Validator assignments
>
> An optimistic node is *not* a full node. It is unable to produce blocks, since an execution engine cannot produce a payload upon an unknown parent. It cannot faithfully attest to the head block of the chain, since it has not fully
verified that block.
>
> ### Block Production
>
> An optimistic validator MUST NOT produce a block (i.e., sign across the `DOMAIN_BEACON_PROPOSER` domain).
>
> ### Attesting
>
> An optimistic validator MUST NOT participate in attestation (i.e., sign across the `DOMAIN_BEACON_ATTESTER`, `DOMAIN_SELECTION_PROOF` or `DOMAIN_AGGREGATE_AND_PROOF` domains).
>
> ### Participating in Sync Committees
>
> An optimistic validator MUST NOT participate in sync committees (i.e., sign across the
> `DOMAIN_SYNC_COMMITTEE`, `DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF` or
> `DOMAIN_CONTRIBUTION_AND_PROOF` domains).

To comply with the spec, this serialization enforced by `await` is required: if one wants to attest, and attest as soon as possible modulo the make-sure-everyone-gets-the-block-first delay, the block must be actually verified, not only optimistic.

So for a non-validating node, the literal stakes are lower, and for a validating node, there's ultimately no choice but to wait. At least this way, if the node has other things to do, it can do them in the meantime.

beacon_chain/gossip_processing/block_processor.nim
zah: There are no new tricks in 1.6 that you can use to prevent the copying. Perhaps you meant "prevent" in the sense of "outlaw", forcing a compile-time error when it happens, which is already available in Nim 1.2 with a definition like:

```nim
proc `=copy`(dest: var Foo; source: Foo) {.error: "outlawed".}
```

beacon_chain/gossip_processing/consensus_manager.nim
zah: I was not advocating that they should be used. I was explaining that there is no "cyclic import detection" as the message above implied (it would be relatively easy to add to Nim though).

Otherwise, I have advocated many times that cyclic imports should be supported better in the language because their current state leads to the proliferation of the "types module" which makes encapsulation difficult to achieve in Nim:

https://forum.nim-lang.org/t/4293

beacon_chain/gossip_processing/block_processor.nim
zah: Knowing the compiler internals, It's a bit hard to guess the effects on compilation times of this technique. You are right that each module in Nim is visited only once during compilation, so effectively a large project is likely to end up compiling the same set of files regardless of the import strategy. 

The `nim check` command can benefit though because it may need to see a smaller subset of the project when you are close to the leaf modules in the import tree. There might also be some slight speedup brought by reducing the scope lookup tables and the numbers of candidates considered during overload resolution in some situations.

For me, the main downside of this technique are that:

1) It requires quite a lot of work (unless your editor does it for you automatically)
2) It makes it even harder to spot inappropriate import cycles.

Both can be addressed with better tooling.

arnetheduck: > vastly more symbols than they need imported

this is true, but it is also often the case that a missing symbol is a lot more dangerous than an extra one - the former means a more generic overload will be chosen leading to subtle miserialization and similar issues due to nim-ser's open matching policy, the latter a (presumably fixable) compile-time error.

> single monolithic imports hinder attempts at improving that.

we have two categories of monoliths - giants like chronos that import and export everything under their project, and "small" modules that have hard dependencies of their own - the former should of course be split up, but with the latter, the following happens very often: module A imports B, and C imports A - to use A from C, B must also be imported, so C now imports A and B - then A changes and no longer needs B it remains in C as an explicit import and nim's (broken) import analysis indeed rarely identifies it as unused.

For the latter case, I've found that it's easier to export hard dependencies - modules that use A then have a shorter import list of their own, which helps keep the list tight overall - there is no compile-time downside in this example because A has a hard dep on B, so B must be compiled no matter what..

We have two more issues:

* we reuse types from implementation libraries - for example , we reuse eth1 block hashes in eth2, which brings in _a lot_ of junk from eth1 instead of using just a minimal eth2-specific definition (ditto libp2p) - presumably this is done because copying is bad, but it also creates rigit ties between the project that are not always correct (just because an eth1 block has is serialized to hex in one way in eth1 doesn't mean eth2 does it the same way - these kinds of false carry-overs are quite problematic)
* the p2p macro defines both client and server at the same time - this requires clients (like the VC) to import practically all of the server as well (VC doesn't need libp2p, but still imports it) - similar problems happen with other "categories" of imports, and the result is that "auxiliary" compiles like tests and simulations take longer even though they don't use a lot of the stuff that transitively ends up being processed - bearssl was a good example as well

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3842 might address this. It needs testing, but should be a viable approach. It keeps enough of these invariants that it should remain valid, but be less invasive.

beacon_chain/gossip_processing/block_processor.nim
tersec: > the former means a more generic overload will be chosen leading to subtle miserialization and similar issues due to nim-ser's open matching policy

That seems like an issue with nim-ser's open matching policy.

arnetheduck: > That seems like an issue with nim-ser's open matching policy.

yes, that's the biggest and most obvious culprit, and @zah has promised progress on this front, but it happens with other things generic as well (`hash` function which messes up `Table`, `$`, etc)

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3842

beacon_chain/sync/sync_protocol.nim
```diff
@@ -454,6 +458,13 @@ p2pProtocol BeaconSync(version = 1,
 
     for i in startIndex..endIndex:
       if dag.getBlockSZ(blocks[i], bytes):
+        # In general, there is not much intermediate time between post-merge
+        # blocks all being optimistic and none of them being optimistic. The
+        # EL catches up, tells the CL the head is verified, and that's it.
+        if  blocks[i].slot.epoch >= dag.cfg.BELLATRIX_FORK_EPOCH and
+            dag.is_optimistic(dag.head.root):
```
etan-status, line 465: Why are we not responding with blocks, while our primary head is SYNCING?

https://github.com/ethereum/consensus-specs/blob/v1.5.0/specs/bellatrix/p2p-interface.md#why-allow-invalid-payloads-on-the-p2p-network

> The specification allows blocks with invalid execution payloads to propagate across gossip and via RPC calls

...

> Therefore, optimistic nodes must send optimistic blocks via RPC.

Is the intention here to block sending of known INVALID payloads? But then, it should not be based on `dag.head.root`...

Note, this is now refactored to the following snippet, but still lives in `sync_protocol.nim`.

```nim
        # In general, there is not much intermediate time between post-merge
        # blocks all being optimistic and none of them being optimistic. The
        # EL catches up, tells the CL the head is verified, and that's it.
        if  blocks[i].slot.epoch >= dag.cfg.BELLATRIX_FORK_EPOCH and
            not dag.head.executionValid:
          continue
```

etan-status: image

There's also a weaker "may" version.

In any case, if the intention is to avoid optimistic blocks, probably best to use `ResourcesUnavailable`.
Or, maybe we should just allow optimistic blocks to be made available. The spec talks about "RPC" which suggests beacon-API, but it's on the p2p document which suggests Req/Resp.

etan-status: also, our status message indicates optimistic head. so we say "we are on root X", then someone asks us "give me block X", and we say "it doesn't exist" (current logic) / "we cannot give you the head / it's unavailable" (ResourceUnavailable case).

etan-status: These are possible solutions, with increasing distance from current logic:

1. Current solution is not good for the network, as we provide incorrect response ("it really doesn't exist" instead of "I am still syncing / I have pruned it")
2. `ResourceUnavailable` probably better, would make our response honest / consistent with reality.
3. Look at the actual block that we provide, vs. the `dag.head`? As in, historical blocks that are already finalized are implicitly valid and should always be made available, they are not optimistic...
4. Restricting the limit to known-invalid blocks vs all optimistic blocks, that would allow blocks to propagate faster that are still syncing (which SHOULD NOT cause down-scores or disconnects according to spec)
5. As the request is by range, it is relative to the DAG head anyway, which is.. not a descendant of a known invalid. So we can (and possibly should) just remove the check and provide all the blocks known to us...

etan-status: These are possible solutions, with increasing distance from current logic:

1. Current solution is not good for the network, as we provide incorrect response ("it really doesn't exist" instead of "I am still syncing / I have pruned it")
2. `ResourceUnavailable` probably better, would make our response honest / consistent with reality.
3. Look at the actual block that we provide, vs. the `dag.head`? As in, historical blocks that are already finalized are implicitly valid and should always be made available, they are not optimistic... We should consider also only advertising the finalized head (or the latest valid head) in status message while syncing, in this case.
4. Restricting the limit to just known-invalid blocks (instead of all optimistic blocks). That would allow blocks to propagate faster that are still syncing (which SHOULD NOT cause down-scores or disconnects according to spec)
5. As the request is by range, it is relative to the DAG head anyway, which is.. not a descendant of a known invalid. So we can (and possibly should) just remove the check and provide all the blocks known to us...

etan-status: #7198

PR #3798, opened 2022-06-24
vc: initial documentation

arnetheduck:
bare-bones, but enough to get started

### DIFF
--- a/docs/the_nimbus_book/src/validator-client.md
+++ b/docs/the_nimbus_book/src/validator-client.md
@@ -1,0 +1,54 @@
+# Run a separate validator client
+
+> ⚠️ This feature is currently in BETA - we are still testing it and implementation details may change in response to community feedback. **We strongly advise against using it on mainnet** - your validators may get slashed
+
+By default, Nimbus loads validator keys into the main beacon node process, which is a simple, safe and efficient way to run a validator.
+
+Advanced users may wish to run validators in a separate process, allowing more flexible deployment strategies. The Nimbus beacon node supports both it's own and third-party validator clients via the built-in [REST API](./rest-api.md).
+
+> ⚠️ So far, all slashings with known causes have been linked to overly complex setups involving separation between beacon node and validator client! Only use this setup if you've taken steps to mitigate the increased risk.
+
+## Build
+
+The validator client is currently only available when built from source. To build the validator client, [build the beacon node](./build.md), then issue:
+
+```sh
+make -j4 nimbus_validator_client
+```
+
+When upgrading, don't forget to upgrade the validator client also, with the same command as above!
+
+## Setup
+
+To run a separate validator client, you must first make sure that your beacon node has its REST API enabled - make sure that you've started with the `--rest` option.
+
+Next, choose a data directory for the validator client and import the keys there:
+
+```sh
+build/nimbus_beacon_node deposits import \
+  --data-dir:build/data/vc_shared_prater_0 "<YOUR VALIDATOR KEYS DIRECTORY>"
+```
+
+> ⚠️ Do not use the same data directory for beacon node and validator client - they will both try to load the same keys which may result in slashing!
+
+> ⚠️ If you are migrating your keys from the beacon node to the validator client, simply move the `secrets` and `validators` folders in the beacon node data directory to the data directory of the validator client
+
+With the keys imported, you are ready to start validator client:
+
+```sh
+build/nimbus_validator_client --network:prater --data-dir:build/data/vc_shared_prater_0
+```
+
+## Options
+
+`nimbus_validator_client` supports command line options similar to the beacon node:
+```sh
+# See help
+build/nimbus_validator_client --help
+```
+
+The client will by defualt connect to a beacon node on the same machine as the validator client. Pick a different node with `--beacon-node`:
+
+```sh
+build/nimbus_validator_client --network:prater --beacon-node:http://host:port/
+```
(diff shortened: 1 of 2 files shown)


docs/the_nimbus_book/src/validator-client.md
```diff
@@ -0,0 +1,59 @@
+# Run a separate validator client
+
+> ⚠️ This feature is currently in BETA - we are still testing it and implementation details may change in response to community feedback. **We strongly advise against using it on mainnet** - your validators may get slashed
+
+By default, Nimbus loads validator keys into the main beacon node process, which is a simple, safe and efficient way to run a validator.
+
+Advanced users may wish to run validators in a separate process, allowing more flexible deployment strategies. The Nimbus beacon node supports both it's own and third-party validator clients via the built-in [REST API](./rest-api.md).
```
arnetheduck:
```suggestion
Advanced users may wish to run validators in a separate process, allowing more flexible deployment strategies. The Nimbus beacon node supports both its own and third-party validator clients via the built-in [REST API](./rest-api.md).
```

docs/the_nimbus_book/src/validator-client.md
```diff
@@ -0,0 +1,59 @@
...
+
+> ⚠️ Do not use the same data directory for beacon node and validator client - they will both try to load the same keys which may result in slashing!
+
+> ⚠️ If you are migrating your keys from the beacon node to the validator client, simply move the `secrets` and `validators` folders in the beacon node data directory to the data directory of the validator client
+
+With the keys imported, you are ready to start validator client:
+
+```sh
+build/nimbus_validator_client \
+  --network:prater \
```
jclapis: This fails with `Unrecognized option 'network'`, same with the other example further below (the `--beacon-node` one).

docs/the_nimbus_book/src/validator-client.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Build
+
+The validator client is currently only available when built from source. To build the validator client, [build the beacon node](./build.md), then issue:
+
+```sh
+make -j4 nimbus_validator_client
+```
+
+When upgrading, don't forget to upgrade the validator client also, with the same command as above!
```
zah: It may be better to explicitly state that the user must follow the same upgrade steps as with the beacon node (with link to the respective page).

docs/the_nimbus_book/src/validator-client.md
arnetheduck: right, shouldn't be there, 527915a9f820d1b97c95d1bd95f832801b907689

docs/the_nimbus_book/src/validator-client.md
arnetheduck: 527915a9f820d1b97c95d1bd95f832801b907689

docs/the_nimbus_book/src/validator-client.md
```diff
@@ -0,0 +1,62 @@
...
+Advanced users may wish to run validators in a separate process, allowing more flexible deployment strategies. The Nimbus beacon node supports both its own and third-party validator clients via the built-in [REST API](./rest-api.md).
+
+> ⚠️ So far, all slashings with known causes have been linked to overly complex setups involving separation between beacon node and validator client! Only use this setup if you've taken steps to mitigate the increased risk.
+
+## Build
+
+The validator client is currently only available when built from source. To build the validator client, [build the beacon node](./build.md), then issue:
+
+```sh
+make -j4 nimbus_validator_client
```
tersec, line 16: The documentation typically uses this rather than something like `-j$(nproc)`. Presumably, this compromises between a reasonable guess at lowest-common-denominator number of cores and compatibility across Linux and macOS (which doesn't support, per se, `nproc`, though it has its own approaches to this), but it's not ideal.

docs/the_nimbus_book/src/validator-client.md
```diff
@@ -0,0 +1,62 @@
...
+With the keys imported, you are ready to start validator client:
+
+```sh
+build/nimbus_validator_client \
+  --data-dir:build/data/vc_shared_prater_0
+```
+
+## Options
+
+`nimbus_validator_client` supports command line options similar to the beacon node:
```
tersec: Probably better to be consistent here and either refer to beacon nodes generically as a non-specific noun (as is now) but in a way that's accurate and precise with regard to what beacon nodes in general do, or specifically `nimbus_beacon_node`, with which it's being compared.

docs/the_nimbus_book/src/validator-client.md
arnetheduck: indeed - this lowest-common-denominator approach gives most of the benefit without the mess

tersec: This is an issue of not-great defaults though, and while it's messy, it's messiness that should be handled by default by the build system: https://github.com/status-im/nimbus-eth2/issues/3806

tersec: But, out of scope of this PR, sure.

arnetheduck: Yes, apparently that's difficult to do, at a make level  for whatever reason. I'm guessing `nimble` will learn this at some point

docs/the_nimbus_book/src/validator-client.md
arnetheduck: 78048fca9b9d614508b1efd5c37baa200b5be896

PR #3799, opened 2022-06-25
add option to configure max historic LC data periods

etan-status:
Adds a `--light-client-data-max-periods` option to override the number
of sync committee periods to retain light client data.
Raising it above the default enables archive nodes to serve full data.
Lowering below the default speeds up import times (still no persistence)


tersec (approved):

beacon_chain/conf.nim
```diff
@@ -456,6 +456,11 @@ type
               "Must be one of: none, only-new, full (slow startup), on-demand (may miss validator duties)"
         name: "light-client-data-import-mode" .}: Option[LightClientDataImportMode]
 
+      lightClientDataMaxPeriods* {.
```
arnetheduck, line 459: I've toyed with the idea to call this a `sync-horizon` - ie the same will happen for blocks and states eventually .. not a blocker for this PR but ..

etan-status: Yeah, I have seen those, but they currently have a different meaning in sync manager, to sync up through at most horizon within wall clock. This here is a limit for syncing backwards.

PR #3804, opened 2022-06-27
use eth_chainId instead of net_version

tersec:
`net_version` isn't one of the supported engine API methods in https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#underlying-protocol while `eth_chainId` is.

Until https://github.com/ethereum/go-ethereum/pull/25166 was merged, Geth effectively didn't support `eth_chainId` in merge networks in situations where it mattered, but with that fixed, it's now worth switching Nimbus to use it.

Tested with
- `rm ~/.ethereum/ -rv && go-ethereum/build/bin/geth --http --mainnet` and `build/nimbus_beacon_node --network=mainnet web3 test --url=http://127.0.0.1:8545`
- `rm ~/.ethereum/ -rv && go-ethereum/build/bin/geth --http --ropsten` and `build/nimbus_beacon_node --network=ropsten web3 test --url=http://127.0.0.1:8545`
- `rm ~/.ethereum/ -rv && go-ethereum/build/bin/geth --http --goerli` and `build/nimbus_beacon_node --network=prater web3 test --url=http://127.0.0.1:8545`
- `rm ~/.ethereum/ -rv && go-ethereum/build/bin/geth --http --sepolia` and `build/nimbus_beacon_node --network=sepolia web3 test --url=http://127.0.0.1:8545`

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -1061,8 +1061,7 @@ proc detectPrimaryProviderComingOnline(m: Eth1Monitor) {.async.} =
     # Use one of the get/request-type methods from
     # https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.9/src/engine/specification.md#underlying-protocol
     # which does nit take parameters and returns a small structure, to ensure
-    # this works with engine API endpoints. Either eth_chainId or eth_syncing
-    # works for this purpose.
+    # this works with engine API endpoints.
     let testRequest = tempProvider.web3.provider.eth_syncing()
 
     yield testRequest or sleepAsync(web3Timeouts)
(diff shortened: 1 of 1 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1343,18 +1342,27 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
...
+          of mainnet: 1'u64
+          of ropsten: 3'u64
+          of rinkeby: 4'u64
+          of goerli:  5'u64
+          of sepolia: 11155111'u64   # https://chainid.network/
+      if expectedChain != providerChain:
+        fatal "The specified web3 provider serves data for a different chain",
+               expectedChain, providerChain
+        quit 1
+    except CatchableError as exc:
```
etan-status, line 1362: Shouldn't `DataProviderFailure` be enough here?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1343,18 +1342,27 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
...
+        providerChain =
+          (awaitWithRetries m.dataProvider.web3.provider.eth_chainId()).uint64
+
+        # https://eips.ethereum.org/EIPS/eip-155#list-of-chain-ids
+        expectedChain = case m.eth1Network.get
+          of mainnet: 1'u64
+          of ropsten: 3'u64
+          of rinkeby: 4'u64
+          of goerli:  5'u64
+          of sepolia: 11155111'u64   # https://chainid.network/
```
etan-status:
```suggestion
        providerChain =
          awaitWithRetries m.dataProvider.web3.provider.eth_chainId()

        # https://eips.ethereum.org/EIPS/eip-155#list-of-chain-ids
        expectedChain = case m.eth1Network.get
          of mainnet: 1.Quantity
          of ropsten: 3.Quantity
          of rinkeby: 4.Quantity
          of goerli:  5.Quantity
          of sepolia: 11155111.Quantity   # https://chainid.network/
```
A bit more future-proof to notice when `eth_chainId` suddenly returns a different type (if `==` is borrowed). Otherwise, `distinctBase` etc probably make it too verbose, so current solution would win.

etan-status (approved):
Nice that all major ELs implement this and no fallback seems necessary.

beacon_chain/eth1/eth1_monitor.nim
tersec: Maybe. Didn't want to risk it, because debugging that in an async context is problematic.

tersec: Also, `{.push raises: [Defect].}` effectively doesn't apply to `{.async.}` functions, which means that even if currently, `DataProviderFailure` might be enough, changes in the library implementation could undermine that with no compiler-time warnings or errors.

PR #3807, opened 2022-06-27
REST: fix timing check for bellatrix epoch.

cheatfate:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -253,7 +267,15 @@ proc initFullNode(
                                     finalizedEpochRef.eth1_data,
                                     finalizedEpochRef.eth1_deposit_index)
       node.updateLightClientFromDag()
-      eventBus.finalQueue.emit(data)
+      # TODO (cheatfate): Proper implementation required
```
tersec, line 270: With this PR, would this not have been accomplished? Of course, maybe what will be viewed as a proper implementation will change, but I'm not sure what this is standing in for (likewise the other `TODO (cheatfate): Proper implementation required` comments in similar place).

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -310,6 +332,9 @@ proc initFullNode(
       getFrontfillSlot, dag.backfill.slot, blockVerifier, maxHeadAge = 0)
 
   dag.setFinalizationCb makeOnFinalizationCb(node.eventBus, node.eth1Monitor)
+  dag.setBlockCb(onBlockAdded)
```
tersec, line 335: This would mean that only full nodes, not light nodes, call `onBlockAdded`, `onHeadChanged`, and `onChainReorg`, but none of them should happen in a light node anyway (`onBlockAdded` is called from `addResolvedHeadBlock`, and `onHeadChanged` and  `onReorgHappened` are functions of running fork choice and `ChainDAGRef.updateHead`, so not relevant to light clients.

tersec (approved):

beacon_chain/nimbus_beacon_node.nim
etan-status: Yeah, `blockchain_dag_light_client.nim` only emits light client finality / optimistic updates, and does not depend on those other callbacks.

beacon_chain/nimbus_beacon_node.nim
etan-status: Well, it always is either `none` or `some(false)` -- is it never intended to be `some(true)`?

What is fixed here is that the `optimistic` field is included as soon as wall clock passes past bellatrix, instead of this depending on the locally synced DAG head slot.

etan-status (approved):

beacon_chain/nimbus_beacon_node.nim
tersec: For the moment, yes, it's never intended to be `some(true)`. https://github.com/status-im/nimbus-eth2/pull/3793 might change this, but Nimbus `unstable` is never an optimistic node, nor does it have optimistically imported blocks.

PR #3809, opened 2022-06-27
Avoid overflow in the initial 'nextExchangeTransitionConfTime' calculation

zah:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -693,11 +693,19 @@ proc init*(T: type BeaconNode,
       else:
         nil
 
-    bellatrixEpochTime =
-      genesisTime + cfg.BELLATRIX_FORK_EPOCH * SLOTS_PER_EPOCH * SECONDS_PER_SLOT
+    maxSecondsInMomentType = int64.high div 1_000_000_000
```
tersec: This seems like awfully leaky abstraction -- ideally, this shouldn't depend on the representation of the `Moment` type, while this can break if what should be an irrelevant aspect of the `Moment` type changes.

zah: Yes, this was unpleasant to write.

I've addressed it now by adding `Moment.high` operator in Chronos:
https://github.com/status-im/nim-chronos/pull/290

tersec (approved):

PR #3810, opened 2022-06-27
remove web3 url prompt in launcher script

arnetheduck:
The interactive prompt for web3 has outlived its utility as we now load
url:s from command line params and config files, preventing the prompt
from correctly detecting when it's needed.

Also, after the merge, a JWT secret will (likely) be needed.

* log notice when web3 url is missing
* fix docs to not mention default that doesn't exist
* fix scripts to properly quote arguments

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -474,6 +474,12 @@ proc init*(T: type BeaconNode,
     else:
       none(DepositContractSnapshot)
 
+  if config.web3Urls.len() == 0:
+    if BELLATRIX_FORK_EPOCH == FAR_FUTURE_EPOCH:
+      notice "Running without execution client - validator features partially disabled (see https://nimbus.guide/eth1.html)"
+    else:
+      notice "Running without execution client - validator features disabled (see https://nimbus.guide/eth1.html)"
+
   var eth1Monitor: Eth1Monitor
   if not ChainDAGRef.isInitialized(db).isOk():
     var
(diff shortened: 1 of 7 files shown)


run-mainnet-beacon-node.sh
```diff
@@ -9,5 +9,5 @@
 
 cd "$(dirname $0)"
 # Allow the binary to receive signals directly.
-exec scripts/run-beacon-node.sh nimbus_beacon_node mainnet $@
+exec scripts/run-beacon-node.sh nimbus_beacon_node mainnet "$@"
```
zah, line 12: `$@` here can expand to multiple command-line options. I'm not sure these quotes are appropriate.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -474,6 +474,12 @@ proc init*(T: type BeaconNode,
     else:
       none(DepositContractSnapshot)
 
+  if config.web3Urls.len() == 0:
+    if BELLATRIX_FORK_EPOCH == FAR_FUTURE_EPOCH:
```
zah: `BELLATRIX_FORK_EPOCH` is not supposed to be a constant I think. It should be part of the `RuntimeConfig`

run-mainnet-beacon-node.sh
arnetheduck: it's inappropriate without the quotes, pretty much always: https://stackoverflow.com/a/3816747

beacon_chain/nimbus_beacon_node.nim
arnetheduck: oops 7473049dba54a4b58b31e30128a3c73f4f16b0b6

PR #3812, opened 2022-06-27
Revert "Switch back to nim-websock"

tersec:
Reverts status-im/nimbus-eth2#3665

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -118,6 +118,10 @@ switch("passL", "-fno-omit-frame-pointer")
 # for heap-usage-by-instance-type metrics and object base-type strings
 --define:nimTypeNames
 
+# TODO https://github.com/status-im/nimbus-eth2/issues/3130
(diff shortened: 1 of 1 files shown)


arnetheduck:
what are the chances of making this a command-line parameter instead? cc @Menduist

tersec:
> what are the chances of making this a command-line parameter instead? cc @Menduist

It can be a command-line parameter once it's out of the critical path and not preventing the Ropsten fleet from working.

zah:
A run-time option would be difficult, but we can enable websock on some of our fleet nodes. We are already supporting custom compilation flags there.

zah:
Compile-time setup for the fleet has been proposed here: https://github.com/status-im/infra-nimbus/pull/113
It requires https://github.com/status-im/nimbus-eth2/pull/3820 to be merged first.

PR #3819, opened 2022-06-28
cleanups

arnetheduck:
* avoid circular panda imports
* move deposit merkleization helpers to spec/
* normalize validator signature helpers to spec names / params
* remove redundant functions for remote signing

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -14,6 +14,7 @@ import
   stew/[byteutils, io2],
   eth/p2p/discoveryv5/[enr, random2],
   eth/keys,
+  ./consensus_object_pools/vanity_logs/pandas,
   ./rpc/[rest_api, state_ttl_cache],
   ./spec/datatypes/[altair, bellatrix, phase0],
   ./spec/[engine_authentication, weak_subjectivity],
@@ -29,8 +30,6 @@ from
   libp2p/protocols/pubsub/gossipsub
 import
   TopicParams, validateParameters, init
-
-from "."/consensus_object_pools/vanity_logs/pandas import getPandas
 
 when defined(windows):
   import winlean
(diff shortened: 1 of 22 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -145,6 +144,22 @@ versionGauge.set(1, labelValues=[fullVersionStr, gitRevision])
 
 logScope: topics = "beacnde"
 
+func getPandas(stdoutKind: StdoutLogKind): VanityLogs =
```
tersec, line 147: Keeping 🐼s isolated in one module (both the Unicode codepoints and the general idea) was intentional.

arnetheduck: it also led to a circular import of `conf`, the way it was structured, which in turn led to a cascade of other imports in unrelated places - is it .. important that it stays there? if so , the logging output stuff needs to go into its own module

tersec: Avoiding circular imports is more important, sure, and this seems basically fine.

Text editors should all be fairly good at Unicode by now, and if not, it'll show up in diffs, so maybe one of the concerns (to keep non-Latin-1-ish characters out of frequently modified files) isn't so crucial.

tersec (approved):

tersec:
Jenkins CI failing because
```
./scripts/launch_local_testnet.sh: line 860: EL_RPC_PORTS: unbound variable
script returned exit code 1
```
but this PR doesn't modify any of those scripts, so it's something pre-existing.

etan-status:
> Jenkins CI failing because
> 
> ```
> ./scripts/launch_local_testnet.sh: line 860: EL_RPC_PORTS: unbound variable
> script returned exit code 1
> ```
> 
> but this PR doesn't modify any of those scripts, so it's something pre-existing.

need to merge from `unstable`, there was some regression that seems to have been fixed there.

tersec:
Not sure it is, see https://github.com/status-im/nimbus-eth2/pull/3822

https://github.com/status-im/nimbus-eth2/commit/a48d741022f6b0da1bb679e0ede4e38c019242cf caused the `EL_RPC_PORTS` error, then https://github.com/status-im/nimbus-eth2/commit/73227508db6f8ece8657f33c76d2762691ac172e fixed that by defining said variable, but didn't fix the JWT setup.

Maybe one of the commits since then picked that up.

tersec:
At least through https://ci.status.im/blue/organizations/jenkins/nimbus%2Fnimbus-eth2%2Fmultibranch/detail/unstable/1460/pipeline/149 (https://github.com/status-im/nimbus-eth2/pull/3796) it's still failing with:
```
{"lvl":"FAT","ts":"2022-06-28 23:04:24.832+00:00","msg":"Specified a JWT secret file which couldn't be loaded","topics":"beacnde","err":"couldn't open specified JWT secret file"}
```

https://github.com/status-im/nimbus-eth2/pull/3822 should fix this though.

PR #3821, opened 2022-06-28
Version 22.6.1

zah:


arnetheduck:
shouldn't sepolia ttd be in here as well?

arnetheduck:
ok nevermind, looks like it's on stable already

PR #3823, opened 2022-06-28
persist LC data across restarts

etan-status:
With the Altair spec `LightClientUpdate` structure taking its final form
it is finally possible to persist LC data across restarts without having
to worry about data migration due to spec changes. A separate `lcdataV1`
database is created in the `caches` subdirectory to hold known LC data.
A full database with default settings (129 periods) uses <15 MB disk.

### DIFF
--- a/beacon_chain/light_client_data_db.nim
+++ b/beacon_chain/light_client_data_db.nim
@@ -86,6 +86,12 @@ type
       ## Tracks the finalized sync committee periods for which complete data
       ## has been imported (from `dag.tail.slot`).
 
+# No `uint64` support in Sqlite
+template isSupportedBySQLite(slot: Slot): bool =
+  slot <= int64.high.Slot
+template isSupportedBySQLite(period: SyncCommitteePeriod): bool =
+  period <= int64.high.SyncCommitteePeriod
+
 proc initCurrentSyncCommitteeBranchStore(
     backend: SqStoreRef): KvResult[CurrentSyncCommitteeBranchStore] =
   ? backend.exec("""
@@ -124,8 +130,8 @@ proc initCurrentSyncCommitteeBranchStore(
 
 func hasCurrentSyncCommitteeBranch*(
     db: LightClientDataDB, slot: Slot): bool =
-  if slot > int64.high.Slot:
-    return false  # No `uint64` support
+  if not slot.isSupportedBySQLite:
+    return false
   var exists: int64
   for res in db.currentBranches.containsStmt.exec(slot.int64, exists):
     res.expect("SQL query OK")
@@ -135,8 +141,8 @@ func hasCurrentSyncCommitteeBranch*(
 
 proc getCurrentSyncCommitteeBranch*(
     db: LightClientDataDB, slot: Slot): altair.CurrentSyncCommitteeBranch =
-  if slot > int64.high.Slot:
-    return  # No `uint64` support
+  if not slot.isSupportedBySQLite:
+    return default(altair.CurrentSyncCommitteeBranch)
   var branch: seq[byte]
   for res in db.currentBranches.getStmt.exec(slot.int64, branch):
     res.expect("SQL query OK")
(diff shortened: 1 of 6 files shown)


etan-status:
[lcdataV1.sqlite3.zip](https://github.com/status-im/nimbus-eth2/files/9006029/lcdataV1.sqlite3.zip)
Prater DB for testing.

beacon_chain/light_client_data_db.nim
```diff
@@ -0,0 +1,342 @@
...
+    getStmt: getStmt,
+    putStmt: putStmt,
+    delStmt: delStmt,
+    delFromStmt: delFromStmt,
+    keepFromStmt: keepFromStmt)
+
+proc getBestUpdate*(
+    db: LightClientDataDB, period: SyncCommitteePeriod
+): altair.LightClientUpdate =
+  doAssert period <= int64.high.SyncCommitteePeriod
```
tersec: This exact bounds check is performed in several functions, operating on slightly different variables (`period`, et cetera) -- might be useful to pull it out as a template or similar.

beacon_chain/light_client_data_db.nim
```diff
@@ -0,0 +1,342 @@
...
+  for res in db.currentBranches.containsStmt.exec(slot.int64, exists):
+    res.expect("SQL query OK")
+    doAssert exists == 1
+    return true
+  return false
+
+proc getCurrentSyncCommitteeBranch*(
+    db: LightClientDataDB, slot: Slot): altair.CurrentSyncCommitteeBranch =
+  if slot > int64.high.Slot:
+    return  # No `uint64` support
```
tersec: This relies on a quirk of how `result` works in terms of returning a `default(altair.CurrentSyncCommitteeBranch)`, and it seems better to avoid this implicit dependence on what empty return does in a function. In particular, it's path-dependent in a potentially nonobvious way, i.e. the same `return` statement could mean different things depending on if something had set `result` prior to any particular return.

If the goal is to return an empty or default `altair.CurrentSyncCommitteeBranch`, that can be explicit and in theory RVO or NRVO should handle the nominal inefficiency.

Likewise the other plain `return` in this function.

beacon_chain/light_client_data_db.nim
```diff
@@ -0,0 +1,342 @@
...
+func hasCurrentSyncCommitteeBranch*(
+    db: LightClientDataDB, slot: Slot): bool =
+  if slot > int64.high.Slot:
+    return false  # No `uint64` support
+  var exists: int64
+  for res in db.currentBranches.containsStmt.exec(slot.int64, exists):
+    res.expect("SQL query OK")
+    doAssert exists == 1
+    return true
+  return false
```
tersec:
```suggestion
  false
```

beacon_chain/light_client_data_db.nim
```diff
@@ -0,0 +1,342 @@
...
+
+func isPeriodSealed*(
+    db: LightClientDataDB, period: SyncCommitteePeriod): bool =
+  doAssert period <= int64.high.SyncCommitteePeriod
+  var exists: int64
+  for res in db.sealedPeriods.containsStmt.exec(period.int64, exists):
+    res.expect("SQL query OK")
+    doAssert exists == 1
+    return true
+  return false
```
tersec, line 274:
```suggestion
  false
```

tersec:
```suggestion
  false
```

beacon_chain/light_client_data_db.nim
etan-status: 0fa7ad9a59872cbb63752ec4347cf6aabcb41aaf

beacon_chain/light_client_data_db.nim
etan-status: 0fa7ad9a59872cbb63752ec4347cf6aabcb41aaf

beacon_chain/light_client_data_db.nim
etan-status: 0fa7ad9a59872cbb63752ec4347cf6aabcb41aaf

beacon_chain/light_client_data_db.nim
etan-status: 0fa7ad9a59872cbb63752ec4347cf6aabcb41aaf

beacon_chain/light_client_data_db.nim
etan-status: df946ac12868574ceda3697d5f2544b9f43abb60

beacon_chain/light_client_data_db.nim
etan-status: df946ac12868574ceda3697d5f2544b9f43abb60

arnetheduck:
why keep this in a separate database? this is supposed to be part of the "standard feature set" of every beacon node running the full protocol, so it should really live in the main db, no?

PR #3827, opened 2022-06-29
ci: refactor Jenkinsfile to be a pipeline

jakubgs:
Changes:
- Name local testnet output folders same as the `make` taget.
- Move both `Jenkinsfile`s to `ci` folder to avoid cluttering repo root.
- Separate builds by platform so logs from macos and linux hosts don't get mixed.
- Detect platform and architecture from Jenkins Job path to use one Jenkinsfile.
- Divide shell commands into as many stages as possible to make debugging easier.
- Generalize running testnets via a `launchLocalTestnet()` Groovy method.
- Handle uploading of results of running testnets stage-by-stage basis.
- Use `catchError()` to upload test results while marking job as failed.
- Abort previously started PR build jobs usin `disableConcurrentBuilds()`.
- Throttle jobs using the new `throttleJobProperty()` function.

Builds:
- https://ci.status.im/job/nimbus-eth2/job/platforms/job/linux/job/x86_64/
- https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/x86_64/
- https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/aarch64/

Result:

image

image

This should be easier to follow and debug.


tersec:
Your description says
> - Move both Jenkinsfiles to ci folder to avoid cluttering repo root

But while it does move `Jenkinsfiles.benchmarks`, it keeps both the existing `Jenkinsfile` in https://github.com/status-im/nimbus-eth2/blob/ci/refactor-jenkinsfile/Jenkinsfile and adds the new one in https://github.com/status-im/nimbus-eth2/blob/ci/refactor-jenkinsfile/ci/Jenkinsfile

jakubgs:
Ah, good point, must have lost that during one of the rebases.

zah:
I'm slightly confused by the different Jenkins results. Are they referring to the same simulation runs? Or are we running more simulations now?

jakubgs:
It should be running exactly the same thing as before. It's just organized differently. Two testnets per each platform.

Issue #3828, opened 2022-06-30
VC: keeps spamming BN on failure


arnetheduck:
When the BN is unable to publish an attestation (for example because it has no peers on the topic), VC starts a tight spam loop, reposting the same attestation over and over again - it seems like it never stops until it finally succeeds - below is an example of such an attestation - the same thing happens with aggregates.

If a BN fails to process an attestation, the VC should retry another BN perhaps, but not keep sending the message to the same BN.

Below is an example (with a new log message):
```
00000000000000000000000000000000000000000000000, data: (slot: 3338476, index: 43, beacon_block_root: \"33e1201b\", source: \"104326:b5382e88\", target: \"104327:01d74b6f\"), signature: \"844e51a3\")" error="No peers on libp2p topic"
NOT 2022-06-30 08:15:20.120+02:00 No peers for topic, skipping publish       topics="libp2p gossipsub" peersOnTopic=0 connectedPeers=0 topic=/eth2/82f4a72b/beacon_attestation_43/ssz_snappy
NOT 2022-06-30 08:15:20.120+02:00 Produced attestation not sent              attestation="(aggregation_bits: 0b00000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 3338476, index: 43, beacon_block_root: \"33e1201b\", source: \"104326:b5382e88\", target: \"104327:01d74b6f\"), signature: \"844e51a3\")" error="No peers on libp2p topic"
NOT 2022-06-30 08:15:20.163+02:00 No peers for topic, skipping publish       topics="libp2p gossipsub" peersOnTopic=0 connectedPeers=0 topic=/eth2/82f4a72b/beacon_attestation_43/ssz_snappy
NOT 2022-06-30 08:15:20.163+02:00 Produced attestation not sent              attestation="(aggregation_bits: 0b00000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 3338476, index: 43, beacon_block_root: \"33e1201b\", source: \"104326:b5382e88\", target: \"104327:01d74b6f\"), signature: \"844e51a3\")" error="No peers on libp2p topic"
[...]

PR #3829, opened 2022-06-30
Introduce message router

arnetheduck:
Whether new blocks/attestations/etc are produced internally or received
via REST, their journey through the node is the same - to ensure that
they get the same treatment (logging, metrics, processing), this PR
moves the routing to a dedicated module and fixes several small
differences that existed before.

* `xxxValidator` -> `processMessageName` - the processor also was adding
messages to pools, so we want the name to reflect that action
* add missing "sent" metrics for some messages
* document ignore policy better - already-seen messages are not actaully
rebroadcast by libp2p
* skip redundant signature checks for internal validators consistently


tersec:
Jenkins artifacts on `unstable`:
2022-06-30-213250_900x643_scrot

Jenkins artifacts for this PR:
2022-06-30-213448_903x636_scrot

Those artifacts got about 100x larger. It's not that it's paramount that they be minimally sized, but that's unreasonably larger. Furthermore, what used to take literal seconds (3 seconds in https://ci.status.im/blue/organizations/jenkins/nimbus%2Fnimbus-eth2%2Fmultibranch/detail/unstable/1473/pipeline) here requires 5.5 minutes:
2022-06-30-213844_898x118_scrot

I ran into this with my PR along these lines, and it's part of what motivated switching to gzip -- bzip2 took a half hour or more. Unsurprising, because:
```
local_testnet1_data$ du -ah
2.7G    ./log_val0.txt
3.9G    ./log3.txt
2.8G    ./log_val3.txt
4.0K    ./bootstrap_nodes.txt
4.0G    ./log0.txt
2.8G    ./log_val1.txt
2.8G    ./log_val2.txt
3.9G    ./log1.txt
128K    ./log_lc0.txt
3.9G    ./log2.txt
27G     .
```
That's for the Linux testnets. The two macOS ones look similar.

While this likely is all a symptom rather than cause, and is merely indicating a real underlying issue which should be fixed, it also might be better to keep a CI regression of this level out of `unstable`, even if that means temporarily isolating the part of this PR which is causing that change. CI is there to detect bugs, yes, but this potentially interferes with finding other bugs.

In particular the sheer mass of circumstantial evidence points to an unfortunate interaction between the VC, BN, and sync committee messages:
```
local_testnet1_data$ jq .msg < log0.txt | sort | uniq -cd | sort -rn | head -n10
3020340 "Received request"
3020334 "Serving API request"
2414640 "Received response from handler"
2090479 "Sync committee message received"
2070184 "Sync committee message sent"
2046724 "Dropping sync committee message"
1583597 "No peers for topic, skipping publish"
 605694 "Received error response from handler"
  64881 "Contribution received"
  61218 "Contribution sent"
```
Probably they're failing in an ongoing way and the VC is spamming the BN with retry requests, or similar.

tersec:
Probably related to https://github.com/status-im/nimbus-eth2/issues/3828

arnetheduck:
"oops" :)

arnetheduck:
> Probably related to https://github.com/status-im/nimbus-eth2/issues/3828

this made the bug obvious, though there was a real underlying issue with return codes as well

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -502,8 +501,8 @@ proc runQueueProcessingLoop*(self: ref BlockProcessor) {.async.} =
       # sequence of
       # - newPayload(execution payload with block hash `h`) followed by
       # - forkchoiceUpdated(head = `h`)
-      # This is intrinsically somewhat optimistic, because determining the 
-      # validity of an execution payload requires the forkchoiceUpdated 
+      # This is intrinsically somewhat optimistic, because determining the
+      # validity of an execution payload requires the forkchoiceUpdated
```
tersec: https://github.com/status-im/nimbus-eth2/pull/3793 removes this entire comment block.

arnetheduck:
rebased

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -341,6 +341,25 @@ proc initFullNode(
...
       getLocalWallSlot, getFirstSlotAtFinalizedEpoch, getBackfillSlot,
       getFrontfillSlot, dag.backfill.slot, blockVerifier, maxHeadAge = 0)
+    router = (ref MessageRouter)(
+      processor: processor,
+      network: node.network,
+    )
+
+  if node.config.lightClientDataServe.get:
+    proc scheduleSendingLightClientUpdates(slot: Slot) =
+      if node.lightClientPool[].broadcastGossipFut != nil:
```
zah, line 351: Hrm, the logic here is slightly dubious.

Let's say this function is called for slot 10 and and it takes a while for the broadcast future to complete.
If it receives a request to broadcast slot 11 in the meantime, why should that request be dropped?

beacon_chain/validators/message_router.nim
```diff
@@ -0,0 +1,500 @@
...
+
+  let
+    sendTime = router[].processor.getCurrentBeaconTime()
+    delay = sendTime - attestation.data.slot.attestation_deadline()
+
+  block:
+    let res = await router[].network.broadcastAttestation(
+      subnet_id, attestation)
+    if not res.isOk:
+      notice "Attestation not sent",
```
zah, line 190: It would be good to add a metric for the `not_sent` case.

beacon_chain/validators/message_router.nim
```diff
@@ -0,0 +1,500 @@
...
+              headSyncCommittees.current_sync_committee, subcommitteeIdx):
+            let index = keysCur.getOrDefault(uint64(valKey), -1)
+            if index >= 0:
+              resIndices.add(index)
+              resFutures.add(router.routeSyncCommitteeMessage(
+                msgs[index], subcommitteeIdx, true))
+        for subcommitteeIdx in SyncSubcommitteeIndex:
+          for valKey in syncSubcommittee(
+              headSyncCommittees.next_sync_committee, subcommitteeIdx):
+            let index = keysNxt.getOrDefault(uint64(valKey), -1)
```
zah, line 355: The sync committee message is tied to a slot (and thus to a particular sync committee). Here, you seem to be accepting the message if the validator is sitting on the current or the next sync committee which doesn't seem right.

beacon_chain/nimbus_beacon_node.nim
arnetheduck: I don't disagree, but this is copied from the current code so fixing it is a different PR - cc @etan-status

beacon_chain/validators/message_router.nim
arnetheduck: we have one, at the libp2p level

beacon_chain/validators/message_router.nim
arnetheduck: I don't disagree, but this is also a 1:1 copy from the existing code - it might have been written at a time where the off-by-one scenarios of the `slot` were not well understood  where checking the next committee would work around the problem

zah: I don't recognize this code at all, so I'll have to do some git archeology to see how it ended up here.

arnetheduck: I thought about rewriting but decided against (to keep this PR focused) - if it was rewritten, it should probably handle only one message at a time - the multiplexing can happen on the rest side, like it does for attestations - this helper would stay, in order to find the correct subnet to send on, just like attestations - in both cases, internal validators already know the subnet from the duty selection process but for external validators don't, so we must look it up

zah (approved):

PR #3830, opened 2022-06-30
avoid unnecessary async copies in broadcast

arnetheduck:


beacon_chain/networking/eth2_network.nim
```diff
@@ -2419,31 +2419,30 @@ proc addAsyncValidator*[MsgType](node: Eth2Node,
...
-    # This is only for messages we create. A message this large amounts to an
-    # internal logic error.
-    doAssert uncompressed.len <= maxGossipMaxSize()
+  snappy.encode(uncompressed)
 
-    let compressed =
-      try: snappy.encode(uncompressed)
-      except InputTooLarge:
-        raiseAssert "More than 4gb? not likely.."
+proc broadcast(node: Eth2Node, topic: string, msg: seq[byte]):
```
zah, line 2430: Is the claim that this PR is reducing the number of copies? There will be still some copying here (both `topic` and `msg` will be copied)

arnetheduck: it avoids the (extremely slow) copy of `msg` via `genericAssign` as well as the lifetime extension of keeping copies of the unserialized object and `uncompessed` in the closure environment

zah: We can perhaps start optimizing away even the `string` and `seq` copies like the ones here by using more `ref` parameters.

arnetheduck: not worth it imo, this is by far not the biggest culprit (most copying happens in libp2p which would require major changes) - this was just a small drive-by fix that doesn't change semantics or introduce extra mutability like ref would - it gets rid of the extra try/catch layers that also were introducing overhead and noise (no longer needed thanks to the new snappy impl)

PR #3832, opened 2022-06-30
merge LC db into main BN db

etan-status:
To treat derived LC data similar to derived state caches, merge it into
the main beacon node DB.


etan-status:
Note that this will create the new LC tables even when LC data collection is disabled (they won't be populated though), because `BeaconChainDB` does not know about the config. Not sure if that's a problem, or if we should pass that info in there as well.

etan-status:
Screenshot 2022-07-01 at 08 17 02

PR #3836, opened 2022-07-02
Implement SSZ responses in the light client REST APIs

zah:
Other changes:

* More DRY encoding of the Nimbus content type preference.
* Switch if/elif/else to exhaustive case statements to guard
  the code better from future changes.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -785,22 +785,18 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
       return RestApiResponse.jsonError(
         Http404, BlockNotFoundError, "v1 API supports only phase 0 blocks")
 
-    let contentType =
-      block:
-        let res = preferredContentType(jsonMediaType,
-                                       sszMediaType)
-        if res.isErr():
-          return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
-        res.get()
+    let responseType = request.pickResponseType().valueOr:
+      return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
 
     return
-      if contentType == sszMediaType:
+      case responseType
+      of sszResponseType:
         var data: seq[byte]
         if not node.dag.getBlockSSZ(bid, data):
           return RestApiResponse.jsonError(Http404, BlockNotFoundError)
 
         RestApiResponse.response(data, Http200, $sszMediaType)
-      elif contentType == jsonMediaType:
+      of jsonResponseType:
         let bdata = node.dag.getForkedBlock(bid).valueOr:
           return RestApiResponse.jsonError(Http404, BlockNotFoundError)
 
@@ -811,8 +807,6 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
           # issue..
           RestApiResponse.jsonError(
             Http404, BlockNotFoundError, "v1 API supports only phase 0 blocks")
-      else:
-        RestApiResponse.jsonError(Http500, InvalidAcceptError)
 
   # https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockV2
   router.api(MethodGet, "/eth/v2/beacon/blocks/{block_id}") do (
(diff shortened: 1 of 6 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -785,22 +785,18 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-        if res.isErr():
-          return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
-        res.get()
+    let responseType = request.pickResponseType().valueOr:
+      return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
 
     return
-      if contentType == sszMediaType:
+      case responseType
+      of sszResponseType:
```
etan-status, line 793: jsonError for SSZ response? The other endpoints also seem to do it this way though.

beacon_chain/rpc/rest_light_client_api.nim
```diff
@@ -69,16 +81,26 @@ proc installLightClientApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
       if update.isSome:
         updates.add update.get
-    return RestApiResponse.jsonResponse(updates)
+
+    return
+      case responseType
+      of jsonResponseType:
+        RestApiResponse.jsonResponse(updates)
+      of sszResponseType:
+        RestApiResponse.sszResponse(updates.asSszList(MAX_CLIENT_UPDATES))
```
etan-status, line 90: Updates may be from different fork digests in the future, i.e., each `update` may become a `ForkedLightClientUpdate`. The response should be future proofed for that, similar to the libp2p protocol which already did this.

See https://github.com/ethereum/beacon-APIs/pull/181#issuecomment-1172910910

And, originally, point 3.3 from https://github.com/ethereum/consensus-specs/pull/2802#issuecomment-1115457016

beacon_chain/rpc/rest_utils.nim
```diff
@@ -310,3 +310,40 @@ const
...
+type
+  ResponseContentType* = enum
+    # Please note that the order of content types here determines
+    # the order of preference for the returned result by Nimbus.
+    # This Nimbus preference is used when the request doesn't
+    # explicitly state another preference.
+    jsonResponseType = "application/json"
+    sszResponseType = "application/octet-stream"
+
+proc pickResponseType*(request: HttpRequestRef): Result[ResponseContentType, void] =
```
etan-status, line 323: Would be great to introduce this for the other APIs as well for consistency (maybe separate PR ahead of this)

etan-status (commented):
Thanks for adding SSZ support! This was indeed still missing.

The `nim-stew` bump intentional?

beacon_chain/rpc/rest_utils.nim
etan-status: Ah, nvm, this already covers all of the other endpoints that support SSZ.

etan-status:
LC events are in `rest_event_api.nim`, but it seems there is no SSZ support there in general.

arnetheduck:
the risk of overimplementing unspecced stuff is that the specced stuff later becomes incompatible .. because this is not in the spec, right?

etan-status:
Well, the LC specific endpoints are `/eth/v0` or `_v0` (events), there is no adopted spec so far.
https://github.com/ethereum/beacon-APIs/pull/181

etan-status:
The problem with the fork digest also exists for other endpoints, but for those only an individual object is returned in the response, so heuristics can be applied such as peeking the first 8 bytes for slot number etc, or try which deserializations succeed and which don't  (that sounds error prone but seems to work well in practice for existing use).

What's new in the LC spec is that there is a way to request entire ranges of updates, whereas for the blocks the REST API only allows single block download per request. Requiring a light client to perform multiple round trips to be consistent with this existing behaviour is kinda dumb (also considering that LC updates are much smaller than blocks, so the roundtrips add significant overhead), so a new solution to exchange SSZ lists of potentially different objects over REST should be designed. Of course, this only starts mattering once there actually are different objects, but at the very least the mechanism should be future-proof to that.

For the other endpoints (finality / optimistic updates), they also have the problem of not including fork context with them, but at the very least there it can be argued that there is existing prior art of just attempting deserializations / heuristics which seems to be acceptable for now.

For JSON response, the problem is way smaller as JSON does not enforce a specific memory layout, e.g. if the slot is moved in some version the client can still find it and infer fork digest from it. But for SSZ, the payload does not itself describe where each property is located.

etan-status:
SSZ is now supported as of https://github.com/ethereum/beacon-APIs/pull/247 and implemented in Nimbus in https://github.com/status-im/nimbus-eth2/pull/4213

Note that the event stream uses SSE which is a text encoding, so only supports JSON. This is in line with the other routes.

This PR here still has unrelated adjustments that may still be useful. Suggest untying them from the LC related parts if still relevant.

PR #3837, opened 2022-07-04
api: add skip_randao_verification for produceBlockV2

michaelsproul:
Implement https://github.com/ethereum/beacon-APIs/pull/222.

If my Nim isn't too atrocious hopefully I might be able to get this to a mergeable state. I'm putting it up as a draft for _gentle_ feedback (this is my first time ever writing Nim).

Does the broad angle of attack seem acceptable? My next step would be to work out how to put a zero `randao_reveal` in when none is provided, and to add a new `UpdateFlag` to disable verification of the RANDAO signature only.


beacon_chain/extras.nim
```diff
@@ -24,6 +24,9 @@ type
     ## Skip verification of BLS signatures in block processing.
     ## Predominantly intended for use in testing, e.g. to allow extra coverage.
     ## Also useful to avoid unnecessary work when replaying known, good blocks.
+    skipRandaoValidation ##\
+    ## Skip verification of the proposer's randao reveal in block processing.
+    ## This is implied by `skipBblsValidation`.
```
michaelsproul:
```suggestion
    ## This is implied by `skipBlsValidation`.
```

arnetheduck:
> this is my first time ever writing Nim

welcome to the bright side - fortunately, it's not that easy to write atrocious Nim code (which perhaps is the main raison d'etre of it), and yours is already looking quite good ;)

I guess there are two things to look at here: is the PR good, and is the feature desired in general  - I'll focus on the former and then we can have the latter discussion in the spec PR, upon which this PR would depend :)

beacon_chain/spec/state_transition_block.nim
```diff
@@ -85,7 +85,7 @@ proc process_randao(
   let
     epoch = state.get_current_epoch()
 
-  if skipBlsValidation notin flags:
+  if skipBlsValidation notin flags and skipRandaoValidation notin flags:
```
arnetheduck: instead of skipping, it probably makes sense to target a specific signature, like the infinity point  - this lowers the risk of "random" signatures being propagated in the system in general

beacon_chain/extras.nim
```diff
@@ -24,6 +24,9 @@ type
     ## Skip verification of BLS signatures in block processing.
     ## Predominantly intended for use in testing, e.g. to allow extra coverage.
     ## Also useful to avoid unnecessary work when replaying known, good blocks.
+    skipRandaoValidation ##\
```
arnetheduck: not for this PR of course where consistency would be the right thing, but as a curiosity, `##\` is a leftover from old nim versions and in new code it's no longer needed (we'll get rid of it here at some point as well)

beacon_chain/spec/state_transition_block.nim
arnetheduck:
```suggestion
  if {skipBlsValidation, skipRandaoValidation} * flags != {}:
```

can use a set intersection here, thought you might enjoy at least some cryptic syntax :) that said, if we go for infinity, we'll need different logic here of course - `skipBlsValidation` is there for the test suite mainly which has all kinds of junk

narimiran: If those two are *not* in `flags`, the intersection *should be* empty, no?

`if {skipBlsValidation, skipRandaoValidation} * flags == {}:`

arnetheduck: right ;) good catch!

michaelsproul: I _do_ enjoy the cryptic syntax. And I guess I can infer from this that set union is `+` :relieved: 

I think your point about only accepting the Inf signature makes sense though, I'll amend the PR along those lines

narimiran: > And I guess I can infer from this that set union is `+`

Correct. And a set difference is, you guessed it, `-`; a subset is `<`.

arnetheduck:
@michaelsproul since https://github.com/ethereum/beacon-APIs/pull/222 is merged, we can go ahead with this PR as well as soon as it's rebased / updated to the new param names etc

michaelsproul:
I've updated this PR to use the `skip_randao_verification` flag that was decided upon in the `beacon-APIs` repo. I've tested it manually and it seems to be working well. If you'd like me to add some automated tests please point me in the right direction.

arnetheduck:
>  If you'd like me to add some automated tests please point me in the right direction.

The relevant location is here: https://github.com/status-im/nimbus-eth2/blob/2ad5239faebb3b55ec148a6b4bdd0fdb0a06a419/tests/testblockutil.nim#L99: we need to teach this function to take a valid/invalid randao and pass on the new randao flag, then teach https://github.com/status-im/nimbus-eth2/blob/2ad5239faebb3b55ec148a6b4bdd0fdb0a06a419/tests/test_blockchain_dag.nim#L30 to call it with the various combinations - although the suite creates some dummy blocks, a new `test` in it will do fine.

arnetheduck:
Thanks for the feature and the upstream spec! I'll merge this now and throw in some tests in a separate PR.

michaelsproul:
Thanks for having me on as a novice Nimbus contributor! I'll wear the badge with pride :grin:

PR #3842, opened 2022-07-06
less async optimistic sync

tersec:
It keeps the fcU-await-then-dag.updateHead flow but doesn't require `storeBlock` or its callers to wait.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -164,7 +164,6 @@ proc storeBackfillBlock(
 
   res
 
-from chronicles/chronos_tools import traceAsyncErrors
 from ../consensus_object_pools/attestation_pool import addForkChoice
 from ../consensus_object_pools/spec_cache import get_attesting_indices
 from ../spec/datatypes/phase0 import TrustedSignedBeaconBlock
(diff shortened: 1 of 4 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -264,11 +264,8 @@ proc storeBlock*(
     # called valid blocks have already been registered as verified. The head
     # can lag a slot behind wall clock, complicating detecting synced status
     # for validating, otherwise.
-    #
-    # TODO have a third version which is fire-and-forget for when it is merge
-    # but payloadValid is true, i.e. fcU is for EL's benefit, not CL. Current
-    # behavior adds unnecessary latency to CL event loop.
-    await self.consensusManager.updateHeadWithExecution(wallTime.slotOrZero)
+    traceAsyncErrors self.consensusManager.updateHeadWithExecution(
```
arnetheduck: is it a problem if uhwe calls get reordered in-flight? also, tendentially, `asyncSpawn` is the better version, with a `try/catch` inside `uhwe`

arnetheduck (approved):

beacon_chain/gossip_processing/block_processor.nim
tersec: It's a potential issue -- probably somewhat okay (if they're re-ordered, might as well fcU to the later block if one has it), but it'd speak to a missed opportunity to have fcU'd to an earlier block earlier. So not, in itself, bad, but indicative of suboptimal functioning.

tersec: > also, tendentially, `asyncSpawn` is the better version, with a `try/catch` inside `uhwe`

https://github.com/status-im/nimbus-eth2/pull/3842/commits/afe719e244940135bbb7eebd33cfbbc45f496dd4

PR #3851, opened 2022-07-09
MEV boost script for ropsten and sepolia builder network

tersec:
Slight MEV protocol update in how validator registration works and completely replacing the MEV test script from using `mergemock`, which was not particularly realistic: it needs `forkchoiceUpdated` calls to its EL-part which don't make sense in a MEV context, but doesn't need `registerValidator`, and does less strict validation than what's on ropsten and sepolia.

2022-07-09-224805_943x1022_scrot


scripts/test_mev_boost.nim
```diff
@@ -1,64 +1,131 @@
...
+    restUrl: "https://builder-relay-sepolia.flashbots.net/",
+    runtimeConfig: sepoliaMetadata.cfg,
+    genesisValidatorsRoot: Eth2Digest.fromHex(
+      "0xd8ea171f3c94aea21ebc42a1ed61052acf3f9209c00e4efbaaddac09ed9b8078"),
+    builderSigningDomain: Eth2Domain.fromHex(
+      "0x00000001d3010778cd08ee514b08fe67b6c503b510987a4ce43f42306d97c67c"),
+    proposerSigningDomain: Eth2Domain.fromHex(
+      "0x0000000036fa50131482fe2af396daf210839ea6dcaaaa6372e95478610d7e08")
+  )
+  networkInfo = sepoliaInfo
```
etan-status: `networkInfo` should probably become a parameter for the other functions, rather than a global.

`DeploymentPhase.Devnet` currently points to `Ropsten` / `Sepolia`, in case that's any use for filtering feature availability here.

tersec: https://github.com/status-im/nimbus-eth2/pull/3851/commits/e38c8491a1871249e724aa41e75a1efb0ff157e7 changes `networkInfo` from being a global variable.

As far as `DeploymetPhase.Devnet`, this is rather instead tied to Flashbot's schedule/deployments, rather than testnets more generally. They'll generally line up, but these are custom network definitions for a specific purpose. Unless https://builder-relay-ropsten.flashbots.net/ / https://builder-relay-sepolia.flashbots.net/ or similar is available for a network, it's not listed here. Goerli will, presumably, gain such a builder after it merges next, but whether it's included in this script or not will be a function of that, not any other definition or filter.

PR #3855, opened 2022-07-11
ncli verifyDeposit: a simple helper for inspecting deposit events

zah:


tersec:
This `nim-web3` bump isn't safe: https://github.com/status-im/nimbus-eth2/pull/3850

That's an issue in `nim-web3`, and I'll revert the changes there. This `Quantity`  thing is  a real spec divergence, but also, shouldn't block other things in the meantime.

Also, because of the other spec ambiguities detailed there, already reported, it's not clear exactly what "correct" handling of the "Quantity" `difficulty`/`totalDifficulty` would be, so it can't be made unambiguously correct, pending JSON-RPC clarifications.

None of which is directly relevant here, just, why the fix isn't just-add-one-more-change-to-`ethtypes` instead of revert.

tersec:
It shouldn't need a full revert -- there's one specific condition which causes problems currently due to the somewhat less-than-rigorous `Quantity` usage (e.g., for various hashes which are definitely not defined as integer quantities, too): https://github.com/status-im/nim-web3/pull/55

ncli/ncli.nim
```diff
@@ -7,15 +7,18 @@ import
   ../beacon_chain/spec/eth2_apis/eth2_rest_serialization,
   ../beacon_chain/spec/datatypes/[phase0, altair, bellatrix],
   ../beacon_chain/spec/[
-    eth2_ssz_serialization, forks, helpers, state_transition],
-  ../beacon_chain/networking/network_metadata
+    eth2_ssz_serialization, forks, helpers, state_transition, signatures],
+  ../beacon_chain/networking/network_metadata,
+  ../beacon_chain/conf/eth2_types_confutils_defs,
+  ../beacon_chain/eth1/deposit_contract_abi
```
arnetheduck, line 13: this import is quite unfortunate: it brings in the entirerty of web3 which transitively imports lots and lots of completely irrelevant stuff just to get access to a trivial byte converter

tersec:
Due to EIP-6110, removing support for deposit events, and definitely not adding new support.

PR #3858, opened 2022-07-12
enforce epoch alignment for --finalized-checkpoint-state/block

tersec:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -489,6 +494,13 @@ proc init*(T: type BeaconNode,
       except IOError as err:
         fatal "Failed to load the checkpoint block", err = err.msg
         quit 1
+
+      withBlck(checkpointBlock):
+        if not blck.message.slot.is_epoch:
+          fatal "--finalized-checkpoint-block must point to a block for an epoch slot",
+            slot = blck.message.slot
+          quit 1
```
```suggestion
      if not checkpointBlock.slot.is_epoch:
        fatal "--finalized-checkpoint-block must point to a block for an epoch slot",
          slot = checkpointBlock.slot
        quit 1
```
To visually align with the implementation for `checkpointState` (which also reads out `getStateField` twice).

etan-status (approved):
Even better would be if this actually worked, but at least this describes the current implementation accurately.

tersec:
> Even better would be if this actually worked, but at least this describes the current implementation accurately.

Well, if one currently tries this, there's an assertion deep in the DAG, which isn't particularly informative to typical end-users.

PR #3861, opened 2022-07-12
attestation target epochs are epochs of attestation head

tersec:
https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#ffg-vote
> Set `attestation_data.target = Checkpoint(epoch=get_current_epoch(head_state), root=epoch_boundary_block_root)` where `epoch_boundary_block_root` is the root of block at the most recent epoch boundary.

Aside from renaming `bs` to the more informative `attestationHead` (what its callers name it), `current_epoch` is not the correct `target.epoch`.

### DIFF
--- a/beacon_chain/consensus_object_pools/spec_cache.nim
+++ b/beacon_chain/consensus_object_pools/spec_cache.nim
@@ -179,18 +179,18 @@ proc is_valid_indexed_attestation*(
   ok()
 
 func makeAttestationData*(
-    epochRef: EpochRef, bs: BlockSlot,
+    epochRef: EpochRef, attestationHead: BlockSlot,
     committee_index: CommitteeIndex): AttestationData =
-  ## Create an attestation / vote for the block `bs` using the
+  ## Create an attestation / vote for the block `attestationHead` using the
   ## data in `epochRef` to fill in the rest of the fields.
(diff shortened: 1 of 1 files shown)


beacon_chain/consensus_object_pools/spec_cache.nim
```diff
@@ -179,29 +179,29 @@ proc is_valid_indexed_attestation*(
...
   # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/phase0/validator.md#attestation-data
   AttestationData(
     slot: slot,
     index: committee_index.asUInt64,
-    beacon_block_root: bs.blck.root,
+    beacon_block_root: attestationHead.blck.root,
     source: epochRef.checkpoints.justified,
     target: Checkpoint(
-      epoch: current_epoch,
+      epoch: attestationHead.slot.epoch,
```
etan-status, line 204: `current_epoch` was `slot.epoch()`, and `slot` is `attestationHead.slot`. Not sure I understand the problem being fixed here.

tersec: Oh, indeed.

(There is a problem though, closing this, will reconsider approaches.)

PR #3862, opened 2022-07-12
handle discrepancy between subnet tracking/nextaction tracking

tersec:
Fixes a crash/regression.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1018,7 +1018,15 @@ proc trackCurrentSyncCommitteeTopics(node: BeaconNode, slot: Slot) =
   node.network.updateSyncnetsMetadata(currentSyncCommitteeSubnets)
 
 func getNextSyncCommitteeSubnets(node: BeaconNode, epoch: Epoch): SyncnetBits =
-  let epochToSyncPeriod = nearSyncCommitteePeriod(epoch)
+  let epochsToSyncPeriod = nearSyncCommitteePeriod(epoch)
+
+  # The end-slot tracker might call this when it's theoretically applicable,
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1018,7 +1018,15 @@ proc trackCurrentSyncCommitteeTopics(node: BeaconNode, slot: Slot) =
...
 func getNextSyncCommitteeSubnets(node: BeaconNode, epoch: Epoch): SyncnetBits =
-  let epochToSyncPeriod = nearSyncCommitteePeriod(epoch)
+  let epochsToSyncPeriod = nearSyncCommitteePeriod(epoch)
+
+  # The end-slot tracker might call this when it's theoretically applicable,
+  # but more than SYNC_COMMITTEE_SUBNET_COUNT epochs from when the next sync
+  # committee period begins, in which case `epochsToNextSyncPeriod` is none.
+  if  epochsToSyncPeriod.isNone or
+      forkVersionAtEpoch(node.dag.cfg, epoch + epochsToSyncPeriod.get) ==
+        node.dag.cfg.GENESIS_FORK_VERSION:
```
etan-status: Semantically this probably means >= Altair as that is when sync committees got introduced.
```suggestion
      node.dag.cfg.stateForkAtEpoch(epoch + epochsToSyncPeriod.get) <
        BeaconStateFork.Altair
```

If yes, there's another occurrence of this further down.

etan-status (approved):

beacon_chain/nimbus_beacon_node.nim
tersec: Yeah, that's the intent -- early exit if there's no sync committee at the point in the chain.

PR #3866, opened 2022-07-13
rename verifyFinalization internal flag to strictVerification

tersec:
It began as `verifyFinalization`, but it's more general now. The unifying theme is that it covers stricter reactions, e.g., quitting via assertions, that would be inappropriate in a production, or even live-testnet (ropsten, sepolia, prater, etc) setting but useful for debugging. Its main user is `launch_local_testnet.sh`, but in principle, it can be used in any situation where one wants to run `nimbus_beacon_node` or related commands such that, e.g., invalid signatures or other gossip `REJECT` conditions cause it to quit via assert.

This PR deliberately changes only the internal flag name; the command-line option will be a separate PR.

### DIFF
--- a/beacon_chain/extras.nim
+++ b/beacon_chain/extras.nim
@@ -26,7 +26,9 @@ type
     ## Also useful to avoid unnecessary work when replaying known, good blocks.
     skipStateRootValidation ##\
     ## Skip verification of block state root.
-    strictVerification
+    strictVerification ##\
+    ## Strictly assert on unexpected conditions to aid debugging.
+    ## Should not be used in production, as additional asserts are reachable.
     slotProcessed ##\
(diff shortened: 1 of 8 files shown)


beacon_chain/extras.nim
```diff
@@ -26,7 +26,7 @@ type
     ## Also useful to avoid unnecessary work when replaying known, good blocks.
     skipStateRootValidation ##\
     ## Skip verification of block state root.
-    verifyFinalization
+    strictVerification
```
etan-status:
```suggestion
    strictVerification ##\
    ## Strictly assert on unexpected conditions to aid debugging.
    ## Should not be used in production, as additional asserts are reachable.
```

etan-status (approved):
Good to see `finalization` gone :)

PR #3869, opened 2022-07-13
do not disconnect on receiving canceled request

etan-status:
LC cancels concurrent requests if one peer sent a correct response and
waiting for other peers is no longer useful. On the server side this
resulted in a descore (-500) and a likely disconnect. The behaviour is
changed to match `UnexpectedEOF`, `PotentiallyExpectedEOF` handling that
return an error response without disconnecting from the peer.


beacon_chain/networking/eth2_network.nim
```diff
@@ -1024,8 +1024,11 @@ proc handleIncomingStream(network: Eth2Node,
       try:
         awaitWithTimeout(
           readChunkPayload(conn, peer, maxChunkSize(MsgRec), MsgRec), deadline):
-            returnInvalidRequest(
-              errorMsgLit "Request full data not sent in time")
+            # Request was canceled, e.g., because a different peer fulfilled it.
```
tersec: Is this timeout always a canceled request, rather than some other reason?

etan-status: I guess the timeout could also be hit on slow connections, disconnections etc, i.e., network issues in general. But as we already have similar handling for other Eof behaviour further down that does not contain a descore, I guess it's fine to treat this similarly. I reworded the comment to mention a more broadly applicable "timeout" instead.

PR #3872, opened 2022-07-14
Fix a VC crash observed in the local_testnet simulation

zah:
It's not quite clear why this condition was triggered in the local
simulation, but it seems a viable scenario after the Keymanager API
is integrated in the validator client.

The user can temporarily remove all validator keys from a running
client before adding another set of keys.


zah:
The crash happened because `currentRoot` ended up uninitialized in the `while` loop at the beginning of the function. Later in the code, there are multiple unconditional `currentRoot.get` calls which resulted in `Defect` being raised.

PR #3874, opened 2022-07-14
Allow the user to use 'goerli' instead of 'prater'

zah:


arnetheduck:
This is somewhat problematic from a UX perspective in that if a user had a "prater" node and switches to using "goerli" (because they read that in some guide) , they'll lost their history and setup because of the different data directory names..

PR #3876, opened 2022-07-14
optimistic block gossip validation

tersec:
Nice-to-have, but a bit risky to merge before this release. Aiming for early next release cycle.

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -175,51 +175,31 @@ template checkedReject(error: ValidationError): untyped =
   err(error)
 
 template validateBeaconBlockBellatrix(
-       signed_beacon_block: phase0.SignedBeaconBlock |
-                            altair.SignedBeaconBlock,
-       parent: BlockRef): untyped =
+    signed_beacon_block: phase0.SignedBeaconBlock | altair.SignedBeaconBlock,
+    parent: BlockRef): untyped =
   discard
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.7/specs/merge/p2p-interface.md#beacon_block
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/p2p-interface.md#beacon_block
 template validateBeaconBlockBellatrix(
        signed_beacon_block: bellatrix.SignedBeaconBlock,
        parent: BlockRef): untyped =
   # If the execution is enabled for the block -- i.e.
   # is_execution_enabled(state, block.body) then validate the following:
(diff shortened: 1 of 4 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -175,51 +175,31 @@ template checkedReject(error: ValidationError): untyped =
...
+  # `is_merge_transition_complete(state) or is_execution_block(block)`
+  #
+  # `is_merge_transition_complete(state)` tests for
+  # `state.latest_execution_payload_header != ExecutionPayloadHeader()`, while
+  # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#block-processing
+  # shows that `state.latest_execution_payload_header` being default or not is
+  # exactly equivalent to whether that block's execution payload is default or
+  # not, so test cached block information rather than reconstructing a state.
+  if  signed_beacon_block.message.is_execution_block or
+      not dag.loadExecutionBlockRoot(parent).isZero:
```
etan-status, line 202: Is there any way how `dag.loadExecutionBlockRoot(parent)` could be zero, if `not signed_beacon_block.message.is_execution_block`? Wouldn't such a block be rejected even before this function?

etan-status: Ah, that would just be a pre-merge block. But, isn't the `signed_beacon_block.message.is_execution_block` enough to check?

tersec: Someone might try to send a non-merge block on top of a merge parent. This is/should be rejected (well, with some dubious nuances/caveats regarding optimistic sync, but it should at least be detected).

etan-status: But this function doesn't reject such a sequence, so I assume the rejection logic already applied before this is called, i.e., by the time this function is hit, `doAssert dag.loadExecutionBlockRoot(parent).isZero or signed_beacon_block.message.is_execution_block`

tersec: Sure it rejects such a sequence. If `not dag.loadExecutionBlockRoot(parent).isZero`, regardless of whether `signed_beacon_block.message.is_execution_block`, it will run its conditional body, which includes
```nim
if not (signed_beacon_block.message.body.execution_payload.timestamp ==
    timestampAtSlot):
  ...
  return errReject("BeaconBlock: mismatched execution payload timestamp")
```

It will `REJECT` such a non-execution block which attempts to build upon an execution block.

Of course, the next step would be to fake just the `timestamp`, but then it will get rejected elsewhere.

Issue #3683, opened 2022-05-31
VC: implement doppelganger protection


arnetheduck:
`--doppelganger-detection` missing from VC

85d023452494b348033be85033317ebd22799980

cheatfate:
I think commit which you put there in the issue is wrong.

arnetheduck:
> I think commit which you put there in the issue is wrong.

looks right to me? ie it's part of `stable`, which I was testing with

tersec:
https://github.com/status-im/nimbus-eth2/pull/3877

PR #3877, opened 2022-07-14
VC: doppelganger protection

cheatfate:
This PR should fix https://github.com/status-im/nimbus-eth2/issues/3683


beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -263,6 +263,66 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+      block:
+        if epoch.isErr():
+          return RestApiResponse.jsonError(Http400, InvalidEpochValueError,
+                                           $epoch.error())
+        let
+          res = epoch.get()
+          wallTime = node.beaconClock.now()
+          wallEpoch = wallTime.slotOrZero().epoch
+          nextEpoch =
+            if wallEpoch == Epoch(0xFFFF_FFFF_FFFF_FFFF'u64):
```
tersec: `FAR_FUTURE_EPOCH`

beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -263,6 +263,66 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+              wallEpoch
+            else:
+              wallEpoch + 1
+          prevEpoch =
+            if wallEpoch == Epoch(0):
```
tersec: `GENESIS_EPOCH`

tersec: But there's already a function `get_previous_epoch` which does this:
https://github.com/status-im/nimbus-eth2/blob/798fa69647d4a19a708f97bbe0e98e81cf76cc39/beacon_chain/spec/beacon_time.nim#L211-L217

beacon_chain/validator_client/common.nim
```diff
@@ -346,3 +367,93 @@ proc forkAtEpoch*(vc: ValidatorClientRef, epoch: Epoch): Fork =
...
+                      index: ValidatorIndex) =
+  if vc.config.doppelgangerDetection:
+    let
+      startEpoch = vc.currentSlot().epoch()
+      state = DoppelgangerState(startEpoch: startEpoch, epochsCount: 0'u64,
+                                lastAttempt: DoppelgangerAttempt.None,
+                                status: DoppelgangerStatus.Checking)
+      res = vc.doppelgangerDetection.validators.hasKeyOrPut(index, state)
+    if res:
+      warn "Validator is already in doppelganger's table", pubkey = pubkey,
```
tersec:
```suggestion
      warn "Validator is already in doppelganger table", pubkey = pubkey,
```

beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -263,6 +263,66 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        res
+    let qepoch =
+      block:
+        if epoch.isErr():
+          return RestApiResponse.jsonError(Http400, InvalidEpochValueError,
+                                           $epoch.error())
+        let
+          res = epoch.get()
+          wallTime = node.beaconClock.now()
+          wallEpoch = wallTime.slotOrZero().epoch
```
tersec: Elsewhere there's a
```nim
proc currentSlot*(vc: ValidatorClientRef): Slot =
  vc.beaconClock.now().slotOrZero()
```
defined which seems like could be used here? Not getting `wallTime` directly and just `wallEpoch = node.currentSlot.epoch` or similar.

beacon_chain/conf.nim
```diff
@@ -758,6 +758,15 @@ type
       abbr: "d"
       name: "data-dir" .}: OutDir
 
+    doppelgangerDetection* {.
+      desc: "If enabled, the validator client prudently listens for 2 epochs " &
+            "for attestations from a validator with the same index " &
+            "(a doppelganger), before sending an attestation itself. This " &
+            "protects against slashing (due to double-voting) but means you " &
+            "will miss two attestations when restarting."
```
tersec, line 768: Not sure how it's phrased elsewhere, but it's two attestations per validator, not two globally.

cheatfate: > Not sure how it's phrased elsewhere, but it's two attestations per validator, not two globally.

The only change i have made here is concatenation of strings to put everything in 80 characters per line. I have not changed text actually.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -756,3 +756,11 @@ proc prune*(pool: var AttestationPool) =
     # If pruning fails, it's likely the result of a bug - this shouldn't happen
     # but we'll keep running hoping that the fork chocie will recover eventually
     error "Couldn't prune fork choice, bug?", err = v.error()
+
+proc validatorSeenAtEpoch*(pool: var AttestationPool, epoch: Epoch,
+                           vindex: ValidatorIndex): bool =
+  if uint64(vindex) < uint64(len(pool.nextAttestationEpoch)):
+    let mark = pool.nextAttestationEpoch[int(vindex)]
```
tersec: This isn't a safe conversion at runtime, since `ValidatorIndex* = distinct uint32`. While that's not unique to this code, as they're frequently used as array indices (I think there have been discussions about making them some other type -- according to @arnetheduck they consume enough memory that making them a 64-bit integer type needlessly is wasteful, but I don't remember if/why regular `int32` wouldn't at least make this situation clearer rather than failing at random times).

`validatorIndexLimit = min(uint64(int32.high), VALIDATOR_REGISTRY_LIMIT)` in theory prevents this from being an issue, we also have
https://github.com/status-im/nimbus-eth2/blob/20d45e69b51655b42f0ea5781ff373c3b3dffdde/beacon_chain/spec/datatypes/base.nim#L652-L657 to centralize this rather than chase down random `ValidatorIndex` to `int` conversions exactly for array indexing later on, if this is changed. It's not (strictly) runtime-safe, but at least the lack of safety is localized in one bit of code.

So should just be able to use `pool.nextAttestationEpoch[vindex]`.

beacon_chain/validator_client/api.nim
```diff
@@ -10,6 +10,133 @@ type
...
+    try:
+      await vc.waitOnlineNodes(timerFut)
+      vc.onlineNodes()
+    except CancelledError:
+      var default: seq[BeaconNodeServerRef]
+      if not(isNil(timerFut)) and not(timerFut.finished()):
+        await timerFut.cancelAndWait()
+      default
+    except CatchableError:
+      # This case could not be happened.
```
tersec, line 44: If this truly can't happen, maybe log to check if it does? Are you sure enough about this to, e.g., `assert` or let it crash on a `Defect`?

beacon_chain/conf.nim
tersec: Ah, so at least there's not some duplicate copy of the text somewhere else which also needs updating.

beacon_chain/consensus_object_pools/attestation_pool.nim
etan-status: The conversion should be safe, becaues `len(...) <= high(int)`, and the `if` at 762 checks against `len`.

tersec: Also, separately,
https://github.com/status-im/nimbus-eth2/blob/6ff00a5694002b9a0a09eb5dedc8ee6d60c0fdc2/beacon_chain/spec/datatypes/base.nim#L724-L725
already captures this `uint64(len(foo))` pattern -- `uint64(someSignedInt)` isn't in general safe, but the characteristics of `len` in particular are safe. This is another one example of each case being checkable separately, but this means that remaining raw integer conversions can be more carefully scrutinized/searched for, by excluding some common and provably safe cases.

tersec: > The conversion should be safe, becaues `len(...) <= high(int)`, and the `if` at 762 checks against `len`.

That requires searching more involved reasoning than using the existing `ValidatorIndex` sequence access mechanisms. I agree, this code is fine, but does these manual integer type conversions which need to be verified manually, when there's already infrastructure in Nimbus to automate that away.

beacon_chain/validator_client/api.nim
```diff
@@ -10,6 +10,135 @@ type
...
+
+  var timerFut =
+    if timeout != InfiniteDuration:
+      sleepAsync(timeout)
+    else:
+      nil
+
+  let onlineNodes =
+    try:
+      await vc.waitOnlineNodes(timerFut)
```
zah:
```suggestion
      if not isNil(timerFut):
        await vc.waitOnlineNodes(timerFut)
```

beacon_chain/conf.nim
```diff
@@ -758,6 +758,15 @@ type
       abbr: "d"
       name: "data-dir" .}: OutDir
 
+    doppelgangerDetection* {.
+      desc: "If enabled, the validator client prudently listens for 2 epochs " &
```
zah, line 764:
```suggestion
      # TODO This description is shared between the BN and the VC.
      #      Extract it in a constant (confutils fix may be needed).
      desc: "If enabled, the validator client prudently listens for 2 epochs " &
```

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2097,6 +2097,52 @@ proc readValue*(reader: var JsonReader[RestJson],
     keystores: keystores, passwords: passwords, slashing_protection: slashing
   )
 
+proc writeValue*(writer: var JsonWriter[RestJson],
```
zah, line 2100: These overloads shouldn't be necessary.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -756,3 +756,11 @@ proc prune*(pool: var AttestationPool) =
     # If pruning fails, it's likely the result of a bug - this shouldn't happen
     # but we'll keep running hoping that the fork chocie will recover eventually
     error "Couldn't prune fork choice, bug?", err = v.error()
+
+proc validatorSeenAtEpoch*(pool: var AttestationPool, epoch: Epoch,
+                           vindex: ValidatorIndex): bool =
+  if uint64(vindex) < lenu64(pool.nextAttestationEpoch):
+    let mark = pool.nextAttestationEpoch[vindex]
```
tersec, line 763: This happens regardless of whether there are any validators the BN knows about. It's safe for the VC to rely on.

https://github.com/status-im/nimbus-eth2/blob/b7ed0aa23573803337461246220cf2327ad6e1c6/beacon_chain/gossip_processing/gossip_validation.nim#L547-L554
https://github.com/status-im/nimbus-eth2/blob/b7ed0aa23573803337461246220cf2327ad6e1c6/beacon_chain/gossip_processing/gossip_validation.nim#L752-L759

Issue #3200, opened 2021-12-15
Implement the required changes for interoperability with mev-boost


zah:
mev-boost is a new agent that will enable the Eth2 validators to optimize their proposed blocks for "maximum extracted value" (aka MEV) obtained by delegating the selection of execution layer transactions to the FlashBots network. The agent sits between the consensus and the execution layer client and communicates with both. You can find more about the overall architecture of mev-boost and Flashbots here:

https://ethresear.ch/t/mev-boost-merge-ready-flashbots-architecture/11177

The users of Nimbus will enable the use of mev-boost by supplying a `--mev-url` parameter. When provided, it will take precedence over any other `--web3-url` parameter for the purposes of execution engine communication. The communication protocol with the MEV agent is slightly different because it requires the consensus node to be able to sign a beacon block 'header" without seeing the full list of transactions that will be included in the block. Until now, the only pilot implementation of this functionality is present in the following Lighthouse branch:

https://github.com/realbigsean/lighthouse/tree/mev-lighthouse

arnetheduck:
Beacon API issue: https://github.com/ethereum/beacon-APIs/issues/179

tersec:
https://github.com/status-im/nimbus-eth2/pull/3883

PR #3883, opened 2022-07-17
MEV block proposal

tersec:
https://github.com/status-im/nimbus-eth2/issues/3200

This doesn't affect node operations unless `--payload-builder-enable` is specified.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -537,6 +537,11 @@ type
         desc: "Suggested fee recipient"
         name: "suggested-fee-recipient" .}: Option[Address]
 
+      payloadBuilder* {.
+        hidden
+        desc: "Payload builder URL"
+        name: "payload-builder" .}: Option[string]
+
     of BNStartUpCmd.createTestnet:
       testnetDepositsFile* {.
         desc: "A LaunchPad deposits file for the genesis state validators"
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -546,6 +546,7 @@ type
       payloadBuilderUrl* {.
         hidden
         desc: "Payload builder URL"
+        defaultValue: ""
         name: "payload-builder-url" .}: string
 
     of BNStartUpCmd.createTestnet:
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -537,10 +537,15 @@ type
         desc: "Suggested fee recipient"
         name: "suggested-fee-recipient" .}: Option[Address]
 
-      payloadBuilder* {.
+      payloadBuilderEnable* {.
+        hidden
+        desc: "Enable external payload builder"
+        name: "payload-builder-enable" .}: bool
+
+      payloadBuilderUrl* {.
         hidden
         desc: "Payload builder URL"
-        name: "payload-builder" .}: Option[string]
+        name: "payload-builder-url" .}: string
 
     of BNStartUpCmd.createTestnet:
       testnetDepositsFile* {.
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -540,6 +540,7 @@ type
       payloadBuilderEnable* {.
         hidden
         desc: "Enable external payload builder"
+        defaultValue: false
         name: "payload-builder-enable" .}: bool
 
       payloadBuilderUrl* {.
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -767,6 +767,17 @@ proc init*(T: type BeaconNode,
       max(Moment.init(bellatrixEpochTime, Second),
           Moment.now)
 
+  let payloadBuilderRestClient =
+    if config.payloadBuilder.isSome:
+      RestClientRef.new(
+          config.payloadBuilder.get,
+          httpFlags = {HttpClientFlag.NewConnectionAlways}).valueOr:
+        warn "Payload builder REST client setup failed",
+          payloadBuilderUrl = config.payloadBuilder.get
+        nil
+    else:
+      nil
+
   let node = BeaconNode(
     nickname: nickname,
     graffitiBytes: if config.graffiti.isSome: config.graffiti.get
@@ -777,6 +788,7 @@ proc init*(T: type BeaconNode,
     config: config,
     attachedValidators: validatorPool,
     eth1Monitor: eth1Monitor,
+    payloadBuilderRestClient: payloadBuilderRestClient,
     restServer: restServer,
     keymanagerServer: keymanagerServer,
     keymanagerToken: keymanagerToken,
@@ -798,7 +810,7 @@ proc init*(T: type BeaconNode,
 
   node
 
-func strictVerification(node: BeaconNode, slot: Slot) =
+func verifyFinalization(node: BeaconNode, slot: Slot) =
   # Epoch must be >= 4 to check finalization
   const SETTLING_TIME_OFFSET = 1'u64
   let epoch = slot.epoch()
@@ -1373,7 +1385,7 @@ proc onSlotStart(node: BeaconNode, wallTime: BeaconTime,
     wallSlot.epoch.toGaugeValue - finalizedEpoch.toGaugeValue)
 
   if node.config.strictVerification:
-    strictVerification(node, wallSlot)
+    verifyFinalization(node, wallSlot)
 
   node.consensusManager[].updateHead(wallSlot)
 
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -768,12 +768,12 @@ proc init*(T: type BeaconNode,
           Moment.now)
 
   let payloadBuilderRestClient =
-    if config.payloadBuilder.isSome:
+    if config.payloadBuilderEnable:
       RestClientRef.new(
-          config.payloadBuilder.get,
+          config.payloadBuilderUrl,
           httpFlags = {HttpClientFlag.NewConnectionAlways}).valueOr:
         warn "Payload builder REST client setup failed",
-          payloadBuilderUrl = config.payloadBuilder.get
+          payloadBuilderUrl = config.payloadBuilderUrl
         nil
     else:
       nil
--- a/beacon_chain/spec/signatures.nim
+++ b/beacon_chain/spec/signatures.nim
@@ -317,7 +317,6 @@ proc get_contribution_and_proof_signature*(
 
   blsSign(privkey, signing_root.data)
 
-
 # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/altair/validator.md#aggregation-selection
 func is_sync_committee_aggregator*(signature: ValidatorSig): bool =
   let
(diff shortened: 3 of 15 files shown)


tersec:
Nim 1.6 failures are, well, actual Nim 1.6 build failures:
```
Building: Nim compiler
make[2]: *** [build-nim] Error 1
vendor/nimbus-build-system/makefiles/targets.mk:77: recipe for target 'build-nim' failed
make[1]: *** [vendor/nimbus-build-system/vendor/Nim/bin/nim] Error 2
vendor/nimbus-build-system/makefiles/targets.mk:120: recipe for target 'vendor/nimbus-build-system/vendor/Nim/bin/nim' failed
vendor/nimbus-build-system/makefiles/targets.mk:103: recipe for target 'update-common' failed
make: *** [update-common] Error 2
```

tersec:
2022-07-17-210215_911x819_scrot
Probably related to this.

The `version-1-6` branch isn't entirely healthy in CI at the moment.

beacon_chain/conf.nim
```diff
@@ -532,6 +532,11 @@ type
         desc: "Suggested fee recipient"
         name: "suggested-fee-recipient" .}: Option[Address]
 
+      payloadBuilder* {.
+        hidden
+        desc: "Payload builder URL"
+        name: "payload-builder" .}: Option[string]
```
zah: Technically, it's possible to use multiple builders, right?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -778,6 +789,7 @@ proc init*(T: type BeaconNode,
     config: config,
     attachedValidators: validatorPool,
     eth1Monitor: eth1Monitor,
+    restClient: restClient,
```
zah: Shouldn't this name be more descriptive?

beacon_chain/spec/signatures.nim
```diff
@@ -333,3 +332,38 @@ proc verify_contribution_and_proof_signature*(
...
+
+  let domain = get_domain(
+    fork, DOMAIN_APPLICATION_BUILDER, GENESIS_EPOCH, ZERO_HASH)
+
+  # This is only supported for specific networks.
+  const supportedBuilderSigningDomains = [
+    # Ropsten
+    Eth2Domain.fromHex(
+      "0x00000001d5531fd3f3906407da127817ef33c71868154c6021bdaac6866406d8"),
+    # Sepolia
```
zah: Prater TTD has been accepted now, so we can add it to the list.

beacon_chain/spec/state_transition.nim
```diff
@@ -537,6 +540,33 @@ proc makeBeaconBlock*(
       rollback(state)
       return err(res.error())
 
+    # Override for MEV
+    if transactions_root.isSome and execution_payload_root.isSome:
+      withState(state):
+        when stateFork == BeaconStateFork.Bellatrix:
```
zah, line 547: It may be beneficial to turn this into an exhaustive case statement, so you are reminded to revisit the code when new forks arrive.

beacon_chain/validators/validator_duties.nim
```diff
@@ -476,6 +495,237 @@ proc makeBeaconBlockForHeadAndSlot*(node: BeaconNode,
...
+    block:
+      let res = await validator.getBlockSignature(
+        fork, genesis_validators_root, slot, blockRoot, blindedBlock.message)
+      if res.isErr():
+        return err("Unable to sign block: " & res.error())
+      res.get()
+
+  return ok blindedBlock
+
+# TODO temporary, don't merge with this still here
```
zah: Does this need to be removed now?

beacon_chain/validators/validator_duties.nim
```diff
@@ -905,6 +1170,76 @@ proc updateValidatorMetrics*(node: BeaconNode) =
...
+    validatorRegistration
+
+  if signature.isErr:
+    return err signature.error
+
+  validatorRegistration.signature = signature.get
+
+  return ok validatorRegistration
+
+proc registerValidators(node: BeaconNode) {.async.} =
```
zah, line 1202: Maybe the registrations can be evenly spread out through the period instead of producing a spike of CPU usage from time to time.

beacon_chain/validators/validator_pool.nim
```diff
@@ -210,9 +211,12 @@ proc getBlockSignature*(v: AttachedValidator, fork: Fork,
           fork, genesis_validators_root, slot, block_root,
           v.data.privateKey).toValidatorSig())
     of ValidatorKind.Remote:
-      let request = Web3SignerRequest.init(
-        fork, genesis_validators_root, blck.Web3SignerForkedBeaconBlock)
-      await v.signData(request)
+      when blck is BlindedBeaconBlock:
+        err "Web3 remote signing and MEV combination unsupported"
```
zah: Hrm, I think the web3signer protocol will be extended to handle this, so we can leave a TODO mark here.

zah: Update: I've talked to the web3signer team and they seem to support signing blinded blocks through a change that was introduced for Bellatrix (allowing a block header to be signed, instead of a full block):

https://github.com/ConsenSys/web3signer/pull/547/files

beacon_chain/conf.nim
tersec: One is supposed to do this with an external builder/relay setup, but yes, it's technically possible to incorporate that multiplexing into a CL. This was discussed at DevConnect and the recommendation was not to do this within the CL, however.

beacon_chain/validators/validator_duties.nim
tersec: I want to keep it for another day or two -- it's useful to track the Nimbus Sepolia MEV node: https://sepolia.beaconcha.in/blocks?q=Nimbus-MEV
2022-07-22-115539_946x1048_scrot

But sure, it can't be released or merged this way. I'll remove this after some of the other changes.

beacon_chain/validators/validator_pool.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3883/commits/74883198f74e918a5d87e8f2f802a5383af9e756

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3883/commits/74883198f74e918a5d87e8f2f802a5383af9e756

beacon_chain/conf.nim
tersec: In particular, what you're describing is the role of a "relay", e.g., https://github.com/flashbots/mev-boost/
> This project is part of the Flashbots research towards proposer/builder separation for Maximal Extractable Value (MEV) mitigation. `mev-boost` can connect to relays that aggregate multiple builders. The builders prepare full blocks, optimizing for MEV extraction and fair distribution of the rewards. The Consensus Layer client of the validator proposes the most profitable block received from `mev-boost`.

i.e. the question of whether a CL should be its own relay, or meta-relay. The protocol allows it -- there used to be a distinction between the "builder" and "relay" protocols, which is gone now, to allow composing these layers more freely -- but they're still generally better to outsource to more specialized software.

For example, the publicly available URLs are not of builders but relays:
> ### Sepolia testnet
>
> Run mev-boost pointed at our [Sepolia Relay](https://builder-relay-sepolia.flashbots.net/):
>
> ```
> ./mev-boost -sepolia -relays
> https://0x845bd072b7cd566f02faeb0a4033ce9399e42839ced64e8b2adcfc859ed1e8e1a5a293336a49feac6d9a5edb779be53a@builder-relay-sepolia.flashbots.net
```

File: beacon_chain/spec/signatures.nim (hunk 3 above)
Comment (tersec) [reply beacon_chain/spec/signatures.nim]: It's not really compatible until/unless either https://github.com/flashbots/mev-boost/#usage contains it -- that's currently only GSF5, Ropsten, Kiln, and Sepolia.

File: beacon_chain/validators/validator_pool.nim (hunk 7 above)
Comment (zah) [reply beacon_chain/validators/validator_pool.nim]: See comment above. Implementing this should already be possible.

File: beacon_chain/conf.nim (hunk 1 above)
Comment (zah) [reply beacon_chain/conf.nim]: That's a reasonable point, but I think users would still be concerned about losing connectivity to a particular relay (either due to a particular internet link going down or due to the relay going offline), so having the option to specify more than one may be desirable.

Comment (tersec) [reply beacon_chain/conf.nim]: https://github.com/flashbots/mev-boost should accomplish this, and it's already maintained and will be improved by Flashbots.

File: beacon_chain/spec/signatures.nim (hunk 3 above)
Comment (tersec) [reply beacon_chain/spec/signatures.nim]: https://github.com/status-im/nimbus-eth2/pull/3883/commits/be3ab6e97ead88c31d967fc77d43f6c246405bd9 removes this check entirely. It was initially there as a sanity check, but now that it's clear it basically works, better to let it naturally fail if the relay or builder specified rejects the (valid) signature for some reason, rather than manually maintaining a list.

There's no reasonable security concern, or feature-gating utility to this particular check.

File: beacon_chain/spec/state_transition.nim (hunk 4 above)
Comment (tersec) [reply beacon_chain/spec/state_transition.nim:547]:
```nim
      withState(state):
        case stateFork
        of BeaconStateFork.Phase0, BeaconStateFork.Altair:
          doAssert false,
            "makeBeaconBlock: CL MEV not supported in Phase 0 and Altair"
        of BeaconStateFork.Bellatrix:
          state.data.latest_execution_payload_header.transactions_root =
            transactions_root.get

          # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#beaconblockbody
          # Effectively hash_tree_root(ExecutionPayload) with the beacon block
          # body, with the execution payload replaced by the execution payload
          # header. htr(payload) == htr(payload header), so substitute.
          state.data.latest_block_header.body_root = hash_tree_root(
            [hash_tree_root(randao_reveal),
             hash_tree_root(eth1_data),
             hash_tree_root(graffiti),
             hash_tree_root(exits.proposer_slashings),
             hash_tree_root(exits.attester_slashings),
             hash_tree_root(List[Attestation, Limit MAX_ATTESTATIONS](attestations)),
             hash_tree_root(List[Deposit, Limit MAX_DEPOSITS](deposits)),
             hash_tree_root(exits.voluntary_exits),
             hash_tree_root(sync_aggregate),
             execution_payload_root.get])
```
results in
> nimbus-eth2/beacon_chain/spec/state_transition.nim(551, 21) Error: undeclared field: 'latest_execution_payload_header' for type altair.BeaconState [type declared in nimbus-eth2/beacon_chain/spec/datatypes/altair.nim(249, 3)]

because the `case` is more similar to the runtime `if` than compile-time `when`. Is there a compile-time `case`? Enclosing the whole thing in `static` shouldn't work, because it's not really static.

In the meantime, https://github.com/status-im/nimbus-eth2/pull/3883/commits/b1c3afa0254e76667dc3cbde6866b332914c24eb at least provides the reminder to check when capella starts getting supported to update this code.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3883/commits/7780e21b42baf6ca21fd8ee208d55e2e8602c410 adds a TODO here -- I agree with this as a goal, but not sure it works best as part of this PR.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3883/commits/7780e21b42baf6ca21fd8ee208d55e2e8602c410 removes this graffiti.

beacon_chain/validators/validator_pool.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3883/commits/b218a3a4bc43fd0b8812842e849e6962612a20c5

beacon_chain/conf.nim
```diff
@@ -537,6 +537,18 @@ type
         desc: "Suggested fee recipient"
         name: "suggested-fee-recipient" .}: Option[Address]
 
+      payloadBuilderEnable* {.
+        hidden
+        desc: "Enable external payload builder"
+        defaultValue: false
+        name: "payload-builder-enable" .}: bool
```
zah, line 544: We don't use this `-enable` suffix on other similar flags (e.g. `--rest:on`, `--keymanager:on`, `--metrics:on`, etc). `--light-client-enable` is the only existing precedent, but we haven't released this functionality officially yet, so we can fix it as well.

zah: https://github.com/status-im/nimbus-eth2/pull/3928
https://github.com/status-im/infra-role-beacon-node-linux/commit/df1059949f22a87f18c5641d4f35fb6c6e225318

etan-status: Note that macOS and Windows use separate `infra-role-beacon-node` repos and need separate updating.

PR #3887, opened 2022-07-18
classify prater as merged network

etan-status:
Moves `prater` to the group of to be merged testnets.


etan-status:
(don't include into 22.07 release, this is just preparations for the next release cycle)

PR #3888, opened 2022-07-18
Make CI logs great again

narimiran:
An (ongoing) attempt to greatly reduce the noise in the CI logs.
Scrolling through a zillion of `Hint: 'some_func' cannot raise 'Defect'` is no fun. And it crashes my browser.

### DIFF
--- a/beacon_chain/beacon_chain_db_light_client.nim
+++ b/beacon_chain/beacon_chain_db_light_client.nim
@@ -5,7 +5,11 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
+
 
 # This implements the pre-release proposal of the libp2p based light client sync
 # protocol. See https://github.com/ethereum/consensus-specs/pull/2802
--- a/beacon_chain/validators/slashing_protection_v2.nim
+++ b/beacon_chain/validators/slashing_protection_v2.nim
@@ -5,7 +5,10 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
 
 import
   # Standard library
@@ -237,7 +240,10 @@ func version*(_: type SlashingProtectionDB_v2): static int =
 # Internal
 # -------------------------------------------------------------
 
-{.push raises: [Defect].}
+when (NimMajor, NimMinor) < (1, 4):
+  {.push raises: [Defect].}
+else:
+  {.push raises: [].}
 logScope:
   topics = "antislash"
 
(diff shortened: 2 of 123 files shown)


arnetheduck:
as long as 1.2 is supported, this is the way we need to annotate - we can conditionally silence the warning on 1.6+ - @markspanbroek  has written some alternatives like https://github.com/markspanbroek/upraises/blob/main/upraises.nim but they introduce a lot of complexity for what ultimately is a cosmetic problem that will go away soon.

arnetheduck:
see also https://status-im.github.io/nim-style-guide/errors.exceptions.html

beacon_chain/beacon_chain_db_light_client.nim
```diff
@@ -5,7 +5,11 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [Defect].}
```
tersec, line 8: Would all of the `push raises` pragmas at the beginning of modules across `nimbus-eth2` be updated this way? It seems like a lot of churn, when hopefully within a month or two, they'll be able to be replaced by simply unconditional `{.push raises: [].}` after dropping Nim 1.2 support.

arnetheduck: I don't mind, fwiw - specially as long as every file in the project has the exact same header with licenses and stuff - we can simply use this across all libraries and modules - I sort of prefer low-tech solutions like this .. I did have a high-tech alternative in mind which would be `include stew/strict` or something like this which would allow us to set strict options for all our code, and potentially not limit it to push raises, but .. it's a bit over the top.

narimiran: > hopefully within a month or two (...) after dropping Nim 1.2 support

I've got an impression that won't happen any time soon.
And even if it does, this `when` statement can stay there.

In the mean time, we get much cleaner logs for the 1.6 branch.

arnetheduck: basically, we can get rid of this block at a time of our choosing with a simple search&replace in all files - at that point, it's easier that all files have the exact same "block" of stuff

narimiran:
Some numbers:

Currently (the latest commit in `unstable` branch), the CI log for the 1.6 branch has **107.000 lines**.
Removing every hint about `cannot raise 'Defect'` (including the ones in the `vendor` directory) leaves **16.000 lines**. 85% off.

beacon_chain/beacon_chain_db_light_client.nim
narimiran: > I did have a high-tech alternative in mind which would be include stew/strict or something like this

If this materializes one day, even after this PR the way to update to that solution is still:
> a simple search&replace in all files

----

So, IMO, no harm done, and it immediately improves the situation. By 85%. (see the comment below)

narimiran:
Rebased onto the latest `unstable`, and IMO ready for merging.

beacon_chain/validators/slashing_protection_v2.nim
```diff
@@ -237,7 +240,10 @@ func version*(_: type SlashingProtectionDB_v2): static int =
 # Internal
 # -------------------------------------------------------------
 
-{.push raises: [Defect].}
```
tersec, line 240: Doesn't the first `push raises` cover this?

narimiran: It should.

I was not sure why there was this duplication originally, so I left it as is. But you're right: now might be a good time to clean it up (if there really is no reason why this exists).

tersec: Either seems reasonable, though it should be cleaned up. But there's also value in this very mechanical approach, where it's a basically 1:1 equivalence before/after.

tersec (approved):

PR #3891, opened 2022-07-20
cleanup magic strings

etan-status:
Replaces hardcoded IP address strings with constants.
Adjusts some LC log levels.
Removes no longer used `RpcServer` type.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -47,6 +47,10 @@ const
   defaultSigningNodeRequestTimeout* = 60
   defaultBeaconNode* = "http://127.0.0.1:" & $DefaultEth2RestPort
 
+  defaultListenAddressDesc* = $defaultListenAddress
+  defaultAdminListenAddressDesc* = $defaultAdminListenAddress
+  defaultBeaconNodeDesc* = $defaultBeaconNode
+
 when defined(windows):
   {.pragma: windowsOnly.}
   {.pragma: posixOnly, hidden.}
@@ -224,7 +228,7 @@ type
       listenAddress* {.
         desc: "Listening address for the Ethereum LibP2P and Discovery v5 traffic"
         defaultValue: defaultListenAddress
-        defaultValueDesc: "0.0.0.0"
+        defaultValueDesc: defaultListenAddressDesc
         name: "listen-address" .}: ValidIpAddress
 
       tcpPort* {.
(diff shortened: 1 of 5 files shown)


beacon_chain/conf.nim
```diff
@@ -224,7 +228,7 @@ type
       listenAddress* {.
         desc: "Listening address for the Ethereum LibP2P and Discovery v5 traffic"
         defaultValue: defaultListenAddress
-        defaultValueDesc: "0.0.0.0"
+        defaultValueDesc: defaultListenAddressDesc
```
arnetheduck: did you check if this actually shows the string in the help? there was some issue with that, previously

kdeme: Indeed, I think it will show the content of the string only if you do here `$defaultListenAddressDesc`, else it prints just "defaultListenAddressDesc". In fluffy that silliness made it work: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/conf.nim#L64

arnetheduck: cc @zah is this fixable?

zah: I wasn't aware of the silly work-around - it may hold the key to making this work automatically.

etan-status: ec3200a5a2582ae0c25192c6307bb09868ecdcbd

beacon_chain/conf.nim
```diff
@@ -860,13 +864,13 @@ type
     bindPort* {.
       desc: "Port for the REST (BETA version) HTTP server"
       defaultValue: DefaultEth2RestPort
-      defaultValueDesc: "5052"
+      defaultValueDesc: DefaultEth2RestPortDesc
```
```suggestion
      defaultValueDesc: $DefaultEth2RestPortDesc
```

arnetheduck: why is this a capital D?

arnetheduck (approved):

beacon_chain/conf.nim
etan-status: Because the `defaultValue` also has capital D 🙈 I guess I'll just align them as well

beacon_chain/conf.nim
```diff
@@ -374,14 +378,14 @@ type
 
       restPort* {.
         desc: "Port for the REST server"
-        defaultValue: DefaultEth2RestPort
+        defaultValue: defaultEth2RestPort
```
etan-status, line 381: oh, one more to fix here (the 5052)

PR #3901, opened 2022-07-22
Implement the /eth/v1/validator/prepare_beacon_proposer end-point

zah:


arnetheduck:
needs a documentation update to highlight the new source of fee recipient

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -239,6 +239,10 @@ type
     epoch*: Epoch
     active*: bool
 
+  PrepareBeaconProposerBody* = object
+    validator_index*: ValidatorIndex
```
arnetheduck, line 243: `RestValidatorIndex` I believe?

beacon_chain/validators/validator_duties.nim
```diff
@@ -374,8 +378,11 @@ proc getExecutionPayload(
           terminalBlockHash
       latestFinalized =
         node.dag.loadExecutionBlockRoot(node.dag.finalizedHead.blck)
-      feeRecipient = node.config.getSuggestedFeeRecipient(pubkey).valueOr:
-        node.config.defaultFeeRecipient
+      dynamicFeeRecipient = node.dynamicFeeRecipientsStore.getDynamicFeeRecipient(
```
arnetheduck, line 381: fwiw, I *think* this can be written as `gDFR() or gSFR() or Opt.some(default)`

beacon_chain/spec/eth2_apis/rest_types.nim
zah: `RestValidatorIndex` was something that raised in prominence when we tried to introduce a wider `ValidatorIndex` reform (which wasn't merged). In the current code, it's used only for a single field and all other REST data types use the spec `ValidatorIndex` type.

zah:
> needs a documentation update to highlight the new source of fee recipient

Will be addressed with a separate PR against the `unstable` branch.

beacon_chain/validators/validator_duties.nim
zah: The `valueOr` construct effectively works as a right-associative operator while `or` is left-associative. For this particular purpose, right-associativity produces better code which is equivalent to a hand-written tree of if statements.

beacon_chain/spec/eth2_apis/rest_types.nim
arnetheduck: The wider VI is unrelated - `RestValidatorIndex` is used to differentiate between errors: a validator index that is larger than `2**32` but smaller than `2**41` (or something like that) is a valid but not yet assigned validator index - the `ValidatorIndex` parser is unable to distinguish these two cases, resulting in a poor error message, or in some requests, invalid responses (because some requests give an empty response for a not-yet-assigned validator index)

PR #3902, opened 2022-07-22
Use the latest Nim 1.6.7 by default

zah:


etan-status:
So far, compiler versions were selected globally as part of `nimbus-build-system`. What keeps us from reusing that approach?

arnetheduck:
we don't want an unreleased nim to be used in general - what's the plan here / reason for exception? that would be one reason to not push it to nbs as well, given how it's shared between projects and the "standard" is to use released versions.

env.sh
```diff
@@ -30,5 +30,7 @@ if [ -f "${USER_ENV_FILE}" ]; then
   set +o allexport
 fi
 
-source ${ABS_PATH}/vendor/nimbus-build-system/scripts/env.sh
+# Unless expliticly specified otherwise, use the latest Nim 1.6 as of 2022-07-22
+export NIM_COMMIT="${NIM_COMMIT:-5f61f1594d0c11caf0fb650e3d3bc32cf8f38890}"
```
tersec: Using explicit `NIM_COMMIT` overrides tends to create unclean `git status`es in NBS.

tersec:
```
$ git status && make update OVERRIDE=1 && ./env.sh nim --version 
On branch switch-to-nim-1-6
Your branch is up to date with 'origin/switch-to-nim-1-6'.

nothing to commit, working tree clean
[using Nim version 5f61f1594d0c11caf0fb650e3d3bc32cf8f38890]
Nim Compiler Version 1.2.16 [Linux: amd64]
Compiled at 2022-07-24
Copyright (c) 2006-2020 by Andreas Rumpf

git hash: c6a9f27b3e36bae9aacec1bd6c37893fb98fd33f
active boot switches: -d:release
```

zah:
> we don't want an unreleased nim to be used in general - what's the plan here / reason for exception? that would be one reason to not push it to nbs as well, given how it's shared between projects and the "standard" is to use released versions.

The latest released Nim version doesn't have the closure iterator fixes from @Menduist and we are aiming to switch to `1.6.8` as soon as it's available, so being closer to it reduces the likelihood of surprises.

zah:
> So far, compiler versions were selected globally as part of nimbus-build-system. What keeps us from reusing that approach?

In the past, we didn't have this additional option, so bumping the Nim commit inside nimbus-build-system was the only supported way. It's a bit more polluting towards other projects using nimbus-build-system, so I wanted to try this alternative method.

zah:
@tersec, I noticed this problem as well, but I believe it gets resolved the first time you run a `make` command such as `make nimbus_beacon_node`.

tersec:
> @tersec, I noticed this problem as well, but I believe it gets resolved the first time you run a `make` command such as `make nimbus_beacon_node`.

```
nimbus-eth2$ git checkout switch-to-nim-1-6 
Switched to branch 'switch-to-nim-1-6'
Your branch is up to date with 'origin/switch-to-nim-1-6'.
nimbus-eth2$ make update OVERRIDE=1
nimbus-eth2$ make nimbus_beacon_node 
Building: install/usr/lib/libbacktracenim.a
Building: build/generate_makefile
[using Nim version 5f61f1594d0c11caf0fb650e3d3bc32cf8f38890]
Build completed successfully: build/generate_makefile
Building: build/nimbus_beacon_node
[using Nim version 5f61f1594d0c11caf0fb650e3d3bc32cf8f38890]
nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1824, 7) Warning: Please use MetricsHttpServerRef API; startMetricsHttpServer is deprecated [Deprecated]
Build completed successfully: build/nimbus_beacon_node
nimbus-eth2$ ./env.sh nim --version
[using Nim version 5f61f1594d0c11caf0fb650e3d3bc32cf8f38890]
Nim Compiler Version 1.2.16 [Linux: amd64]
Compiled at 2022-07-24
Copyright (c) 2006-2020 by Andreas Rumpf

git hash: c6a9f27b3e36bae9aacec1bd6c37893fb98fd33f
active boot switches: -d:release
```
At no point did it appear to build this newer version of Nim (https://github.com/nim-lang/Nim/commit/5f61f1594d0c11caf0fb650e3d3bc32cf8f38890), either.

tersec:
I do have in my `.bashrc`:
```
MAKEFLAGS="-j$(nproc)"
export MAKEFLAGS
```

Not sure whether that should matter

zah:
Apparently, I did something to resolve this issue, but I'm not sure what that was, so let's stick to the beaten path - I've bumped nimbus-build-system in the usual way.

tersec:
Jenkins failure is just
```
./scripts/setup_scenarios.sh— Shell Script14s
[2022-07-25T22:17:09.271Z] + ./scripts/setup_scenarios.sh
[2022-07-25T22:17:09.271Z] Downloading consensus spec test vectors
script returned exit code 1
```
and not relevant to PR

PR #3904, opened 2022-07-22
fix concurrent Jenkins stages

etan-status:
The ports for the concurrently executing REST and Minimal testnet clash,
leading to some CI failures since #3827 introduced further concurrency.
Adjusting the ports to be distinct across various tests should fix this.


Makefile
```diff
@@ -182,12 +182,12 @@ local-testnet-mainnet:
 		--stop-at-epoch 5 \
 		--disable-htop \
 		--enable-logtrace \
-		--base-port $$(( 9100 + EXECUTOR_NUMBER * 100 )) \
-		--base-rest-port $$(( 7100 + EXECUTOR_NUMBER * 100 )) \
-		--base-metrics-port $$(( 8108 + EXECUTOR_NUMBER * 100 )) \
+		--base-port $$(( 39000 + EXECUTOR_NUMBER * 100 )) \
```
tersec: Might useful to avoid the https://dataplane.org/ephemeralports.html selections by common OSes to avoid a scenario where one process selects a port as a source port, then another the same port as a listening port.

17k, 18k, 19k, 26k, 27k, 28k, and 29k look fine according to that list, but 36k, 37k, 38k, and 39k look like they overlap Linux defaults, and I can confirm locally:
```
$ cat /proc/sys/net/ipv4/ip_local_port_range 
32768   60999
```

Putting aside FreeBSD (it eats enough ports this way I'd want that to be confirmed first), the safe port numbers according to that document seem to start at 5001 (avoid Windows ephemeral ports) and end at 32,767 (avoid Linux ephemeral ports).

If one stipulates the FreeBSD one is true, incorporating that leaves just 5001 through 9999 inclusive.

tersec (approved):

Makefile
etan-status: EL default net port is 30303, so I guess in practice, 32768+ is the safe range to avoid for eth.
Updated to condense into 5001-9999, though, it doesn't hurt. Needs revisiting if EXECUTOR_NUMBER ever is increased beyond 0/1.

etan-status: btw, source ports don't clash with listening ports. Connections are tracked as a 4-tuple (src-ip/src-port/dst-ip/dst-port), so you can reuse the same port as many times as you want as long as it is connected to different remotes. 

What can happen, though, is that a different process already selected the listening port randomly that we try to select for the test. So it is still good to avoid ephemeral port ranges for this reason.

Issue #3686, opened 2022-05-31
Prevent running multple BN:s and VC:s from the same data directory


arnetheduck:
This will cause keys to be loaded by both VC and BN resulting in high slashing risk (ie the slashing DB _may_ prevent it, but it's not certain)

This is specially true in the case of a VC and BN sharing data directory.

cheatfate:
This is not only VC problem, it is also problem of BN itself.

arnetheduck:
#3080 suggests one approach

It's indeed a problem both for VC:s and BN:s, though with the VC, it's very easy to put in the same data dir in the two processes by accident or misunderstanding.

zah:
This was fixed in the v22.8.0 release through the addition of keystore locking.

Issue #3080, opened 2021-11-10
Implement keystore file locking


zah:
When a validator is loaded by the beacon node or the validator client, Nimbus could place a lock on the `keystore.json` file. This can prevent the user from accidentally launching another Nimbus instance that loads the same keystore. Similar functionality is available in Lighthouse, so we can try to ensure interoperability between the clients as well. Care should be taken that we use OS APIs that will automatically release the lock on process termination.

zah:
Implemented in https://github.com/status-im/nimbus-eth2/pull/3907

PR #3907, opened 2022-07-24
Exclusive keystore locking

cheatfate:
Address #3686 and #3080
This PR requires https://github.com/status-im/nim-stew/pull/116 to be merged.

### DIFF
--- a/beacon_chain/filepath.nim
+++ b/beacon_chain/filepath.nim
@@ -13,10 +13,57 @@ else:
 
 import chronicles
 import stew/io2
-export io2
+import spec/keystore
 
 when defined(windows):
   import stew/[windows/acl]
+
+type
+  ByteChar = byte | char
+
+const
+  INCOMPLETE_ERROR =
+    when defined(windows):
+      IoErrorCode(996) # ERROR_IO_INCOMPLETE
+    else:
+      IoErrorCode(ENOSPC)
+
+proc openLockedFile*(keystorePath: string): IoResult[FileLockHandle] =
+  let
+    flags = {OpenFlags.Read, OpenFlags.Exclusive}
+    handle = ? openFile(keystorePath, flags)
+
+  var success = false
+  defer:
+    if not(success):
+      discard closeFile(handle)
+
+  let lock = ? lockFile(handle)
+  success = true
+  ok(FileLockHandle(ioHandle: lock, opened: true))
+
+proc getData*(lockHandle: FileLockHandle,
+              maxBufferSize: int): IoResult[string] =
+  let length = min(lockHandle.ioHandle.size, maxBufferSize)
+  var buffer = newString(length)
+  let bytesRead = ? readFile(lockHandle.ioHandle.handle, buffer)
+  if uint64(bytesRead) != uint64(len(buffer)):
+    err(INCOMPLETE_ERROR)
+  else:
+    ok(buffer)
+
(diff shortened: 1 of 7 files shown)


etan-status:
`Attempt to workaround macos issue`, this is triggered by LTO. 

```
         let cookedKey =
           block:
             let res = publicKey.load()
             if res.isNone():
               # Skip folders which has invalid ValidatorPubKey.
               continue
             res.get()

         yield cookedKey
```

With LTO, the `publicKey.load()` block is inlined, but only partially updates `res`. The final step that loads it into an `Option` 
using `some()` only initializes the option with zeroes, but then forgets to actually copy the computed result into it. Likewise, `cookedKey` is not updated with the result, and it happens to end up as a copy of `publicKey` instead.

Slightly changing the syntax makes LTO no longer inline `publicKey.load()`, and works around this problem. It is unclear, though, where else the LTO introduces such bugs.

The problem occurs using both `-flto=thin` and `-flto=full` and goes away when not using LTO.

Reproducible on macOS 12.5 with Xcode 13.4.1 on Nim compiler 1.6 (5f61f1594d0c11caf0fb650e3d3bc32cf8f38890) on an Intel MacBook Pro (16-inch, 2019) when doing a clean re-build:
```
rm -rf build nimcache test_keymanager_api && ulimit -n 1024 && \
    NIM_COMMIT=version-1-6 make -j test 
```
Disabling LTO fixes it (ensure clean re-build)
```
rm -rf build nimcache test_keymanager_api && ulimit -n 1024 && \
    NIM_COMMIT=version-1-6 make NIMFLAGS="-d:disableLTO" -j test
```

At this time, I have not identified a minimal program to reproduce this problem.

etan-status:
In `spec/crypto.nim`, tagging `func load*(v: ValidatorPubKey): Option[CookedPubKey]` as `{.inline.}` makes this also fail with `-d:disableLTO`, confirming the problem with inlining when not applying a workaround.

Curiously, `{.noinline.}` does not fully fix this, though. As expected, the `load` function remains un-inlined and the `Option` is correctly constructed. However, the `get()` after the `load` is still forgotten, so `cookedKey` is still incorrect, making `el` incorrect when the iterator is inlined into `listLocalValidators`.

```
proc listLocalValidators(validatorsDir,
                         secretsDir: string): seq[ValidatorPubKey] {.
     raises: [Defect].} =
  var validators: seq[ValidatorPubKey]
  try:
    for el in listLoadableKeys(validatorsDir, secretsDir,
                               {KeystoreKind.Local}):
      validators.add el.toPubKey()
  except OSError as err:
    error "Failure to list the validator directories",
          validatorsDir, secretsDir, err = err.msg
  validators
``` 

What works, though, is to drop the `lent` keyword for `proc get*[T](self: Option[T]): lent T {.inline.}` in `Nim/lib/pure/options.nim`. Without the `lent` keyword, even if `load` is annotated with `{.inline.}` explicitly, the tests succeed with LTO enabled!

etan-status:
Regarding `lent`, not sure if it is just a coincidence, as even putting a single `debugEcho` into `load` also works around the bug. Or, as in the original workaround, using a slightly different syntax instead of the `block` syntax.

etan-status:
After going through the generated C code thoroughly once more, this issue is actually rooted in the Nim --> C stage, not in C --> ASM! It is surfaced by aggressive optimization levels (UB), and promoted by inlining (which LTO increases).

Reported upstream with minimal example: https://github.com/nim-lang/Nim/issues/20107

PR #3910, opened 2022-07-25
Move from --rpc to --rest in docker-compose examples

hukkin:


tersec (approved):

docker/dist/binaries/docker-compose-example1.yml
```diff
@@ -28,9 +28,9 @@ services:
       --log-level=info
       --tcp-port=9000
       --udp-port=9000
-      --rpc
-      --rpc-address=0.0.0.0
-      --rpc-port=9190
+      --rest
+      --rest-address=0.0.0.0
+      --rest-port=9190
```
etan-status:
```suggestion
      --rest-port=5052
```

hukkin: If we change this, then the port forwarding above needs to change too.

hukkin: Committed the change. Arguably the example made more sense when configuring to a non-default port, but maybe it was confusing to use the legacy RPC port...

etan-status: Well, previously, the example was using the default port (9190 for RPC), so using 5052 now just makes it consistent with how it used to be.

PR #3912, opened 2022-07-25
launch_local_testnet: show processes killed by pkill

jakubgs:
I want to verify that processes being killed are not from other parallel jobs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ local-testnet-minimal:
 		--base-el-ws-port $$(( 6203 + EXECUTOR_NUMBER * 500 )) \
 		--base-el-auth-rpc-port $$(( 6204 + EXECUTOR_NUMBER * 500 )) \
 		--el-port-offset 5 \
-		--timeout 30 \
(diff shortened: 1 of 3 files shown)


jakubgs:
One thing worth noting is that `cleanup` method appears to be called twice:
```
21:58:43  Timeout reached. Aborting.
21:58:43  
21:58:43  ++ ps --forest -o pid,tty,stat,time,cmd -g 3374482
21:58:43      PID TT       STAT     TIME CMD
21:58:43  + cleanup
21:58:43  + set -x
21:58:43  + ps --forest -o pid,tty,stat,time,cmd -g 3374482
21:58:43      PID TT       STAT     TIME CMD
21:58:43  make: *** [Makefile:201: local-testnet-mainnet] Error 1
```
Not sure why though.

jakubgs:
Resolved in:

* https://github.com/status-im/nimbus-eth2/pull/3914

Caused by immediate assignment of `EXECUTOR_NUMBER` in `Makefile` instead of a lazy one.

PR #3928, opened 2022-08-01
Keep CLI options consistent by removing the '-enable' suffix from the outliers

zah:


tersec (approved):

etan-status (commented):
The `--light-client` change is alright; The flag is not used outside of developer and testing contexts for now. 

The flag may be renamed to something entirely different in the future, though, when we have actual use cases for it, e.g., trusting sync committees for syncing, so renaming this right now may not be useful in the end (but I'm also not opposed).

etan-status (approved):

PR #3932, opened 2022-08-02
split block storage

tersec:


tersec:
```
2022-08-02T07:01:14.1824551Z /home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/beacon_chain_db.nim: In function ‘getSplitBlock__beacon95chain95db_37951’:
2022-08-02T07:01:14.1826047Z /home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/beacon_chain_db.nim:903:200: error: dereferencing pointer to incomplete type ‘tyObject_TrustedSignedPartialBeaconBlock__nZPmS9bKi6dcEhRUFGxsqgg {aka struct tyObject_TrustedSignedPartialBeaconBlock__nZPmS9bKi6dcEhRUFGxsqgg}’
2022-08-02T07:01:14.1826904Z    if GetResult.found != db.partialBlocks[T.toFork].getSZSSZ(
2022-08-02T07:01:14.1827349Z                                                                                                                                                                                                         ^                                                                               
2022-08-02T07:01:14.2158988Z make[1]: *** [/home/runner/work/nimbus-eth2/nimbus-eth2/nimcache/release/nimbus_light_client/@mbeacon_chain_db.nim.c.o] Error 1
2022-08-02T07:01:14.2164643Z nimcache/release/nimbus_light_client/nimbus_light_client.makefile:1547: recipe for target '/home/runner/work/nimbus-eth2/nimbus-eth2/nimcache/release/nimbus_light_client/@mbeacon_chain_db.nim.c.o' failed
2022-08-02T07:01:14.2167366Z make: *** [nimbus_light_client] Error 2
2022-08-02T07:01:14.2167872Z Makefile:333: recipe for target 'nimbus_light_client' failed
```

beacon_chain/beacon_chain_db.nim
```diff
@@ -689,6 +703,17 @@ proc putBlock*(
 proc putBlock*(
     db: BeaconChainDB,
     value: bellatrix.TrustedSignedBeaconBlock) =
+  db.withManyWrites:
+    db.partialBlocks[type(value).toFork].putSZSSZ(
```
etan-status, line 719: Is `SZSSZ` still the intended format? The other forks use `SnappySSZ`. I recall one of the original reasons for `SZSSZ` was to allow serving blocks without having to uncompress and recompress between the different formats, but with the partial storage this now seems to be necessary anyway.

beacon_chain/beacon_chain_db.nim
```diff
@@ -452,6 +454,12 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("blocks").expectDb(),
       kvStore db.openKvStore("altair_blocks").expectDb(),
       kvStore db.openKvStore("bellatrix_blocks").expectDb()]
+    partialBlocks = [
+      nil, nil,
+      kvStore db.openKvStore("bellatrix_partialblocks").expectDb()]
+    executionPayloads = [
+      nil, nil,
+      kvStore db.openKvStore("bellatrix_executionpayloads").expectDb()]
```
etan-status, line 474: Hmm. If the end goal is to have the EL client store the `executionpayloads`, one could also just change the type of `bellatrix_blocks` to the partial format, without introducing the two additional tables. Or is the idea to first store them in `_executionpayloads` and then syncing them to the EL client later, then dropping them from our table?

beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -342,6 +342,39 @@ type
...
+    # Operations
+    proposer_slashings*: List[TrustedProposerSlashing, Limit MAX_PROPOSER_SLASHINGS]
+    attester_slashings*: List[TrustedAttesterSlashing, Limit MAX_ATTESTER_SLASHINGS]
+    attestations*: List[TrustedAttestation, Limit MAX_ATTESTATIONS]
+    deposits*: List[Deposit, Limit MAX_DEPOSITS]
+    voluntary_exits*: List[TrustedSignedVoluntaryExit, Limit MAX_VOLUNTARY_EXITS]
+    sync_aggregate*: TrustedSyncAggregate
+
+    # Execution
+    execution_payload* {.dontSerialize.}: ExecutionPayload  # [New in Bellatrix]
```
etan-status, line 369: Would there be a benefit in only splitting away the transactions, but to keep the `ExecutionPayloadHeader`?

e.g., if in the end, the EL will store the payloads, keeping the header could potentially allow re-syncing the EL if user decides to change to a different one, using `forkChoiceUpdated` calls. Or is the copy of the header in the `BeaconState` sufficient for that?

beacon_chain/beacon_chain_db.nim
```diff
@@ -818,12 +847,81 @@ proc getBlock*(
...
+  # optimize away some of these encoding/decoding and compression/decompression
+  # rounds which unfortunately reappear.
+  #
+  # Furthermore, even gimmicks such as splitting the beacon block into before,
+  # then `ExecutionPayload`, then the trailing signature, then combining using
+  # concatenated compressed frames, would only help when `ExecutionPayload` is
+  # stored locally whereas one underlying reason to do this at all is allowing
+  # execution clients to handle `ExecutionPayload` storage sans duplication.
+  if GetResult.found != db.partialBlocks[T.toFork].getSZSSZ(
+      key.data, isomorphicCast[TrustedSignedPartialBeaconBlock](data)):
```
etan-status: In `forks.nim` there are helpers like `asTrusted()` to add type safety around `isomorphicCast`.

beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -342,6 +342,39 @@ type
...
+    proposer_index*: uint64 # `ValidatorIndex` after validation
+    parent_root*: Eth2Digest
+    state_root*: Eth2Digest
+    body*: TrustedPartialBeaconBlockBody
+
+  TrustedSignedPartialBeaconBlock* = object
+    message*: TrustedPartialBeaconBlock
+    signature*: TrustedSig
+
+    root*: Eth2Digest # serialized signal of whether to fetch execution payload
```
etan-status, line 382: Hmm, this doesn't seem to be used as a signal in the end (and in the DB, the root is already known from the table key)

beacon_chain/beacon_chain_db.nim
```diff
@@ -818,12 +847,81 @@ proc getBlock*(
...
-    db: BeaconChainDB, key: Eth2Digest,
-    T: type X): Opt[T] =
+    db: BeaconChainDB, key: Eth2Digest, T: type X): Opt[T] =
   # We only store blocks that we trust in the database
   result.ok(default(T))
+
+  if db.getSplitBlock(key, result.get, T):
+    return result
+
   if db.blocks[T.toFork].getSZSSZ(key.data, result.get) == GetResult.found:
```
etan-status, line 938: While it's nice to fall back to the table with full blocks, it is only useful for testnets. The alternative would be to migrate existing data to the new structure on startup, or to have testnets re-sync.

beacon_chain/beacon_chain_db.nim
tersec: > Or is the idea to first store them in _executionpayloads and then syncing them to the EL client later, then dropping them from our table?

Yes

tersec: Well, not even syncing to, exactly. That's where the canonical source of truth for them is anyway.

beacon_chain/spec/datatypes/bellatrix.nim
tersec: Not here, no, because it's always local.

beacon_chain/beacon_chain_db.nim
tersec: Yes, we can get this out before `mainnet`, and I'm fine with having no backwards compatibility. It's slightly nicer, just a question of whether one's willing to break existing Goerli/Prater setups until they CL-resync. The other testnets are either deprecated or otherwise only used by people more used to that, who resync frequently anyway.

beacon_chain/beacon_chain_db.nim
tersec: I don't have a strong preference here. It also has some CRCs, for example, which are nice. When in doubt, the framed format should probably be default.

beacon_chain/spec/datatypes/bellatrix.nim
tersec: Potentially interesting, yeah. Worth considering. Small additional amount of disk space for some maybe-useful information always being in the CL database.

beacon_chain/beacon_chain_db.nim
tersec: This would allow dropping the `partial_blocks` tables as separate from just `blocks` because each fork would only ever use exactly one. That's already the case pre-Bellatrix and post-Bellatrix (Capella won't have compatibility for this; it will be born split-block). So it's really just Bellatrix.

beacon_chain/spec/datatypes/bellatrix.nim
tersec: The issue with just using pre-existing information is that Bellatrix/Capella/etc don't imply merge. They could have no corresponding nonzero `ExecutionPayload`. This means that what will become a more expensive RPC operation, getting this information from the EL, could be avoided entirely.

tersec:
https://github.com/nim-lang/Nim/issues/20141 is what hit the Linux CIs. They're using Ubuntu 18.04, and thus gcc 7.x, while the VM in which I recreated/tested this was Ubuntu 20.04 and gcc 9.x, so presumably it spans at least those versions of gcc.

tersec:
> [nim-lang/Nim#20141](https://github.com/nim-lang/Nim/issues/20141) is what hit the Linux CIs. They're using Ubuntu 18.04, and thus gcc 7.x, while the VM in which I recreated/tested this was Ubuntu 20.04 and gcc 9.x, so presumably it spans at least those versions of gcc.

https://github.com/status-im/nimbus-eth2/pull/3932/commits/248dae025e6434ed0bf93282e2a0a4ae3e872f4f still causes the crucial line to be compiled without any additional copies:
```c
T3_ = (tyEnum_GetResult__KebNXtrEUaKiSLaf7LLUJA)0;              T3_ = getSZSSZ__beacon95chain95db_38039((*db).partialBlocks[(((tyEnum_BeaconBlockFork__CJmpmLh9awLxuRuGu9ayPu9aQ) 2))- 0], (*key_0).data, 32, (&(*((tyObject_TrustedSignedPartialBeaconBlock__nZPmS9bKi6dcEhRUFGxsqgg*) (data_0)))));           if (!!((((tyEnum_GetResult__KebNXtrEUaKiSLaf7LLUJA) 0) == T3_))) goto LA4_;
```

beacon_chain/beacon_chain_db.nim
arnetheduck: it also supports streaming (for lower mem usage), so it's probably more future proof once blocks start growing..

beacon_chain/beacon_chain_db.nim
arnetheduck: one thing is that we don't _really_ need a table-per-fork - I've been thinking about getting rid of that, and this might be a good opportunity - ie we know the fork by slot when loading, or can use the load-header trick - there are already other places where it's done that way (era files, genesis states etc etc)

beacon_chain/beacon_chain_db.nim
arnetheduck: I would keep the backwards compat for now, and potentially drop it in a future release - we don't want to confuse prater users now, when they're busy with their testing

arnetheduck: this also helps a potential migration feature: if version N-1 can read split table and N performs migration to split table, it's still possible to downgrade to N-1.

tersec:
Probably not worthwhile once enforcing `MIN_EPOCHS_FOR_BLOCK_REQUESTS` of around 5 months.

Issue #3624, opened 2022-05-09
Checkpoint sync doesn't accept `charset` in `content-type` response header


acolytec3:
**Describe the bug**
The Nimbus client requires that response headers be set to strictly `application/json` without any `charset` specified when trying to parse REST responses during checkpoint sync so is not able to parse responses from the Lodestar client on the Kiln testnet.

**To Reproduce**
1. OS - Ubuntu 22 running on `AMD64` platform
2. Branch/commit used: `kiln-auth-dev`
3. Commands being executed: `build/nimbus_beacon_node trustedNodeSync --trusted-node-url=https://lodestar-kiln.chainsafe.io --network=vendor/merge-testnets/kiln --web3-url=ws://127.0.0.1:8551 --log-level=DEBUG  --jwt-secret="/home/jim/development/ethjs/packages/client/kiln/datadir/jwtsecret" --data-dir=build/kiln`
4. Relevant log lines: 
```
NOT 2022-05-09 15:56:40.867-04:00 Starting trusted node sync                 databaseDir=build/kiln/db restUrl=https://lodestar-kiln.chainsafe.io blockId=finalized backfill=true reindex=false
DBG 2022-05-09 15:56:41.146-04:00 Sending REST request to remote server      remote=lodestar-kiln.chainsafe.io:443 request=/eth/v1/beacon/blocks/genesis/root http_method=GET
DBG 2022-05-09 15:56:41.399-04:00 Got REST response headers from remote server status=200 http_method=GET remote=lodestar-kiln.chainsafe.io:443 request=/eth/v1/beacon/blocks/genesis/root connection=1
DBG 2022-05-09 15:56:41.399-04:00 Received REST response body from remote server status=200 http_method=GET remote=lodestar-kiln.chainsafe.io:443 request=/eth/v1/beacon/blocks/genesis/root connection=1 contentType="application/json; charset=utf-8" size=77
ERR 2022-05-09 15:56:41.399-04:00 Unable to download genesis block root      error="Unable to decode REST response, error [Content-Type not supported]" restUrl=https://lodestar-kiln.chainsafe.io
```

zah:
This issue was addressed by the linked PR above and the fixes have been shipped in Nimbus v22.8.0

PR #3934, opened 2022-08-03
Address #3624

cheatfate:
This PR requires:

1. https://github.com/status-im/nim-chronos/pull/302
2. https://github.com/status-im/nim-presto/pull/32

to be merged.

### DIFF
--- a/beacon_chain/rpc/rest_nimbus_api.nim
+++ b/beacon_chain/rpc/rest_nimbus_api.nim
@@ -169,7 +169,8 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
     if contentBody.isNone:
       return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
 
-    template setGraffitiAux(node: BeaconNode, graffitiStr: string): RestApiResponse =
+    template setGraffitiAux(node: BeaconNode,
+                            graffitiStr: string): RestApiResponse =
       node.graffitiBytes = try:
         GraffitiBytes.init(graffitiStr)
       except CatchableError as err:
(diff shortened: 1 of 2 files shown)


beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -177,17 +177,17 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
       node.graffitiBytes = graffitiBytes.get
       return RestApiResponse.jsonResponse((result: true))
-    of "text/plain":
+    elif contentType == TextPlainMediaType:
       return node.setGraffitiAux contentBody.get.strData
-    of "application/x-www-form-urlencoded":
+    elif contentType == UrlEncodedMediaType:
       return node.setGraffitiAux decodeUrl(contentBody.get.strData)
     else:
       return RestApiResponse.jsonError(Http400, "Unsupported content type: " &
```
etan-status, line 194: The `$contentBody.get.contentType` on next line --> `contentType`

etan-status (approved):
Have not looked into the `nim-http-utils` changes thoroughly, but the submodule is updated to `master` so I guess those are already reviewed. 

I did some checks re case-sensitive handling, and all seems correct ✅

RFC 2045
>    The type, subtype, and parameter names are not case sensitive.  For
   example, TEXT, Text, and TeXt are all equivalent top-level media
   types.  Parameter values are normally case sensitive, but sometimes
   are interpreted in a case-insensitive fashion, depending on the
   intended use.  (For example, multipart boundaries are case-sensitive,
   but the "access-type" parameter for message/External-body is not
   case-sensitive.)

The new Chronos tests seem to cover the linked issue, so LGTM 👍

beacon_chain/rpc/rest_nimbus_api.nim
cheatfate: Could you please be more specific, because i'm missing the point.

etan-status: On the next line (194) there is still a mention to `$contentBody.get.contentType`, which could also be replaced with the new `contentType` variable (instead of re-computing that once more). GH did not allow me to comment on that line.

PR #3937, opened 2022-08-03
log NimVersion

narimiran:
After this PR:

```
NOT 2022-08-03 19:22:16.939+02:00 Starting beacon node                       topics="beacnde" version=v22.7.0-52b32c-stateofus nimVersion=1.2.16 enr=(...)
```

IMO, the commit hash for Nim is not needed if we stick to the released stable versions of the compiler. (e.g. 1.2.16 or 1.6.8)

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -145,6 +145,9 @@ declareGauge next_action_wait,
 declareGauge versionGauge, "Nimbus version info (as metric labels)", ["version", "commit"], name = "version"
 versionGauge.set(1, labelValues=[fullVersionStr, gitRevision])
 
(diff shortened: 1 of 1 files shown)


cheatfate:
I do not like idea to put `nimVersion` in `Starting beacon node` log line.
Also i do not like idea to keep only `nimVersion` because even right now specific nim version commit hash do not equal to nim version commit hash which is distributed and maintained by `nimbus-build-system`.

mratsim:
> the commit hash for Nim is not needed if we stick to the released stable versions of the compiler. (e.g. 1.2.16 or 1.6.8)

We might still need Nim specific commits in the future, especially as people try early for compat with 1.7/1.8 or before Nim 2.0

PR #3938, opened 2022-08-04
The body of prepare_beacon_proposer is an array (Fixes #3920)

### LINKED ISSUE #3920: `prepare_beacon_proposer` not spec-compliant
**Describe the bug**

The `prepare_beacon_proposer` REST API appears to take only a single proposer object, whereas the spec states that it should take an array of objects.

**To Reproduce**
```
$ curl -w %{http_code} -H 'content-type:application/json' --data '{"validator_index":"1","fee_recipient":"0x5a454afa7e7d9b4d914067715c3b16cf3e3c899d"}' http://localhost:5052/eth/v1/validator/prepare_beacon_proposer
200
$ curl -s -H 'content-type:application/json' --data '[{"validator_index":"1","fee_recipient":"0x5a454afa7e7d9b4d914067715c3b16cf3e3c899d"}]' http://localhost:5052/eth/v1/validator/prepare_beacon_proposer 
{"code":"400","message":"Unable to decode prepare beacon proposer request","stacktraces":[]}
```

**Additional context**

```
$ nimbus_beacon_node --version
Nimbus beacon node v22.7.0-75ccfc-stateofus
Copyright (c) 2019-2022 Status Research & Development GmbH

eth2 specification v1.2.0-rc.1

Nim Compiler Version 1.2.16 [Linux: amd64]
```

zah:


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -764,21 +764,22 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-      proposerData.validator_index,
-      proposerData.fee_recipient,
-      currentEpoch)
+    for proposerData in body:
+      node.dynamicFeeRecipientsStore.addMapping(
+        proposerData.validator_index,
+        proposerData.fee_recipient,
+        currentEpoch)
 
     return RestApiResponse.response("", Http200, "text/plain")
```
cheatfate, line 784: Are you sure that `text/plain` should be specified?

etan-status: https://github.com/ethereum/beacon-APIs/blob/master/apis/validator/prepare_beacon_proposer.yaml
(lists `application/json`)

zah: `application/json` is specified only regarding the *request* contents.

Since the response is empty, I don't think `application/json` would be appropriate. This may trigger a JSON parsing attempt of an empty string which should result in an error in most clients. I think we can raise an issue with the spec for being under-specified in this particular case.

zah: Some people argue that the HTTP spec implies that `Content-Type` should not be specified when the `Content-Length` is zero:

https://stackoverflow.com/questions/29784398/should-content-type-header-be-present-when-the-message-body-is-empty

PR #3943, opened 2022-08-08
fix obtaining deposits after connection loss

etan-status:
When an error occurs during Eth1 deposits import, the already imported
blocks are kept while the connection to the EL is re-established.
However, the corresponding merkleizer is not persisted, leading to any
future deposits no longer being properly imported. This is quite common
when syncing a fresh Nimbus instance against an already-synced Geth EL.
Fixed by persisting the head merkleizer together with the blocks.

### DIFF
--- a/beacon_chain/beacon_chain_db_light_client.nim
+++ b/beacon_chain/beacon_chain_db_light_client.nim
@@ -20,8 +20,7 @@ import
   eth/db/kvstore_sqlite3,
   # Beacon chain internals
   spec/datatypes/altair,
-  spec/[eth2_ssz_serialization, helpers],
-  ./filepath
+  spec/[eth2_ssz_serialization, helpers]
 
 logScope: topics = "lcdata"
 
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -651,7 +651,7 @@ when hasDepositRootChecks:
   const
     contractCallTimeout = 60.seconds
 
-  func fetchDepositContractData(p: Web3DataProviderRef, blk: Eth1Block):
+  proc fetchDepositContractData(p: Web3DataProviderRef, blk: Eth1Block):
                                 Future[DepositContractDataStatus] {.async.} =
     let
       depositRoot = p.ns.get_deposit_root.call(blockNumber = blk.number)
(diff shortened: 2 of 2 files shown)


zah:
The head merkleizer is not kept around on purpose, because the web3 providers are known to miss certain deposit events from time to time which causes the merkleizer to get infected with invalid values. These are eventually discovered, the monitor is restarted and we continue from the known safe finalized state.

etan-status:
Yes, that case still works. But in the case where there are no invalid values, there are still regular disconnects of the websocket connection every couple minutes, and because there is nothing invalid in there, the sync starts from the previous head (instead of from finalized head). In this case, the old head merkleizer still needs to be available.

Memory consumption should be similar, whether it is living in a `newClone` of a persistent background task, or whether it is retained across followup invocations of the same task as part of `Eth1Chain`.

etan-status:
The scenario you are describing goes through this flow:

```nim
  if m.latestEth1Block.isSome and m.depositsChain.blocks.len > 0:
    let needsReset = m.depositsChain.hasConsensusViolation or (block:
      let
        lastKnownBlock = m.depositsChain.blocks.peekLast
        matchingBlockAtNewProvider = awaitWithRetries(
          m.dataProvider.getBlockByNumber lastKnownBlock.number)

      lastKnownBlock.voteData.block_hash.asBlockHash != matchingBlockAtNewProvider.hash)

    if needsReset:
      m.depositsChain.clear()
      m.latestEth1Block = none(FullBlockId)

```

and then, because `m.depositsChain.blocks.len == 0` due to the reset:

```nim
  if shouldProcessDeposits and m.depositsChain.blocks.len == 0:
    let startBlock = awaitWithRetries(
      m.dataProvider.getBlockByHash(
        m.depositsChain.finalizedBlockHash.asBlockHash))

    m.depositsChain.addBlock Eth1Block(
      number: Eth1BlockNumber startBlock.number,
      timestamp: Eth1BlockTimestamp startBlock.timestamp,
      voteData: eth1DataFromMerkleizer(
        m.depositsChain.finalizedBlockHash,
        m.depositsChain.finalizedDepositsMerkleizer))

    eth1SyncedTo = Eth1BlockNumber startBlock.number

    eth1_synced_head.set eth1SyncedTo.toGaugeValue
    eth1_finalized_head.set eth1SyncedTo.toGaugeValue
    eth1_finalized_deposits.set(
      m.depositsChain.finalizedDepositsMerkleizer.getChunkCount.toGaugeValue)

    m.depositsChain.headMerkleizer = copy m.finalizedDepositsMerkleizer

    debug "Starting Eth1 syncing", `from` = shortLog(m.depositsChain.blocks[0])
```

The fix is for the case where `needsReset == false`, where `blocks.len > 0`. The merkleizer was lost and sync got stuck.

zah:
Alright, thank you for the explanation.

It's also curious why your WebSocket connection was dying so frequently. Have you tried to look into the root cause behind this?

PR #3945, opened 2022-08-09
Minor post-merge cleanups

zah:
https://github.com/status-im/nimbus-eth2/pull/3944

The use of nested `awaitWithRetries` calls would have
resulted in an unexpected number of retries (3x3).
We now use regular `await` in outer layer to avoid the problem.

https://github.com/status-im/nimbus-eth2/pull/3943

The new code has an invariant that the `headMerkleizer` field in
the `Eth1Chain` is always kept in sync with the blocks stored in
the chain.

This invariant is now enforced better by doing the necessary merkleizer updates
in the `Eth1Chain.addBlock` function, in the `Eth1Chain.init` function and in the
`Eth1Chain.reset` function.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -70,10 +70,19 @@ type
   GenesisStateRef = ref phase0.BeaconState
 
   Eth1Block* = ref object
+    hash*: Eth2Digest
     number*: Eth1BlockNumber
     timestamp*: Eth1BlockTimestamp
+      ## Basic properties of the block
+      ## These must be initialized in the constructor
+
     deposits*: seq[DepositData]
-    voteData*: Eth1Data
+      ## Deposits inside this particular block
+
+    depositRoot*: Eth2Digest
+    depositCount*: uint64
+      ## Global deposits count and hash tree root of the entire sequence
+      ## These are computed when the block is added to the chain (see `addBlock`)
 
     when hasGenesisDetection:
       activeValidatorsCount*: uint64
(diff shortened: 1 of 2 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -628,11 +660,6 @@ func depositEventsToBlocks(depositsList: JsonNode): seq[Eth1Block] {.
       amount: bytes_to_uint64(amount.toArray),
       signature: ValidatorSig.init(signature.toArray))
 
-proc fetchTimestamp(p: Web3DataProviderRef, blk: Eth1Block) {.async.} =
-  let web3block = awaitWithRetries(
-    p.getBlockByHash(blk.voteData.block_hash.asBlockHash))
-  blk.timestamp = Eth1BlockTimestamp web3block.timestamp
-
```
etan-status, line 635: Does it still compile with `-d:hasDepositRootChecks`?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -781,36 +798,38 @@ func lowerBound(chain: Eth1Chain, depositCount: uint64): Eth1Block =
...
       return
     result = eth1Block
 
 proc trackFinalizedState(chain: var Eth1Chain,
                          finalizedEth1Data: Eth1Data,
                          finalizedStateDepositIndex: uint64,
                          blockProposalExpected = false): bool =
-  # Returns true if the Eth1Monitor is synced to the finalization point
+  ## This function will return true if the Eth1Monitor is synced
+  ## to the finalization point.
```
etan-status, line 810: Wasn't there a convention that function comments should use the imperative form (`Return true if the Eth1Monitor is synced to the finalization point`) instead of descriptive form (the current version) or the most common third person form (`ReturnS true if ...`)? I can't find it in the style-guide anymore but I remember seeing something suggesting imperative form after running into this style in a couple modules. Maybe it's also an Ethereum style rule (they use imperative as well).

Also, `Indicate whether` vs `Return true if`, to better describe the `false` case

research/block_sim.nim
```diff
@@ -101,10 +101,7 @@ cli do(slots = SLOTS_PER_EPOCH * 6,
 
   eth1Chain.addBlock Eth1Block(
     number: Eth1BlockNumber 1,
-    timestamp: Eth1BlockTimestamp genesisTime,
-    voteData: Eth1Data(
-      deposit_root: merkleizer.getDepositsRoot,
-      deposit_count: merkleizer.getChunkCount))
```
etan-status, line 107: These two intentionally removed? In rest of the code, those two fields were simply moved out of `voteData`.

etan-status (approved):

research/block_sim.nim
zah: Yes, because they are now consistently computed when the block is added to the chain (in `addBlock`).

beacon_chain/eth1/eth1_monitor.nim
zah: I've changed the wording because yesterday while reading the code I was confused for a moment that the comment was referring to the `if` statement below which didn't make sense.

beacon_chain/eth1/eth1_monitor.nim
zah: Yes

PR #3947, opened 2022-08-10
avoid materializing potentially long deposits seq

etan-status:
When fetching eth1 data and deposits for a new block proposal, the list
of deposits from previous eth1 data to the next one is fully loaded into
a `seq`. This can potentially be a very long list in active periods.
Changing this to an `iterator` saves memory by ensuring that the entire
list is no longer materialized; only the `DepositData` roots are needed.


arnetheduck:
Nice, more eth1monitor tests are long overdue!

PR #3955, opened 2022-08-11
book edits

arnetheduck:
* reorganise navigation menus
* update light client guide with comparison table
* add suggested fee recipient and JWT secrets to the merge guide
* add some background info to book readme


docs/the_nimbus_book/src/migration.md
```diff
@@ -1,38 +1,36 @@
...
-
 # Migrate from another client
 
 This guide will take you through the basics of how to migrate to Nimbus from another client. *See [here](./migration-options.md) for advanced options*.
 
-The main pain point involves the exporting and importing of the [slashing protection database](https://eips.ethereum.org/EIPS/eip-3076), since each client takes a slightly different approach here.
+**Please take your time to get this right.** Don't hesitate to reach out to us in the `#helpdesk` channel of [our discord](https://discord.gg/j3nYBUeEad) if you come across a stumbling block. We are more than happy to help guide you through the migration process. Given what's at stake, there is no such thing as a stupid question.
+
+!!! info ""
+    Unlike other clients, Nimbus does not require a separate validator client - instead, validators are by default run inside the beacon node process.
```
zah:
```suggestion
    Unlike other clients, Nimbus does not require a separate validator client - instead, the validators duties can be handled by the beacon node process.
```

zah: The previous wording seemed a bit unusual to me.

docs/the_nimbus_book/src/suggested-fee-recipient.md
```diff
@@ -1,10 +1,25 @@
-# Suggested Fee Recipient
+# Set up suggested fee recipient
 
-A suggested fee recipient offers an execution client, in a merged Ethereum network, a coinbase it might use. The execution client might not use this offered coinbase, unless one ensures that by running, controlling, and similarly configuring this execution client oneself.
+A suggested fee recipient offers an execution client, in a merged Ethereum network, a coinbase it might use, base on the validator that is proposing the block.
```
zah:
```suggestion
A suggested fee recipient offers an execution client, in a merged Ethereum network, a coinbase it might use, based on the validator that is proposing the block.
```

docs/the_nimbus_book/src/migration.md
arnetheduck:
```suggestion
    Unlike other clients, Nimbus does not require a separate validator client - instead, validators run in the beacon node process.
```

arnetheduck: this wording lacks a `'` possessive - but I think we say that validators "run" elsewhere as well - it's a bit ambiguous whether "validator" refers to a single key, the role of being a validator in general with one or more keys or something else. Checking the other clients, they also talk about running validators interchangeable with "performing validator duties" and other forms - no clear consensus that I can find - anyway, you're right that it  read wrong, fixed it another way

PR #3956, opened 2022-08-11
import EL deposits even when EL is stuck

etan-status:
The `eth1_monitor` only starts importing deposits once the EL reports a
new head block. However, the EL may be stuck at a block, e.g., the TTD.
By polling the latest EL block once after subscribing to new EL block
events it is ensured that deposits are still imported in this situation.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -1314,6 +1314,9 @@ proc syncBlockRange(m: Eth1Monitor,
 func init(T: type FullBlockId, blk: Eth1BlockHeader|BlockObject): T =
   FullBlockId(number: Eth1BlockNumber blk.number, hash: blk.hash)
 
+func isNewLastBlock(m: Eth1Monitor, blk: Eth1BlockHeader|BlockObject): bool =
+  blk.number.uint64 > m.latestEth1BlockNumber
+
 proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
   if m.state == Started:
     return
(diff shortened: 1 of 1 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1446,31 +1447,29 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
...
 
-    let nextBlock = if mustUsePolling:
+    let nextBlock = if mustUsePolling or not didPollOnce:
       let blk = awaitWithRetries(
         m.dataProvider.web3.provider.eth_getBlockByNumber(blockId("latest"), false))
 
-      let fullBlockId = FullBlockId.init(blk)
-
-      if m.latestEth1Block.isSome and
-         m.latestEth1Block.get == fullBlockId:
```
zah, line 1467: I think this condition must be preserved. The execution engine can pivot to another block at the same height or even to another block at a lower height. We we need to sleep and continue (which means retrying the request) only if we have obtained exactly the same block as before.

etan-status: The new logic simply replicates the existing logic in `newBlockHeadersHandler`, having separate logic for events-based vs polling-based API does not seem right.

The `eth1_monitor` subtracts 1024 blocks (follow distance) from this latest EL block to determine its syncing target, so smaller reorgs around the chain tip are not a problem.

The `>` logic seems more correct, because we don't react to reorgs going backward and simply keep the blocks that we already have imported.

etan-status: Also, if there's multiple ELs connected with different syncing progress, we should probably only follow the one with higher sync progress, which is guaranteed by the `>` logic.

etan-status: More accurate would probably be a check based on timestamp, but I'm not sure if there's a reason why block number is used instead of timestamp to determine newer block.

zah: These are fair points. I guess we can name this operation to have it consistent in both places - e.g. `m.isNewLatestBlock(blk)`

PR #3960, opened 2022-08-12
track request chunk size across EL reconnects

etan-status:
When the EL connection is interrupted, deposits are once more requested
in chunks of 5000 blocks. This is a problem when the response takes over
a minute to produce and consistently times out as followup requests with
lower chunk sizes may no longer work after a request was canceled, e.g.,
when using Geth with websockets. By keeping track of `blocksPerRequest`
across EL reconnections, it is possible to recover from this by avoiding
to continuously repeat the initial request with the full 5000 blocks.
Also cleans up one more "retry of retry" instance; `DataProviderTimeout`
is a `CatchableError` and already handled by the existing retry logic.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -62,6 +62,8 @@ const
   hasDepositRootChecks = defined(has_deposit_root_checks)
   hasGenesisDetection* = defined(has_genesis_detection)
 
+  targetBlocksPerLogsRequest = 5000'u64  # This is roughly a day of Eth1 blocks
+
 type
   Eth1BlockNumber* = uint64
   Eth1BlockTimestamp* = uint64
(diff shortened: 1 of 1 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1177,15 +1175,22 @@ proc syncBlockRange(m: Eth1Monitor,
           toBlock = some blockId(maxBlockNumberRequested))
 
         depositLogs = try:
-          # Downloading large amounts of deposits can be quite slow
+          # Downloading large amounts of deposits may take several minutes
           awaitWithTimeout(jsonLogsFut, web3Timeouts):
-            retryOrRaise newException(DataProviderTimeout,
```
zah, line 1182: The intention here was to avoid raising the exception if that's not necessary. You are right that the changed code is logically equivalent, but it does go through all the trouble of raising and catching an exception in an async function.

zah (approved):

beacon_chain/eth1/eth1_monitor.nim
etan-status: The outer `except` block has yet another retry loop, leading to quadratic retries.
In practice, with Geth websocket, I have never seen the connection to recover from a timeout.

PR #3967, opened 2022-08-15
Keystore deserialization should be able to handle optional description.

cheatfate:


zah:
Fix implemented in https://github.com/status-im/nimbus-eth2/pull/3976

PR #3974, opened 2022-08-16
packaging updates

arnetheduck:
* one package per binary (nimbus_beacon_node, nimbus_validator_client)
* use `-` in package name (`_` is separating the version)
* don't include (un)installation scripts in package
* default metrics port 8009 for vc
* fix several upgrade/install errors in scripts
* add JWT option to service files


scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -27,8 +29,10 @@ Environment=NETWORK=mainnet
 
 # You need to have have access to an execution client - by default, we assume
 # a compatible execution client is running on the same machine on the default
-# websocket port.
+# websocket port. Make sure to supply the right path to the JWT secret.
+
 Environment=WEB3_URL=ws://127.0.0.1:8546
+Environment=JWT_SECRET=/tmp/jwtsecret
```
zah, line 35: The `/tmp` folder is a system-wide namespace and `jwtsecret` is too generic as a name. We should use a prefix like `nimbus` in the filename.

scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service
```diff
@@ -51,7 +55,8 @@ TimeoutSec=1200
 Restart=always
 
 # Don't restart when Doppelganger detection has been activated
-RestartPreventExitStatus=1031
+# TODO https://github.com/status-im/nimbus-eth2/issues/3973
+# RestartPreventExitStatus=1031
```
zah: I've already submitted a PR changing this here:
https://github.com/status-im/nimbus-eth2/pull/3977

scripts/package_src/nimbus_validator_client/after_install
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+DISTRO=$(lsb_release -si)
+if ! id -u nimbus > /dev/null 2>&1; then
+  case $DISTRO in
+    Ubuntu|Debian)
+      adduser --system --no-create-home --group nimbus
```
zah, line 10: `adduser` will fail if the user `nimbus` already exists, which will result in a failing installation due to the `set -e` directive above.

Here are some ideas how to make the command idempotent:
https://unix.stackexchange.com/questions/28526/add-a-user-to-the-system-only-if-it-doesnt-exist

scripts/package_src/nimbus_validator_client/deb_after_purge
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+# TODO We purge the nimbus user / group / folder from the beacon node uninstall
```
zah: Right, this looks like an open question.

scripts/package_src/nimbus_validator_client/image/lib/systemd/system/nimbus_validator_client.service
```diff
@@ -0,0 +1,55 @@
...
+Environment=METRICS_PORT=8009
+
+# Interaction and monitoring
+Environment=METRICS_ENABLED=Yes
+
+# Default group = nimbus
+User=nimbus
+WorkingDirectory=/var/lib/nimbus
+TimeoutSec=1200
+Restart=always
```
zah, line 46: The validator client support the doppelganger exit code as well.

beacon_chain/conf.nim
```diff
@@ -841,7 +841,7 @@ type
 
     metricsPort* {.
       desc: "Listening HTTP port of the metrics server"
-      defaultValue: 8008
+      defaultValue: 8009
```
zah: How about a port like `8108` to make it easier to use our current practice of creating scripts that use the `BASE_PORT + NODE_ID` pattern?

scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service
arnetheduck: the jwtsecret is owned by the execution client (it's the EC that creates it), and both nethermind, geth and several other guides use `/tmp/jwtsecret` in their [documentation](https://geth.ethereum.org/docs/interface/consensus-clients) - see the example command line.

scripts/package_src/nimbus_validator_client/after_install
arnetheduck: hm, so according to the debian manpage, `adduser` returns 0 if the user already exists: https://manpages.debian.org/jessie/adduser/adduser.8.en.html

this is contrary to `useradd`, which indeed gives an error :/ it also ties in with the TODO as to which package should be responsible for the user - if all packages create-if-don't-exist `nimbus`, none should remove it until all are removed - it's a messy problem.

finally, `useradd` actually exists on debian - `adduser` is just a wrapper - we could simply use `useradd` always, and then with the `id` trick..

arnetheduck: btw, the whole section is already gated by an `if ! id` check above, meaning it already checks if the user exists

scripts/package_src/nimbus_validator_client/deb_after_purge
arnetheduck: actually, turns out it it isn't - most guides recommend not to remove created users

scripts/package_src/nimbus_validator_client/image/lib/systemd/system/nimbus_validator_client.service
arnetheduck: rebased

beacon_chain/conf.nim
arnetheduck: done

scripts/package_src/nimbus_validator_client/after_install
zah: Ah, I've missed that, sorry.

scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service
zah: That's still quite ridiculous, but, oh, well...

scripts/package_src/nimbus_beacon_node/before_remove
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -e
+
+SERVICE_NAME="nimbus_beacon_node"
+
+if systemctl --all --type service | grep -q "$SERVICE_NAME";then
```
zah:
```suggestion
if systemctl --all --type service | grep -q "$SERVICE_NAME"; then
```

PR #3980, opened 2022-08-17
Bellatrix metadata for Mainnet

zah:


etan-status:
`beacon_chain/spec/presets.nim` needs a bump as well.

zah:
I would say it needs refactoring - using any constant from there which should have been read from the run-time config is a potential bug. But are you talking about files such as `bellatrix_preset.nim` or to the definition of the `RuntimeConfig` (e.g. missing fields, etc).

zah:
Nevermind, found out what you meant!

zah:
In `presets.nim`, there is a constant `defaultRuntimeConfig` which holds values that will be used as defaults in case the network specific run-time config doesn't override them. The specs currently don't specify what should be the treatment of configs with missing values, but the Gnosis network is known to use such configs with the expectation that "all other settings are the same as mainnet". This seems relatively reasonable for most settings, but it's a bit less clear about values such as `TERMINAL_TOTAL_DIFFICULTY` and `X_FORK_EPOCH`. A more appropriate conservative stance would be to assign these parameters to known safe values such as "far future epoch" until the network config hasn't specified them explicitly.

tersec (approved):

Issue #3963, opened 2022-08-13
Sync progress should distinguish between optmistic and real progress


arnetheduck:
```
INF 2022-08-12 16:40:24.163+02:00 Slot start                                 topics="beacnde" slot=3650602 epoch=114081 sync="00h00m (99.99%) 6.5529slots/s (RPQQQQQQQQ:3650367)" peers=35 head=85f92e16:3650389 finalized=114072:cf9d2bd5 delay=163ms932us33ns
```

Here, we see `sync=...` showing progress making it appear that we will be synced soon, but what's really happening is that the blocks are optimistic. Instead, if optimistic sync is happening, it should be spelled out somehow (`sync="opt 00h:00m..."` for example)

arnetheduck:
ideally we would also show eth1 progress here, when opt-synced (so as to give an estimate when duties will resume)

tersec:
> ideally we would also show eth1 progress here, when opt-synced (so as to give an estimate when duties will resume)

Perhaps, but that's also a distinct issue, and the issue as opened has been addressed. There are thornier issues with that, in general. It's reasonable to continue to discuss those in this issue or to open another issue for that, but either way, closing.

PR #3987, opened 2022-08-18
distinguish between optimistic and non-optimistic sync progress

tersec:
https://github.com/status-im/nimbus-eth2/issues/3963

I'm not certain this is the best formatting, but it's arguably better to put the `opt` after rather than before the time stamp, since there's a race-condition like setup where exactly when the execution client verifies a payload can vary, and this makes the churn more readable in that case.

There's a case for doing this for "opt synced" too, but there, at least, there isn't any additional changing information to read (the estimated time).

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1343,13 +1343,13 @@ func syncStatus(node: BeaconNode): string =
   let optimistic_head = node.dag.is_optimistic(node.dag.head.root)
   if node.syncManager.inProgress:
     if optimistic_head:
-      node.syncManager.syncStatus & ": opt"
+      node.syncManager.syncStatus & "/opt"
     else:
       node.syncManager.syncStatus
   elif node.backfiller.inProgress:
     "backfill: " & node.backfiller.syncStatus
   elif optimistic_head:
-    "opt synced"
+    "synced/opt"
   else:
     "synced"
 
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1340,11 +1340,15 @@ proc onSlotEnd(node: BeaconNode, slot: Slot) {.async.} =
...
   if node.syncManager.inProgress:
-    node.syncManager.syncStatus
+    if optimistic_head:
+      node.syncManager.syncStatus & ": opt"
+    else:
+      node.syncManager.syncStatus
   elif node.backfiller.inProgress:
     "backfill: " & node.backfiller.syncStatus
-  elif node.dag.is_optimistic(node.dag.head.root):
+  elif optimistic_head:
```
arnetheduck, line 1351: I think it makes sense to keep the same formatting for both syncing and synced (so one can quickly grep for `: opt` for example

arnetheduck: maybe `/opt` instead which allows grepping without "" :)

arnetheduck (approved):

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/3987/commits/e991be6b2c08170f24111bc1980e06fbca0e42cc

PR #3991, opened 2022-08-18
Fix the block header computation when proposing an empty execution payload

zah:


beacon_chain/spec/helpers.nim
```diff
@@ -352,17 +376,14 @@ func build_empty_execution_payload*(state: bellatrix.BeaconState): ExecutionPayl
...
     prev_randao: randao_mix,
     gas_limit: latest.gas_limit, # retain same limit
     gas_used: 0, # empty block, 0 gas
     timestamp: timestamp,
     base_fee_per_gas: latest.base_fee_per_gas) # retain same base_fee
 
-  payload.block_hash = withEth2Hash:
-    h.update payload.hash_tree_root().data
-    h.update cast[array[13, uint8]]("FAKE RLP HASH")
+  payload.block_hash = payload.toBlockHeader.blockHash
```
etan-status: Interesting. Are these still correct?
```
func asEth2Digest*(x: BlockHash): Eth2Digest =
  Eth2Digest(data: array[32, byte](x))

template asBlockHash*(x: Eth2Digest): BlockHash =
  BlockHash(x.data)
```

And, does `asEngineExecutionPayload` and `asConsensusExecutionPayload` need updating as well?

beacon_chain/spec/helpers.nim
```diff
@@ -352,17 +376,14 @@ func build_empty_execution_payload*(state: bellatrix.BeaconState): ExecutionPayl
   var payload = ExecutionPayload(
     parent_hash: latest.block_hash,
     state_root: latest.state_root, # no changes to the state
-    receipts_root: static(Eth2Digest.fromHex(
```
etan-status, line 355: spec function reference at top of this function is outdated.

beacon_chain/spec/helpers.nim
```diff
@@ -340,6 +344,26 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
...
+    coinbase      : EthAddress payload.fee_recipient.data,
+    stateRoot     : payload.state_root,
+    txRoot        : BLANK_ROOT_HASH,
+    receiptRoot   : payload.receipts_root,
+    bloom         : payload.logs_bloom.data,
+    difficulty    : default(DifficultyInt),
+    blockNumber   : payload.block_number.u256,
+    gasLimit      : GasInt payload.gas_limit,
+    gasUsed       : GasInt payload.gas_used,
+    timestamp     : fromUnix(int64 payload.timestamp),
```
arnetheduck: this looks like it could defect?

zah: The `timestamp` is not controlled by a third party in the current code, but are you suggesting a saturating conversion?

arnetheduck: yeah, I'd avoid the possibility altogether, just in case - these kinds of functions end up in unexpected contexts sooner or later

beacon_chain/spec/helpers.nim
zah: The cited spec function from a test module is pretty scandalous :) It has many TODO comments and spec-violating values.

beacon_chain/spec/helpers.nim
zah: Yes, these are merely conversion functions while the algorithm here aims to replicate the actions of the Engine API when building payloads.

beacon_chain/spec/helpers.nim
etan-status: Well, originally this was only used in testing code, hence the "FAKE RLP HASH" etc. Only later it got repurposed 😅 I think the link should be replaced entirely

beacon_chain/spec/helpers.nim
```diff
@@ -340,29 +344,51 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
   let slots_since_genesis = slot - GENESIS_SLOT
   state.genesis_time + slots_since_genesis * SECONDS_PER_SLOT
 
-# https://github.com/ethereum/consensus-specs/blob/v1.1.10/tests/core/pyspec/eth2spec/test/helpers/execution_payload.py#L1-L31
+proc emptyPayloadToBlockHeader*(payload: ExecutionPayload): ExecutionBlockHeader =
```
tersec, line 347:
```suggestion
func emptyPayloadToBlockHeader*(payload: ExecutionPayload): ExecutionBlockHeader =
```

zah: Let's sneak this in another PR. I don't want to restart the CI run just for it.

tersec: Sure, reasonable.

PR #3996, opened 2022-08-18
Ignore errors from Sqlite's ROLLBACK TRANSACTION

zah:
Multiple fleet nodes have been observed to fail on this line.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -400,7 +400,23 @@ template withManyWrites*(dbParam: BeaconChainDB, body: untyped) =
     if commit:
       expectDb db.db.exec("COMMIT TRANSACTION;")
     else:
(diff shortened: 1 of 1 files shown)


arnetheduck:
> Multiple fleet nodes have been observed to fail on this line.

but why did they fail? this seems pretty unusual?

zah:
> but why did they fail? this seems pretty unusual?

I haven't gotten to the bottom of this, but the question is whether this change is an improvement?
Technically, the code is more Sqlite API-compliant now, but perhaps it can hide a database write problem under the carpet, which may end up putting the database in an inconsistent state.

arnetheduck:
well, "technically" we should have gotten an error from an earlier call to sqlite, in order for the transaction to have been auto-aborted - I guess this change is safe, but it would be good to go deeper into the reason why it failed in the first place - ie this change likely won't fix any crashes: the "original" error will likely cause the application to crash regardless because we fail-fast on database write errors.

PR #3997, opened 2022-08-19
Doppelganger detection bug fix

zah:
When the client was started without any validators, the doppelganger
detection structures were never initialized properly. Later, when
validators were added through the Keymanager API, they interacted
with the uninitialized doppelganger detection structures and their
duties were inappropriately skipped.

### DIFF
--- a/beacon_chain/validator_client/doppelganger_service.nim
+++ b/beacon_chain/validator_client/doppelganger_service.nim
@@ -59,7 +59,7 @@ proc processActivities(service: DoppelgangerServiceRef, epoch: Epoch,
                     validator_index = vindex
             else:
               inc(value.epochsCount)
-              notice "Validator's activity was not seen",
+              debug "Validator's activity was not seen",
                      validator_index = vindex, epoch = epoch,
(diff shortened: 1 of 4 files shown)


beacon_chain/validator_client/doppelganger_service.nim
```diff
@@ -59,7 +59,7 @@ proc processActivities(service: DoppelgangerServiceRef, epoch: Epoch,
                     validator_index = vindex
             else:
               inc(value.epochsCount)
-              notice "Validator's activity was not seen",
+              debug "Validator's activity was not seen",
```
mratsim, line 62: This is something that appears only once so cannot spam, and it's related to validator earnings/security so I think notice is OK.

mratsim (approved):

beacon_chain/validator_client/doppelganger_service.nim
zah: It's being logged for every validator and it just tells the user something relatively obvious (the doppelganger detection is active), but with a somewhat unclear log message.

PR #3999, opened 2022-08-19
don't use empty execution payload when newPayload rejects it

tersec:
This still allows the empty payload fallback, but catches the specific `newPayload`-returns-`INVALID` scenario that has been occurring. Even with that fixed in general, it's useful to have this guardrail function.

Furthermore, https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#execution-engine-errors effectively specifies that only blocks for which the execution engine returns either `SYNCING` or `ACCEPTED` may be optimistically imported.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -398,9 +398,9 @@ proc getExecutionPayload(
           node.consensusManager.eth1Monitor.newExecutionPayload(payload),
           NEWPAYLOAD_TIMEOUT):
             info "getExecutionPayload: newPayload timed out"
-            PayloadExecutionStatus.syncing
+            Opt.none PayloadExecutionStatus
 
-    if executionPayloadStatus in [
+    if executionPayloadStatus.isNone or executionPayloadStatus.get in [
         PayloadExecutionStatus.invalid,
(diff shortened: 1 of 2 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -445,8 +447,33 @@ proc makeBeaconBlockForHeadAndSlot*(
...
+             not (
+               is_merge_transition_complete(proposalState.bellatrixData.data) or
+               ((not node.eth1Monitor.isNil) and
+                node.eth1Monitor.terminalBlockHash.isSome)):
+          # https://github.com/nim-lang/Nim/issues/19802
+          (static(default(bellatrix.ExecutionPayload)))
+        else:
+          let
+            pubkey = node.dag.validatorKey(validator_index)
+            maybeExecutionPayload = (await getExecutionPayload(
```
arnetheduck, line 466: looking at this code in general, we should really be running `getExecutionPayload` and attestation packing in parallel (ie call gep and store the future but await it after the packing) - the serial approach taken here is easier to reason about, but we've seen packing taking quite some time so..

tersec: Sure. This is meant as plausible for 22.8.0. That would not be.

PR #4002, opened 2022-08-19
reduce LC optsync latency

etan-status:
The optimistic sync spec was updated since the LC based optsync module
was introduced. It is no longer necessary to wait for the justified
checkpoint to have execution enabled; instead, any block is okay to be
optimistically imported to the EL client, as long as its parent block
has execution enabled. Complex syncing logic has been removed, and the
LC optsync module will now follow gossip directly, reducing the latency
when using this module. Note that because this is now based on gossip
instead of using sync manager / request manager, that individual blocks
may be missed. However, EL clients should recover from this by fetching
missing blocks themselves.


etan-status:
Tested manually with `nimbus_light_client`, as well as with `nimbus_beacon_node` on a fresh Sepolia sync with `--light-client`.

etan-status:
Spec: https://github.com/ethereum/consensus-specs/pull/2844 and https://github.com/ethereum/consensus-specs/pull/2881

beacon_chain/gossip_processing/optimistic_processor.nim
```diff
@@ -0,0 +1,216 @@
...
+  # Start of block processing - in reality, we have already gone through SSZ
+  # decoding at this stage, which may be significant
+  debug "Optimistic block received", delay
+
+  let v = self.validateBeaconBlock(signedBlock, wallTime)
+  if v.isErr:
+    debug "Dropping optimistic block", error = v.error
+    return err(v.error)
+
+  # Note that validation of blocks is delayed by ~4/3 slots because we have to
```
tersec, line 104: Is this 4-or-3 slots, or (4/3) slots?

tersec (approved):

beacon_chain/gossip_processing/optimistic_processor.nim
etan-status: 4 INTERVALS_PER_SLOT

PR #4004, opened 2022-08-19
Reviewed the Engine API calls for missing error handling

zah:


beacon_chain/validators/validator_duties.nim
```diff
@@ -390,11 +399,17 @@ proc getExecutionPayload(
...
-        GETPAYLOAD_TIMEOUT):
-          info "getExecutionPayload: getPayload timed out; using empty execution payload"
+      payload = try:
+          awaitWithTimeout(
+            get_execution_payload(payload_id, node.consensusManager.eth1Monitor),
+            GETPAYLOAD_TIMEOUT):
+              warn "Getting execution payload from Engine API timed out", payload_id
+              empty_execution_payload
+        except CatchableError as err:
+          warn "Getting execution payload from Engine API failed",
```
tersec, line 409: Why is one `error` and one `warn`?

tersec (approved):

beacon_chain/validators/validator_duties.nim
zah: My reasoning was that we have a recovery path in this case (the empty payload), but I agree that's quite debatable.

PR #4006, opened 2022-08-20
Require properly configured Engine API connection after the merge

zah:


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1341,6 +1355,17 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
 
   await m.ensureDataProvider()
 
+  if m.currentEpoch >= m.cfg.BELLATRIX_FORK_EPOCH:
+    let status = await m.exchangeTransitionConfiguration()
+    if status == localConfigurationUpdated:
+      info "Obtained terminal block from Engine API",
+           terminalBlockNumber = m.terminalBlockNumber.get.uint64,
+           terminalBlockHash = m.terminalBlockHash.get
+    elif isFirstRun:
```
tersec: This does only trigger in specific conditions, but even so, without https://github.com/ethereum/execution-apis/pull/285 being fully supported across the ecosystem, this might be a bit aggressive to quit on mismatch.

zah: I came up with a slightly different plan during the weekend. We can quit here, but we'll allow the user to silence the check with a command-line option. The goal is to inform the users that they must do something for the upgrade while allowing them to continue operating their node while they are setting up a full Ethereum node with an Engine API.

tersec: Sounds reasonable.

docs/the_nimbus_book/src/merge.md
```diff
@@ -22,6 +22,8 @@ As a node operator, you will need to run both an execution client and a consensu
 
 If you were running an execution client before, make sure to update its configuration to include an option for [JWT secrets](./eth1.md#3-pass-the-jwt-secret-to-nimbus) and engine API.
 
+Please note that once the Bellatrix fork epoch is reached on 6th of September 2022, Nimbus will refuse to start unless connected to a properly configured execution client. If you need more time to complete the transition, you can temporarily run the beacon node with the command-line option `--require-engine-api-in-bellatrix=no`, but please note that such a setup will stop working once the network TTD is reached (currently estimated to happen on 13th of September, see https://wenmerge.com/ for more up-to-date information).
```
tersec, line 25: https://bordel.wtf/ currently states
> Total Terminal Difficulty of 58750000000000000000000 is expected around Thu Sep 15 02:29 2022 UTC, i.e. between Thu Sep 15 01:00 2022 UTC and Thu Sep 15 03:59 2022 UTC 

But probably better here to go with the earlier of the plausible estimates (so the wenmerge one is probably better)

PR #4008, opened 2022-08-20
do not prune LC data by default

etan-status:
Aligns the default retention policy for LC data with the one for blocks.
Minimum spec requirement for both blocks and LC data is ~5 months.
Additional use cases are better supported by retaining data for longer.


etan-status:
Verified on Sepolia by manually testing with `--light-client-data-import-mode=full`, and checking that historic LC data is pruned when specifying a low `--light-client-data-max-periods`, and restored when removing the `--light-client-data-max-periods` override.

tersec (approved):

PR #4009, opened 2022-08-20
update EL head on slot and before attesting

etan-status:
`forkChoiceUpdated` was only triggered when selecting a new head as part
of storing a block. However, fork choice is also refreshed on slot start
and before sending attestations. Code has been streamlined to share the
same logic regardless of the context for fork choice.


etan-status:
This has become even messier with the addition of `runProposalForkchoiceUpdated`; should split the pure consolidation refactoring of the various fcu callers into separate PR.

PR #4011, opened 2022-08-20
support connecting to peers without bellatrix

etan-status:
Make discovery fork ID aware of scheduled Bellatrix fork to enable
connections to peers that don't have Bellatrix scheduled yet.
Without this, has peering issues with peers on older SW version.

### DIFF
--- a/beacon_chain/spec/network.nim
+++ b/beacon_chain/spec/network.nim
@@ -127,10 +127,10 @@ func getENRForkID*(cfg: RuntimeConfig,
     next_fork_epoch: cfg.nextForkEpochAtEpoch(epoch))
 
 func getDiscoveryForkID*(cfg: RuntimeConfig,
-                   epoch: Epoch,
-                   genesis_validators_root: Eth2Digest): ENRForkID =
-  # Until 1 epoch from fork, returns pre-fork value
-  if epoch + 1 >= cfg.ALTAIR_FORK_EPOCH:
+                         epoch: Epoch,
+                         genesis_validators_root: Eth2Digest): ENRForkID =
+  # Until 1 epoch from fork, return pre-fork value.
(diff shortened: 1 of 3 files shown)


etan-status:
This started breaking with https://github.com/status-im/nimbus-eth2/pull/3980

etan-status:
Originally found by @tersec as part of testing

tersec (approved):
I've tested this on Goerli and Mainnet.

beacon_chain/spec/network.nim
```diff
@@ -123,10 +123,10 @@ func getENRForkID*(cfg: RuntimeConfig,
...
 
 func getDiscoveryForkID*(cfg: RuntimeConfig,
-                   epoch: Epoch,
-                   genesis_validators_root: Eth2Digest): ENRForkID =
-  # Until 1 epoch from fork, returns pre-fork value
-  if epoch + 1 >= cfg.ALTAIR_FORK_EPOCH:
+                         epoch: Epoch,
+                         genesis_validators_root: Eth2Digest): ENRForkID =
+  # Until 1 epoch from fork, return pre-fork value.
+  if cfg.nextForkEpochAtEpoch(epoch) - epoch <= 1:
```
mratsim, line 129: What do other clients do there?

etan-status: Don't know, but it's the same logic as for Altair before. Note there is two values, an `enrForkId` and a `discoveryForkId`. The `discoveryForkId` is used to disconnect from peers who do not indicate support for the new fork 1 epoch before fork; it will remain compatible with both old and new peers until that epoch.

Testing at least shows that Altair Mainnet can connect to peers again (right now it can't on `unstable`, or only to other Nimbuses). And Bellatrix Goerli works as well. The transition I wrote unit tests for in this PR, the compatibility matrix in bottom of the test seems sane to me.

Issue #3953, opened 2022-08-11
Nimbus doesn't provide enough time for the EL node for block building


zah:
I won't provide detailed description here because @tersec is familiar with the issue.

zah:
Addressed by the linked PR above

Issue #3952, opened 2022-08-11
The MEV block building process should execute in parallel with the regular execution API block building process


zah:
This behavior is prescribed in the spec here:
https://github.com/ethereum/builder-specs/pull/46/files

Currently, Nimbus doesn't try to create a block through the Engine API when the block builder is enabled (see `proposeBlock` and `proposeBlockMEV` in the code).

zah:
Addressed by the linked PR above

PR #4012, opened 2022-08-22
allow execution clients several seconds to construct blocks

tersec:
Addresses https://github.com/status-im/nimbus-eth2/issues/3952 and https://github.com/status-im/nimbus-eth2/issues/3953

### DIFF
--- a/beacon_chain/consensus_object_pools/consensus_manager.nim
+++ b/beacon_chain/consensus_object_pools/consensus_manager.nim
@@ -27,6 +27,7 @@ type
     headBlockRoot*: Eth2Digest
     safeBlockRoot*: Eth2Digest
     finalizedBlockRoot*: Eth2Digest
+    timestamp*: uint64
     feeRecipient*: Eth1Address
 
   ConsensusManager* = object
@@ -53,7 +54,7 @@ type
 
     # Tracking last proposal forkchoiceUpdated payload information
     # ----------------------------------------------------------------
-    forkchoiceUpdatedInfo*: ForkchoiceUpdatedInformation
+    forkchoiceUpdatedInfo*: Opt[ForkchoiceUpdatedInformation]
 
 # Initialization
 # ------------------------------------------------------------------------------
@@ -72,7 +73,8 @@ func new*(T: type ConsensusManager,
     quarantine: quarantine,
     eth1Monitor: eth1Monitor,
     dynamicFeeRecipientsStore: dynamicFeeRecipientsStore,
-    keymanagerHost: keymanagerHost
+    keymanagerHost: keymanagerHost,
+    forkchoiceUpdatedInfo: Opt.none ForkchoiceUpdatedInformation
   )
 
 # Consensus Management
@@ -260,10 +262,11 @@ proc runProposalForkchoiceUpdated(self: ref ConsensusManager) {.async.} =
           fcResult.payloadId.isNone:
         return
 
-      self.forkchoiceUpdatedInfo = ForkchoiceUpdatedInformation(
+      self.forkchoiceUpdatedInfo = Opt.some ForkchoiceUpdatedInformation(
         payloadId: bellatrix.PayloadID(fcResult.payloadId.get),
         headBlockRoot: headBlockRoot,
         finalizedBlockRoot: finalizedBlockRoot,
+        timestamp: timestamp,
         feeRecipient: feeRecipient)
     except CatchableError as err:
       error "Engine API fork-choice update failed", err = err.msg
--- a/beacon_chain/consensus_object_pools/consensus_manager.nim
+++ b/beacon_chain/consensus_object_pools/consensus_manager.nim
@@ -16,7 +16,19 @@ import
   ../consensus_object_pools/[blockchain_dag, block_quarantine, attestation_pool],
   ../eth1/eth1_monitor
 
+from ../spec/eth2_apis/dynamic_fee_recipients import
+  DynamicFeeRecipientsStore, getDynamicFeeRecipient
+from ../validators/keystore_management import
+  KeymanagerHost, getSuggestedFeeRecipient
+
 type
+  ForkChoiceUpdatedInformation* = object
+    payloadId*: PayloadID
+    headBlockRoot*: Eth2Digest
+    safeBlockRoot*: Eth2Digest
+    finalizedBlockRoot*: Eth2Digest
+    feeRecipient*: Eth1Address
+
   ConsensusManager* = object
     expectedSlot: Slot
     expectedBlockReceived: Future[bool]
@@ -34,6 +46,15 @@ type
     # ----------------------------------------------------------------
     eth1Monitor*: Eth1Monitor
 
+    # Allow determination of preferred fee recipient during proposals
+    # ----------------------------------------------------------------
+    dynamicFeeRecipientsStore: DynamicFeeRecipientsStore
+    keymanagerHost: ref KeymanagerHost
+
+    # Tracking last proposal forkchoiceUpdated payload information
+    # ----------------------------------------------------------------
+    forkchoiceUpdatedInfo*: ForkchoiceUpdatedInformation
+
 # Initialization
 # ------------------------------------------------------------------------------
 
@@ -41,13 +62,17 @@ func new*(T: type ConsensusManager,
           dag: ChainDAGRef,
           attestationPool: ref AttestationPool,
           quarantine: ref Quarantine,
-          eth1Monitor: Eth1Monitor
+          eth1Monitor: Eth1Monitor,
+          dynamicFeeRecipientsStore: DynamicFeeRecipientsStore,
+          keymanagerHost: ref KeymanagerHost
          ): ref ConsensusManager =
   (ref ConsensusManager)(
     dag: dag,
     attestationPool: attestationPool,
     quarantine: quarantine,
-    eth1Monitor: eth1Monitor
+    eth1Monitor: eth1Monitor,
+    dynamicFeeRecipientsStore: dynamicFeeRecipientsStore,
+    keymanagerHost: keymanagerHost
   )
 
 # Consensus Management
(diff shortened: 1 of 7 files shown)


beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -182,6 +207,67 @@ proc updateHead*(self: var ConsensusManager, wallSlot: Slot) =
...
+        self.keymanagerHost[].defaultFeeRecipient
+    else:
+      self.keymanagerHost[].defaultFeeRecipient
+
+from ../spec/datatypes/bellatrix import PayloadID
+
+proc runProposalForkchoiceUpdated(self: ref ConsensusManager) {.async.} =
+  withState(self.dag.headState):
+    let
+      nextSlot = state.data.slot + 1
```
etan-status, line 237: Edge cases:
- During sync
- When the latest N slot were empty, e.g., we will propose `wallSlot + 1`, but `dag.headState.slot < wallSlot`

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -182,6 +207,67 @@ proc updateHead*(self: var ConsensusManager, wallSlot: Slot) =
...
+          fcResult.payloadId.isNone:
+        return
+
+      self.forkchoiceUpdatedInfo = ForkchoiceUpdatedInformation(
+        payloadId: bellatrix.PayloadID(fcResult.payloadId.get),
+        headBlockRoot: headBlockRoot,
+        finalizedBlockRoot: finalizedBlockRoot,
+        feeRecipient: feeRecipient)
+    except CatchableError as err:
+      error "Engine API fork-choice update failed", err = err.msg
```
etan-status, line 278: `strictVerification` assert

beacon_chain/validators/validator_duties.nim
```diff
@@ -405,11 +394,25 @@ proc getExecutionPayload(
...
-        proposalState.bellatrixData.data, latestHead, latestFinalized,
-        feeRecipient,
-        node.consensusManager.eth1Monitor))
+      feeRecipient = node.consensusManager.getFeeRecipient(pubkey, validator_index, epoch)
+      lastFcU = node.consensusManager.forkchoiceUpdatedInfo
+      payload_id =
+        if  lastFcU.headBlockRoot == latestHead and
+            lastFcU.finalizedBlockRoot == latestFinalized and
+            lastFcU.feeRecipient == feeRecipient:
+          some bellatrix.PayloadID(lastFcU.payloadId)
```
etan-status: How long do these `payloadId` stay valid? Considering we also have other fcu tasks scheduled concurrently that may select different heads.

beacon_chain/validators/validator_duties.nim
```diff
@@ -354,17 +354,6 @@ proc get_execution_payload(
     asConsensusExecutionPayload(
       await execution_engine.getPayload(payload_id.get))
 
-proc getFeeRecipient(node: BeaconNode,
-                     pubkey: ValidatorPubKey,
```
etan-status: Looks correct, but it doesn't seem necessary to move this if slated for 22.8. It introduces quite some noise to this PR, also with the `keymanagerHost` moving.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -197,6 +283,10 @@ proc updateHeadWithExecution*(self: ref ConsensusManager, newHead: BlockRef)
     # justified and finalized
     self.dag.updateHead(newHead, self.quarantine[])
 
+    # TODO after things stabilize with this, check for upcoming proposal and
+    # don't bother sending first fcU, but initially, keep both in place
+    asyncSpawn self.runProposalForkchoiceUpdated()
```
etan-status, line 297: There is yet another `forkChoiceUpdated` in `block_processor.nim` (`expectValidForkchoiceUpdated`). The function here (`updateHeadWithExecution`) is only called when a block is processed with a payload that initially returned an inconclusive result (syncing/accepted). The other one is called on blocks with payloads that already have `valid` verdict. It seems a bit counterintuitive to only do this if there was any recent `syncing/accepted` block but not if there only were`valid` block.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -182,6 +207,67 @@ proc updateHead*(self: var ConsensusManager, wallSlot: Slot) =
 
   self.updateHead(newHead)
 
+proc checkNextProposer(dag: ChainDAGRef, slot: Slot):
+    Opt[(ValidatorIndex, ValidatorPubKey)] =
+  let proposer = dag.getProposer(dag.head, slot + 1)
+  if proposer.isNone():
+    return Opt.none((ValidatorIndex, ValidatorPubKey))
+  Opt.some((proposer.get, dag.validatorKey(proposer.get).get().toPubKey))
```
etan-status, line 220: lgtm, `dag.getProposer` already validates that `dag.validatorKey` will return `some` result.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -182,6 +207,67 @@ proc updateHead*(self: var ConsensusManager, wallSlot: Slot) =
...
+
+    # Approximately lines up with validator_duties version. Used optimistcally/
+    # opportunistically, so mismatches are fine if not too frequent.
+    let
+      timestamp = compute_timestamp_at_slot(state.data, nextSlot)
+      randomData =
+        get_randao_mix(state.data, get_current_epoch(state.data)).data
+      feeRecipient = self.getFeeRecipient(
+        nextProposer, validatorIndex, nextSlot.epoch)
+      headBlockRoot = self.dag.loadExecutionBlockRoot(self.dag.head)
```
etan-status, line 250: As this is called from an async context, `dag.head` may already point to a new head by this time. Around the merge, this could be a non-execution block with `headBlockRoot.isZero`, which is not documented as an allowed value for `forkChoiceUpdated` (in the `runForkchoiceUpdated` wrapper, it even asserts for that)

beacon_chain/validators/validator_duties.nim
```diff
@@ -405,11 +394,25 @@ proc getExecutionPayload(
...
-      payload_id = (await forkchoice_updated(
-        proposalState.bellatrixData.data, latestHead, latestFinalized,
-        feeRecipient,
-        node.consensusManager.eth1Monitor))
+      feeRecipient = node.consensusManager.getFeeRecipient(pubkey, validator_index, epoch)
+      lastFcU = node.consensusManager.forkchoiceUpdatedInfo
+      payload_id =
+        if  lastFcU.headBlockRoot == latestHead and
+            lastFcU.finalizedBlockRoot == latestFinalized and
+            lastFcU.feeRecipient == feeRecipient:
```
etan-status: If `latestHead` ends up as zero (due to no `terminalBlockHash` being reached / pre-merge), this may be triggered unintentionally (as `lastFcu` was never initialized), and resolve to `some default(PayloadId)`, so will run `execution_engine.getPayload` instead of using the pre-merge `default(bellatrix.ExecutionPayload)`.

etan-status (approved):
The changes are mostly of best-effort nature and might be okay for 22.8, given that validator performance is tied to it. I don't have the metrics though to see how frequently this change actually would improve the situation.

tersec:
> The changes are mostly of best-effort nature and might be okay for 22.8, given that validator performance is tied to it. I don't have the metrics though to see how frequently this change actually would improve the situation.

Tha's the intent, yes. It's not an optimal design from scratch.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: Wastes a bit of time on the EL locally, nothing else.

beacon_chain/validators/validator_duties.nim
tersec: Hm, yeah, it was one of the issues I wondered about. For now, I agree, it's better to leave in place, reduce risk, though the duplicate logic should be removed eventually.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: Not sure if ideal. Those are maybe best when it's verifying something truly Nimbus-internal, but this isn't. There are exogenous factors involved here, and `strictVericiation`, though strict, shouldn't trigger except on actual Nimbus logic errors fixable solely in Nimbus code.

beacon_chain/validators/validator_duties.nim
tersec: A few seconds at least, within a slot. That's fine though, if the head/finalized/fee recipient (i.e. fork choice parameters, though maybe prudent to add the rest, e.g., `timestamp`) mismatch, it just won't be used.

beacon_chain/validators/validator_duties.nim
tersec: `getExecutionPayload` is only called in a merge context:
https://github.com/status-im/nimbus-eth2/blob/e7f6aac08c30def31c451ce8f8631cfd6b6b6e9b/beacon_chain/validators/validator_duties.nim#L487-L506

tersec: I did consider this and, could wrap the whole thing in an `Opt`/`Optional` to not rely on this though. It's not something I'd do in a more optimal design, probably, but might make sense here.

tersec: https://github.com/status-im/nimbus-eth2/pull/4012/commits/223f949fe9372a7ddc5e2c1686ebc036c73fb583 (a) adds a `timestamp` check which should never accidentally match and (b) adds an explicit initialization check (which makes things uglier enough I'd typically avoid it, but it's more risk-conservative)

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4012/commits/223f949fe9372a7ddc5e2c1686ebc036c73fb583 restores the `validator_duties` `getFeeRecipient` setup, except for adding a comment to remove it later.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/223f949fe9372a7ddc5e2c1686ebc036c73fb583 adds `timestamp`, so the validity length per se isn't that crucial. It's always within a slot this way, and that's definitely fine.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: It has to wait until `dag.updateHead` runs, which happens after the bulk of the `block_processor`/`consensus_manager` logic, because that's when/where `dag.headState` is updated, which is necessary to get the RANDAO information from said state.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4012/commits/f8b4984bcb2445f9ef2fa395585d840b807480a0

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4012/commits/f8b4984bcb2445f9ef2fa395585d840b807480a0 runs it when `newPayload` is `VALID`

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -34,20 +47,34 @@ type
     # ----------------------------------------------------------------
     eth1Monitor*: Eth1Monitor
 
+    # Allow determination of preferred fee recipient during proposals
+    # ----------------------------------------------------------------
+    dynamicFeeRecipientsStore: DynamicFeeRecipientsStore
```
zah: Er, is this a second instance besides the one in the `BeaconNode`? This is not reasonable, because this object is configured through the REST API (see "/eth/v1/validator/prepare_beacon_proposer"). A separate instance would be useless unless hooked with that API end-point.

tersec: https://github.com/status-im/nimbus-eth2/pull/4012/commits/bea11e893f344c6b21af9c42509325df52e683cc

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -182,6 +209,71 @@ proc updateHead*(self: var ConsensusManager, wallSlot: Slot) =
...
+
+proc getFeeRecipient*(
+    self: ref ConsensusManager, pubkey: ValidatorPubKey, validatorIdx: ValidatorIndex,
+    epoch: Epoch): Eth1Address =
+  self.dynamicFeeRecipientsStore[].getDynamicFeeRecipient(validatorIdx, epoch).valueOr:
+    if self.keymanagerHost != nil:
+      self.keymanagerHost[].getSuggestedFeeRecipient(pubkey).valueOr:
+        self.keymanagerHost[].defaultFeeRecipient
+    else:
+      self.keymanagerHost[].defaultFeeRecipient
```
zah: `self.keymanagerHost` is `nil` here (see the check above). That's why the original definition of this function needed to access the value from the `BeaconNode` config.

tersec: https://github.com/status-im/nimbus-eth2/pull/4012/commits/00ba2b2e0a4d2fa2f61cfded73c884530140b983

PR #4013, opened 2022-08-22
Harden block proposal against expired slashings/exits

arnetheduck:
When a message is signed in a phase0 domain, it can no longer be
validated under bellatrix due to the correct fork no longer being
available in the `BeaconState`.

To ensure that all slashing/exits are still valid, in this PR we re-run
the checks in the state that we're proposing for, thus hardening against
both signatures and other changes in the state that might have
invalidated the message.


arnetheduck:
Hats off to @paulhauner for this one!

zah:
Will there be a spec change for this?

arnetheduck:
> Will there be a spec change for this?

this has been in the spec since phase0 but it only becomes an issue now with a third fork (the beaconstate holds two forks, current and previous)

PR #4019, opened 2022-08-24
remove some Bellatrix-specific references

tersec:
As Capella approaches, various of these assumptions need to be removed.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -376,22 +376,23 @@ proc getFeeRecipient(node: BeaconNode,
 from web3/engine_api_types import PayloadExecutionStatus
 
 proc getExecutionPayload(
-    node: BeaconNode, proposalState: auto,
+    node: BeaconNode, proposalState: ref ForkedHashedBeaconState,
     epoch: Epoch,
     validator_index: ValidatorIndex,
     pubkey: ValidatorPubKey):
     Future[Opt[ExecutionPayload]] {.async.} =
   # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
-
-  # Only current hardfork with execution payloads is Bellatrix
-  static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
-  template empty_execution_payload(): auto =
-    build_empty_execution_payload(proposalState.bellatrixData.data)
+  func empty_execution_payload(): ExecutionPayload =
+    withState(proposalState[]):
+      when stateFork >= BeaconStateFork.Bellatrix:
+        build_empty_execution_payload(state.data)
+      else:
+        default(ExecutionPayload)
 
   if node.eth1Monitor.isNil:
     beacon_block_payload_errors.inc()
     warn "getExecutionPayload: eth1Monitor not initialized; using empty execution payload"
-    return Opt.some empty_execution_payload
+    return Opt.some empty_execution_payload()
 
   try:
     # Minimize window for Eth1 monitor to shut down connection
@@ -399,6 +400,8 @@ proc getExecutionPayload(
 
     # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#request-2
     const GETPAYLOAD_TIMEOUT = 1.seconds
+
+    static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
 
     let
       terminalBlockHash =
@@ -448,12 +451,12 @@ proc getExecutionPayload(
             GETPAYLOAD_TIMEOUT):
               beacon_block_payload_errors.inc()
               warn "Getting execution payload from Engine API timed out", payload_id
-              empty_execution_payload
+              empty_execution_payload()
         except CatchableError as err:
           beacon_block_payload_errors.inc()
           warn "Getting execution payload from Engine API failed",
                 payload_id, err = err.msg
-          empty_execution_payload
+          empty_execution_payload()
 
       executionPayloadStatus =
         awaitWithTimeout(
@@ -474,7 +477,7 @@ proc getExecutionPayload(
     beacon_block_payload_errors.inc()
     error "Error creating non-empty execution payload; using empty execution payload",
       msg = err.msg
-    return Opt.some empty_execution_payload
+    return Opt.some empty_execution_payload()
 
 proc makeBeaconBlockForHeadAndSlot*(
     node: BeaconNode, randao_reveal: ValidatorSig,
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -316,21 +316,16 @@ proc getBlockProposalEth1Data*(node: BeaconNode,
 
 from web3/engine_api import ForkchoiceUpdatedResponse
 
-# TODO: This copies the entire BeaconState on each call
-proc forkchoice_updated(state: bellatrix.BeaconState,
-                        head_block_hash: Eth2Digest,
-                        safe_block_hash: Eth2Digest,
-                        finalized_block_hash: Eth2Digest,
-                        fee_recipient: ethtypes.Address,
-                        execution_engine: Eth1Monitor):
-                        Future[Option[bellatrix.PayloadID]] {.async.} =
+proc forkchoice_updated(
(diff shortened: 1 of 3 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -369,21 +369,22 @@ proc getFeeRecipient(node: BeaconNode,
 from web3/engine_api_types import PayloadExecutionStatus
 
 proc getExecutionPayload(
-    node: BeaconNode, proposalState: auto,
+    node: BeaconNode, proposalState: ref ForkedHashedBeaconState,
```
tersec, line 374: Otherwise, the usual issue from the `state` function exported deep in Chronos emerges, e.g.
```nim
# Can be in library, or other code not directly controlled
proc state(x: int) = discard

type A = object
  data: string

template withState(x: A, body: untyped): untyped =
  # `state` here sort of can be changed, but still costly
  template state: untyped {.inject.} = x
  body

withState(A()):
  echo(state.data)
```

When `data` here is just a function one is calling using the method syntax, that's irrelevant syntax sugar that's trivial to work around, but since `data`  is a field of the result of the `state` template, there aren't as many good workarounds.

beacon_chain/validators/validator_duties.nim
```diff
@@ -369,21 +369,22 @@ proc getFeeRecipient(node: BeaconNode,
...
     validator_index: ValidatorIndex,
     pubkey: ValidatorPubKey):
     Future[Opt[ExecutionPayload]] {.async.} =
   # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
-
-  # Only current hardfork with execution payloads is Bellatrix
-  static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
-  template empty_execution_payload(): auto =
-    build_empty_execution_payload(proposalState.bellatrixData.data)
+  func empty_execution_payload(): ExecutionPayload =
```
tersec: This can't be a `template` and can't be generic for similar reasons relating to allowing the `withState` template to work in the presence of the aforementioned exported Chronos HTTP server function. It's unfortunate, because a template would be better, but it's run not very often, at most once every slot in some testnet, so it's an acceptable workaround for the `state`-reference ambiguity.

beacon_chain/validators/validator_duties.nim
arnetheduck: oops, right this is actually really bad: `auto` introduces a _copy_ of the state (in `async` functions) - I think @zah was the first to point this out in some unrelated PR... in 1.6 we can turn the copy operator into an error, but not in 1.2 :/

tersec: Seems odd, my mental model is that it just gets specialized/monomorphized/lowered into a non-generic function at compile time.

Anyway, does this mean that this change is a fix for that regardless?

arnetheduck: `async` creates a closure iterator which effectively is a heap-based copy of any state that you pass into the function, so as to ensure that the data is available when async execution resumes after having been interrupted by a wait.

changing to `ref` indeed removes the copy of the data it in the sense that the reference is copied into the closure instead - this is dangerous for other reasons: in particular, the referenced object that you're passing into the function may now change between `await` calls in the function because `ref` implies shared mutability. We've had several bugs where this happens: an async operation starts, then the underlying state is reorged while an await is pending.

In this case, it would be better not to pass a state at all, but rather pass (a copy of) the minimal data needed to construct the empty payload.

arnetheduck: actually, the proposal state is local to each block production (it's already a copy of a state) so it should be safe to use `ref` - the danger remains that someone modifies the ref, but in this case it "shouldn't" happen - we can't express `ref to const` in nim unfortunately

beacon_chain/validators/validator_duties.nim
zah: Your other `forkyState` PR is merged now, so you can revisit this if you wish.
I think using a function here is nice, but it doesn't need to be a closure IMO. The proposal state can be passed in as an argument (a helper template can keep the call-site code as is today).

beacon_chain/validators/validator_duties.nim
zah: > Seems odd, my mental model is that it just gets specialized/monomorphized/lowered into a non-generic function at compile time.

If this function was already being passed a `ref` then it wouldn't have copied the argument (`auto` will bind happily to a `ref` value), so your mental model is mostly right.

beacon_chain/validators/validator_duties.nim
tersec: I'm not sure what the advantage of the function + helper template version is though, compared with the previous template.
```nim
    withState(proposalState[]):
      when stateFork >= BeaconStateFork.Bellatrix:
        build_empty_execution_payload(state.data)
      else:
        default(ExecutionPayload)
```
expands into something reasonably small, in actual code (not compile-time template overhead)

zah: Using a template is fine. My comment was that you already made `empty_execution_payload` a function, but it's currently a closure while it doesn't need to be. The closures are not very expensive in Nim, but they still introduce memory allocations, GC cycles and other subtle changes to the codegen when there is more than one closure in a particular function, so it's good to avoid them when that's trivial (as it is here).

tersec: After looking more at https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#executionpayload it's clear that this needs larger changes and that the static assert can't simply be removed, since it returns an `ExecutionPayload`, but soon that might mean either a `bellatrix.ExecutionPayload` or `capella.ExecutionPayload`. The required changes are outside the scope of what this PR was initially intended for, which was kind of a pass through the static Bellatrix-asserts where they could be readily, in local isolation, replaced by something more adaptable.

My general preference here is that the `getExecutionPayload`-realm end up using generically specified return types, rather than yet another `ForkedFoo` type which is fundamentally unnecessary (haven't looked at the developing branch along these lines, don't know how it's organized here) with its own suite of `withExecutionPayload`, etc. These things are part of the state transition function, and largely are only processed in any detail in already fork-specific parts of code.

tersec: https://github.com/status-im/nimbus-eth2/pull/4019/commits/6fa6f5dfbca8a0233e52a780e9f18b3e2a92388e uses a template.

The `BlindedBeaconBlock` parts need a slightly more custom solution, outside the scope of this PR.

beacon_chain/validators/validator_duties.nim
```diff
@@ -623,20 +618,22 @@ macro copyFields(
         # unblinded objects, and can't simply be copied.
         "transactions_root", "execution_payload",
         "execution_payload_header", "body"]:
+      # TODO use stew/assign2
       result.add newAssignment(
         newDotExpr(dst, ident(name)), newDotExpr(src, ident(name)))
 
-proc getBlindedBeaconBlock(
+proc getBlindedBeaconBlock[T](
```
arnetheduck, line 625: does `T: bellatrix.SignedBlindedBeaconBlock | some.OtherBlock` work?

arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
tersec: It won't fully work yet, since there's a macro-based setup which copies all but certain fields between the blinded/unblinded paylods and blocks. Finishing that part is outside the scope of this PR, but will involve changing
```nim
  const
    blckFields = getFieldNames(typeof(forkedBlock.bellatrixData))
    blckBodyFields = getFieldNames(typeof(forkedBlock.bellatrixData.body))

  # https://github.com/ethereum/builder-specs/blob/v0.2.0/specs/validator.md#block-proposal
  var blindedBlock: T

  copyFields(blindedBlock.message, forkedBlock.bellatrixData, blckFields)
  copyFields(
    blindedBlock.message.body, forkedBlock.bellatrixData.body, blckBodyFields)
  blindedBlock.message.body.execution_payload_header = executionPayloadHeader
```
so that instead of `forkedBlock.bellatrixData` it should be able to use `T` now and instead of `forkedBlock.bellatrixData.body`, `T.data`, and checking that the `copyFields` macro system still works (since it was initially designed to operate off the non-blinded block type, but it should work either way, since by design it copies the shared subset of fields between them).

PR #4026, opened 2022-08-24
allow accessing withState forky state via `forkyState`

tersec:
```nim
proc state*(server: HttpServerRef): HttpServerState {.raises: [Defect].} =
  ## Returns current HTTP server's state.
  if server.lifetime.finished():
  ...
```
in Chronos httpserver.nim and
```nim
template withState*(x: ForkedHashedBeaconState, body: untyped): untyped =
  case x.kind
  of BeaconStateFork.Bellatrix:                                       
    const stateFork {.inject, used.} = BeaconStateFork.Bellatrix
    template state: untyped {.inject, used.} = x.bellatrixData
    body
```
in nimbus-eth2 `spec/forks.nim` regularly conflict, e.g., https://github.com/status-im/nimbus-eth2/pull/4019

Nim sees `state` in attempts to access the `state` template within a `withState` block and uses the Chronos procedure instead. There are easier ways around this sometimes, but fewer when it's a genuine field access.

A pure reproduction of this phenomenon is
```nim
# Can be in library, or other code not directly controlled
proc state(x: int) = discard

type A = object
  data: string

template withState(x: A, body: untyped): untyped =
  # `state` here sort of can be changed, but still costly
  template state: untyped {.inject.} = x
  body

withState(A()):
  echo(state.data)
```
Which will result in a compile-time error:
```
repro.nim(12, 10) template/generic instantiation of `withState` from here
repro.nim(13, 15) Error: undeclared field: 'data'
```

There are various workarounds, e.g., when `data` is a function, but when it's a field like this, there are fewer. So instead of continuing to gamble each `withState` usage that one will have to spend some time working around this, just rename the `withState` variable. `forkyState` is a bit clearer too, as often currently people name both the outer (forked) and inner (forky) state `state`, shadowing one with the other.

This PR doesn't complete the conversion, because there are some large PRs outstanding, and it's better to let them land, and gradually do this, while unblocking progress in the PR inspiring this which involves adding a bunch of `withState` blocks in different places exposed to Chronos's `state` function.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -487,10 +487,10 @@ proc getState(
     state = (ref ForkedHashedBeaconState)(kind: fork)[]
 
   withState(state):
-    if not db.getState(state_root, state.data, rollback):
+    if not db.getState(state_root, forkyState.data, rollback):
       return false
 
-    state.root = state_root
+    forkyState.root = state_root
 
   true
 
@@ -608,7 +608,7 @@ proc currentSyncCommitteeForPeriod*(
   dag.withUpdatedState(tmpState, bsi) do:
     withState(state):
       when stateFork >= BeaconStateFork.Altair:
-        ok state.data.current_sync_committee
+        ok forkyState.data.current_sync_committee
       else: err()
   do: err()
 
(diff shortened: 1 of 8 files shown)


etan-status:
Would it be viable to simply rename the HttpServer `state` proc to `serverState`?

etan-status:
What also works in your minimal example is to use a `block` scope.

```
# Can be in library, or other code not directly controlled
proc state(x: int) = discard

type A = object
  data: string

template withState(x: A, body: untyped): untyped =
  # `state` here sort of can be changed, but still costly
  block:
    template state: untyped {.inject.} = x
    body

withState(A()):
  echo(state.data)
```

tersec:
> Would it be viable to simply rename the HttpServer `state` proc to `serverState`?

Potentially, though it might be necessary not to simply rename but rather to gradually deprecate the existing symbol, which helps, but not soon enough. I agree that'd have been the better approach to begin with, but it might be a bit late now to rely solely on that. @cheatfate can comment more though.

etan-status:
If the `block:` workaround also works with `withState`, no renaming is needed.

zah:
The `block` work-around should not change anything. The problem with this construct is Nim's "pre-pass" for generics and templates which tries to makes sense of the code before the template is expanded. Since the compiler is able to see an existing `state` symbol before expanding the template, it turns the "state" identifier into something called a "a sym choice" symbol and this prevents further lookups after the template is expanded. 

It could be argued that in this particular case, Nim is not following its own rules, because the sym choice should have been "open" (since `state` is overloaded), which means that other scopes can later contribute to it, but this doesn't kick in here.

PR #4027, opened 2022-08-24
fix engine API crash when EL disconnected

etan-status:
When issuing an engine API call while the EL is disconnected, a `nil`
pointer is dereferenced. Fixed by correctly initializing futures.

```
Traceback (most recent call last, using override)
vendor/nim-libp2p/libp2p/protocols/pubsub/pubsub.nim(890) main
beacon_chain/nimbus_beacon_node.nim(2139) main
beacon_chain/nimbus_beacon_node.nim(0) handleStartUpCmd
beacon_chain/nimbus_beacon_node.nim(0) doRunBeaconNode
beacon_chain/nimbus_beacon_node.nim(0) start
beacon_chain/nimbus_beacon_node.nim(1589) run
vendor/nimbus-build-system/vendor/Nim/lib/system/iterators_1.nim(107) poll
vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
beacon_chain/consensus_object_pools/consensus_manager.nim(297) updateHeadWithExecution
vendor/nim-chronos/chronos/asyncmacro2.nim(213) runProposalForkchoiceUpdated
vendor/nim-chronos/chronos/asyncfutures2.nim(365) futureContinue
beacon_chain/consensus_object_pools/consensus_manager.nim(259) runProposalForkchoiceUpdated
beacon_chain/eth1/eth1_monitor.nim(0) forkchoiceUpdated
vendor/nim-chronos/chronos/asyncfutures2.nim(219) complete
vendor/nim-chronos/chronos/asyncfutures2.nim(149) cancelled
vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(610) signalHandler
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -478,7 +478,7 @@ proc getPayload*(p: Eth1Monitor,
   # Eth1 monitor can recycle connections without (external) warning; at least,
   # don't crash.
   if p.isNil or p.dataProvider.isNil:
-    var epr: Future[engine_api.ExecutionPayloadV1]
+    let epr = newFuture[engine_api.ExecutionPayloadV1]("getPayload")
     epr.complete(default(engine_api.ExecutionPayloadV1))
     return epr
 
@@ -489,7 +489,7 @@ proc newPayload*(p: Eth1Monitor, payload: engine_api.ExecutionPayloadV1):
   # Eth1 monitor can recycle connections without (external) warning; at least,
   # don't crash.
   if p.dataProvider.isNil:
-    var epr: Future[PayloadStatusV1]
+    let epr = newFuture[PayloadStatusV1]("newPayload")
     epr.complete(PayloadStatusV1(status: PayloadExecutionStatus.syncing))
     return epr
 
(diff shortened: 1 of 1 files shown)


arnetheduck (approved):

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -478,7 +478,7 @@ proc getPayload*(p: Eth1Monitor,
   # Eth1 monitor can recycle connections without (external) warning; at least,
   # don't crash.
   if p.isNil or p.dataProvider.isNil:
-    var epr: Future[engine_api.ExecutionPayloadV1]
+    let epr = newFuture[engine_api.ExecutionPayloadV1]("getPayload")
```
arnetheduck, line 481: it's slightly odd that we complete with an empty execution payload - shouldn't this be completed an exception instead via `fail`? cc @tersec

tersec: `p.isNil` and `p.dataProvider.isNil` are slightly different cases in that regard.

`p.isNil` is effectively a misconfiguration, and should be detected elsewhere. What it does here isn't that important. It's up to convenience, and it's convenient to make it the same as `p.dataProvider.isNil`.

(Really, there's just not much use case for `nil` `Eth1Monitor` anymore. That's in itself kind of a bug post-merge, even to just sync or follow the chain.)

`p.dataProvider.isNil` is the recycling-in-progress case, and probably, yes, it should return some kind of more explicit error, whether that's an exception or something else.

zah: When there is no active connection, I think the code here can wait on a future signalling the moment the connection is restored. You'll do that for up to a second for example instead of bailing out immediately.

etan-status: Expected connection resets only happen during sync though, when `eth1_monitor` doesn't see new blocks for 5 minutes because EL is syncing and not providing them; so, not sure how much it helps in practice for `getPayload` (which is only called when in sync). For now, I fixed the nil pointer crash (`var` initializes to `0` and `Future` is a `ref`)

PR #4028, opened 2022-08-25
bump `news` to fix concurrency bug with >1MB frames

etan-status:
When calling `newPayload` on a >1MB payload (can happen post-merge),
`news` splits up that payload into 1MB chunks. The chunks are each sent
individually, though, with `await` in-between. This means that when we
send concurrent `forkChoiceUpdated` calls, that those may end up getting
in-between the `newPayload` chunks, leading to invalid data being sent.
The EL then returns an error message with a `null` `id` entry (as it
could not read the request `id` due to the mangling) and disconnects.
A PR has been submitted to fix this in `news`, and merged into `status`
branch early as this fix is critical for reliable post-merge operation:
https://github.com/Tormund/news/pull/22


tersec:
We've mostly kept the Status news repo used aligned with upstream, but this is an important enough fix that if we need to, it's okay to push it regardless.

The other possibility is trying to resurrect `nim-websock` once again, but maybe not the thing to gamble on given the not-great record the last few times.

arnetheduck:
the workaround being to use http instead, I guess?

arnetheduck (approved):

etan-status:
I guess, HTTP has a different set of issues. WebSocket based setups should work, though. They are not documented to be broken, and expecting node operators to update their configurations yet another time is not feasible. But yes, experience with HTTP may vary in case of trouble.

etan-status:
Manually tested by doing a full Sepolia sync on empty Geth / Nimbus DB with websocket.

PR #4030, opened 2022-08-25
update engine API spec ref URLs from alpha.9 to beta.1

tersec:
https://github.com/ethereum/execution-apis/releases/tag/v1.0.0-beta.1 lists other changes.

In the engine API namespace:
-  "Auth spec shouldn't specify which methods are provided over it by @lightclient" properly applies only to ELs
- "Engine API: respond with error if payload attributes are invalid by @mkalinin" similarly modifies constraints on ELs, not directly CLs -- correct usage on CLs won't hit that either way
- "engine timeouts (redux) by @djrtwo" is done
- "Engine API: adjust error codes with JSON-RPC 2.0 spec by @mkalinin"  is done
- "Engine API: return error if forkchoice state is inconsistent by @mkalinin" applies to ELs -- to the extent it affects CLs, something has already gone wrong and CL just sees whatever error message, up to EL (but hopefully within spec)
- "Engine API: remove unauth port and refer to auth spec by @mkalinin" is likewise basically for ELs
- "Clarify the difference between `SYNCING` and `ACCEPTED` by @djrtwo" Nimbus treats them the same, which is fine and typical
- "Engine API: replace INVALID_TERMINAL_BLOCK with INVALID + lvh=0x00..00 by @mkalinin" done
- "Engine API: a few clarifications by @mkalinin" nothing relevant to change in Nimbus
- "Make td optional field on block by @lightclient" I see why it shows up here, but it's not really about the engine API directly
- "require JWT keys to be exactly 32 bytes by @tersec" -- constraint mostly on EL (JWT-token-consuming) side in this case. Nimbus always used 32 by default and now constrains JWT tokens it reads as well
- "Engine API: narrow down fcU skip conditions by @mkalinin" applies to ELs
- "Engine API: recommend to retry the call after timeout by @mkalinin" happens now by reinserting a timed-out block in the block processing queue
- "Increase JWT issued-at window to 60s by @dapplion" doesn't directly affect Nimbus, though it's nice. Updated some comment text.

### DIFF
--- a/beacon_chain/spec/engine_authentication.nim
+++ b/beacon_chain/spec/engine_authentication.nim
@@ -25,9 +25,9 @@ proc base64urlEncode(x: auto): string =
   base64.encode(x, safe = true).replace("=", "")
 
 func getIatToken*(time: int64): JsonNode =
-  # https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.9/src/engine/authentication.md#jwt-claims
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/authentication.md#jwt-claims
   # "Required: iat (issued-at) claim. The EL SHOULD only accept iat timestamps
-  # which are within +-5 seconds from the current time."
+  # which are within +-60 seconds from the current time."
   #
   # https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6 describes iat
   # claims.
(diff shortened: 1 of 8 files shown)


beacon_chain/spec/engine_authentication.nim
```diff
@@ -92,10 +91,10 @@ proc checkJwtSecret*(
       # Secret JWT key is parsed in constant time using nimcrypto:
       # https://github.com/cheatfate/nimcrypto/pull/44
       let secret = utils.fromHex(lines[0])
-      if secret.len >= MIN_SECRET_LEN:
+      if secret.len == SECRET_LEN:
         ok(secret)
       else:
-        err("JWT secret not at least 256 bits")
+        err("JWT secret not 256 bits")
```
etan-status, line 97: On the EF test fleet, there is an extra `0x` prefix for historical reason. Maybe better to just warn in that case, in case `utils.fromHex` doesn't already catch that

tersec: A `0x` prefix is fine though. It was decided that the two variations which were fine were:
- `0x` or not `0x` prefix
- trailing newline or not

Issue #4015, opened 2022-08-22
Document the external block builder options


zah:

tersec:
Done: https://github.com/status-im/nimbus-eth2/pull/4032

PR #4032, opened 2022-08-26
document external block builder configuration

tersec:
https://github.com/status-im/nimbus-eth2/issues/4015

### DIFF
--- a/docs/the_nimbus_book/src/external-block-builder.md
+++ b/docs/the_nimbus_book/src/external-block-builder.md
@@ -1,0 +1,21 @@
+# Set up MEV
+
+[Maximal extractable value](https://ethereum.org/en/developers/docs/mev/) involves consensus clients contacting an external execution payload builder which might maximize profit or some other defined metric in ways hindered for a purely local consensus and execution client setup. This external builder network uses the [builder API](https://ethereum.github.io/builder-specs/) which consensus clients use to access MEV bundles found by searchers. In exchange, such searchers and builders might choose to retain some of the profit gained from such bundles. A builder API relay provides access to multiple searchers via a single URL.
+
+Nimbus supports this builder API to access these MEV builders. If one is configured, the block production flow becomes modified:
+1. attempt to use the specified MEV relay or builder to create an execution payload
+2. if the MEV builder or relay doesn't function, then fall back to existing local execution client
+
+There does exist a failure mode, intrinsic to the builder API, wherein the consensus client has signed a blinded proposal and therefore even if the MEV relay or builder doesn't provide a full block, a consensus client such as Nimbus cannot safely proceed with step 2 and fall back on its local execution client.
+
+## Command line option
+
+=== "Mainnet"
+    ```sh
+    ./run-mainnet-beacon-node.sh --payload-builder=true --payload-builder-url=https://${HOST}:${PORT}/
+    ```
+
+=== "Prater"
+    ```sh
+    ./run-prater-beacon-node.sh --payload-builder=true --payload-builder-url=https://${HOST}:${PORT}/
+    ```
--- a/docs/the_nimbus_book/src/external-block-builder.md
+++ b/docs/the_nimbus_book/src/external-block-builder.md
@@ -1,4 +1,4 @@
-# Set up MEV
+# Set up block builders
 
 [Maximal extractable value](https://ethereum.org/en/developers/docs/mev/) involves consensus clients contacting an external execution payload builder which might maximize profit or some other defined metric in ways hindered for a purely local consensus and execution client setup. This external builder network uses the [builder API](https://ethereum.github.io/builder-specs/) which consensus clients use to access MEV bundles found by searchers. In exchange, such searchers and builders might choose to retain some of the profit gained from such bundles. A builder API relay provides access to multiple searchers via a single URL.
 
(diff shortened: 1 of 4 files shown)


docs/the_nimbus_book/src/external-block-builder.md
```diff
@@ -0,0 +1,21 @@
+# Set up MEV
```
arnetheduck:
```suggestion
# Set up block builders
```
... or something else that relates to the actual act?

docs/the_nimbus_book/src/external-block-builder.md
```diff
@@ -0,0 +1,21 @@
...
+
+Nimbus supports this builder API to access these MEV builders. If one is configured, the block production flow becomes modified:
+1. attempt to use the specified MEV relay or builder to create an execution payload
+2. if the MEV builder or relay doesn't function, then fall back to existing local execution client
+
+There does exist a failure mode, intrinsic to the builder API, wherein the consensus client has signed a blinded proposal and therefore even if the MEV relay or builder doesn't provide a full block, a consensus client such as Nimbus cannot safely proceed with step 2 and fall back on its local execution client.
+
+## Command line option
```
arnetheduck, line 14: we should unhide the options too if they're getting documented, they're stable now? Also, do we need both url:s and `--payload-builder=true`? isn't the flag redundant?

docs/the_nimbus_book/src/external-block-builder.md
```diff
@@ -0,0 +1,21 @@
+# Set up MEV
+
+[Maximal extractable value](https://ethereum.org/en/developers/docs/mev/) involves consensus clients contacting an external execution payload builder which might maximize profit or some other defined metric in ways hindered for a purely local consensus and execution client setup. This external builder network uses the [builder API](https://ethereum.github.io/builder-specs/) which consensus clients use to access MEV bundles found by searchers. In exchange, such searchers and builders might choose to retain some of the profit gained from such bundles. A builder API relay provides access to multiple searchers via a single URL.
+
+Nimbus supports this builder API to access these MEV builders. If one is configured, the block production flow becomes modified:
+1. attempt to use the specified MEV relay or builder to create an execution payload
+2. if the MEV builder or relay doesn't function, then fall back to existing local execution client
```
arnetheduck: do we not do both requests in parallel and pick the more profitable one?

docs/the_nimbus_book/src/external-block-builder.md
```diff
@@ -0,0 +1,21 @@
...
+
+There does exist a failure mode, intrinsic to the builder API, wherein the consensus client has signed a blinded proposal and therefore even if the MEV relay or builder doesn't provide a full block, a consensus client such as Nimbus cannot safely proceed with step 2 and fall back on its local execution client.
+
+## Command line option
+
+=== "Mainnet"
+    ```sh
+    ./run-mainnet-beacon-node.sh --payload-builder=true --payload-builder-url=https://${HOST}:${PORT}/
```
arnetheduck, line 18:
```suggestion
    ./run-mainnet-beacon-node.sh --payload-builder=true --payload-builder-url=https://localhost:4444/
```
do payload builders have a "standard" port? ideally they would have a default, then we can put that default here so that copy-paste just works - elsewhere, we assume the builder is  running on the same host as the node, is that the typical setup for builders too? if yes, might as well make that copy-pasteable too for consistency

docs/the_nimbus_book/src/external-block-builder.md
zah: That's how we do these options in general. For example `--metrics:on --metrics-server-address=0.0.0.0`. One rationale is that it makes it easy for DevOps to control purely the on/off flag, without messing with the URL property that stays a constant.

docs/the_nimbus_book/src/external-block-builder.md
tersec: There are a couple of options:
```go
defaultListenAddr         = getEnv("BOOST_LISTEN_ADDR", "localhost:18550")
```
and the relays in the wild so far as using the standard HTTPS port 443.

If we want something copy-pastable, `http://localhost:18550` probably, but people might not run anything locally on their machines, because unlike the EL, in general, the builder is not running on the same host as the node. There are basically builder-API-proxies which run on the same host, which is as close as one can reasonably get.

docs/the_nimbus_book/src/external-block-builder.md
tersec: My initial version was `--payload-builder=${URL}`, adjusted partly for the reasons @zah noted.

A couple other advantages:
- it's possible that some other way of specifying the payload builders will emerge, and then it'll be useful not to have a privileged interaction between the randomly-first-configuration-option and the enabling/disabling functionality; and
- it avoids a kind of semi-opaque tristate option where it was not possible to describe or specify any kind of simple "default" values for those options, because even the existence of the `--payload-builder=${URL}` option activated a bunch of code, regardless of what `${URL}` was. This is both a documentation issue and also makes diagnosing potential issues trickier, since there was no minimal-difference way to turn on `--payload-builder` but not specify a URL.

docs/the_nimbus_book/src/external-block-builder.md
tersec: In parallel, yes, more profitable one, not yet.

docs/the_nimbus_book/src/external-block-builder.md
tersec: > we should unhide the options too if they're getting documented, they're stable now? Also, do we need both url:s and `--payload-builder=true`? isn't the flag redundant?

https://github.com/status-im/nimbus-eth2/pull/4032/commits/a0b05227c6d19fe9e90cf7af737e768e3401797d unhides the options

Issue #3998, opened 2022-08-19
Cache and resend, rather than recreate, builder API registrations


tersec:
Currently, Nimbus recreates and re-signs all of these every epoch. This is viable for smallish nodes, but does not scale, and isn't necessary. Instead, cache registrations and simply resend the cached and signed registrations every epoch.

https://github.com/status-im/nimbus-eth2/blob/c8fb447020940eeb720e778e79f0ae963137d276/beacon_chain/validators/validator_duties.nim#L1213-L1218

This also is better for the builder/relay API network servers, because they otherwise have to reverify potentially thousands of BLS signatures per node, every few minutes.

tersec:
https://github.com/status-im/nimbus-eth2/pull/4040

PR #4040, opened 2022-08-26
Cache and resend, rather than recreate, builder API registrations

tersec:
https://github.com/status-im/nimbus-eth2/issues/3998

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -1313,6 +1313,10 @@ proc registerValidators(node: BeaconNode, epoch: Epoch) {.async.} =
     # https://github.com/ethereum/builder-specs/blob/v0.2.0/specs/validator.md#validator-registration
     var validatorRegistrations: seq[SignedValidatorRegistrationV1]
     for key in attachedValidatorPubkeys:
+      # Time passed during awaits; REST keymanager API might have removed it
+      if key notin node.attachedValidators[].validators:
+        continue
+
       let validator = node.attachedValidators[].validators[key]
 
       if validator.index.isNone:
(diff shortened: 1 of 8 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -1306,14 +1305,18 @@ proc registerValidators(node: BeaconNode, epoch: Epoch) {.async.} =
...
 
-      let validatorRegistration =
-        await node.getValidatorRegistration(validator, epoch)
-      if validatorRegistration.isErr:
-        error "registerValidators: validatorRegistration failed",
-               validatorRegistration
-        continue
+      if validator.externalBuilderRegistration.isNone:
+        let validatorRegistration =
+          await node.getValidatorRegistration(validator, epoch)
```
arnetheduck, line 1340: if someone adds a validator via rest api, I'm guessing this await can get interleaved (cc @zah?)

tersec: In general, there are lots of the usual kinds approaches people use when modifying sequences in place, but for this purpose, it mostly matters whether table `mvalues` iteration works with that. It's fine if some new validator is missed this epoch around in that scenario, and it's even fine to just abort the table scan in a given epoch if there's any issues of this sort. I'd prefer either, ideally, to adding more internal buffering to this function of information on all attached validators, but that could, ultimately, solve this.

zah: The problem here is that the contents of the `validators` table may change while you are iterating over it. This would be detected as a `Defect` by the standard library.

The standard work-around is to copy the list of validators prior to starting the iteration by iterating over `dup(node.attachedValidators[].validators.values)`.

tersec: https://github.com/status-im/nimbus-eth2/pull/4040/commits/cf539c921e03ef764ce46e2e92187d3fbe12e0ac works around this. Rather than copy the values in their entirety, which would be a better approach for a read-only application, here it copies the keys, because it's more obviously symmetrical that the access mechanism that works for reading also works for writing.

PR #4041, opened 2022-08-27
accelerate EL sync with LC with `--sync-light-client`

etan-status:
When the BN-embedded LC makes sync progress, pass the corresponding
execution block hash to the EL via `engine_forkchoiceUpdatedV1`.
This allows the EL to sync to wall slot while the chain DAG is behind.
Renamed `--light-client` to `--sync-light-client` for clarity, and
`--light-client-trusted-block-root` to `--trusted-block-root` for
consistency with `nimbus_light_client`.

Note that this does not work well in practice at this time:
- Geth sticks to the optimistic sync:
  "Ignoring payload while snap syncing" (when passing the LC head)
  "Forkchoice requested unknown head" (when updating to LC head)
- Nethermind syncs to LC head but does not report ancestors as VALID,
  so the main forward sync is still stuck in optimistic mode:
  "Pre-pivot block, ignored and returned Syncing"

To aid EL client teams in fixing those issues, having this available
as a hidden option is still useful.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -284,16 +284,16 @@ type
         desc: "Weak subjectivity checkpoint in the format block_root:epoch_number"
         name: "weak-subjectivity-checkpoint" .}: Option[Checkpoint]
 
-      lightClientEnable* {.
+      syncLightClient* {.
         hidden
-        desc: "BETA: Accelerate sync using light client."
+        desc: "Accelerate sync using light client"
         defaultValue: false
-        name: "light-client" .}: bool
+        name: "sync-light-client" .}: bool
 
-      lightClientTrustedBlockRoot* {.
+      trustedBlockRoot* {.
         hidden
-        desc: "BETA: Recent trusted finalized block root to initialize light client from."
-        name: "light-client-trusted-block-root" .}: Option[Eth2Digest]
+        desc: "Recent trusted finalized block root to initialize light client from"
+        name: "trusted-block-root" .}: Option[Eth2Digest]
 
       finalizedCheckpointState* {.
         desc: "SSZ file specifying a recent finalized state"
--- a/beacon_chain/consensus_object_pools/consensus_manager.nim
+++ b/beacon_chain/consensus_object_pools/consensus_manager.nim
@@ -56,6 +56,7 @@ type
     # Tracking last proposal forkchoiceUpdated payload information
     # ----------------------------------------------------------------
     forkchoiceUpdatedInfo*: Opt[ForkchoiceUpdatedInformation]
+    optimisticHead: tuple[bid: BlockId, execution_block_hash: Eth2Digest]
 
 # Initialization
 # ------------------------------------------------------------------------------
@@ -117,6 +118,45 @@ from web3/engine_api_types import
 
 func `$`(h: BlockHash): string = $h.asEth2Digest
 
+func shouldSyncOptimistically*(
+    optimisticSlot, dagSlot, wallSlot: Slot): bool =
+  ## Determine whether an optimistic execution block hash should be reported
+  ## to the EL client instead of the current head as determined by fork choice.
+
+  # Check whether optimistic head is sufficiently ahead of DAG
+  const minProgress = 8 * SLOTS_PER_EPOCH  # Set arbitrarily
+  if dagSlot + minProgress > optimisticSlot:
+    return false
+
+  # Check whether optimistic head has synced sufficiently close to wall slot
+  const maxAge = 2 * SLOTS_PER_EPOCH  # Set arbitrarily
+  if optimisticSlot < max(wallSlot, maxAge.Slot) - maxAge:
+    return false
+
+  true
+
+func shouldSyncOptimistically*(self: ConsensusManager, wallSlot: Slot): bool =
+  if self.eth1Monitor == nil:
+    return false
+  if self.optimisticHead.execution_block_hash.isZero:
+    return false
+
+  shouldSyncOptimistically(
+    optimisticSlot = self.optimisticHead.bid.slot,
+    dagSlot = getStateField(self.dag.headState, slot),
+    wallSlot = wallSlot)
+
+func optimisticHead*(self: ConsensusManager): BlockId =
+  self.optimisticHead.bid
+
+func optimisticExecutionPayloadHash*(self: ConsensusManager): Eth2Digest =
+  self.optimisticHead.execution_block_hash
+
+func setOptimisticHead*(
+    self: var ConsensusManager,
+    bid: BlockId, execution_block_hash: Eth2Digest) =
+  self.optimisticHead = (bid: bid, execution_block_hash: execution_block_hash)
+
 proc runForkchoiceUpdated*(
     eth1Monitor: Eth1Monitor,
     headBlockRoot, safeBlockRoot, finalizedBlockRoot: Eth2Digest):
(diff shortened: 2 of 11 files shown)


beacon_chain/conf.nim
```diff
@@ -284,16 +284,16 @@ type
         desc: "Weak subjectivity checkpoint in the format block_root:epoch_number"
         name: "weak-subjectivity-checkpoint" .}: Option[Checkpoint]
 
-      lightClientEnable* {.
+      syncLightClient* {.
         hidden
-        desc: "BETA: Accelerate sync using light client."
+        desc: "Accelerate sync using light client"
```
tersec, line 289: Given that this doesn't work currently, is reasonable here and for `--trusted-block-root` to remove the `BETA: ` designation?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1285,13 +1285,21 @@ proc onSlotEnd(node: BeaconNode, slot: Slot) {.async.} =
...
+func syncStatus(node: BeaconNode, wallSlot: Slot): string =
   let optimistic_head = node.dag.is_optimistic(node.dag.head.root)
   if node.syncManager.inProgress:
-    if optimistic_head:
-      node.syncManager.syncStatus & "/opt"
-    else:
-      node.syncManager.syncStatus
+    let
+      optimisticSuffix =
+        if optimistic_head:
```
tersec, line 1293: I deliberately didn't factor this out because it introduces an additional memory allocation/GC churn and string concatenation in some cases, but probably it's reasonable enough, so, sure.

docs/the_nimbus_book/src/options.md
```diff
@@ -35,6 +35,8 @@ The following options are available:
      --secrets-dir             A directory containing validator keystore passwords.
      --wallets-dir             A directory containing wallet files.
      --web3-url                One or more execution layer Web3 provider URLs.
+     --require-engine-api-in-bellatrix  Require Nimbus to be configured with an Engine API end-point after the Bellatrix
```
tersec, line 38: I'm not sure what the right level of documentation to have for this is.

In an ideal world, it goes away in a couple of months, because the merge has happened.

For it to be useful, it can't be that hidden, so, it's in `--help`, for example, but any additional visibility keeps it retained longer, probably.

I'm envisioning treating this as something people can discover if they truly need it, not something so easy to find that people start tweaking it regardless, and, it's really only useful anyway for about a week or two, then it's obsolete.

Not sure it should be added here.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -117,6 +118,45 @@ from web3/engine_api_types import
...
 func `$`(h: BlockHash): string = $h.asEth2Digest
 
+func shouldSyncOptimistically*(
+    optimisticSlot, dagSlot, wallSlot: Slot): bool =
+  ## Determine whether an optimistic execution block hash should be reported
+  ## to the EL client instead of the current head as determined by fork choice.
+
+  # Check whether optimistic head is sufficiently ahead of DAG
+  const minProgress = 8 * SLOTS_PER_EPOCH  # Set arbitrarily
+  if dagSlot + minProgress > optimisticSlot:
```
tersec: Maybe document, at least, in a comment the boundary conditions here. These kind of issues came up in the audit, and while I'm not convinced that caring about anything happening in a trillion years is worthwhile if it makes the code/logic more complex, it's worth acknowledging a bit, e.g.,
```
# 6 second (minimal preset) slots => overflow at ~1.75 trillion years
# Overflows in 14 trillion years (minimal) or 112 trillion years (mainnet)
```
or whatever the appropriate numbers are here.

beacon_chain/conf.nim
etan-status: It's still `hidden` (and the reason why it doesn't work well is not Nimbus).
We are only using the "BETA" tag for non-hidden options elsewhere.
But no strong pref on this, I can re-add it as well.

docs/the_nimbus_book/src/options.md
etan-status: This excerpt is literally a copy of `--help` that was forgotten to be updated in the .md. As in, `build/nimbus_beacon_node`, copy the first block into the .md, done.

beacon_chain/nimbus_beacon_node.nim
etan-status: It's only called in "Slot start" log and as part of the status bar, so not a frequent path.

An alternative would be:
```
    let lc_head = node.consensusManager[].shouldSyncOptimistically(wallSlot)
    if optimistic_head or lc_head:
      var res = newStringOfCap(100)
      res.add node.syncManager.syncStatus
      if optimistic_head:
        res.add "/opt"
      if lc_head:
        res.add " - lc: "
        res.add $shortLog(node.consensusManager[].optimisticHead)
      res
    else:
      node.syncManager.syncStatus
```

but it loses a lot on readability.

etan-status: Not sure. It needs to be fast enough even in the worst path of both suffixes, I think it's fine.

beacon_chain/consensus_object_pools/consensus_manager.nim
etan-status: rewrote the condition to make it clearer to reason about.

beacon_chain/conf.nim
tersec: Probably, sure, the feature itself can be viewed as non exactly "beta", but not supported by the ecosystem at large.

beacon_chain/nimbus_beacon_node.nim
tersec: > It's only called in "Slot start" log and as part of the status bar, so not a frequent path.
> 
> An alternative would be:
> 
> ```
>     let lc_head = node.consensusManager[].shouldSyncOptimistically(wallSlot)
>     if optimistic_head or lc_head:
>       var res = newStringOfCap(100)
>       res.add node.syncManager.syncStatus
>       if optimistic_head:
>         res.add "/opt"
>       if lc_head:
>         res.add " - lc: "
>         res.add $shortLog(node.consensusManager[].optimisticHead)
>       res
>     else:
>       node.syncManager.syncStatus
> ```
> 
> but it loses a lot on readability.

I certainly prefer what you wrote to this, yes.

tersec (approved):

beacon_chain/nimbus_beacon_node.nim
etan-status: The string concatenation actually seems quite efficient, given that Nim stores the length it does not require a linear scan (essentially a memcpy), and can have similar performance to a string builder in other languages.

The allocations are real, though, there is no extra capacity allocated after a string by default. Not sure how much of it gets optimized away at C compilation step, I doubt it's a lot though as the generated logic is a bit.. bulky.

Still, I think here the frequency is 1 / second, so definitely not the bottleneck for now ^^

beacon_chain/conf.nim
etan-status: That describes it quite accurately

beacon_chain/nimbus_beacon_node.nim
tersec: Alright, this can be optimized if (unlikely) it ever does become a bottleneck.

PR #4042, opened 2022-08-27
add REST endpoint for fork choice context

etan-status:
Implements a proposed REST endpoint for analyzing fork choice behaviour.
See https://github.com/ethereum/beacon-APIs/pull/232

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -484,6 +484,12 @@ type
         defaultValue: true # the use of the nimbus_signing_process binary by default will be delayed until async I/O over stdin/stdout is developed for the child process.
         name: "in-process-validators" .}: bool
 
+      debugForkChoice* {.
+        hidden
+        desc: "Enable debug API for fork choice (https://github.com/ethereum/beacon-APIs/pull/232)"
+        defaultValue: false
+        name: "debug-fork-choice" .}: bool
+
       discv5Enabled* {.
         desc: "Enable Discovery v5"
         defaultValue: true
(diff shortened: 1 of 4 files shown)


tersec:
Aside from the code quality [I say this, not having even looked yet], I'd prefer not adding endpoints which need indefinite support unless there's either a particular compelling reason or they get actually added to a spec (which counts, obviously, as a compelling reason).

etan-status:
Scripts:

Dump fork choice periodically (cd to a directory first):
`
while :; do mkdir -p "$(date -u +"%Y-%m-%d_%H")" && curl -s "http://localhost:5052/eth/v1/debug/fork_choice" -o "$(date -u +"%Y-%m-%d_%H/%Y-%m-%dT%H:%M:%SZ").txt"; sleep 0.75; done
`

Combine multiple files into one:
`jq -n '[inputs | .time=(input_filename | split("/")[-1] | split(".")[0]) | .protoArray=.data | del(.data)]' *.txt`

etan-status:
> indefinite support

If it's not part of the spec, there should be no need for indefinite support, no? or is it about /v1 vs /v0?

tersec:
> > indefinite support
> 
> If it's not part of the spec, there should be no need for indefinite support, no? or is it about /v1 vs /v0?

That's a concern, yes: it's exactly there is a spec thing that'll be merged, and if this gets merged too, before the spec version, and they diverge, there'll be pressure to support both in an ongoing way. That it's part of the `debug` API helps, and it's definitely useful functionality, so as long as it can remain aligned with a likely-looking of piece of specification to be merged without creating redundancy, it seems fine.

If it means Nimbus will be stuck with a microcosm of the REST vs RPC server thing for a while, where there's this Nimbus-specific code that just can't be gotten rid for a while of even though it's been supplanted, that's less fine.

etan-status:
I've put the API behind a startup flag now, so that it is possible to fetch this debug data when needed, without committing to "supporting it forever"; it's clearly meant as a developer tool.

The handler itself only imports readonly functions of proto array, so it cannot mess with fork choice by accident.

tersec (approved):

PR #4045, opened 2022-08-29
refactor `newPayload` timeout/retry handling

tersec:
This addresses:
- it should, per spec, be retrying calls such as `newPayload` when they're crucial to forward progress. `newPayload` occurs in two timing contexts, the block processing queue which doesn't constrain the timing of the rest of Nimbus, and during block proposal, where to the extent it's required, it also constrains forward progress
- it centralizes timeout handling at the same time, because all calls to `newPayload` must respect its 8 second timeout, and simplifies calling code as a result.
- if `newPayload` isn't callable, then that's not equivalent to `SYNCING`, because `SYNCING` allows optimistic import, whereas an inability to call the EL does not, so signal that separately. All `newPayload` callers already handle `CatchableError`, so use that with an informative error message.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -485,15 +485,22 @@ proc getPayload*(p: Eth1Monitor,
   p.dataProvider.web3.provider.engine_getPayloadV1(FixedBytes[8] payloadId)
 
 proc newPayload*(p: Eth1Monitor, payload: engine_api.ExecutionPayloadV1):
-    Future[PayloadStatusV1] =
+    Future[PayloadStatusV1] {.async.} =
   # Eth1 monitor can recycle connections without (external) warning; at least,
   # don't crash.
   if p.dataProvider.isNil:
-    let epr = newFuture[PayloadStatusV1]("newPayload")
-    epr.complete(PayloadStatusV1(status: PayloadExecutionStatus.syncing))
-    return epr
+    raise newException(CatchableError, "Web3 provider is nil")
 
-  p.dataProvider.web3.provider.engine_newPayloadV1(payload)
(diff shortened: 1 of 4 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -485,15 +485,22 @@ proc getPayload*(p: Eth1Monitor,
...
+    raise newException(CatchableError, "Web3 provider is nil")
+
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#request
+  const NEWPAYLOAD_TIMEOUT = 8.seconds
+
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#timeouts
+  # Consensus Layer client software **MUST** wait for a specified `timeout`
+  # before aborting the call. In such an event, the Consensus Layer client
+  # software **SHOULD** retry the call when it is needed to keep progressing.
+  return awaitWithRetries(
```
zah, line 501: Please note that this can still fail, but now more than 30 seconds later. Is this reasonable from the perspective of the code calling this function?

cheatfate: I think `awaitWithRetries` is incorrect in timeout handling. It should use its timeout value for all requests, but right now it is using new timeout value for every attempt. One more thing, this template is unable to handle cancellation properly and exploits usage of `yield` keyword.

tersec: For in the block queue processing loop, definitely, yes, it's better to retry, in order, even if takes more than 30 seconds.

For the `validator_duties` version, it's not as clear, because it could cause a long wait. Probably yes if it needs to propose next slot too (because then it's important that the EL client see the payload in time for the fcU setting up the next slot) and, more debatably otherwise, if the next slot isn't another proposal.

tersec:
Obsoleted by other PRs.

PR #4050, opened 2022-08-30
descore on empty response for range w known block

etan-status:
The sync protocol does not distinguish between:
- All requested slots are empty
- Peer does not have data available about requested range

Therefore, we treat EOF for `beacon_blocks_by_range` and for
`beacon_blocks_by_range` as valid responses, as if the entire epoch
really contained no single block for any slot. Once a followup response
provides new blocks, we detect that some blocks were missing and rewind.

During backfill, we also request the known-to-exist `backfill.slot`,
so we can actually detect whether an epoch really does not have blocks
or whether a response is incomplete (`PeerScoreNoBlocks`).

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -348,6 +348,20 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
              request = req
         return
 
+      if len(data) == 0 and man.direction == SyncQueueKind.Backward and
+          req.contains(man.getSafeSlot()):
+        # The sync protocol does not distinguish between:
+        # - All requested slots are empty
+        # - Peer does not have data available about requested range
+        #
+        # However, we include the `backfill` slot in backward sync requests.
+        # If we receive an empty response to a request covering that slot,
+        # we know that the response is incomplete and can descore.
+        peer.updateScore(PeerScoreNoBlocks)
+        man.queue.push(req)
+        debug "Response does not include known-to-exist block", request = req
+        return
+
       # Scoring will happen in `syncUpdate`.
       man.workers[index].status = SyncWorkerStatus.Queueing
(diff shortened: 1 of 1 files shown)


etan-status:
Similar to this, we also don't distinguish between unexpected disconnects and graceful disconnects:
- https://github.com/status-im/nim-libp2p/issues/755

etan-status:
Seeing backfill performance of continuously increasing slots/s on Goerli, around 50/s. I ran it for multiple days before and did not make meaningful backfill sync progress.

etan-status:
Successful backfill on a manual Goerli checkpoint sync.

beacon_chain/sync/sync_manager.nim
```diff
@@ -348,6 +348,20 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
              request = req
         return
 
+      if len(data) == 0 and man.direction == SyncQueueKind.Backward and
+          req.contains(man.getSafeSlot()):
```
arnetheduck, line 352: so we request the safe slot redundantly?

etan-status: Yes, that was always the case.

etan-status: But only for `Backward` sync

arnetheduck: talking to @cheatfate, we realised that we should do the same for peers the report a head block via `status` update and then fail to provide it (this indicates they're buggy and should get a significant penalty)

arnetheduck (approved):

beacon_chain/sync/sync_manager.nim
etan-status: Yeah, for the head block it works, but sync manager only syncs until 1 epoch behind wall slot. it could be a scoring for request manager though.

PR #4055, opened 2022-08-31
more efficient forkchoiceUpdated usage

tersec:


beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -47,6 +48,10 @@ type
     # ----------------------------------------------------------------
     eth1Monitor*: Eth1Monitor
 
+    # Allow determination of whether there's an upcoming proposal
+    # ----------------------------------------------------------------
+    attachedValidators*: ref ValidatorPool
```
arnetheduck: we need to consider proposers that come in via rest api as well

etan-status: `ActionTracker` could be an alternative that takes care of VCs, no?

```
func getNextProposalSlot*(tracker: ActionTracker, slot: Slot): Slot
```

tersec: https://github.com/status-im/nimbus-eth2/pull/4055/commits/2809f413085f0e489d4d9c1655071affb64bd908 uses `getNextProposalSlot`

arnetheduck: looking over the code, in addition to the action tracker we should also be checking https://github.com/status-im/nimbus-eth2/blob/7cac6f46db950ed7bdf758f82d3533526a1c1f4f/beacon_chain/spec/eth2_apis/dynamic_fee_recipients.nim which collects calls from https://ethereum.github.io/beacon-APIs/#/Validator/prepareBeaconProposer which is the offiical way of preparing - actiontracker derives "active" proposers from attestations which is a good enough proxy in many cases but not all

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -328,15 +331,22 @@ proc storeBlock*(
...
+          # No attached validator is next proposer, so use non-proposal fcU
+          asyncSpawn eth1Monitor.expectValidForkchoiceUpdated(
+            headBlockRoot = headExecutionPayloadHash,
+            safeBlockRoot = newHead.get.safeExecutionPayloadHash,
+            finalizedBlockRoot = newHead.get.finalizedExecutionPayloadHash)
+        else:
+          # Some attached validator is next proposer, so prepare payload. As
+          # updateHead() updated the DAG head, runProposalForkchoiceUpdated,
+          # which needs the state corresponding to that head block, can run.
+          asyncSpawn self.consensusManager.runProposalForkchoiceUpdated(
```
arnetheduck, line 345: is the spawn here guaranteed to complete before the next fcu call?

tersec: As far as Nimbus is concerned, yes
https://github.com/status-im/nimbus-eth2/blob/e367f612e89c81eb0dbde7531363cce279b83362/beacon_chain/consensus_object_pools/consensus_manager.nim#L304-L319

`FORKCHOICEUPDATED_TIMEOUT` is 8 seconds.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4055/commits/0782a71a82c69f71018ecec8faef76d57609dc3a

Has risk of false positives, but if people prefer that, sure.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -343,9 +367,14 @@ proc updateHeadWithExecution*(self: ref ConsensusManager, newHead: BeaconHead)
     # justified and finalized
     self.dag.updateHead(newHead.blck, self.quarantine[])
 
-    # TODO after things stabilize with this, check for upcoming proposal and
-    # don't bother sending first fcU, but initially, keep both in place
-    asyncSpawn self.runProposalForkchoiceUpdated()
+    if self.checkNextProposer(wallSlot).isSome:
```
arnetheduck: this is called inside rpfu as well?

tersec: https://github.com/status-im/nimbus-eth2/pull/4055/commits/10e8b3879c6bbd9877bbfcd4b09d8ededae73524

arnetheduck (approved):

Issue #3954, opened 2022-08-11
Descendants of `INVALID` block are not deleted in optsync


etan-status:
**Describe the bug**
On `testing-large-01.aws-eu-central-1a.nimbus.prater`, the Nimbus beacon node entered a restart loop and crashes on start:
```
{"lvl":"INF","ts":"2022-08-11 10:14:13.534+00:00","msg":"Loading block DAG from database","topics":"beacnde","path":"/data/beacon-node-prater-testing/data/db"}
/data/beacon-node-prater-testing/repo/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(456) main
/data/beacon-node-prater-testing/repo/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(449) NimMain
/data/beacon-node-prater-testing/repo/beacon_chain/nimbus_beacon_node.nim(2177) main
/data/beacon-node-prater-testing/repo/beacon_chain/nimbus_beacon_node.nim(2045) handleStartUpCmd
/data/beacon-node-prater-testing/repo/beacon_chain/nimbus_beacon_node.nim(1863) doRunBeaconNode
/data/beacon-node-prater-testing/repo/beacon_chain/nimbus_beacon_node.nim(646) init
/data/beacon-node-prater-testing/repo/beacon_chain/nimbus_beacon_node.nim(191) loadChainDag
/data/beacon-node-prater-testing/repo/beacon_chain/consensus_object_pools/blockchain_dag.nim(836) init
/data/beacon-node-prater-testing/repo/vendor/nim-stew/stew/results.nim(756) expect
/data/beacon-node-prater-testing/repo/vendor/nim-stew/stew/results.nim(348) raiseResultDefect
Error: unhandled exception: not nil [ResultDefect]
```

The problem occurs because in the DB, when loading blocks from `head` => `finalizedHead`, one block is missing from the DB.
Examining the logs reveals that the missing block actually got judged as `INVALID` by the EL. Whether this is due to the config issue is a separate problem.
```
{"lvl":"DBG","ts":"2022-08-11 09:42:07.883+00:00","msg":"newPayload: succeeded","parentHash":"893a5a27","blockHash":"1642d9bc","blockNumber":7384505,"payloadStatus":2}
[...]

PR #4058, opened 2022-09-01
clean up more properly after fcU-`INVALID`

tersec:
https://github.com/status-im/nimbus-eth2/issues/3954

Ends up conflicting too much with https://github.com/status-im/nimbus-eth2/pull/4055 so will pause and continue/rebase after that's merged.

Basic approach is visible, though -- what remains is (a) some tests and (b) the plumbing around time functionality to enable proper forkchoice re-updating after an `INVALID`.


tersec:
Superceded by https://github.com/status-im/nimbus-eth2/pull/4081

PR #4071, opened 2022-09-06
fork choice: support marking roots/nodes invalid

tersec:


beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -435,6 +435,17 @@ func prune*(
 func prune*(self: var ForkChoice): FcResult[void] =
   self.backend.prune(self.checkpoints.finalized.root)
 
+func mark_root_invalid*(self: var ForkChoice, root: Eth2Digest) =
```
etan-status, line 438: Will the caller mark descendants as well? Otherwise may have to propagate to them in here.

beacon_chain/fork_choice/proto_array.nim
```diff
@@ -550,6 +552,8 @@ func nodeIsViableForHead(self: ProtoArray, node: ProtoNode): bool =
   ) and (
     (node.checkpoints.finalized == self.checkpoints.finalized) or
     (self.checkpoints.finalized.epoch == GENESIS_EPOCH)
+  ) and (
+    not node.invalid
```
etan-status: moving this to the start of the function would ensure it applies on all networks.

```nim
if node.invalid:
  return false
```

the gh suggest feature doesn't work so far away.

beacon_chain/fork_choice/fork_choice.nim
tersec: `maybeUpdateBestChildAndDescendant` and `nodeLeadsToViableHead` handle the propagation to descendents:
https://github.com/status-im/nimbus-eth2/blob/3a8abd60108ff94b340620e2b12fd2cb7229ffbe/beacon_chain/fork_choice/proto_array.nim#L517-L530

beacon_chain/fork_choice/proto_array.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4071/commits/25a4466b2c42eee30b4dfc4150c70d7dbe60d5c0

arnetheduck:
why do we mark things invalid in fork choice? why not remove them / move them outside?

tersec:
> why do we mark things invalid in fork choice? why not remove them / move them outside?

Risk tolerance of doing this now. This is a low-risk approach. There is `prune` functionality in fork choice, but it prunes old blocks and does some rearranging of the protoarray that is too high-risk at this point to use similar logic with. Furthermore, this function will remove old protoarray nodes, including all the invalid ones, next finalization.

As for why not do this purely in the DAG, issues with that:

(1) the DAG only knows about parents as it is, and without consuming more memory and more complicated/novel accounting mechanisms, there's no ready way to query for descendents to remove when marking something in the DAG invalid, which means that the general case is inefficient if it's dependent on clearing the DAG but leaving fork choice intact;

(2) if fork choice isn't adjusted too, the risk is that the fork choice head will keep pointing to some block that the DAG has removed, and this is recoverable via node restart, but only that;

(3) the way the DAG updates are handled in https://github.com/status-im/nimbus-eth2/pull/4058 (which this is a spinoff of) is to just look at the head and go backwards, trading off scanning the whole DAG against leaving some blocks (albeit in sidechains) in the database in cases of invalids. Those can be cleaned up after next finalization, for example, or via some other garbage collection mechanism which can run without as much time pressure. But this means that it's important to inform fork choice that no other side-chain deriving from an invalid block can be selected instead, too.

beacon_chain/fork_choice/proto_array.nim
```diff
@@ -593,6 +600,7 @@ iterator items*(self: ProtoArray): ProtoArrayItem =
       checkpoints: node.checkpoints,
       unrealized: unrealized,
       weight: node.weight,
+      invalid: node.invalid,
```
etan-status, line 603: wondering if invalid items should be skipped in `rest_debug_api.nim`, but for now just getting the fix in has priority.

tersec: Considered this, deliberately left them in but including information to see their status. The debug API can be used to, well, debug fork choice and as long as they're in the protoarray probably better not to skip them.

etan-status: For the `invalid` flag to be transmitted as part of the debug API, would have to add it to `ForkChoiceResponseExtraData`

etan-status (approved):

beacon_chain/fork_choice/proto_array.nim
tersec: Will do this, in a follow-up PR.

tersec: https://github.com/status-im/nimbus-eth2/pull/4074

PR #4072, opened 2022-09-06
only mark post-finalized blocks invalid

tersec:
It doesn't make sense, in general, to treat finalized blocks as mutable. If the consensus layer chain finalized blocks later declared invalid, well, it still finalized them. It becomes a social consensus problem and requires manual intervention.


etan-status (approved):
`dag.finalizedHead` may be unrealized on the current particular fork. But point still holds that there exists a fork that finalized through `dag.finalizedHead.slot`.

PR #4075, opened 2022-09-06
checking for merge terminal block should be debug-level

tersec:


etan-status:
Or incorporated into `Slot start` / `Slot end`?

etan-status (approved):

tersec:
> Or incorporated into `Slot start` / `Slot end`?

It really is kind of a debugging thing. End-users shouldn't have to care.

PR #4076, opened 2022-09-06
Fix messageId generation for forks following Altair

Menduist:


etan-status (approved):
Validated that `topic.startsWith(altairPrefix) or topic.startsWith(bellatrixPrefix)` at the `everything >= altair`. Probably fine without an assert, because if there is a new prefix scheme introduced in the future this portion is probably re-visited as part of potential work for that.

etan-status:
Actually, are messageId only put to GossipSub, or also to req/resp? If also applicable to req/rsp, maybe `/eth2/beacon_chain/req/` would be another prefix to ignore.

it _looks_ like it is only applied to GossipSub, so assuming it's fine.

Menduist:
Only to GossipSub topics AFAIK, req/resp use regular semver instead

PR #4077, opened 2022-09-06
exchangeTransitionConfiguration fix

tersec:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1376,8 +1361,9 @@ proc onSecond(node: BeaconNode, time: Moment) =
   ## This procedure will be called once per minute.
   # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1
   if time > node.nextExchangeTransitionConfTime and not node.eth1Monitor.isNil:
-    node.nextExchangeTransitionConfTime = time + chronos.minutes(1)
-    traceAsyncErrors node.eth1Monitor.exchangeTransitionConfiguration()
+    node.nextExchangeTransitionConfTime = time + chronos.seconds(45)
```
etan-status, line 1367:
```nim
  if time >= node.nextExchangeTransitionConfTime and not node.eth1Monitor.isNil:
    # The EL client SHOULD log a warning when not receiving an exchange message
    # at least once every 120 seconds. If we only attempt to exchange every 60
    # seconds, the warning would be triggered if a single message is missed.
    # To accommodate for that, exchange slightly more frequently.
    # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1
    node.nextExchangeTransitionConfTime = time + chronos.seconds(45)
    
    if node.currentSlot.epoch >= node.dag.cfg.BELLATRIX_FORK_EPOCH:
      traceAsyncErrors node.eth1Monitor.exchangeTransitionConfiguration()
```

etan-status: - \>= so that the exchange runs right at the start
- rationale for 45 seconds

etan-status (approved):

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4077/commits/f50916af5fa5aa57181e8e0de0e68500b83d8722

etan-status (approved):

Issue #3699, opened 2022-06-03
Improved logging/docs for validator flow


lightclient:
I was recently gifted a stack of Görli eth to stake on Prater and thought this would be a good time to run a serious staking setup with nimbus! While configuring the setup, I ran into a few problems (all self-imposed) that I feel could be better illuminated through logging. These are kind of from memory so please forgive me if I get something confused.

*  Incrementally output validator pubkeys as their keystores are unencrypted. Lighthouse currently does this. For only a handful of validators it's probably fine to do a summary, but for a larger number it's sort of weird for it to appear to block for a long period of time (even with debug log level set).
* `NOT` is _not_ a terribly reassuring log prefix msg to be seeing while importing validator keys (`nimbus deposits import`).
* Both `--validators-dir` and `--secrets-dir` state the argument is "a directory containing X", but I believe the actual expected value is a directory containing another directory, which contains X.
* Incorrect permissions on `validators` or `secrets` causes nimbus to simply ignore them. IMO if those directories exist, but have invalid permissions that should at least be logged as a `WARN` if not outright `FATAL`.
* `--insecure` is a kind of a concerning flag name to pass to a piece of software I'm about to delegate tens of thousands of (hypothetical) euros to (yes I see if you don't want metrics you don't need it, but still).
* Overall, I find the docs sort of difficult to follow because it's very focused on these scripts in the root of the repo. I think it would be great to actually remove these scripts to run nimbus and put things that should go into the client in the client and things that don't make sense into the docs. A good example of this is in the "[Import your validator keys](https://nimbus.guide/keys.html#import-your-validator-keys-into-nimbus)"; it asks to copy the `validator_keys` directory into the nimbus repository and then run a command (`nimbus deposits import`) whose path defaults to the current directory. Would much rather it be agnostic to where the user has placed their files.
* The "[Joe Clapis](https://nimbus.guide/metrics-pretty-pictures.html#joe-clapis)" dashboard is now extremely rocketpool focused - not sure if it makes sense as the first community dashboard to point people to anymore.

arnetheduck:
> --insecure is a kind of a concerning flag name to pass to a piece of software I'm about to delegate tens of thousands of (hypothetical) euros to (yes I see if you don't want metrics you don't need it, but still).

The `insecure` options was used in very early versions of Nimbus to enable metrics support - metrics have since been added to the stable feature set and no longer require the insecure flag (for a good while too) - it looks like there's a guiide somewhere that needs updating?

[...]

PR #4078, opened 2022-09-06
 load suggested fee recipient file also when keymanager is disabled 

arnetheduck:
Since these files may have been created in a previous run or manually,
we want to keep loading them even on nodes that don't enable the
keystore API (for example static setups)

* log keystore loading progressively (https://github.com/status-im/nimbus-eth2/issues/3699)


zah:
LGTM, but needs rebasing

arnetheduck:
rebased, and added logging of the fee recipient at different stages

PR #4081, opened 2022-09-06
handle INVALIDATED forkchoiceUpdated better

tersec:
https://github.com/status-im/nimbus-eth2/issues/3954

### DIFF
--- a/beacon_chain/consensus_object_pools/consensus_manager.nim
+++ b/beacon_chain/consensus_object_pools/consensus_manager.nim
@@ -211,17 +211,21 @@ proc runForkchoiceUpdatedDiscardResult*(
   discard await eth1Monitor.runForkchoiceUpdated(
     headBlockRoot, safeBlockRoot, finalizedBlockRoot)
 
-proc updateExecutionClientHead(self: ref ConsensusManager, newHead: BeaconHead)
-    {.async.} =
+from ../beacon_clock import GetBeaconTimeFn
+from ../fork_choice/fork_choice import mark_root_invalid
+
+proc updateExecutionClientHead(
+    self: ref ConsensusManager, newHead: BeaconHead):
+    Future[Opt[void]] {.async.} =
   if self.eth1Monitor.isNil:
-    return
+    return Opt[void].ok()
 
   let headExecutionPayloadHash = self.dag.loadExecutionBlockRoot(newHead.blck)
 
   if headExecutionPayloadHash.isZero:
     # Blocks without execution payloads can't be optimistic.
     self.dag.markBlockVerified(self.quarantine[], newHead.blck.root)
-    return
+    return Opt[void].ok()
 
   # Can't use dag.head here because it hasn't been updated yet
   let payloadExecutionStatus = await self.eth1Monitor.runForkchoiceUpdated(
@@ -235,8 +239,11 @@ proc updateExecutionClientHead(self: ref ConsensusManager, newHead: BeaconHead)
   of PayloadExecutionStatus.invalid, PayloadExecutionStatus.invalid_block_hash:
     self.dag.markBlockInvalid(newHead.blck.root)
     self.quarantine[].addUnviable(newHead.blck.root)
+    return Opt.none(void)
   of PayloadExecutionStatus.accepted, PayloadExecutionStatus.syncing:
     self.dag.optimisticRoots.incl newHead.blck.root
+
+  return Opt[void].ok()
 
 proc updateHead*(self: var ConsensusManager, newHead: BlockRef) =
   ## Trigger fork choice and update the DAG with the new head block
@@ -352,8 +359,8 @@ proc runProposalForkchoiceUpdated*(
     error "Engine API fork-choice update failed", err = err.msg
 
 proc updateHeadWithExecution*(
-    self: ref ConsensusManager, newHead: BeaconHead, wallSlot: Slot)
-    {.async.} =
+    self: ref ConsensusManager, initialNewHead: BeaconHead,
+    getBeaconTimeFn: GetBeaconTimeFn) {.async.} =
   ## Trigger fork choice and update the DAG with the new head block
   ## This does not automatically prune the DAG after finalization
   ## `pruneFinalized` must be called for pruning.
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -228,9 +232,20 @@ proc updateExecutionClientHead(self: ref ConsensusManager, newHead: BeaconHead)
   of PayloadExecutionStatus.invalid, PayloadExecutionStatus.invalid_block_hash:
     self.dag.markBlockInvalid(newHead.blck.root)
     self.quarantine[].addUnviable(newHead.blck.root)
+
+    # Need to select a new head, because the nominal head isn't valid anymore
+    self.attestationPool[].forkChoice.mark_root_invalid(newHead.blck.root)
```
etan-status: Is this also needed in `block_processor` for `newPayload` failures? i.e., should this be moved into `markBlockInvalid`?

Could fcu return SYNCING forever once `newPayload` reports `INVALID`, as the EL already dropped the invalid block?

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -328,16 +343,34 @@ proc runProposalForkchoiceUpdated*(self: ref ConsensusManager) {.async.} =
...
   ## This does not automatically prune the DAG after finalization
   ## `pruneFinalized` must be called for pruning.
 
   # Grab the new head according to our latest attestation data
   try:
     # Ensure dag.updateHead has most current information
-    await self.updateExecutionClientHead(newHead)
+    var newHead = initialNewHead
+    var attempts = 0
+    const maxAttempts = 3
```
etan-status: hmm, technically this should be retried until it works, but considering that multiple of these `updateHeadWithExecution` may run in parallel and that this is called every time a new block is processed, the 3 attempts _should_ be fine as a change that is not too invasive.

Eventually, having a background task that always ensures that the EL matches DAG, that re-selects new head until a `VALID` one is found (and periodically re-issues fcU as long as it's `SYNCING`), and that manages the concurrency / memory use more predictably (e.g., 10 fcu at a time), may make this clearer.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -228,9 +232,20 @@ proc updateExecutionClientHead(self: ref ConsensusManager, newHead: BeaconHead)
   of PayloadExecutionStatus.invalid, PayloadExecutionStatus.invalid_block_hash:
     self.dag.markBlockInvalid(newHead.blck.root)
     self.quarantine[].addUnviable(newHead.blck.root)
+
+    # Need to select a new head, because the nominal head isn't valid anymore
+    self.attestationPool[].forkChoice.mark_root_invalid(newHead.blck.root)
+    let wallTime = getBeaconTimeFn()
+    return self.attestationPool[].selectOptimisticHead(wallTime).valueOr:
```
etan-status: Note that returning from an `async` function inserts a yield point, so the head selection will NOT run as part of the same chunk as the caller's `await self.updateExecutionClientHead(newHead, getBeaconTimeFn)`.

Alternatively, could have an `Opt[void]` result here, and do the `selectOptimisticHead` in the caller to guarantee that `self.updateExecutionClientHead` really gets the most recent value. It would also separate the logic better, by having `updateExecutionClientHead` _only_ worry about updating EL to certain head and marking valid/invalid, and the calling function deal with head selection and coordination of recovery logic. 

The `Opt[void]` could make the recovery logic read like the following. I think it reflects the intention a bit better than interpreting the result of "newHead != nextHead" as "requested head was invalid".
```nim
var 
  attempts = 0
  newHead = initialNewHead
while (await self.updateExecutionClientHead(newHead)).isErr:
  # This proc is called on every new block; guarantee timely return
  inc attempts
  const maxAttempts = 3  
  if attempts >= maxAttempts:
    warn "updateHeadWithExecution: too many attempts to recover from invalid payload",
      attempts, maxAttempts, newHead, initialNewHead
    break
  
  # Select new head for next attempt
  let 
    wallTime = getBeaconTimeFn()
    nextHead = self.attestationPool[].selectOptimisticHead(wallTime).valueOr:
      warn "Head selection failed after invalid block, using previous head",
        newHead, wallSlot = wallTime.slotOrZero
      break
  warn "updateHeadWithExecution: attempting to recover from invalid payload",
    attempts, maxAttempts, newHead, initialNewHead, nextHead
  newHead = nextHead
```

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -328,16 +343,34 @@ proc runProposalForkchoiceUpdated*(self: ref ConsensusManager) {.async.} =
     except CatchableError as err:
       error "Engine API fork-choice update failed", err = err.msg
 
-proc updateHeadWithExecution*(self: ref ConsensusManager, newHead: BeaconHead)
-    {.async.} =
+proc updateHeadWithExecution*(
+    self: ref ConsensusManager, initialNewHead: BeaconHead,
+    getBeaconTimeFn: GetBeaconTimeFn) {.async.} =
```
etan-status: Does `runProposalForkchoiceUpdated` require similar recovery logic? Or is it intentionally excluded there because it is already only run as an optimization? Or is it suspended anyway because while we are in optimistic mode we are not allowed to do proposals anyway, so head is not possible to become invalid?

Likewise, the fcu in `validator_duties`, I guess that one definitely cannot run into `INVALID` anymore? And if i does, recovery is too slow to be viable anyway?

etan-status (approved):
LGTM.

- fcu invalid result is propagated to fork choice
- fork choice is run again when current head becomes invalid.

Open question regarding treatment of `INVALID` in `newPayload`. Rest of functionality looks correct to me.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: No, the previous version had this and did more refactoring around this case, but if `newPayload` catches it, it doesn't get near the DAG or fork choice to begin with.

Generally, if `newPayload` returns `INVALID`, it just stops there. It shouldn't affect fcUs because Nimbus never further processes it beyond noting its inviability in case it gets that block again, as a quick shortcut to refusing it.

`newPayload` is the easy case.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: The most correct spec answer to this is `latestValidHash`:
https://github.com/ethereum/execution-apis/blob/4accc7ab19832aba6c484d70b32acfe000401e4a/src/engine/specification.md#payloadstatusv1
> `latestValidHash`: `DATA|null`, 32 Bytes - the hash of the most recent *valid* block in the branch defined by payload and its ancestors

That's in theory exactly what's desired here, though it's a problem when first syncing, because there might not be any `VALID` hash because that's how opt sync works. So we still might need this kind of more manual approach, and to treat LVH as an optimization.

The other aspect is that until maybe a month or two ago, ELs disagreed even on whether the LVH was `VALID` or `INVALID` (despite its name), so this would need to be tested for EL compatibility, which is out of scope of this PR. This uses only existing interfaces with the EL.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: `runProposalForkchoiceUpdated` is called from two places,
https://github.com/status-im/nimbus-eth2/blob/5d16719dbc538c6680f3e233d653f47a0620ca9b/beacon_chain/gossip_processing/block_processor.nim#L326-L337
and
https://github.com/status-im/nimbus-eth2/blob/5d16719dbc538c6680f3e233d653f47a0620ca9b/beacon_chain/consensus_object_pools/consensus_manager.nim#L353-L383

In both cases: (1) it runs after `dag.updateHead` is called, i.e. the code has decided that it's found a valid head, and (2) it runs `asyncSpawn`, i.e. it's not expecting to have errors to handle.

In the second `updateHeadWithExecution` case, it explicitly runs after the head already has been selected.

In the first, `storeBlock` case, it's being called when the block already has been determined to be `VALID`.

And, yes, by the time it's in `validator_duties`, it's a non-optimistic, `VALID` block already. It doesn't try to propose optimistically.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4081/commits/a71b6c0d3d12296cd330eb90bccbc27b57e72826

PR #4082, opened 2022-09-07
improve deposit processing performance

etan-status:
When there are a lot of deposits, we decompress the public key into a crypto cache. To avoid having those caches grow unreasonably big, make sure to operate on the decompressed pubkey instead.

### DIFF
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -530,16 +530,14 @@ const
   DOMAIN_CONTRIBUTION_AND_PROOF* = DomainType([byte 0x09, 0x00, 0x00, 0x00])
 
 func getImmutableValidatorData*(validator: Validator): ImmutableValidatorData2 =
-  let cookedKey = validator.pubkey.load() # Loading the pubkey is slow!
-  doAssert cookedKey.isSome,
-    "Cannot parse validator key: " & toHex(validator.pubkey)
+  let cookedKey = validator.pubkey.loadValid()  # `Validator` has valid key
   ImmutableValidatorData2(
-    pubkey: cookedKey.get(),
+    pubkey: cookedKey,
     withdrawal_credentials: validator.withdrawal_credentials)
 
 func getImmutableValidatorData*(
     deposit: DepositData): Opt[ImmutableValidatorData2] =
-  let cookedKey = deposit.pubkey.load()
+  let cookedKey = deposit.pubkey.load()  # Loading the pubkey is slow!
   if cookedKey.isNone:
     return err()
   ok ImmutableValidatorData2(
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -537,6 +537,15 @@ func getImmutableValidatorData*(validator: Validator): ImmutableValidatorData2 =
     pubkey: cookedKey.get(),
     withdrawal_credentials: validator.withdrawal_credentials)
 
+func getImmutableValidatorData*(
+    deposit: DepositData): Opt[ImmutableValidatorData2] =
+  let cookedKey = deposit.pubkey.load()
+  if cookedKey.isNone:
+    return err()
+  ok ImmutableValidatorData2(
+    pubkey: cookedKey.get(),
+    withdrawal_credentials: deposit.withdrawal_credentials)
+
 template makeLimitedUInt*(T: untyped, limit: SomeUnsignedInt) =
   # A "tigher" type is often used for T, but for the range check to be effective
   # it must make sense..
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -535,15 +535,6 @@ func getImmutableValidatorData*(validator: Validator): ImmutableValidatorData2 =
     pubkey: cookedKey,
     withdrawal_credentials: validator.withdrawal_credentials)
 
-func getImmutableValidatorData*(
-    deposit: DepositData): Opt[ImmutableValidatorData2] =
-  let cookedKey = deposit.pubkey.load()  # Loading the pubkey is slow!
-  if cookedKey.isNone:
-    return err()
-  ok ImmutableValidatorData2(
-    pubkey: cookedKey.get(),
-    withdrawal_credentials: deposit.withdrawal_credentials)
-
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/datatypes/base.nim
```diff
@@ -537,6 +537,15 @@ func getImmutableValidatorData*(validator: Validator): ImmutableValidatorData2 =
     pubkey: cookedKey.get(),
     withdrawal_credentials: validator.withdrawal_credentials)
 
+func getImmutableValidatorData*(
+    deposit: DepositData): Opt[ImmutableValidatorData2] =
+  let cookedKey = deposit.pubkey.load()
```
arnetheduck: this is _the_ operation that is slow ;)

etan-status: Yes, but if not done here, it additionally ends up in the `blsVerify` cache which is unbounded and never pruned.

Screenshot 2022-09-07 at 09 45 19

Screenshot 2022-09-07 at 09 45 51

beacon_chain/spec/datatypes/base.nim
```diff
@@ -530,13 +530,20 @@ const
   DOMAIN_CONTRIBUTION_AND_PROOF* = DomainType([byte 0x09, 0x00, 0x00, 0x00])
 
 func getImmutableValidatorData*(validator: Validator): ImmutableValidatorData2 =
-  let cookedKey = validator.pubkey.load() # Loading the pubkey is slow!
-  doAssert cookedKey.isSome,
-    "Cannot parse validator key: " & toHex(validator.pubkey)
+  let cookedKey = validator.pubkey.loadValid()  # `Validator` has valid key
```
etan-status, line 533: this effectively skips `blst_p1_affine_in_g1` just to assert :)

arnetheduck: `Validator` instances may come from an untrusted source (like the REST API or an invalid checkpoint sync state) - as long as this function is public, there's no guarantee that the pubkey inside is "valid enough"

arnetheduck: it's an odd function in general - I think it's worth rethinking this flow in general and see if there's low-hanging room for improvement

beacon_chain/spec/signatures.nim
```diff
@@ -200,13 +200,23 @@ func get_deposit_signature*(message: DepositMessage, version: Version,
...
 
-  blsVerify(deposit.pubkey, signing_root.data, deposit.signature)
+  blsVerify(pubkey, signing_root.data, deposit.signature)
+
+proc verify_deposit_signature*(preset: RuntimeConfig,
+                               deposit: DepositData): bool =
+  # Deposits come with compressed public keys; uncompressing them is expensive.
+  # `blsVerify` fills an internal cache when using `ValidatorPubKey`.
+  # To avoid filling this cache unnecessarily, uncompress explicitly here.
+  let validator = getImmutableValidatorData(deposit).valueOr:
```
arnetheduck: there's no need for this function here - to avoid the cache, you can simply `load` the pubkey directly (and thus avoid copying the credentials) - `getImmutableValidatorData` is really meant for the long-term validator cache and we don't seed it from deposits alone (rather, we seed it from the state after the block has been applied - in case the block is invalid, we don't want to seed the cache with invalid data)

mratsim:
To give some figures on the pubkey deserialization speed.

Source https://github.com/mratsim/constantine/pull/183

image

Pubkey deserialization without check: 11us
Pubkey deserialization with checks: 45 ns (ratio ~4.5x)

This is the same speed in Constantine, which will allow me to extrapolate uncompressed deserialization speed

image

## Compression explanation

A BLS12-381 G1 (pubkey) point has coordinate (x, y) with formula: `y² = x³ + 4`, when you compress, you only store the x coordinate and because sqrt(x³+4) has 2 solutions, you can store which y you refer to (+y or -y) in a single bit. Thankfully BLS12-381 only needs 381-bit and so we have 3 extra bits for metadata (since we can only store bytes, so 48 bytes = 384 bits).

## Difference between compressed and uncompressed speed

So the differences between compressed and uncompressed representation are:
- The size, compressed is 2x smaller
- The sqrt operation which cost 11us in Constantine, and I assume about the same in BLST (I actually have optimizations to save ~15% compared to BLST but in this particular use-case they are irrelevant)
  image

## Conclusion

99.9% of deserialization cost without checks is due to compression and we can make it virtually instant with uncompressed pubkeys

arnetheduck:
> we can make it virtually instant with uncompressed pubkeys

unfortunately we don't have uncompressed deposit pubkeys - so the only thing we can do with deposits that I know is to batch-verify all deposits in a block (unless there's a way to batch-load keys as well?).. 

oh, and as etan points out, we could avoid decompressing the keys twice, but that's probably a non-trivial flow to introduce

arnetheduck:
hm, maybe we can have a variant pubkey type that holds either compressed _or_ uncompressed key

mratsim:
>  (unless there's a way to batch-load keys as well?)..

We can load them in parallel but there is no math / cryptography or number-theoretic construct to improve batch-load like we can batch-verify.

> hm, maybe we can have a variant pubkey type that holds either compressed or uncompressed key

In terms of memory, the variant take as much space as the bigger type + 1 byte.
Also while in the previous case we were aligned, due to the runtime variant tag, the key will be unaligned, on x86 unaligned load aren't a problem, unsure about ARM.

Anyway the question becomes, are there cases where we would load a compressed pubkey without using it right away and so we want to delay paying that cost?

etan-status:
The flow here is (1) deposit processing, as part of which pubkey is fully loaded and checked; (2) create `Validator` from it which contains compressed form once more; (3) when the entire block checks out, as part of saving it to DB, load the pubkeys once more to have efficient DB access from then onward.

In this PR, (1) is modified to not pollute the last-resort crypto key cache, and (3) is modified to use the decompression without repeating the entire verification process.

arnetheduck (approved):

mratsim:
> (1) is modified to not pollute the last-resort crypto key cache

If someone becomes a validator they will attest once every 6 min anyway so it will be cached and we only delay the inevitable, or did I miss something?

etan-status:
There is a separate crypto key cache when using `ValidatorPubKey` with `blsVerify`. We have a separate cache of `CookedKey` in `beacon_chain_db` that already will hold validator keys. The changes here avoid the use of the separate crypto key cache in addition to the DB cache.

PR #4085, opened 2022-09-07
avoid rechecking bellatrix+ block signatures in state transition

arnetheduck:
Like the other forks, these are already batch-verified - this also avoids polluting the last-resort global key cache

### DIFF
--- a/beacon_chain/spec/datatypes/bellatrix.nim
+++ b/beacon_chain/spec/datatypes/bellatrix.nim
@@ -145,11 +145,11 @@ type
     finalized_checkpoint*: Checkpoint
 
     # Inactivity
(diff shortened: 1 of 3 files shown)


etan-status (approved):
🥲

beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -352,9 +361,6 @@ type
   BoolReturnSuccessRPC = object
     success*: bool
 
-func encodeQuantityHex*(x: auto): string =
-  "0x" & x.toHex
-
```
etan-status, line 357: intended?

arnetheduck: yes, it was replaced by `to0xHex`

PR #4088, opened 2022-09-07
increase blinded block submission timeout from 1s to 4s

tersec:
At this point, why not -- one can't fallback to a local EL anymore.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -49,7 +49,7 @@ const
   delayBuckets = [-Inf, -4.0, -2.0, -1.0, -0.5, -0.1, -0.05,
                   0.05, 0.1, 0.5, 1.0, 2.0, 4.0, 8.0, Inf]
 
-  BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE = 1.seconds
+  BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE = 4.seconds
   BUILDER_STATUS_DELAY_TOLERANCE = 3.seconds
   BUILDER_VALIDATOR_REGISTRATION_DELAY_TOLERANCE = 3.seconds
 


beacon_chain/validators/validator_duties.nim
```diff
@@ -49,7 +49,7 @@ const
   delayBuckets = [-Inf, -4.0, -2.0, -1.0, -0.5, -0.1, -0.05,
                   0.05, 0.1, 0.5, 1.0, 2.0, 4.0, 8.0, Inf]
 
-  BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE = 1.seconds
+  BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE = 4.seconds
```
etan-status, line 52: should it be tied to `INTERVALS_PER_SLOT` to take account for faster slots on e.g. gnosis?

etan-status (approved):

beacon_chain/validators/validator_duties.nim
tersec: Still better to try to propose. As long as it's less than slot time, it's fine.

arnetheduck: this doesn't leave any time for the empty block to propagate, no?

tersec: What empty block? It's too late for any kind of local fallback, empty or otherwise, by this point. To quote https://nimbus.guide/external-block-builder.html
> There exists a failure mode, intrinsic to the builder API, wherein the consensus client has signed a blinded proposal and therefore even if the external block builder relay or builder doesn't provide a full block, a consensus client such as Nimbus cannot safely proceed with step 2 and fall back on its local execution client.

"Safely proceed" == "without high risk of slashing".

arnetheduck: ah, this is for when we've already signed it?

tersec: Yes

PR #2429, opened 2021-03-18
Use the ValidatorIndex type in VoluntaryExit

zah:
Using the `ValidatorIndex` consistently in the spec datatypes can improve type safety in theory and it simplifies the implementation of the REST API modules where the numeric types need special treatment. This is a draft PR exploring the cost of such a refactoring.

PR #4089, opened 2022-09-07
Feature/capella

### LINKED ISSUE #3942: Implement the Capella beacon chain spec
The Capella beacon chain spec changes are defined here:
https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#introduction

As a first step, we should implement the Capella state transition in expectation of test vectors that might be published in the future. The task will include the following:

1) Define a new `capella` module in the `beacon_chain/spec/datatypes` folder. The module will hold the new definitions for types such as `ExecutionPayload`, `BeaconBlockBody`, `BeaconState`, etc, as well as the brand new types such as `Withdrawal` and `BLSToExecutionChange`.

2) Extend the `BeaconStateFork` enum with additional `Capella` value and handle the fallout from the change.

3) Introduce the modifications to the state transition functions (as done for previous forks such as Altair and Bellatrix). All existing code should continue to compile normally (i.e. `make` and `make test` should complete successfully). `block_sim` can be extended to test Capella transitions.

tavurth:
# Table of Contents

1.  [Bellatrix -> Capella (August 2022)](#org5a7261e)
    1.  [Getting stuck](#org835ddfe)
    2.  [Setup and overview](#orgbaf02eb)
    3.  [Spec files (capella.nim)](#org843501a)
    4.  [Fork file (forks.nim)](#org39f67a2)
    5.  [Review](#org6603b1e)
    6.  [Typing](#org50b8842)
    7.  [Duplication](#org7846de1)
        1.  [Duplication should be preferred](#org055837f)
    8.  [Rebase on `unstable`](#org1d98c5f)
    9.  [Tracking compiler errors](#orge9b96e9)
    10. [Ran into a serialization problem with the spec](#orgc07342f)
    11. [Large amount of typing errors](#org2334a4b)
    12. [Polymorphic errors](#orgedb1c82)
    13. [Ambiguous calls](#org2df0fac)
    14. [sizeof(U) !== sizeof(T)](#org883415d)

<a id="org5a7261e"></a>

# Bellatrix -> Capella (August 2022)

In this guide I mostly lay out the problems I ran into
while making the forward migration.

In general I would suggest getting the spec/datatypes/\\\* files
done first and them performing a search-and replace sweep on the
whole project.

Look for all complex matches and then prepare macros for different
files to sweep.

Stop at unviable areas, tag with TODO and get it to compile first.

&#x2014;

<a id="org835ddfe"></a>

## Getting stuck

If you run into some difficult type or other error that you can&rsquo;t process.
Try to compile a different part of the project using the command:

`./env.sh nim c beacon_chain/foo.nim`

<a id="orgbaf02eb"></a>

## Setup and overview

Started working on adding the relevant part of the spec files to the capella.nim I added
More cleaning up of the spec files

<a id="org843501a"></a>

## Spec files (capella.nim)

Here it was useful to copy from previous spec files rather than the spec itself
mostly because the spec file types can differ slightly
Functions at the bottom of the page were required for later compilation steps
Better to copy them and try to remove later if possible

<a id="org39f67a2"></a>

## Fork file (forks.nim)

Find and duplicate all bellatrix (previous entry) functions and type names
Better to search next and modify in this case, there are many hidden ones

<a id="org6603b1e"></a>

## Review

Found a few places that should be checked later because the previous
specversion had inconsistencies with the previous spec
marked with TODO and left for the end

<a id="org50b8842"></a>

## Typing

Started working on moving some types into helper and rest<sub>types</sub>
Sorted code in datatypes/capella.nim, and Added TODO to cleanup
these parts later when assured I could remove them (I could not),
More helpers work and slowly tracking the compiler errors

<a id="org7846de1"></a>

## Duplication

Duplication starts with finding functions that match your
previous fork name and duplicating them. I found that
quick-search and replace (in selected region) was very
helpful here.

Tried to generalise some functionality but was unable to
easily because of typing errors.

<a id="org055837f"></a>

### Duplication should be preferred

As it prevents errors with previous forks.

Compiler errors tracked me to more duplicates, however these
errors are not always helpful at telling you where the issue
is as the compiler complains a lot about non-matching types.

<a id="org1d98c5f"></a>

## Rebase on `unstable`

which was thankfully easy and required a few easy refactor

Understand the scope is much more spread
Search and replace prev + new fork everywhere

<a id="orge9b96e9"></a>

## Tracking compiler errors

In general these are helpful at this stage and tell you where
to start looking. However,

at this stage it would be more helpful to
start combing through files. I suggest:

-   prepare a list of files containing `complex(prev_fork)` declarations
-   comb those files using a macro for cases one by one

these steps above will likely save you many compiler errors
be sure to TODO tag the ones which differ from spec to spec

<a id="orgc07342f"></a>

## Ran into a serialization problem with the spec

seemed likely there was no spec version present,
but the code was a bit out of date in one part.
There was a better implementation below which
takes the version (in REST) from the JSON body

<a id="org2334a4b"></a>

## Large amount of typing errors

This was likely because I culled out some
important functions from `capella.nim` too early.

At this point I also ran into `sizeof(U) or sizeof(T)` problems
which was caused because the spec is duplicated in a few places
and also made immutable. All areas must be updated correctly.

More typing problems, this time with `shortLog` who&rsquo;s asking
for a BeaconBlock and getting one, but does not seem happy.
Found some more missed areas from the first code sweep and fixed
them up.

<a id="orgedb1c82"></a>

## Polymorphic errors

Sometimes we don&rsquo;t import the spec files directly, but rather export
them from some other file.

Search `export {previous_fork_name}` and be sure that all of them are
updated with your new fork name.

<a id="org2df0fac"></a>

## Ambiguous calls

I found that some places the spec has new features from the bellatrix
upgrade which need to be modified or in some way made non-ambiguous.

So far in the slow process of migration type ambiguity has been the single
largest slowdown to progress.

<a id="org883415d"></a>

## sizeof(U) !== sizeof(T)

Go into datatypes/base.nim and uncomment the following lines:

    # NOTE: Uncomment for debugging type size mismatch
    echo alignLeft($T.typeof & ":", 50), T.sizeof
    echo alignLeft($U.typeof & ":", 50), U.sizeof, "\n", repeat("-", 20)

Then rebuild. This should print out the typenames and the corresponding
size when an isomorphicCast attempt is made during compile time.
Hopefully this will allow you to narrow down exactly which types are
causing the issue.

Most of these issues are caused by a missing duplicate item in one of your spec
files. It could also be in the file `beacon_chain_db_immutable.nim`

### DIFF
--- a/.projectile
+++ b/.projectile
@@ -1,0 +1,2 @@
+- ConsensusSpecPreset-mainnet.md
+- ConsensusSpecPreset-minimal.md
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -743,7 +743,7 @@ proc putState*(
   db.statesNoVal[type(value).toFork()].putSnappySSZ(
     key.data, toBeaconStateNoImmutableValidators(value))
 
-proc putState*(db: BeaconChainDB, key: Eth2Digest, value: bellatrix.BeaconState) =
+proc putState*(db: BeaconChainDB, key: Eth2Digest, value: bellatrix.BeaconState | capella.BeaconState) =
   db.updateImmutableValidators(value.validators.asSeq())
   db.statesNoVal[type(value).toFork()].putSZSSZ(
     key.data, toBeaconStateNoImmutableValidators(value))
@@ -1033,7 +1033,6 @@ proc getStateOnlyMutableValidators(
 
 proc getStateOnlyMutableValidators(
     immutableValidators: openArray[ImmutableValidatorData2],
-    # TODO: @tavurth check
     store: KvStoreRef, key: openArray[byte], output: var bellatrix.BeaconState,
     rollback: RollbackProc): bool =
   ## Load state into `output` - BeaconState is large so we want to avoid
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -719,6 +719,10 @@ template toBeaconStateNoImmutableValidators(state: bellatrix.BeaconState):
     BellatrixBeaconStateNoImmutableValidators =
   isomorphicCast[BellatrixBeaconStateNoImmutableValidators](state)
 
+template toBeaconStateNoImmutableValidators(state: capella.BeaconState):
+    CapellaBeaconStateNoImmutableValidators =
+  isomorphicCast[CapellaBeaconStateNoImmutableValidators](state)
+
 proc putState*(
     db: BeaconChainDB, key: Eth2Digest,
     value: phase0.BeaconState | altair.BeaconState) =
@@ -893,6 +897,8 @@ proc getBlockSSZ*(
     getBlockSSZ(db, key, data, altair.TrustedSignedBeaconBlock)
   of BeaconBlockFork.Bellatrix:
     getBlockSSZ(db, key, data, bellatrix.TrustedSignedBeaconBlock)
+  of BeaconBlockFork.Capella:
+    getBlockSSZ(db, key, data, capella.TrustedSignedBeaconBlock)
 
 proc getBlockSZ*(
     db: BeaconChainDB, key: Eth2Digest, data: var seq[byte],
@@ -936,6 +942,8 @@ proc getBlockSZ*(
     getBlockSZ(db, key, data, altair.TrustedSignedBeaconBlock)
   of BeaconBlockFork.Bellatrix:
     getBlockSZ(db, key, data, bellatrix.TrustedSignedBeaconBlock)
+  of BeaconBlockFork.Capella:
+    getBlockSZ(db, key, data, capella.TrustedSignedBeaconBlock)
 
 proc getStateOnlyMutableValidators(
     immutableValidators: openArray[ImmutableValidatorData2],
@@ -1168,6 +1176,7 @@ proc containsState*(db: BeaconChainDBV0, key: Eth2Digest): bool =
     db.backend.contains(subkey(phase0.BeaconState, key)).expectDb()
 
 proc containsState*(db: BeaconChainDB, key: Eth2Digest, legacy: bool = true): bool =
+  db.statesNoVal[BeaconStateFork.Capella].contains(key.data).expectDb or
   db.statesNoVal[BeaconStateFork.Bellatrix].contains(key.data).expectDb or
   db.statesNoVal[BeaconStateFork.Altair].contains(key.data).expectDb or
   db.statesNoVal[BeaconStateFork.Phase0].contains(key.data).expectDb or
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -889,6 +889,16 @@ proc getBlockSSZ*(
 
 proc getBlockSSZ*(
     db: BeaconChainDB, key: Eth2Digest, data: var seq[byte],
+    T: type capella.TrustedSignedBeaconBlock): bool =
+  let dataPtr = addr data # Short-lived
+  var success = true
+  func decode(data: openArray[byte]) =
+    try: dataPtr[] = decodeFramed(data)
+    except CatchableError: success = false
+  db.blocks[T.toFork].get(key.data, decode).expectDb() and success
+
+proc getBlockSSZ*(
+    db: BeaconChainDB, key: Eth2Digest, data: var seq[byte],
     fork: BeaconBlockFork): bool =
   case fork
   of BeaconBlockFork.Phase0:
@@ -926,6 +936,15 @@ proc getBlockSZ*(
 proc getBlockSZ*(
     db: BeaconChainDB, key: Eth2Digest, data: var seq[byte],
     T: type bellatrix.TrustedSignedBeaconBlock): bool =
+  let dataPtr = addr data # Short-lived
+  var success = true
+  func decode(data: openArray[byte]) =
+    assign(dataPtr[], data)
+  db.blocks[T.toFork].get(key.data, decode).expectDb() and success
+
+proc getBlockSZ*(
+    db: BeaconChainDB, key: Eth2Digest, data: var seq[byte],
+    T: type capella.TrustedSignedBeaconBlock): bool =
   let dataPtr = addr data # Short-lived
   var success = true
   func decode(data: openArray[byte]) =
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -17,7 +17,7 @@ import
   eth/db/[kvstore, kvstore_sqlite3],
   ./networking/network_metadata, ./beacon_chain_db_immutable,
   ./spec/[eth2_ssz_serialization, eth2_merkleization, forks, state_transition],
-  ./spec/datatypes/[phase0, altair, bellatrix],
+  ./spec/datatypes/[phase0, altair, bellatrix, capella],
   "."/[beacon_chain_db_light_client, filepath]
 
 export
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -471,8 +471,7 @@ proc new*(T: type BeaconChainDB,
     lcData = db.initLightClientDataDB(LightClientDataDBNames(
       altairCurrentBranches: "lc_altair_current_branches",
       altairBestUpdates: "lc_altair_best_updates",
-      sealedPeriods: "lc_sealed_periods")
-    ).expectDb()
+      sealedPeriods: "lc_sealed_periods")).expectDb()
 
   # Versions prior to 1.4.0 (altair) stored validators in `immutable_validators`
   # which stores validator keys in compressed format - this is
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -451,14 +451,18 @@ proc new*(T: type BeaconChainDB,
     blocks = [
       kvStore db.openKvStore("blocks").expectDb(),
       kvStore db.openKvStore("altair_blocks").expectDb(),
-      kvStore db.openKvStore("bellatrix_blocks").expectDb()]
+      kvStore db.openKvStore("bellatrix_blocks").expectDb(),
+      kvStore db.openKvStore("capella_blocks").expectDb()
+    ]
 
     stateRoots = kvStore db.openKvStore("state_roots", true).expectDb()
 
     statesNoVal = [
       kvStore db.openKvStore("state_no_validators2").expectDb(),
       kvStore db.openKvStore("altair_state_no_validators").expectDb(),
-      kvStore db.openKvStore("bellatrix_state_no_validators").expectDb()]
+      kvStore db.openKvStore("bellatrix_state_no_validators").expectDb(),
+      kvStore db.openKvStore("capella_state_no_validators").expectDb()
+    ]
 
     stateDiffs = kvStore db.openKvStore("state_diffs").expectDb()
     summaries = kvStore db.openKvStore("beacon_block_summaries", true).expectDb()
@@ -467,7 +471,8 @@ proc new*(T: type BeaconChainDB,
     lcData = db.initLightClientDataDB(LightClientDataDBNames(
       altairCurrentBranches: "lc_altair_current_branches",
       altairBestUpdates: "lc_altair_best_updates",
-      sealedPeriods: "lc_sealed_periods")).expectDb()
+      sealedPeriods: "lc_sealed_periods")
+    ).expectDb()
 
   # Versions prior to 1.4.0 (altair) stored validators in `immutable_validators`
   # which stores validator keys in compressed format - this is
@@ -481,6 +486,7 @@ proc new*(T: type BeaconChainDB,
   if immutableValidatorsDb.len() < immutableValidatorsDb1.len():
     notice "Migrating validator keys, this may take a minute",
       len = immutableValidatorsDb1.len()
+
     while immutableValidatorsDb.len() < immutableValidatorsDb1.len():
       let val = immutableValidatorsDb1.get(immutableValidatorsDb.len())
       immutableValidatorsDb.add(ImmutableValidatorDataDb2(
@@ -693,6 +699,13 @@ proc putBlock*(
     db.blocks[type(value).toFork].putSZSSZ(value.root.data, value)
     db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
 
+proc putBlock*(
+    db: BeaconChainDB,
+    value: capella.TrustedSignedBeaconBlock) =
+  db.withManyWrites:
+    db.blocks[type(value).toFork].putSZSSZ(value.root.data, value)
+    db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
+
 proc updateImmutableValidators*(
     db: BeaconChainDB, validators: openArray[Validator]) =
   # Must be called before storing a state that references the new validators
@@ -824,6 +837,18 @@ proc getBlock*(
 
 proc getBlock*[
     X: bellatrix.TrustedSignedBeaconBlock](
+    db: BeaconChainDB, key: Eth2Digest,
+    T: type X): Opt[T] =
+  # We only store blocks that we trust in the database
+  result.ok(default(T))
+  if db.blocks[T.toFork].getSZSSZ(key.data, result.get) == GetResult.found:
+    # set root after deserializing (so it doesn't get zeroed)
+    result.get().root = key
+  else:
+    result.err()
+
+proc getBlock*[
+    X: capella.TrustedSignedBeaconBlock](
     db: BeaconChainDB, key: Eth2Digest,
     T: type X): Opt[T] =
   # We only store blocks that we trust in the database
@@ -1008,7 +1033,51 @@ proc getStateOnlyMutableValidators(
 
 proc getStateOnlyMutableValidators(
     immutableValidators: openArray[ImmutableValidatorData2],
+    # TODO: @tavurth check
     store: KvStoreRef, key: openArray[byte], output: var bellatrix.BeaconState,
+    rollback: RollbackProc): bool =
+  ## Load state into `output` - BeaconState is large so we want to avoid
+  ## re-allocating it if possible
+  ## Return `true` iff the entry was found in the database and `output` was
+  ## overwritten.
+  ## Rollback will be called only if output was partially written - if it was
+  ## not found at all, rollback will not be called
+  # TODO rollback is needed to deal with bug - use `noRollback` to ignore:
+  #      https://github.com/nim-lang/Nim/issues/14126
+  # TODO RVO is inefficient for large objects:
+  #      https://github.com/nim-lang/Nim/issues/13879
+
+  case store.getSZSSZ(key, toBeaconStateNoImmutableValidators(output))
+  of GetResult.found:
+    let numValidators = output.validators.len
+    doAssert immutableValidators.len >= numValidators
+
+    for i in 0 ..< numValidators:
+      let
+        # Bypass hash cache invalidation
+        dstValidator = addr output.validators.data[i]
+
+      assign(
+        dstValidator.pubkey,
+        immutableValidators[i].pubkey.toPubKey())
+      assign(
+        dstValidator.withdrawal_credentials,
+        immutableValidators[i].withdrawal_credentials)
+
+    output.validators.resetCache()
+
+    true
+  of GetResult.notFound:
+    false
+  of GetResult.corrupted:
+    rollback()
+    false
+
+# TODO: Could be split into generic
+proc getStateOnlyMutableValidators(
+    immutableValidators: openArray[ImmutableValidatorData2],
+    # TODO: @tavurth check
+    store: KvStoreRef, key: openArray[byte], output: var capella.BeaconState,
     rollback: RollbackProc): bool =
   ## Load state into `output` - BeaconState is large so we want to avoid
   ## re-allocating it if possible
@@ -1099,7 +1168,7 @@ proc getState*(
 
 proc getState*(
     db: BeaconChainDB, key: Eth2Digest,
-    output: var (altair.BeaconState | bellatrix.BeaconState),
+    output: var (altair.BeaconState | bellatrix.BeaconState | capella.BeaconState),
     rollback: RollbackProc): bool =
   ## Load state into `output` - BeaconState is large so we want to avoid
   ## re-allocating it if possible
@@ -1174,7 +1243,7 @@ proc containsBlock*(
     db.v0.containsBlock(key)
 
 proc containsBlock*[
-    X: altair.TrustedSignedBeaconBlock | bellatrix.TrustedSignedBeaconBlock](
+    X: altair.TrustedSignedBeaconBlock | bellatrix.TrustedSignedBeaconBlock | capella.TrustedSignedBeaconBlock](
     db: BeaconChainDB, key: Eth2Digest, T: type X): bool =
   db.blocks[X.toFork].contains(key.data).expectDb()
 
@@ -1184,7 +1253,8 @@ proc containsBlock*(db: BeaconChainDB, key: Eth2Digest, fork: BeaconBlockFork): 
   else: db.blocks[fork].contains(key.data).expectDb()
 
 proc containsBlock*(db: BeaconChainDB, key: Eth2Digest): bool =
-  db.containsBlock(key, bellatrix.TrustedSignedBeaconBlock) or
+  db.containsBlock(key, capella.TrustedSignedBeaconBlock) or
+    db.containsBlock(key, bellatrix.TrustedSignedBeaconBlock) or
     db.containsBlock(key, altair.TrustedSignedBeaconBlock) or
     db.containsBlock(key, phase0.TrustedSignedBeaconBlock)
 
@@ -1317,6 +1387,8 @@ iterator getAncestorSummaries*(db: BeaconChainDB, root: Eth2Digest):
     elif (let blck = db.getBlock(res.root, altair.TrustedSignedBeaconBlock); blck.isSome()):
       res.summary = blck.get().message.toBeaconBlockSummary()
     elif (let blck = db.getBlock(res.root, bellatrix.TrustedSignedBeaconBlock); blck.isSome()):
+      res.summary = blck.get().message.toBeaconBlockSummary()
+    elif (let blck = db.getBlock(res.root, capella.TrustedSignedBeaconBlock); blck.isSome()):
       res.summary = blck.get().message.toBeaconBlockSummary()
     else:
       break
--- a/beacon_chain/beacon_chain_db_immutable.nim
+++ b/beacon_chain/beacon_chain_db_immutable.nim
@@ -11,7 +11,7 @@ else:
   {.push raises: [].}
 
 import
-  ./spec/datatypes/[base, altair, bellatrix],
+  ./spec/datatypes/[base, altair, bellatrix, capella],
   ./spec/[eth2_ssz_serialization, eth2_merkleization]
 
 type
@@ -181,4 +181,68 @@ type
     next_sync_committee*: SyncCommittee
 
     # Execution
-    latest_execution_payload_header*: ExecutionPayloadHeader  # [New in Bellatrix]
+    latest_execution_payload_header*: bellatrix.ExecutionPayloadHeader  # [New in Bellatrix]
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/capella/beacon-chain.md#beaconstate
+  # Memory-representation-equivalent to a Capella BeaconState for in-place SSZ
+  # reading and writing
+  CapellaBeaconStateNoImmutableValidators* = object
+    # Versioning
+    genesis_time*: uint64
+    genesis_validators_root*: Eth2Digest
+    slot*: Slot
+    fork*: Fork
+
+    # History
+    latest_block_header*: BeaconBlockHeader ##\
+    ## `latest_block_header.state_root == ZERO_HASH` temporarily
+
+    block_roots*: HashArray[Limit SLOTS_PER_HISTORICAL_ROOT, Eth2Digest] ##\
+    ## Needed to process attestations, older to newer
+
+    state_roots*: HashArray[Limit SLOTS_PER_HISTORICAL_ROOT, Eth2Digest]
+    historical_roots*: HashList[Eth2Digest, Limit HISTORICAL_ROOTS_LIMIT]
+
+    # Eth1
+    eth1_data*: Eth1Data
+    eth1_data_votes*:
+      HashList[Eth1Data, Limit(EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH)]
+    eth1_deposit_index*: uint64
+
+    # Registry
+    validators*: HashList[ValidatorStatus, Limit VALIDATOR_REGISTRY_LIMIT]
+    balances*: HashList[uint64, Limit VALIDATOR_REGISTRY_LIMIT]
+
+    # Randomness
+    randao_mixes*: HashArray[Limit EPOCHS_PER_HISTORICAL_VECTOR, Eth2Digest]
+
+    # Slashings
+    slashings*: HashArray[Limit EPOCHS_PER_SLASHINGS_VECTOR, uint64] ##\
+    ## Per-epoch sums of slashed effective balances
+
+    # Participation
+    previous_epoch_participation*: EpochParticipationFlags
+    current_epoch_participation*: EpochParticipationFlags
+
+    # Finality
+    justification_bits*: JustificationBits
+
+    previous_justified_checkpoint*: Checkpoint ##\
+    ## Previous epoch snapshot
+
+    current_justified_checkpoint*: Checkpoint
+    finalized_checkpoint*: Checkpoint
+
+    # Inactivity
+    inactivity_scores*: HashList[uint64, Limit VALIDATOR_REGISTRY_LIMIT]
+
+    # Sync
+    current_sync_committee*: SyncCommittee
+    next_sync_committee*: SyncCommittee
+
+    # Execution
+    latest_execution_payload_header*: capella.ExecutionPayloadHeader
+
+    # Withdrawals
+    withdrawal_queue*: List[Withdrawal, Limit capella.WITHDRAWALS_QUEUE_LIMIT]  # [New in Capella]
+    next_withdrawal_index*: WithdrawalIndex  # [New in Capella]
--- a/beacon_chain/beacon_chain_db_immutable.nim
+++ b/beacon_chain/beacon_chain_db_immutable.nim
@@ -194,47 +194,40 @@ type
     fork*: Fork
 
     # History
-    latest_block_header*: BeaconBlockHeader ##\
-    ## `latest_block_header.state_root == ZERO_HASH` temporarily
-
-    block_roots*: HashArray[Limit SLOTS_PER_HISTORICAL_ROOT, Eth2Digest] ##\
-    ## Needed to process attestations, older to newer
+    latest_block_header*: BeaconBlockHeader
+    block_roots*: HashArray[Limit SLOTS_PER_HISTORICAL_ROOT, Eth2Digest]
+      ## Needed to process attestations, older to newer
 
     state_roots*: HashArray[Limit SLOTS_PER_HISTORICAL_ROOT, Eth2Digest]
     historical_roots*: HashList[Eth2Digest, Limit HISTORICAL_ROOTS_LIMIT]
 
     # Eth1
     eth1_data*: Eth1Data
-    eth1_data_votes*:
-      HashList[Eth1Data, Limit(EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH)]
+    eth1_data_votes*: List[Eth1Data, Limit EPOCHS_PER_ETH1_VOTING_PERIOD * SLOTS_PER_EPOCH]
     eth1_deposit_index*: uint64
 
     # Registry
-    validators*: HashList[ValidatorStatus, Limit VALIDATOR_REGISTRY_LIMIT]
-    balances*: HashList[uint64, Limit VALIDATOR_REGISTRY_LIMIT]
+    validators*: HashList[Validator, Limit VALIDATOR_REGISTRY_LIMIT]
+    balances*: HashList[Gwei, Limit VALIDATOR_REGISTRY_LIMIT]
 
     # Randomness
     randao_mixes*: HashArray[Limit EPOCHS_PER_HISTORICAL_VECTOR, Eth2Digest]
 
     # Slashings
-    slashings*: HashArray[Limit EPOCHS_PER_SLASHINGS_VECTOR, uint64] ##\
-    ## Per-epoch sums of slashed effective balances
+    slashings*: HashArray[Limit EPOCHS_PER_SLASHINGS_VECTOR, Gwei]  # Per-epoch sums of slashed effective balances
 
     # Participation
     previous_epoch_participation*: EpochParticipationFlags
     current_epoch_participation*: EpochParticipationFlags
 
     # Finality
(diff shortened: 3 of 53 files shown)


tersec:
2022-09-09-151411_274x443_scrot
why does it modify all these submodule commits?

scripts/compile_nim_program.sh
```diff
@@ -38,8 +38,8 @@ build/generate_makefile "nimcache/release/${BINARY}/${PROJECT_NAME}.json" "nimca
 if uname | grep -qi darwin || [[ -n "${FORCE_DSYMUTIL}" ]]; then
   [[ -z "${DSYMUTIL}" ]] && DSYMUTIL="dsymutil"
   # Scary warnings in large volume: https://github.com/status-im/nimbus-eth2/issues/3076
-  "${DSYMUTIL}" build/${BINARY} 2>&1 \
-    | grep -v "failed to insert symbol" \
-    | grep -v "could not find object file symbol for symbol" \
-    || true
+  "${DSYMUTIL}" build/${BINARY} 2>&1 |
```
tersec, line 41: This could/should be in a separate PR. It's unrelated to Capella.

beacon_chain/validators/validator_duties.nim
```diff
@@ -551,6 +607,33 @@ proc makeBeaconBlockForHeadAndSlot*(
...
+        # https://github.com/nim-lang/Nim/issues/19802
+        (static(default(capella.ExecutionPayload)))
+      else:
+        let pubkey = node.dag.validatorKey(validator_index)
+        (await getExecutionPayload(
+          node, proposalState,
+          slot.epoch, validator_index,
+          # TODO https://github.com/nim-lang/Nim/issues/19802
+          if pubkey.isSome: pubkey.get.toPubKey else: default(ValidatorPubKey))),
+>>>>>>> b97c30aa (Try to fix ambiguous calls)
```
tersec: This `>>>>>>>` is a rebasing artifact, it looks like

beacon_chain/validators/validator_duties.nim
```diff
@@ -505,8 +561,8 @@ proc makeBeaconBlockForHeadAndSlot*(
       warn "Eth1 deposits not available. Skipping block proposal", slot
       return ForkedBlockResult.err("Eth1 deposits not available")
 
-    # Only current hardfork with execution payloads is Bellatrix
-    static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
+    # Only current hardforks with execution payloads are Bellatrix & Capella
+    static: doAssert high(BeaconStateFork) == BeaconStateFork.Capella
```
tersec, line 552: One goal of this change is to generalize enough to make it clear what the path is for forks beyond Capella. They'll presumably all have execution payloads (unless they change even more fundamentally).

beacon_chain/spec/beaconstate.nim
```diff
@@ -158,7 +160,7 @@ func get_whistleblower_reward*(validator_effective_balance: Gwei): Gwei =
 func get_proposer_reward(state: ForkyBeaconState, whistleblower_reward: Gwei): Gwei =
   when state is phase0.BeaconState:
     whistleblower_reward div PROPOSER_REWARD_QUOTIENT
-  elif state is altair.BeaconState or state is bellatrix.BeaconState:
+  elif state is ForkyUpgradedBeaconState:
```
tersec: This might be worth keeping fork-by-fork. Some of these fork-selections are based on that sync committees came into Altair and probably will remain in post-Altair forks, litkewise `ExecutionPayload`s in Bellatrix, but this is not some a constant which has already changed and could easily change again. It probably should be something that by default fails to compile in newer hardforks until someone checks that, yes, it's like the previous hardfork, rather than assuming this for all future hardforks.

beacon_chain/spec/beaconstate.nim
```diff
@@ -356,6 +358,19 @@ func get_initial_beacon_block*(state: bellatrix.HashedBeaconState):
...
+  # The genesis block is implicitly trusted
+  let message = capella.TrustedBeaconBlock(
+    slot: state.data.slot,
+    state_root: state.root)
+    # parent_root, randao_reveal, eth1_data, signature, and body automatically
+    # initialized to default values.
+  capella.TrustedSignedBeaconBlock(
+    message: message, root: hash_tree_root(message))
+
+
```
tersec: Extra newline

beacon_chain/spec/beaconstate.nim
```diff
@@ -562,8 +577,8 @@ func check_attestation_index*(
     Result[CommitteeIndex, cstring] =
   check_attestation_index(data.index, committees_per_slot)
 
-# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.3/specs/altair/beacon-chain.md#get_attestation_participation_flag_indices
-func get_attestation_participation_flag_indices(state: altair.BeaconState | bellatrix.BeaconState,
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/altair/beacon-chain.md#get_attestation_participation_flag_indices
```
tersec: Why change this from rc.3 to rc.1?

beacon_chain/spec/beaconstate.nim
```diff
@@ -881,9 +896,15 @@ func upgrade_to_altair*(cfg: RuntimeConfig, pre: phase0.BeaconState):
   post
 
 # https://github.com/ethereum/consensus-specs/blob/v1.1.7/specs/merge/fork.md#upgrading-the-state
-func upgrade_to_bellatrix*(cfg: RuntimeConfig, pre: altair.BeaconState):
+proc upgrade_to_bellatrix*(cfg: RuntimeConfig, pre: altair.BeaconState):
```
tersec: What necessitates this change from `func` to `proc`?

beacon_chain/spec/datatypes/altair.nim
```diff
@@ -102,6 +102,9 @@ type
     beacon_block_root*: Eth2Digest
       ## Block root for this signature
 
+    # TODO: This should probably be `ValidatiorIndex`
```
tersec, line 105: https://github.com/status-im/nimbus-eth2/pull/2429 attempted this. Probably a good idea, but also, not related to Capella and shouldn't be part of a Capella PR. Already enough changes/noise, no need to add unrelated ones.

beacon_chain/spec/datatypes/capella.nim
```diff
@@ -0,0 +1,455 @@
...
+
+  MAX_BLS_TO_EXECUTION_CHANGES*: uint64 = 16
+
+  # 2**40
+  WITHDRAWALS_QUEUE_LIMIT*: uint64 = (uint64) 1_099_511_627_776
+
+  DOMAIN_BLS_TO_EXECUTION_CHANGE* = DomainType([byte 0x0A, 0x00, 0x00, 0x00])
+
+type
+  # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#custom-types
```
tersec: The current spec version is 1.2.0-rc.3 -- should ensure that everything matches that and update the URLs. I think these are all fine.

beacon_chain/spec/datatypes/capella.nim
```diff
@@ -0,0 +1,455 @@
...
+  Transaction* = List[byte, Limit MAX_BYTES_PER_TRANSACTION]
+
+  ExecutionAddress* = object
+    data*: array[20, byte]  # TODO there's a network_metadata type, but the import hierarchy's inconvenient
+
+  BloomLogs* = object
```
tersec: There's no need to re-declare `Transaction`, `ExecutionAddress`, or `BloomLogs`. https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#containers does not change them, so the `bellatrix`-module versions are fine. They're abstractions from https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#custom-types and https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#executionpayload which Capella also uses. The execution layer addresses, for example, certainly don't change just because the consensus layer hardfork changes.

beacon_chain/spec/datatypes/capella.nim
```diff
@@ -0,0 +1,455 @@
...
+    voluntary_exits*: List[SignedVoluntaryExit, Limit MAX_VOLUNTARY_EXITS]
+    sync_aggregate*: SyncAggregate # TODO TrustedSyncAggregate after batching
+
+    # Execution
+    execution_payload*: ExecutionPayload
+
+    # Capella operations
+    bls_to_execution_changes*: List[SignedBLSToExecutionChange, Limit MAX_BLS_TO_EXECUTION_CHANGES]  # [New in Capella]
+
+  #
```
tersec, line 221: This entire "Duplicates from bellatrix" can be removed

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -749,16 +752,18 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: Eth2Digest) {.
      raises: [IOError, Defect].} =
   writeValue(writer, hexOriginal(value.data))
 
+type ForkyBloomLogs = bellatrix.BloomLogs | capella.BloomLogs
```
tersec: This is why not to re-declare pointlessly different `BloomLogs` types. Just use the `bellatrix` one and the changes here aren't necessary.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -749,16 +752,18 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: Eth2Digest) {.
...
+
 ## BloomLogs
-proc readValue*(reader: var JsonReader[RestJson], value: var BloomLogs) {.
+proc readValue*(reader: var JsonReader[RestJson], value: var ForkyBloomLogs) {.
```
tersec: `ForkyBloomLogs` shouldn't exist as a type

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -749,16 +752,18 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: Eth2Digest) {.
...
      raises: [IOError, SerializationError, Defect].} =
   try:
     hexToByteArray(reader.readValue(string), value.data)
   except ValueError:
     raiseUnexpectedValue(reader,
                          "BloomLogs value should be a valid hex string")
 
-proc writeValue*(writer: var JsonWriter[RestJson], value: BloomLogs) {.
+proc writeValue*(writer: var JsonWriter[RestJson], value: ForkyBloomLogs) {.
```
tersec: `ForkyBloomLogs` shouldn't exist as a type

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -310,6 +311,15 @@ type
     BYTES_PER_LOGS_BLOOM*: uint64
     MAX_EXTRA_DATA_BYTES*: uint64
 
+    # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/presets/mainnet/capella.yaml
```
tersec, line 314: More 1.2.0-rc.1 stuff, see elsewhere re 1.2.0-rc.3

beacon_chain/spec/state_transition_block.nim
```diff
@@ -23,10 +23,10 @@ else:
   {.push raises: [].}
 
 import
-  std/[algorithm, sequtils, sets, tables],
+  std/[algorithm, options, sequtils, sets, tables],
```
tersec, line 26: Preference is to use `Opt` from `stew/results` in place of Nim's stdlib `Option`, because it allows usage of `valueOr` and other advantages.

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -747,7 +747,30 @@ iterator get_inactivity_penalty_deltas*(
           state.inactivity_scores[vidx]
       yield (vidx, Gwei(penalty_numerator div penalty_denominator))
 
-# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.3/specs/phase0/beacon-chain.md#rewards-and-penalties-1
+# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/bellatrix/beacon-chain.md#modified-get_inactivity_penalty_deltas
```
tersec, line 750: more v1.2.0-rc.3/rc.1 issues (won't point out all of them, but some)

.projectile
```diff
@@ -0,0 +1,2 @@
+- ConsensusSpecPreset-mainnet.md
```
tersec, line 1: What is this file?

beacon_chain/spec/forks.nim
```diff
@@ -42,43 +42,61 @@ type
...
 
   ForkyBeaconState* =
     phase0.BeaconState |
     altair.BeaconState |
-    bellatrix.BeaconState
+    bellatrix.BeaconState |
+    capella.BeaconState
+
+  # Beacon state which does not include the base fork
+  ForkyUpgradedBeaconState* =
```
tersec, line 55: Presumably Bellatrix is an upgrade, too -- it's reasonable to have a name for this typeclass, e.g., to deal with states/blocks which have sync committees, but `ForkyUpgradedBeaconState` is ambiguous already will only become moreso with Capella (which is also an upgrade).

beacon_chain/spec/forks.nim
```diff
@@ -590,31 +703,43 @@ func bellatrixFork*(cfg: RuntimeConfig): Fork =
...
 func forkVersionAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): Version =
   case cfg.stateForkAtEpoch(epoch)
+  of BeaconStateFork.Capella:   cfg.CAPELLA_FORK_VERSION
   of BeaconStateFork.Bellatrix: cfg.BELLATRIX_FORK_VERSION
   of BeaconStateFork.Altair:    cfg.ALTAIR_FORK_VERSION
   of BeaconStateFork.Phase0:    cfg.GENESIS_FORK_VERSION
 
 func nextForkEpochAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): Epoch =
   case cfg.stateForkAtEpoch(epoch)
+  # TODO: Is Capella a FAR_FUTURE?
```
tersec, line 731: Yes, in Capella, for now, there's no next fork defined, so it can only be `FAR_FUTURE_EPOCH`. But the `BeaconStateFork.Bellatrix` case should then know about `cfg.CAPELLA_FORK_EPOCH`. At any given time, there should be exactly one `FAR_FUTURE_EPOCH` here.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -318,18 +318,18 @@ func asEth2Digest*(x: BlockHash): Eth2Digest =
 template asBlockHash*(x: Eth2Digest): BlockHash =
   BlockHash(x.data)
 
-func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV1):
+func asConsensusExecutionPayload*(payload_id: bellatrix.PayloadID, rpcExecutionPayload: ExecutionPayloadV1):
```
tersec: What does `asConsensusExecutionPayload` use `payload_id` for? It's not part of either `ExecutionPayloadV1` or `bellatrix.ExecutionPayload`?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -318,18 +318,18 @@ func asEth2Digest*(x: BlockHash): Eth2Digest =
...
     bellatrix.ExecutionPayload =
   template getTransaction(tt: TypedTransaction): bellatrix.Transaction =
     bellatrix.Transaction.init(tt.distinctBase)
 
   bellatrix.ExecutionPayload(
     parent_hash: rpcExecutionPayload.parentHash.asEth2Digest,
     feeRecipient:
-      ExecutionAddress(data: rpcExecutionPayload.feeRecipient.distinctBase),
+      bellatrix.ExecutionAddress(data: rpcExecutionPayload.feeRecipient.distinctBase),
```
tersec: Should not need this -- the EL addresses have not changed format

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -340,7 +340,32 @@ func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV1):
         rpcExecutionPayload.extraData.distinctBase),
     base_fee_per_gas: rpcExecutionPayload.baseFeePerGas,
     block_hash: rpcExecutionPayload.blockHash.asEth2Digest,
-    transactions: List[bellatrix.Transaction, MAX_TRANSACTIONS_PER_PAYLOAD].init(
+    transactions: List[bellatrix.Transaction, bellatrix.MAX_TRANSACTIONS_PER_PAYLOAD].init(
+      mapIt(rpcExecutionPayload.transactions, it.getTransaction)))
+
+func asConsensusExecutionPayload*(payload_id: capella.PayloadID, rpcExecutionPayload: ExecutionPayloadV1):
```
tersec: Same comments as the Bellatrix version about:
1) the PayloadID isn't fork-dependent, so it shouldn't be duplicated to begin with;
2) neither `ExecutionPayloadV1` nor `capella.ExecutionPayload` (https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#executionpayload) have PayloadID, so it's unnecessary as a parameter to begin with.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -340,7 +340,32 @@ func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV1):
...
+    block_number: rpcExecutionPayload.blockNumber.uint64,
+    gas_limit: rpcExecutionPayload.gasLimit.uint64,
+    gas_used: rpcExecutionPayload.gasUsed.uint64,
+    timestamp: rpcExecutionPayload.timestamp.uint64,
+    extra_data:
+      List[byte, MAX_EXTRA_DATA_BYTES].init(
+        rpcExecutionPayload.extraData.distinctBase),
+    base_fee_per_gas: rpcExecutionPayload.baseFeePerGas,
+    block_hash: rpcExecutionPayload.blockHash.asEth2Digest,
+    transactions: List[capella.Transaction, capella.MAX_TRANSACTIONS_PER_PAYLOAD].init(
```
tersec: See elsewhere about `Transaction` -- it's not changed. https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#custom-types defines a transaction as a
```python
ByteList[MAX_BYTES_PER_TRANSACTION]
```
and https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#custom-types does not change this.

tersec: As a result, it should not be necessary to specify that it's specifically a `capella` transaction rather than a `bellatrix` transaction.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -514,6 +573,7 @@ proc forkchoiceUpdated*(p: Eth1Monitor,
       finalizedBlockHash: finalizedBlock.asBlockHash),
     none(engine_api.PayloadAttributesV1))
 
+
```
tersec: ?

beacon_chain/spec/beaconstate.nim
```diff
@@ -934,9 +955,10 @@ func upgrade_to_bellatrix*(cfg: RuntimeConfig, pre: altair.BeaconState):
     next_sync_committee: pre.next_sync_committee,
 
     # Execution-layer
-    latest_execution_payload_header: ExecutionPayloadHeader()
+    latest_execution_payload_header: executionHeader
```
tersec, line 961: This really should be able to just directly call `bellatrix.ExecutionPayloadHeader` directly. If there's a genuine Nim bug hwere, that can be tracked down in a reproducible test case.

tests/test_eth1_monitor.nim
```diff
@@ -499,4 +499,5 @@ suite "Eth1 monitor":
     for executionPayload in executionPayloads:
       check:
         executionPayload == asConsensusExecutionPayload(
+          bellatrix.PayloadID(),
```
tersec: ?

ncli/ncli_db.nim
```diff
@@ -316,12 +323,16 @@ proc cmdBench(conf: DbConf, cfg: RuntimeConfig) =
               of BeaconStateFork.Bellatrix:
                 doAssert dbBenchmark.getState(
                   state.root, loadedState[2][].data, noRollback)
+              of BeaconStateFork.Capella:
+                doAssert dbBenchmark.getState(
+                  state.root, loadedState[3][].data, noRollback)
```
tersec, line 328: These states are sizable, and needed to be managed carefully to keep i386 builds running tests, so worth ensuring it keeps that test runner within a 2GB address space.

tavurth:
> why does it modify all these submodule commits?

That's a good question 🤔 Before I had some problems with these vendor areas during a rebase I will revert them.

beacon_chain/validators/validator_duties.nim
tavurth: I'm not exactly sure what you mean to change here?

beacon_chain/spec/beaconstate.nim
tavurth: I implemented the new type `ForkyUpgradedBeaconState` for everyone above `phase0`

It seems like we will have many many of these or statements very soon if we leave it in the current style 🤔 

I agree with the checking before implementation though, I've used an array check here instead 👍

beacon_chain/spec/beaconstate.nim
tavurth: Hmm this was probably a copy paste from above

beacon_chain/spec/beaconstate.nim
tavurth: Thanks for the catch!

I found many many size errors while upgrading due to the fact that the specs have to be changed in multiple places while upgrading. 

Due to the size errors I had to log the sizes of objects in `base.nim` and that required side effects I think. Reverted 👍

beacon_chain/spec/datatypes/altair.nim
tavurth: Ok will leave it as a TODO?

beacon_chain/spec/datatypes/capella.nim
tavurth: Thank you! 

I just duplicated everything with the idea to go back and cleanup later when everything was working. However it's taken far longer than expected 😢

beacon_chain/spec/datatypes/altair.nim
tersec: That's fine, but (a) typo in `ValidatiorIndex`, and (b) not directly related to Capella, so if you want to add these TODOs, that should be a separate PR.

beacon_chain/spec/beaconstate.nim
tersec: https://nim-lang.org/docs/system.html#debugEcho%2Cvarargs%5Btyped%2C%5D

beacon_chain/spec/datatypes/altair.nim
arnetheduck: Also - see the explanation here: https://github.com/status-im/nimbus-eth2/blob/stable/beacon_chain/spec/datatypes/base.nim#L49 - during serialization we use a wider type so as to be able to correctly parse all messages, including those containing nonsensical data - validating whether data is correct requires a context that the parser does not have and that changes over time (thus can't be represented well with type).

Long story short: it should not be a `ValidatorIndex` here.

beacon_chain/beacon_chain_db.nim
```diff
@@ -467,7 +471,8 @@ proc new*(T: type BeaconChainDB,
     lcData = db.initLightClientDataDB(LightClientDataDBNames(
       altairCurrentBranches: "lc_altair_current_branches",
       altairBestUpdates: "lc_altair_best_updates",
-      sealedPeriods: "lc_sealed_periods")).expectDb()
+      sealedPeriods: "lc_sealed_periods")
```
arnetheduck: generally avoid unnecessary reformatting in branches that have other, meaningful changes

arnetheduck: ie we don't have an auto-formatter, so small differences in style are expected

beacon_chain/spec/beaconstate.nim
arnetheduck: > It seems like we will have many many of these or statements very soon if we leave it in the current style 

this is fine - forks happen rarely - if it becomes an _actual_) problem we can deal with it then (probably via some property-based mechanism - something like `when hasAltairWhistleBlower(state)`.


For now though, the aim is to get to a working state _without_ introducing any new style or constructs (ie such introduction of new constructs is a separate PR / work item and can be done  either before or after, but not together)

arnetheduck: Also, `Upgraded` is somewhat arbitrary - it doesn't say when it was upgraded.

beacon_chain/spec/datatypes/capella.nim
tersec: Adding that duplication adds (as you discovered) its own in this case unnecessary complexity, due to creating symbols for types with the same names in different modules and requiring modifying all the functions using those types.

beacon_chain/spec/beaconstate.nim
tersec: Yes, see my comment elsewhere about the name of `ForkyUpgradedBeaconState` (it's too broad/generic/ambiguous). But that aside, I'm distinguishing here from structural features of these hardforks which won't likely disappear in, say, whatever hardforks end up starting with "d" or "e" (the Capella+1 and Capella+2 hardforks), where I'd agree it's appropriate to avoid the `or`-filled conditions by having a typeclass that's basically "forky beacon state which supports sync committees" [just, not phase0, so Altair or newer] or  "forky beacon state which supports execution payloads" [Bellatrix or newer] from examples such as this, where what's being selected for isn't such a structural feature of a hardfork but a constant that can and has been changed before already.

The difference is that if one uses the general typeclass for upgraded states, then it will automatically compile when that typeclass is updated to include the new hardfork. But that's not necessarily ideal, because even though it's highly likely Capella+1 and Capella+2 will support all of sync committees, execution payloads, and withdrawals, so typeclasses based on selecting for that functionality should be fine, it could easily change these non-structural numeric constants, so it's safer to ensure that doesn't compile until someone hand-checks it.

.projectile
tavurth: It's used to prevent emacs from searching inside certain files.
Usually I jump around by looking up names, and this file contains far too many instances

Screenshot 2022-09-12 at 14 12 55

tersec: In general, the repo doesn't contain editor configs like this.

beacon_chain/validators/validator_duties.nim
tersec: An approach general enough to handle both Bellatrix and Capella `ExecutionPayload`s/`ExecutionPayloadHeader`s ideally would be able not to rely on this static-`doAssert` of what the highest known fork is.

The reason these `static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix` statements are around in some places is because in their vicinity, something directly references `someForkedState.bellatrixData` or `someForkedBlock.bellatrixData`, which works currently because the only execution payloads are in Bellatrix. But once it works for Bellatrix and Capella, it might not be able use that kind of hard-coding anyway, in which case the static-`doAsserts` add nothing.

.projectile
tersec: Also, regardless, another non-Capella-related change.

beacon_chain/spec/beaconstate.nim
etan-status: You can use `typeof(state).toFork >= BeaconStateFork.Altair` inside `when`.

Advantage is that someone reviewing bellatrix spec can find all changes that started in bellatrix by searching for bellatrix :) no more noise from altair-introduced changes.

beacon_chain/eth1/eth1_monitor.nim
tavurth: Normally I have a tool such as prettier remove these for me

tests/test_eth1_monitor.nim
tavurth: Still not refactored as I moved from a `bellatrix` `capella` payloadId to a base payload ID

Still stuck in the endless compile errors

tersec: (1) there aren't different types of payload IDs, really

(2) `asConsensusExecutionPayload` doesn't know/care about a payload ID anyway that I can tell, just the payload contents (to which the ID is something of a temporary handle)? It's not used in `asConsensusExecutionPayload`.

beacon_chain/eth1/eth1_monitor.nim
tersec: If you can make `nimpretty` or some other more automated tool behave consistently with the https://status-im.github.io/nim-style-guide/ that's fine, of course, but only if it doesn't require en masse reformatting of entire modules along the way, since in general Nimbus and its vendored Nim-language repositories haven't been built using any such tools.

beacon_chain/beacon_chain_db_immutable.nim
```diff
@@ -181,4 +181,66 @@ type
...
+    # Execution
+    latest_execution_payload_header*: capella.ExecutionPayloadHeader
+
+    # Withdrawals
+    withdrawal_queue*: List[Withdrawal, Limit capella.WITHDRAWALS_QUEUE_LIMIT]  # [New in Capella]
+    next_withdrawal_index*: WithdrawalIndex  # [New in Capella]
+
+    # TODO: @tavurth We should probably use ValidatorIndex here
+    #       but I see that in Altair we have a workaround
+    #       by simply putting uint64, which seems to be against the spec
```
tersec, line 245: (1) `ValidatorIndex` isn't used, I don't think, in any of the SSZ object definitions in Nimbus, not just Altair

(2) https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#custom-types specifies that `ValidatorIndex`'s SSZ-equivalent type is `uint64`. How is using `uint64` against the spec?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -473,8 +478,7 @@ proc getBlockByNumber*(p: Web3DataProviderRef,
   except ValueError as exc: raiseAssert exc.msg # Never fails
   p.web3.provider.eth_getBlockByNumber(hexNumber, false)
 
-proc getPayload*(p: Eth1Monitor,
-                 payloadId: bellatrix.PayloadID): Future[engine_api.ExecutionPayloadV1] =
+proc getPayload*(p: Eth1Monitor, payloadId: base.PayloadID): Future[engine_api.ExecutionPayloadV1] =
```
tersec: This is a 100-character line; https://status-im.github.io/nim-style-guide/formatting.style.html notes that

beacon_chain/spec/datatypes/base.nim
```diff
@@ -139,6 +139,9 @@ type
   # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.3/specs/phase0/beacon-chain.md#custom-types
   Eth2Domain* = array[32, byte]
 
+  # Used in capella and bellatrix
+  PayloadID* = array[8, byte]
```
tersec: `base` has generally contained things of use across all supported forks. `PayloadID` is not useful or meaningful in phase0 or Altair.

beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -341,6 +339,9 @@ type
     BeaconBlockBody |
     SigVerifiedBeaconBlockBody |
     TrustedBeaconBlockBody
+  # NOTE: for nicer errors
+  SomeBellatrixBeaconBlock =
```
tersec: Fine as a debugging tool, not for merging.

beacon_chain/validators/validator_duties.nim
```diff
@@ -315,6 +315,11 @@ proc getBlockProposalEth1Data*(node: BeaconNode,
       finalizedEpochRef.eth1_deposit_index)
 
 from web3/engine_api import ForkchoiceUpdatedResponse
+from web3/engine_api_types import PayloadExecutionStatus, PayloadID
+
+
+converter to_buf(payload_id: engine_api_types.PayloadID): array[8, byte] =
```
tersec: https://status-im.github.io/nim-style-guide/language.converters.html
> Avoid using converters.

beacon_chain/validators/validator_duties.nim
```diff
@@ -375,18 +408,30 @@ proc getFeeRecipient(node: BeaconNode,
...
+    proposalState.bellatrixData.data
+
+
 proc getExecutionPayload(
     node: BeaconNode, proposalState: auto,
     epoch: Epoch,
     validator_index: ValidatorIndex,
     pubkey: ValidatorPubKey):
-    Future[Opt[ExecutionPayload]] {.async.} =
+    Future[bellatrix.ExecutionPayload | capella.ExecutionPayload] {.async.} =
```
tersec, line 413: A type class as as a return "type" isn't going to work, because per https://nim-lang.org/docs/manual.html#generics-type-classes
> Whilst the syntax of type classes appears to resemble that of ADTs/algebraic data types in ML-like languages, it should be understood that type classes are static constraints to be enforced at type instantiations. Type classes are not really types in themselves but are instead a system of providing generic "checks" that ultimately resolve to some singular type. Type classes do not allow for runtime type dynamism, unlike object variants or methods.

tersec: Also, what happened to the `Opt[]` part? `getExecutionPayload` can fail.

beacon_chain/validators/validator_duties.nim
```diff
@@ -375,18 +408,30 @@ proc getFeeRecipient(node: BeaconNode,
 
 from web3/engine_api_types import PayloadExecutionStatus
 
+
+template get_fork_data(epoch: Epoch): auto {.inline.} =
```
tersec: `template` is always effectively `inline`. I'm not sure if it's valid Nim, but it's not useful either way.

beacon_chain/validators/validator_duties.nim
```diff
@@ -375,18 +408,30 @@ proc getFeeRecipient(node: BeaconNode,
...
+  if epoch == BeaconStateFork.Capella:
```
tersec: (1) `epoch == BeaconStateFork.Capella` (which shows up elsewhere too) doesn't mean anything. It's comparing an `Epoch` (a `distinct uint64`) with an enum value (`BeaconStateFork.Capella`), where the latter is semi-arbitrary, the exact integer values mostly important in terms of their ordering rather than exact value. To determine what fork an epoch is in:
https://github.com/status-im/nimbus-eth2/blob/8be964a152a25d6954aa38d48e81d7575ea1fc6c/beacon_chain/spec/forks.nim#L384-L399

But this needs to know the `RuntimeConfig`, which tells it `ALTAIR_FORK_EPOCH`, `BELLATRIX_FORK_EPOCH`, and (soon) `CAPELLA_FORK_EPOCH`.

(2) This kind of `if` statement won't work because `if` is a creature of runtime and `proposalState.capellaData.data` is of a different type than `proposalState.bellatrixData.data`. For example:
```nim
let a =
  if true:
    3
  else:
    "4"
```
results in
`if_expr.nim(5, 5) Error: type mismatch: got <string> but expected 'int literal(3)'`
even though a human looking at that can see it always takes the branch resulting in a literal `int`. Nim's AST requires that all branches of an `if` or `case` expression have consistent types, so that the expression as a whole has a consistent, resolved type at compile time.

Sometimes, one can use `when` expressions rather than `if` because `when` is a compile-time creature and creates the AST, so it's not subject to this constraint. But, even so...

(3) Any sort of `if/when is_capella: foo else: bar` where `bar` is taken to mean `bellatrix` just perpetuates the fragility in the current codebase and will break and need updating with Capella+1, and/or fail in mysterious ways.

One way to enforce this more broadly is to use `case`, if one can engineer around the consistent-type issue (see point (2), because
> If there are no else or elif parts and not all possible values that expr can hold occur in a slicelist, a static error occurs. This holds only for expressions of ordinal types. "All possible values" of expr are determined by expr's type.
>
> ...
>
> Because case statements are checked for exhaustiveness during semantic analysis, ...

So it won't silently generate incorrect code in this circumstance, and is one reason that `case` statements and expressions are favored when feasible to handle different forks.

beacon_chain/validators/validator_duties.nim
```diff
@@ -375,18 +408,30 @@ proc getFeeRecipient(node: BeaconNode,
...
+    Future[bellatrix.ExecutionPayload | capella.ExecutionPayload] {.async.} =
   # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
 
-  # Only current hardfork with execution payloads is Bellatrix
-  static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
+  # Only current hardfork with execution payloads is Bellatrix & Capella
+  static: doAssert high(BeaconStateFork) == BeaconStateFork.Capella
   template empty_execution_payload(): auto =
-    build_empty_execution_payload(proposalState.bellatrixData.data)
+    if Epoch == BeaconStateFork.Capella:
```
tersec: See https://github.com/status-im/nimbus-eth2/pull/4089/files#r968991028

beacon_chain/validators/validator_duties.nim
```diff
@@ -440,12 +485,19 @@ proc getExecutionPayload(
            latestHead, latestSafe, latestFinalized,
            feeRecipient, node.consensusManager.eth1Monitor))
       payload = try:
+          let execution_payload =
+            if Epoch == BeaconStateFork.Capella:
```
tersec: https://github.com/status-im/nimbus-eth2/pull/4089/files#r968991028

beacon_chain/spec/datatypes/bellatrix.nim
tavurth: I see this here also: 🤔 

https://github.com/status-im/nimbus-eth2/blob/33122a208d5ceeafbcbeb4aeac3065431f2d3035/beacon_chain/spec/state_transition_block.nim#L599

beacon_chain/validators/validator_duties.nim
tavurth: Thank you very helpful 🙏

beacon_chain/validators/validator_duties.nim
```diff
@@ -380,13 +410,17 @@ proc getExecutionPayload(
...
   # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
 
-  # Only current hardfork with execution payloads is Bellatrix
-  static: doAssert high(BeaconStateFork) == BeaconStateFork.Bellatrix
+  # Only current hardfork with execution payloads is Bellatrix & Capella
+  static: doAssert high(BeaconStateFork) == BeaconStateFork.Capella
+
   template empty_execution_payload(): auto =
-    build_empty_execution_payload(proposalState.bellatrixData.data)
+    if epoch >= node.dag.cfg.CAPELLA_FORK_EPOCH:
```
tersec, line 420: When Capella+1 comes out, this will break without further warning because it only checks for `>= node.dag.cfg.CAPELLA_FORK_EPOCH`, which will be true. but then access the `capellaData` field of the case object, which will cause a defect.

beacon_chain/validators/validator_duties.nim
```diff
@@ -440,12 +474,21 @@ proc getExecutionPayload(
            latestHead, latestSafe, latestFinalized,
            feeRecipient, node.consensusManager.eth1Monitor))
       payload = try:
+          let execution_payload =
+            if epoch >= node.dag.cfg.CAPELLA_FORK_EPOCH:
```
tersec, line 478: When Capella+1 comes out, this will break without further warning because it only checks for `>= node.dag.cfg.CAPELLA_FORK_EPOCH`, which will be true. but then access the `capellaData` field of the case object, which will cause a defect.

arnetheduck:
thanks for the docs and PR - capella support has now been added via other PR:s

PR #4103, opened 2022-09-09
move i386 testing to nightlies + remove testing of Nim 1.4

narimiran:


tersec (approved):

arnetheduck:
This should be in the job named `Daily` - `nightly` creates a release - it doesn't run the test suite

.github/workflows/nightly_build.yml
```diff
@@ -56,6 +56,52 @@ jobs:
           path: ./dist/${{ steps.make_dist.outputs.archive_dir }}/build/nimbus_validator_client.sha512sum
           retention-days: 2
 
+  build-i386:
+    name: Linux i386 release asset
```
zah, line 60: This nightly job only packages the i386 build. What we want to do is to run the full test suite as well.

narimiran:
> This should be in the job named `Daily` - `nightly` creates a release - it doesn't run the test suite

If I'm not missing something, the daily job is in the file called `cron.yml`, and it already contains i386 Linux.
Then the only question is — do we want to also create a release for i386 Linux (in `nightly_build.yml`) or keep it as it was? (So the only change would be removing i386 Linux from the CI job)

(cc @zah, as this also answers your remark)

zah:
OK, but why is the Daily job in such a bad health?

https://github.com/status-im/nimbus-eth2/actions/runs/3063418614

This should be investigated.
Also, its failures should be communicated to the team in a better way (looks like it has deteriorated long time ago).

PR #4114, opened 2022-09-12
delay first exchangeTransitionConfiguration

tersec:


arnetheduck:
wouldn't it make more sense to ignore zero-TTD responses completely?

tersec:
I'm not sure what the issue is?

etan-status (approved):

PR #4121, opened 2022-09-15
update default preset

arnetheduck:


beacon_chain/spec/presets.nim
```diff
@@ -186,7 +186,7 @@ when const_preset == "mainnet":
     # ---------------------------------------------------------------
     # TBD, 2**256-2**10 is a placeholder
     TERMINAL_TOTAL_DIFFICULTY:
-      u256"115792089237316195423570985008687907853269984665640564039457584007913129638912",
+      u256"58750000000000000000000",
```
mratsim, line 189: We should use 0x, here, all the specs use 0x and from the numbers alone, it doesn't seem like hex, it could as well be decimal.

Furthermore u256 defaults to decimal parsing, we should use `Uint256.fromHex` here:

- https://github.com/status-im/nim-stint/blob/e656ad40d1815f618c12c7636a37e6971959aa62/stint.nim#L23
- https://github.com/status-im/nim-stint/blob/e656ad40d1815f618c12c7636a37e6971959aa62/stint/io.nim#L270

mratsim (dismissed):

beacon_chain/spec/presets.nim
arnetheduck: this is straight from the yaml: https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.3/configs/mainnet.yaml#L16 - it uses a decimal value

zah: I think this change is not appropriate. 

The `defaultRuntimeConfig` value that is being specified here is an object that is not explicitly covered by the spec, but semantically, in our codebase, it is not the "run-time config for mainnet".  We have the mainnet run-time config stored in another constant which is `mainnetMetadata.cfg`. This one is appropriately loaded from the mainnet `config.yaml` file present in the `eth2-networks` repo.

 `defaultRuntimeConfig` is a value that we use as a fallback when a config file for a particular network doesn't specify a particular run-time setting. The Gnosis network in particular is using such a sparse config and for most fields it's quite natural to assume that the designers of the Gnosis network meant to use the same settings as mainnet (thus, the `defaultRuntimeConfig` indeed matches the mainnet settings). 
 
 When this line of thinking gets a bit more controversial is exactly in regard to settings such as `TERMINAL_TOTAL_DIFFICULTY` where it makes very little sense for new custom networks to use the value `58750000000000000000000` as a default. Instead, the long constant that is replaced here is actually referenced in various other specs (such as the Engine API) exactly to denote networks where the TTD is not yet specified, so it makes a lot of sense to keep it as the default value that should be used until overridden by the network config file.
 
 What we can do is suggest that this notion of "default run-time config" should be described properly in the spec to give an unambiguous interpretation to sparse config such as the one from Gnosis. Please note that any run-time config for a phase0 or Altair chain will be sparse with regard to Bellatrix and it should get the safe default value.

etan-status: Gnosis uses a custom preset / config, with custom block time even.

The EF tests depend on `defaultRuntimeConfig`, which is based on the `mainnet.yaml` and `minimal.yaml` values from consensus-specs (with the real TTD and fork epoch values, actually).

But also, the EF tests for Altair don't actually use Bellatrix specific constants, as they have a completely separate implementation generating those tests that is not bellatrix aware.

zah: I've recently removed many usages of `defaultRuntimeConfig` and replaced them with `mainnetMetadata.cfg` exactly for the reasons I'm describing above.

etan-status: Would `minimal` need an equivalent `minimalMetadata.cfg`?

arnetheduck: according to the comment above the variable, the values herein are derived from the given yml file - it seems like keeping it in sync makes sense - whatever other chain derives their config from `mainnet.yml` will take a look at a snapshot of it and make a diff - this means we have two unsavoury choices: either break "recent" chains that target the same spec version as us _or_ break old chains that target an old version of the yml. given these are more or less equally bad, seems more simply to keep _all_ our code locked in on one version rather than to mix and match  randomly.

arnetheduck: the alternative is to drop back all parameters to some dummy values - but halving half/half with altair but not bellatrix seems like a lose-lose approach - I have no particular preference on default vs explicit mainnet beyond noting that bringing in the full metadata significantly adds to compilation time (I've tested this recently - linking takes significantly longer when including the states)

zah: > according to the comment above the variable

We should change the comment to match the actual role of this object (which is the one I described)

> either break "recent" chains that target the same spec version as us 

How are we breaking these recent chains if the default value stays as I suggest? Was my comment easy to understand? Perhaps I should try to rephrase it.

arnetheduck: > How are we breaking these recent chains if the default value stays as I suggest?

imagine a chain being created today off mainnet today (a fork for example, like ETC or a shadow fork for the next HF) - they would create a network config and change maybe a few values, but they'd use the "current" mainnet as template changing only what's needed so the diff wouldn't contain the bellatrix epoch and thus be broken - ie the diff depends on when you make it because upstream, this file changes over time.

I'm not against keeping dummies and defaults in this file - but then the altair epoch should be reset as well to far-future - either is fine, it's the mixing of the two approaches that will be hard to sort out

zah: People who are creating new testnet configs are likely to test them against the actual clients and thus it's highly unlikely that someone creating a forked mainnet config will remove the `TERMINAL_TOTAL_DIFFICULTY` setting.

This would be especially true if we also do the next step and suggest that these type of sparse configs should be actually described in the spec (and I'm arguing that the treatment I'm proposing is much more reasonable than assuming that new and old networks will get mainnet's TERMINAL_TOTAL_DIFFICULTY by default).

arnetheduck:
#4147 replaces this

PR #4124, opened 2022-09-15
docs: add note about data-dir

Shiti:
The trusted node sync has to be executed over an empty `data-dir`. Added this as a note in the docs.


tersec:
Should target `unstable`

Shiti:
@tersec I updated the target branch but I see that some checks have not even started. Is this expected ?

tersec:
> @tersec I updated the target branch but I see that some checks have not even started. Is this expected ?

Yes, there's a glitch in how the GitHub actions CIs are triggered in this case, where a PR doesn't modify anything they (correctly) view as not modifying the aspect of the Nim code which affects their purview. The flaw here isn't that they aren't running -- there's no real reason to run CI on this particular change -- but that they show as pending.

It's fine.

PR #4126, opened 2022-09-16
clean up exchange configuration handling

arnetheduck:
Per spec, we should not be sending our detected terminal block to EL - the EL configuration exchange should only look at values from configuration and report mismatches.

* note TODO about usage of terminal block hash configuration

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -559,7 +559,7 @@ proc exchangeTransitionConfiguration*(p: Eth1Monitor): Future[EtcStatus] {.async
   if dataProvider.isNil:
     return EtcStatus.exchangeError
 
-  # https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#engine_exchangetransitionconfigurationv1
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1
   let consensusCfg = TransitionConfigurationV1(
     terminalTotalDifficulty: p.depositsChain.cfg.TERMINAL_TOTAL_DIFFICULTY,
     terminalBlockHash: p.depositsChain.cfg.TERMINAL_BLOCK_HASH,
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -136,7 +136,6 @@ type
 
     exchangedConfiguration*: bool
     terminalBlockHash*: Option[BlockHash]
-    terminalBlockNumber*: Option[Quantity]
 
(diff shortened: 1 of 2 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -561,67 +559,46 @@ proc exchangeTransitionConfiguration*(p: Eth1Monitor): Future[EtcStatus] {.async
   if dataProvider.isNil:
     return EtcStatus.exchangeError
 
+  # https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#engine_exchangetransitionconfigurationv1
```
tersec: https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1

arnetheduck:
```suggestion
  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1
```

beacon_chain/spec/presets.nim
```diff
@@ -38,8 +38,7 @@ type
 
     # Transition
     TERMINAL_TOTAL_DIFFICULTY*: UInt256
-    TERMINAL_BLOCK_HASH*: BlockHash
-    # TODO TERMINAL_BLOCK_HASH_ACTIVATION_EPOCH*: Epoch
+    TERMINAL_BLOCK_HASH*: BlockHash # TODO use in eht1monitor
```
tersec, line 41:
```suggestion
    TERMINAL_BLOCK_HASH*: BlockHash # TODO use in eth1monitor
```

tersec (approved):

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -561,67 +559,46 @@ proc exchangeTransitionConfiguration*(p: Eth1Monitor): Future[EtcStatus] {.async
...
-      if p.terminalBlockHash.isSome:
-        p.terminalBlockHash.get
-      else:
-        (static default BlockHash),
-    terminalBlockNumber:
-      if p.terminalBlockNumber.isSome:
-        p.terminalBlockNumber.get
-      else:
-        (static default Quantity))
+    terminalBlockHash: p.depositsChain.cfg.TERMINAL_BLOCK_HASH,
```
zah, line 565: The old code was not fully wrong here I think. We still need to consider the command-line override that the user may supply.

arnetheduck: we consider the command line here: https://github.com/status-im/nimbus-eth2/blob/667c3c97ebe3c40fd724cf80c991410e1cd32900/beacon_chain/nimbus_beacon_node.nim#L1830

PR #4127, opened 2022-09-16
Post merge polish

mratsim:
Some post-merge polish after looking at some log files:

- Say that we are connected to an Execution Client (not just Engine API which is a technical term users don't care about)
- Use Execution/Consensus instead of eth1 / eth2 in user logs (not done for debug)

One thing that was recurring as well is the people not updating --web3-url=http://localhost:8545 to --web3-url=http://localhost:8551 (or websocket).

I think we need to add a log whether we are connected as a mere web3 (to monitor deposits), or as a fully-fledged CL. Though now that mainnet is switch to PoS, maybe there is no interest at all in pre-PoS with deposit monitoring as all chains of interest have moved to PoS? Or use a different consensus algorithms already integrated within an EL (for rollups for example).

Notably missing: do we send a stub to the ELs on mainnet or do we like today send the actual expected values?

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -597,6 +597,14 @@ proc exchangeTransitionConfiguration*(p: Eth1Monitor): Future[EtcStatus] {.async
       terminalBlockHash = executionCfg.terminalBlockHash,
       terminalBlockNumber = executionCfg.terminalBlockNumber.uint64
 
+  if executionCfg.terminalBlockHash == default BlockHash:
+    # If TERMINAL_BLOCK_HASH is stubbed with
+    # 0x0000000000000000000000000000000000000000000000000000000000000000 then
+    # TERMINAL_BLOCK_HASH and TERMINAL_BLOCK_NUMBER parameters MUST NOT take
+    # an effect.
+    trace "Engine API using stub hash"
+    return EtcStatus.match
+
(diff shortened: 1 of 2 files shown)


mratsim:
TODO overlap with #4126 @arnetheduck

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -580,45 +580,46 @@ proc exchangeTransitionConfiguration*(p: Eth1Monitor): Future[EtcStatus] {.async
           consensusCfg),
         timeout = 1.seconds)
     except CatchableError as err:
-      error "Failed to exchange transition configuration", err = err.msg
+      error "Failed to exchange transition configuration witrh Execution Client", err = err.msg
```
tersec, line 583:
```suggestion
      error "Failed to exchange transition configuration with execution client", err = err.msg
```

tersec:
In general, are "execution layer" and "execution client" treated as proper nouns by the spec? I agree with using that rather than engine API, but I'm not sure about the capitalization here.

arnetheduck:
If I were changing the strings, I'd probably hint about `engine` configuration somewhere at least, given that in executions clients one has to enable the `engine` api specifically

arnetheduck:
#4126 needs to go in before this one though - until then, we're simply not sending the right messages at the right time and whatever we print here will not be right

arnetheduck:
these changes have been, or are being, incorporated in various other PR:s, closing for now

PR #4129, opened 2022-09-16
Remove optimistic sync candidate check

tersec:
As https://github.com/prysmaticlabs/prysm/pull/11453 notes, the window of hazard has passed.


etan-status (approved):
LC will need similar updates. Will follow-up on that.

PR #4134, opened 2022-09-16
allow custom configs in Gnosis binary

etan-status:
When running as Gnosis-chain binary the config was no longer adjustable. Restores loading custom configs when running as Gnosis-chain binary, as long as the following keys remain same:
- SLOTS_PER_EPOCH=16
- SECONDS_PER_SLOT=5
- BASE_REWARD_FACTOR=25
- EPOCHS_PER_SYNC_COMMITTEE_PERIOD=512

This allows running the Gnosis-chain binary on custom test networks.

### DIFF
--- a/beacon_chain/networking/network_metadata.nim
+++ b/beacon_chain/networking/network_metadata.nim
@@ -200,76 +200,79 @@ template mergeTestnet(path: string, eth1Network: Eth1Network): Eth2NetworkMetada
   loadCompileTimeNetworkMetadata(mergeTestnetsDir & "/" & path,
                                  some eth1Network)
 
-when not defined(gnosisChainBinary):
-  when const_preset == "mainnet":
-    const
-      mainnetMetadata* = eth2Network("shared/mainnet", mainnet)
-      praterMetadata* = eth2Network("shared/prater", goerli)
-      ropstenMetadata* = mergeTestnet("ropsten-beacon-chain", ropsten)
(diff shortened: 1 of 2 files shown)


beacon_chain/networking/network_metadata.nim
```diff
@@ -200,76 +200,79 @@ template mergeTestnet(path: string, eth1Network: Eth1Network): Eth2NetworkMetada
...
+    quit 1
 
-    if eth2Network.isSome and eth2Network.get == "gnosis-chain":
-      warn "`--network:gnosis-chain` is deprecated, use `--network:gnosis` instead"
+  metadata
 
-  proc getRuntimeConfig*(eth2Network: Option[string]): RuntimeConfig {.raises: [Defect, IOError].} =
-    checkNetworkParameterUse eth2Network
-    gnosisMetadata.cfg
+proc getRuntimeConfig*(
```
zah, line 271: This function should be removed now and all usages can be replaced with `loadEth2Network(ethNetwork).cfg` (or its moral equivalent).

In particular, the use of `defaultRuntimeConfig` is inappropriate due to the reasoning provided in https://github.com/status-im/nimbus-eth2/pull/4121#discussion_r971967818

etan-status: Sounds more like a followup rather than a part of relaxing the Gnosis specific restrictions, but generally agree that this can now be done.

PR #4137, opened 2022-09-16
fix TTD before bellatrix

etan-status:
When TTD hits before Bellatrix, avoid waiting for new blocks and detect the TTD block as the terminal block hash even before Bellatrix hits. Also allow detecting EL genesis block as merge transition block. This fixes the local testnet simulation with Geth to actually merge.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -1322,7 +1322,7 @@ func init(T: type FullBlockId, blk: Eth1BlockHeader|BlockObject): T =
   FullBlockId(number: Eth1BlockNumber blk.number, hash: blk.hash)
 
 func isNewLastBlock(m: Eth1Monitor, blk: Eth1BlockHeader|BlockObject): bool =
-  blk.number.uint64 > m.latestEth1BlockNumber
+  m.latestEth1Block.isNone or blk.number.uint64 > m.latestEth1BlockNumber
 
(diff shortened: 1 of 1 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1508,7 +1508,7 @@ proc startEth1Syncing(m: Eth1Monitor, delayBeforeStart: Duration) {.async.} =
     #      we should try to fetch that block from the EL - this facility is not
     #      in use on any current network, but should be implemented for full
     #      compliance
-    if m.currentEpoch >= m.cfg.BELLATRIX_FORK_EPOCH and m.terminalBlockHash.isNone:
+    if m.terminalBlockHash.isNone:
       var terminalBlockCandidate = nextBlock
 
       debug "startEth1Syncing: checking for merge terminal block",
```
zah, line 1519: Hmm, but wouldn't this line be printed too early in networks that are still running (or just simulating) Altair rules? 

Does the merge happen in the simulation when the Bellatrix epoch is reached by the CL?

etan-status: Those have a `TERMINAL_TOTAL_DIFFICULTY` of `~0`. So, it will work fine, but could lead to annoying logs. Will add a check for that.

etan-status: Regarding your second question: Yes

PR #4138, opened 2022-09-16
run local finalization testnets with Geth

etan-status:
Enable full 🐼 testing in local finalization testnets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,7 +200,7 @@ local-testnet-minimal:
 		--el-port-offset 5 \
 		--timeout 600 \
 		--kill-old-processes \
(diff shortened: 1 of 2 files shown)


etan-status:
image

etan-status:
Geth doesn't provide binaries for ARM macOS, so restricted to Intel for now.

etan-status:
Checked Jenkins CI, and confirmed that Geth is being run on Linux and macOS-Intel.
On macOS-AS, there are no official Geth binaries, so running without Geth there.

arnetheduck:
nice!

PR #4146, opened 2022-09-19
Trusted node sync clarifications in nimbus book

kdeme:
- Make it clear that it is a separate command from actually
running the beacon node
- Fix and extend the note on starting from a clean folder

^Those are two common mistakes seen on helpdesk, attempt to clarify that here. Second was apparently already added in an earlier PR, but it splitted the above Hint in two. 

Technically, it is also only the `db` directory that needs to be clean, but it is probably only more confusing to add/explain that in the book here?

### DIFF
--- a/docs/the_nimbus_book/src/trusted-node-sync.md
+++ b/docs/the_nimbus_book/src/trusted-node-sync.md
@@ -21,10 +21,12 @@ It is possibly to use trusted node sync with a third-party API provider -- see [
 !!! tip
     Make sure to replace `http://localhost:5052` in the commands below with the appropriate endpoint of the trusted beacon node. `http://localhost:5052` is the default endpoint exposed by Nimbus, but this is not consistent across all clients.
 
-!!! note ""
-    The path specified for data-dir must be an empty directory
-
     For example, if your trusted node is a [Prysm node](https://docs.prylabs.network/docs/how-prysm-works/ethereum-public-api#performing-requests-against-a-local-prysm-node), it exposes `127.0.0.1:3500` by default. Which means you would run the commands below with `--trusted-node-url=http://127.0.0.1:3500`
+
+!!! note
+    The path specified for `--data-dir` must be an empty directory as trusted node sync needs to be started from a fresh database.
+
+To start trusted node sync, run:
 
 === "Mainnet"
     ```sh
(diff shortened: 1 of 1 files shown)


docs/the_nimbus_book/src/trusted-node-sync.md
```diff
@@ -21,11 +21,13 @@ It is possibly to use trusted node sync with a third-party API provider -- see [
 !!! tip
     Make sure to replace `http://localhost:5052` in the commands below with the appropriate endpoint of the trusted beacon node. `http://localhost:5052` is the default endpoint exposed by Nimbus, but this is not consistent across all clients.
 
-!!! note ""
-    The path specified for data-dir must be an empty directory
-
     For example, if your trusted node is a [Prysm node](https://docs.prylabs.network/docs/how-prysm-works/ethereum-public-api#performing-requests-against-a-local-prysm-node), it exposes `127.0.0.1:3500` by default. Which means you would run the commands below with `--trusted-node-url=http://127.0.0.1:3500`
 
+!!! note
+    The path specified for `--data-dir` must be an empty directory as trusted node sync needs to be started from a fresh database.
```
arnetheduck, line 27: this is not strictly correct: if it's started with an existing dir, it'll resume the backfill

kdeme: You mean if it was an existing db that had been trusted synced before, right? I had actually first written down something else as I hadn't noticed the small PR which added this line. 

My original line was something like this:
```
!!! note
    If a node was run before with the default syncing mechanism and the same `--data-dir` is used for the `trustedNodeSync` command, then no trusted node sync will be done. The trusted node sync needs to be started from a fresh database and thus it is required to have an empty or no `db` folder in the provided `--data-dir`.
```
Still not fully correct I guess, but at least it clarifies somewhat that Trusted node sync and the default sync can't work together. I can try to adjust to something more condense that tackles both.

arnetheduck: yeah, so if you start a normal node, then run tns, it'll do nothing because the data dir already exists with _some_ state in it - we should probably improve TNS itself at some point, but for now, TNS will never re-establish sync on an existing database

kdeme: Yes, an initial improvement would be to have at least different log messages based on TNS run on a previous TNS versus TNS run on a regular synced database. For the latter case it doesn't show anything obvious in the logs that the outcome is not what the user is probably expecting. The user will only notice something is off when the nimbus node is restarted.

PR #4147, opened 2022-09-19
Fix and document some usages of defaultRuntimeConfig

zah:
#4134 follow-up work, discussed here:
https://github.com/status-im/nimbus-eth2/pull/4134#discussion_r973539811

### DIFF
--- a/beacon_chain/spec/presets.nim
+++ b/beacon_chain/spec/presets.nim
@@ -103,8 +103,12 @@ when const_preset == "mainnet":
   # TODO Move this to RuntimeConfig
   const SECONDS_PER_SLOT* {.intdefine.}: uint64 = 12
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/configs/mainnet.yaml
-  # TODO Read these from yaml file
+  # The default run-time config specifies the default configuration values
+  # that will be used if a particular run-time config is missing specific
+  # confugration values (which will be then taken from this config object).
+  # It mostly matches the mainnet config with the exception of few properties
+  # such as `CONFIG_NAME`, `TERMINAL_TOTAL_DIFFICULTY`, `*_FORK_EPOCH`, etc
+  # which must be effectively overriden in all network (including mainnet).
   const defaultRuntimeConfig* = RuntimeConfig(
     # Mainnet config
 
@@ -118,7 +122,7 @@ when const_preset == "mainnet":
     # * 'ropsten' - testnet
     # * 'sepolia' - testnet
     # Must match the regex: [a-z0-9\-]
-    CONFIG_NAME: "mainnet",
+    CONFIG_NAME: "",
 
     # Transition
     # ---------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


beacon_chain/networking/network_metadata.nim
```diff
@@ -270,9 +270,29 @@ proc getMetadataForNetwork*(
 
 proc getRuntimeConfig*(
     eth2Network: Option[string]): RuntimeConfig {.raises: [Defect, IOError].} =
+  ## Returns the run-time config for a network specified on the command line
+  ## If the network is not explicitly specified, the function will act as the
+  ## regular Nimbus binary, returning the mainnet config.
+  ##
+  ## TODO the assumption that the input variable is a CLI config option is not
+  ## quite appropriate in such as low-level function. Usagege of this function
+  ## can be replaced with `getMetadataForNetwork(eth2Network.get).cfg` which
```
arnetheduck: the network metadata includes the genesis state, bootnodes etc - things that apps don't want to include in general - the network metadata is relevant only to the beacon node while tools and tests often need a runtime configuration alone

beacon_chain/spec/presets.nim
```diff
@@ -152,7 +153,7 @@ when const_preset == "mainnet":
 
     # Altair
     ALTAIR_FORK_VERSION: Version [byte 0x01, 0x00, 0x00, 0x00],
-    ALTAIR_FORK_EPOCH: Epoch(74240), # Oct 27, 2021, 10:56:23am UTC
+    ALTAIR_FORK_EPOCH: Epoch(uint64.high), # Oct 27, 2021, 10:56:23am UTC
```
etan-status:
```suggestion
    ALTAIR_FORK_EPOCH: Epoch(uint64.high),
```

beacon_chain/networking/network_metadata.nim
arnetheduck: importing `network_metadata` adds 10:s of seconds to compile time and 10:s of mb to binary size

beacon_chain/spec/presets.nim
arnetheduck: comment

beacon_chain/networking/network_metadata.nim
zah: This PR is not changing any imports though. Also, Nim can do the run-time config expression evaluation without spilling a run-time constant for the genesis data.  I'll take a look at the generated code to confirm.

arnetheduck: well, the comment recommends calling `getMetadataForNetwork(eth2Network.get).cfg` - this is a very expensive way to get a `RuntimeConfig` instance because it first creates a `Eth2NetworkMetadata`

zah: OK, this particular recommendation was removed

arnetheduck:
The windows test failure here looks consistent, so it should be looked into before merging

zah:
The failure was tracked down to the fact that we were not using the latest Nim 1.6 fixes because the respective branch in the Status Nim repo was stale.

arnetheduck (approved):

PR #4149, opened 2022-09-19
Support Prysm and Ethdo Keystores (Fixes #4107)

### LINKED ISSUE #4107: Nimbus cannot import a keystore exported from Prysm
**Describe the bug**

`The pubkey was formatted wrong. Error: Invalid keystore objects found`

Possibly because of `pbkdf2` instead of `scrypt`?

```
{
        "crypto": {
                "checksum": {
                        "function": "sha256",
                        "message": "54fc80f6d0676bdae7c968e0d462f90a4e3a028fc7669ef8527e2f74386c9b36",
                        "params": {}
                },
                "cipher": {
                        "function": "aes-128-ctr",
                        "message": "3c2540f69cbe7e66c0c4a6e416e99bf0d1056399c21b4c45552561da920871fa",
                        "params": {
                                "iv": "98a15bd46d258aceecaeeab25bddf5e2"
                        }
                },
                "kdf": {
                        "function": "pbkdf2",
                        "message": "",
                        "params": {
                                "c": 262144,
                                "dklen": 32,
                                "prf": "hmac-sha256",
                                "salt": "c0abbbbda36e588824865a71b5b34d5a95335fe1077c286d4e9c844f7193c62b"
                        }
                }
        },
        "uuid": "39796eb1-2e43-4353-9f13-5211c7ddc58c",
        "pubkey": "8ed78a5495b54d5b6cc8bf170534ecb633b9694fba121ca680744fa9633f1b67cc77c045f88a6f97be781fe6c2867646",
        "version": 4,
        "description": "keystore",
        "path": ""
}
```

zah:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2034,100 +2034,23 @@ proc readValue*(reader: var JsonReader[RestJson], value: var ScryptParams) {.
...
-  writer.writeField("crypto", value.crypto)
-  if value.description.isSome:
-    writer.writeField("description", value.description.get)
-  writer.writeField("pubkey", value.pubkey)
-  writer.writeField("path", string(value.path))
-  writer.writeField("uuid", value.uuid)
-  writer.writeField("version", JsonString(
-    Base10.toString(uint64(value.version))))
-  writer.endRecord()
+     error: "keystores must be converted to json with Json.encode(keystore). " &
```
arnetheduck, line 2037: we use the "default" json serialization for logging - we should use a dedicated standard serializer for the keystore standard, which may or may not match our logging json serializer

arnetheduck: ie the correct way to do this is to introduce a `KeystoreJson` that implements the keystore standard, whatever that may be - we cannot keep on using `Json`, this prevents migrating nim-json-rpc to json-ser for example

zah: While this would be a theoretical improvement, the current state of affairs is that `Json.encode(keystore)` is a correct implementation of the standard, while the removed functions here were not.

arnetheduck: well, architecturally, this PR is a regression because it moves further away from dedicated serializers - we _could_ do the right thing here and introduce the KeystoreJson already - otherwise, it deserves documenting at least, because it's a fragile setup causing easily preventable bugs

arnetheduck: in particular, the outright recommendation to use `Json.encode` is simply poor advice - it's one hack/workaround replacing another

zah: In case that's not clear, we have been using `Json.decode` for keystores all along. The encoding/decoding routines were defined before the support for flavours were added to the JSON library and they are tested in the `test_keystore.nim` module. This PR is certainly not a regression as it removes an unnecessary duplicated definition that should not have been created at all. It also addresses a very specific issue, encountered by our end users.

I don't disagree regarding introducing `KeystoreJson`, but this can happen in another PR.

arnetheduck (approved):

PR #4152, opened 2022-09-19
Don't search for the TTD block after the merge

zah:


tersec:
In general, https://github.com/status-im/nimbus-eth2/pull/4144 and this PR will have a bunch of basically trivial merge conflicts

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -644,7 +645,8 @@ proc init*(T: type BeaconNode,
       eth1Network,
       config.web3ForcePolling,
       optJwtSecret,
-      config.requireEngineAPI)
+      config.requireEngineAPI,
+      ttdReached = not dag.loadExecutionBlockRoot(dag.head).isZero)
```
etan-status: may be optimistic / not a valid payload

zah: That's a good point. I've switched to `finalizedHead` as a safer option.

PR #4154, opened 2022-09-19
REST API: add ssz encoding for publishBlock

cheatfate:
https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock implementation for SSZ encoded blocks.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -769,14 +769,18 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
       block:
         if contentBody.isNone():
           return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
-        let body = contentBody.get()
-        let res = decodeBody(RestPublishedSignedBeaconBlock, body)
-        if res.isErr():
-          return RestApiResponse.jsonError(Http400, InvalidBlockObjectError,
-                                           $res.error())
-        var forked = ForkedSignedBeaconBlock(res.get())
+        let
+          body = contentBody.get()
+          version = request.headers.getString("eth-consensus-version")
+        var
+          restBlock = decodeBody(RestPublishedSignedBeaconBlock, body,
+                                 version).valueOr:
+            return RestApiResponse.jsonError(Http400, InvalidBlockObjectError,
+                                             $error)
+          forked = ForkedSignedBeaconBlock(restBlock)
+
         if forked.kind != node.dag.cfg.blockForkAtEpoch(
-            getForkedBlockField(forked, slot).epoch):
+             getForkedBlockField(forked, slot).epoch):
           return RestApiResponse.jsonError(Http400, InvalidBlockObjectError)
 
         withBlck(forked):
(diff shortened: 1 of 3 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -769,14 +769,18 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+          restBlock = decodeBody(RestPublishedSignedBeaconBlock, body,
+                                 version).valueOr:
+            return RestApiResponse.jsonError(Http400, InvalidBlockObjectError,
+                                             $error)
+          forked = ForkedSignedBeaconBlock(restBlock)
+
         if forked.kind != node.dag.cfg.blockForkAtEpoch(
-            getForkedBlockField(forked, slot).epoch):
+             getForkedBlockField(forked, slot).epoch):
           return RestApiResponse.jsonError(Http400, InvalidBlockObjectError)
```
etan-status, line 784: Hmm, I think this check should be moved to `gossip_validation.validateBeaconBlock` (so that it applies to all sources of blocks). It's also a new check that wasn't present in the pure JSON version.

beacon_chain/spec/eth2_apis/rest_beacon_calls.nim
```diff
@@ -116,6 +116,42 @@ proc publishBlock*(body: bellatrix.SignedBeaconBlock): RestPlainResponse {.
...
+       client: RestClientRef,
+       blck: bellatrix.SignedBeaconBlock
+     ): Future[RestPlainResponse] {.async.} =
+  ## https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock
+  let
+    consensus = BeaconBlockFork.Bellatrix.toString()
+    resp = await client.publishBlock(
+      blck, restContentType = $OctetStreamMediaType,
+      extraHeaders = @[("eth-consensus-version", consensus)])
+  return resp
```
etan-status:
```suggestion
proc publishSszBlock*(
       client: RestClientRef,
       blck: ForkySignedBeaconBlock
     ): Future[RestPlainResponse] {.async.} =
  ## https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock
  let
    consensus = typeof(blck).toFork.toString()
    resp = await client.publishBlock(
      blck, restContentType = $OctetStreamMediaType,
      extraHeaders = @[("eth-consensus-version", consensus)])
  return resp
```

etan-status (approved):

beacon_chain/rpc/rest_beacon_api.nim
cheatfate: nope it was there too.

arnetheduck: gossip checks this implicitly via signature check, if nothing else

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2321,6 +2323,64 @@ proc parseRoot(value: string): Result[Eth2Digest, cstring] =
...
+              data = string.fromBytes(body.data)
+        return err("Unable to deserialize data")
+      except CatchableError:
+        return err("Unexpected deserialization error")
+    ok(data)
+  elif body.contentType == OctetStreamMediaType:
+    let blockFork = ? BeaconBlockFork.decodeString(version)
+    case blockFork
+    of BeaconBlockFork.Phase0:
+      if len(body.data) < sizeof(phase0.SignedBeaconBlock):
```
arnetheduck: this check is inaccurate / redundant: the SSZ data size does not match its nim representation in-memory size (nim sequences are bigger and the objects contain dontserialize fields for example) - SSZ.decode already detects too-short data on its own, so the check should be removed

arnetheduck (approved):

PR #4155, opened 2022-09-20
add nim commit to the gauge

narimiran:
I was wrong in https://github.com/status-im/nimbus-eth2/pull/3937 — Nim commit _is_ useful :)

This also fixes `shortNimBanner` func, which was wrong before — it never produced `gitHash` because `nimBanner` only contains the line with version info.
Extracted that logic into `getNimGitHash`, so it could be use separately from `shortNimBanner` (which now correctly produces e.g. `Nim Compiler Version 1.6.7 [Linux: amd64] (09d85d8b)`)

And per @jakubgs's request, `nim_version` is now lowercase :)

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -116,8 +116,8 @@ declareGauge next_action_wait,
 declareGauge versionGauge, "Nimbus version info (as metric labels)", ["version", "commit"], name = "version"
 versionGauge.set(1, labelValues=[fullVersionStr, gitRevision])
 
-declareGauge nimVersionGauge, "Nim version info", ["nim_version"], name = "Nim_version"
-nimVersionGauge.set(1, labelValues=[NimVersion])
+declareGauge nimVersionGauge, "Nim version info", ["version", "nim_commit"], name = "nim_version"
+nimVersionGauge.set(1, labelValues=[NimVersion, getNimGitHash()])
 
 logScope: topics = "beacnde"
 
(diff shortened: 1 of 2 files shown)


narimiran:
Also, is Nim commit hash something that would be useful to have in the notice when starting beacon node, or is the current situation, where just `NimVersion` is printed, enough?

e.g. To have instead of `version=v22.7.0-52b32c-stateofus nimVersion=1.2.16 enr=(...)` something like `version=v22.7.0-52b32c-stateofus nimVersion=1.2.16-09d85t8b enr=(...)`

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -116,8 +116,8 @@ declareGauge next_action_wait,
 declareGauge versionGauge, "Nimbus version info (as metric labels)", ["version", "commit"], name = "version"
 versionGauge.set(1, labelValues=[fullVersionStr, gitRevision])
 
-declareGauge nimVersionGauge, "Nim version info", ["nim_version"], name = "Nim_version"
-nimVersionGauge.set(1, labelValues=[NimVersion])
+declareGauge nimVersionGauge, "Nim version info", ["version", "nim_commit"], name = "nim_version"
```
jakubgs, line 119: Did you really have to add the `nim_` prefix to the label? Seems redundant considering the metric name.

zah: Yes, the labels could have been `version` and `revision` (or `rev` for short), but where are these labels displayed precisely? Is changing the label name a breaking change in Prometheus/Grafana?

jakubgs: Well, I don't believe we use this anywhere yet, so not really.
But if they were used it's just a matter of adjusting the label for the graph query.

PR #4159, opened 2022-09-22
local_launch_testnet: use -SIGKILL instead of -9

jakubgs:
Moving the `-f` flag after the signal flag will resolve this error:
```
pkill: illegal option -- 9
usage: pkill [-signal] [-ILfilnovx] [-F pidfile] [-G gid]
             [-P ppid] [-U uid] [-g pgrp] [-t tty] [-u euid]
             pattern ...
```
Also using `-SIGKILL` is more explicit than `-9`.

### DIFF
--- a/scripts/launch_local_testnet.sh
+++ b/scripts/launch_local_testnet.sh
@@ -390,7 +390,7 @@ kill_by_port() {
       echo -n "Found old process listening on port ${PORT}, with PID ${PID}. "
       if [[ "${KILL_OLD_PROCESSES}" == "1" ]]; then
         echo "Killing it."
-        kill -9 "${PID}" || true
+        kill -SIGKILL "${PID}" || true
       else
         echo "Aborting."
         exit 1
@@ -645,7 +645,7 @@ cleanup() {
 
   for proc in "${PROCS_TO_KILL[@]}"
   do
-    pkill -f -9 -P $$ "${proc}" || true
+    pkill -SIGKILL -f -P $$ "${proc}" || true
   done
 
   # Delete all binaries we just built, because these are unusable outside this


etan-status:
Not sure if it works with `-signal 9`. If `signal` is meant to be replaced with the actual name, `killall -KILL`, or `killall -SIGKILL` both should work.

jakubgs:
What the fuck:
```
admin@macm1-01.ms-eu-dublin.ci.devel:~ % sudo pkill -KILL -f -l 'sleep 100'
kill -9 84664
admin@macm1-01.ms-eu-dublin.ci.devel:~ % sudo pkill -9 -f -l 'sleep 100'   
kill -9 84675
admin@macm1-01.ms-eu-dublin.ci.devel:~ % sudo pkill -signal 9 -f -l 'sleep 100'
pkill: illegal option -- s
usage: pkill [-signal] [-ILfilnovx] [-F pidfile] [-G gid]
             [-P ppid] [-U uid] [-g pgrp] [-t tty] [-u euid]
             pattern ...
```
Then why the fuck did we get:
```
pkill: illegal option -- 9
```
In the logs?
https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Fmacos%2Faarch64/detail/PR-4157/1/pipeline/90

etan-status:
https://ss64.com/osx/pkill.html
https://ss64.com/osx/kill.html

From the man page, numeric should also work (and `killall -9 geth`, for example, works on my machine). Maybe a recent addition though, as I'm on macOS 12.6.

etan-status:
Wait, why do we do `pkill` in the bottom, and regular `kill` at the top?!

jakubgs:
Don't ask me, I didn't write this monstrosity.

etan-status:
Oh, from `pkill` man:

```
     -signal           A non-negative decimal number or symbolic signal name
                       specifying the signal to be sent instead of the default
                       TERM.  This option is valid only when given as the first
                       argument to pkill.
```

So, the problem was that `-f` was before `-9`.

jakubgs:
Oooh, well spotted. Then this should fix it.

Also, terrible command line flag parsing...

tersec (approved):
Odd distinctions between `kill` and `pkill` use aside, this at least addresses the observed issue in what should be an at least as portable way.

jakubgs:
I can't find the errors in the minimal testnet run logs:
https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Fmacos%2Faarch64/detail/PR-4159/2/pipeline/72
Seems to work.

etan-status (approved):

PR #4161, opened 2022-09-22
fix check for attaching to pre-finalized parent

etan-status:
When the BN's head is reorged while shut down, reloading the BN will not assign `BlockRef` to alternate branches. However, blocks from other branches are still present in the database, leading to their descendants incorrectly marked as `UnviableFork`. By restricting the check to blocks that have been finalized, they should be reported as `MissingParent` instead, eventually re-assigning a `BlockRef` to them.


tersec (approved):

arnetheduck:
this looks like it could use a test case

greenertux:
Problem is not fixed with v22.10. If I read correctly, this fix has not been included in that release.

PR #4170, opened 2022-09-23
rm unused `dump` function for `KeystoresAndSlashingProtection`

etan-status:
In `eth2_rest_serialization` there was a `dump` function for `KeystoresAndSlashingProtection` that does not seem to be used. Removes that unused function.


etan-status:
image

tersec:
One usual approach here is to mark it as `{.deprecated.}`, and see if a warning shows up. It does look like this isn't used, and even if it is, `dump` functions' overloads' correct selections generally aren't critical to the correctness of Nimbus, but there are cases where Nim does choose oddly.

tersec (approved):
I agree, it's not used anywhere in `nimbus-eth2`.

PR #4171, opened 2022-09-23
add `sszResponsePlain` for pre-encoded SSZ data

etan-status:
For pre-encoded JSON REST responses we have `jsonResponsePlain`. Adds a `sszResponsePlain` function to serve similar purpose for SSZ. This avoids caller having to explicitly specify Http200 and media type.


etan-status:
```
RestKeyValueTuple* = tuple[key: string, value: string]
```
(presto/common.nim)

PR #4172, opened 2022-09-23
rm unused `GraffitiBytes` reader

etan-status:
All REST `readValue` parsers have the signature:
```nim
proc readValue*(reader: var JsonReader[RestJson], value: var T) {.
     raises: [IOError, SerializationError, Defect].} =
```

Only for `GraffitiBytes`, there also exists an overload with signature:
```nim
proc readValue*(reader: var JsonReader[RestJson], T: type GraffitiBytes): T
               {.raises: [IOError, SerializationError, Defect].}
```

This additional overload seems unused, and also incorrect for REST, as REST always sends Graffiti as a hex string. `GraffitiBytes.init` is used for other purposes such as command line parsing.

Removes the additional `GraffitiBytes` overload.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -799,13 +799,6 @@ proc writeValue*(writer: var JsonWriter[RestJson], value: GraffitiBytes)
                 {.raises: [IOError, Defect].} =
   writeValue(writer, hexOriginal(distinctBase(value)))
 
-proc readValue*(reader: var JsonReader[RestJson], T: type GraffitiBytes): T
-               {.raises: [IOError, SerializationError, Defect].} =
-  try:
-    init(GraffitiBytes, reader.readValue(string))
-  except ValueError as err:
-    reader.raiseUnexpectedValue err.msg
-
 ## Version
 proc readValue*(
     reader: var JsonReader[RestJson],


tersec:
Similar question/prompt as other PR -- does `{.deprecated: "foo".}` on each confirm this, across all of `make LOG_LEVEL=TRACE` and `make LOG_LEVEL=TRACE test`?

tersec:
This is used; with `deprecated`:
```
beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim(1017, 30) Warning: readValue is deprecated [Deprecated]
beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim(1314, 22) template/generic instantiation of `decode` from here
vendor/nim-serialization/serialization.nim(48, 13) template/generic instantiation of `readValue` from here
vendor/nim-serialization/serialization.nim(29, 9) template/generic instantiation of `readValue` from here
vendor/nim-json-serialization/json_serialization/reader.nim(618, 24) template/generic instantiation of `fieldReadersTable` from here
vendor/nim-serialization/serialization/object_serialization.nim(247, 34) template/generic instantiation of `makeFieldReadersTable` from here
vendor/nim-serialization/serialization/object_serialization.nim(208, 26) template/generic instantiation of `enumAllSerializedFields` from here
vendor/nim-serialization/serialization/object_serialization.nim(151, 32) template/generic instantiation of `enumAllSerializedFieldsImpl` from here
vendor/nim-serialization/serialization/object_serialization.nim(224, 62) template/generic instantiation of `readFieldIMPL` from here
vendor/nim-serialization/serialization.nim(29, 9) template/generic instantiation of `readValue` from here
vendor/nim-json-serialization/json_serialization/reader.nim(618, 24) template/generic instantiation of `fieldReadersTable` from here
vendor/nim-serialization/serialization/object_serialization.nim(247, 34) template/generic instantiation of `makeFieldReadersTable` from here
vendor/nim-serialization/serialization/object_serialization.nim(208, 26) template/generic instantiation of `enumAllSerializedFields` from here
vendor/nim-serialization/serialization/object_serialization.nim(151, 32) template/generic instantiation of `enumAllSerializedFieldsImpl` from here
vendor/nim-serialization/serialization/object_serialization.nim(224, 62) template/generic instantiation of `readFieldIMPL` from here
vendor/nim-serialization/serialization.nim(29, 9) template/generic instantiation of `readValue` from here
vendor/nim-json-serialization/json_serialization/reader.nim(618, 24) template/generic instantiation of `fieldReadersTable` from here
vendor/nim-serialization/serialization/object_serialization.nim(247, 34) template/generic instantiation of `makeFieldReadersTable` from here
vendor/nim-serialization/serialization/object_serialization.nim(208, 26) template/generic instantiation of `enumAllSerializedFields` from here
vendor/nim-serialization/serialization/object_serialization.nim(151, 32) template/generic instantiation of `enumAllSerializedFieldsImpl` from here
vendor/nim-serialization/serialization/object_serialization.nim(224, 62) template/generic instantiation of `readFieldIMPL` from here
```
It probably still builds because the serialization framework fails open/permissive, but this would change its behavior.

zah:
Since Nimbus has a Nimbus-specific end-point for setting the Graffiti, I think the intention of this overload is to allow the graffiti bytes to be specified there just like they are on the command line. Looks like we'll need a separate type such as `CliGraffitiBytes` to implement this properly.

arnetheduck:
`CliGraffitiBytes`

Rather than this, the right thing to do is a `CliJson` format - it's the CLI that is the "standard" here (just like `RestJson` is a separate standard, `LogJson` should be another and `Web3RpcJson` yet another - neither of these json standards should infect each other except by explicit choice)

zah:
There is no such thing as `CliJson`. We don't accept JSON on the command-line. Instead, we have two different REST API end-points that accept different representations for the Graffiti bytes. One is quite strict (how the bytes are encoded within a block), the other is more liberal (how the user may be entering them in a curl request, changing the run-time settings). The `Cli` is put there to suggest that the user may enter them in the curl request exactly in the same way as they are entered on the command-line (through the `--graffiti` flag, which expects a textual format, not JSON).

arnetheduck:
> There is no such thing

well, the point remains the same: we have a `GraffitiBytes` type that does what it should, and a number of standards: it's the job of the serialization framework to translate the standar type to the "internal" type, ergo we should have specific serializers that deal with this "conversion" without leaking into each other and without "crutch" types - we can call them `LaxSerializer` and `StrictJsonSerializer` for all that it matters, as long as it's clear which is used in which context. Having `RestJson` specialized for `CliGraffitiBytes` is not the right semantic, neither is `Json`.

etan-status:
Leaving this aside for now to avoid breaking Graffiti endpoint unintentionally.

PR #4174, opened 2022-09-25
implement v1.2.0 optimistic sync tests

tersec:
This also requires
- implementing latest valid hash support, hitherto treated as a noncritical optimization
- adding merge (not just Bellatrix) support to the test block infrastructure
- covering some additional LVH cases the EF consensus spec tests don't

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1898,9 +1898,9 @@ proc updateHead*(
         dag.finalizedHead.blck.root, stateRoot, dag.finalizedHead.slot.epoch)
       dag.onFinHappened(dag, data)
 
-proc getEarliestInvalidRoot*(
+proc getEarliestInvalidBlockRoot*(
     dag: ChainDAGRef, initialSearchRoot: Eth2Digest,
-    latestValidHash: Eth2Digest, defaultEarliestInvalidRoot: Eth2Digest):
+    latestValidHash: Eth2Digest, defaultEarliestInvalidBlockRoot: Eth2Digest):
     Eth2Digest =
   # Earliest within a chain/fork in question, per LVH definition. Intended to
   # be called with `initialRoot` as the parent of the block regarding which a
@@ -1920,14 +1920,14 @@ proc getEarliestInvalidRoot*(
   # parent of the reported invalid block
   if  curBlck.executionBlockRoot.isSome and
       curBlck.executionBlockRoot.get == latestValidHash:
-    return defaultEarliestInvalidRoot
+    return defaultEarliestInvalidBlockRoot
 
   while true:
     # This was supposed to have been either caught by the pre-loop check or the
     # parent check.
     if  curBlck.executionBlockRoot.isSome and
         curBlck.executionBlockRoot.get == latestValidHash:
-      doAssert false, "getEarliestInvalidRoot: unexpected LVH in loop body"
+      doAssert false, "getEarliestInvalidBlockRoot: unexpected LVH in loop body"
 
     if (curBlck.parent.isNil) or
        curBlck.parent.executionBlockRoot.get(latestValidHash) ==
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1929,9 +1929,7 @@ proc getEarliestInvalidRoot*(
       doAssert false, "getEarliestInvalidRoot: unexpected LVH in loop body"
 
     if (curBlck.parent.isNil) or
-       curBlck.parent.executionBlockRoot.isNone or
-       (curBlck.parent.executionBlockRoot.isSome and
-        curBlck.parent.executionBlockRoot.get == lvh):
+       curBlck.parent.executionBlockRoot.get(lvh) == lvh:
       break
     curBlck = curBlck.parent
 
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1899,8 +1899,9 @@ proc updateHead*(
       dag.onFinHappened(dag, data)
 
 proc getEarliestInvalidRoot*(
-    dag: ChainDAGRef, initialSearchRoot: Eth2Digest, lvh: Eth2Digest,
-    defaultEarliestInvalidRoot: Eth2Digest): Eth2Digest =
+    dag: ChainDAGRef, initialSearchRoot: Eth2Digest,
+    latestValidHash: Eth2Digest, defaultEarliestInvalidRoot: Eth2Digest):
+    Eth2Digest =
   # Earliest within a chain/fork in question, per LVH definition. Intended to
   # be called with `initialRoot` as the parent of the block regarding which a
   # newPayload or forkchoiceUpdated execution_status has been received as the
@@ -1918,18 +1919,19 @@ proc getEarliestInvalidRoot*(
   # Only allow this special case outside loop; it's when the LVH is the direct
   # parent of the reported invalid block
   if  curBlck.executionBlockRoot.isSome and
-      curBlck.executionBlockRoot.get == lvh:
+      curBlck.executionBlockRoot.get == latestValidHash:
     return defaultEarliestInvalidRoot
 
   while true:
     # This was supposed to have been either caught by the pre-loop check or the
     # parent check.
     if  curBlck.executionBlockRoot.isSome and
-        curBlck.executionBlockRoot.get == lvh:
+        curBlck.executionBlockRoot.get == latestValidHash:
       doAssert false, "getEarliestInvalidRoot: unexpected LVH in loop body"
 
     if (curBlck.parent.isNil) or
-       curBlck.parent.executionBlockRoot.get(lvh) == lvh:
+       curBlck.parent.executionBlockRoot.get(latestValidHash) ==
+         latestValidHash:
       break
     curBlck = curBlck.parent
 
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1898,6 +1898,45 @@ proc updateHead*(
         dag.finalizedHead.blck.root, stateRoot, dag.finalizedHead.slot.epoch)
       dag.onFinHappened(dag, data)
 
+proc getEarliestInvalidRoot*(
+    dag: ChainDAGRef, initialSearchRoot: Eth2Digest, lvh: Eth2Digest,
+    defaultEarliestInvalidRoot: Eth2Digest): Eth2Digest =
+  # Earliest within a chain/fork in question, per LVH definition. Intended to
+  # be called with `initialRoot` as the parent of the block regarding which a
+  # newPayload or forkchoiceUpdated execution_status has been received as the
+  # tests effectively require being able to access this before the BlockRef's
+  # made. Therefore, to accommodate the EF consensus spec sync tests, and the
+  # possibilities that the LVH might be an immediate parent or a more distant
+  # ancestor special-case handling of an earliest invalid root as potentially
+  # not being from this function's search, but being provided as a default by
+  # the caller with access to the block.
+  var curBlck = dag.getBlockRef(initialSearchRoot).valueOr:
+    # Being asked to traverse a chain which the DAG doesn't know about -- but
+    # that'd imply the block's otherwise invalid for CL as well as EL.
+    return static(default(Eth2Digest))
+
+  # Only allow this special case outside loop; it's when the LVH is the direct
+  # parent of the reported invalid block
+  if  curBlck.executionBlockRoot.isSome and
+      curBlck.executionBlockRoot.get == lvh:
+    return defaultEarliestInvalidRoot
+
+  while true:
+    # This was supposed to have been either caught by the pre-loop check or the
+    # parent check.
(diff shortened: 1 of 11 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1898,6 +1898,45 @@ proc updateHead*(
...
+    # This was supposed to have been either caught by the pre-loop check or the
+    # parent check.
+    if  curBlck.executionBlockRoot.isSome and
+        curBlck.executionBlockRoot.get == lvh:
+      doAssert false, "getEarliestInvalidRoot: unexpected LVH in loop body"
+
+    if (curBlck.parent.isNil) or
+       curBlck.parent.executionBlockRoot.isNone or
+       (curBlck.parent.executionBlockRoot.isSome and
+        curBlck.parent.executionBlockRoot.get == lvh):
```
```suggestion
       curBlck.parent.executionBlockRoot.get(lvh) == lvh:
```

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1898,6 +1898,45 @@ proc updateHead*(
         dag.finalizedHead.blck.root, stateRoot, dag.finalizedHead.slot.epoch)
       dag.onFinHappened(dag, data)
 
+proc getEarliestInvalidRoot*(
+    dag: ChainDAGRef, initialSearchRoot: Eth2Digest, lvh: Eth2Digest,
```
etan-status: `lvh` --> `latestValidHash` maybe? Could be confusing if revisited in a couple months.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1898,6 +1898,45 @@ proc updateHead*(
         dag.finalizedHead.blck.root, stateRoot, dag.finalizedHead.slot.epoch)
       dag.onFinHappened(dag, data)
 
+proc getEarliestInvalidRoot*(
```
etan-status: `getEarliestInvalidHashBlockRoot` or `getDescendantOfLatestValidHash` would be clearer. Or, at the very least, `getEarliestInvalidBlockRoot` to remove ambiguity with other roots e.g. state root

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1898,6 +1898,45 @@ proc updateHead*(
...
+  # tests effectively require being able to access this before the BlockRef's
+  # made. Therefore, to accommodate the EF consensus spec sync tests, and the
+  # possibilities that the LVH might be an immediate parent or a more distant
+  # ancestor special-case handling of an earliest invalid root as potentially
+  # not being from this function's search, but being provided as a default by
+  # the caller with access to the block.
+  var curBlck = dag.getBlockRef(initialSearchRoot).valueOr:
+    # Being asked to traverse a chain which the DAG doesn't know about -- but
+    # that'd imply the block's otherwise invalid for CL as well as EL.
+    return static(default(Eth2Digest))
```
etan-status, line 1917: Hmm, would it be correct to return `defaultEarliestInvalidRoot` here as well?

If yes, the return type could be changed to `Opt[Eth2Digest]`, and the caller could then do `dag.getEarliestInvalidRoot(initialSearchRoot, lvh).get(defaultEarliestInvalidRoot)`. This would avoid polluting this function with `defaultEarliestInvalidRoot`.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -228,17 +228,31 @@ proc updateExecutionClientHead(
...
+    let earliestKnownInvalidRoot =
+      if latestValidHash.isSome:
+        self.dag.getEarliestInvalidRoot(
+          newHead.blck.root, latestValidHash.get.asEth2Digest,
+          newHead.blck.root)
+      else:
+        newHead.blck.root
+
+    self.attestationPool[].forkChoice.mark_root_invalid(
+      earliestKnownInvalidRoot)
```
etan-status: Hmm. Not sure whether this is correct in situations where the EL is restarted or changed to a different EL while Nimbus is running. Or, multi-EL scenarios which may not have the same sync progress.

The previous mechanism of only marking the specific block invalid, and then re-selecting a new head, doing fcu to it and then marking it invalid/valid one at a time seems more robust, in general, while not losing correctness (just a bit of latency), especially when considering how persistent an `addUnviable` can be (requiring a restart if done so incorrectly due to intermittent EL bug for example).

We could shortcut the recovery mechanism though, by using the `getEarliestInvalidRoot` for the next fcu instead. If that is indeed reported as invalid, then we don't need all the extra fcu for the intermediate blocks. however, if it is reported as valid, it just means that the EL was not synced as far as the DAG at the time it was asked for verdict.

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -442,6 +442,7 @@ func mark_root_invalid*(self: var ForkChoice, root: Eth2Digest) =
         self.backend.proto_array.nodes.offset
     if nodePhysicalIdx < self.backend.proto_array.nodes.buf.len:
       self.backend.proto_array.nodes.buf[nodePhysicalIdx].invalid = true
+    self.backend.proto_array.propagateInvalidity(nodePhysicalIdx)
```
etan-status, line 445: Shouldn't this already be covered for purpose of scoring / selecting heads? 
`nodeLeadsToViableHead` should already be `false` if the current block is `invalid`, treating the entire branch as invalid?

beacon_chain/fork_choice/proto_array.nim
```diff
@@ -558,6 +558,28 @@ func nodeIsViableForHead(self: ProtoArray, node: ProtoNode): bool =
...
+      parentLogicalIdx = nodeParent.unsafeGet()
+      parentPhysicalIdx = parentLogicalIdx - self.nodes.offset
+
+    # Former case is orphaned, latter is invalid, but caught in score updates
+    if parentPhysicalIdx < 0 or parentPhysicalIdx >= self.nodes.len:
+      continue
+
+    # Invalidity transmits to all descendents
+    if self.nodes.buf[parentPhysicalIdx].invalid:
+      self.nodes.buf[nodePhysicalIdx].invalid = true
```
etan-status, line 581: What about new descendants of an invalid block that are being added after `propagateInvalidity` was originally called?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1898,6 +1898,45 @@ proc updateHead*(
...
+      doAssert false, "getEarliestInvalidRoot: unexpected LVH in loop body"
+
+    if (curBlck.parent.isNil) or
+       curBlck.parent.executionBlockRoot.isNone or
+       (curBlck.parent.executionBlockRoot.isSome and
+        curBlck.parent.executionBlockRoot.get == lvh):
+      break
+    curBlck = curBlck.parent
+
+  curBlck.root
```
etan-status, line 1938: This is not necessarily correct.

When someone switches from a different CL to Nimbus but retains their EL, the EL's `lvh` may be far in the future. This means that none of our DAG blocks contains `lvh`. This implementation would then return the first descendant of `dag.finalizedHead` (or `dag.finalizedHead` itself if no descendants were yet added).

If `lvh` is not found, maybe better to return the `defaultEarliestInvalidRoot` or `ZERO_HASH` instead.

etan-status: Also, in extended periods of non-finality, the linear scan may become expensive if performed repeatedly.
One way to accelerate could be to stop the scan at the latest known-VALID block, instead of all the way back at finalized. Or, also having a constant maximum search depth.

etan-status (approved):
Thanks for the added tests and for merge simulation in `testblockutil`!

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: I'm fine with keeping the previous logic in place (plus `mark_invalid_root`), yes, but then it starts diverging from the optimistic sync tests. As much as feasible, I want to keep them aligned, so what's tested is what's running. The non-`earliestInvalid` versions are in this PR mostly kept for as a fallback.

The previous mechanism still exists, too, and it's necessary in intial optimistic sync, where there can't be an LVH because there hasn't yet been a `VALID` block (here, LVH only reports actual-EL-`VALID`, not just not-`INVALIDATED`). So that recovery mechanism is still necessary.

Intermittent EL `INVALIDATED` bugs should be fixed. It seems unwise to contort Nimbus excessively to handle that "well", beyond, as it does, not persisting them so at least restarts clear them.

I'm not sure exactly what the multi-EL scenario will look like, but if they disagree with each other on this, well, that's not going to work out well.

But, I'm fine with keeping this unchanged in this PR and exploring this in a future PR, while the LVH handling remains in place to pass the EF optimistic sync tests.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4174/commits/313a35d7dcad904f7f23f0ed7d26ab386820ef0d

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4174/commits/313a35d7dcad904f7f23f0ed7d26ab386820ef0d reverts the LVH parts, while still adding `mark_root_invalid`.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: I'd want to use `Root` in some capacity here, because while it's dealing with the LVH (a hash, from the EL), what it's returning is firmly in CL-land, an SSZ root. `HashBlockRoot` is a bit redundant in that sense, and confusing the boundary between EL-"hash" and CL-"root".

But yes, state vs block root is a useful distinction to make, so `getEarliestInvalidBlockRoot` is a good change.

https://github.com/status-im/nimbus-eth2/pull/4174/commits/840d097e52d728f06914caf5f96e8ffdf45ee702

beacon_chain/fork_choice/fork_choice.nim
tersec: The updating is done in `maybeUpdateBestChildAndDescendant`
https://github.com/status-im/nimbus-eth2/blob/ad286b9aadf25c995294aa96e8c6408a747912b2/beacon_chain/fork_choice/proto_array.nim#L424-L451

Which is called by `applyScoreChanges` which iterates from descendants to ancestors
https://github.com/status-im/nimbus-eth2/blob/ad286b9aadf25c995294aa96e8c6408a747912b2/beacon_chain/fork_choice/proto_array.nim#L270-L282

Which is correct for the scores as such -- it allows quick updates. However, invalidity propagates the other direction, from ancestors to descendants, unlike the scores, and there's no existing mechanism in fork choice to propagate information from ancestors to descendants.

beacon_chain/fork_choice/proto_array.nim
tersec: In theory, they shouldn't be added in the first place, as they'd be rejected as unviable by the DAG/quarantine.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4205

In theory, latest-known `VALID` block is exactly where this will stop. If it doesn't, that's an EL bug. EL bugs shouldn't crash Nimbus, or result in durably incorrect behavior, but EL bugs can trigger suboptimally slow behavior. It's meant to be a mutually trusted system, and I'm wary of adding complication to Nimbus to handle buggy ELs, when that complication can carry its own risks .

In particular, https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md#payloadstatusv1 defines:
> `latestValidHash`: `DATA|null`, 32 Bytes - the hash of the most recent *valid* block in the branch defined by payload and its ancestors

This also addresses the concern about the EL's `lvh` being far in the future in the new-CL-database/existing-EL-database case: the `lvh` is relative to the payload provided by the CL. Here, the failure mode for buggy EL is that Nimbus needs to be restarted, which seems reasonable.

As far as cost in general, this should be a relatively one-time thing per invalid branch -- once invalidated, it shouldn't be doing that backwards search again, so potentially better to do it once than multiple times, along the lines of how pruning huge swaths of finalized blocks after a long-unfinalizing network finalizes is expensive, but one-time. Doing halfway versions introduces less-well-defined state and doesn't necessarily save time overall, in a throughput sense.

There is an issue here where the quarantine unviables are sometimes checked only as direct parents rather than ancestors, which is an argument for not using the LVH-based search for adding to quarantine.

The other aspect is that it should rediscover all of this by just asking the EL, if it misses something. So everything here should be considered a cache.

For the moment, this is all only used in tests, so another approach is to move this out from `blockchain_dag` and into `tests/`, so ensure that it doesn't accidentally get used in `blockchain_dag`.

While the initial intent, and still one that seems broadly ideal, is to maximally match test behavior with non-test behavior (otherwise, what is the test testing?), just the status quo here is that this LVH infrastructure is test-only, so it can/should reflect that.

beacon_chain/fork_choice/proto_array.nim
tersec: The other fallback here is that they'll ask the EL again if someone does try to build on them, and get that block instead. Since by then things should have mostly caught up to head, it's more likely to be the only remaining set of blocks on which building is still happening, and therefore should iteratively converge.

Issue #4162, opened 2022-09-22
Adjust logging levels so there exists something between `INF` and current `DBG`


tersec:
This logging level might, for example, be a tweaked `DBG`.

The challenge currently is that `DBG` is verbose enough to discourage people even who might prefer something more than `INF` to run it:
```
$ zcat service.20220922_1700.log.gz | jq .lvl | sort | uniq -c
 372545 "DBG"
    600 "INF"
   3313 "NOT"
      2 "WRN"
```
There are orders of magnitude more `DBG` than `INF` messages, in this basically well-functioning Sepolia node's logs.
```
$ zcat service.20220922_1700.log.gz | jq -r '[.lvl, .msg]|@csv' | sort | uniq -c | sort -rn | head
 131992 "DBG","Sync committee message resolved"
 131992 "DBG","Sync committee message received"
  39917 "DBG","Got incoming request from peer"
  12367 "DBG","Dropping message after validation, reason: ignore"
  10985 "DBG","Contribution received"
   9179 "DBG","Attestation resolved"
   9179 "DBG","Attestation received"
   4591 "DBG","Ignoring eth1 vote"
   3294 "DBG","Peer status"
   2841 "DBG","Aggregate received"
```

By pruning moving some of less useful messages to `TRC` would roughly halve the total number of logged messages.

Furthermore, in another Sepolia node, currently stuck in syncing pending https://github.com/status-im/nimbus-eth2/pull/4161
```
$ zcat service.20220922_1700.log.gz | jq -r '[.lvl, .msg]|@csv' | sort | uniq -c | sort -rn | head
  38346 "DBG","Got incoming request from peer"
  12542 "DBG","Block range request done"
   2844 "DBG","Message sent to RPC server"
   2507 "DBG","newPayload: succeeded"
   2507 "DBG","newPayload: inserting block into execution engine"
   2507 "DBG","Block parent unknown or finalized already"
   2380 "DBG","Peer status"
   1800 "DBG","Current syncing state"
   1329 "DBG","starting pubsub read loop"
   1308 "DBG","exiting pubsub read loop"
```

Sync committee messages almost always (literally 100% of the time here) resolve, so one set of potential changes is to drop one of the `Sync committee message` messages to `TRC`.

One challenge here is that `logtrace` depends on seeing both sides of `Sync committee message sent` and `Sync committee message received`, so that would have to be compromised some.

So the gossip messages just aren't there because it's behind.

[...]

PR #4175, opened 2022-09-26
adjust block resolved log levels

tersec:
https://github.com/status-im/nimbus-eth2/issues/4162

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -81,7 +81,7 @@ proc addResolvedHeadBlock(
     epochRef = dag.getEpochRef(state, cache)
     epochRefTick = Moment.now()
 
-  debug "Block resolved",
+  info "Block resolved",
     blockRoot = shortLog(blockRoot),
     blck = shortLog(trustedBlock.message),
     executionValid, heads = dag.heads.len(),


scripts/launch_local_testnet.sh
```diff
@@ -66,7 +66,7 @@ USE_HTOP="1"
 USE_VC="1"
 LIGHTHOUSE_VC_NODES="0"
 USE_GANACHE="0"
-LOG_LEVEL="DEBUG; TRACE:networking"
+LOG_LEVEL="TRACE"
```
etan-status: `TRACE` can be very spammy inside libp2p, should double-check that log sizes are roughly similar to before

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -76,7 +76,7 @@ proc addResolvedHeadBlock(
     epochRef = dag.getEpochRef(state, cache)
     epochRefTick = Moment.now()
 
-  debug "Block resolved",
+  info "Block resolved",
```
etan-status, line 84: This one is very useful for debugging, thanks!

etan-status (approved):

beacon_chain/consensus_object_pools/block_clearance.nim
arnetheduck: Hm, on sync, this will be quite noisy (ie millions of lines).. should it be limited to gossip blocks?

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -485,7 +485,7 @@ proc processSyncCommitteeMessage*(
 
   # Potential under/overflows are fine; would just create odd metrics and logs
   let delay = wallTime - syncCommitteeMsg.slot.start_beacon_time
-  debug "Sync committee message received", delay
+  trace "Sync committee message received", delay
```
arnetheduck: hm, attestations are more noisy than this - why single out the scm:s?

tersec: https://github.com/status-im/nimbus-eth2/issues/4162

tersec: They're not more noisy

scripts/launch_local_testnet.sh
tersec: Either this change occurs or `logtrace` can't trace sync committee messages, because it will only see one side of the send/receive. But it's also not something one has to immediately have working -- logtrace is currently set only to run on aggregate attestations for this, so it's fine to sacrifice this.

2022-09-26-220037_885x220_scrot
2022-09-26-215901_876x213_scrot

Before and after this PR. that's enough of a change that I'd want to address that separately, if someone does want to use `logtrace` to track sync committee messages in the local testnet. It'll be an obvious, nonsubtle, easily reproducible issue, so it can be addressed later if people think it's worth solving.

https://github.com/status-im/nimbus-eth2/pull/4175/commits/c509035f485d56ae33bc3143b9200b4edb2cbfab reverts this change.

etan-status: hmm, or add `TRACE` specifically for `beacval`?

beacon_chain/gossip_processing/eth2_processor.nim
arnetheduck: so .. with 400k validators and a fully subscribed node, we can expect, hourly:
```
>>> 512*5*60
153600
>>> 400000/32*5*60
3750000
```

ie two orders of magnitude more attestations - I'm guessing sepolia doesn't have enough validators to be representative

zah:
I find the inconsistency created here a bit hard to justify. What is the evidence that we have users reluctant to use the DEBUG log level specifically due to the volume of sync committee messages?

tersec:
> I find the inconsistency created here a bit hard to justify. What is the evidence that we have users reluctant to use the DEBUG log level specifically due to the volume of sync committee messages?

Sure, it's worth getting data from a large network, to make sure it captures scaling corrrectly.

But the volume of `DEBUG` logs, abstractly, is definitely an issue, and it's worth inconsistency to do something to address https://github.com/status-im/nimbus-eth2/issues/4162. I agree that the sync committee changes probably are not that thing.

We do need, I would argue, something that's more useful for debugging than `INFO`, but not literally `DEBUG`.

beacon_chain/gossip_processing/eth2_processor.nim
tersec: This appears to be the case, yes. https://github.com/status-im/nimbus-eth2/issues/4162#issuecomment-1269679003 shows on Prater/Goerli:
```
$ zcat service.20221006_0900.log.gz | jq -r '[.lvl, .msg]|@csv' | sort | uniq -c | sort -rn | head
2212428 "DBG","Attestation received"
2200777 "DBG","Attestation resolved"
 161811 "DBG","Aggregate received"
 130052 "DBG","Dropping message after validation, reason: ignore"
 110091 "DBG","Dropping aggregate"
  59843 "DBG","Got incoming request from peer"
  47929 "DBG","Aggregate resolved"
  47892 "DBG","Sync committee message resolved"
  47891 "DBG","Sync committee message received"
  18488 "NOT","Attestation sent"
```
Attestations dominate significantly.

tersec: https://github.com/status-im/nimbus-eth2/pull/4175/commits/a0ae8235b9018423a02a21d0ef891409e1ec730e

tersec:
Because that specific inconsistency isn't worthwhile, https://github.com/status-im/nimbus-eth2/pull/4175/commits/a0ae8235b9018423a02a21d0ef891409e1ec730e removes it

arnetheduck:
Been thinking a bit about this message, and here's a bit of a wishlist I would want from "positive" logging, I think:

* the useful part of "slot start" - ie time, basic peer connectivity stats, basic dag stats
* the useful part of "slot end" - ie next actions

I *think* this by a single "Slot updated" message which happens at the first of:

* when expected gossip block arrives and is chosen as head
* when +4s happens and head is "updated" to the new slot without a block

Apart from this, we would log "Block resolved" from gossip *and* introduce a new "Syncing" log that replaces "Slot start" but only prints when we're actually syncing.

With this logging, I expect a synced node to display 2 messages:
* Block resolved
* Slot updated

An unsynced node would similarly display 2 messages / slot:
* Syncing
* Slot updated

Potentially, we would highlight oddities with additional logging:

* missing expected block at +4s, when synced (noting the expected validator)
* Block resolved that did *not* become head
* etc

arnetheduck:
More thoughts on this and how to manage log volume and make the message useful during syncing as well:

Every slot, we wait for the expected block of that slot to arrive, or slot+4s to happen - these are the general conditions for attesting anyway.

We then break down logging the following way:

* if more than one block arrives and/or it isn't the expect block, we log a summary of received blocks
* if the expected block arrives (ie slot == wall slot and proposer == expected proposer based on previous head), we log it (regardless of the above "extra" block logging)
* if none of the 2 above were logged, we log that no block was received for a full slot
  * if we are in sync, this is the point where we start producing attestations for the previous head
  * if we are not in sync, this would indicate syncing is stuck


This is inspired by nethermind which logs something like the following when the client is synced:
```
29 Feb 12:37:50 | Received new block:  1042807 (0xd7308d...978f33) 
29 Feb 12:37:50 | Processed             1042807     |    697.06 ms  |  slot     11,309 ms | Gas gwei: 2.06 .. 2.21 (3.43) .. 48.00 
29 Feb 12:37:50 | - Block              28.34 MGas   |     67    txs |  calls  1,339 (130) | sload   3,816 | sstore    714 | create   3 
29 Feb 12:37:50 | - Block throughput   40.65 MGas/s |     96.12 t/s |          1.43 Blk/s | recv        0 | proc        0 
```

On the other hand, when it's bulk-processing blocks, it logs this instead:
```
28 Feb 21:14:14 | Processed     1038209...  1038213 |  1,723.25 ms  |  slot      1,724 ms | Gas gwei: 1.03 .. 1.34 (3.08) .. 54.00 
28 Feb 21:14:14 | - Blocks 5           75.95 MGas   |    192    txs |  calls  2,318 (  9) | sload   5,664 | sstore  2,676 | create 402(-402) 
28 Feb 21:14:14 | - Block throughput   44.07 MGas/s |    111.42 t/s |          2.90 Blk/s | recv        0 | proc       49 
```

The strategy of logging details on the expected block if it arrives and a summary of any syncing blocks gives us the best of both worlds - it also covers the case where someone sends "rogue" blocks that drop in too late and become potential but not really viable heads.

jakubgs:
This would make managing the volume of logs that got via Logstash into our ElasticSearch cluster much easier.

tersec:
- #8539

PR #4178, opened 2022-09-26
Harden EpochRef loading against bogus block root at tail

arnetheduck:
* add more error information when things go wrong with database
* lower log level when reloading attestations from no-block epoch start slot


etan-status (approved):

arnetheduck:
unrelated mac fail

PR #4180, opened 2022-09-26
newPayload `INVALIDATED` should be `unviableFork`

tersec:
This makes it more consistent with `forkchoiceUpdated` `INVALIDATED`s and also accounts for where a sequence of optimistically synced `NOT_VALIDATED` blocks are followed by an `INVALIDATED` block. In that case, all the block processor knows is that the block or one of its ancestors is invalid. It does not know that the block itself, in isolation, is invalid.

This changes the peer descoring, but in a way consistently with `forkchoiceUpdated`.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -300,7 +300,7 @@ template toGaugeValue(x: Quantity): int64 =
 func compute_time_at_slot(genesis_time: uint64, slot: Slot): uint64 =
   genesis_time + slot * SECONDS_PER_SLOT
 
-# https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.1/specs/phase0/validator.md#get_eth1_data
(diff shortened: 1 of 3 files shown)


etan-status (approved):
This aligns the behaviour with the flow where `newPayload` returns `SYNCING/ACCEPTED`, and then `forkchoiceUpdated` returns `INVALID`, leading to block being marked as unviable for future occurrences.

tersec:
> This aligns the behaviour with the flow where `newPayload` returns `SYNCING/ACCEPTED`, and then `forkchoiceUpdated` returns `INVALID`, leading to block being marked as unviable for future occurrences.

It's marked unviable either way, already. For `newPayload`:
https://github.com/status-im/nimbus-eth2/blob/9750cd3a38e3f971541affb2c52ff7a8baeeeaae/beacon_chain/gossip_processing/block_processor.nim#L558-L559

And for `forkchoiceUpdated`:
https://github.com/status-im/nimbus-eth2/blob/9750cd3a38e3f971541affb2c52ff7a8baeeeaae/beacon_chain/consensus_object_pools/consensus_manager.nim#L240-L241

This just changes the scoring.

PR #4184, opened 2022-09-27
don't verify BLS signatures in produced blocks

tersec:
This is time-consuming and, since blocks are verified anyway, superfluous.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -532,7 +532,11 @@ proc makeBeaconBlockForHeadAndSlot*(
       effectiveExecutionPayload,
       noRollback, # Temporary state - no need for rollback
       cache,
-      verificationFlags = if skip_randao_verification_bool: {skipRandaoVerification} else: {},
+      verificationFlags =
+        if skip_randao_verification_bool:
+          {skipBlsValidation, skipRandaoVerification}
+        else:
+          {skipBlsValidation},
       transactions_root =
         if transactions_root.isSome:
           Opt.some transactions_root.get


beacon_chain/validators/validator_duties.nim
```diff
@@ -532,7 +532,11 @@ proc makeBeaconBlockForHeadAndSlot*(
       effectiveExecutionPayload,
       noRollback, # Temporary state - no need for rollback
       cache,
-      verificationFlags = if skip_randao_verification_bool: {skipRandaoVerification} else: {},
+      verificationFlags =
+        if skip_randao_verification_bool:
+          {skipBlsValidation, skipRandaoVerification}
+        else:
+          {skipBlsValidation},
```
arnetheduck, line 539: we still need to validate the randao - unlike attestations et al, it hasn't gone through sig-checks at all (when coming from rest)

tersec: This only avoids validating RANDAO when `skip_randao_verification_bool`, as before

arnetheduck: https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/spec/state_transition_block.nim#L94

arnetheduck: also, I wonder if we should use a `SigVerified` cast instead which would generate better code

tersec: > This only avoids validating RANDAO when `skip_randao_verification_bool`, as before

> also, I wonder if we should use a `SigVerified` cast instead which would generate better code

https://github.com/status-im/nimbus-eth2/pull/4187

PR #4185, opened 2022-09-27
Build block in parallel with attestation packing

arnetheduck:
* fix block proposal in first slot after checkpoint


arnetheduck:
will trivially conflict with #4184, will fix once that one is merged

tersec (approved):

Issue #4181, opened 2022-09-26
Cannot create non-wall-slot block proposal


arnetheduck:
Here we see an attempt to create a block for `3975925` while head is at `3975927` - this should not be an invalid operation given that 25 has not finalized yet:

```
INF 2022-09-26 21:05:31.134+02:00 State replayed                             topics="chaindag" blocks=15 slots=18 current=94282677:3975927 ancestor=a923334c:3975903@3975904 target=fabbf676:3975922 ancestorStateRoot=08822ff0 targetStateRoot=5403b0d6 found=false assignDur=47ms478us146ns replayDur=608ms552us126ns
ERR 2022-09-26 21:05:31.146+02:00 Engine API fork-choice update failed       topics="beacval" err="{\"code\":-38003,\"message\":\"Invalid payload attributes\",\"data\":{\"err\":\"invalid timestamp, parent 1664219124 given 1664219076\"}}" finalized_block_hash=1df2bf99 head_block_hash=07b54e4f
INF 2022-09-26 21:05:31.172+02:00 getExecutionPayload: newExecutionPayload invalid topics="beacval" executionPayloadStatus=ok(INVALID_BLOCK_HASH)
WRN 2022-09-26 21:05:31.173+02:00 Unable to get execution payload. Skipping block proposal topics="beacval" slot=3975923 validator_index=271550
```

321c87279050d6f4f2014f2c81731bb4835cd26d - this results from a call to the proposer rest api:

http://localhost:5052/eth/v2/validator/blocks/3975925?randao_reveal=0xc00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000&skip_randao_verification

tersec:
The EL is apparently rejecting this block. The fix, such as it is, would be to remove
https://github.com/status-im/nimbus-eth2/blob/ad286b9aadf25c995294aa96e8c6408a747912b2/beacon_chain/validators/validator_duties.nim#L424-L429

However, it was added initially because otherwise back-to-back blocks being proposed by the same node weren't being registered with the EL, i.e. the first block was proposed, but the EL never learned about its payload via any other means from the CL (and EIP-3675 prohibits post-merge EL block gossip), so the second block was not successfully produced, either not at all or with incorrect EL-parent-block.

More recently, it's been a useful safeguard that the empty execution payload blocks are verified to be actually correct before being sent out.

[...]

PR #4186, opened 2022-09-27
remove `newPayload` from block production flow

tersec:
https://github.com/status-im/nimbus-eth2/issues/4181#issuecomment-1258568058 has more context.

Upshot is that at this point it's a safeguard which probably costs more than is worth.

This specifically requires testing in situations which trigger multiple consecutive block proposals by the same node.


arnetheduck (approved):

arnetheduck:
So this still needs the update to `BlockProcessor`?

tersec:
> So this still needs the update to `BlockProcessor`?

Yes, and that couldn't happen (sans rebasing/merge conflicts) until https://github.com/status-im/nimbus-eth2/pull/4192 was merged, because it modifies that part of `BlockProcessor` too, so the rebasing here to `unstable` was to pull that in pre-emptively.

tersec:
It would also be preferable to merge https://github.com/status-im/nimbus-eth2/pull/4182 before adding `BlockProcessor` changes to this -- it's cleanly mergeable at the moment, but it becomes messy if it waits, since it's going to affect the whole module.

https://github.com/status-im/nimbus-eth2/pull/4182/files#diff-a09c99ea4dc7b4118405b5b5c2985124f3ac10247f3d69c9611c22ca7c496f48

tersec:
> So this still needs the update to `BlockProcessor`?

https://github.com/status-im/nimbus-eth2/pull/4186/commits/03aba9dddff1e48a815cd936c85fbbb0f31132ca

This has some tradeoffs, but I'm not sure they're avoidable:
- it means `processBlock` and `storeBlock` in `block_processor` are async. In theory, this shouldn't matter until it reaches the first yield point, in terms of compounding scheduling latency upon scheduling latency, but that wasn't even a possibility before;

- it relies slightly more on the guarantee that `BlockProcessor` will be a singleton than before, because as a result of this `async` transformation, it can't use `var BlockProcessor` but rather `ref BlockProcessor`. In theory harmless, in practice a slightly increased risk;

- somewhat more avoidable, but more a case of it not making sense to return to outer layers of callers in the REST and `validator_duties` layers which have no good mechanism for this block rescheduling that `block_processor` can do, it complicates the `BlockError`-based `storeBlock` return type somewhat to signal the whether to continue to effectively loop on that block. This should at least replicate existing semantics: loop on the block in a similar way, and, albeit differently (before, `validator_duties` detected `newPayload` failures more directly, treat failure of that first `newPayload` as an error. There's probably not time to try again, in terms of retries, because that first `newPayload` can already take up to 8 seconds per specification to time out. If it hasn't worked by then, `validator_duties` needs to move on regardless; and

- as a more philosophical difference, it takes the `storeBlock` notion more seriously than before of validating, always, without the fast path it used to have of being able to specify from the outside to `storeBlock` that the EL block/execution payload was presumed `VALID`. This is arguably more correct, but depends on one's views of how much to the REST API is supposed to be able to provide arbitrary input, and also increases the baseline latency of `storeBlock`.

It's probably an overall net improvement, but definitely comes with its ugliness. Because this `newPayload` still happens within `storeBlock`, `storeBlock` time as a whole won't decrease, and might increase, but the actual block broadcast does happen first regardless.

PR #4187, opened 2022-09-27
re-enable randao checks

tersec:
followup to https://github.com/status-im/nimbus-eth2/pull/4184

### DIFF
--- a/beacon_chain/spec/state_transition.nim
+++ b/beacon_chain/spec/state_transition.nim
@@ -422,6 +422,7 @@ proc makeBeaconBlock*(
                                 randao_reveal, eth1_data, graffiti, attestations, deposits,
                                 exits, sync_aggregate, execution_payload)
 
+  # Signatures are verified elsewhere, so don't duplicate inefficiently here
   let res = process_block(
     cfg, state.data, blck.asSigVerified(), verificationFlags, cache)
 
--- a/beacon_chain/spec/state_transition.nim
+++ b/beacon_chain/spec/state_transition.nim
@@ -353,8 +353,7 @@ proc makeBeaconBlock*(
                                 exits, sync_aggregate, execution_payload)
 
   let res = process_block(
-    cfg, state.data, isomorphicCast[phase0.SigVerifiedBeaconBlock](blck),
-    verificationFlags, cache)
+    cfg, state.data, blck.asSigVerified(), verificationFlags, cache)
 
   if res.isErr:
     rollback(state)
@@ -424,8 +423,7 @@ proc makeBeaconBlock*(
                                 exits, sync_aggregate, execution_payload)
 
   let res = process_block(
-    cfg, state.data, isomorphicCast[altair.SigVerifiedBeaconBlock](blck),
-    verificationFlags, cache)
+    cfg, state.data, blck.asSigVerified(), verificationFlags, cache)
 
   if res.isErr:
     rollback(state)
@@ -496,8 +494,7 @@ proc makeBeaconBlock*(
                                 exits, sync_aggregate, execution_payload)
 
   let res = process_block(
-    cfg, state.data, isomorphicCast[bellatrix.SigVerifiedBeaconBlock](blck),
-    verificationFlags, cache)
+    cfg, state.data, blck.asSigVerified(), verificationFlags, cache)
 
   if res.isErr:
     rollback(state)
--- a/beacon_chain/spec/state_transition.nim
+++ b/beacon_chain/spec/state_transition.nim
@@ -352,7 +352,9 @@ proc makeBeaconBlock*(
                                 randao_reveal, eth1_data, graffiti, attestations, deposits,
                                 exits, sync_aggregate, execution_payload)
 
-  let res = process_block(cfg, state.data, blck, verificationFlags, cache)
+  let res = process_block(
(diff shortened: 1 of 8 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -555,6 +554,20 @@ proc makeBeaconBlockForHeadAndSlot*(
...
+
+    # Verify RANDAO, since block production otherwise doesn't verify signatures
+    if not skip_randao_verification_bool:
+      let proposer_pubkey = node.dag.validatorKey(validator_index)
+
+      if proposer_pubkey.isNone:
+        return ForkedBlockResult.err("makeBeaconBlockForHeadAndSlot: unable to find pubkey for validator index")
+
+      withStateAndBlck(state, res.get):
+        if not verify_epoch_signature(
```
tersec: This is a very literal/direct translation from the state transition function, to mitigate risk, thus the `withStateAndBlck`.

beacon_chain/spec/state_transition.nim
```diff
@@ -352,7 +352,9 @@ proc makeBeaconBlock*(
                                 randao_reveal, eth1_data, graffiti, attestations, deposits,
                                 exits, sync_aggregate, execution_payload)
 
-  let res = process_block(cfg, state.data, blck, verificationFlags, cache)
+  let res = process_block(
+    cfg, state.data, isomorphicCast[phase0.SigVerifiedBeaconBlock](blck),
```
arnetheduck: `asSigVerified`?

arnetheduck: ideally we would move the cast to `makeBeaconBlockForHeadAndSlot` since that is where the sig checking is actually done

beacon_chain/validators/validator_duties.nim
```diff
@@ -555,6 +554,20 @@ proc makeBeaconBlockForHeadAndSlot*(
       error "Cannot create block for proposal",
         slot, head = shortLog(head), error = res.error()
       return err($res.error)
+
+    # Verify RANDAO, since block production otherwise doesn't verify signatures
+    if not skip_randao_verification_bool:
```
arnetheduck: when true, must be `ValidatorSig.infinity`

tersec:
```
Wrote local_testnet_data/bootstrap_nodes.txt
Wrote local_testnet_data/config.yaml:
PRESET_BASE: minimal
MIN_GENESIS_ACTIVE_VALIDATOR_COUNT: 128
MIN_GENESIS_TIME: 0
GENESIS_DELAY: 10
DEPOSIT_CONTRACT_ADDRESS: 0x0000000000000000000000000000000000000000
ETH1_FOLLOW_DISTANCE: 1
ALTAIR_FORK_EPOCH: 1
BELLATRIX_FORK_EPOCH: 2
TERMINAL_TOTAL_DIFFICULTY: 0
Waiting for Altair finalization
Altair finalized, launching 1 light client(s)
That was run #1
```

beacon_chain/validators/validator_duties.nim
tersec: handled already by passing `skip_randao_verification_bool`
https://github.com/status-im/nimbus-eth2/blob/7a444769d5a8a8a77c165e21150fbf7d245f00ce/beacon_chain/validators/validator_duties.nim#L534-L537

beacon_chain/spec/state_transition.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4187/commits/a2618dd5a1ff1f5924e165777a0fc72fbf090450

tersec: > ideally we would move the cast to `makeBeaconBlockForHeadAndSlot` since that is where the sig checking is actually done

there's no `ForkedSigVerifiedBeaconBlock` or similar. Of course it could be created, but this adds enough infrastructure for a last-minute PR

tersec: Also, `state_transition.makeBeaconBlock` is what runs `process_block`, and `process_block` is what does all those signature verifications. So it has to be one of the trusted/sigverified types within `state_transition.makeBeaconBlock`

On the other side of that, if `makeBeaconBlockForHeadAndSlot` returns anything except what it does, that needs changes beyond the scope of what should be in this PR too, because it's used by the VC, etc.

So there aren't that many options.

arnetheduck (approved):

beacon_chain/spec/state_transition.nim
```diff
@@ -421,7 +422,8 @@ proc makeBeaconBlock*(
                                 randao_reveal, eth1_data, graffiti, attestations, deposits,
                                 exits, sync_aggregate, execution_payload)
 
-  let res = process_block(cfg, state.data, blck, verificationFlags, cache)
+  let res = process_block(
+    cfg, state.data, blck.asSigVerified(), verificationFlags, cache)
```
arnetheduck, line 427: should probably document why we consider asSigVerified safe (ie because  the sigs have been verified elsewhere)

tersec: https://github.com/status-im/nimbus-eth2/pull/4187/commits/7a9bffba0f7d52b2fc9f40ddf9fd38f58496b692

PR #4189, opened 2022-09-27
Fix REST generic error parsing.

cheatfate:
Fix REST generic error parser and add ability to decode `code` field as number of string.
Add tests.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -328,13 +328,11 @@ proc jsonMsgResponse*(t: typedesc[RestApiResponse],
     block:
       var default: seq[byte]
       try:
-        var defstrings: seq[string]
         var stream = memoryOutput()
         var writer = JsonWriter[RestJson].init(stream)
         writer.beginRecord()
-        writer.writeField("code", "200")
+        writer.writeField("code", 200)
         writer.writeField("message", msg)
-        writer.writeField("stacktraces", defstrings)
         writer.endRecord()
         stream.getOutput(seq[byte])
       except SerializationError:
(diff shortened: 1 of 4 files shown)


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2235,6 +2235,59 @@ proc writeValue*(writer: var JsonWriter[RestJson],
...
+      let ires =
+        try:
+          let res = reader.readValue(int)
+          if res < 0:
+            reader.raiseUnexpectedValue("Invalid `code` field value")
+          Opt.some(uint64(res))
+        except SerializationError:
+          Opt.none(uint64)
+      if ires.isNone():
+        let sres = Base10.decode(uint64, reader.readValue(string)).valueOr:
```
arnetheduck, line 2247: why do we need to decode as string? https://github.com/ethereum/beacon-APIs/blob/master/types/http.yaml#L86 says it should always be a json number

cheatfate: I dont see any reasons to be extremely strict as specification said. Specification said many things which are not implemented in different REST servers (for example limitation of 30 validator keys for arrays in query strings https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/states/validators.yaml#L23).
So if we going to be strict for some reasons we will hide our errors inside `Serialization error` log message, just because we are unable to decode `code` delivered as `String`.

tests/test_serialization.nim
```diff
@@ -0,0 +1,134 @@
...
+              message: string,
+              stacktraces: openArray[string]): RestGenericError =
+      RestGenericError(
+        code: uint64(status), message: message,
+        stacktraces: some(@stacktraces)
+      )
+
+    const GoodTestVectors = [
+      (
+        "{\"code\":500,\"message\":\"block not found\"}",
```
arnetheduck:
```suggestion
        """{"code":500,"message":"block not found"}""",
```

arnetheduck (approved):
After this change, we should also rename to `RestErrorMessage`, like the yaml type is named (also because `Error` is used in the rest client for exceptions, so it becomes confusing)

PR #4200, opened 2022-10-01
Updated nomenclature for eth2 naming

julian-st:


tersec:
Should target the `unstable` branch; otherwise, looks good

PR #4207, opened 2022-10-03
deprecate /eth2/beacon_chain/req/beacon_blocks_by_{range,root}/1/

tersec:
For when https://github.com/ethereum/consensus-specs/pull/3014 is merged

This could wait until an actual consensus spec release, but doesn't have to, because as noted:
> There haven't been any phase0 blocks in mainnet or any of the still-supported testnets in [`MIN_EPOCHS_FOR_BLOCK_REQUESTS`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#configuration) regardless, so the spec already technically supports not serving said blocks.

So it can be merged as soon as that spec PR is merged.


etan-status (approved):
Should we still do the rate limiting?

```nim
     peer.updateRequestQuota(blockByRangeLookupCost)
     peer.awaitNonNegativeRequestQuota()
```

tersec:
> Should we still do the rate limiting?
> 
> ```nim
>      peer.updateRequestQuota(blockByRangeLookupCost)
>      peer.awaitNonNegativeRequestQuota()
> ```

Sure, why not: https://github.com/status-im/nimbus-eth2/pull/4207/commits/f1960c8237690da9f45e0cb649e8d01fa0994293

Discourages use of it, since it's not entirely free.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -274,142 +274,23 @@ p2pProtocol BeaconSync(version = 1,
...
-
-        await response.writeBytesSZ(uncompressedLen, bytes, []) # phase0
-        inc found
-
-    debug "Block root request done",
-      peer, roots = blockRoots.len, count, found
+    # https://github.com/tersec/consensus-specs/blob/deprecate-phase0-beacon-blocks-req-resp/specs/phase0/p2p-interface.md#beaconblocksbyroot
+    # `/eth2/beacon_chain/req/beacon_blocks_by_root/1/` is deprecated. Clients
+    # MAY respond with an empty list during the deprecation transition period.
+    discard
```
zah:
```suggestion
    peer.updateRequestQuota(blockByRootLookupCost)
    peer.awaitNonNegativeRequestQuota()
```

tersec:
https://github.com/ethereum/consensus-specs/pull/3014 was merged

tersec:
https://github.com/status-im/nimbus-eth2/pull/4378

PR #4225, opened 2022-10-06
Light client decrease min time between queries

KonradStaniec:
As it currently stands light client takes a bit long to start. Main reason is that currently not that many peers serve light client data (especially on testnets like goerli) and light client can't retrieve bootstrap. The other reason is that querying 2 random peers every 1min(at most) makes it pretty slow to cycle through all available peers.

This fix decreases time between queries to 20s (at most).

I have experimented with other more complex fixes, like keeping table of peers which recently failed provide data and avoid asking them, but ultimately the most impactful and simple fix was to decrease query interval.


etan-status (approved):

etan-status:
this actually changes to 10-30 seconds (random), but yes, average is now 20s

PR #4229, opened 2022-10-11
bump nim-eth: move eth1 specific code to nimbus-eth1

jangko:


arnetheduck:
will be making a bump, closing this since it points to the wrong nim-eth commit

PR #4236, opened 2022-10-13
VC: Fix doppelganger protection never allow attestations.

cheatfate:
* Also it fixes place where doppelganger protection starts working.
* Add `/nimbus/v1/debug/chronos/metrics` call which provides chronos trackers values.
* Improve doppelganger protection logging.

### DIFF
--- a/beacon_chain/validator_client/attestation_service.nim
+++ b/beacon_chain/validator_client/attestation_service.nim
@@ -407,7 +407,7 @@ proc spawnAttestationTasks(service: AttestationServiceRef,
   var dutiesSkipped: seq[string]
   for index, duties in dutiesByCommittee:
     let (protectedDuties, skipped) = vc.doppelgangerFilter(duties)
-    dutiesSkipped.add(skipped)
+    if len(skipped) > 0: dutiesSkipped.add(skipped)
     if len(protectedDuties) > 0:
       asyncSpawn service.publishAttestationsAndAggregates(slot, index,
                                                           protectedDuties)
--- a/beacon_chain/validator_client/attestation_service.nim
+++ b/beacon_chain/validator_client/attestation_service.nim
@@ -403,11 +403,19 @@ proc spawnAttestationTasks(service: AttestationServiceRef,
       for item in attesters:
         res.mgetOrPut(item.data.committee_index, default).add(item)
       res
+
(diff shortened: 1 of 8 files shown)


beacon_chain/validator_client/attestation_service.nim
```diff
@@ -421,9 +403,19 @@ proc spawnAttestationTasks(service: AttestationServiceRef,
...
-      asyncSpawn service.publishAttestationsAndAggregates(slot, index, duties)
+    let (protectedDuties, skipped) = vc.doppelgangerFilter(duties)
+    if len(skipped) > 0: dutiesSkipped.add(skipped)
+    if len(protectedDuties) > 0:
+      asyncSpawn service.publishAttestationsAndAggregates(slot, index,
+                                                          protectedDuties)
+  if len(dutiesSkipped) > 0:
+    info "Doppelganger protection disabled validator duties",
+         validators = len(dutiesSkipped)
+    debug "Doppelganger protection disabled validator duties dump",
```
arnetheduck: this is likely too expensive for large setups (>1000 validators) - trace perhaps

cheatfate: It is one time log and it uses format `shortLog(validatorPubKey) & "@" & validatorIndex` and everything is mapped into array.

arnetheduck: for 1000 validators, that's a 10k character long line which will up the log parser most likely (and we want to be able to run 10k validators on one instance in the future) - I'd be careful.

arnetheduck:
Eventually I think we should probably move the doppelganger activation check to `ValidatorPool` so that we use the same doppelganger tracking mechanism in both VC and BN: validators can be added "dynamically" in both cases (either via deposit activation or keymanager API).

tersec:
The doppelganger detection only operates at startup either way regardless. It doesn't have to care whether some validators get added later.

arnetheduck:
> The doppelganger detection only operates at startup either way regardless.

which seems wrong? ie adding a validator to one setup and forgetting to remove it from the other should be caught by the detection..

cheatfate:
Doppelganger detection in `VC` actually cares about validators which could be added at any time, while `BN` doppelganger detection only works for validators which was present at `BN` startup.

PR #4238, opened 2022-10-14
temporarily disable LC sync in testnet

etan-status:
LC sync seems to trigger libp2p disconnects in launch_local_testnet in some cases. Disable the testing flag for now until investigated.

### DIFF
--- a/scripts/launch_local_testnet.sh
+++ b/scripts/launch_local_testnet.sh
@@ -994,7 +994,6 @@ for NUM_NODE in $(seq 0 $(( NUM_NODES - 1 ))); do
     --keymanager-token-file="${DATA_DIR}/keymanager-token" \
     --rest-port="$(( BASE_REST_PORT + NUM_NODE ))" \
     --metrics-port="$(( BASE_METRICS_PORT + NUM_NODE ))" \
-    --sync-light-client=on \
     ${EXTRA_ARGS} \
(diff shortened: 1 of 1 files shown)


arnetheduck:
sure that wasn't just a one-time fluke?

tersec:
For a while, this was a typical failure mode happening every few times; this isn't the first time this failure mode has emerged. There were some libp2p fixes for it, but there's reason to believe that while a fluke in some sense, it points to a real bug to be fixed.

In the meantime, it's more important that the finalization CI be trustworthy.

arnetheduck:
> , it points to a real bug to be fixed.

I'd agree this points to a bug (libp2p likely), but how does disabling LC help?

tersec:
> > , it points to a real bug to be fixed.
> 
> I'd agree this points to a bug (libp2p likely), but how does disabling LC help?

It removes the trigger.

tersec (approved):

etan-status:
> I'd agree this points to a bug (libp2p likely), but how does disabling LC help?

Only light client manager cancels pending requests that are no longer needed. This has been an issue for a while, with fixes over and over again. Seems like it just needs another round of fixing to not disconnect on cancel :)

Disabling it unblocks other PRs from randomly failing CI, I'll investigate the issue separately after devcon.

Issue #4087, opened 2022-09-07
VC: call prepareBeaconProposer to register suggested fee recipient


arnetheduck:
Before proposing, the VC needs to prepare the BN (which prepares the EL) for proposal - to do so, it should call https://ethereum.github.io/beacon-APIs/#/Validator/prepareBeaconProposer

v22.8.2

arnetheduck:
fixed in #4240

PR #4240, opened 2022-10-14
VC: some fixes

cheatfate:
1. Optimize doppelganger protection to skip protection for validators which are not activated yet.
2. Cache sync committee duties to avoid spam of `prepareSyncCommitteeSubnets()` calls.
3. Fix lazy missing `lazyWait()` for `firstSuccessParallel()` template.
4. Address #4087.

### DIFF
--- a/beacon_chain/validator_client/duties_service.nim
+++ b/beacon_chain/validator_client/duties_service.nim
@@ -632,7 +632,8 @@ proc mainLoop(service: DutiesServiceRef) {.async.} =
     # become safe to combine loops, breaks and exception handlers.
     let breakLoop =
       try:
-        discard await race(attestFut, proposeFut, indicesFut, syncFut)
+        discard await race(attestFut, proposeFut, indicesFut, syncFut,
+                           prepareFut)
         checkAndRestart(AttesterLoop, attestFut, service.attesterDutiesLoop())
         checkAndRestart(ProposerLoop, proposeFut, service.proposerDutiesLoop())
         checkAndRestart(IndicesLoop, indicesFut, service.validatorIndexLoop())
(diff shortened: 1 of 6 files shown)


beacon_chain/validator_client/duties_service.nim
```diff
@@ -96,6 +99,7 @@ proc pollForValidatorIndices*(vc: ValidatorClientRef) {.async.} =
          updated_validators = len(updated)
     trace "Validator indices update dump", missing_validators = missing,
           updated_validators = updated
+    vc.indicesAvailable.fire()
```
arnetheduck, line 102: what's this event for? ie indices are not always available until later and it might be the case that some indices are avail and some are not (ie one active validator and one that was "just" deposited and hasn't been processed yet) - whatever code relies on this will have to handle both "more" indices becoming available and "some" indices not being available

arnetheduck (approved):

beacon_chain/validator_client/duties_service.nim
cheatfate: This event is just for ordering different independent loops, all validators which do not have indices will be queried for index every slot.
```nim
proc validatorIndexLoop(service: DutiesServiceRef) {.async.} =
  let vc = service.client
  while true:
    await vc.pollForValidatorIndices()
    await service.waitForNextSlot(IndicesLoop)
```
The only problem that beacon proposer preparation loop requires this indices. And because all loops are started in parallel, the only way to order things up is to have such event.
```nim
proc proposerPreparationsLoop(service: DutiesServiceRef) {.async.} =
  let vc = service.client
  debug "Beacon proposer preparation loop waiting for validator indices update"
  await vc.indicesAvailable.wait()
  while true:
    await service.prepareBeaconProposers()
    await service.waitForNextSlot(ProposerPreparationLoop)
```
So if at the start `VC` do not have any validators attached `proposerPreparationsLoop` will wait for first indices to be available to start working...

PR #4242, opened 2022-10-15
fix false-positive warnings on expected VALID fcU status; adjust log levels

tersec:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -183,16 +183,39 @@ from ../eth1/eth1_monitor import
...
+  # relative lack of immediate evidence.
+  if receivedExecutionBlockHash != headBlockHash:
+    return
+
+  case payloadExecutionStatus
+  of PayloadExecutionStatus.valid:
+    # situation nominal
+    discard
+  of PayloadExecutionStatus.accepted, PayloadExecutionStatus.syncing:
+    info "expectValidForkchoiceUpdate: forkChoiceUpdated ACCEPTED/SYNCING, not VALID",
```
arnetheduck: info logs should be intelligable for users ideally - the function name here is quite unfriendly

tersec: https://github.com/status-im/nimbus-eth2/pull/4242/commits/48c81ea0a4512c9e81933d91a4fddf5ae9fd3d68

PR #4246, opened 2022-10-18
only support x86 CPUs supported by non-EOL macOS software

tersec:
Can't assume ADX yet, but start using BMI2 if available and supported by BLST for the [`MULX`](https://www.felixcloutier.com/x86/mulx) instruction; https://github.com/status-im/nimbus-eth2/blob/stable/docs/cpu_features.md#bmi2--adx discusses this.


zah:
Still works fine on my trusty Macbook from 2012!

PR #4251, opened 2022-10-21
State-only checkpoint state startup

arnetheduck:
Currently, we require genesis and a checkpoint block and state to start from an arbitrary slot - this PR relaxes this requirement so that we can start with a state alone.

The current trusted-node-sync algorithm works by first downloading blocks until we find an epoch aligned non-empty slot, then downloads the state via slot.

However, current [proposals](https://github.com/ethereum/beacon-APIs/pull/226) for checkpointing prefer finalized state as the main reference - this allows more simple access control and caching on the server side - in particular, this should help checkpoint-syncing from sources that have a fast `finalized` state download (like teku) but are slow when accessing state via slot.

Earlier versions of Nimbus will not be able to read databases created without a checkpoint block and genesis. In most cases, backfilling makes the database compatible except where genesis is also missing (custom networks).

* backfill checkpoint block from libp2p instead of checkpoint source, when doing trusted node sync
* allow starting the client without genesis / checkpoint block
* perform epoch start slot lookahead when loading tail state, so as to deal with the case where the epoch start slot does not have a block
* replace `--blockId` with `--state-id` in TNS command line
* when replaying, also look at the parent of the last-known-block (even if we don't have the parent block data, we can still replay from a "parent" state) - in particular, this clears the way for implementing state pruning
* deprecate `--finalized-checkpoint-block` option (no longer needed)

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -201,8 +201,81 @@ func getBlockIdAtSlot*(dag: ChainDAGRef, slot: Slot): Opt[BlockSlotId] =
 
   err() # not backfilled yet
 
+proc containsBlock(
+    cfg: RuntimeConfig, db: BeaconChainDB, slot: Slot, root: Eth2Digest): bool =
+  db.containsBlock(root, cfg.blockForkAtEpoch(slot.epoch))
+
 proc getForkedBlock*(db: BeaconChainDB, root: Eth2Digest):
-    Opt[ForkedTrustedSignedBeaconBlock] {.gcsafe.}
+    Opt[ForkedTrustedSignedBeaconBlock] =
+  # When we only have a digest, we don't know which fork it's from so we try
+  # them one by one - this should be used sparingly
+  if (let blck = db.getBlock(root, bellatrix.TrustedSignedBeaconBlock);
+      blck.isSome()):
+    ok(ForkedTrustedSignedBeaconBlock.init(blck.get()))
+  elif (let blck = db.getBlock(root, altair.TrustedSignedBeaconBlock);
+      blck.isSome()):
+    ok(ForkedTrustedSignedBeaconBlock.init(blck.get()))
+  elif (let blck = db.getBlock(root, phase0.TrustedSignedBeaconBlock);
+      blck.isSome()):
+    ok(ForkedTrustedSignedBeaconBlock.init(blck.get()))
+  else:
+    err()
+
+proc containsBlock(dag: ChainDAGRef, bid: BlockId): bool =
+  let fork = dag.cfg.blockForkAtEpoch(bid.slot.epoch)
+  if dag.db.containsBlock(bid.root, fork):
+    return true
+
+  # TODO avoid loading bytes from era
+  var bytes: seq[byte]
+  (bid.slot <= dag.finalizedHead.slot and
+    getBlockSZ(
+      dag.era, getStateField(dag.headState, historical_roots).asSeq,
+      bid.slot, bytes).isOk and bytes.len > 0)
+
+proc getBlock*(
+    dag: ChainDAGRef, bid: BlockId,
+    T: type ForkyTrustedSignedBeaconBlock): Opt[T] =
+  dag.db.getBlock(bid.root, T) or
+    getBlock(
+      dag.era, getStateField(dag.headState, historical_roots).asSeq,
+      bid.slot, Opt[Eth2Digest].ok(bid.root), T)
+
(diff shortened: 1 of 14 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -579,74 +642,48 @@ proc getState(
...
-  # content
-  let fork = dag.cfg.blockForkAtEpoch(bid.slot.epoch)
-  dag.db.getBlockSSZ(bid.root, bytes, fork) or
-    (bid.slot <= dag.finalizedHead.slot and
-      getBlockSSZ(
-        dag.era, getStateField(dag.headState, historical_roots).asSeq,
-        bid.slot, bytes).isOk)
+  let rollbackAddr =
+    # Any restore point will do as long as it's not the object being updated
+    if unsafeAddr(state) == unsafeAddr(dag.headState):
```
tersec, line 668: Nim 1.6 deprecates `unsafeAddr`

arnetheduck: yeah, this is simply to keep it in line with other uses in this module which use the same construct - we can fix them all in one go when we drop 1.2

beacon_chain/trusted_node_sync.nim
```diff
@@ -295,15 +239,18 @@ proc doTrustedNodeSync*(
...
-    for i in 0..<checkpointSlot.int:
-      if dbCache.slots[i].isNone():
-        total += 1
-    total
+  dbCache.updateSlots(checkpointSlot)
+
+  let
+    missingSlots = block:
+      var total = 0
+      for i in 0..<checkpointSlot.int:
```
tersec: In theory, one can avoid this `Slot` <-> `int` conversion by writing
```
for i in 0.Slot ..< checkpointSlot:
  discard dbCache.isKnown(i)
```
where the rest of the loop aside, `i` remains a slot the entire time. At some point, there were Nim issues around this construct, so maybe not reliable enough to use.

tersec: Also, the range of `int` isn't necessarily the range of `Slot` -- seems better to at least use `uint64` here.

beacon_chain/trusted_node_sync.nim
```diff
@@ -23,13 +23,16 @@ type
     summaries: Table[Eth2Digest, BeaconBlockSummary]
     slots: seq[Option[Eth2Digest]]
 
+proc updateSlots(cache: var DbCache, slot: Slot) =
+  if cache.slots.len() < slot.int + 1:
```
tersec: Another `int`/`uint64` thing -- ultimately, it uses `setLen` in the line below, which means that it's dealing with an `int` regardless, but still seems better to avoid potentially lossy conversions when feasible, e.g.,
```nim
  if cache.slots.lenu64() < slot + 1
```
The dependent on `int` is ultimately a dependence on `seq` and that built-in set of types. My preference is, even if that's going to be the case indefinitely, centralize as much of these type conversions to/from `int` from not-at-all-`int` types in one place, even when they're currently not feasible to avoid entirely.

beacon_chain/trusted_node_sync.nim
```diff
@@ -122,167 +126,107 @@ proc doTrustedNodeSync*(
...
-        # for now, we ask the user to fix it instead
-        error "A checkpoint block from the first slot of an epoch could not be found with the given block id - pass an epoch slot with a block using the --block-id parameter",
-          blockId
+    let
+      state = try:
+        let id = block:
+          let tmp = StateIdent.decodeString(stateId).valueOr:
+            error "Cannot decode checkpoint state id, must be a slot, hash, 'finalized' or 'head'"
+            quit 1
+          if tmp.kind == StateQueryKind.Slot:
```
tersec: Equivalent to checking for `tmp.kind == StateQueryKind.Slot and not tmp.slot.is_epoch()` and else, `tmp`, no?

Seems clearer to avoid unnecessarily nested `if` statements.

tersec (approved):

beacon_chain/trusted_node_sync.nim
arnetheduck: ae0d1cabfc6d10cd7baa9d32be072aec03e4066b

beacon_chain/trusted_node_sync.nim
arnetheduck: 68e54467b598dd2c41edc38614f46d3009693360 - that said, in constructs like `for a in 0..<slot.int`, there's a real advantage of doing the `int` conversion up-front in that the loop won't start at all if the slot is off (say because the code is missing a check or has an underflow) meaning less risk of corruption due to side effects in the loop..

PR #4255, opened 2022-10-24
Reduce fee recipient log spam

arnetheduck:
Once is enough

### DIFF
--- a/beacon_chain/spec/eth2_apis/dynamic_fee_recipients.nim
+++ b/beacon_chain/spec/eth2_apis/dynamic_fee_recipients.nim
@@ -19,10 +19,21 @@ proc addMapping*(store: var DynamicFeeRecipientsStore,
                  validator: ValidatorIndex,
                  feeRecipient: Eth1Address,
                  currentEpoch: Epoch) =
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/eth2_apis/dynamic_fee_recipients.nim
```diff
@@ -19,10 +19,21 @@ proc addMapping*(store: var DynamicFeeRecipientsStore,
...
                  feeRecipient: Eth1Address,
                  currentEpoch: Epoch) =
-  info "Updating fee recipient",
-    validator, feeRecipient = feeRecipient.toHex(), currentEpoch
-  store.mappings[validator] = Entry(recipient: feeRecipient,
-                                    addedAt: currentEpoch)
+  var
+    found, updated = false
+  store.mappings.withValue(validator, entry) do:
+    updated = not (entry[].recipient == feeRecipient)
```
tersec, line 25:
```suggestion
    updated = entry[].recipient != feeRecipient
```

tersec (approved):

Issue #4116, opened 2022-09-13
VC: support external block builders by using `/eth/v1/validator/register_validator`


tersec:
Followup to https://github.com/status-im/nimbus-eth2/issues/4051 / https://github.com/status-im/nimbus-eth2/pull/4115

tersec:
https://github.com/status-im/nimbus-eth2/pull/4260

PR #4260, opened 2022-10-26
VC: Fix for #4116 (external block builders support)

cheatfate:


beacon_chain/conf.nim
```diff
@@ -903,6 +903,12 @@ type
       defaultValue: 0
       name: "stop-at-epoch" .}: uint64
 
+    builderRegistration* {.
+      desc: "Provide beacon node with registrations for the given validators " &
+            "to the external builder network"
+      defaultValue: false
+      name: "builder-api-registration" .}: bool
```
arnetheduck: based on discussion in our weekly call, this flag will also cause block production to use a separate call (https://ethereum.github.io/beacon-APIs/#/Validator/produceBlindedBlock) meaning that the name should be updated to reflect "both" registration and block production - as such, `--builder-proposals` for example is appropriate.

arnetheduck: actually, even better is to use the same name as the beacon node (`--payload-builder`): https://github.com/status-im/nimbus-eth2/blob/e6a346d4714c3daef37693db5291b7efd2dde02e/beacon_chain/conf.nim#L566

beacon_chain/validator_client/api.nim
```diff
@@ -2087,3 +2087,43 @@ proc prepareBeaconProposer*(
...
+      debug "Validator registration was interrupted"
+      return 00
+    of ApiOperation.Failure:
+      debug "Unexpected error happened while registering validators"
+      return 0
+  else:
+    var count = 0
+    for apiResponse in resp.data:
+      if apiResponse.data.isErr():
+        debug "Unable to register validator with beacon node",
```
arnetheduck, line 2120: this error message seems important for users to resolve their issues - should maybe be logged at higher level

PR #4261, opened 2022-10-26
capella test vector support

tersec:
This is not all of Capella
- it pointedly avoids `Forked` things -- that's another round of intrusive changes conceptually unrelated to the Capella state transition function itself
- it's enough to implement the test vectors available at https://github.com/ethereum/consensus-spec-tests/commits/master as of a couple of hours ago, except those with depended on `Forked` types (transition tests, slot tests, and block tests)
- it should be safe to merge whenever, and does not interfere with Bellatrix operation

[minimal.log](https://github.com/status-im/nimbus-eth2/files/9872297/minimal.log)
[mainnet.log](https://github.com/status-im/nimbus-eth2/files/9872298/mainnet.log)

### DIFF
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -19,6 +19,8 @@ import
   ./datatypes/[phase0, altair, bellatrix],
   "."/[eth2_merkleization, forks, signatures, validator]
 
+from ./datatypes/capella import BeaconState, ExecutionPayloadHeader, Withdrawal
+
 export extras, forks, validator
 
 # https://github.com/ethereum/consensus-specs/blob/v1.2.0/specs/phase0/beacon-chain.md#increase_balance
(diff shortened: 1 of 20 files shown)


arnetheduck (approved):

arnetheduck:
```
/home/arnetheduck/status/nimbus-eth2/beacon_chain/spec/state_transition_block.nim(635, 34) Warning: use `delete(s, first..last)`; delete is deprecated [Deprecated]

```

tersec:
> ```
> /home/arnetheduck/status/nimbus-eth2/beacon_chain/spec/state_transition_block.nim(635, 34) Warning: use `delete(s, first..last)`; delete is deprecated [Deprecated]
> ```

https://nim-lang.org/docs/sequtils.html#delete%2Cseq%5BT%5D%2CNatural%2CNatural and https://nim-lang.org/docs/sequtils.html#delete%2Cseq%5BT%5D%2CSlice%5Bint%5D are the functions in question here.

The former is deprecated in 1.6, but 1.2 doesn't have the latter.

It might be worth working around this warning by doing this slightly more manually, but I'm also anticipating that within another couple months we might be able to use purely 1.6.

PR #4262, opened 2022-10-26
Optimistic mode

arnetheduck:
In optimistic mode, Nimbus will sync optimistically even when the execution client is offline / not available.

An optimistic node is less secure because it has not validated block transactions via the execution client and can thus not be used for validation duties.


docs/the_nimbus_book/src/optimistic-sync.md
```diff
@@ -0,0 +1,27 @@
+# Optimistic sync
+
+Optimistic sync is the process of syncing an execution and consensus client concurrently, without having the consensus client wait for the execution client. During optimistic sync, the consensus client quickly syncs up to the latest conesensus but delays verifying block execution payloads - it continuously informs the execution client of the latest consensus head, allowing the execution client to perform a snapshot sync directly to the latest state.
```
tersec, line 3: In theory -- practically, because of snap sync, ELs wait for CLs to near-finish. This is one motivation for LC sync. Currently optimistic sync isn't that concurrent.

tersec (approved):

docs/the_nimbus_book/src/optimistic-sync.md
arnetheduck: because of trusted node sync, the CL near-finishes instantly

PR #4276, opened 2022-11-01
add Capella support to Forked*

tersec:
- a large no-op, essentially, the goal of which is from this point to allow safe, incremental Capella development/fleshing-out within the `unstable` branch
- general pattern was to flesh out `spec/forks` to its almost Capella-final form, but the rest to stub where feasible, to preserve the no-op aspect in a PR which already in an essentially minimal form adds 500 lines and changes 42 files
- `capellaImplementationMissing` serves both as a `grep` (`rg`, `ag`, etc) anchor, but also as a compile-time anchor. Once the `capellaImplementationMissing`s in an area of the codebase is complete, and it compiles without it, it implies that the Capella aspects should be done there
- where the data is under internal control, it does `raiseAssert $capellaImplementationMissing` to be as visible as possible, because after all, `checkForkConsistency()` ensures that `nimbus_beacon_node` can't run yet with a Capella network:
https://github.com/status-im/nimbus-eth2/blob/3ef09ff596b54b02c6d71bc0291999f6e6f93429/beacon_chain/spec/datatypes/base.nim#L955-L958

However, `beacon_chain/spec/eth2_apis/eth2_rest_serialization` has places where otherwise, this would allow random unauthenticated external input to crash Nimbus at will, obviously suboptimal -- therefore it returns as a more controlled parsing error, but still both greppable and dependent on this symbol that allows compile-time checking. When writing, it stays with the `raiseAssert`, because that is under internal control and if that happens, that's a logic error.

- There's a change between Nim 1.2 and 1.6 whereby
```nim
proc f() =
  raiseAssert "foo"
  echo "hello world"
```
changes from an error around the unreachable `echo` statement in Nim 1.2 to a warning. Therefore, in several places, for compatibility between Nim 1.2 and 1.6, this PR uses:
```nim
proc f() =
  if true: raiseAssert "foo"
  echo "hello world"
```

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -18,6 +18,9 @@ import
     state_transition, state_transition_epoch],
   "."/[block_dag, blockchain_dag, blockchain_dag_light_client]
 
+# TODO remove when forks re-exports this
+from ../spec/datatypes/capella import asSigVerified, asTrusted, shortLog
+
 export results, signatures_batch, block_dag, blockchain_dag
 
 # Clearance
@@ -36,7 +39,7 @@ proc addResolvedHeadBlock(
        trustedBlock: ForkyTrustedSignedBeaconBlock,
        blockVerified: bool,
        parent: BlockRef, cache: var StateCache,
-       onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded | OnBellatrixBlockAdded,
+       onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded | OnBellatrixBlockAdded | OnCapellaBlockAdded,
        stateDataDur, sigVerifyDur, stateVerifyDur: Duration
      ): BlockRef =
   doAssert state.matches_block_slot(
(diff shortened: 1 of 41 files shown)


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -36,7 +39,7 @@ proc addResolvedHeadBlock(
        trustedBlock: ForkyTrustedSignedBeaconBlock,
        blockVerified: bool,
        parent: BlockRef, cache: var StateCache,
-       onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded | OnBellatrixBlockAdded,
+       onBlockAdded: OnPhase0BlockAdded | OnAltairBlockAdded | OnBellatrixBlockAdded | OnCapellaBlockAdded,
```
arnetheduck: time for `OnForkyBlockAdded` I think

tests/consensus_spec/fixtures_utils.nim
```diff
@@ -120,6 +124,9 @@ proc loadForkedState*(
   # prevent temporaries created by case objects
   let forkedState = new ForkedHashedBeaconState
   case fork
+  of BeaconStateFork.Capella:
+    doAssert capellaImplementationMissing
```
arnetheduck: cruft

tersec: https://github.com/status-im/nimbus-eth2/pull/4276/commits/b91875d5351843d3042dba1beb71a41f452bad62

arnetheduck (approved):

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: This isn't something I do routinely now, since https://github.com/nim-lang/Nim/issues/18095 still isn't fixed and it's a gamble each time whether attempting to do so will have been a waste of time, but in this case, https://github.com/status-im/nimbus-eth2/pull/4276/commits/bf785d654ae3aa3b6ac1a036f3acdf2489d97cac seems to work on both Nim 1.2 and Nim 1.6.

PR #4281, opened 2022-11-02
Sync committee subscription fixes.

cheatfate:
* Optimize server `/eth/v1/validator/sync_committee_subscriptions` call. There is no need to create sequence.
* Fix `prepareBeaconCommitteeSubnet()` should be sent to all corresponding beacon nodes.
* Fix `prepareSyncCommitteeSubnets()` should  be sent to all corresponding beacon nodes.
* Show error response in log for calls.

### DIFF
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -716,24 +716,31 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
   router.api(MethodPost,
              "/eth/v1/validator/sync_committee_subscriptions") do (
     contentBody: Option[ContentBody]) -> RestApiResponse:
-    if contentBody.isNone():
-      return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
-    let dres = decodeBody(seq[RestSyncCommitteeSubscription],
-                          contentBody.get())
-    if dres.isErr():
-      return RestApiResponse.jsonError(Http400,
+    let subscriptions =
+      block:
+        var res: seq[RestSyncCommitteeSubscription]
+        if contentBody.isNone():
+          return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
+        let dres = decodeBody(seq[RestSyncCommitteeSubscription],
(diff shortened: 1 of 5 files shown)


beacon_chain/validator_client/api.nim
```diff
@@ -1771,180 +1771,85 @@ proc publishBlock*(
...
-    strategy = $strategy
-
-  const
-    ErrorMessage = "Unable to prepare sync committee subnet"
-    NoErrorMessage = "Commitee subnet was successfully prepared"
-
-  case strategy
-  of ApiStrategyKind.First, ApiStrategyKind.Best:
-    var status = false
-    let res = vc.firstSuccessParallel(RestPlainResponse, OneThirdDuration,
```
arnetheduck, line 1876: doesn't this _also_ send the request to all beacon nodes? the difference is how long it waits, right?

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -716,37 +716,33 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-                                             EpochFromTheIncorrectForkError)
-          if uint64(item.validator_index) >=
-             lenu64(getStateField(node.dag.headState, validators)):
-            return RestApiResponse.jsonError(Http400,
-                                             InvalidValidatorIndexValueError)
-          let validator_pubkey =
-            getStateField(node.dag.headState, validators).item(
-              item.validator_index).pubkey
+    let subs = dres.get()
+    for item in subs:
```
arnetheduck: if there's an error in the second item, the first item will be added resulting in a "partial success" - ideally the validation would be done in a separate read-only loop (the existing code is also wrong / strange in this regard - I think the idea with the extra seq was to do this validation first, but somehow the mutation ended up in the same loop...)

beacon_chain/validator_client/api.nim
cheatfate: `firstSuccessParallel()` sends to all too, but its not going to wait for result from all, so as soon as you received first successful result it will ignore all other results.

arnetheduck (approved):

PR #4283, opened 2022-11-03
Make trusted node sync era-aware

arnetheduck:
This PR removes a bunch of code to make TNS aware of era files, avoiding a duplicated backfill when era files are available.

* reuse chaindag for loading backfill state, replacing the TNS homebrew
* fix era block iteration to skip empty slots
* add tests for `can_advance_slots`

### DIFF
--- a/beacon_chain/era_db.nim
+++ b/beacon_chain/era_db.nim
@@ -352,10 +352,15 @@ proc getPartialState(
     false
 
 iterator getBlockIds*(
-    db: EraDB, historical_roots: openArray[Eth2Digest], start_slot: Slot): BlockId =
+    db: EraDB, historical_roots: openArray[Eth2Digest],
+    start_slot: Slot, prev_root: Eth2Digest): BlockId =
+  ## Iterate over block roots starting from the given slot - `prev_root` must
+  ## point out the last block added to the chain before `start_slot` such that
+  ## empty slots can be filtered out correctly
   var
     state = (ref PartialBeaconState)() # avoid stack overflow
     slot = start_slot
+    prev_root = prev_root
 
   while true:
     # `case` ensures we're on a fork for which the `PartialBeaconState`
@@ -372,11 +377,14 @@ iterator getBlockIds*(
       break
 
     let
-      x = slot.int mod state[].block_roots.len
-    for i in x..<state[].block_roots.len():
-      # TODO these are not actually valid BlockId instances in the case where
-      #      the slot is missing a block - use index to filter..
-      yield BlockId(root: state[].block_roots.data[i], slot: slot)
+      x = slot.uint64 mod state[].block_roots.lenu64
+
+    for i in x..<state[].block_roots.lenu64():
+      # When no block is included for a particular slot, the block root is
+      # repeated
+      if slot == 0 or prev_root != state[].block_roots.data[i]:
+        yield BlockId(root: state[].block_roots.data[i], slot: slot)
+        prev_root = state[].block_roots.data[i]
       slot += 1
 
 proc new*(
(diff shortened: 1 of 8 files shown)


beacon_chain/era_db.nim
```diff
@@ -399,31 +407,50 @@ when isMainModule:
...
-  var got8191 = false
-  var slot4: Eth2Digest
-  for bid in db.getBlockIds(historical_roots, Era(0)):
-    if bid.slot == Slot(1):
+  var
+    got0 = false
+    got8191 = false
+    got8192 = false
+    got8193 = false
+  for bid in db.getBlockIds(historical_roots, Slot(0), Eth2Digest()):
```
tersec, line 423: Nothing new in this PR, but some modules having the in-line `isMainModule` tests and others depending on the test suite is a bit odd.

beacon_chain/trusted_node_sync.nim
```diff
@@ -210,148 +137,114 @@ proc doTrustedNodeSync*(
...
-        if blck.message.slot == checkpointSlot:
-          if blck.root != checkpointRoot:
-            error "Downloaded block does not match checkpoint history",
-              blck = shortLog(blck),
-              expectedRoot = shortLog(checkpointRoot)
+    proc downloadBlock(slot: Slot):
+        Future[Option[ref ForkedSignedBeaconBlock]] {.async.} =
+      # Download block at given slot, retrying a few times,
+      var lastError: ref CatchableError
+      for i in 0..<3:
```
tersec, line 179: Presumably this `0..<3` is the "a few times" part -- maybe useful to clarify that if so, that the loop isn't a fundamental part of each request

beacon_chain/trusted_node_sync.nim
```diff
@@ -210,148 +137,114 @@ proc doTrustedNodeSync*(
...
-            error "Expected child block not found in checkpoint history"
-            quit 1
+      raise lastError
 
-        if blck.root notin dbCache.summaries:
-          db.putBlock(blck.asTrusted())
+    # Download several blocks in parallel but process them serially
+    proc processBlock(blck: Option[ref ForkedSignedBeaconBlock]) =
+      let newStamp = SyncMoment.now(processed)
+      if newStamp.stamp - stamp.stamp > 12.seconds:
```
tersec, line 201: why `12.seconds` rather than `SECONDS_PER_SLOT`?

beacon_chain/era_db.nim
arnetheduck: yeah - in this particular case it's motivated by the lack of reasonable test data - not sure if it's better to generate it on the fly in the test suite or download the first two era files like we do with test scenarios in spec

arnetheduck:
```
-rw-r--r--. 1 arnetheduck arnetheduck 1.8M May 30 09:56 mainnet-00000-4b363db9.era
-rw-r--r--. 1 arnetheduck arnetheduck  18M May 30 09:56 mainnet-00001-40cf2f3c.era
```
20mb roughly .. using these files would mean not testing writing - testing writing means  observable side effects in tests => flakey :/

beacon_chain/trusted_node_sync.nim
arnetheduck: this delay is kind of arbitrary (controls progress logging frequency) - ie it doesn't have anything to do with slots really, except by convenient coincidence

beacon_chain/trusted_node_sync.nim
arnetheduck: clarify how? ie the loop returns on first succes

tersec: Alright, indeed

PR #4284, opened 2022-11-03
Show last 10 lines from the test log file on test failure

zah:


Makefile
```diff
@@ -36,6 +36,8 @@ VALIDATORS := 1
 CPU_LIMIT := 0
 BUILD_END_MSG := "\\x1B[92mBuild completed successfully:\\x1B[39m"
 
+TEST_MODULES_FLAGS := -d:chronicles_log_level=TRACE -d:chronicles_sinks=json[file]
```
arnetheduck, line 39: on trace level, 10 lines is .. not a lot

zah: Yes, learned this the hard way and made another commit in `unstable` after merging this.

PR #4286, opened 2022-11-07
implement /eth/v1/beacon/blinded_blocks BN endpoint

tersec:


beacon_chain/validators/message_router_mev.nim
```diff
@@ -0,0 +1,122 @@
...
+    dst: untyped, src: untyped, fieldNames: static[seq[string]]): untyped =
+  result = newStmtList()
+  for name in fieldNames:
+    if name notin [
+        # These fields are the ones which vary between the blinded and
+        # unblinded objects, and can't simply be copied.
+        "transactions_root", "execution_payload",
+        "execution_payload_header", "body"]:
+      # TODO use stew/assign2
+      result.add newAssignment(
```
arnetheduck, line 43: we should ideally use `stew/assign2` here (which is orders of magnitude faster), but that's unrelated to this PR

beacon_chain/validators/message_router_mev.nim
```diff
@@ -0,0 +1,122 @@
...
+      awaitWithTimeout(
+        node.payloadBuilderRestClient.submitBlindedBlock(blindedBlock),
+        BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE):
+          error "Submitting blinded block timed out",
+            blk = shortLog(blindedBlock)
+          return err("Submitting blinded block timed out")
+      # From here on, including error paths, disallow local EL production by
+      # returning Opt.some, regardless of whether on head or newBlock.
+    except RestDecodingError as exc:
+      error "unblindAndRouteBlockMEV: REST decoding error submitting blinded block",
```
arnetheduck: avoid function names in non-debug logs - also, both logging _and_ returning an error creates unclarity as to whose responsibility it is to log

arnetheduck (approved):

beacon_chain/validators/message_router_mev.nim
tersec: Well, per the `TODO`, I agree. I want to add tests for this (useful/good anyway) for this first though before changing this.

arnetheduck: oops didn't see todo :)

beacon_chain/validators/message_router_mev.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4286/commits/faae1cf73efcdaa35d0fcfa30f193b74f1b167ef

arnetheduck (approved):

beacon_chain/validators/message_router_mev.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4286/commits/1a3ca5fe94c7bd5c69d1072d3826bebe447f5f94  removes the `unblindAndRouteBlockMEV` error logging, after verification that both BN and VC calling codepaths adequately log this.

PR #4291, opened 2022-11-07
set fee recipient in empty payload fallback

### LINKED ISSUE #4289: Empty execution payload should have fee recipient set
https://github.com/status-im/nimbus-eth2/blob/06ccf5b80c316a1eae4f8dbcdc3a4b822e80e7ad/beacon_chain/spec/helpers.nim#L383 is missing the fee recipient

etan-status:
When the EL/Builder fails to produce an execution payload, we fall back to an empty `ExecutionPayload`. Even though it contains no transactions it should refer to the configured fee recipient. This is useful for privacy reasons (do not reveal the reason for the empty payload) and for compliance with additional fee recipient rules by staking pools.

### DIFF
--- a/tests/test_helpers.nim
+++ b/tests/test_helpers.nim
@@ -69,6 +69,7 @@ suite "Spec helpers":
     const recipient1 = default(Eth1Address)
 
     var recipient2: Eth1Address
+    randomize(123)
     let p = cast[ptr UncheckedArray[byte]](addr recipient2)
     for i in 0 ..< sizeof(recipient2):
       p[i] = byte.rand()
--- a/tests/test_helpers.nim
+++ b/tests/test_helpers.nim
@@ -8,13 +8,12 @@
 {.used.}
 
 import
-  # Standard library
-  std/random,
   # Status libraries
   stew/bitops2,
+  web3/ethtypes,
   # Beacon chain internals
   ../beacon_chain/spec/[forks, helpers, state_transition],
-  ../beacon_chain/spec/datatypes/bellatrix,
+  ../beacon_chain/eth1/eth1_monitor,
   # Test utilities
   ./unittest2, mocking/mock_genesis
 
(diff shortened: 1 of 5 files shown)


tests/test_helpers.nim
```diff
@@ -58,3 +60,25 @@ suite "Spec helpers":
...
+    var cfg = defaultRuntimeConfig
+    cfg.ALTAIR_FORK_EPOCH = GENESIS_EPOCH
+    cfg.BELLATRIX_FORK_EPOCH = GENESIS_EPOCH
+
+    const recipient1 = default(Eth1Address)
+
+    var recipient2: Eth1Address
+    let p = cast[ptr UncheckedArray[byte]](addr recipient2)
+    for i in 0 ..< sizeof(recipient2):
+      p[i] = byte.rand()
```
tersec: Seems better to avoid even this kind of mostly benign per-run randomness in the test suite, so there's less reason it can sometimes fail in a way that's tricky to reproduce.

Also, in general, we've mostly avoided using `std/random`, e.g.
https://github.com/status-im/nimbus-eth2/blob/c585b0a5b1ae4d55af38ad7f4715ad455e791552/beacon_chain/nimbus_beacon_node.nim#L608-L609

For tests, the weaknesses of `std/random` don't matter as much, but might nicer to pregenerate as many random recipients as desired and hardcode them in.

tersec (approved):

tests/test_helpers.nim
arnetheduck: +1 hardcode some value, doesn't seem like there's added value from a random number here

tests/test_helpers.nim
```diff
@@ -58,3 +61,26 @@ suite "Spec helpers":
...
+  test "build_empty_execution_payload":
+    var cfg = defaultRuntimeConfig
+    cfg.ALTAIR_FORK_EPOCH = GENESIS_EPOCH
+    cfg.BELLATRIX_FORK_EPOCH = GENESIS_EPOCH
+
+    const recipient1 = default(Eth1Address)
+
+    var recipient2: Eth1Address
+    randomize(123)
+    let p = cast[ptr UncheckedArray[byte]](addr recipient2)
```
tersec: An `Eth1Address` is [already](https://github.com/status-im/nim-web3/blob/master/web3/ethtypes.nim#L19)
```nim
Address* = distinct array[20, byte]
```
so this `cast`/`ptr`/etc to a `byte` array is unnecessary. It's already one and can just be filled that way (but if one hardcodes it, then it can look like `Eth1Address.fromHex(...)` and it also avoids this cast. Either way).

etan-status: Using `fromHex` now

arnetheduck: https://github.com/status-im/nim-eth/blob/master/eth/common/eth_types.nim#L33 is the canonical one for low-level  code (though I'm of a mind to change it to something more distinct)

tersec: Sure, it's a 20-`byte` array regardless

arnetheduck: yeah, I'm just waiting for @zah to do the strict serializer, then we can use these common types a bit more with less risk that one standard infects the other.. that said, I'm not entirely sure there's good economy in using these common types, ie the dependency is often more costly than the conversion, in terms of having to "consider" consequences when updating `nim-eth`

Issue #4057, opened 2022-09-01
Don't sync eth1 deposits when EL is unsynced


arnetheduck:
Trying to fetch deposits on an unsynced EL results in addtional load on the EL further delaying its sync and leads to timeouts in the API - it's also bound to fail most of the time since the EL is known to be out of date.

22.8.1

zah:
We currently detect that the EL is not synced by trying to request a block from time to time (the requested block is the starting point for syncing the deposits). Is the suggesting here to switch to a different specialized API for obtaining the sync status (we'll still have to request the starting block and handle the possibility that it may not be returned)?

arnetheduck:
well, when the beacon node is optimistically synced, the execution data has not been verified yet and the execution client is by definition out of sync - we could simply use the optimistic-or-not status to determine whether deposit sync should be active

zah:
There are no downsides of syncing the deposits as early as possible. Nimbus will not persist the deposits in the database until the data has been confirmed by the finalization process on the beacon chain. If the finalized data indicates that the downloaded deposits are not authentic, the client will automatically restart the syncing process from the latest known confirmed point. The information stored in the downloaded deposits won't be used until the first block proposal, which would require the client to already be fully synced and the deposits to be sufficiently verified.

The risk of starting the syncing process later is that the client might miss a block proposal during the "synced, but not deposits-synced" state.

arnetheduck:
Alright - this was also written at a time when not syncing deposits let to a lot of ugly `ERR` in the logs, when in fact, nothing was wrong, ie the user had just started syncing their eth1 node and was met with ERR all over the place, even though the system was correctly configured.

I think the logging is much better after the multi-el branch and we can close this issue for now - I think I'll open a separate one about the non-synced-UX in general, ie we should be clear about what is an error and what is simply the natural state of things: ideally, a correctly configured client would have NOT:ces as most when it's doing well and we should reserve the higher warning levels for actual problems (by being more refined about what consistutes a problem and what is simply "expected").

PR #4303, opened 2022-11-08
Support for obtaining deposit snapshots during trustedNodeSync

### LINKED ISSUE #4037: trustedNodeSync may hang forever due to lack of timeout handling in all REST requests

zah:
Other changes:

* More optimal search for TTD block.

* Add timeouts to all REST requests during trusted node sync. Fixes #4037

* Removed support for storing a deposit snapshot in the network metadata.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -17,9 +17,9 @@ import
   chronos, metrics, chronicles/timings, stint/endians2,
   web3, web3/ethtypes as web3Types, web3/ethhexstrings, web3/engine_api,
   eth/common/eth_types,
-  eth/async_utils, stew/[byteutils, objects, shims/hashes],
+  eth/async_utils, stew/[byteutils, objects, results, shims/hashes],
   # Local modules:
-  ../spec/[eth2_merkleization, forks, helpers],
+  ../spec/[deposit_snapshots, eth2_merkleization, forks, helpers],
   ../spec/datatypes/[base, phase0, bellatrix],
   ../networking/network_metadata,
   ../consensus_object_pools/block_pools_types,
@@ -30,8 +30,7 @@ from std/times import getTime, inSeconds, initTime, `-`
 from ../spec/engine_authentication import getSignedIatToken
 
 export
-  web3Types, deques, base,
-  beacon_chain_db.DepositContractSnapshot
+  web3Types, deques, base, DepositTreeSnapshot
 
 logScope:
   topics = "eth1"
@@ -124,6 +123,7 @@ type
     web3Urls: seq[string]
     eth1Network: Option[Eth1Network]
     depositContractAddress*: Eth1Address
+    depositContractDeployedAt: BlockHashOrNumber
     forcePolling: bool
     jwtSecret: Option[seq[byte]]
     blocksPerLogsRequest: uint64
@@ -809,6 +809,7 @@ proc onBlockHeaders(p: Web3DataProviderRef,
     p.web3.subscribeForBlockHeaders(blockHeaderHandler, errorHandler))
 
 proc pruneOldBlocks(chain: var Eth1Chain, depositIndex: uint64) =
+  ## Called on block finalization to delete old and now redundant data.
   let initialChunks = chain.finalizedDepositsMerkleizer.getChunkCount
   var lastBlock: Eth1Block
 
@@ -824,9 +825,11 @@ proc pruneOldBlocks(chain: var Eth1Chain, depositIndex: uint64) =
 
   if chain.finalizedDepositsMerkleizer.getChunkCount > initialChunks:
     chain.finalizedBlockHash = lastBlock.hash
-    chain.db.putEth2FinalizedTo DepositContractSnapshot(
+    chain.db.putDepositTreeSnapshot DepositTreeSnapshot(
       eth1Block: lastBlock.hash,
-      depositContractState: chain.finalizedDepositsMerkleizer.toDepositContractState)
+      depositContractState: chain.finalizedDepositsMerkleizer.toDepositContractState,
+      blockHeight: lastBlock.number,
+    )
 
     eth1_finalized_head.set lastBlock.number.toGaugeValue
     eth1_finalized_deposits.set lastBlock.depositCount.toGaugeValue
@@ -1057,12 +1060,6 @@ proc new*(T: type Web3DataProvider,
 
   return ok Web3DataProviderRef(url: web3Url, web3: web3, ns: ns)
 
-proc putInitialDepositContractSnapshot*(db: BeaconChainDB,
-                                        s: DepositContractSnapshot) =
-  let existingStart = db.getEth2FinalizedTo()
-  if not existingStart.isOk:
-    db.putEth2FinalizedTo(s)
-
 template getOrDefault[T, E](r: Result[T, E]): T =
   type TT = T
   get(r, default(TT))
@@ -1071,9 +1068,9 @@ proc init*(T: type Eth1Chain, cfg: RuntimeConfig, db: BeaconChainDB): T =
   let
     finalizedDeposits =
       if db != nil:
-        db.getEth2FinalizedTo().getOrDefault()
+        db.getDepositTreeSnapshot().getOrDefault()
       else:
-        default(DepositContractSnapshot)
+        default(DepositTreeSnapshot)
     m = DepositsMerkleizer.init(finalizedDeposits.depositContractState)
 
   T(db: db,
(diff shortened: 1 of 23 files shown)


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1066,6 +1040,49 @@ proc init*(T: type Eth1Monitor,
...
+                      depositContractDeployedAt: BlockHashOrNumber) {.async.} =
+  if db.hasDepositTreeSnapshot():
+    return
+
+  let providerRes = await Web3DataProvider.new(depositContractAddress, web3Url, jwtSecret)
+  if providerRes.isErr:
+    fatal "Failed to initialize web3 provider",
+          depositContract = depositContractAddress,
+          web3Url, err = providerRes.error
+    quit 1
```
arnetheduck: do we close the db when `quit` runs? also, this prevents starting eth1 and eth2 in "random" order which is not nice - ie I should be able to start the two services in any order and they should be "prone" to resume as soon as the "other" is up

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1066,6 +1040,49 @@ proc init*(T: type Eth1Monitor,
...
+    # If there is no DCS record at all, create one pointing to the deployment block
+    # of the deposit contract and insert it as a starting point.
+    let blk = try:
+      awaitWithRetries getBlock(provider, depositContractDeployedAt)
+    except CatchableError as e:
+      fatal "Failed to fetch deployment block",
+            depositContract = depositContractAddress,
+            deploymentBlock = $depositContractDeployedAt,
+            err = e.msg
+      quit 1
```
arnetheduck, line 1149: ditto... and how does this work when you start nimbus with an unsynced eth1?

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -123,6 +124,24 @@ proc toString*(kind: ValidatorFilterKind): string =
...
+      return RestApiResponse.jsonResponse(
+        RestDepositSnapshot(
+          finalized: snapshot.depositContractState.branch,
+          deposit_root: snapshot.getDepositRoot(),
+          deposit_count: snapshot.getDepositCountU64(),
+          execution_block_hash: snapshot.eth1Block,
+          execution_block_height: snapshot.blockHeight))
+    else:
+      # This can happen in a very short window after the client is started, but the
+      # snapshot record still haven't been upgraded in the database. Returning 404
```
arnetheduck, line 141: 404 is generally used for some semi-permanent conditions ("block doesn't exist at slot") - it would be better to return .. 503 (syncing) seems more appropriate

beacon_chain/trusted_node_sync.nim
```diff
@@ -229,6 +267,16 @@ proc doTrustedNodeSync*(
       databaseDir, head = shortLog(dbHead.get())
     (headSlot, dbHead.get())
 
+  # Fetch deposit snapshot.  This API endpoint is still optional.
+  let depositSnapshot = await fetchDepositSnapshot(client)
```
arnetheduck: I think this is better to put behind a flag (default false) while v0 is a thing

arnetheduck:
> Removed support for storing a deposit snapshot in the network metadata.

since no snapshot sources support the new snapshots, isn't this a regression? it seems reasonable to keep the old support until the new version has reached some reasonable penetration (basically when it's supported by https://eth-clients.github.io/checkpoint-sync-endpoints/)

beacon_chain/trusted_node_sync.nim
```diff
@@ -64,6 +71,30 @@ proc update(cache: var DbCache, blck: ForkySignedBeaconBlock) =
 proc isKnown(cache: DbCache, slot: Slot): bool =
   slot < cache.slots.lenu64 and cache.slots[slot.int].isSome()
 
+const
+  restRequestsTimeout = 5.seconds
```
arnetheduck: 10s at least but probably 30 .. ie TNS is fairly harmless when it stalls for a bit but annoying when it fails - specially for states it might take much longer to generate one

arnetheduck: actually.. 60s for states (also because they're large and might have limited constrained bandwidth to download on) and 30s for blocks (ditto on shared service providers, with future sidecars)

beacon_chain/rpc/rest_beacon_api.nim
zah: This should be addressed to the spec PR though:
https://github.com/ethereum/beacon-APIs/pull/245/files#diff-1c56455f973510b25b565b012b8bbd0d947a16e301a0807cae1c173a5d8af4c9R31

beacon_chain/trusted_node_sync.nim
zah: The failure is relatively benign. Only a `INFO` level message is printed.

zah:
> since no snapshot sources support the new snapshots, isn't this a regression? it seems reasonable to keep the old support until the new version has reached some reasonable penetration (basically when it's supported by https://eth-clients.github.io/checkpoint-sync-endpoints/)

The old support was baking a snapshot in the binary. I haven't been updating these snapshots in a while and every running client will have a more recent record in its database. New clients that didn't obtain a snapshot from trusted node sync will be initialized to the deposit contract deployment block which seems reasonable (i.e. this is the behavior of all other clients)

beacon_chain/trusted_node_sync.nim
zah: Indeed, this choice was a bit quick and arbitrary and I didn't revisit it. Will adjust.

arnetheduck:
> New clients that didn't obtain a snapshot from trusted node sync will be initialized to the deposit contract deployment block which seems reasonable (i.e. this is the behavior of all other clients)

this is the regression indeed - there's no support in servers for the eip, and there won't be fore the foreseeable future - since the code for reading the baked-in snapshot is still there, it seems reasonable to keep using it until the eip has reached penetration (which can take anything from days to months) - I'd prefer seeing this feature as a pure addon until the standard has settled (both EIP and beacon API are drafts at this point - it's quite possible there will be changes in the format meaning that we might have to go through a potentially painful DB migration, so a way out is needed)

beacon_chain/trusted_node_sync.nim
arnetheduck: nonetheless, it's a draft feature - also, disabling it completely is useful in case something breaks - this was part of the rationale for moving TNS to a finalized state alone: this is the minimal set of data needed to get started, and we should keep the possibility to start like this regardless of what the server supports - blocks and deposit snapshots are optional extras.

beacon_chain/trusted_node_sync.nim
```diff
@@ -229,6 +267,16 @@ proc doTrustedNodeSync*(
       databaseDir, head = shortLog(dbHead.get())
     (headSlot, dbHead.get())
 
+  # Fetch deposit snapshot.  This API endpoint is still optional.
+  let depositSnapshot = await fetchDepositSnapshot(client)
+  if depositSnapshot.isOk:
+    info "Writing deposit contracts snapshot",
```
arnetheduck: shouldn't the snapshot be validated against the state before writing to disk?

zah:
I still don't think this is a regression. The non-genesis snapshots were tested only in the testnets. The metadata for mainnet was never changed after the genesis event.

arnetheduck:
> I still don't think this is a regression.

the difference would be that we don't need to download deposits for the genesis validators, no?

beacon_chain/trusted_node_sync.nim
```diff
@@ -143,6 +182,21 @@ proc doTrustedNodeSync*(
...
+      if depositSnapshot.isOk:
+        if depositSnapshot.get.matches(getStateField(state[], eth1_data)):
+          info "Writing deposit contracts snapshot",
+               depositRoot = depositSnapshot.get.getDepositRoot(),
+               depositCount = depositSnapshot.get.getDepositCountU64
+          db.putDepositTreeSnapshot(depositSnapshot.get)
+        else:
+          warn "The downloaded deposit snapshot does not agree with the downloaded state"
+      else:
+        warn "Deposit tree snapshot was not imported", reason = depositSnapshot.error
```
arnetheduck, line 198: the download is optional and nothing will happen if it doesn't work except a small delay, so this is a notice at most

beacon_chain/trusted_node_sync.nim
```diff
@@ -178,7 +232,9 @@ proc doTrustedNodeSync*(
       var lastError: ref CatchableError
       for i in 0..<3:
         try:
-          return await client.getBlockV2(BlockIdent.init(slot), cfg)
+          return awaitWithTimeout(client.getBlockV2(BlockIdent.init(slot), cfg),
+                                  smallRequestsTimeout):
+            raise newException(CatchableError, "Request timed out")
```
arnetheduck, line 237: if there's a timeout, we should reattempt - that's what the retries are for

beacon_chain/eth1/eth1_monitor.nim
zah: This merely maintains the Status quo as Nimbus would already exit at start-up when paired with an EL that hasn't synced the deposit contract deployment block.

I've improved the behavior in the multi EL branch where the BN is able to wait for the EL to sync before the rest of the Eth1Monitor activities start.

beacon_chain/rpc/rest_beacon_api.nim
zah: Unfortunately, the spec has been merged before this suggestion was addressed

beacon_chain/trusted_node_sync.nim
zah: Changed to the suggested timeouts

beacon_chain/trusted_node_sync.nim
zah: I've added the suggested CLI flag, but I'm contemplating removing it since the API is now officially in the spec.

beacon_chain/eth1/eth1_monitor.nim
zah: I agree in general with the suggestions, but this requires more invasive changes to the Eth1Monitor that will be delivered in another PR.

beacon_chain/trusted_node_sync.nim
zah: There will be retry on this error. An exception is raised here, so it can be handled in the except block below, which will produce the following message:

```
warn "Retrying download of block", slot, err = "Request timed out"
```

I didn't want to duplicate the logic in multiple places and the cost of raising an exception is not that high in this infrequent code path.

beacon_chain/trusted_node_sync.nim
zah: Well, with the new CLI option the user explicitly requested the download of a deposit snapshot and it didn't work. I think a warning is reasonable. The user should check their configuration.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: how do you mean, shut down? we can run the BN without an EL entirely (in optmistic mode). Also, geth doesn't even start syncing without a BN. Since when do we require a synced EL?

arnetheduck: in fact, we're quite buggy in this aspect: we run the deposit monitoring even when we _know_ the EL is not synced - see https://github.com/status-im/nimbus-eth2/issues/4057 - this causes many support requests in general because of all the error / warning logging that ensues

zah: The deposit contract deployment block is part of the PoW history that Geth will sync on its own.
Otherwise, yes, the general behavior can be improved, but as I explained these are more invasive changes that are currently part of another PR.

PR #4304, opened 2022-11-09
Per-validator doppelganger protection.

cheatfate:


beacon_chain/validators/validator_pool.nim
```diff
@@ -215,6 +226,47 @@ iterator items*(pool: ValidatorPool): AttachedValidator =
   for item in pool.validators.values():
     yield item
 
+proc doppelgangerCheck*(validator: AttachedValidator,
```
arnetheduck, line 229: tests for this function?

beacon_chain/validators/validator_pool.nim
```diff
@@ -215,6 +226,47 @@ iterator items*(pool: ValidatorPool): AttachedValidator =
...
+      # Because validator was started at `GENESIS_EPOCH` we going to skip
+      # doppelganger detection.
+      ok(true)
+    else:
+      if activationEpoch.get() == epoch:
+        # Validator just activated so we going to skip doppelganger detection.
+        ok(true)
+      else:
+        let actualStartEpoch = max(startEpoch, broadcastEpoch)
+        if epoch - actualStartEpoch <= DOPPELGANGER_EPOCHS_COUNT:
```
arnetheduck: can `epoch < actualStartEpoch` happen? also, when COUNT == 2, this will be true for 3 epochs, is that intended?

arnetheduck: actually, it's normal that it happens: when we enable gossip traffic in slot 0 of epoch, there might still be attestations from epoch-1 bouncing around (attestations are propagated  in gossip for up to 32 slots)

beacon_chain/validators/validator_pool.nim
```diff
@@ -215,6 +226,53 @@ iterator items*(pool: ValidatorPool): AttachedValidator =
...
+    # from current wall epoch - it should not participate in the network.
+    err("Validator is not activated yet, or beacon node clock is invalid")
+  else:
+    if currentStartEpoch > epoch:
+      err("Validator is not started or broadcast is not started, or " &
+          "beacon node clock is invalid")
+    else:
+      let actEpoch = activationEpoch.get()
+      # startEpoch == broadcastEpoch == activateEpoch == GENESIS_EPOCH
+      if (currentStartEpoch == GENESIS_EPOCH) and (actEpoch == GENESIS_EPOCH):
```
arnetheduck: I don't think this condition is needed: basically, genesis is not special: what is important is that we _don't_ activate doppelganger detection for validators that were "recently" activated.

As an example, if a validator is activated in epoch 5 and it's currently epoch 5, we want the validator to start sending attestations at once (same if it's epoch 6 and validator was activated at 5): genesis is just a special case of this logic for epoch 0.

The idea here is that if we have two instances running and both wait 2 epochs after activation, doppelganger detection will not work _and_ we'll lose the funds. On activation, validators should start sending attestations without waiting.

zah: This would mean we don't have doppelganger protection during the activation which seems like a potentially non-zero risk scenario (the user accidentally deployed a key to multiple machines). We can potentially try to tackle this problem with randomness (i.e. the validators can send their first attestation after a random delay to increase the chance of noticing the attestation from the doppelganger).

I can also think of other potential solutions such as having a specialized topic for announcing your presence at the slot start during activation or evil hacks such as sending an attestation for a known deterministic value that would not produce slashing (i.e. the genesis block).

arnetheduck: > This would mean we don't have doppelganger protection during the activation

true, but adding randomness to the process I think is a separate feature/task to consider - ie as long as we have a fixed observation period there's no benefit waiting on activation

beacon_chain/validators/validator_pool.nim
```diff
@@ -215,6 +226,53 @@ iterator items*(pool: ValidatorPool): AttachedValidator =
...
+  ## perform any validator actions.
+  ##
+  ## Returns error, if its impossible to perform doppelganger check.
+  let
+    startEpoch = validator.startSlot.epoch() # startEpoch is epoch when /
+      # validator appeared in beacon_node.
+    activationEpoch = validator.activationEpoch # validator's activation_epoch
+    currentStartEpoch = max(startEpoch, broadcastEpoch)
+
+  if activationEpoch.isNone() or activationEpoch.get() > epoch:
```
zah, line 247: When is `activationEpoch` set to `None`? I see many places in the code that produce this value, but it's hard to understand the root cause. If a validator ends up being added with an `activationEpoch = None`, where is the code that will eventually get it out of this state?

arnetheduck: my guess would be when adding a validator key whose deposit has not yet been processed (and therefore it isn't part of the validator set)

PR #4315, opened 2022-11-11
`block_sim` runs capella by default

tersec:
This means it is proposing blocks, but the full infrastructure suitable for `nimbus_beacon_chain` to do so isn't there yet.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -485,7 +485,8 @@ func init(T: type AttestationCache, state: phase0.HashedBeaconState): T =
 
 func init(
     T: type AttestationCache,
-    state: altair.HashedBeaconState | bellatrix.HashedBeaconState,
+    state: altair.HashedBeaconState | bellatrix.HashedBeaconState |
+           capella.HashedBeaconState,
     cache: var StateCache): T =
   # Load attestations that are scheduled for being given rewards for
   let
@@ -582,11 +583,9 @@ proc getAttestationsForBlock*(pool: var AttestationPool,
     attCache =
       when state is phase0.HashedBeaconState:
         AttestationCache.init(state)
-      elif state is altair.HashedBeaconState or state is bellatrix.HashedBeaconState:
+      elif state is altair.HashedBeaconState or state is bellatrix.HashedBeaconState or
+           state is capella.HashedBeaconState:
         AttestationCache.init(state, cache)
-      elif state is capella.HashedBeaconState:
-        if true: raiseAssert $capellaImplementationMissing
-        default(AttestationCache)
       else:
         static: doAssert false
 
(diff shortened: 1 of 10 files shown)


beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -582,11 +583,9 @@ proc getAttestationsForBlock*(pool: var AttestationPool,
     attCache =
       when state is phase0.HashedBeaconState:
         AttestationCache.init(state)
-      elif state is altair.HashedBeaconState or state is bellatrix.HashedBeaconState:
+      elif state is altair.HashedBeaconState or state is bellatrix.HashedBeaconState or
```
arnetheduck, line 586: I wonder if `when forkOf(state) in StateFork.Altair..StateFork.Capella` would be something to figure out here

arnetheduck: or `{StateFork.Altair..StateFork.Capella}` or something similar - ie statically get an enum from the type then use that enum for convenient range access

etan-status: `state.toFork() >= BeaconStateFork.Altair`

arnetheduck: yeah, then we can also use `when state.toFork of ...`

etan-status: Ah, yeah. Point was that it already exists today.

etan-status: Actually, it's `typeof(state).toFork() >= ...`

See
```
func is_execution_block*(blck: SomeForkyBeaconBlock): bool =
  when typeof(blck).toFork >= BeaconBlockFork.Bellatrix:
    const defaultExecutionPayload =
      default(typeof(blck.body.execution_payload))
    blck.body.execution_payload != defaultExecutionPayload
  else:
    false
```

arnetheduck (approved):

beacon_chain/consensus_object_pools/attestation_pool.nim
arnetheduck: `>=`: we probably want to avoid >= so that we get an explicit compile-time error on a new fork - ie `case` is better

tersec: Yes, but not as part of this PR. It's a pattern that's probably worth having available though.

tersec: https://github.com/status-im/nimbus-eth2/blob/c585b0a5b1ae4d55af38ad7f4715ad455e791552/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim#L23-L30

shows what this might look like -- my next PR extends both of those to Capella, for the light client's purposes.

PR #4324, opened 2022-11-11
allow other response types in `SyncQueue`

etan-status:
`SyncQueue` currently only supports `Slot -> ForkedSignedBeaconBlock`. Extend to support other types as well to prepare for EIP4844 sidecars. `SyncManager` will receive similar treatment in a followup.

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -62,7 +62,7 @@ type
     notInSyncEvent: AsyncEvent
     rangeAge: uint64
     chunkSize: uint64
-    queue: SyncQueue[A]
+    queue: BeaconBlocksSyncQueue[A]
     syncFut: Future[void]
     blockVerifier: BlockVerifier
     inProgress*: bool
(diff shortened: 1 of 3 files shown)


beacon_chain/sync/sync_manager.nim
```diff
@@ -95,10 +95,11 @@ proc speed*(start, finish: SyncMoment): float {.inline.} =
 proc initQueue[A, B](man: SyncManager[A, B]) =
   case man.direction
   of SyncQueueKind.Forward:
-    man.queue = SyncQueue.init(A, man.direction, man.getFirstSlot(),
-                               man.getLastSlot(), man.chunkSize,
-                               man.getSafeSlot, man.blockVerifier, 1,
-                               man.ident)
+    man.queue =
+      initBeaconBlocksSyncQueue(A, man.direction, man.getFirstSlot(),
```
cheatfate, line 99: We are not using such names, this should be just `init`. It could be
```nim
BeaconBlocksSyncQueue.init(A, man.direction, man.getFirstSlot(),
                           man.getLastSlot(), man.chunkSize,
                           man.getSafeSlot, man.blockVerifier, 1,
                           man.ident)

File: beacon_chain/sync/sync_queue.nim [hunk 2]
```diff
@@ -644,9 +668,9 @@ proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
       else:
         let rewindSlot = sq.getRewindPoint(sq.outSlot, sq.getSafeSlot())
         warn "Got incorrect sync result in queue, rewind happens",
-             blocks_map = getShortMap(sq.readyQueue[0].request,
+             values_map = getShortMap(sq.readyQueue[0].request,
```
cheatfate, line 671: I'm not an english guru but `value` and `values_map` looks weird, what about `item` and `items_map`?

tests/test_sync_manager.nim
```diff
@@ -962,18 +967,21 @@ suite "SyncManager test suite":
 
     check waitFor(runTest()) == true
 
+  type Request = SyncRequest[SomeTPeer]
```
cheatfate, line 970: I would not recommend you to overload `Result` and `Request` words even for tests.

cheatfate:
My personal feeling that this PR complicates this simple data structure a lot. All this additional generic constructions and renames could cause issues. Also such big changes just modified existing tests but no new tests added, this looks suspicious.

beacon_chain/sync/sync_queue.nim
etan-status: `item` is already used to refer to the associated peer.

etan-status:
Seems like latest spec tends to bundle blocks with sidecars, making the sync more specialized. Stashing this for now.

PR #4329, opened 2022-11-15
quicker Geth startup in local testnet

etan-status:
Local testnet simulation currently waits 5 seconds when starting each individual Geth instance. Waiting a shorter amount saves almost a minute per minimum + mainnet CI finalization job.
Measured startup times per Geth: Linux ~100ms, macOS Intel ~300ms.


arnetheduck:
can we loop a `curl` to the RPC instead, so the rest starts "when it's ready"? ie it will be different on heavily loaded CI or PI vs fast machine

etan-status:
"when it's ready" is indicated by Geth with the socket file being written, so that should already be the case.
Or, you mean, repeating the `geth attach` line a couple times until successful? Seems a bit weird as an availability check but should work as well. Which one you prefer?

arnetheduck:
whatever works - ie as long as it's reliable.. I was thinking that doing curl and parsing the response of some health request would be one way but if the socket file is written "late enough", that's fine too

etan-status:
Waiting for the socket seems to work well-enough (and is also plausible, given that the OS manages the socket backlog).

The problem with `curl` is that it may connect to a different Geth if one is erroneously running on that port already (what we had before #4328). 

`geth attach` _should_ succeed when the file is there (so far, it's working out well, but of course limited sample size so far). If the socket file doesn't work, looping on `geth attach` is the next-best thing, but I'm not sure if that would have unpredictable timings like looping on `ssh` (which can introduce lagspikes due to some timeouts being in the minutes).

zah:
Merging this as it is for now. If we see problems again, we can try the `geth attach` treatment.

PR #4332, opened 2022-11-16
VC: blinded block publishing support

cheatfate:


beacon_chain/validator_client/api.nim
```diff
@@ -1772,6 +1772,200 @@ proc publishBlock*(
...
+    if res.isErr():
+      raise newException(ValidatorApiError, res.error())
+    return res.get().data
+
+  of ApiStrategyKind.Priority:
+    vc.firstSuccessSequential(
+      RestResponse[ProduceBlindedBlockResponse],
+      SlotDuration, {BeaconNodeRole.BlockProposalData},
+      produceBlindedBlock(it, slot, randao_reveal, graffiti)):
+      if apiResponse.isErr():
```
zah, line 1828: Isn't this block the same as the one above? Why don't you extract it in a helper function?

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -962,14 +967,69 @@ proc readValue*[BlockType: ForkedBeaconBlock](
...
+    let res =
+      try:
+        some(RestJson.decode(string(data.get()),
+                             phase0.BeaconBlock,
+                             requireAllFields = true,
+                             allowUnknownFields = true))
+      except SerializationError:
+        none[phase0.BeaconBlock]()
+    if res.isNone():
+      reader.raiseUnexpectedValue("Incorrect phase0 block format")
```
zah: Here and elsewhere, you can capture the error message from the `SerializationError` and provide a more detailed error response.

beacon_chain/spec/forks.nim
```diff
@@ -283,6 +295,29 @@ template init*(T: type ForkedSignedBeaconBlock, forked: ForkedBeaconBlock,
                                              root: blockRoot,
                                              signature: signature))
 
+template init*(T: type ForkedSignedBlindedBeaconBlock,
+               forked: ForkedBlindedBeaconBlock, blockRoot: Eth2Digest,
+               signature: ValidatorSig): T =
+  case forked.kind
```
zah, line 301: Why is this a template? It suffers from potential multiple evaluation of its parameters.

beacon_chain/validator_client/api.nim
```diff
@@ -1772,6 +1772,200 @@ proc publishBlock*(
...
+    BlockPublished = "Block was successfully published"
+    BlockBroadcasted = "Block not passed validation, but still published"
+    ErrorMessage = "Unable to publish block"
+
+  case strategy
+  of ApiStrategyKind.First, ApiStrategyKind.Best:
+    let res = block:
+      vc.firstSuccessParallel(RestPlainResponse, SlotDuration,
+                              {BeaconNodeRole.BlockProposalPublish}):
+        case data.kind
```
zah, line 1875: Ditto (Isn't this block the same as the one below? Why don't you extract it in a helper function?)

beacon_chain/validator_client/block_service.nim
```diff
@@ -25,123 +151,162 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
...
-           blck = shortLog(beaconBlock),
-           signature = shortLog(signature),
-           validator = shortLog(validator),
-           wall_slot = currentSlot
+      let signingRoot = beaconBlocks.normalBlock.get().signing_root
+      vc.attachedValidators[]
+      .slashingProtection
+      .registerBlock(vindex, validator.pubkey, slot, signingRoot)
+
+  if notSlashable.isOk():
```
zah: If the blinded block is slashable, but the normal block is not slashable, the logic here will refuse to send the normal block.

An alternative way would be to write the checks in turn. If there is a blinded block, check it's slashing protection status and potentially send it. If any of the above fails, check the slashing protection status of the normal block and potentially send it.

beacon_chain/validator_client/block_service.nim
```diff
@@ -25,123 +151,162 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
...
-    if res:
-      let delay = vc.getDelay(slot.block_deadline())
-      beacon_blocks_sent.inc()
-      beacon_blocks_sent_delay.observe(delay.toFloatSeconds())
-      notice "Block published", blockRoot = shortLog(blockRoot),
-             blck = shortLog(beaconBlock), signature = shortLog(signature),
-             validator = shortLog(validator)
+  let notSlashable =
+    if beaconBlocks.blindedBlock.isSome():
+      let signingRoot = beaconBlocks.blindedBlock.get().signing_root
```
arnetheduck: fwiw, recomputing the signing root is likely cheaper than introducing an extra copy to create the new type - ie it's just a hash of a few fields, specially when the beacon block root is already computed

beacon_chain/validator_client/block_service.nim
arnetheduck: a block is slashable if another block with a different signing root has been sent already in the same slot - ie the only time the scenario you describe would happen would be iff the exact same non-blinded block had been sent already (for example during an extremely fast restart where the same duty is done twice _and_  the same block is produced twice)

beacon_chain/validator_client/block_service.nim
```diff
@@ -25,123 +151,162 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
...
+      vc.attachedValidators[]
+      .slashingProtection
+      .registerBlock(vindex, validator.pubkey, slot, signingRoot)
```
arnetheduck: the order of operations seems wrong here - we should first add the block to the slashing protection, _then_ sign it - ie it's the signing that we want to protect so that under no circumstances do we sign several blocks - even if in theory the code doesn't have a path today where the signed block is not added to the slashing db, the two things should happen "together" with protection leading then immediately signing

arnetheduck: notably, doing it this way will also enable a flow like:
```
let blockRoot = if haveBlinded: htr(blinded) else: htr(full)
if slashprot.add(blockRoot, other data).isOk): signAndPublish(...)
```
thanks to the construction of the blinded block, the eventual "unblinding" results in the same block root at the end - eventually, we might even create something like `toBlinded` which enableds working with blinded block types alone in relevant flows, though that is likely  for later

cheatfate: Current algorithm is just more advanced version of your algorithm.
```nim
if blindedEnabled:
  spawn produceBlindedBlockWithSignature()
spawn produceNormalBlockWithSignature()

let notSlashable =
  if blindedEnabled and blindedReady:
    slashing.registerBlock(blindedBlock)
  else:
    slashing.registerBlock(normalBlock)

if notSlashable:
  if blindedEnabled and blindedReady:
    publishBlindedBlock()
  else:
    publishNormalBlock()
```

beacon_chain/validator_client/block_service.nim
cheatfate: i'm not sure i understand you, i think you missed an algorithm at all.

beacon_chain/validator_client/block_service.nim
cheatfate: At this stage there could be only one block available either blinded or normal. If we was not able to produce blinded block we will try to produce normal block.

beacon_chain/validator_client/api.nim
cheatfate: This block is the same like one above just because actual scoring is missing, as soon as scoring this blocks will become different.

beacon_chain/spec/forks.nim
cheatfate: @zah all the previous `init` declarations in this file are templates, why should i change it to something else? Do you want to make it `func` or `proc`?

zah: Looks like there are no side-effects, so a `func` would do.

cheatfate: @zah but there will be just single one `init` declaration which will be `func` all others will be `template`. Maybe its better to address this issue in different PR?

beacon_chain/validator_client/block_service.nim
arnetheduck: Here's how it should be:

```nim
if blindedEnabled:
  spawn produceBlindedBlock()
spawn produceNormalBlock()

let notSlashable =
  if blindedEnabled and blindedReady:
    slashing.registerBlock(blindedBlock)
  else:
    slashing.registerBlock(normalBlock)

if notSlashable:
  if blindedEnabled and blindedReady:
    signAndPublishBlindedBlock()
  else:
    signAndPublishNormalBlock()
```

Notice two differences: we only ever sign one of the normal and blinded blocks, and we don't ever sign slashable blocks

beacon_chain/spec/signatures.nim
```diff
@@ -110,6 +110,10 @@ func get_block_signature*(
 
   blsSign(privkey, signing_root.data)
 
+func get_block_signature*(signing_root: Eth2Digest,
```
arnetheduck: we should not introduce this function without also using it in the beacon node - ie the flows should be the same for both vc and bn in this regard, so they call the same signing function (otherwise, it's easy to confuse signing root and block root since they have the same type)

beacon_chain/validator_client/block_service.nim
arnetheduck: what I mean is that `SlashingForkedSignedBeaconBlock` as a type likely isn't a good performance tradeoff, even if signing root has to be computed twice - also, it would be good to keep the block signing flow the same in BN and VC as much as possible - this includes using the same roots in `get_block_signature` and friends (ie either both use signing root or both use block root, but not a mix)

beacon_chain/spec/forks.nim
zah: I took a look at the whole file and there was one more template that needed such a fix. Since the change is completely benign, I've committed it in the branch here.

beacon_chain/validator_client/block_service.nim
cheatfate: I have noticed that you treat signature process as process which could not fail. But its not true. Remote signature brings one more point of failure, so its why i'm introducing new helpers and its why i think my approach is much more resilient, because if for some reason remote signature of blinded block will fail you still will be able to publish signed normal block.

arnetheduck: we can consider failing signatures in a separate PR / issue, but for now this behavior should be the same on BN and VC, meaning first slashing prot, then signature.

arnetheduck (approved):

Issue #4290, opened 2022-11-07
Tighten sync committee duty polling


arnetheduck:
The current VC polls sync committee duties for current and next epoch:

https://github.com/status-im/nimbus-eth2/blob/a1201f13a482a84e4b6d230c1d648b6a49023d9a/beacon_chain/validator_client/duties_service.nim#L445

This is both too much and not enough - we should:

* not poll the same period twice (even if the epoch is different)
* start polling the "next" period 4 epochs before the period starts (and consequently register those duties - currently this happens 1 epoch before which might not be enough

cheatfate:
Fixed in #4333.

PR #4333, opened 2022-11-16
Address #4290

cheatfate:
Address #4290


beacon_chain/validator_client/duties_service.nim
```diff
@@ -11,6 +11,7 @@ import common, api, block_service
 
 const
   ServiceName = "duties_service"
+  SUBSCRIPTION_LOOKAHEAD_EPOCHS* = 4'u64
```
arnetheduck, line 14: ideally there would be a single shared constant between BN and VC implementation

PR #4334, opened 2022-11-17
cancel non-started finalization testnets on cancel

etan-status:
When cancelling the `minimal` CI finalization testnet on Jenkins, the internal Jenkins cancellation exception is being caught by `catchError` and the `mainnet` CI finalization testnet will still run. Replacing the logic with `try` / `finally` instead, and moving the `archiveArtifacts` step to the post-build (to run that even on cancellation / failure).

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -89,6 +89,7 @@ pipeline {
         }
       }
     }
+  }
 
   post {
     always {
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -78,19 +78,19 @@ pipeline {
       stages {  /* parallel builds of minimal / mainnet not yet supported */
         stage('minimal') {
           steps { timeout(26) {
-            launchLocalTestnet('minimal')
+            sh 'make local-testnet-minimal'
           } }
           post { always {
-            collectLocalTestnetArtifacts('minimal')
+            sh 'tar cjf local-testnet-minimal.tar.gz local-testnet-minimal/*.txt'
           } }
         }
 
         stage('mainnet') {
           steps { timeout(62) {
-            launchLocalTestnet('mainnet')
+            sh 'make local-testnet-mainnet'
           } }
           post { always {
-            collectLocalTestnetArtifacts('mainnet')
+            sh 'tar cjf local-testnet-mainnet.tar.gz local-testnet-mainnet/*.txt'
           } }
         }
       }
@@ -108,14 +108,6 @@ pipeline {
       )
     }
   }
-}
-
-def launchLocalTestnet(String name) {
-  sh "make local-testnet-${name}"
-}
-
-def collectLocalTestnetArtifacts(String name) {
-  sh "tar cjf local-testnet-${name}.tar.gz local-testnet-${name}/*.txt"
 }
 
 def isMainBranch() {
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -90,15 +90,11 @@ pipeline {
       }
     }
 
-    stage('Upload') {
-      steps { timeout(5) {
-        archiveArtifacts('*.tar.gz')
-      } }
-    }
-  }
-
   post {
     always {
+      catchError { timeout(5) {
+        archiveArtifacts(artifacts: '*.tar.gz', allowEmptyArchive: true)
+      } }
       cleanWs(
         disableDeferredWipeout: true,
         deleteDirs: true
@@ -109,15 +105,11 @@ pipeline {
 
 def launchLocalTestnet(String name) {
   /* We want to mark job as failed, but save the results. */
-  catchError(
-    message: "Local ${name} testnet finalization failure!",
-    buildResult: 'FAILURE',
-    stageResult: 'FAILURE'
-  ) {
+  try {
     sh "make local-testnet-${name}"
+  } finally {
+    sh "tar cjf local-testnet-${name}.tar.gz local-testnet-${name}/*.txt"
   }
-  /* Archive test results regardless of outcome. */
-  sh "tar cjf local-testnet-${name}.tar.gz local-testnet-${name}/*.txt"
 }
 
 def isMainBranch() {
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile
```diff
@@ -109,15 +106,11 @@ pipeline {
 
 def launchLocalTestnet(String name) {
   /* We want to mark job as failed, but save the results. */
```
jakubgs: This makes this comment incorrect, since the error will not actually be caught by the job like this.

ci/Jenkinsfile
```diff
@@ -89,16 +89,13 @@ pipeline {
...
-      } }
-    }
   }
 
   post {
     always {
+      catchError { timeout(5) {
+        archiveArtifacts(artifacts: '*.tar.gz', allowEmptyArchive: true)
+      } }
       cleanWs(
```
jakubgs: We can move the `cleanWs` call to a separate `cleanup` post stage instead of bunching things up in `always`.

ci/Jenkinsfile
```diff
@@ -89,16 +89,13 @@ pipeline {
...
-    stage('Upload') {
-      steps { timeout(5) {
-        archiveArtifacts('*.tar.gz')
-      } }
-    }
   }
 
   post {
     always {
+      catchError { timeout(5) {
```
jakubgs: What's the point of this `catchError`?

ci/Jenkinsfile
jakubgs: Are we sure we want to ignore errors from failed testnet launches? Seems wrong.

ci/Jenkinsfile
```diff
@@ -109,15 +106,11 @@ pipeline {
 
 def launchLocalTestnet(String name) {
   /* We want to mark job as failed, but save the results. */
-  catchError(
```
jakubgs, line 112: You know, you could just set `catchInterruptions: false`:

>`catchInterruptions` : boolean (optional)
>If true, certain types of exceptions that are used to interrupt the flow of execution for Pipelines will be caught and handled by the step. If false, those types of exceptions will be caught and immediately rethrown. Examples of these types of exceptions include those thrown when a build is manually aborted through the UI and those thrown by the timeout step. Defaults to true.

https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#catcherror-catch-error-and-set-build-result-to-failure

No need for `try`/`finally`, since that ignores errors entirely.

ci/Jenkinsfile
etan-status: So that workspace still gets cleaned up even when archiving artifacts times out.

ci/Jenkinsfile
etan-status: With `catchInterruptions: false`, the logs won't get saved in case of timeout though.

ci/Jenkinsfile
etan-status: The stage will still fail on error. It is `try` / `finally`, not `try` / `catch`

ci/Jenkinsfile
jakubgs: Makes sense. Then if you use `cleanup` for `cleanWs` you don't need that in `always.

ci/Jenkinsfile
jakubgs: Are you sure about that? I don't think that's true. Can we test it?

ci/Jenkinsfile
jakubgs: Hmmm, I see, that makes sense. Then if use of `finally` doesn't hide the error that's fine.

jakubgs: If it does not then we can handle that by using `updateBuildStatus('FAILURE')` in a `catch`.

etan-status: I have switched it to `post`, this way the syntax is consistent with the other post step for cleaning up workspace

ci/Jenkinsfile
```diff
@@ -108,15 +111,10 @@ pipeline {
...
-  /* We want to mark job as failed, but save the results. */
-  catchError(
-    message: "Local ${name} testnet finalization failure!",
-    buildResult: 'FAILURE',
-    stageResult: 'FAILURE'
-  ) {
-    sh "make local-testnet-${name}"
-  }
-  /* Archive test results regardless of outcome. */
+  sh "make local-testnet-${name}"
```
jakubgs: If the function just makes one `sh` call, then the function is pointless. Just use `sh` directly.

ci/Jenkinsfile
```diff
@@ -108,15 +111,10 @@ pipeline {
...
+}
+
+def collectLocalTestnetArtifacts(String name) {
   sh "tar cjf local-testnet-${name}.tar.gz local-testnet-${name}/*.txt"
```
jakubgs: Same here, if the function just does one call, it's useless, just use `sh` directly in the stage.

ci/Jenkinsfile
jakubgs: Yeah that's cleaner.

ci/Jenkinsfile
etan-status: alright

jakubgs (approved):
Looks good.

ci/Jenkinsfile
```diff
@@ -77,23 +77,26 @@ pipeline {
...
+          } }
         }
       }
-    }
-
-    stage('Upload') {
-      steps { timeout(5) {
-        archiveArtifacts('*.tar.gz')
-      } }
+      post { always { timeout(5) {
```
jakubgs, line 97: Small nitpick but maybe:
```suggestion
      post {
        cleanup { timeout(5) {
```

etan-status: Shouldn't matter unless multiple `post` actions are defined. `always` runs at the beginning of `post`, `cleanup` at the end. As the rest works using `always`, prob better to keep as is for now rather than finding out that the assumption is incorrect ^^

jakubgs: We already use `cleanup` in desktop. It works as expected.

etan-status: Ah, great to know for next time (for the `cleanupWorkspace` action it already was on `always` before this PR btw)

PR #4339, opened 2022-11-21
periodically log LC sync progress

### LINKED ISSUE #4048: Periodic progress information
No "Slot start" message is printed, thus it's hard to tell if the light client is doing anything, and what its latest view of the network is (what it thinks the head is etc

etan-status:
Adds a "Slot start" log to the LC that behaves similar to BN to inform the user that the light client is doing something, and to indicate the latest view of the network (finalized / optimistic).

### DIFF
--- a/beacon_chain/nimbus_light_client.nim
+++ b/beacon_chain/nimbus_light_client.nim
@@ -220,6 +220,61 @@ programMain:
 
     blocksGossipState = targetGossipState
 
+  proc onSlot(wallTime: BeaconTime, lastSlot: Slot) =
+    let
+      wallSlot = wallTime.slotOrZero()
+      expectedSlot = lastSlot + 1
+      delay = wallTime - expectedSlot.start_beacon_time()
+
+      finalizedHeader = lightClient.finalizedHeader
+      optimisticHeader = lightClient.optimisticHeader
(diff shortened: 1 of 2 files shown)


tersec (approved):

beacon_chain/nimbus_light_client.nim
```diff
@@ -220,6 +220,61 @@ programMain:
...
+      nextSlot = curSlot + 1
+      timeToNextSlot = nextSlot.start_beacon_time() - getBeaconTime()
+    while true:
+      await sleepAsync(timeToNextSlot)
+
+      let
+        wallTime = getBeaconTime()
+        wallSlot = wallTime.slotOrZero()
+
+      onSlot(wallTime, curSlot)
```
tersec, line 272: Presumably this varied usage of `wallSlot` (for the timer) and `curSlot` as the sort of expected-wallslot is intentional? It creates a certain policy implied in catching up with slots if there's clock/timer skew

etan-status: Yeah, it is the same in beacon node. The `curSlot` passed here is the previous slot at which `onSlot` is called. It is used to compute `delay` in the log

PR #4343, opened 2022-11-22
Bugfix: The VC was producing invalid sync committee contributions

zah:
Since the sync committee duties are no longer updated on every slot and previously the sync committee aggregators selection proofs were generated during the duties update, this now resulted in the client using stale selection proofs (they must be generated at each slot).

The fix consists of moving the selection proof generation logic in a different function which is properly executed on each slot.

Other changes:

* The logtrace tool has been enhanced with a framework for adding new simpler log aggregation and analysis algorithms. The default CI testnet simulation will now ensure that the blocks in the network have reasonable sync committee participation.

### DIFF
--- a/beacon_chain/validator_client/sync_committee_service.nim
+++ b/beacon_chain/validator_client/sync_committee_service.nim
@@ -27,16 +27,16 @@ type
 
 proc serveSyncCommitteeMessage*(service: SyncCommitteeServiceRef,
                                 slot: Slot, beaconBlockRoot: Eth2Digest,
-                                duty: SyncDutyAndProof): Future[bool] {.
+                                duty: SyncCommitteeDuty): Future[bool] {.
      async.} =
   let
     vc = service.client
     fork = vc.forkAtEpoch(slot.epoch)
     genesisValidatorsRoot = vc.beaconGenesis.genesis_validators_root
-    vindex = duty.data.validator_index
+    vindex = duty.validator_index
     subcommitteeIdx = getSubcommitteeIndex(
-      duty.data.validator_sync_committee_index)
-    validator = vc.getValidator(duty.data.pubkey).valueOr: return false
+      duty.validator_sync_committee_index)
+    validator = vc.getValidator(duty.pubkey).valueOr: return false
     message =
       block:
         let res = await getSyncCommitteeMessage(validator, fork,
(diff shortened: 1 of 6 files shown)


beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -203,38 +202,81 @@ proc serveContributionAndProof*(service: SyncCommitteeServiceRef,
...
+        vc.beaconGenesis.genesis_validators_root,
+        slot,
+        subCommitteeIdx)
+
+    slotSignatureReqs.add(future)
+    validators.add((validator, subCommitteeIdx))
 
-  if len(contributionItems) > 0:
+  try:
+    await allFutures(slotSignatureReqs)
```
arnetheduck, line 230: fwiw, this call is not really needed - ie one can add the future to `validators` then await them one by one without the extra call to `allFutures` (since we wait for all, we wait for the first one too etc)

beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -203,38 +202,81 @@ proc serveContributionAndProof*(service: SyncCommitteeServiceRef,
...
+  except CancelledError as exc:
+    var pendingCancel: seq[Future[void]]
+    for future in slotSignatureReqs:
+      if not(future.finished()):
+        pendingCancel.add(future.cancelAndWait())
+    await allFutures(pendingCancel)
+    raise exc
+
+  var
+    contributionsFuts: array[SYNC_COMMITTEE_SUBNET_COUNT,
```
arnetheduck, line 240: shouldn't we create two contributions if two validators are in the same subnet and making contributions? ie low probablility but possible

beacon_chain/validator_client/sync_committee_service.nim
arnetheduck: technically, this code could also benefit from being factored in a different way, where getting selection proof then signing contribution is done as one async function - then each part "naturally" can complete without each validator stopping and waiting for _all_ others

arnetheduck: in fact, factoring this way could be quite important - imagine you're running many validators on a slow remote signer and the operation as a whole times out (because the slot ends) - you want the ones that managed to get a selection proof to proceed to sign and publish their contribution without waiting for the others

arnetheduck: (we can leave it for a future PR but ..)

beacon_chain/validator_client/sync_committee_service.nim
zah: Both validators will send the same contribution content. The logic here ensures we are not generating it twice.

arnetheduck: ok, can we comment that here?

beacon_chain/validator_client/sync_committee_service.nim
zah: You are missing some of the details (see my other comment), but there is some validity to these ideas. Will revisit in a follow-up PR.

PR #4344, opened 2022-11-22
implement more missing capella functionality

tersec:


beacon_chain/validators/validator_duties.nim
```diff
@@ -367,7 +367,7 @@ proc getExecutionPayload[T](
   template empty_execution_payload(): auto =
     withState(proposalState[]):
       when stateFork >= BeaconStateFork.Capella:
-        raiseAssert $capellaImplementationMissing
+        raiseAssert $capellaImplementationMissing & ": beacon_chain/validators/validator_duties.nim: getExecutionPayload"
```
arnetheduck, line 370: are any of these asserts rechable from `REST`? ie can a request be made via rest that triggers an assert? if yes, that should really be a non-assert

tersec: Not unless the BN is set to `CAPELLA_FORK_EPOCH` being not-`FAR_FUTURE_EPOCH`, which isn't the case for any existing supported network, mainnet, testnet, or devnet.

PR #4345, opened 2022-11-23
Fix doppelganger detection issue.

cheatfate:
Fix missing activationEpoch setup.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -138,34 +138,56 @@ proc getAttachedValidator(node: BeaconNode,
   node.attachedValidators[].getValidator(pubkey)
 
 proc getAttachedValidator(node: BeaconNode,
-                          state_validators: auto,
-                          idx: ValidatorIndex): AttachedValidator =
-  if uint64(idx) < state_validators.lenu64:
-    let validator = node.getAttachedValidator(state_validators[idx].pubkey)
-    if validator != nil and validator.index != some(idx):
-      # Update index, in case the validator was activated!
-      notice "Validator activated", pubkey = validator.pubkey, index = idx
-      validator.index = Opt.some(idx)
-    validator
-  else:
-    warn "Validator index out of bounds",
-      idx, validators = state_validators.len
-    nil
-
-proc getAttachedValidator(node: BeaconNode,
-                          idx: ValidatorIndex): AttachedValidator =
+                          idx: ValidatorIndex,
+                          slot: Slot): Opt[AttachedValidator] =
   let key = node.dag.validatorKey(idx)
-  if key.isSome():
-    let validator = node.getAttachedValidator(key.get().toPubKey())
-    if validator != nil and validator.index != Opt.some(idx):
-      # Update index, in case the validator was activated!
-      notice "Validator activated", pubkey = validator.pubkey, index = idx
-      validator.index = Opt.some(idx)
-    validator
+  if key.isNone():
+    return Opt.some(AttachedValidator(nil))
+
+  let validator = node.getAttachedValidator(key.get().toPubKey())
+  if isNil(validator):
+    return Opt.some(AttachedValidator(nil))
+
+  if validator.index != Opt.some(idx):
+    # Update index and activation_epoch, in case the validator was activated!
+    notice "Validator activated", pubkey = validator.pubkey, index = idx,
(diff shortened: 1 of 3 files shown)


tersec (approved):

beacon_chain/validators/validator_duties.nim
```diff
@@ -945,8 +967,9 @@ proc handleAttestations(node: BeaconNode, head: BlockRef, slot: Slot) =
       epochRef.shufflingRef, slot, committee_index)
 
     for index_in_committee, validator_index in committee:
-      let validator = node.getAttachedValidator(validator_index)
-      if validator == nil:
+      let validator = node.getAttachedValidator(validator_index, slot).valueOr:
+        continue
+      if isNil(validator):
```
arnetheduck, line 972: when would this return nil in a non-error Result?

beacon_chain/validators/validator_duties.nim
```diff
@@ -138,34 +138,56 @@ proc getAttachedValidator(node: BeaconNode,
...
+  if not(node.config.doppelgangerDetection):
+    Opt.some(validator)
   else:
-    warn "Validator key not found",
-      idx, head = shortLog(node.dag.head)
-    nil
+    let
+      doppelgangerDetection = node.processor[].doppelgangerDetection
+      broadcastStartEpoch = doppelgangerDetection.broadcastStartEpoch
+      res = validator.doppelgangerCheck(slot.epoch(), broadcastStartEpoch)
```
arnetheduck, line 165: it's a bit strange that we do the doppelganger check in a function called `getAttachedValidator` - such a function could be used for other purposes than performing duties - I think it would make sense to split it into two parts, one `getValidator` and one `getActiveValidator` which also does the doppelganger check - also, the distinction between `nil` and `none` return is somewhat unusual / confusing - as a convention when we return `Opt` we typically no longer want `nil` - is the difference significant? else `Opt.none` could be returned for all failure cases (not found by key, doppel active, etc)

arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
cheatfate: `Opt.some(nil)` - there is no such validator
`Opt.some(validator)` - validator has been found and it passed doppelganger protection.
`Opt.none()` - validator has been found but it not yet passed doppelganger detection.

etan-status: Isn't `Opt.none()` implemented internally as `Opt.some(nil)`?

From stew:
>     ## One difference is `Option[ref|ptr T]` which disallows `nil` - `Opt[T]`
>     ## allows an "ok" result to hold `nil` - this can be useful when `nil` is
>     ## a valid outcome of a function, but increases complexity for the caller.

etan-status: Ah, misread. That's only the case for `Option`. In `Opt` it works fine.

arnetheduck: > `Opt.some(nil)` - there is no such validator `Opt.some(validator)` - validator has been found and it passed doppelganger protection. `Opt.none()` - validator has been found but it not yet passed doppelganger detection.

After careful reading of the code, I understood this also - but as is, the code is hard to understand and read - ie the name of the function and the difference between opt.none and opt.some(nil) is .. confusing / arbitrary - as suggested, this could would be easier to read if factored differently, for example by having an enum for no-such and not-passed, then naming the function accordingly.

beacon_chain/validators/validator_duties.nim
```diff
@@ -138,34 +138,56 @@ proc getAttachedValidator(node: BeaconNode,
...
+  if isNil(validator):
+    return Opt.some(AttachedValidator(nil))
+
+  if validator.index != Opt.some(idx):
+    # Update index and activation_epoch, in case the validator was activated!
+    notice "Validator activated", pubkey = validator.pubkey, index = idx,
+                                  activation_epoch = slot.epoch()
+    validator.index = Opt.some(idx)
+    if validator.activationEpoch.isNone():
+      validator.activationEpoch = Opt.some(slot.epoch())
```
arnetheduck, line 157: this is a bug actually - the activation epoch is the epoch when the validator is activated in the state (ie some time after the deposit) - it's the first epoch in which they are supposed to perform duties, but this function may be called with arbitrary slots)

PR #4349, opened 2022-11-24
Fix the REST API navigation typo

cheny0:
This is a minor fix to the API table of content in the Nimbus guide.

### DIFF
--- a/docs/the_nimbus_book/mkdocs.yml
+++ b/docs/the_nimbus_book/mkdocs.yml
@@ -97,7 +97,7 @@ nav:
       - 'logging.md'
       - 'validator-client-options.md'
       - 'validator-monitor.md'
-    - "REST API:s":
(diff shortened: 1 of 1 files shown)


arnetheduck (approved):

arnetheduck:
This is fine, though it would probably be good to review other such abbreviations as well, so that we gain consistency in use

jorge-campo:
> This is fine, though it would probably be good to review other such abbreviations as well, so that we gain consistency in use

We have a [writing style guide](https://write.status.im/) that we'll adapt progressively to cover infrastructure and API documentation. This will help us to standardize writing styles across different docs and writers.

cheny0:
> This is fine, though it would probably be good to review other such abbreviations as well, so that we gain consistency in use

Will edit other such abbreviations in follow-up RPs

PR #4350, opened 2022-11-24
more capella

tersec:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -1201,6 +1213,7 @@ proc readValue*(reader: var JsonReader[RestJson],
   if voluntary_exits.isNone():
     reader.raiseUnexpectedValue("Field `voluntary_exits` is missing")
 
+  discard $capellaImplementationMissing & ": autodetect via added field"
```
arnetheduck, line 1216: what's this?

tersec: make sure this isn't missed, in a compile-time-enforced-way, later.

The final step for all this will be removing the `capellaImplementationMissing` variable, and this makes sure that won't work unless there's truly nothing left to do with it.

arnetheduck (approved):

PR #4357, opened 2022-11-25
bump `nim-eth`, extend empty block fallback for Capella

etan-status:
Implements `emptyPayloadToBlockHeader` for Capella. https://github.com/status-im/nim-eth/pull/562

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -18,7 +18,7 @@ import
   # Status libraries
   stew/[bitops2, byteutils, endians2, objects, saturation_arith],
   chronicles,
-  eth/eip1559, eth/common/[eth_types, eth_types_rlp],
+  eth/eip1559, eth/common/[eth_types, eth_types_rlp], eth/trie/[db, hexary],
   # Internal
   ./datatypes/[phase0, altair, bellatrix, capella],
   "."/[eth2_merkleization, forks, ssz_codec]
@@ -29,6 +29,7 @@ export
   forks, eth2_merkleization, ssz_codec
 
 type
+  ExecutionWithdrawal = eth_types.Withdrawal
   ExecutionBlockHeader = eth_types.BlockHeader
 
   FinalityCheckpoints* = object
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/helpers.nim
```diff
@@ -360,32 +361,101 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
...
+    difficulty     : default(DifficultyInt),
+    blockNumber    : payload.block_number.u256,
+    gasLimit       : cast[GasInt](payload.gas_limit),
+    gasUsed        : cast[GasInt](payload.gas_used),
+    timestamp      : fromUnix(int64.saturate payload.timestamp),
+    extraData      : payload.extra_data.asSeq,
+    mixDigest      : payload.prev_randao, # EIP-4399 `mixDigest` -> `prevRandao`
+    nonce          : default(BlockNonce),
+    fee            : some payload.base_fee_per_gas,
+    withdrawalsRoot: withdrawalsRoot)
```
tersec, line 425: Does it work to unconditionally include this, regardless of fork or engine API version?

beacon_chain/validators/validator_duties.nim
```diff
@@ -388,9 +388,10 @@ proc getExecutionPayload[T](
 
   template empty_execution_payload(): auto =
     withState(proposalState[]):
-      when stateFork >= BeaconStateFork.Bellatrix:
-        build_empty_execution_payload[typeof forkyState.data, T](
-          forkyState.data, feeRecipient)
+      when stateFork >= BeaconStateFork.Capella:
+        raiseAssert $capellaImplementationMissing
```
tersec, line 392: Doesn't this PR add such an implementation?

beacon_chain/spec/helpers.nim
```diff
@@ -360,32 +361,101 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
...
+func toExecutionWithdrawal*(
+    withdrawal: capella.Withdrawal): ExecutionWithdrawal =
+  ExecutionWithdrawal(
+    index: withdrawal.index,
+    validatorIndex: withdrawal.validatorIndex,
+    address: EthAddress withdrawal.address.data,
+    amount: gweiToWei withdrawal.amount)
+
+# https://eips.ethereum.org/EIPS/eip-4895
+proc computeWithdrawalsTrieRoot*(
```
tersec, line 377: Presumably one of the RLP or `initHexaryTrie` or `newMemoryDB` calls renders this non-`func`?

etan-status: Yes

beacon_chain/spec/helpers.nim
etan-status: Yes, the RLP serializer checks whether the field is a `some` or a `none`, and emits the correct block hash. 

I tested the logic using the test script here:
https://github.com/status-im/nim-eth/pull/562#issuecomment-1326830428

etan-status: For Capella blocks you need engine API v2 calls. For pre-Capella blocks, you may also use v1 but v2 also works.

beacon_chain/validators/validator_duties.nim
etan-status: The problem here is that the various `when` return different types. So, either the entire function needs to change to ForkyState, or there needs to be a ForkedExecutionPayload. Too big to just extend the empty payload fallback. 

Also, regardless of empty vs non-empty, we need to pass expected withdrawals into this function so that engine_getPayload or the empty block fallback are correct. Again too big for the scope of this PR.

tersec: Plan is to make the whole `makeBeaconBlockForHeadAndSlot` call tree generic, avoid yet another forked type.

tersec: But, sure, agree with scope of PR.

beacon_chain/spec/helpers.nim
tersec: Fortunately, once ELs support Capella in any production capacity, then Nimbus can start getting rid of the v1 calls then. Presumably gating on Gnosis capella fork schedule, but fundamentally see no reason to keep using v1 calls, ever, once the ELs support v2 calls.

tersec (approved):

PR #4360, opened 2022-11-25
fix order of MEV signed block root assignment/checking

tersec:


beacon_chain/validators/message_router_mev.nim
```diff
@@ -83,11 +83,10 @@ proc unblindAndRouteBlockMEV*(
         getFieldNames(typeof(signedBlock.message.body)))
       signedBlock.message.body.execution_payload = unblindedPayload.data.data
 
-      if signedBlock.root != hash_tree_root(blindedBlock.message):
-        return err("Unblinded block doesn't match blinded block SSZ root")
-
       signedBlock.root = hash_tree_root(signedBlock.message)
 
+      doAssert signedBlock.root == hash_tree_root(blindedBlock.message)
```
arnetheduck, line 88: this has (not insignificant) runtime cost for release builds too, ie that of hashing the block again - worth it?

arnetheduck (approved):

beacon_chain/validators/message_router_mev.nim
tersec: Yes, because it's tricky to test MEV in a reasonable way, and barring that, it should fail safe.

That condition not holding is about the most castastrophic plausible non-slashing condition that can occur.

One alternate approach, and this makes sense anyway to add these tests, is to get the `copyFields` aspects factored and tested well enough to really trust them. That would also allow one to confidently switch them to use `assign2`.

arnetheduck (approved):

PR #4362, opened 2022-11-25
more strict read-only database mode

arnetheduck:
* avoid creating pre-altair backwards compatibility tables
* allow running ncli_db era export without above tables present
* drop unused pre-altair backwards compatibility tables
* run benchmark on read-ronly database
* fix running benchmark from genesis

needs https://github.com/status-im/nim-eth/pull/563

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -289,7 +289,7 @@ proc close*(s: var DbSeq) =
   reset(s)
 
 proc add*[T](s: var DbSeq[T], val: T) =
-  doAssert(distinctBase(s.insertStmt) != nil, "database closed or read-only")
+  doAssert(distinctBase(s.insertStmt) != nil, "database closed or table not preset")
   var bytes = SSZ.encode(val)
   s.insertStmt.exec(bytes).expectDb()
   inc s.recordCount
@@ -299,7 +299,7 @@ template len*[T](s: DbSeq[T]): int64 =
 
 proc get*[T](s: DbSeq[T], idx: int64): T =
   # This is used only locally
-  doAssert(distinctBase(s.selectStmt) != nil, $T & " not found at index " & $(idx))
+  doAssert(distinctBase(s.selectStmt) != nil, $T & " table not present for read at " & $(idx))
 
   let resultAddr = addr result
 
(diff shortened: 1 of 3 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -240,45 +240,56 @@ template expectDb(x: auto): untyped =
...
+    ? db.exec("""
+      CREATE TABLE IF NOT EXISTS '""" & name & """'(
+        id INTEGER PRIMARY KEY,
+        value BLOB
+      );
+    """)
+    true
+  if hasTable:
+    let
+      insertStmt = db.prepareStmt(
```
zah, line 258: Elsewhere you are checking whether this value is `nil` and declaring that the database is read-only if it is. Here, you are not setting it to `nil` for read-only databases.

beacon_chain/beacon_chain_db.nim
```diff
@@ -288,6 +299,8 @@ template len*[T](s: DbSeq[T]): int64 =
 
 proc get*[T](s: DbSeq[T], idx: int64): T =
   # This is used only locally
+  doAssert(distinctBase(s.selectStmt) != nil, $T & " not found at index " & $(idx))
```
zah: This error message is quite confusing:  "ImmutableValidatorData not found at index 19".

You probably meant to say something along the lines of "The table for storing ImmutableValidatorData is missing".

zah (approved):

beacon_chain/beacon_chain_db.nim
arnetheduck: the message is actually correct:ish - a "missing" table is treated the same as an empty table or one that doesn't hold enough many items so from that POV, the message makes sense - it's a bit like nim does with an empty seq being `nil` underneath

zah: This is an assert producing a crash with a stack trace. I think such a message should be as precise as possible.
What you describe makes sense if the application was recovering from the error and treating a missing table and an incomplete table in the same way.

beacon_chain/beacon_chain_db.nim
arnetheduck: a4f516f8edaf7fe5cac13ecc32ec257e299f93f1

beacon_chain/beacon_chain_db.nim
arnetheduck: a4f516f8edaf7fe5cac13ecc32ec257e299f93f1

PR #2962, opened 2021-10-06
extend REST config with more spec constants

etan-status:
So far, the REST/RPC config response did not include all spec constants.
The spec for `/eth/v1/config/spec` defines that the response should
include constants for all hard forks known by the beacon node. This
patch extends the corresponding responses to include more constants.

PR #4365, opened 2022-11-28
Types and scaffolding for EIP-4844 

henridf:
This commit adds the EIP-4844 spec types, and fills in
scaffolding/boilerplate for the use of these types across the repo.

None of the actual EIP-4844 logic is introduced yet.

This follows the pattern used by @tersec when introducing Capella (https://github.com/status-im/nimbus-eth2/pull/4276).


beacon_chain/spec/datatypes/eip4844.nim
```diff
@@ -0,0 +1,321 @@
...
+  SomeBeaconBlockBody* =
+    BeaconBlockBody |
+    SigVerifiedBeaconBlockBody |
+    TrustedBeaconBlockBody
+
+  BlockParams = object
+    parentHash*: string
+    timestamp*: string
+
+  BoolReturnValidRPC = object
```
tersec: `BoolReturnValidRPC` and `BoolReturnSuccessRPC` aren't used anywhere -- https://github.com/status-im/nimbus-eth2/pull/4366 removes them from bellatrix/capella, and they shouldn't be here either. It was an oversight to leave them in

.gitmodules
```diff
@@ -165,7 +165,7 @@
 	branch = master
 [submodule "vendor/eth2-networks"]
 	path = vendor/eth2-networks
-	url = https://github.com/eth-clients/eth2-networks.git
+	url = https://github.com/henridf/eth2-networks.git
```
tersec: This isn't mergeable as such -- okay for a draft, but what does this change?

tersec: ah, https://github.com/henridf/eth2-networks/commit/64324476b843b88302814d85faa856d000b7e3ab

beacon_chain/spec/presets.nim
```diff
@@ -159,8 +161,11 @@ when const_preset == "mainnet":
     # Capella
     CAPELLA_FORK_VERSION: Version [byte 0x03, 0x00, 0x00, 0x00],
     CAPELLA_FORK_EPOCH: Epoch(uint64.high),
+    # eip4844
+    EIP4844_FORK_VERSION: Version [byte 0x04, 0x00, 0x00, 0x00],
```
tersec, line 166: It's failing CI currently because
```
/home/runner/work/nimbus-eth2/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/fatal.nim(49, 5) Error: unhandled exception: /home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/spec/datatypes/base.nim(964, 16) `distinctBase(forkVersions[i]) != distinctBase(forkVersions[j])`  
```
presumably there's another conflicting fork version around for this

henridf: Tests pass fine when running locally; I think this is due to my eth2-networks fork not getting picked up. (not sure why that isn't the case 😬 )

tersec: Another option is to leave out the `SHARDING_FORK_VERSION` from that comparison list -- presumably, `EIP4844_FORK_VERSION` is effectively its replacement for the next few months. Then it wouldn't need the `eth2-networks` fork bump. As long as `SHARDING_FORK_EPOCH` is set to still be checked to be always `FAR_FUTURE_EPOCH`, it's safe enough because it can't ever be triggered.

Yet another option is to fix `eth2-networks` upstream to be 4844-correct. This depends how much commitment and clarify there is from the EF, but is a bit more fundamentally correct and doesn't need workarounds in Nimbus.

beacon_chain/spec/datatypes/base.nim
```diff
@@ -958,7 +958,7 @@ func checkForkConsistency*(cfg: RuntimeConfig) =
   let forkVersions =
     [cfg.GENESIS_FORK_VERSION, cfg.ALTAIR_FORK_VERSION,
      cfg.BELLATRIX_FORK_VERSION, cfg.CAPELLA_FORK_VERSION,
-     cfg.SHARDING_FORK_VERSION]
+     cfg.EIP4844_FORK_VERSION, cfg.SHARDING_FORK_VERSION]
```
tersec: Are the `EIP4844_FOO` and `SHARDING_FOO` forks ever going to meaningfully coexist?

beacon_chain/spec/presets.nim
henridf: Yep, I've started looking into the second option. I'll open that PR and see how that goes, and if it doesn't look likely to be merged soon, I'll go for your first option that I hadn't thought of. 👍

beacon_chain/spec/datatypes/base.nim
```diff
@@ -958,7 +958,7 @@ func checkForkConsistency*(cfg: RuntimeConfig) =
   let forkVersions =
```
tersec, line 959: Around here in the code, there's a static check for `SHARDING_FORK_EPOCH == FAR_FUTURE_EPOCH`. Would probably be a bit safer to start out with that also for `EIP4844_FORK_EPOCH == FAR_FUTURE_EPOCH`, because it can still run most of the EF consensus spec tests, and once we're a bit more confident things are fine, then it can be separately allowed to be non-`FAR_FUTURE_EPOCH`.

beacon_chain/spec/datatypes/base.nim
henridf: I'm not really sure what SHARDING_FORK_VERSION is supposed to represent at this stage

beacon_chain/spec/presets.nim
henridf: > I'll open that PR 

https://github.com/eth-clients/eth2-networks/pull/81

beacon_chain/spec/datatypes/base.nim
henridf: 👍 71081b473083889d61dec6d9a22ca44d6fb359d0

beacon_chain/rpc/rest_config_api.nim
```diff
@@ -153,6 +153,10 @@ proc installConfigApiHandlers*(router: var RestRouter, node: BeaconNode) =
             "0x" & $cfg.CAPELLA_FORK_VERSION,
           CAPELLA_FORK_EPOCH:
             Base10.toString(uint64(cfg.CAPELLA_FORK_EPOCH)),
+          EIP4844_FORK_VERSION:
```
tersec: Making this visible unconditionally is a externally visible behavior change, and one which isn't necessary for the type system. Presumably this would make sense when it's not at `FAR_FUTURE_EPOCH`, though.

One way to keep the compile-time aspect of the other `raiseAssert`/etc "TODOs" is to use something like `discard $eip4844ImplementationMissing`, while just sidestepping this for the moment.

I'm not sure, in hindsight, having `SHARDING_FORK_VERSION` and `SHARDING_FORK_EPOCH` here makes sense either, but if nothing else it's already included in shipping versions, so I'd leave potentially removing that as separate from an EIP4844 PR.

henridf: Addressed in 64e348b1ecc2e577df920c0986e8d295f47d04b9.

I'm happy to put up a quick PR to remove SHARDING_FORK_{VERSION,EPOCH} if we want to do that? (they've been in for a while, #2962).

beacon_chain/spec/datatypes/eip4844.nim
poxelman: Leave reply

beacon_chain/rpc/rest_config_api.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4385

tersec: merged, should unblock this PR

henridf: merged unstable into this branch

.gitmodules
tersec: is this still necessary after `checkForkConsistency` doesn't include `SHARDING_FORK_VERSION`? If not, it's be better to continue pointing to the canonical `eth-clients` project repository.

beacon_chain/spec/datatypes/eip4844.nim
tersec: This is marked as `resolved`, but I still see `BoolReturnValidRPC` and `BoolReturnSuccessRPC` being added?

henridf: oops, my bad for the mixup, fixed in 5fa61d605c532cb18c923ca062456efceb74b8a4

.gitmodules
henridf: shouldn't be necessary any more - reverted

henridf:
Switched to regular (non-draft) PR. Thanks @tersec for the comments and for the unblocking PR!

ncli/resttest-rules.json
```diff
@@ -2610,7 +2610,7 @@
...
...","SECONDS_PER_SLOT":"","SECONDS_PER_ETH1_BLOCK":"","MIN_VALIDATOR_WITHDRAWABILITY_DELAY":"","SHARD_COMMITTEE_PERIOD":"","ETH1_FOLLOW_DISTANCE":"","INACTIVITY_SCORE_BIAS":"","INACTIVITY_SCORE_RECOVERY_RATE":"","EJECTION_BALANCE":"","MIN_PER_EPOCH_CHURN_LIMIT":"","CHURN_LIMIT_QUOTIENT":"","PROPOSER_SCORE_BOOST":"","DEPOSIT_CHAIN_ID":"","DEPOSIT_NETWORK_ID":"","DEPOSIT_CONTRACT_ADDRESS":"","BLS_WITHDRAWAL_PREFIX":"","ETH1_ADDRESS_WITHDRAWAL_PREFIX":"","DOMAIN_BEACON_PROPOSER":"","DOMAIN_BEACON_ATTESTER":"","DOMAIN_RANDAO":"","DOMAIN_DEPOSIT":"","DOMAIN_VOLUNTARY_EXIT":"","DOMAIN_SELECTION_PROOF":"","DOMAIN_AGGREGATE_AND_PROOF":"","TIMELY_SOURCE_FLAG_INDEX":"","TIMELY_TARGET_FLAG_INDEX":"","TIMELY_HEAD_FLAG_INDEX":"","TIMELY_SOURCE_WEIGHT":"","TIMELY_TARGET_WEIGHT":"","TIMELY_HEAD_WEIGHT":"","SYNC_REWARD_WEIGHT":"","PROPOSER_WEIGHT":"","WEIGHT_DENOMINATOR":"","DOMAIN_SYNC_COMMITTEE":"","DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF":"","DOMAIN_CONTRIBUTION_AND_PROOF":"","TARGET_AGGREGATORS_PER_COMMITTEE":"","RANDOM_SUBNETS_PER_VALIDATOR":"","EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION":"","ATTESTATION_SUBNET_COUNT":"","TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE":"","SYNC_COMMITTEE_SUBNET_COUNT":""}}]
```
tersec: It's not passing Jenkins CI because of
```
----------------------------------------------------------------------------------------------------------------^M
TEST                                         STATUS              ERROR  ELAPSED             MESSAGE             ^M
----------------------------------------------------------------------------------------------------------------
/eth/v1/config/spec#236                      ValidationError     --R    44ms576us753ns                          
```
from `resttest0/client_log.txt`

One can run this locally by `make nimbus_beacon_node resttest && make restapi-test`; probably there are other combinations that work too.

beacon_chain/rpc/rest_config_api.nim
```diff
@@ -272,6 +272,8 @@ proc installConfigApiHandlers*(router: var RestRouter, node: BeaconNode) =
              "/eth/v1/config/spec") do () -> RestApiResponse:
     return RestApiResponse.response(cachedConfigSpec, Http200,
                                     "application/json")
+  # EIP4844_FORK_EPOCH and EIP4844_FORK_VERSION not yet in config
```
tersec, line 275: This is fine in isolation, it can be marked as a `TODO` essentially via `eip4844ImplementationMissing`, but then `resttest-rules.json` needs to be consistent to pass Jenkins CI.

henridf: 8e3c9eb

ncli/resttest-rules.json
henridf: ah thanks! I didn't this make target.

tersec (approved):

Issue #4353, opened 2022-11-24
Validator Client with 7000 validators produces a large amount of "Block proposal awaited" log messages


zah:
Observe the same slot number:

```
{"lvl":"DBG","ts":"2022-11-24 15:14:04.018+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"491ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.018+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"391ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.018+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"351ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.018+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"501ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.018+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"441ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.018+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"601ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.019+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"370ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.019+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"391ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.019+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"450ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.019+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"481ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.019+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"501ns","service":"attestation_service"}
{"lvl":"DBG","ts":"2022-11-24 15:14:04.019+00:00","msg":"Block proposal awaited","slot":4399570,"duration":"461ns","service":"attestation_service"}
[...]

PR #4368, opened 2022-11-29
Address #4353

cheatfate:
Address #4353.

### DIFF
--- a/beacon_chain/validator_client/block_service.nim
+++ b/beacon_chain/validator_client/block_service.nim
@@ -426,17 +426,39 @@ proc addOrReplaceProposers*(vc: ValidatorClientRef, epoch: Epoch,
           res
       vc.proposers[epoch] = ProposedData.init(epoch, dependentRoot, tasks)
 
-proc waitForBlockPublished*(vc: ValidatorClientRef, slot: Slot) {.async.} =
+proc waitForBlockPublished*(vc: ValidatorClientRef,
+                            slot: Slot, timediff: TimeDiff) {.async.} =
   ## This procedure will wait for all the block proposal tasks to be finished at
-  ## slot ``slot``
-  let pendingTasks =
-    block:
-      var res: seq[Future[void]]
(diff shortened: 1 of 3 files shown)


beacon_chain/validator_client/block_service.nim
```diff
@@ -426,17 +426,39 @@ proc addOrReplaceProposers*(vc: ValidatorClientRef, epoch: Epoch,
...
+        res
+  logScope:
+    start_time = startTime
+    pending_tasks = len(pendingTasks)
+    slot = slot
+    timediff = timediff
+
   if len(pendingTasks) > 0:
-    await allFutures(pendingTasks)
+    let waitTime = (start_beacon_time(slot) + timediff) - vc.beaconClock.now()
```
arnetheduck, line 451: this can be `vc.beaconClock.fromNow(...)` (makes these easier to grep for since most code uses it)

beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -332,23 +332,11 @@ proc produceAndPublishContributions(service: SyncCommitteeServiceRef,
...
-  let
-    vc = service.client
-    startTime = Moment.now()
+                                         duties: seq[SyncCommitteeDuty]) {.
+     async.} =
+  let vc = service.client
 
-  try:
-    let timeout = syncCommitteeMessageSlotOffset
-    await vc.waitForBlockPublished(slot).wait(nanoseconds(timeout.nanoseconds))
```
arnetheduck, line 343: btw, why move the timeout into waitForBlockPublished? could it stay here? that seemed more easy to read

arnetheduck (approved):

beacon_chain/validator_client/sync_committee_service.nim
cheatfate: To avoid code duplication and why not? Old version forces to do more work on exception handling, with just single procedure which do not perform any cleaning it is much more easy.

PR #4369, opened 2022-11-29
verify `CAPELLA_FORK_EPOCH == FAR_FUTURE_EPOCH` for supported networks

tersec:
Also clean up some unused string constants and update some spec references.

No runtime changes.

### DIFF
--- a/beacon_chain/networking/network_metadata.nim
+++ b/beacon_chain/networking/network_metadata.nim
@@ -209,6 +209,7 @@ elif const_preset == "mainnet":
     for network in [
         mainnetMetadata, praterMetadata, ropstenMetadata, sepoliaMetadata]:
       checkForkConsistency(network.cfg)
+      doAssert network.cfg.CAPELLA_FORK_EPOCH == FAR_FUTURE_EPOCH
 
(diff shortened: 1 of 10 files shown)


beacon_chain/networking/network_metadata.nim
```diff
@@ -209,6 +209,7 @@ elif const_preset == "mainnet":
     for network in [
         mainnetMetadata, praterMetadata, ropstenMetadata, sepoliaMetadata]:
       checkForkConsistency(network.cfg)
+      doAssert network.cfg.CAPELLA_FORK_EPOCH == FAR_FUTURE_EPOCH
```
arnetheduck, line 212: since these are loaded at compile time, it *should* be possible to make this check compile-time also

tersec: The whole thing is in a `static:` block:
https://github.com/status-im/nimbus-eth2/blob/df54470b13112ea853ea0b0bfb5f85d50fa0f413/beacon_chain/networking/network_metadata.nim#L208-L212

arnetheduck: ah.

PR #4371, opened 2022-11-29
persist LC sync progress across restarts

### LINKED ISSUE #4046: Resume with last root when starting
So that it doesn't have to be supplied on every restart, we should persist basic light client information such as the current head / finalized checkpoint, so that it can resume automatically after a restart as long as the chain is finalizing.

etan-status:
Persist the latest finalized header and sync committee across restarts of `nimbus_light_client` to avoid redoing time-consuming bootstrap step.

### DIFF
--- a/beacon_chain/light_client_db.nim
+++ b/beacon_chain/light_client_db.nim
@@ -149,7 +149,7 @@ proc getSyncCommittee*(
     res.expect("SQL query OK")
     try:
       return ok SSZ.decode(syncCommittee, altair.SyncCommittee)
-    except SszError:
+    except SszError as exc:
       error "LC store corrupted", store = "syncCommittees",
         period, exc = exc.msg
       return err()
--- a/beacon_chain/light_client_db.nim
+++ b/beacon_chain/light_client_db.nim
@@ -88,9 +88,9 @@ proc getLatestFinalizedHeader*(db: LightClientDB): Opt[BeaconBlockHeader] =
     res.expect("SQL query OK")
     try:
       return ok SSZ.decode(header, BeaconBlockHeader)
-    except MalformedSszError, SszSizeMismatchError:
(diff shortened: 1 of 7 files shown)


beacon_chain/light_client_db.nim
```diff
@@ -0,0 +1,187 @@
...
+  store.putStmt.dispose()
+
+proc getLatestFinalizedHeader*(db: LightClientDB): Opt[BeaconBlockHeader] =
+  var header: seq[byte]
+  for res in db.headers.getStmt.exec(
+      LightClientHeaderKind.Finalized.int64, header):
+    res.expect("SQL query OK")
+    try:
+      return ok SSZ.decode(header, BeaconBlockHeader)
+    except MalformedSszError, SszSizeMismatchError:
```
zah: The list of exceptions here seems peculiarly specific. Handling `SszError` would be less fragile in the face of potential refactorings in the SSZ library.

beacon_chain/light_client_db.nim
```diff
@@ -0,0 +1,187 @@
...
+
+proc getSyncCommittee*(
+    db: LightClientDB, period: SyncCommitteePeriod): Opt[altair.SyncCommittee] =
+  doAssert period.isSupportedBySQLite
+  var syncCommittee: seq[byte]
+  for res in db.syncCommittees.getStmt.exec(period.int64, syncCommittee):
+    res.expect("SQL query OK")
+    try:
+      return ok SSZ.decode(syncCommittee, altair.SyncCommittee)
+    except MalformedSszError, SszSizeMismatchError:
```
zah: ditto

zah (approved):

beacon_chain/light_client_db.nim
etan-status: It's consistent with the existing `beacon_chain_db_light_client`.

PR #4374, opened 2022-11-30
avoid port re-use across unit test runs

etan-status:
In Jenkins CI we run two instances of unit tests concurrently. This can trigger CI failure when the same port numbers are re-used by the different test instances. Fixed one more issue of this by allowing user configuration of the base port number.


zah:
That's great! Thanks for tracking this down.

PR #4376, opened 2022-11-30
don't access Result.error when not error

tersec:


beacon_chain/validators/validator_duties.nim
```diff
@@ -755,10 +755,15 @@ proc proposeBlockMEV(
     # the block failed to validate and integrate into the DAG, which for the
     # purpose of this return value, is equivalent. It's used to drive Beacon
     # REST API output.
+    let errMsg =
+      if unblindedBlockRef.isOk:
+        unblindedBlockRef.error
```
etan-status, line 760: doesn't this access `Result.error` when not error?

tersec: yeah, https://github.com/status-im/nimbus-eth2/pull/4376/commits/59167879a9fcd80c2759bba002f6fef9b4daf472

tersec: https://github.com/status-im/nimbus-eth2/pull/4376/commits/30e0b594e6efa61e75afd84d7ac10dff6b132be1

PR #4378, opened 2022-11-30
deprecate `/eth2/beacon_chain/req/beacon_blocks_by_{range,root}/1/`

tersec:
https://github.com/ethereum/consensus-specs/pull/3014

Update/recreate https://github.com/status-im/nimbus-eth2/pull/4207


beacon_chain/sync/sync_protocol.nim
```diff
@@ -269,139 +269,27 @@ p2pProtocol BeaconSync(version = 1,
       response: MultipleChunksResponse[
         phase0.SignedBeaconBlock, MAX_REQUEST_BLOCKS])
       {.async, libp2pProtocol("beacon_blocks_by_range", 1).} =
```
arnetheduck: I think at this point we're better off removing it completely which will give clients that try to access it an error at the libp2p level

tersec: https://github.com/status-im/nimbus-eth2/pull/4378/commits/777c3adab0bdb63bc4ab0f9c6813849a9f1cfabe

arnetheduck (approved):

PR #4381, opened 2022-12-01
Switch to official doppelganger detection implementation.

cheatfate:
Switch VC to use /eth/v1/validator/liveness/{epoch} endpoint instead of nimbus one.

### DIFF
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -4,7 +4,7 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import std/[typetraits, sets]
+import std/[typetraits, sets, sequtils]
 import stew/[results, base10], chronicles
 import ".."/[beacon_chain_db, beacon_node],
        ".."/networking/eth2_network,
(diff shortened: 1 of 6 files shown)


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -853,3 +853,66 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+              if wallEpoch == FAR_FUTURE_EPOCH:
+                wallEpoch
+              else:
+                wallEpoch + 1
+            prevEpoch = get_previous_epoch(wallEpoch)
+          if (res < prevEpoch) or (res > nextEpoch):
+            return RestApiResponse.jsonError(Http400, InvalidEpochValueError,
+                    "Requested epoch is more than one epoch from current epoch")
+          res
+      indexList =
```
arnetheduck, line 878: reading the index list should really be a function, given how we read the same list in several places (attestations etc)

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2364,6 +2364,43 @@ proc readValue*(reader: var JsonReader[RestJson],
...
+  writer.writeField("is_live", value.is_live)
+  writer.endRecord()
+
+proc readValue*(reader: var JsonReader[RestJson],
+                value: var RestLivenessItem) {.
+     raises: [SerializationError, IOError, Defect].} =
+  var index: Option[ValidatorIndex]
+  var isLive: Option[bool]
+
+  for fieldName in readObjectFields(reader):
```
arnetheduck, line 2382: why does this need a custom reader? isn't it simple enough?

PR #4400, opened 2022-12-07
add local testnet output to gitignore

etan-status:
Latest local testnet output folder was not yet in gitignore.


arnetheduck:
why is this a PR even?

PR #4403, opened 2022-12-08
change all :s to s

cheny0:
Now all ":s" in the Nimbus guide are replaced with "s".

### DIFF
--- a/docs/the_nimbus_book/src/eth1.md
+++ b/docs/the_nimbus_book/src/eth1.md
@@ -81,7 +81,7 @@ It is safe to start the beacon node even if the execution client is not yet full
 
 ### 3. Pass the URL and JWT secret to Nimbus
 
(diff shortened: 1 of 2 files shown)


arnetheduck:
there's one more btw:
```
[arnetheduck@praeceps src]$ ag ":s"
eth1.md
84:The `--web3-url` option informs the beacon node how to connect to the execution client - both `http://` and `ws://` URL:s are supported.
103:    You can pass one or more `--web3-url` parameters to the node as long as they share JWT secret. Any additional web3 url:s will be used for backup, should the first one become unavailable:

index.md
18:* Fast [Beacon](./rest-api.md) and [KeyManager](./keymanager-api.md) API:s with extensions
```

cheny0:
> 

Thanks, it's fixed.

PR #4410, opened 2022-12-09
add eip4844 fork choice tests and remove test vector workarounds

tersec:
https://github.com/ethereum/consensus-spec-tests/releases/tag/v1.3.0-alpha.1-hotfix.2 fixes the issues that were being seen, so now EIP4844 tests can be/are straightforwardly implemented, in their entirety as much as the other forks.


tersec:
closing, something's going wrong with the test fixture downloading (works locally), but better to test it in isolation than tie it to otherwise useful changes

PR #4414, opened 2022-12-11
bump nim-eth2-scenarios to fix EIP4844 consensus spec test vectors

tersec:


tersec:
```
2022-12-11T20:52:25.5011771Z     Unhandled exception: cannot open: D:\a\nimbus-eth2\nimbus-eth2\vendor\nim-eth2-scenarios\tests-v1.3.0-alpha.1-hotfix.2\minimal\bellatrix\light_client\sync\pyspec_tests\supply_sync_committee_from_past_update\update_0xd0275ac71f8143bd665d9...
```
too long for Windows, exceeds `MAX_PATH` of 260 characters.

Will probably pass on these hotfixes, since while the underlying issues should be fixed, they're likely several which would all have to be fixed, in various libraries, and not ordinarily relevant so long as an alpha.2 is released soon enough.

PR #4418, opened 2022-12-12
consolidate slot sanity checks across forks

tersec:


beacon_chain/spec/forks.nim
```diff
@@ -266,17 +266,28 @@ template toFork*[T: eip4844.BeaconState | eip4844.HashedBeaconState](
...
-#   T(kind: BeaconStateFork.Altair, altairData: data)
-# template init*(T: type ForkedHashedBeaconState, data: bellatrix.HashedBeaconState): T =
-#   T(kind: BeaconStateFork.Bellatrix, bellatrixData: data)
-# template init*(T: type ForkedHashedBeaconState, data: capella.HashedBeaconState): T =
-#   T(kind: BeaconStateFork.Capella, capellaData: data)
-# template init*(T: type ForkedHashedBeaconState, data: eip4844.HashedBeaconState): T =
-#   T(kind: BeaconStateFork.EIP4844, eip4844Data: data)
+# TODO using "ref T" doesn't work -- some kind of reference to type being
+# treated differently than type?
+func init*(T: type ForkedHashedBeaconState, data: phase0.BeaconState):
```
arnetheduck: these should either be `new(T: type Forked...` or `init(T: type (ref Forked)...` - ie `new` means "introduce `ref` to `T`

arnetheduck (approved):

beacon_chain/spec/forks.nim
tersec: Well, this particular issue was https://github.com/nim-lang/Nim/issues/21086

tersec: And, being parameterized by `T` but returning a `ref T` was/is the goal -- avoid the `TODO these cause stack overflows due to large temporaries getting allocated` hazard. If preferred, I can keep them isolated in `test_fixture_sanity_slots`, which is their only user, but they'd find use at some point in other test fixtures. Would also be reasonable to put them in `fixtures_utils` and revert `beacon_chain_spec/forks` to the status quo ante, with the `TODO` comment block, and at least it's inarguably not worse than before.

But it's necessary to have some helper functions somewhere like this, to avoid the need for spelling out `Phase0.HashedBeaconState`, etc in the consolidated test fixtures.

arnetheduck: > being parameterized by T but returning a ref T was/is the goal

then it should be called `new` - it's no longer an `init` at that point (init should return its own type or possibly Opt/Result, but not introduce `ref`): https://status-im.github.io/nim-style-guide/language.objconstr.html

tersec: https://github.com/status-im/nimbus-eth2/pull/4418/commits/f1993e368da1e4e8cd859cf29d3840fec366444c

arnetheduck: btw, we detect stack overflow in CI these days so we're a bit more safe against sudden unwanted copies

PR #4420, opened 2022-12-12
validate EL block hash in EL simulation

etan-status:
When simulating EL with `--optimistic` flag, perform block hash check.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -357,14 +357,28 @@ proc storeBlock*(
     self.consensusManager.quarantine[].addUnviable(signedBlock.root)
     return err((VerifierError.UnviableFork, ProcessingStatus.completed))
 
-  if NewPayloadStatus.noResponse == payloadStatus and not self[].optimistic:
-    # Disallow the `MissingParent` from leaking to the sync/request managers
-    # as it will be descored. However sync and request managers interact via
-    # `processBlock` (indirectly). `validator_duties` does call `storeBlock`
-    # directly, so is exposed to this, but only cares about whether there is
-    # an error or not.
-    return err((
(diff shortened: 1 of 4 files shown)


etan-status:
Doing `--optimistic` sync on Mainnet to validate the computation is correct

arnetheduck:
nice - with this in place, we could enable optimistic mode by default and eventually hide the flag

etan-status:
Not sure about default. We would still need to keep it as a flag at least to disable though, in case there is an EL bug that leads to INVALID becoming finalized due to majority using broken EL, to protect the rest from accidentally syncing that chain. After some more time, new deposits / inactivity leaks would kick those others out and there could be new finality on the correct chain.

arnetheduck:
> INVALID becoming finalized due to majority using broken EL

an actively connected EL still has priority over this flag but it's kind of useless anyway since it will be returning SYNCING which is equivalent to it not being there at all. If we ever run into the situation where INVALID spills into the finalized chain, we have bigger problems

etan-status:
Yes, you are right. It would need a new `--pessimistic` flag that guarantees awaiting for block execution before importing each individual block. SYNCING would be interpreted same as no response and be retried until there is a final verdict.

`--pessimistic` would also need widespread support for ELs actually honoring sync requests for blocks not close to wall clock.

arnetheduck:
we've thought about adding `--pessimistic` indeed - that's how a "lock-step" sync would work

zah (approved):

arnetheduck:
we good here?

etan-status:
```
 peers: 74 ❯ finalized: 8d829bee:168489 ❯ head: 362f79ac:168491:5 ❯ time: 168491:5 (5391717) ❯ sync: synced/opt  
```

full Mainnet genesis sync with `--optimistic`, with an extra assert on block hash validation error

PR #4421, opened 2022-12-13
Backfill only up to MIN_EPOCHS_FOR_BLOCK_REQUESTS blocks

arnetheduck:
When backfilling, we only need to download blocks that are newer than MIN_EPOCHS_FOR_BLOCK_REQUESTS - the rest cannot reliably be fetched from the network and does not have to be provided to others.

This change affects only trusted-node-synced clients - genesis sync continues to work as before (because it needs to construct a state by building it from genesis).

Those wishing to complete a backfill should do so with era files instead.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1619,6 +1619,9 @@ proc start*(node: BeaconNode) {.raises: [Defect, CatchableError].} =
 
   waitFor node.initializeNetworking()
 
+  if node.eth1Monitor != nil:
+    node.eth1Monitor.start()
+
   node.run()
 
 func formatGwei(amount: uint64): string =
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -271,10 +271,7 @@ proc initFullNode(
     dag.backfill.slot
 
   func getFrontfillSlot(): Slot =
-    if dag.frontfill.isSome():
-      dag.frontfill.get().slot
-    else:
-      GENESIS_SLOT
+    max(dag.frontfill.get(BlockId()).slot, dag.horizon)
 
   let
     quarantine = newClone(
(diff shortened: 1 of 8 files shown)


arnetheduck:
mac test failure needs investigation, could be something real

arnetheduck:
was something real: pre-merge block production broken without EL

beacon_chain/sync/sync_queue.nim
```diff
@@ -793,7 +783,7 @@ proc push*[T](sq: SyncQueue[T], sr: SyncRequest[T],
           if safeSlot > failSlot:
             let rewindSlot = sq.getRewindPoint(failSlot, safeSlot)
             # It's quite common peers give us fewer blocks than we ask for
-            info "Gap in block range response, rewinding", request = req,
+            debug "Gap in block range response, rewinding", request = req,
```
tersec, line 786: Is this a general/unrelated change or does this become a more frequent noise-level message after this PR?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -445,10 +442,7 @@ proc init*(T: type BeaconNode,
   let optJwtSecret = rng[].loadJwtSecret(config, allowCreate = false)
 
   if config.web3Urls.len() == 0:
-    if cfg.BELLATRIX_FORK_EPOCH == FAR_FUTURE_EPOCH:
-      notice "Running without execution client - validator features partially disabled (see https://nimbus.guide/eth1.html)"
-    else:
-      notice "Running without execution client - validator features disabled (see https://nimbus.guide/eth1.html)"
+    notice "Running without execution client - validator features disabled (see https://nimbus.guide/eth1.html)"
```
tersec, line 445: So long as this remains only informational either way, it still allows compatibility with a testnet/devnet which decides not to go into Bellatrix

beacon_chain/sync/sync_queue.nim
arnetheduck: Unrelated, but it's spammy when syncing which I've been doing a lot while testing

beacon_chain/nimbus_beacon_node.nim
arnetheduck: Yeah, there's nuance here about fork timing etc that could be expressed better but .. the old check was also slightly wrong (for the pre-merge period) - not sure it's worth the lines of code to fix properly

PR #4432, opened 2022-12-15
update EF consensus spec URLs from v1.3.0-alpha.1 to v1.3.0-alpha.2

tersec:
automated via https://github.com/status-im/nimbus-eth2/blob/stable/scripts/find_unchanged_consensus_spec_files.sh

### DIFF
--- a/beacon_chain/beacon_chain_db_immutable.nim
+++ b/beacon_chain/beacon_chain_db_immutable.nim
@@ -18,7 +18,7 @@ from ./spec/datatypes/capella import ExecutionPayloadHeader, Withdrawal
 from ./spec/datatypes/eip4844 import ExecutionPayloadHeader
 
 type
-  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.1/specs/phase0/beacon-chain.md#beaconstate
+  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/phase0/beacon-chain.md#beaconstate
   # Memory-representation-equivalent to a phase0 BeaconState for in-place SSZ reading and writing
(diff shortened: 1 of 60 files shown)


henridf:
Nit, the comment 

>  Can sync optimistically without this condition, see note on `is_data_available` 

at the end of `state_transition_block.nim` is moot as of this PR.

tersec:
> Nit, the comment
> 
> > Can sync optimistically without this condition, see note on `is_data_available`
> 
> at the end of `state_transition_block.nim` is moot as of this PR.

Yeah, will adjust that next PR.

Balance with this giant/sprawling PR (and that particular file I hand-checked the diff and added it to the script's output, having noticed that as well) is to keep it low-risk by doing strictly what it says.

Another option, maybe better, would have been to hold off on that file and update it as a whole later. There's another case where there's a chance in whether a "New in Altair" comment is there, and that one I intentionally ignored either way.

tersec:
> > Nit, the comment
> > > Can sync optimistically without this condition, see note on `is_data_available`
> > 
> > 
> > at the end of `state_transition_block.nim` is moot as of this PR.
> 
> Yeah, will adjust that next PR.
> 
> Balance with this giant/sprawling PR (and that particular file I hand-checked the diff and added it to the script's output, having noticed that as well) is to keep it low-risk by doing strictly what it says.
> 
> Another option, maybe better, would have been to hold off on that file and update it as a whole later. There's another case where there's a chance in whether a "New in Altair" comment is there, and that one I intentionally ignored either way.

https://github.com/status-im/nimbus-eth2/pull/4428/commits/3737a4493abb8a229e0f75a3413555ced6844d42

PR #4434, opened 2022-12-16
60% state replay speedup

arnetheduck:
* don't use HashList for epoch participation - in addition to the code currently clearing the caches several times redundantly, clearing has to be done each block nullifying the benefit (35%)
* introduce active balance cache - computing it is slow due to cache unfriendliness in the random access pattern and bounds checking and we do it for every block - this cache follows the same update pattern as the active validator index cache (20%)
* avoid recomputing base reward several times per attestation (5%)

Applying 1024 blocks goes from 20s to ~8s on my laptop - these kinds of requests happen on historical REST queries but also whenever there's a reorg.

image

### DIFF
--- a/beacon_chain/spec/ssz_codec.nim
+++ b/beacon_chain/spec/ssz_codec.nim
@@ -17,7 +17,7 @@ import
   ./datatypes/base
 
 from ./datatypes/altair import
-  ParticipationFlags, EpochParticipationFlags, asHashList
+  ParticipationFlags, EpochParticipationFlags
 
 export codec, base, typetraits, EpochParticipationFlags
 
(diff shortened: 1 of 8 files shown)


beacon_chain/spec/ssz_codec.nim
```diff
@@ -60,4 +60,6 @@ func fromSszBytes*(T: type JustificationBits, bytes: openArray[byte]): T {.raise
   copyMem(result.addr, unsafeAddr bytes[0], sizeof(result))
 
 func fromSszBytes*(T: type EpochParticipationFlags, bytes: openArray[byte]): T {.raisesssz.} =
-  readSszValue(bytes, HashList[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT] result)
+  # TODO https://github.com/nim-lang/Nim/issues/21123
+  let tmp = cast[ptr List[ParticipationFlags, Limit VALIDATOR_REGISTRY_LIMIT]](addr result)
```
zah, line 64: The previous code is also based on a distinct type. How come it was working?

zah (approved):

beacon_chain/spec/ssz_codec.nim
arnetheduck: `List` is a `distinct seq` - `HashList` is an object - the bug happens with double distinct + generic (see upstream issue)

The previous code thus had `distinct[generic-object]`, not `distinct[generic-distinct]` - as a side note, this code remains subtly wrong relative to the standard it implements - it should really use a `distinct RestByte = uint8` and create a list over that but fixing it is beyond the scope here.

arnetheduck: apparently fixed for nim 2.0, but still there for 1.6

PR #4437, opened 2022-12-17
Allow reindexing to start from an arbitrary state

arnetheduck:
When not backfilling all the way to genesis (#4421), it becomes more useful to start rebuilding the historical indices from an arbitrary starting point.

To rebuild the index from non-genesis, a state and an unbroken block history is needed - here, we allow loading the state from an era file and recreating the history from there onwards.

* speed up partial era state loading

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -2161,21 +2161,13 @@ func needsBackfill*(dag: ChainDAGRef): bool =
 proc rebuildIndex*(dag: ChainDAGRef) =
   ## After a checkpoint sync, we lack intermediate states to replay from - this
   ## function rebuilds them so that historical replay can take place again
-  ## TODO handle databases without genesis state
-  if dag.backfill.slot > 0:
-    debug "Backfill not complete, cannot rebuild archive"
-    return
-
-  if dag.tail.slot == GENESIS_SLOT:
-    # The tail is the earliest slot for which we're supposed to have states -
-    # if it's sufficiently recent, don't do anything
-    debug "Archive does not need rebuilding"
-    return
-
+  ## TODO the pruning of junk states could be moved to a separate function that
+  ##      runs either on startup
   # First, we check what states we already have in the database - that allows
   # resuming the operation at any time
   let
     roots = dag.db.loadStateRoots()
+    historicalRoots = getStateField(dag.headState, historical_roots).asSeq()
 
   var
     canonical = newSeq[Eth2Digest](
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -2215,19 +2209,53 @@ proc rebuildIndex*(dag: ChainDAGRef) =
...
+            dag.cfg.stateForkAtEpoch(slot.epoch), state_root, state[], noRollback):
+          fatal "Cannot load state, database corrupt or created for a different network?",
+            state_root, slot
+          quit 1
+        tailBid = Opt.some state[].latest_block_id()
+
+      continue
+
+    if i == 0 or canonical[i - 1].isZero:
+      reset(tailBid) # No unbroken history!
```
zah, line 2250: What's the point of computing the `tailBid` above when you are going to reset it here?
You can put the above block in an else condition for this if.

beacon_chain/spec/forks.nim
```diff
@@ -866,24 +866,29 @@ type
     genesis_validators_root: Eth2Digest
     slot: Slot
 
+func readSszForkedHashedBeaconState*(
+    cfg: RuntimeConfig, slot: Slot, data: openArray[byte]):
+    ForkedHashedBeaconState {.raises: [Defect, SszError].} =
+  # TODO https://github.com/nim-lang/Nim/issues/19357
+  result = ForkedHashedBeaconState(
+    kind: cfg.stateForkAtEpoch(slot.epoch()))
```
zah, line 874: I suspect there is a more optimal way to implement this by fully initializing the `ForkedHashedBeaconState` object in a switch statement (switching by fork).

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: the tail computation needs to detect the most recent bid from which there is an unbroken history to `head` - therefore, when we start computing, we set it optimistically to that point - then we continue computing the state moving towards head, but if there's a gap in the state snapshot continuum (ie canonical[-1] is missing because we fail to compute it because for example only block header index and not block data is there), we need to move the tail again past that gap.

arnetheduck: notably, the tail bid is a block coordinate for which we don't necessarily have block data - ie when we have a state for slot 32 it will sometimes have been created by an earlier block + empty slot processing _unless_ it's a checkpoint state - so a state at slot 32 might have a bid from slot 31 where no block data exists

beacon_chain/spec/forks.nim
arnetheduck: not without generating temporaries - ie I've tried other constructs and they all led to RVO being disabled and a temporary being introduced - if you know of any way to do it, would indeed be nice

beacon_chain/consensus_object_pools/blockchain_dag.nim
zah: Ah, I've missed the `continue` statement in the if above. Sorry for the noise

PR #4438, opened 2022-12-19
Migrating the deposit contract snapshot can no longer fail on start-up

zah:
The missing piece of data that had to be obtained previously from the configured EL client is now part of the network metadata baked into the binary.

### DIFF
--- a/beacon_chain/spec/presets.nim
+++ b/beacon_chain/spec/presets.nim
@@ -202,7 +202,7 @@ when const_preset == "mainnet":
     # Ethereum PoW Mainnet
     DEPOSIT_CHAIN_ID: 1,
     DEPOSIT_NETWORK_ID: 1,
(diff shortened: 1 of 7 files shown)


beacon_chain/spec/presets.nim
```diff
@@ -202,7 +202,7 @@ when const_preset == "mainnet":
     # Ethereum PoW Mainnet
     DEPOSIT_CHAIN_ID: 1,
     DEPOSIT_NETWORK_ID: 1,
-    DEPOSIT_CONTRACT_ADDRESS: Eth1Address.fromHex("0x00000000219ab540356cBB839Cbe05303d7705Fa")
+    DEPOSIT_CONTRACT_ADDRESS: default(Eth1Address)
```
zah, line 205: The more strict validation of the network metadata uncovered this issue where all network were assumed to have a deposit contract. This required some of our local simulations which don't deploy a deposit contract to specify its deployment block.

arnetheduck (approved):

PR #4444, opened 2022-12-21
eip4844 gossip

henridf:
Adds gossip for the `beacon_block_and_blobs_sidecar` topic. Validation is mostly stubbed out pending integration of nim-kzg library.

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -15,7 +15,7 @@ import
   stew/results,
   chronicles, chronos, metrics, taskpools,
   ../spec/[helpers, forks],
-  ../spec/datatypes/[altair, phase0],
+  ../spec/datatypes/[altair, phase0, eip4844],
   ../consensus_object_pools/[
     block_clearance, block_quarantine, blockchain_dag, exit_pool, attestation_pool,
     light_client_pool, sync_committee_msg_pool],
@@ -239,6 +239,62 @@ proc processSignedBeaconBlock*(
     beacon_blocks_dropped.inc(1, [$v.error[0]])
 
   v
+
+proc processSignedBeaconBlockAndBlobsSidecar*(
+  self: var Eth2Processor, src: MsgSource,
+  signedBlockAndBlobsSidecar: SignedBeaconBlockAndBlobsSidecar): ValidationRes =
+  let
+    wallTime = self.getCurrentBeaconTime()
+    (afterGenesis, wallSlot) = wallTime.toSlot()
+  let signedBlock = signedBlockAndBlobsSidecar.beacon_block
+
+  logScope:
+    blockRoot = shortLog(signedBlock.root)
+    blck = shortLog(signedBlock.message)
+    signature = shortLog(signedBlock.signature)
+    wallSlot
+
+  if not afterGenesis:
+    notice "Block before genesis"
+    return errIgnore("Block before genesis")
+
+  # Potential under/overflows are fine; would just create odd metrics and logs
+  let delay = wallTime - signedBlock.message.slot.start_beacon_time
+
+  # Start of block processing - in reality, we have already gone through SSZ
+  # decoding at this stage, which may be significant
+  debug "Block received", delay
+
+  let blockRes =
+    self.dag.validateBeaconBlock(self.quarantine, signedBlock, wallTime, {})
+  if blockRes.isErr():
+    debug "Dropping block", error = blockRes.error()
+    self.blockProcessor[].dumpInvalidBlock(signedBlock)
+    beacon_blocks_dropped.inc(1, [$blockRes.error[0]])
+    return blockRes
+
+  let sidecarRes =  validateBeaconBlockAndBlobsSidecar(signedBlockAndBlobsSidecar)
+  if sidecarRes.isOk():
+    # Block passed validation - enqueue it for processing. The block processing
+    # queue is effectively unbounded as we use a freestanding task to enqueue
+    # the block - this is done so that when blocks arrive concurrently with
+    # sync, we don't lose the gossip blocks, but also don't block the gossip
+    # propagation of seemingly good blocks
+    trace "Block validated"
+    self.blockProcessor[].addBlock(
+      src, ForkedSignedBeaconBlock.init(signedBlock),
+      validationDur = nanoseconds(
+        (self.getCurrentBeaconTime() - wallTime).nanoseconds))
+
+    # Validator monitor registration for blocks is done by the processor
+    beacon_blocks_received.inc()
+    beacon_block_delay.observe(delay.toFloatSeconds())
+  else:
+    debug "Dropping block", error = sidecarRes.error()
+    self.blockProcessor[].dumpInvalidBlock(signedBlock)
+    beacon_blocks_dropped.inc(1, [$sidecarRes.error[0]])
+
+  sidecarRes
 
 proc setupDoppelgangerDetection*(self: var Eth2Processor, slot: Slot) =
   # When another client's already running, this is very likely to detect
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -275,25 +275,26 @@ proc processSignedBeaconBlockAndBlobsSidecar*(
     return blockRes
 
   let sidecarRes =  validateBeaconBlockAndBlobsSidecar(signedBlockAndBlobsSidecar)
-  if sidecarRes.isOk():
-    # Block passed validation - enqueue it for processing. The block processing
-    # queue is effectively unbounded as we use a freestanding task to enqueue
-    # the block - this is done so that when blocks arrive concurrently with
-    # sync, we don't lose the gossip blocks, but also don't block the gossip
-    # propagation of seemingly good blocks
-    trace "Block validated"
-    self.blockProcessor[].addBlock(
-      src, ForkedSignedBeaconBlock.init(signedBlock),
-      validationDur = nanoseconds(
-        (self.getCurrentBeaconTime() - wallTime).nanoseconds))
-
-    # Validator monitor registration for blocks is done by the processor
-    beacon_blocks_received.inc()
-    beacon_block_delay.observe(delay.toFloatSeconds())
-  else:
+  if sidecarRes.isErr():
     debug "Dropping block", error = sidecarRes.error()
     self.blockProcessor[].dumpInvalidBlock(signedBlock)
     beacon_blocks_dropped.inc(1, [$sidecarRes.error[0]])
+    return sidecarRes
+
+  # Block passed validation - enqueue it for processing. The block processing
+  # queue is effectively unbounded as we use a freestanding task to enqueue
+  # the block - this is done so that when blocks arrive concurrently with
+  # sync, we don't lose the gossip blocks, but also don't block the gossip
+  # propagation of seemingly good blocks
+  trace "Block validated"
+  self.blockProcessor[].addBlock(
+    src, ForkedSignedBeaconBlock.init(signedBlock),
+    validationDur = nanoseconds(
+        (self.getCurrentBeaconTime() - wallTime).nanoseconds))
+
+  # Validator monitor registration for blocks is done by the processor
+  beacon_blocks_received.inc()
+  beacon_block_delay.observe(delay.toFloatSeconds())
 
   sidecarRes
 
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -246,7 +246,8 @@ proc processSignedBeaconBlockAndBlobsSidecar*(
   let
     wallTime = self.getCurrentBeaconTime()
     (afterGenesis, wallSlot) = wallTime.toSlot()
-  let signedBlock = signedBlockAndBlobsSidecar.beacon_block
+
+  template signedBlock: auto = signedBlockAndBlobsSidecar.beacon_block
 
   logScope:
     blockRoot = shortLog(signedBlock.root)
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -25,6 +25,7 @@ import
   ./batch_validation
 
 from ../spec/datatypes/capella import SignedBeaconBlock
+from ../spec/datatypes/eip4844 import SignedBeaconBlock, SignedBeaconBlockAndBlobsSidecar
 
 from libp2p/protocols/pubsub/pubsub import ValidationResult
 
@@ -182,7 +183,8 @@ template validateBeaconBlockBellatrix(
 
 # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/bellatrix/p2p-interface.md#beacon_block
 template validateBeaconBlockBellatrix(
-       signed_beacon_block: bellatrix.SignedBeaconBlock | capella.SignedBeaconBlock,
+       signed_beacon_block: bellatrix.SignedBeaconBlock |
+       capella.SignedBeaconBlock | eip4844.SignedBeaconBlock,
        parent: BlockRef): untyped =
   # If the execution is enabled for the block -- i.e.
   # is_execution_enabled(state, block.body) then validate the following:
@@ -225,7 +227,8 @@ template validateBeaconBlockBellatrix(
 proc validateBeaconBlock*(
     dag: ChainDAGRef, quarantine: ref Quarantine,
     signed_beacon_block: phase0.SignedBeaconBlock | altair.SignedBeaconBlock |
-                         bellatrix.SignedBeaconBlock | capella.SignedBeaconBlock,
+                         bellatrix.SignedBeaconBlock | capella.SignedBeaconBlock |
+                         eip4844.SignedBeaconBlock,
     wallTime: BeaconTime, flags: UpdateFlags): Result[void, ValidationError] =
   # In general, checks are ordered from cheap to expensive. Especially, crypto
   # verification could be quite a bit more expensive than the rest. This is an
@@ -387,14 +390,50 @@ proc validateBeaconBlock*(
 
   ok()
 
-from ../spec/datatypes/eip4844 import SignedBeaconBlock
+proc validateBeaconBlockAndBlobsSidecar*(signedBlock: SignedBeaconBlockAndBlobsSidecar):
+                         Result[void, ValidationError] =
+  # [REJECT] The KZG commitments of the blobs are all correctly encoded
+  # compressed BLS G1 points -- i.e. all(bls.KeyValidate(commitment) for
+  # commitment in block.body.blob_kzg_commitments)
+  # todo ^
 
-proc validateBeaconBlock*(
-    dag: ChainDAGRef, quarantine: ref Quarantine,
-    signed_beacon_block: eip4844.SignedBeaconBlock,
-    wallTime: BeaconTime, flags: UpdateFlags): Result[void, ValidationError] =
-  debugRaiseAssert $eip4844ImplementationMissing & ": gossip_validation.nim: validateBeaconBlock not how EIP4844 works anymore"
-  err(default(ValidationError))
+  # [REJECT] The KZG commitments correspond to the versioned hashes in
+  # the transactions list --
+  # i.e. verify_kzg_commitments_against_transactions(block.body.execution_payload.transactions,
+  # block.body.blob_kzg_commitments)
+  if not verify_kzg_commitments_against_transactions(
+    signedBlock.beacon_block.message.body.execution_payload.transactions.asSeq,
+    signedBlock.beacon_block.message.body.blob_kzg_commitments.asSeq):
+    return errReject("KZG blob commitments not correctly encoded")
+
+  let sidecar = signedBlock.blobs_sidecar
+
+  # [IGNORE] the sidecar.beacon_block_slot is for the current slot
+  # (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. sidecar.beacon_block_slot == block.slot.
+  if not (sidecar.beacon_block_slot == signedBlock.beacon_block.message.slot):
+     return errIgnore("sidecar and block slots not equal")
+
+  # [REJECT] the sidecar.blobs are all well formatted, i.e. the
+  # BLSFieldElement in valid range (x < BLS_MODULUS).
+  # todo ^
+
+  # [REJECT] The KZG proof is a correctly encoded compressed BLS G1
+  # point -- i.e. bls.KeyValidate(blobs_sidecar.kzg_aggregated_proof)
+  # todo ^
+
+  # [REJECT] The KZG commitments in the block are valid against the
+  # provided blobs sidecar -- i.e. validate_blobs_sidecar(block.slot,
+  # hash_tree_root(block), block.body.blob_kzg_commitments, sidecar)
+  let res = validate_blobs_sidecar(signedBlock.beacon_block.message.slot,
+                                   hash_tree_root(signedBlock.beacon_block),
+                                   signedBlock.beacon_block.message
+                                   .body.blob_kzg_commitments.asSeq,
+                                   sidecar)
+  if res.isOk():
+    ok()
+  else:
+    errIgnore(res.error())
+
 
 # https://github.com/ethereum/consensus-specs/blob/v1.1.9/specs/phase0/p2p-interface.md#beacon_attestation_subnet_id
 proc validateAttestation*(
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -417,14 +417,13 @@ proc validateBeaconBlockAndBlobsSidecar*(signedBlock: SignedBeaconBlockAndBlobsS
   # BLSFieldElement in valid range (x < BLS_MODULUS).
   for blob in sidecar.blobs:
     for i in 0..<blob.len div 8:
-      let fe = cast[UInt256](sidecar.blobs[i*8..(i+1)*8])
+      let fe = UInt256.fromBytesBE(blob[i*8..(i+1)*8])
       if fe >= BLS_MODULUS:
         return errIgnore("BLSFieldElement outside of valid range")
 
   # TODO
   # [REJECT] The KZG proof is a correctly encoded compressed BLS G1
   # point -- i.e. bls.KeyValidate(blobs_sidecar.kzg_aggregated_proof)
-
 
   # [REJECT] The KZG commitments in the block are valid against the
   # provided blobs sidecar -- i.e. validate_blobs_sidecar(block.slot,
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -25,7 +25,8 @@ import
   ./batch_validation
 
 from ../spec/datatypes/capella import SignedBeaconBlock
-from ../spec/datatypes/eip4844 import SignedBeaconBlock, SignedBeaconBlockAndBlobsSidecar, BLS_MODULUS
+from ../spec/datatypes/eip4844 import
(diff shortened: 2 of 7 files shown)


henridf:
@tersec updated per our offline discussion: nim-kzg dependency removed for now, and fixed URLs to point to alpha.2.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -240,6 +240,62 @@ proc processSignedBeaconBlock*(
 
   v
 
+proc processSignedBeaconBlockAndBlobsSidecar*(
+  self: var Eth2Processor, src: MsgSource,
+  signedBlockAndBlobsSidecar: SignedBeaconBlockAndBlobsSidecar): ValidationRes =
+  let
+    wallTime = self.getCurrentBeaconTime()
+    (afterGenesis, wallSlot) = wallTime.toSlot()
+  let signedBlock = signedBlockAndBlobsSidecar.beacon_block
```
```suggestion
  signedBlock = signedBlockAndBlobsSidecar.beacon_block
```

tersec: But more broadly, this involves copying what might be a large-ish beacon block by value, if it contains a sizable execution payload, for slightly more convenient access.

There are a few ways of dealing with this:

- something like `template signedBlock: auto = signedBlockAndBlobsSidecar.beacon_block` expands in-place, no copies (and on the C and machine code level, it's just adjusting the offset in memory it's using)
- use something like `addr signedBlockAndBlobsSidecar.beacon_block` to get a reference, but then one has to dereference (`[]`) each access
- just use the full `signedBlockAndBlobsSidecar.beacon_block` regardless
- factor out some inner part of the function which uses `signedBlock` and pass ``signedBlockAndBlobsSidecar.beacon_block` to it. that won't copy the whole structure.

my preference is the `template` version. It's an idiom one can spot around `nimbus-eth2` if one looks, to avoid exactly this hazard

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -387,14 +390,50 @@ proc validateBeaconBlock*(
 
   ok()
 
-from ../spec/datatypes/eip4844 import SignedBeaconBlock
+proc validateBeaconBlockAndBlobsSidecar*(signedBlock: SignedBeaconBlockAndBlobsSidecar):
+                         Result[void, ValidationError] =
+  # [REJECT] The KZG commitments of the blobs are all correctly encoded
+  # compressed BLS G1 points -- i.e. all(bls.KeyValidate(commitment) for
+  # commitment in block.body.blob_kzg_commitments)
+  # todo ^
```
tersec: Conventionally, in `nimbus-eth2`, `todo` in this role is capitalized: `TODO`

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -387,14 +390,50 @@ proc validateBeaconBlock*(
...
+  let sidecar = signedBlock.blobs_sidecar
+
+  # [IGNORE] the sidecar.beacon_block_slot is for the current slot
+  # (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. sidecar.beacon_block_slot == block.slot.
+  if not (sidecar.beacon_block_slot == signedBlock.beacon_block.message.slot):
+     return errIgnore("sidecar and block slots not equal")
+
+  # [REJECT] the sidecar.blobs are all well formatted, i.e. the
+  # BLSFieldElement in valid range (x < BLS_MODULUS).
+  # todo ^
```
tersec: Is this feasible without a full KZG library? That is, is it just a matter of literal numeric comparison. `keyValidate` isn't, presumably.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -862,10 +862,32 @@ func process_blob_kzg_commitments(
...
+
+  if expected_kzg_commitments.len != blobs_sidecar.blobs.len:
+    return err("validate_blobs_sidecar: different commitment lengths")
+
+  # todo:
+  # if not kzg_4844.verify_aggregate_kzg_proof(asSeq(blobs_sidecar.blobs), expected_kzg_commitments, blobs_sidecar.kzg_aggregated_proof):
+  #  return err("validate_blobs_sidecar: aggregated kzg proof verification failed")
+
+  ok()
+
```
tersec, line 884: Nit: too many newlines between functions -- the code around it consistently uses a single newline.

beacon_chain/gossip_processing/eth2_processor.nim
tersec: This isn't just a matter of speed but correctness: this could be too large for some stacks, which is what this kind of variable declaration uses (as opposed to `new`, which heap-allocates).

beacon_chain/gossip_processing/gossip_validation.nim
henridf: 976f699cec59369ea9d1ec92de5db3da7655e97e

beacon_chain/spec/state_transition_block.nim
henridf: 976f699cec59369ea9d1ec92de5db3da7655e97e

beacon_chain/gossip_processing/eth2_processor.nim
henridf: Thanks for the context - the `let =` form probably made me assume referential transparency and I didn't give it any thought. Fixed via template in e6096de487ddde888d1e5ef5ff7d6e621e0b07bb.

beacon_chain/gossip_processing/gossip_validation.nim
henridf: Yep, done in 769b251f1702a58e66d245603d248c35f70e1a2f. 
(The implementation that iterates over 8 bytes at a time and does `cast[UInt256]` seems a bit low-level but I couldn't see an other way... if there is i'm happy to adjust).

henridf: (keyValidate is described as a BLS function in the spec, but from a quick look it isn't exposed by nim-blscurve and I haven't dug further yet).

tersec: > Yep, done in [769b251](https://github.com/status-im/nimbus-eth2/commit/769b251f1702a58e66d245603d248c35f70e1a2f). (The implementation that iterates over 8 bytes at a time and does `cast[UInt256]` seems a bit low-level but I couldn't see an other way... if there is i'm happy to adjust).

Yeah, it's dependent on the `UInt256` representation, and endianness. Better is to use `readUintBE` or `fromBytesBE` from
https://github.com/status-im/nim-stint/blob/f36623998fdc20349b887b319ca12d8b5a056459/stint/io.nim#L431-L500

It also handles some other cases (e.g., running in the Nim VM) yours doesn't.

beacon_chain/spec/datatypes/eip4844.nim
```diff
@@ -33,12 +33,17 @@ const
   # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.1/specs/eip4844/beacon-chain.md#blob
   BLOB_TX_TYPE* = 0x05'u8
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/eip4844/polynomial-commitments.md
```
tersec:
```suggestion
  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/eip4844/polynomial-commitments.md#constants
```

beacon_chain/gossip_processing/gossip_validation.nim
henridf: feb9b5f4b8f1d724dd86c3ae59be037657b08eca I'd missed those conversion functions, thanks

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -25,6 +25,7 @@ import
   ./batch_validation
 
 from ../spec/datatypes/capella import SignedBeaconBlock
+from ../spec/datatypes/eip4844 import SignedBeaconBlock, SignedBeaconBlockAndBlobsSidecar, BLS_MODULUS
```
tersec: https://status-im.github.io/nim-style-guide/formatting.style.html
> We strive to follow NEP-1 for style matters - naming, capitalization, 80-character limit etc.

e.g.,
```nim
from ../spec/datatypes/eip4844 import
  SignedBeaconBlock, SignedBeaconBlockAndBlobsSidecar, BLS_MODULUS
```

tersec: for naming and capitalization, it's a bit more aspirational in Nimbus, because we also ... strive ... to mirror the upstream spec names for things. But 80-character lines are consistent with upstream specs.

henridf: > 80-character limit etc.

Fixed this and a couple of other instances. I've been manually limiting line width, but is there a linter I'm unaware of that I could run?

tersec: `nimpretty` is supposed to be able to fix this automatically, but from the same link:
> We do not use `nimpretty` - as of writing (nim 1.2), it is not stable enough for daily use

I'd suggest, still as of now, Nim 1.6 or even Nim `devel`: https://github.com/nim-lang/Nim/issues/18498 https://github.com/nim-lang/Nim/issues/20078 https://github.com/nim-lang/Nim/issues/21018 not worth the risk.

Beyond that, I've used generic approaches in `vim`/`neovim` to highlight long lines in `.nim` files. I'm not sure if there's usable Nim linting around, beyond that.

henridf: 👍 yep i've configured  my editor similarly.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1373,14 +1373,28 @@ proc installMessageValidators(node: BeaconNode) =
...
+    node.network.addValidator(
+      getBeaconBlocksTopic(digest),
+      proc (signedBlock: phase.SignedBeaconBlock): ValidationResult =
+        if node.shouldSyncOptimistically(node.currentSlot):
+          toValidationResult(
+            node.optimisticProcessor.processSignedBeaconBlock(signedBlock))
+        else:
+          toValidationResult(node.processor[].processSignedBeaconBlock(
+            MsgSource.gossip, signedBlock)))
+
```
tersec, line 1386: extra newline

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -240,6 +240,63 @@ proc processSignedBeaconBlock*(
...
+  let blockRes =
+    self.dag.validateBeaconBlock(self.quarantine, signedBlock, wallTime, {})
+  if blockRes.isErr():
+    debug "Dropping block", error = blockRes.error()
+    self.blockProcessor[].dumpInvalidBlock(signedBlock)
+    beacon_blocks_dropped.inc(1, [$blockRes.error[0]])
+    return blockRes
+
+  let sidecarRes =  validateBeaconBlockAndBlobsSidecar(signedBlockAndBlobsSidecar)
+  if sidecarRes.isOk():
```
arnetheduck: codewise, it makes more sense to use `isErr` here and return early so as to keep the same flow for all error returns and have the success condition at the end..

also, I wonder if we should structure the code more towards the blob being decoupled from each other - not in this PR perhaps, but probably we want to run blob/sidecar validation in parallel (in a decoupled design, we'd need a quarantine / holding area for unmatched blob/block pairs etc but it would be interesting to explore, code-wise, what the impact is)

beacon_chain/gossip_processing/eth2_processor.nim
arnetheduck: nim 1.6 allows some borrowing with `let` - these capabilities improve in 2.0 as well, but indeed, for now the template is a cheap hack to avoid it for the worst offenders.

beacon_chain/gossip_processing/eth2_processor.nim
henridf: > codewise, it makes more sense to use isErr here and return early so as to keep the same flow for all error returns and have the success condition at the end..

fully agreed, done in c03e57a21181f38016df307428e97539fa44f405

> but probably we want to run blob/sidecar validation in parallel

Is the motivation to limit validation latency? I guess I don't have much intuition at this point of how much extra sequential processing time blob validation will take, but if the additional latency is notable then yes it makes sense. 

Or maybe this is this future-proofing for a world where gossip ends up being reworked to have separate blocks and blobs topics? I agree with that benefit (ofc depends on if/when that gossip change would be made).

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1373,14 +1373,27 @@ proc installMessageValidators(node: BeaconNode) =
   # subnets are subscribed to during any given epoch.
   let forkDigests = node.dag.forkDigests
 
+  template installBeaconBlocksValidator(digest: auto, phase: auto) =
+    node.network.addValidator(
+      getBeaconBlocksTopic(digest),
+      proc (signedBlock: phase.SignedBeaconBlock): ValidationResult =
```
tersec: If this `phase.SignedBeaconBlock` method works reliably, that'd be useful more broadly.

The current codebase doesn't do this, I'm not sure if it's documented, and Nim does some strange things with it, though:
```nim
let x = 3
template f(y: auto) = echo y.x      # Builds, and
f(system)                           # outputs "3"; same with f(y: untyped)

# template g(y: untyped) = echo y.3 # Error: identifier expected, but got '3'
# echo system.3                     # Error: identifier expected, but got '3'

#echo system.x                       # Error: undeclared identifier: 'x'
```

Also, Nim 1.6 still doesn't properly typecheck these situations -- https://github.com/nim-lang/Nim/issues/1027 was only fixed a couple months ago in Nim `devel`. I've asked for a backport of its fix https://github.com/nim-lang/Nim/pull/20631 to 1.6, since Nimbus uses this pattern extensively.

tersec: One approach used is to pass the entire type (`phase0.SignedBeaconBlock`,  `altair.SignedBeaconBlock`, et cetera) as a type or generics parameter to a template or function. Repeats itself slightly, but the semantics are better-defined -- those things are in themselves types, while it's not clear at least to me what Nim thinks is happening here.

tersec: I'm partly concerned because the local testnets in CI keep failing (e.g., https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Flinux%2Fx86_64/detail/PR-4444/7/pipeline https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Fmacos%2Faarch64/detail/PR-4444/7/pipeline https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/x86_64/job/PR-4444/7/display/redirect).

Block receive gossip just stops working, regardless of fork or slot, entirely with this PR.

So, blocks get either sent by a node, which sends them through the block processing pipeline locally or sent and ... apparently not really processed properly. This works for a while, because nodes manage to request backfill blocks in time via req/resp, e.g.:
```
{"lvl":"DBG","ts":"2023-01-03 21:56:07.723+00:00","msg":"Requesting detected missing blocks","topics":"beacnde","blocks":"[f1db17e9]"}
{"lvl":"DBG","ts":"2023-01-03 21:56:07.723+00:00","msg":"Requesting blocks by root","topics":"requman","peer":"16U*ZvWiek","blocks":"[f1db17e9]","peer_score":1000}
{"lvl":"DBG","ts":"2023-01-03 21:56:07.733+00:00","msg":"Block resolved","blockRoot":"f1db17e9","blck":{"slot":20,"proposer_index":84,"parent_root":"264c4bde","state_root":"07584766","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":12,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":21,"block_number":0,"fee_recipient":"0x0000000000000000000000000000000000000000"},"blockVerified":true,"heads":1,"stateDataDur":{"value":8417},"sigVerifyDur":{"value":6546958},"stateVerifyDur":{"value":229417},"putBlockDur":{"value":210083},"epochRefDur":{"value":4417}}
{"lvl":"DBG","ts":"2023-01-03 21:56:07.733+00:00","msg":"Updated head block","topics":"chaindag","stateRoot":"07584766","justified":"0:00000000","finalized":"0:00000000","isOptHead":false,"newHead":"f1db17e9:20","lastHead":"264c4bde:19"}
{"lvl":"DBG","ts":"2023-01-03 21:56:07.734+00:00","msg":"Block processed","localHeadSlot":20,"blockSlot":20,"validationDur":{"value":0},"queueDur":{"value":197417},"storeBlockDur":{"value":7515041},"updateHeadDur":{"value":282667}}
```

Where node 3 initially sent that:
```
{"lvl":"NOT","ts":"2023-01-03 21:56:05.029+00:00","msg":"Block sent","blockRoot":"f1db17e9","blck":{"slot":20,"proposer_index":84,"parent_root":"264c4bde","state_root":"07584766","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":12,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":21,"block_number":0,"fee_recipient":"0x0000000000000000000000000000000000000000"},"signature":"a66896bc","delay":"29ms419us"}
{"lvl":"DBG","ts":"2023-01-03 21:56:05.037+00:00","msg":"Block resolved","blockRoot":"f1db17e9","blck":{"slot":20,"proposer_index":84,"parent_root":"264c4bde","state_root":"07584766","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":12,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":21,"block_number":0,"fee_recipient":"0x0000000000000000000000000000000000000000"},"blockVerified":true,"heads":1,"stateDataDur":{"value":10625},"sigVerifyDur":{"value":7561541},"stateVerifyDur":{"value":192250},"putBlockDur":{"value":268833},"epochRefDur":{"value":3417}}
{"lvl":"DBG","ts":"2023-01-03 21:56:05.038+00:00","msg":"Updated head block","topics":"chaindag","stateRoot":"07584766","justified":"0:00000000","finalized":"0:00000000","isOptHead":false,"newHead":"f1db17e9:20","lastHead":"264c4bde:19"}
{"lvl":"DBG","ts":"2023-01-03 21:56:05.038+00:00","msg":"Block processed","localHeadSlot":20,"blockSlot":20,"validationDur":{"value":0},"queueDur":{"value":38042},"storeBlockDur":{"value":8647625},"updateHeadDur":{"value":257000}}
```

This is the entirety of "Block received" via gossip, across all nodes and all slots:
```
local-testnet-minimal$ rg "Block received"
log0.txt
6294:{"lvl":"DBG","ts":"2023-01-03 21:54:53.019+00:00","msg":"Block received","topics":"beacnde","delay":"19ms893us","blockRoot":"83cebea6","wallSlot":8,"signature":"91a80241","blck":{"slot":8,"proposer_index":1003,"parent_root":"1aad008d","state_root":"2e2f695d","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":7,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":0,"block_number":0,"fee_recipient":""}}
7098:{"lvl":"DBG","ts":"2023-01-03 21:54:59.023+00:00","msg":"Block received","topics":"beacnde","delay":"23ms518us","blockRoot":"7f2a43ea","wallSlot":9,"signature":"aa987743","blck":{"slot":9,"proposer_index":804,"parent_root":"83cebea6","state_root":"094c2578","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":4,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":5,"block_number":0,"fee_recipient":""}}
7994:{"lvl":"DBG","ts":"2023-01-03 21:55:05.028+00:00","msg":"Block received","topics":"beacnde","delay":"28ms185us","blockRoot":"2a130009","wallSlot":10,"signature":"b72d5a20","blck":{"slot":10,"proposer_index":230,"parent_root":"7f2a43ea","state_root":"e82082e7","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":20,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":13,"block_number":0,"fee_recipient":""}}
8929:{"lvl":"DBG","ts":"2023-01-03 21:55:11.012+00:00","msg":"Block received","topics":"beacnde","delay":"12ms368us","blockRoot":"d1b30239","wallSlot":11,"signature":"b0c231e7","blck":{"slot":11,"proposer_index":920,"parent_root":"2a130009","state_root":"65d0af91","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":4,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":29,"block_number":0,"fee_recipient":""}}
9820:{"lvl":"DBG","ts":"2023-01-03 21:55:17.032+00:00","msg":"Block received","topics":"beacnde","delay":"32ms541us","blockRoot":"73c1b8f5","wallSlot":12,"signature":"8571e96a","blck":{"slot":12,"proposer_index":876,"parent_root":"d1b30239","state_root":"1bd3e6c9","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":4,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":29,"block_number":0,"fee_recipient":""}}
10744:{"lvl":"DBG","ts":"2023-01-03 21:55:23.033+00:00","msg":"Block received","topics":"beacnde","delay":"33ms487us","blockRoot":"0830c947","wallSlot":13,"signature":"ae0d9289","blck":{"slot":13,"proposer_index":456,"parent_root":"73c1b8f5","state_root":"67ba293e","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":8,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":29,"block_number":0,"fee_recipient":""}}
11584:{"lvl":"DBG","ts":"2023-01-03 21:55:29.018+00:00","msg":"Block received","topics":"beacnde","delay":"18ms649us","blockRoot":"f7bec601","wallSlot":14,"signature":"877a66b2","blck":{"slot":14,"proposer_index":686,"parent_root":"0830c947","state_root":"157c5ba6","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":4,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":29,"block_number":0,"fee_recipient":""}}
12431:{"lvl":"DBG","ts":"2023-01-03 21:55:35.032+00:00","msg":"Block received","topics":"beacnde","delay":"32ms632us","blockRoot":"64f7df29","wallSlot":15,"signature":"b484524e","blck":{"slot":15,"proposer_index":96,"parent_root":"f7bec601","state_root":"4548c429","eth1data":{"deposit_root":"1681181313acd0d68b1e8d8d275665fe521a9ce5724858c3121cd2f2d84f3e33","deposit_count":1024,"block_hash":"4242424242424242424242424242424242424242424242424242424242424242"},"graffiti":"Nimbus/v22.12.0-c03e57-stateofus","proposer_slashings_len":0,"attester_slashings_len":0,"attestations_len":5,"deposits_len":0,"voluntary_exits_len":0,"sync_committee_participants":29,"block_number":0,"fee_recipient":""}}
```
There should be 3 block received logs for every slot (not the one which sent the block,, of the 4 nodes).

The trouble with relying on req/resp, aside from that block gossip shouldn't be failing to begin with, is that since the attestations get dropped, these blocks don't necessarily get enough votes to keep the head moving. By the time req/resp gets processed, it's too late for the nodes to count votes in a similar way.

So, the combination of those Nim issues/bugs and this particular failure mode causes me to be skeptical that this refactoring is working as intended.

```
N=0; while ./scripts/launch_local_testnet.sh --preset minimal --nodes 4 --disable-htop --stop-at-epoch 8 -- --verify-finalization --discv5:no; do N=$((N+1)); echo "That was run #${N}"; sleep 67; done
```
where that variation runs the minimal preset version in a loop until it fails for some reason.
```
./scripts/launch_local_testnet.sh --preset minimal --nodes 4 --disable-htop --stop-at-epoch 8 -- --verify-finalization --discv5:no
```
from the `nimbus-eth2` directory runs the local testnet a single time, and the various options can of course be adjusted.

henridf: Thanks for digging into this and for the explanations. I've removed the template. I figured the refactoring was straightforward enough; lesson learned. In any case, it belonged at the least in a separate commit and probably in a different PR. 

As a sidenote, this also changed the order in which validators were installed. I did that that to group related validators for clarity, assuming the change had to be innocuous, but given the specific order in which the validators were being installed, am now wondering if there might have been a reason for that order. Anyway, order restored.

tersec: > Thanks for digging into this and for the explanations. I've removed the template. I figured the refactoring was straightforward enough; lesson learned. In any case, it belonged at the least in a separate commit and probably in a different PR.

Sure. I agree, in principle, with the refactoring, but also that's it probably shouldn't be part of this PR.

> As a sidenote, this also changed the order in which validators were installed. I did that that to group related validators for clarity, assuming the change had to be innocuous, but given the specific order in which the validators were being installed, am now wondering if there might have been a reason for that order. Anyway, order restored.

The order was not intended to be important. They'll get called by `nim-libp2p` in whatever order messages come in, regardless. Might as well keep the order though.

tersec (approved):

PR #4445, opened 2022-12-23
History pruning (fixes #4419)

### LINKED ISSUE #4419: UX for pruning
In offering pruning support, there are several things that can be pruned - https://github.com/status-im/nimbus-eth2/blob/45574021c3d60582a401c21e9095be808b477629/beacon_chain/consensus_object_pools/block_pools_types.nim#L109 uses some terms like "history" and "archive" but these have not made it outside of that module - although the code support in chaindag is mostly there, this has not yet been exposed in the CLI.

When considering pruning, we have several types of data to consider:

* blocks - need to be stored for `MIN_EPOCHS_FOR_BLOCK_REQUESTS` (33024 epochs, ~5 months)
* states - need finalized state and newer - upper bound in the inactivity leak range, ie in the ~3 weeks range (https://eth2book.info/bellatrix/part2/incentives/inactivity/)
* light client updates - `MIN_EPOCHS_FOR_BLOCK_REQUESTS` too - their size is negligeable so we might as well keep them

We also  have to consider indices on top of the raw data:
* block hash -> mini-header (slot+parent_root)
* slot->block (finalized blocks)
* (slot+block) -> state_root

Some statistics:

```
sqlite> select avg(length(value))/1024, sum(length(value))/1024, count(*) from bellatrix_state_no_validators ;
6774.26703344652|4457467|658

sqlite> select avg(length(value))/1024, sum(length(value))/1024, count(*) from bellatrix_blocks;
53.2420313847627|34399729|646101
```

From this table, we can see that blocks take up roughly 8x the size of states as of a fairly recent snapshot, looking at bellatrix+ state & block sizes. Avg block size is `53kb` but there's a lot of them - another way to look at it that for every state, we store ~1k blocks.

Currently, we backfill blocks to genesis and don't purge states - effectively, this leads to a mode of operation where a trusted-node-synced node has state history from the checkpoint-point, a forward-synced node has history from genesis and all nodes have blocks since genesis.

We can also "reindex" a backfilled node to create the intermediate states, turning a "backfill" node into a "genesis" node, but there is no (clean) way to move from "genesis" to "backfilled" (ie purging the states up to a point). There is also no way to move the "checkpoint-point".

Ignoring light client updates (they're small and can follow whatever we do for blocks or simply stay around for now since Nimbus is the majority light client provider), we can consider the following pruning modes:

* minimal - prune states and blocks to their absolute minimums
* normal - prune states and blocks to `MIN_EPOCHS_FOR_BLOCK_REQUESTS` - this allows "some" historic requests and is simple to explain
* archive - no pruning (as today)

Notably, here we talk about pruning only - it doesn't say anything about TNSed nodes, or how far they backfill.

Since in normal mode, we start pruning blocks and states after 5 months, it's reasonable to expect that it will grow harder and harder to backfill blocks from the network. Also, even when backfilling, it's not possible to recreate state history except by going either from genesis or from a snapshot.

Thus, it makes  sense to **never backfill** any further than what is minimally required by the protocol and instead rely on era files completely for the creation of "genesis" nodes.

On top of the above, it's cheap to move the pruning point backwards in time - for example if there's a usecase for "1 year of history" this can be expressed as "normal + X".

A possible UX is thus:

* `--pruning=[minimal/normal/none]`
* `--prune-horizon=N` where N is the number of (epochs/eras) of extra data to keep - it is debatable whether this option is worth the implemenation effort (ie just run an archive node, it's not like they're crazy expensive)

When talking about this, we end up with:

* "genesis archive node" for a node that was synced from genesis and has all history (exists today)
* "checkpoint archive node" for a node that was TNSed and then didn't prune (exists today)
* "minimal node" - a node that has no state history - there's no effective difference between "genesis" and "checkpoint" because pruning will render the two the same
* "normal node" - the default UX which offers about 5 months of history as a convenient middle ground - notably, after 5 months, a  checkpoint node becomes equivalent to a genesis node in this configuration

The above also does not consider the creation of era files - currently, era file creation is done using an external process which works with genesis archive nodes - one could plausibly consider a "prune-to-era" option, but that is outside the scope of this issue.

To make the above work well, it would be advisable to bring to production the state-diff feature - it would further lower the cost of "normal" nodes while at the same time speeding up queries, assuming diffing is done at a higher granularity than the 1024 blocks that we have today for state snapshots.

arnetheduck:
Introduce (optional) pruning of historical data - a pruned node will continue to answer queries for historical data up to `MIN_EPOCHS_FOR_BLOCK_REQUESTS` epochs, or roughly 5 months, capping typical database usage at around 60-70gb.

To enable pruning, add `--history=prune` to the command line - on the first start, old data will be cleared (which may take a while) - after that, data is pruned continuously.

When pruning an existing database, the database will not shrink - instead, the freed space is recycled as the node continues to run - to free up space, perform a trusted node sync with a fresh database.

When switching on archive mode in a pruned node, history is retained from that point onwards.

History pruning is scheduled to be enabled by default in a future release.

In this PR, `minimal` mode from #4419 is not implemented meaning retention periods for states and blocks are always the same - depending on user demand, a future PR may implement `minimal` as well.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -419,7 +419,7 @@ proc loadImmutableValidators(vals: DbSeq[ImmutableValidatorDataDb2]): seq[Immuta
       pubkey: tmp.pubkey.loadValid(),
       withdrawal_credentials: tmp.withdrawal_credentials)
 
-template withManyWrites*(dbParam: BeaconChainDB, body: untyped): untyped =
+template withManyWrites*(dbParam: BeaconChainDB, body: untyped) =
   let
     db = dbParam
     nested = isInsideTransaction(db.db)
@@ -433,13 +433,8 @@ template withManyWrites*(dbParam: BeaconChainDB, body: untyped): untyped =
     expectDb db.db.exec("BEGIN TRANSACTION;")
   var commit = false
   try:
-    when typeof(body) is void:
-      block: body
+      body
       commit = true
-    else:
-      let res = block: body
-      commit = true
-      res
   finally:
     if not nested:
       if commit:
@@ -847,9 +842,11 @@ proc putStateDiff*(db: BeaconChainDB, root: Eth2Digest, value: BeaconStateDiff) 
   db.stateDiffs.putSnappySSZ(root.data, value)
 
 proc delBlock*(db: BeaconChainDB, fork: BeaconBlockFork, key: Eth2Digest): bool =
+  var deleted = false
   db.withManyWrites:
     discard db.summaries.del(key.data).expectDb()
-    db.blocks[fork].del(key.data).expectDb()
+    deleted = db.blocks[fork].del(key.data).expectDb()
+  deleted
 
 proc delState*(db: BeaconChainDB, fork: BeaconStateFork, key: Eth2Digest) =
   discard db.statesNoVal[fork].del(key.data).expectDb()
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -70,7 +70,7 @@ type
     ##
     ## 1.2 moved BeaconStateNoImmutableValidators to a separate table to
     ## alleviate some of the btree balancing issues - this doubled the speed but
-    ## was still
+    ## was still slow
     ##
     ## 1.3 creates `kvstore` with rowid, making it quite fast, but doesn't do
     ## anything about existing databases. Versions after that use a separate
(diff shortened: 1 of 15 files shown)


arnetheduck:
draft until https://github.com/status-im/nim-eth/pull/574 is merged

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -2329,4 +2470,5 @@ proc rebuildIndex*(dag: ChainDAGRef) =
 
     for i in junk:
       dag.db.delStateRoot(i[0][1], i[0][0])
-      dag.db.delState(i[1])
+      dag.db.delState(dag.cfg.stateForkAtEpoch(i[0][0].epoch), i[1])
+
```
tersec: newline at end of file

beacon_chain/beacon_chain_db.nim
```diff
@@ -420,41 +420,43 @@ proc loadImmutableValidators(vals: DbSeq[ImmutableValidatorDataDb2]): seq[Immuta
...
+        # it might be necessary for SQLite to rollback and cancel the entire transaction.
+        # An application can tell which course of action SQLite took by using the
+        # sqlite3_get_autocommit() C-language interface.
+        #
+        # It is recommended that applications respond to the errors listed above by
+        # explicitly issuing a ROLLBACK command. If the transaction has already been
+        # rolled back automatically by the error response, then the ROLLBACK command
+        # will fail with an error, but no harm is caused by this.
+        #
+        if isInsideTransaction(db.db): # calls `sqlite3_get_autocommit`
```
tersec, line 458: Is this nested transaction code newly useful specifically with pruning?

arnetheduck: yes (at least it was, at some stage of the pruning development :smile:) - it's a bit of a mess, but the function is available at different abstraction levels - since it's a tool for performance (and not atomicity of operations) it kind of make sense to apply it liberally when there are multiple writes to group, so it's slightly easier to allow nesting because a "super-composed" operation like wiping lots of states will be made up of "composed" operations like wiping a state and its corresponding state root

beacon_chain/beacon_chain_db.nim
```diff
@@ -1339,6 +1345,12 @@ proc containsState*(db: BeaconChainDBV0, key: Eth2Digest): bool =
     db.backend.contains(sk).expectDb() or
     db.backend.contains(subkey(phase0.BeaconState, key)).expectDb()
 
+proc containsState*(db: BeaconChainDB, fork: BeaconStateFork, key: Eth2Digest,
+    legacy: bool = true): bool =
+  if db.statesNoVal[fork].contains(key.data).expectDb(): return true
+
+  (legacy and fork == BeaconStateFork.Phase0 and db.v0.containsState(key))
```
tersec, line 1351: Is support for this v0 database schema still relevant? At the moment, it could still be backfilled, since the network block availability still stretches back to genesis. Another approach might be to cut off this data and let it be recreated, while that's still feasible.

arnetheduck: the v0 data appears in any database that's been with us since genesis and covers the initial months - there's no simple way to "recreate" it that doesn't introduce complexity or performance issues, since it's a non-trivial amount of data whose performance profile is marred by the without_rowid mistake (ie deletes in particular are slow because they cause btree rebalances that rewrite all data).

We generally have never "rewritten" any tables in the database for the simple reason that there's never really a good time to do so - recreating states for example (as in this particular example, because it is a state function) would require reindexing / replay to reach parity, which is slow.

tersec: I'm not that strongly opposed to keeping it, but it's this special case which cuts across much of `beacon_chain_db`, and there's a relatively finite, albeit not that short, time window where it can be quite naturally reconstructed. Doing so is a one-time cost borne in the background to homogenize the database schema.

arnetheduck: The way forward for "keeping" the data in a pruned world I think is to rely on era files - as such, changing to the current table format is of dubious value - also, we've gone through other changes to the tables such as switching to framed snappy in bellatrix, enabling rowid:s etc - if we did a migration for each of those, we'd be looking at hour-long restart times which certainly would not be popular - or days of background processing for data that ultimately is read only rarely.

Regarding reconstruction, we actually have `--reindex` - it is now able to read era files and create state table entries - I'm seeing this as the way forward for "upgrading" a database to state diffs as well - part of the "pruning" series of PR: was to ensure that this works with era and can start at any point in time (as opposed to genesis-only) - this allows a user to move between a pruned and a full node with relative ease.

I'm somewhat hesitant however to require that "defaults-everything" users that have used nimbus for a long while have to do this kind of manual work to keep using nimbus, so I find it easier to maintain minimal read-only support in chaindb.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: 0b901bb433723c351e203842971ca94d077e6463

tersec (approved):

Issue #4395, opened 2022-12-06
Implement EIP-4844 


henridf:
# Tracking issue for EIP-4844 tasks

### Config & Setup
   - [x] New containers ([beacon block](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#beaconblockbody), [execution payload](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#executionpayload)) (#4365)
   - [x] New [constants](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#constants) and [presets](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#preset). (#4365)

### State transition & processing
- [x] [Block processing](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#beacon-chain-state-transition-function) 
   - [x] Move is_data_available check to fork-choice on_block (https://github.com/ethereum/consensus-specs/pull/3185)
- [x] Nim bindings to [`c-kzg-4844`](https://github.com/ethereum/c-kzg-4844) library. _(in progress https://github.com/henridf/nim-kzg-4844)_
- [x] Support [trusted setup](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md#trusted-setup) loading.

### P2P
- Gossip processing
   - [x] Gossip transition. (https://github.com/status-im/nimbus-eth2/pull/4463)
   - [x] [`beacon_block_and_blobs_sidecar`](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#beacon_block_and_blobs_sidecar) new topic. #4444
      - [x] `beacon_block_and_blobs_sidecar` validation
- ReqResp processing
   - [x] [`BeaconBlocksByRange v2`](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#beaconblocksbyrange-v2)
   - [x] [`BeaconBlocksByRoot v2`](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#beaconblocksbyroot-v2)
[...]

PR #4454, opened 2023-01-03
Blob storage

henridf:
Initial support for storing blobs. 

After a few different false starts, I ended up using `Opt[blob]`s quite extensively as that was the simplest / least invasive change. (Initially, I tried creating separate procs e.g. `AddBlockAndBlobs`, `StoreBlockAndBlobs` that would be used in lieu of the non-blob procs in the presence of blobs, but this was altogether more complex and lead to more repeated code).

Currently blob storage is only wired in for blobs coming from gossip. Sync is not wired in yet but _should_ also slot in relatively easily via passing a blob to `nimbus_beacon_node.blockVerifier()`

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -771,7 +771,7 @@ proc putBlock*(
     db.blocks[type(value).toFork].putSZSSZ(value.root.data, value)
     db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
 
-proc putBlobs*(
+proc putBlobsSidecar*(
     db: BeaconChainDB,
     value: BlobsSidecar) =
   db.blobs.putSZSSZ(value.beacon_block_root.data, value)
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -25,7 +25,7 @@ import
   "."/[beacon_chain_db_light_client, filepath]
 
 from ./spec/datatypes/capella import BeaconState
-from ./spec/datatypes/eip4844 import TrustedSignedBeaconBlock
+from ./spec/datatypes/eip4844 import TrustedSignedBeaconBlock, BlobsSidecar
 
 export
   phase0, altair, eth2_ssz_serialization, eth2_merkleization, kvstore,
@@ -114,6 +114,8 @@ type
 
     keyValues: KvStoreRef # Random stuff using DbKeyKind - suitable for small values mainly!
     blocks: array[BeaconBlockFork, KvStoreRef] # BlockRoot -> TrustedSignedBeaconBlock
+
+    blobs: KvStoreRef # (BlockRoot -> BlobsSidecar)
 
     stateRoots: KvStoreRef # (Slot, BlockRoot) -> StateRoot
 
@@ -503,6 +505,8 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("capella_blocks").expectDb(),
       kvStore db.openKvStore("eip4844_blocks").expectDb()]
 
+    blobs = kvStore db.openKvStore("blobs").expectDb()
+
     stateRoots = kvStore db.openKvStore("state_roots", true).expectDb()
 
     statesNoVal = [
@@ -555,6 +559,7 @@ proc new*(T: type BeaconChainDB,
     checkpoint: proc() = db.checkpoint(),
     keyValues: keyValues,
     blocks: blocks,
+    blobs: blobs,
     stateRoots: stateRoots,
     statesNoVal: statesNoVal,
     stateDiffs: stateDiffs,
@@ -766,6 +771,11 @@ proc putBlock*(
     db.blocks[type(value).toFork].putSZSSZ(value.root.data, value)
     db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
 
+proc putBlobs*(
+    db: BeaconChainDB,
+    value: BlobsSidecar) =
+    db.blobs.putSZSSZ(value.beacon_block_root.data, value)
+
 proc updateImmutableValidators*(
     db: BeaconChainDB, validators: openArray[Validator]) =
   # Must be called before storing a state that references the new validators
@@ -948,6 +958,12 @@ proc getBlock*[
     # set root after deserializing (so it doesn't get zeroed)
     result.get().root = key
   else:
+    result.err()
+
+proc getBlobs*(db: BeaconChainDB, key: Eth2Digest): Opt[BlobsSidecar] =
+  var blobs: BlobsSidecar
+  result.ok(blobs)
+  if db.blobs.getSZSSZ(key.data, result.get) != GetResult.found:
     result.err()
 
 proc getPhase0BlockSSZ(
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -20,6 +20,7 @@ import
           eth2_ssz_serialization,
           eth2_merkleization,
           forks,
+          presets,
           state_transition],
   ./spec/datatypes/[phase0, altair, bellatrix],
   "."/[beacon_chain_db_light_client, filepath]
@@ -478,7 +479,8 @@ proc new*(T: type BeaconChainDBV0,
   )
 
 proc new*(T: type BeaconChainDB,
-          db: SqStoreRef
+          db: SqStoreRef,
+          cfg: RuntimeConfig,
     ): BeaconChainDB =
   if not db.readOnly:
     # Remove the deposits table we used before we switched
@@ -505,8 +507,6 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("capella_blocks").expectDb(),
       kvStore db.openKvStore("eip4844_blocks").expectDb()]
 
-    blobs = kvStore db.openKvStore("blobs").expectDb()
-
     stateRoots = kvStore db.openKvStore("state_roots", true).expectDb()
 
     statesNoVal = [
@@ -524,6 +524,10 @@ proc new*(T: type BeaconChainDB,
       altairCurrentBranches: "lc_altair_current_branches",
       altairBestUpdates: "lc_altair_best_updates",
       sealedPeriods: "lc_sealed_periods")).expectDb()
+
+  var blobs : KvStoreRef
+  if cfg.EIP4844_FORK_EPOCH != FAR_FUTURE_EPOCH:
+    blobs = kvStore db.openKvStore("eip4844_blobs").expectDb()
 
   # Versions prior to 1.4.0 (altair) stored validators in `immutable_validators`
   # which stores validator keys in compressed format - this is
@@ -570,6 +574,7 @@ proc new*(T: type BeaconChainDB,
 
 proc new*(T: type BeaconChainDB,
           dir: string,
+          cfg: RuntimeConfig,
           inMemory = false,
           readOnly = false
     ): BeaconChainDB =
@@ -585,7 +590,7 @@ proc new*(T: type BeaconChainDB,
 
       SqStoreRef.init(
         dir, "nbc", readOnly = readOnly, manualCheckpoint = true).expectDb()
-  BeaconChainDB.new(db)
+  BeaconChainDB.new(db, cfg)
 
 template getLightClientDataDB*(db: BeaconChainDB): LightClientDataDB =
   db.lcData
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -774,7 +774,7 @@ proc putBlock*(
 proc putBlobs*(
     db: BeaconChainDB,
     value: BlobsSidecar) =
-    db.blobs.putSZSSZ(value.beacon_block_root.data, value)
+  db.blobs.putSZSSZ(value.beacon_block_root.data, value)
 
 proc updateImmutableValidators*(
     db: BeaconChainDB, validators: openArray[Validator]) =
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -480,7 +480,7 @@ proc new*(T: type BeaconChainDBV0,
 
 proc new*(T: type BeaconChainDB,
           db: SqStoreRef,
-          cfg: RuntimeConfig,
+          cfg: RuntimeConfig = defaultRuntimeConfig
     ): BeaconChainDB =
   if not db.readOnly:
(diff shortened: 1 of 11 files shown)


tersec:
Haven't gotten a chance to look at this more closely, but in general we're preferring `Opt[Foo]` to `Option[Foo]`.

henridf:
> Haven't gotten a chance to look at this more closely, but in general we're preferring Opt[Foo] to Option[Foo].

Fixed.

beacon_chain/beacon_chain_db.nim
```diff
@@ -766,6 +771,11 @@ proc putBlock*(
     db.blocks[type(value).toFork].putSZSSZ(value.root.data, value)
     db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
 
+proc putBlobs*(
+    db: BeaconChainDB,
+    value: BlobsSidecar) =
+    db.blobs.putSZSSZ(value.beacon_block_root.data, value)
```
```suggestion
  db.blobs.putSZSSZ(value.beacon_block_root.data, value)
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -447,7 +457,7 @@ proc storeBlock*(
         return err((VerifierError.UnviableFork, ProcessingStatus.completed))
 
       if not self.consensusManager.quarantine[].addOrphan(
-          dag.finalizedHead.slot, ForkedSignedBeaconBlock.init(signedBlock)):
+        dag.finalizedHead.slot, ForkedSignedBeaconBlock.init(signedBlock), blobs):
```
tersec: That indentation was intentional, to distinguish the `if` condition from the `if` body.

Same idea as the function parameter/return type declaration if it extends over multiple lines, vs the function body.

```nim
func f(
    someLongParams: SomeLongType) =
  5
```

henridf: fixed in 0978f5d37a2a8efe3c63c45258c604619fd7b2c1

beacon_chain/beacon_chain_db.nim
henridf: fixed in 0978f5d37a2a8efe3c63c45258c604619fd7b2c1

beacon_chain/beacon_chain_db.nim
```diff
@@ -766,6 +771,11 @@ proc putBlock*(
     db.blocks[type(value).toFork].putSZSSZ(value.root.data, value)
     db.putBeaconBlockSummary(value.root, value.message.toBeaconBlockSummary())
 
+proc putBlobs*(
```
arnetheduck:
```suggestion
proc putBlobSidecar*(
```

henridf: 691b6c02d06c077789125de0e2bf0a04bad244e4

beacon_chain/beacon_chain_db.nim
```diff
@@ -950,6 +960,12 @@ proc getBlock*[
   else:
     result.err()
 
+proc getBlobs*(db: BeaconChainDB, key: Eth2Digest): Opt[BlobsSidecar] =
```
arnetheduck:
```suggestion
proc getBlobsSidecar*(db: BeaconChainDB, key: Eth2Digest): Opt[BlobsSidecar] =
```

henridf: 🤦‍♂️ [03ef873](https://github.com/status-im/nimbus-eth2/pull/4454/commits/03ef87307bab5e717dae0bf5dce51c2cdad9f090)

arnetheduck:
hm, what's the plan for blob removal?

henridf:
> hm, what's the plan for blob removal?

It's tracked separately in #4395 (can be lower priority as not needed to join devnet). Haven't thought about it in depth but the obvious way would be to have a (epoch?) ticker, and periodically garbage collect. Does that make sense?

arnetheduck:
> a (epoch?) ticker, and periodically garbage collect. Does that make sense?

more or less, but the blob storage as done in this PR does not directly lend itself well to time-based pruning since there's no time-based index - one way to do it is to ensure that blobs are always pruned before blocks and use the slot->block index for this purpose - I'm raising it mainly so that the design of how blobs are stored takes this quirk into account - it's probably good to have a sketch of a plan for pruning, though we can indeed refactor this later

henridf:
> one way to do it is to ensure that blobs are always pruned before blocks and use the slot->block index for this purpose

Agreed, I was thinking of using `blockchain_dag.getBlockRange()`to get the block roots.

beacon_chain/beacon_chain_db.nim
```diff
@@ -503,6 +505,8 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("capella_blocks").expectDb(),
       kvStore db.openKvStore("eip4844_blocks").expectDb()]
 
+    blobs = kvStore db.openKvStore("blobs").expectDb()
```
arnetheduck:
```suggestion
    blobs = kvStore db.openKvStore("eip4844_blobs").expectDb()
```

to avoid table conflicts down the line with test data

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -175,7 +180,11 @@ proc storeBackfillBlock(
       # Track unviables so that descendants can be discarded properly
       self.consensusManager.quarantine[].addUnviable(signedBlock.root)
     else: discard
+    return res
 
+  if blobs.isSome():
+    # Only store blobs after successfully establishing block viability.
+    self.consensusManager.dag.db.putBlobsSidecar(blobs.get())
```
arnetheduck, line 203: this function is responsible for verifying that the data is "consistent enough" - it seems that part of that verification should be that the blob corresponds to the block data, no?

when backfilling, we compute the hash of the incoming block then verify that it matches the `parent_hash` - similarly, it seems that we should be checking here that the blob belongs to the block

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -446,6 +456,10 @@ proc storeBlock*(
 
     return err((blck.error, ProcessingStatus.completed))
 
+  # write blobs now that block has been written.
+  if blobs.isSome():
```
arnetheduck, line 488: same as backfilling - where do we verify that blob and block match?

beacon_chain/gossip_processing/block_processor.nim
henridf: hmm, good point. 

~I wonder if we should also be verifying that the KZG commitments correspond to the versioned hashes in the transactions list ? Otherwise we can get fed any old blobs that don't go with this block.~ (read too quick, this is exactly what you're suggesting)

beacon_chain/beacon_chain_db.nim
etan-status: maybe also as a `TEMPORARY` table until the star name / spec is finalized. or, only create the table if EIP4844_FORK_EPOCH is configured to be non-FAR_FUTURE_EPOCH

henridf: 8f52e5c1c8a839b83aabecec1175460f96e9fd14 (@etan-status went with the latter approach of guarding on FAR_FUTURE_EPOCH rather than modifying `nim-eth` to add support for creating `TEMPORARY` tables).

beacon_chain/gossip_processing/block_processor.nim
henridf: cf1fdf5d9b305dc8edff3cd1b8c6b816891721a8

I'm not very happy with the approach of adding a new `storeBackfillBlock(.., *eip4844.SignedBeaconBlock*, ..)` but don't currently see a better day to "cast" the generic `ForkySignedBeaconBlock` to an eip4844 block (which is needed to access the `blob_kzg_commitments` field).

henridf: > I'm not very happy with the approach of adding a new storeBackfillBlock(.., *eip4844.SignedBeaconBlock*, ..) but don't currently see a better day to "cast" the generic ForkySignedBeaconBlock to an eip4844 block (which is needed to access the blob_kzg_commitments field).

The better way was right there ... use `when typeof(signedBlock).toFork() >= BeaconBlockFork.EIP4844:` to conditionally check blobs based on type. 

Updated in 99f9e4be30e374a6d0d90f766c2b56bddc743e08.

beacon_chain/gossip_processing/block_processor.nim
henridf: 58bb326c99ccb8331e7e74f33576306a10116d02

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -326,7 +327,7 @@ proc validateBeaconBlock*(
     # in the quarantine for later processing
     if not quarantine[].addOrphan(
         dag.finalizedHead.slot,
-        ForkedSignedBeaconBlock.init(signed_beacon_block)):
+        ForkedSignedBeaconBlock.init(signed_beacon_block), blobs):
```
tersec, line 334: It's not valid for `blobs` to be anything but `Opt.none` for any type of `SignedBeaconBlock` except, currently, `eip4844.SignedBeaconBlock`. It'd be preferable to encode this statically via types, but at least, having an assertion (not an error message or `Result` value, but something which kills `nimbus-eth2`; this is an internal logic error) there to ensure that this holds. The former involves more potential duplication and/or refactoring, but the latter is a couple of lines of code.

beacon_chain/beacon_chain_db.nim
```diff
@@ -521,6 +525,10 @@ proc new*(T: type BeaconChainDB,
       altairBestUpdates: "lc_altair_best_updates",
       sealedPeriods: "lc_sealed_periods")).expectDb()
 
+  var blobs : KvStoreRef
+  if cfg.EIP4844_FORK_EPOCH != FAR_FUTURE_EPOCH:
+    blobs = kvStore db.openKvStore("eip4844_blobs").expectDb()
```
tersec, line 532: I don't see where it closes this?

https://github.com/status-im/nimbus-eth2/blob/0ba9fc4ede9506c8b73dc91b1e0ce206f088b75a/beacon_chain/beacon_chain_db.nim#L722-L742
is where they get closed in general, and it tries to follow first-open-last-close discipline.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -393,6 +420,19 @@ proc storeBlock*(
...
+  # writing the block in case of blob error.
+  when typeof(signedBlock).toFork() >= BeaconBlockFork.EIP4844:
+    if blobs.isSome() and not
+         validate_blobs_sidecar(signedBlock.message.slot,
+                                hash_tree_root(signedBlock.message),
+                                signedBlock.message
+                                .body.blob_kzg_commitments.asSeq,
+                                blobs.get()).isOk():
+       return err((VerifierError.Invalid, ProcessingStatus.completed))
+    else:
```
tersec: It's fine if you want this for stylistic purposes (e.g., avoid dropping two indentation levels in a single line), but it's not a requirement to have this `else: discard`

tersec:
As a general comment, there's a lot of flexibility, false flexibility really, in this architecture.

There's been discussion of a more decoupled set of codepaths for blocks and blobs, for some future architecture, and maybe that end justifies the possibility of having EIP4844 blocks without any blobs (according to the type system), but having pre-EIP4844 blocks with blobs ideally simply should not be possible, not be a potential runtime error.

beacon_chain/gossip_processing/gossip_validation.nim
henridf: took the assert route for now: d0cc76c3289747f8f988138926d322e7acdbb73e

beacon_chain/beacon_chain_db.nim
henridf: oops, thanks for flagging that d0cc76c3289747f8f988138926d322e7acdbb73e

beacon_chain/gossip_processing/block_processor.nim
henridf: d0cc76c3289747f8f988138926d322e7acdbb73e

henridf:
> There's been discussion of a more decoupled set of codepaths for blocks and blobs, for some future architecture, and maybe that end justifies the possibility of having EIP4844 blocks without any blobs (according to the type system), but having pre-EIP4844 blocks with blobs ideally simply should not be possible, not be a potential runtime error.

Fully agreed, that's why the use of `Opt[Blob]` everywhere felt non-ideal (but as noted above, easiest path for a naive nim programmer). 

I'm happy to try to refactor this in a way that uses the type system, here or in a future PR. The challenge is in the fact that rather than add a field to the existing beacon block, eip4844 replaces it by a super-object (`BeaconBlockAndBlobsSidecar`), making it non-obvious how to unify/handle both in the type system. 

Off the type of my head, I would probably start off by defining:

```nim
ForkyBeaconBlockAndBlobsSidecar* =
    phase0.BeaconBlock |
    altair.BeaconBlock |
    bellatrix.BeaconBlock |
    capella.BeaconBlock |
    eip4844.BeaconBlockAndBlobsSidecar
```

as a way to unify the pre- and post-EIP4844 block gossip structure, and then have validateBeaconBlock take that as its input. But  pulling that thread may be quite invasive (and there's maybe a better/simpler way to do it).

tersec:
Yes, something like the `ForkyBeaconBlockAndBlobsSidecar` is what I had in mind -- and, one can have different overloads of the gossip validation functions entirely, which share some common code factored in a way to avoid duplication. So there would be a function of the block validation which handles Capella and earlier which pointedly does not know about sidecars, and a version for EIP4844 which does (and might even require them, depending on how coupled/decoupled it should be, but that then becomes a separate design decision again).

The other part then would be representing the block processing queue, and that might require a `ForkedFoo` type if one wants to address it entirely properly there, but I'm not sure it's worthwhile: it'd be just used as a conversion point in two or three places, which would add both conversion costs and potential for impedance mismatch-based required error handling at either end which I'm not sure would work out any better than the status quo using `Opt[Blob]`

These could be addressed separately, even -- the first I'm fairly sure is worthwhile, the second less so. But the first (beacon block validation) can be done regardless of the second (block processing queue).

I think it makes sense to merge this and do a refactoring in another PR, to unblock your other two RPC PRs, since as you note, it's a bit invasive, and the runtime check is good enough for now.

tersec (approved):

tersec:
Jenkins CI failures are due to unrelated issue fixed by https://github.com/status-im/nimbus-eth2/pull/4476

Should work after that's merged and this branch is updated

henridf:
> I think it makes sense to merge this and do a refactoring in another PR

Refactoring added to tracker at https://github.com/status-im/nimbus-eth2/issues/4395

PR #4455, opened 2023-01-03
capella validator withdrawal credentials aren't immutable

tersec:
Was resulting in
```
nimbus-eth2/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(489) main
nimbus-eth2/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(482) NimMain
nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(2125) main
nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1988) handleStartUpCmd
nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(1812) doRunBeaconNode
nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(562) init
nimbus-eth2/beacon_chain/nimbus_beacon_node.nim(168) loadChainDag
nimbus-eth2/beacon_chain/consensus_object_pools/blockchain_dag.nim(1005) init
nimbus-eth2/vendor/nim-stew/stew/results.nim(791) expect
nimbus-eth2/vendor/nim-stew/stew/results.nim(368) raiseResultDefect
Error: unhandled exception: head blocks should apply: process_withdrawals: different numbers of payload and expected withdrawals [ResultDefect]
```
when restarting a client reaching capella

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -509,8 +509,8 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("state_no_validators2").expectDb(),
       kvStore db.openKvStore("altair_state_no_validators").expectDb(),
       kvStore db.openKvStore("bellatrix_state_no_validators").expectDb(),
-      kvStore db.openKvStore("capella_state_no_validators").expectDb(),
-      kvStore db.openKvStore("eip4844_state_no_validators").expectDb()]
+      kvStore db.openKvStore("capella_state_no_validator_pubkeys").expectDb(),
+      kvStore db.openKvStore("eip4844_state_no_validator_pubkeys").expectDb()]
 
     stateDiffs = kvStore db.openKvStore("state_diffs").expectDb()
     summaries = kvStore db.openKvStore("beacon_block_summaries", true).expectDb()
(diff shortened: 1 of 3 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -509,8 +509,8 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("state_no_validators2").expectDb(),
       kvStore db.openKvStore("altair_state_no_validators").expectDb(),
       kvStore db.openKvStore("bellatrix_state_no_validators").expectDb(),
-      kvStore db.openKvStore("capella_state_no_validators").expectDb(),
-      kvStore db.openKvStore("eip4844_state_no_validators").expectDb()]
+      kvStore db.openKvStore("capella_state_no_validator_pubkeys").expectDb(),
```
arnetheduck, line 512: shall we simply call this pubkeys and let it be a version-less pubkey table?

tersec: Eventually, yes, it'll be a bit pointless to store not-actually-immutable-and-therefore-ignored things in the immutable-append-only-table. But it's still useful for Bellatrix, so might as well get the benefit of that, and keep the migration to a version-less pubkey table for after capella happens.

In this case, there's no critical time window for doing so before before not-yet-extant compatibility concerns come in: `capella` and `eip4844` state storage will be born compatible with this concept, and any phase 0/Altair/Bellatrix states in Nimbus databases now need the not-versionless pubkey-and-credentials immutable table anyway.

If anything, it might make sense to wait for the release until the pruning threshold hits Capella's fork epoch for supported networks (including Gnosis), at which point, it'll still be necessary to make sure there's a migration path for people running archive nodes, but reduces the downside risk for most people, in default configurations.

arnetheduck: > pruning threshold 

this doesn't change anything for as long as we support archive / replay nodes, so I wouldn't plan any features on top of it..

arnetheduck (approved):

PR #4457, opened 2023-01-04
getPayloadV2 support for capella

tersec:


arnetheduck (approved):

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -554,6 +555,18 @@ proc getPayload*(p: Eth1Monitor,
...
   p.dataProvider.web3.provider.engine_getPayloadV1(FixedBytes[8] payloadId)
 
+proc getPayloadV2*(
+    p: Eth1Monitor, payloadId: bellatrix.PayloadID):
+    Future[engine_api.ExecutionPayloadV2] =
+  # Eth1 monitor can recycle connections without (external) warning; at least,
+  # don't crash.
+  if p.isNil or p.dataProvider.isNil:
+    let epr = newFuture[engine_api.ExecutionPayloadV2]("getPayload")
+    epr.complete(default(engine_api.ExecutionPayloadV2))
```
arnetheduck, line 565: I find it weird that we return an empty payload here (and not some exception) but that's outside the scope of this PR I guess

Issue #3969, opened 2022-08-16
Continue attesting when web3 goes down


arnetheduck:
When the connection to the web3 server is lost, it remains possible and valid to keep voting for the last known head for some time (1 epoch? up to 2?) - using an old head/target/source still nets a rewards and remains a valid vote.

Ditto when syncing: after optimistic sync, it's sufficient that the execution layer confirms any block within the last 1-2 epochs to start attesting to _that_ block.

zah:
What's preventing attesting at the moment? Isn't this entirely a function of the current head block stored in the DAG?

arnetheduck:
if we're optimistically synced, we stop attesting completely which is correct (we should not attest to a head that the EL has not fully verified) but also overprotective - what happens, either during upgrades to the el or glitches, is that the EL goes out of sync for a few slots then comes back online - in those cases, we could continue attesting to the _latest_ block that the EL verified before going out of sync.

tersec:
https://github.com/status-im/nimbus-eth2/pull/5313

PR #4462, opened 2023-01-05
capella forkchoiceUpdated support

tersec:


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -623,15 +626,26 @@ proc forkchoiceUpdated*(p: Eth1Monitor,
...
-    some(engine_api.PayloadAttributesV1(
-      timestamp: Quantity timestamp,
-      prevRandao: FixedBytes[32] randomData,
-      suggestedFeeRecipient: suggestedFeeRecipient)))
+  let forkchoiceState = ForkchoiceStateV1(
+    headBlockHash: headBlock.asBlockHash,
+    safeBlockHash: safeBlock.asBlockHash,
+    finalizedBlockHash: finalizedBlock.asBlockHash)
+
+  if withdrawals == @[]:
```
etan-status: this breaks in a post-capella world if there are really no withdrawals, e.g., because the entire network is leaking / not finalizing for a while.

Should distinguish between `@[]` and `nil`.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -623,15 +626,26 @@ proc forkchoiceUpdated*(p: Eth1Monitor,
...
+        prevRandao: FixedBytes[32] randomData,
+        suggestedFeeRecipient: suggestedFeeRecipient)))
+  else:
+    p.dataProvider.web3.provider.engine_forkchoiceUpdatedV2(
+      forkchoiceState,
+      some(engine_api.PayloadAttributesV2(
+        timestamp: Quantity timestamp,
+        prevRandao: FixedBytes[32] randomData,
+        suggestedFeeRecipient: suggestedFeeRecipient,
+        withdrawals: mapIt(withdrawals, it.asEngineWithdrawal))))
```
etan-status: why are the withdrawals even in fcU to begin with?! is this intended or a bug?
newPayload already provides them.

or is this some weird hint to anticipate a followup `getPayload`? that one, curiously, no longer takes withdrawals.

wondering whether the withdrawals need to be re-supplied each time that fork choice selects new head.

etan-status: and, can a light client still use `forkchoiceUpdatedV1` to simply sync to a new head, without knowing expected withdrawals?

tersec: Two types of fcU, proposal and non-proposal. Non-proposal is effectively unchanged and supplies `null` `PayloadAttributes` (`V1` or `V2`, either way). Proposal supplies either a https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#PayloadAttributesV1 or https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#PayloadAttributesV2 so the EL can construct, for, yes, `getPayload`, a payload with the correct `withdrawals` field.

It's the same flow as pre-capella, but with a new field in `PayloadAttributes` to match a new field in `ExecutionPayload`.

This isn't a spec bug. https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2 and https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_forkchoiceupdatedv1 have details.

beacon_chain/eth1/eth1_monitor.nim
tersec: why would it break?

etan-status: https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2

> Similarly, if the functionality is not activated, client software MUST return an INVALID status with the appropriate latestValidHash if payloadAttributes.withdrawals is not null. Blocks without withdrawals MUST be expressed with an explicit empty list [] value.

beacon_chain/eth1/eth1_monitor.nim
tersec: > and, can a light client still use `forkchoiceUpdatedV1` to simply sync to a new head, without knowing expected withdrawals?

yes, presumably. or a `forkchoiceUpdatedV2` with `null` `PayloadAttributes`.

beacon_chain/eth1/eth1_monitor.nim
tersec: also it's not a `ref`, so it can't be `nil`

beacon_chain/eth1/eth1_monitor.nim
etan-status: Ah, so, for non-proposal, the entire `PayloadAttributes` would be `null`.

And, for proposals, a non-null `PayloadAttributes` is passed. Before Capella, that one includes a `null` withdrawals field, and after Capella, a non-`null` one.

As for the light client, it would only need to use the `PayloadAttributes` = null version.

Gotcha, thanks for the explanation!

tersec: `getPayload` didn't take anything but a `PayloadID` to begin with. the whole point is to queue up the EL production process early with fcU with all required information, then retrieve the result via the `PayloadID` handle later

beacon_chain/eth1/eth1_monitor.nim
etan-status: True, whatever `seq` actually are :)

Maybe `Opt[seq[Withdrawal]]`, to distinguish the pre-capella case from post-capella with empty list.

beacon_chain/eth1/eth1_monitor.nim
etan-status: Ah, got it. We still need a way to pass those expected withdrawals to the "empty block proposal fallback" logic somehow, but I guess that's an issue to solve once we wire up `getPayload`. this pr focuses on fcU, so should be alright as is.

beacon_chain/eth1/eth1_monitor.nim
tersec: that's if using fcU V2. The thing I'm not certain of is, can one construct a capella block without withdrawals using a proposal fcU V1. If not, then, yes, this breaks under those conditions, but nowhere in the engine API spec does it specify this that I see.

The part you quote has nothing to do with this, because it addresses trying to use the `V2` fcU pre-Capella. This PR does not do this, because pre-capella, it assigns `@[]` to `withdrawals`, which then in Eth1 monitor uses the `V1` call.

beacon_chain/eth1/eth1_monitor.nim
etan-status: Which, then, actually has a different issue in that our empty block proposal fallback cannot really process the withdrawals. It can include them into the block, but the state root would not reflect the updated balances.

@arnetheduck what you think? Kill the empty block proposal fallback? Or can we make it so that it still works in presence of withdrawals?

beacon_chain/spec/beaconstate.nim
```diff
@@ -818,6 +818,74 @@ func get_next_sync_committee_keys(
     i += 1'u64
   res
 
+# https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/capella/beacon-chain.md#has_eth1_withdrawal_credential
```
zah, line 821: Is there a change here or was these just moved?

tersec: moved, plus alpha.1 URL to alpha.2 URL updated

beacon_chain/eth1/eth1_monitor.nim
etan-status: ahh, missed that part. Would be surprised if fcU V1 still works post-Capella, but interesting question for sure.

tersec: The actual `getPayload` will still be `V2` in capella regardless. So it's a question of whether V2 `getPayload` can pair with `V1` fcU. It it can't, then, yes, probably use `Opt[seq[Withdrawal]]`.

But part of the point of writing this is to bring out the spec question.

tersec: https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_getpayloadv2 is notably vague on this point.

beacon_chain/spec/beaconstate.nim
tersec: Except for lifting `num_validators = lenu64(state.validators)` out of the `for` loop

zah:
I've finally submitted my multi-EL-nodes PR, featuring Capella finalization tests here:
https://github.com/status-im/nimbus-eth2/pull/4465

It incorporates all useful changes from this branch as well, so I recommend closing this PR.
Please note that even it this is merged, the `unstable` branch won't be working properly for validators in capella testnets, because it's missing another important fix in the web3 repo - the current signature of the `getPayloadV2` request is not correct:

https://github.com/status-im/nim-web3/commit/d596e68c14d0cf7fb3d52023ba16cac49b615173

tersec:
> I've finally submitted my multi-EL-nodes PR, featuring Capella finalization tests here: #4465
> 
> It incorporates all useful changes from this branch as well, so I recommend closing this PR. Please note that even it this is merged, the `unstable` branch won't be working properly for validators in capella testnets, because it's missing another important fix in the web3 repo - the current signature of the `getPayloadV2` request is not correct:
> 
> [status-im/nim-web3@d596e68](https://github.com/status-im/nim-web3/commit/d596e68c14d0cf7fb3d52023ba16cac49b615173)

Alright.

arnetheduck:
it seems that #4465 has significant issues (EL going offline breaks it completely) - shall we merge this PR (and the web3 update) first, so that capella testing can proceed? given how it's isolated and solves a separate problem, delaying it further while 4465 issues are being resolved seems counterproductive

beacon_chain/eth1/eth1_monitor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4462/commits/e854e84dc0227f70d98d3c41ef4cf7f287b1d94a ensures it uses matched fcU/getPayloads

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: assuming we cannot skip withdrawals arbitrarily, we probably would need to do some implementation work create a partial eth1 state with proofs, that can apply the balances and nothing else - doable, but not trivial. this needs a bit more thought though, haven't looked at the problem in detail.

etan-status: Note that the empty block proposal fallback is _only_ triggered when the connection to the EL is generally stable (i.e., it can process `engine_newPayload` / `engine_forkchoiceUpdated` normally), but then exactly during the `engine_getPayload` it fails. If the EL connection is overall unstable, the node falls into optimistic mode and will not perform validator duties (hence also not trigger the empty block proposal fallback).

So, not sure, if `engine_getPayload` failure could be resolved with proofs. Would definitely be interested in seeing how such a partial Eth1 state solution would technically work though!

If the partial state thing is complicated, should we still keep the empty block proposal fallback for the 0 withdrawals case? or should we drop it for now post Capella? The 0 withdrawals case can only happen if the entire validator set is leaking (not sure if in that situation it is making the overall situation better or worse to propose empty blocks).

arnetheduck: > `engine_getPayload`

it's been observed to be slow

> falls into optimistic mode

if the last block for example was "difficult" we can still try to create a gappy block - see also https://github.com/status-im/nimbus-eth2/issues/3969


> still keep

I'd say yes - ie it seems better to restore consensus stability in such extreme cases and block production is essential in that process.

tersec: > assuming we cannot skip withdrawals arbitrarily

This is indeed the case.

https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/capella/beacon-chain.md#new-process_withdrawals requires this particularly:
```python
def process_withdrawals(state: BeaconState, payload: ExecutionPayload) -> None:
    expected_withdrawals = get_expected_withdrawals(state)
    assert len(payload.withdrawals) == len(expected_withdrawals)

    for expected_withdrawal, withdrawal in zip(expected_withdrawals, payload.withdrawals):
        ...
```

tersec:
> it seems that #4465 has significant issues (EL going offline breaks it completely) - shall we merge this PR (and the web3 update) first, so that capella testing can proceed? given how it's isolated and solves a separate problem, delaying it further while 4465 issues are being resolved seems counterproductive

https://github.com/status-im/nimbus-eth2/pull/4471 bumps `nim-web3`

arnetheduck (approved):

PR #4464, opened 2023-01-05
Blobs sidecars by range

henridf:
This PR adds the blobs_sidecars_by_range RPC.


beacon_chain/sync/sync_protocol.nim
```diff
@@ -398,7 +399,78 @@ p2pProtocol BeaconSync(version = 1,
     debug "Block root request done",
       peer, roots = blockRoots.len, count, found
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/altair/light-client/p2p-interface.md#getlightclientbootstrap
+  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/eip4844/p2p-interface.md#blobssidecarsbyrange-v1
```
tersec: It's rc.0 now

henridf: fixed (here and above)

henridf:
Re testing, I don't see any existing RPC tests in `tests/`. Maybe there's some indirect testing in the Jenkins CI runs? (not sure where those tests are defined; I can't see their output yet).

tersec:
> Re testing, I don't see any existing RPC tests in `tests/`. Maybe there's some indirect testing in the Jenkins CI runs? (not sure where those tests are defined; I can't see their output yet).

There aren't any of which I'm aware. There are tests for some types of RPC (the REST server, for example, in a couple different ways, including the Jenkins CI runs), but for these p2p-interface req/resp RPC, I don't believe so.

Also, one can run the Jenkins CI finalization tests locally, e.g.,
```
./scripts/launch_local_testnet.sh --preset minimal --nodes 4 --disable-htop --stop-at-epoch 8 -- --verify-finalization
```
does something similar. The Jenkins tests stop at 6 epochs or so, and also run a longer `--preset mainnet` version. https://github.com/status-im/nimbus-eth2/blob/unstable/ci/Jenkinsfile shows their precise definitions, pending your being able to access the results directly.

beacon_chain/spec/datatypes/eip4844.nim
```diff
@@ -35,6 +35,8 @@ const
 
   # https://github.com/ethereum/consensus-specs/blob/v1.3.0-alpha.2/specs/eip4844/polynomial-commitments.md#constants
   BLS_MODULUS* = "52435875175126190479447740508185965837690552500527637822603658699938581184513".u256
+  #https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.0/specs/eip4844/p2p-interface.md#configuration
```
tersec:
```suggestion
  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.0/specs/eip4844/p2p-interface.md#configuration
```

henridf:
@tersec sounds good - I guess we'll exercise this code when we can join the eip4844 devnet.

tersec (approved):

PR #4465, opened 2023-01-06
Support for driving multiple EL nodes from a single Nimbus BN

zah:
Full list of changes:

* Eth1Monitor has been renamed to ELManager to match its current responsibilities better.

* The ELManager is no longer optional in the code (it won't have a nil value under any circumstances).

* The support for subscribing for headers was removed as it only worked with WebSockets and contributed significant complexity while bringing only a very minor advantage.

* Transition configuration is now exchanged 2 weeks prior to the Bellatrix fork (at this point, relevant only for future testnets).

* The `--web3-url` parameter has been deprecated in favor of a new `--el` parameter. The new parameter supports specifying the JWT token per connection. Each connection can also be configured with a different set of responsibilities (e.g. download deposits, validate blocks and/or produce blocks). On the command-line, these properties can be configured through URL properties stored in the #anchor part of the URL. In TOML files, they come with a very natural syntax (althrough the URL scheme is also supported).

* The previously scattered EL-related state and logic is now moved to `eth1_monitor.nim` (this module will be renamed to `el_manager.nim` in a follow-up commit). State is assigned properly either to the `ELManager` or the to individual `ELConnection` objects where appropriate.

  The ELManager executes all Engine API requests against all attached EL nodes, in parallel. It compares their results and if there is a disagreement regarding the validity of a certain payload, this is detected and the beacon node is protected from publishing a block with a potential execution layer consensus bug in it.

  The BN provides metrics per EL node for the number of successful or failed requests for each type Engine API requests. If an EL node goes offline and connectivity is resoted later, we report the problem and the remedy in edge-triggered fashion.

* The client can now produce Capella execution payloads

* The local testnet simulation script now runs Geth and Nimbus by default. The minimal network simulation uses a Nimbus remote signing node for 50% of the validators. The mainnet simulation uses the official web3signer in a 2 out of 3 threshold signing config.

* The local testnet simulation can now use a payload builder. This is currently not activated in CI due to lack of automated procedures for installing third-party relays or builders.

  You are adviced to use mergemock for now, but for most realistic results, we can create a simple builder based on the nimbus-eth1 codebase that will be able to propose transactions from the regular network mempool.

* Updated the `run-geth-el.sh` script to work with the latest Capella and EIP-4844 builds of Geth and added the `run-nimbus-eth2-in-withdrawal-testnet.sh` script that can be used to launch Nimbus instances in the available withdrawal devnets.


arnetheduck:
> The --web3-url parameter has been deprecated in favor of a new --el parameter.

all documentation needs to be updated accordingly

> On the command-line, these properties can be configured through URL properties stored in the #anchor part of the URL. In TOML files, they come with a very natural syntax (althrough the URL scheme is also supported).

ditto, needs an update to the docs

> The minimal network simulation uses a Nimbus remote signing node for 50% of the validators. The mainnet simulation uses the official web3signer in a 2 out of 3 threshold signing config.

does anything test the plain and simple validators on mainnet?

> run-nimbus-eth2-in-withdrawal-testnet.sh

with config file support available, we should be moving away from these - ie they're crutches at best and make the manual hard to write (because they're not consistently used for all commands such as trusted node sync)

beacon_chain/eth1/el_conf.nim
```diff
@@ -0,0 +1,254 @@
...
+when not declared(decodeQuery):
+  # TODO
+  # This is a verbatim copy of the iterator from Nim's std library.
+  # We can remove it from the code once we stop supporting Nim 1.2.
+  iterator decodeQuery*(data: string, sep = '&'): tuple[key, value: string] =
+    ## Reads and decodes the query string `data` and yields the `(key, value)` pairs
+    ## the data consists of. If compiled with `-d:nimLegacyParseQueryStrict`,
+    ## a `UriParseError` is raised when there is an unencoded `=` character in a decoded
+    ## value, which was the behavior in Nim < 1.5.1.
+    runnableExamples:
```
arnetheduck: we've stayed clear of runnableExamples on purpose since they're hard to read and hard to distinguish from the rest of the code - is there a special reason to start using them here?

beacon_chain/eth1/el_conf.nim
```diff
@@ -0,0 +1,254 @@
...
+      assert toSeq(decodeQuery("foo=1&bar=2=3")) == @[("foo", "1"), ("bar", "2=3")]
+      assert toSeq(decodeQuery("foo=1;bar=2=3", ';')) == @[("foo", "1"), ("bar", "2=3")]
+      assert toSeq(decodeQuery("&a&=b&=&&")) == @[("", ""), ("a", ""), ("", "b"), ("", ""), ("", "")]
+
+    proc handleHexChar(c: char, x: var int): bool {.inline.} =
+      ## Converts `%xx` hexadecimal to the ordinal number and adds the result to `x`.
+      ## Returns `true` if `c` is hexadecimal.
+      ##
+      ## When `c` is hexadecimal, the proc is equal to `x = x shl 4 + hex2Int(c)`.
+      runnableExamples:
```
arnetheduck: ditto - this is unwelcome clutter when reading the code - it should be part of the comment instead

arnetheduck:
> new --el parameter.

since it's a new parameter, it's a good opportunity to default it to localhost:8551 so that a default "localhost" setup just works with as few command line parameters as possible (to be further improved with https://github.com/ethereum/execution-apis/pull/302) - it should also be possible to remove the default so that no el connection is attempted

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -266,27 +356,18 @@ when hasGenesisDetection:
           withdrawal_credentials: deposit.withdrawal_credentials)
         m.genesisValidatorKeyToIndex[pubkey] = idx
 
-  proc processGenesisDeposit*(m: Eth1Monitor, newDeposit: DepositData) =
-    m.depositsChain.db.genesisDeposits.add newDeposit
+  proc processGenesisDeposit*(m: ELManager, newDeposit: DepositData) =
```
arnetheduck: With the el manager managing multiple EL:s, it makes even more sense to move the deposit processing logic to a separate module that can be unit tested in isolation - ie there should a module whose role is to take as input eth1 blocks and produce a deposit structure / state, similar to the chaindag - all of this is synchronous logic that is orthogonal to the multiplexing of several connections - it's a very natural line along which this code could be simplified and rendered robust.

arnetheduck: ie "processing deposits" and "managing the live payload interaction between el and CL" are to completely stand-alone tasks - if this was designed like the VC, these would be separate services living on top of a manager whose role is solely to multiplex - a design like this would significantly untangle what right now is a monolith that's hard to work with

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -489,211 +571,575 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+    if not req.finished:
+      req.cancel()
+    elif req.failed:
+      error "Failed to get execution payload from EL",
+            url = m.elConnections[idx].engineUrl.url,
+            err = req.error.msg
+    elif bestPayloadIdx.isNone:
+      bestPayloadIdx = some idx
+    else:
+      if cmpGetPayloadResponses(req.read, requests[bestPayloadIdx.get].read) > 0:
```
arnetheduck: why is the payload comparison done in elmanager? it seems to me this should be the responsibility of the same module that compares blocks from all sources, ie blinded blocks included thus concentrating the block selection logic in one place

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -489,211 +571,575 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
-proc forkchoiceUpdated*(p: Eth1Monitor,
+proc sendNewPayload*(m: ELManager,
+                     payload: engine_api.ExecutionPayloadV1 | engine_api.ExecutionPayloadV2):
+                     Future[PayloadExecutionStatus] {.async.} =
+  let
+    deadline = sleepAsync(NEWPAYLOAD_TIMEOUT)
+    requests = m.elConnections.mapIt(sendNewPayloadToSingleEL(it, payload))
+    requestsCompleted = allFutures(requests)
+
+  await requestsCompleted or deadline
```
arnetheduck: awaits like this are generally redundant - one can just as well `for r in requests: await r or deadline`

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -489,211 +571,575 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+    disagreementAlreadyDetected = false
+
+  for idx, req in requests:
+    if not req.finished:
+      req.cancel()
+    else:
+      let url = m.elConnections[idx].engineUrl.url
+      if req.failed:
+        engine_newPayload_failures.inc(1, [url])
+        error "Sending payload to the EL failed",
```
arnetheduck: per our log guidelines, this is a warning since the application continues to function

beacon_chain/conf.nim
```diff
@@ -170,14 +171,12 @@ type
...
+      name: "web3-url" .}: seq[EngineApiUrlConfigValue]
 
-    web3ForcePolling* {.
-      hidden
-      defaultValue: false
-      desc: "Force the use of polling when determining the head block of Eth1"
-      name: "web3-force-polling" .}: bool
+    elUrls* {.
+      desc: "One or more Execution Layer Engine API URLs"
+      name: "el" .}: seq[EngineApiUrlConfigValue]
```
etan-status: Can we translate the previous parameters to the new ones and issue a deprecation warning? 
So that when users upgrade it doesn't randomly fail to start up?

arnetheduck:
```
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
ERR 2023-01-06 11:03:00.750+01:00 Error while syncing deposits               topics="eth1" url=ws://geth:8551 err="Transport is not initialised (missing a call to connect?)"
```

turning off the EL puts the the beacon node in a busy loop (see timing), with this PR

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -489,211 +571,575 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
-      timestamp: Quantity timestamp,
-      prevRandao: FixedBytes[32] randomData,
-      suggestedFeeRecipient: suggestedFeeRecipient)))
+      finalizedBlockHash: finalizedBlock.asBlockHash)
+    deadline = sleepAsync(FORKCHOICEUPDATED_TIMEOUT)
+    requests = m.elConnections.mapIt(
+      it.forkchoiceUpdatedForSingleEL(state, payloadAttributes))
+    requestsCompleted = allFutures(requests)
+
+  await requestsCompleted or deadline
```
arnetheduck: why do we wait for all el:s here? a single valid response would be enough to proceed with other processing?

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: also, when a single payload is valid, this should be a sufficient condition to return? otherwise, our entire staying-in-sync processing is slowed down to the slowest of all connected execution layers - this will lead to profit loss because a single slow EL (your backup EL on a raspberry) will cause the attestation subsystem to use an old head at the attestation 4s cutoff even if your "fast" EL has responded valid.

arnetheduck: this is already happening by the way, and @etan-status is suggesting an improvement that to a certain extent works around it, but the difference is on the order of several % - introducing the slowest-of-all-el:s behavior here will have a real impact.

etan-status: Not sure if you are referring to validator duties timings here, or something else. If it is the former:

The current in-process logic waits 4 seconds until block is received + processed.
If one is received and processed, it adds 2 seconds delay and then attests (so, at 6s into the slot max).
If none is received or processing is still ongoing, it attests at 4s into the slot to the old head.

Proposal to change in a way that:
If block is received but still processing, give it until 7s into the slot before attesting to the old head.
If block is not received within 4s, keep attesting at 4s.

Makes a noticeable difference in validator rewards on Raspi with Besu, as Besu frequently takes longer than 4s to process blocks on that lowend system. Geth requires periodic multiday pruning which tends to kill the DB if the disk becomes full during that. Nethermind/Erigon need 16G ram so cannot be used on Pi. So, not too many alternatives there.

arnetheduck: yes, validator duties timer - the newpayloadtimeout is 8s meaning that a slow EL pushes it past the 4s mark - while I like your proposed patch and we should do it, it unfortunately trivially only covers in-bn validators - there's currently, I think, no way to transmit "verifying block" to the vc (unless you have some clever trick up your sleeve?)

etan-status: No, haven't. Maybe it is also something that could be addressed in the spec instead. The three intervals per slot could be tweaked overall to leave more room for block execution. e.g., 7s for execution, 3s for aggregation, and 2s for creating the new block, instead of 4/4/4.

arnetheduck: it's been discussed a few times indeed, but without concrete results afair.

it's doable in a few ways, ie either via a special nimbus request or maybe by delaying the response when vc requests what attestation data to sign (the vc would then need to be taught to wait sufficiently long as well, which maybe it does already...)

beacon_chain/eth1/el_conf.nim
zah: This code is copy pasted from the Nim standard library because it was introduced in Nim 1.4. It will be removed once we drop support for Nim 1.2 as indicated in the TODO comment above it.

beacon_chain/eth1/el_conf.nim
zah: ditto :)

beacon_chain/eth1/eth1_monitor.nim
zah: These activities are already independent enough. They run as different "asynchronous threads" of execution which both share access to the `Eth1Chain` structure which is indeed a bit like the chaindag.

Otherwise, I've wanted to reduce the required inputs of this particular function further in order to make it more reusable and easier to test, but the main problem is that it has multiple different responsibilities:

1) It must download all deposits, while forming a sparse chain of all blocks with deposits
2) It must download all recent blocks, even if they don't have deposits
3) It must obtain the timestamps for some of the blocks (currently, it's not doing this optimally and I plan to submit a follow-up PR fixing this).

As a result, it has a somewhat awkward monolithic internal design in order to allow these goals to be achieved with the minimal number of requests. I'll add some comments on the top to explain this. Removing the `ELManager` input is a worthy goal which is relatively within reach.

beacon_chain/eth1/eth1_monitor.nim
zah: This is a simple utility function. I agree that it can be moved to another module once it's applied in more places. 

Since you really want the best block, it doesn't matter if comparisons are done in multiple places. It would be silly to return multiple block candidates from the EL manager, just so there could be a single place where all blocks will be compared at the end.

beacon_chain/conf.nim
arnetheduck: our "standard" way of notifying the user: https://github.com/status-im/nimbus-eth2/blob/05a04236b9f7d85f40fcdc3e3ad88cc2fa311bfe/beacon_chain/nimbus_beacon_node.nim#L1788 - I'd love to see support for handling command line evolution in confutils proper though

beacon_chain/eth1/eth1_monitor.nim
zah: This currently implements the "best" strategy in VC terms. You are arguing that it should use the "first" strategy instead. I would argue that we need more configuration options, for selecting the strategy and controlling the timeout :)

beacon_chain/eth1/eth1_monitor.nim
zah: Err, only "fatal" indicates that the application won't continue to function.
For me, "error" means "the application failed to perform its intended function". A "warning" should be something closer to "there are conditions that may cause the application to fail to perform its function"

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: we removed those configuration options from the VC, mainly because options is what you do when you don't know what to do and pass on the problem to the user, and the complexity seemed unwarranted - in this particular case, once there is a `valid` response, you will not find a "better" `best` option and waiting for it is strictly worse than the alternatives.

beacon_chain/eth1/eth1_monitor.nim
zah: This should be again configurable perhaps, but the current logic in `compareStatuses` is implementing the following protection described in the commit message:

> The ELManager executes all Engine API requests against all attached EL nodes, in parallel. It compares their results and if there is a disagreement regarding the validity of a certain payload, this is detected and the beacon node is protected from publishing a block with a potential execution layer consensus bug in it.

In general, I think we should recommend to all professional operators to run at least 2 highly responsive EL instances. The "Raspberry Pi" fallback that you described can be configured as an EL with another role.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: ie the point here is that we consider "valid" to be authoritative - the "potential" downside is that we don't take majority opinions on valid/invalid for example, but as a general principle, we've stayed away from working around EL processing bugs.

arnetheduck: the same applies to "invalid" btw - the only equivocating response we have is "syncing" - ie in that case, it's better to wait for other responses

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: let's move the conversation to the other comment above

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: > In general, I think we should recommend to all professional operators to run at least 2 highly responsive EL instances. The "Raspberry Pi" fallback that you described can be configured as an EL with another role.

depending on the EL software, this will regularly degrade attestation performance - ie if besu process a certain block faster than nethermind, it doesn't matter that both are run on high-performance setups - we will always see the worst of all worlds.

running 2 instances does not give a majority so you need to run at least 3, but _even then_ you should not wait for the last response if you get 2 agreeing responses (which will be the absolute majority of cases since bugs are rare)

beacon_chain/conf.nim
zah: The old names will continue to be supported for a while. I think I already added a deprecation warning, but will double check.

beacon_chain/eth1/eth1_monitor.nim
zah: I don't think the use case of having a "backup EL on a Raspberry Pi" is particularly useful, so if there are no options, I think it makes more sense for us to have the protection from consensus bugs in place. See my other comment here:

https://github.com/status-im/nimbus-eth2/pull/4465#discussion_r1063500809

Although, I think having this disdain for options is not really serving our users.

arnetheduck: Based on practical experience with rocket pool, this is a very popular option that has caused many users to use clients other than nimbus - ie they have the "fallback node" support where a backup execution layer client is kept in sync so that the main one can be pruned and taken offline for os/hardware upgrades (they do it by running two CL:s). 

It significantly decreases downtime for both of those cases - I expect many stakers will want to use it, professional or not (in particular when geth requires pruning and the other clients all have significant downsides like high mem usage or general stability problems) given its popularity there.

arnetheduck: And again: this is not just for raspberries - it happens regularly that clients stall in block processing and push timings beyond the 4s window for different reasons - memory usage, swap files, garbage collection etc, _and_ waiting for all clients is wrong even if we want protection by majority.

As long as we have this all strategy in place, we need to document that it will lead to lower profits, because it will (and nimbus will fare poorly in comparisons as a result).

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: > These activities are already independent enough. 

this is a comment about the code complexity - the monolithic design has regularly caused bugs and confusion when trying to approach the code for changes.

beacon_chain/eth1/eth1_monitor.nim
zah: Well, in my own setup I run Geth and Nethermind on a relatively fast server and both are equally responsive. Introducing such performance asymmetry in your infrastructure is asking for trouble in general. 

I would also argue that most users would prefer the "Best" option when it comes to block production, but perhaps this is something you don't disagree with?

beacon_chain/eth1/eth1_monitor.nim
zah: The previous code has quite a bit more complex control flow because it was relying on things like async events which can be fired from either the websocket header subscriptions or HTTP polling. The current version is quite a bit simpler and introducing a notion of "services" will only bring back some of the complexity.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: well, the point is to concentrate the selection strategy to a single place - at that point it becomes feasible to create other strategies for block selection (ie ensuring that a block contains a particular tx etc) - the elmanager already has too many responsibilities, many of which are not tested - this is one more which could be moved to a place where it could become a regular part of the test suite - since this is a recurring design problem in the elmanager, it's a good opportunity to get it done sooner rather than later.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: block production is different because it doesn't have the same timeout / timing dependency the way that newpayload does, with respect to attestation production - 

that said, as part of the "best" strategy for block production, we must perform a cutoff early enough that we have time to apply the block and have it broadcast throughout the network. It doesn't make sense to wait 8 seconds for a potentially better block only to have it be discarded by the network.

Notably, this entire comment thread is about newpayload which is used in storeblock - until it has finished processing, the head will not be updated and therefore the async timeouts for attestation production cutoff will kick in and we'll attest to an old head.

arnetheduck: > Introducing such performance asymmetry in your infrastructure is asking for trouble in general.

as noted - from experience this is done by many stakers so that they have a live backup of the EL (which is slow to sync) for upgrades and prunings - it's only a bad idea iff the CL unnecessarily waits for all EL:s - if instead it picks the fastest EL (or the fastest majority), it's a win-win.

zah: Perhaps there is a way to resolve these trade offs with a design that is built around a "stream" of block statuses instead of a single definitive response. You would update the head on first confirmation, but if you see the block being invalidated, you'll blacklist it and move to the next best block.

beacon_chain/eth1/eth1_monitor.nim
zah: Sorry, but I still find the requested design silly. If you want to have a single block selection algorithm, one easy way to implement this would be to inject a `BlockSelector` dependency in the EL manager that will be called in place of the `cmpGetPayloadResponses` function. Since any block selection algorithm is likely to be a simple functions will very little configurable state, it's trivial to test in unit tests without depending on any part of the `ELManager`.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: not really - there's a portion of the logic that is not asynchronous and not dependent on EL communication - it checks for consistency of a new block with respect to the others and has an independent API with error handling and logic - it would be exactly the same if it was driven by an internal eth1 engine without async comms, or indeed a feed of blocks read from disk - this is the part that doesn't belong here, but rather in a separate module which receives tests etc - the code in elmanager at that point is a dumb data router, as it should be.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: this would be one way to do it, but ideally it would be able to also take into consideration blinded blocks (for example, a popular strategy is to use a blinded block iff the reward is greater by a percentage compared to an unblinded block, so as to give slight preference to censorship resistance, but not at the expense of large profits). 

It is the mixing of all these responsiblities that has led to the poor track record of elmanager over time - the tests don't get written _because_ it is too complex and does too many things and the way to solve that is to split its responsiblities into parts that relate to logic and parts that relate to connection management and handling.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: Well, its intended function is to follow consensus - because sending a payload to an EL only mildly interferes with that (ie we keep following consensus, and the problem resolves itself as soon as the EL is back online), this is not a problem - specially when there are multiple EL:s, the failure of a single EL is not an error in any significant way, for the CL.

A recent and related example is https://github.com/status-im/nimbus-eth2/pull/4458 - we can make forward progress without an EL, therefore it is not an error, but rather a warning about a degradation due to a failure in an external component.

beacon_chain/eth1/eth1_monitor.nim
zah: Well, sure - the `Eth1Chain` object can be moved to a separate module. It's already used in the non-asynchronous way that you describe as part of the `block_sim` simulation which injects a random number of deposits and uses them during block production.

beacon_chain/eth1/eth1_monitor.nim
arnetheduck: well, another option is to timeout earlier if we have at least one response - basically, it's fine to not update head as long as we're not sending out the attestation, meaning that we can wait for responses as long as the 4s mark is not hit (with some margin) - this should be quite simple to implement as well without disrupting too much of this branch - if the first deadline hits and there are no answers, we make another wait for the extended past-attestation deadline.

tersec:
> Transition configuration is now exchanged 2 weeks prior to the Bellatrix fork (at this point, relevant only for future testnets).

Future testnets will, by all appearances, start merged and in Bellatrix or newer.

.gitmodules
```diff
@@ -204,3 +204,6 @@
 	path = vendor/capella-testnets
 	url = https://github.com/ethpandaops/withdrawals-testnet.git
 	branch = master
+[submodule "vendor/withdrawals-testnets"]
+	path = vendor/withdrawals-testnets
```
tersec, line 204: It seems like there's already a copy of this at `vendor/capella-testnets`?

beacon_chain/spec/engine_authentication.nim
```diff
@@ -6,7 +6,8 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 import
-  chronicles,
+  std/[base64, json, options, os, strutils],
```
tersec: To articulate one reason why I tend to prefer the explicit `from ... import ...` form: it means that there can, in many cases, then be a 1:1 matching between any given symbol (e.g., `JsonNode` or `isNone`) in the imports and some use of it in that module, either directly or via `export`.

When minimizing reproducible test cases, I very often run across examples of imports the removal of which don't break compilation. My goal in those cases isn't to create a functional Ethereum client, so I don't typically know or investigate whether they're changing some overload selection in an incorrect way, which is one reason I typically don't push this kind of cleanup in the actual codebase, but there's some evidence at least that Nim's unused module warning or hint is conservative, that is, produces false negatives at a significant rate. The `from ... import ...` form helps alleviate the issues that then result of just inexorable buildup of module imports which Nim won't warn about (especially if they're through one of the giant `import`/`export` funnel systems where somewhere, indirectly, something from that import is used) but which it's not obvious how it's used, so it just says, maybe useless.

Here, it's clear that, say, if `replace` or `os`'s `/` symbols are not used visibly, those imports can be dropped. This import refactoring removes that visual cue, and makes it more likely that to even begin investigating if/when some imports can be removed from this (or any other module using just `import foo`), the first step would be to revert this change and enumerate the actual used symbols.

Thus, I disagree with this change.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -375,15 +375,31 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
           return RestApiResponse.jsonError(Http400, InvalidRandaoRevealValue)
 
         let res =
-          if qslot.epoch >= node.dag.cfg.CAPELLA_FORK_EPOCH:
-            await makeBeaconBlockForHeadAndSlot[capella.ExecutionPayload](
+          case node.dag.cfg.consensusForkAtEpoch(qslot.epoch)
+          of ConsensusFork.Deneb:
+            # TODO denebImplementationMissing
```
tersec: When feasible, (such as here) it's preferable to actually ensure `denebImplementationMissing` is a symbol in the code, not just a magic greppable string.

The point is that eventually this `denebImplementationMissing` constant in `datatypes/base.nim` can be removed, and that will mean that there aren't significant known portions of Deneb missing, because if there were, there would be a compiler error, because those things would have been referencing the actual `denebImplementationMissing` constant. That is, it's sort of a compiler-checked TODO.

There's a range of approaches used elsewhere -- here, for example, some variation of
```nim
            discard $denebImplementationMissing
```
might be reasonable.

.gitmodules
arnetheduck: each of goerli and sepolia has its own repo - seems like it would be better to switch to those, since they seem more "canonical"

beacon_chain/conf.nim
```diff
@@ -176,14 +177,17 @@ type
...
-      hidden
+    elUrls* {.
+      desc: "One or more execution layer Engine API URLs"
+      name: "el" .}: seq[EngineApiUrlConfigValue]
+
+    noEl* {.
       defaultValue: false
-      desc: "Force the use of polling when determining the head block of Eth1"
-      name: "web3-force-polling" .}: bool
+      desc: "Don't use an EL. The node will remain optimistically synced and won't be able to perform validator duties"
```
arnetheduck, line 189: might as well reuse the `--optimistic` flag for this

beacon_chain/conf.nim
```diff
@@ -1317,8 +1321,21 @@ proc loadJwtSecret*(
...
+proc engineApiUrls*(config: BeaconNodeConf): seq[EngineApiUrl] =
+  let elUrls = if config.noEl:
+    return newSeq[EngineApiUrl]()
+  elif config.elUrls.len == 0 and config.web3Urls.len == 0:
+    @[defaultEngineApiUrl]
+  else:
+    config.elUrls
+
+  if config.web3Urls.len > 0:
+    warn "Config option is deprecated", option = "web3-url"
```
arnetheduck: I'm not sure we _really_ need this deprecation - ie the flag can live on for some time before we start nagging people about it - it has a well-defined meaning even in the new world and supporting it seems trivial, so why bother users with this?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -275,7 +407,16 @@ func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV1):
...
       mapIt(rpcExecutionPayload.transactions, it.getTransaction)))
 
-func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV2):
+func asConsensusType*(payloadWithValue: BellatrixExecutionPayloadWithValue):
+    bellatrix.ExecutionPayloadForSigning =
+  bellatrix.ExecutionPayloadForSigning(
+    executionPayload: payloadWithValue.executionPayload.asConsensusType,
+    blockValue: payloadWithValue.blockValue)
+
+template deref[T](o: Option[T]): T = o.get
```
arnetheduck: so the potential Defect is .. ok here?

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -275,7 +407,16 @@ func asConsensusExecutionPayload*(rpcExecutionPayload: ExecutionPayloadV1):
...
+template deref[V](v: V): V = v
+
+func asConsensusType*(rpcExecutionPayload: ExecutionPayloadV1OrV2|ExecutionPayloadV2):
```
tersec, line 419: This type presumably comes up because `getPayloadV2` is [documented](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#response-2) to return (effectively) this type. But that's misleading:
> #### Response
>
>* result: `object`
>  - `executionPayload`: [`ExecutionPayloadV1`](./paris.md#ExecutionPayloadV1) | [`ExecutionPayloadV2`](#ExecutionPayloadV2) where:
>      - `ExecutionPayloadV1` **MUST** be returned if the payload `timestamp` is lower than the Shanghai timestamp
>      - `ExecutionPayloadV2` **MUST** be returned if the payload `timestamp` is greater or equal to the Shanghai timestamp
>  - `blockValue` : `QUANTITY`, 256 Bits - The expected value to be received by the `feeRecipient` in wei
>* error: code and message set in case an exception happens while getting the payload.

And the CL provides the `timestamp` via its fcU call. So `nimbus-eth2` knows, with certainty, which of V1 or V2 it should get back, and should reject the other with an error. I see that https://github.com/status-im/nim-web3/pull/77 argues that this might be something required for `nimbus-eth1`, but `nimbus-eth2` can always deal in either `ExecutionPayloadV1` or `ExecutionPayloadV2`, whichever is dictated by the timestamp.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,254 +611,745 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+template payload(response: engine_api.GetPayloadV3Response): engine_api.ExecutionPayloadV3 =
+  response.executionPayload
+
+template toEngineWithdrawals*(withdrawals: seq[capella.Withdrawal]): seq[WithdrawalV1] =
+  mapIt(withdrawals, toEngineWithdrawal(it))
+
+template toFork(T: type ExecutionPayloadV1): ConsensusFork =
+  ConsensusFork.Bellatrix
+
+template toFork(T: typedesc[ExecutionPayloadV1OrV2|ExecutionPayloadV2]): ConsensusFork =
```
tersec, line 850: Is this correct? The other issues with `ExecutionPayloadV1OrV2` aside, what this would look like for a CL is that it's sent getPayloadV2 and because it's still in Bellatrix, gets back a V1 response. That's valid engine API. However, the fork for `ExecutionPayloadV1OrV2` varies depending on the timestamp and whether it has any withdrawals.

But really, it shows more of the inapplicability of this `ExecutionPayloadV1OrV2` type to `nimbus-eth2`.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,254 +611,745 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+      req.cancel()
+    elif req.failed:
+      error "Failed to get execution payload from EL",
+            url = m.elConnections[idx].engineUrl.url,
+            err = req.error.msg
+    else:
+      const payloadFork = PayloadType.toFork
+      when payloadFork >= ConsensusFork.Capella:
+        when payloadFork == ConsensusFork.Capella:
+          # TODO: The engine_api module may offer an alternative API where it is guaranteed
```
tersec, line 902: This API already exists though? https://github.com/status-im/nim-web3/blob/master/web3/engine_api.nim

```nim
template getPayload*(
    rpcClient: RpcClient,
    T: type ExecutionPayloadV1,
    payloadId: PayloadID): Future[ExecutionPayloadV1] =
  engine_getPayloadV1(rpcClient, payloadId)

template getPayload*(
    rpcClient: RpcClient,
    T: type GetPayloadV2Response,
    payloadId: PayloadID): Future[GetPayloadV2Response] =
  engine_getPayloadV2(rpcClient, payloadId)

template getPayload*(
    rpcClient: RpcClient,
    T: type GetPayloadV3Response,
    payloadId: PayloadID): Future[GetPayloadV3Response] =
  engine_getPayloadV3(rpcClient, payloadId)
```

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,254 +611,745 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+              actualChain = distinctBase(providerChain)
+        connection.etcStatus = EtcStatus.mismatch
+        return
+    except CatchableError as exc:
+      # Typically because it's not synced through EIP-155, assuming this Web3
+      # endpoint has been otherwise working.
+      debug "Failed to obtain eth_chainId",
+             error = exc.msg
+
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1
```
tersec: This reverts from a v1.0.0-beta.2 link to a v1.0.0-beta.1 link in this PR

beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -29,6 +29,9 @@ const
   # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.2/src/engine/shanghai.md#request-1
   FORKCHOICEUPDATED_TIMEOUT* = 8.seconds
 
+  # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#request
```
tersec: In `unstable`, this is a beta.2 URL

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1348,33 +1888,44 @@ proc syncBlockRange(m: Eth1Monitor,
...
-  # an arbitary state, so we better reset it (not doing this has resulted
-  # in resource leaks historically).
-  if not m.dataProvider.isNil and needsReset:
-    # We introduce a local var to eliminate the risk of scheduling two
-    # competing calls to `close` below.
-    let provider = m.dataProvider
-    m.dataProvider = nil
-    await provider.close()
+  while true:
+    # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.1/src/engine/specification.md#engine_exchangetransitionconfigurationv1
```
tersec: beta.1 URL (beta.2 is current)

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -213,16 +213,17 @@ proc storeBackfillBlock(
 
 from web3/engine_api_types import PayloadExecutionStatus, PayloadStatusV1
 from ../eth1/eth1_monitor import
-  Eth1Monitor, asEngineExecutionPayload, ensureDataProvider, newPayload
+  ELManager, asEngineExecutionPayload, sendNewPayload,
+  forkchoiceUpdated, forkchoiceUpdatedNoResult
 
 proc expectValidForkchoiceUpdated(
-    eth1Monitor: Eth1Monitor,
-    headBlockHash, safeBlockHash, finalizedBlockHash: Eth2Digest,
```
tersec, line 220: This change from `Root` to `Hash` was intentional: https://github.com/status-im/nimbus-eth2/pull/4614 -- is this an intentional revert?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -233,7 +234,7 @@ proc expectValidForkchoiceUpdated(
   # previous `forkchoiceUpdated` had already marked it as valid. However, if
   # it's not the block that was received, don't info/warn either way given a
   # relative lack of immediate evidence.
-  if receivedExecutionBlockHash != headBlockHash:
+  if receivedExecutionBlockRoot != headBlockRoot:
```
tersec: See above, regarding all these `Hash`/`Root` diffs

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -102,44 +112,96 @@ type
...
-    ReadyToRestartToPrimary
-    Failed
-    Stopping
-    Stopped
-
-  Eth1Monitor* = ref object
-    state: Eth1MonitorState
-    startIdx: int
-    web3Urls: seq[string]
+  ForkedPayloadAttributesKind* {.pure.} = enum
```
tersec, line 115: This `ForkedPayloadAttributes` case object type is unnecessary and doesn't especially simplify things.

It's used in basically two ways, to track `nextExpectedPayloadParams` and to construct payloads for fcU calls.

For the `nextExpectedPayloadParams` though, it only cares about equality -- so it'd be more efficient to use an encoding there (e.g., a small htr of a few-item list of hashes, but there are many approaches which work) to compare so `asSameAs` can be used equivalently. This case object machinery largely adds, well, cases to this flow.

For payloads to fcU calls, there's this odd wrapping/unwrapping thing which happens, where, e.g., `runProposalForkchoiceUpdated` now constructs a forked/case object then immediately attempts to fcU with it:
```nim
 let
    payloadAttributes = withState(self.dag.headState):
      when stateFork >= ConsensusFork.Capella:
        ForkedPayloadAttributes(
          kind: ForkedPayloadAttributesKind.v2,
          v2: PayloadAttributesV2(
            timestamp: Quantity timestamp,
            prevRandao: FixedBytes[32] randomData,
            suggestedFeeRecipient: feeRecipient,
            withdrawals: toEngineWithdrawals get_expected_withdrawals(forkyState.data)))
      else:
        ForkedPayloadAttributes(
          kind: ForkedPayloadAttributesKind.v1,
          v1: PayloadAttributesV1(
            timestamp: Quantity timestamp,
            prevRandao: FixedBytes[32] randomData,
            suggestedFeeRecipient: feeRecipient))
  try:
    let
      safeBlockRoot = beaconHead.safeExecutionPayloadHash
      (status, _) = await self.elManager.forkchoiceUpdated(
        headBlockRoot,
        safeBlockRoot,
        beaconHead.finalizedExecutionPayloadHash,
        payloadAttributes = payloadAttributes)
    debug "Fork-choice updated for proposal", status
  except CatchableError as err:
    error "Engine API fork-choice update failed", err = err.msg
```
The first thing, of course, which `forkchoiceUpdated` does is to unwrap this case object and call the specific engine APIs:
```nim
proc forkchoiceUpdated(rpcClient: RpcClient,
                       state: ForkchoiceStateV1,
                       payloadAttributes: ForkedPayloadAttributes): Future[ForkchoiceUpdatedResponse] =
  if payloadAttributes == nil:
    rpcClient.engine_forkchoiceUpdatedV1(state, none PayloadAttributesV1)
  else:
    case payloadAttributes.kind
    of ForkedPayloadAttributesKind.v1:
      rpcClient.engine_forkchoiceUpdatedV1(state, some payloadAttributes.v1)
    of ForkedPayloadAttributesKind.v2:
      rpcClient.engine_forkchoiceUpdatedV2(state, some payloadAttributes.v2)
```
Why? If `forkchoiceUpdated` is passed, directly, a `PayloadAttributesV1` or `PayloadAttributesV2`, which the (in this example) `runProposalForkchoiceUpdated` function knows because it aleady does some `stateFork >= ConsensusFork.Capella`, it can call the engine API without additional memory allocation or dynamic dispatch, and with full typechecking.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -348,11 +508,17 @@ func asEngineExecutionPayload*(executionPayload: bellatrix.ExecutionPayload):
...
 
-func asEngineExecutionPayload*(executionPayload: capella.ExecutionPayload):
+template toEngineWithdrawal(w: capella.Withdrawal): WithdrawalV1 =
+  WithdrawalV1(
+    index: Quantity(w.index),
+    validatorIndex: Quantity(w.validator_index),
+    address: Address(w.address.data),
+    amount: Quantity(w.amount))
+
+func asEngineExecutionPayload*(executionPayload: capella.ExecutionPayload | deneb.ExecutionPayload):
```
tersec: Why should `asEngineExecutionPayload(executionPayload: deneb.ExecutionPayload)` return `ExecutionPayloadV2`?

Slightly later, in the version of `eth1_monitor.nim` in this PR:
```nim
func asEngineExecutionPayload*(executionPayload: deneb.ExecutionPayload):
    ExecutionPayloadV3 =
```
are these both intentional? The latter seems more correct, and either way, the overlap seems odd.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,254 +611,745 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+      connection.lastPayloadId.get
+    elif not headBlock.isZero:
+      engine_api_last_minute_forkchoice_updates_sent.inc(1, [connection.engineUrl.url])
+
+      let response = await rpcClient.forkchoiceUpdated(
+        ForkchoiceStateV1(
+          headBlockHash: headBlock.asBlockHash,
+          safeBlockHash: safeBlock.asBlockHash,
+          finalizedBlockHash: finalizedBlock.asBlockHash),
+        makeForkedPayloadAttributes(
```
tersec, line 785: This already knows whicih type it's going to create, so per my [comment pertaining to `runProposalForkchoiceUpdated`](https://github.com/status-im/nimbus-eth2/pull/4465#discussion_r1125550993) applies here too. This calls `forkchoiceUpdated` using `makeForkedPayloadAttributes` (which is only used here) on a type it must statically know, because of `GetPayloadResponseType`, which will then be immediately unwrapped, creating a superfluous case object with memory allocation, copying, et cetera, rather than just directly calling the appropriate version of the engine API fcU depending on `GetPayloadResponseType`.

.gitmodules
zah: Right, I'll remove the `capella-testnets` instance as it is renaming the upstream repository which lead to this mistake.

beacon_chain/spec/engine_authentication.nim
zah: I don't remember adding these. They look like a conflict resolution mistake. Removing them worked out fine.

beacon_chain/conf.nim
zah: OK, this can wait indeed.

beacon_chain/conf.nim
zah: I don't think the `--optimistic` flag has gained enough popularity with the userbase to justify preserving it. The `--no-el` name is more descriptive.

beacon_chain/rpc/rest_validator_api.nim
zah: Suggestion applied

beacon_chain/conf.nim
tersec: They're semantically different: optimistic means, well, optimistic. Can use EL if there, doesn't have to if not. no-el suggests a kind of refusal to use the EL, more actively. Which is more accurate/representative of intention here?

.gitmodules
tersec: > each of goerli and sepolia has its own repo - seems like it would be better to switch to those, since they seem more "canonical"

Goerli can't be done until https://github.com/eth-clients/goerli/pull/160 is merged

beacon_chain/eth1/eth1_monitor.nim
zah: Yes, the intention of this helper is to be used just like `get`, but in generic code paths where the affected values are not always options and the `deref` may become a no-op (this situation arises with the various payload types as defined by the `web3` package right now). The derefs are guarded with the usual `isNone` checks.

For now, I've renamed this to `maybeDeref` to make the intention slightly more clear. The need for the helper may be eliminated through a future refactoring of the `web3` package which is mentioned in a TODO comment near the usage site of `maybeDeref`.

.gitmodules
tersec: > each of goerli and sepolia has its own repo - seems like it would be better to switch to those, since they seem more "canonical"

https://github.com/status-im/nimbus-eth2/pull/4694

beacon_chain/eth1/eth1_monitor.nim
zah: The `ExecutionPayloadV1OrV2` was introduced by @AdamSpitz in his effort to implement the server-side of the Engine API in nimbus-eth1. I haven't yet examined carefully the rationale there, but I had to adapt our code here to handle the latest changes in the `web3` package.

Otherwise, what you wrote is correct and we are already requiring that the response from the EL includes this optional field here:

https://github.com/status-im/nimbus-eth2/blob/6a3dc07a80e14df1f93fce89d24a187066672af6/beacon_chain/eth1/eth1_monitor.nim#L905

beacon_chain/eth1/eth1_monitor.nim
zah: The problem is the following definition:

https://github.com/status-im/nim-web3/blob/master/web3/engine_api_types.nim#L68-L70

We need an alternative version of `engine_getPayloadV2` which is guaranteed to return a `ExecutionPayloadV2`.

beacon_chain/eth1/eth1_monitor.nim
zah: The same problem is already discussed in my other answers:
https://github.com/status-im/nimbus-eth2/pull/4465#discussion_r1125564798
https://github.com/status-im/nimbus-eth2/pull/4465#discussion_r1125565473

beacon_chain/eth1/eth1_monitor.nim
tersec: Alright, I see the issue at least. It should be fixable in `nim-web3` before introducing all these workarounds in `nimbus-eth2` which (ideally) are then removed and (practically) might not be.

beacon_chain/eth1/eth1_monitor.nim
zah: This is a potentially good idea, but I would prefer exploring it in another PR as the current code is correct, even if less elegant.

beacon_chain/conf.nim
zah: Yes, this is correct.

beacon_chain/eth1/eth1_monitor.nim
zah: The suggested interim strategy has been implemented. The client will wait try to detect a consensus bug by giving the multiple ELs 1 second to respond. If some of them fail to respond in time, their response will not be taken into account if there is already an answer from another EL. In case, nobody responds within 1 second, we still wait for the spec-mandated 8 seconds.

Future PRs will explore potential designs based on a "streaming" notion of the FcU and newPayload statuses.

beacon_chain/eth1/eth1_monitor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4696

beacon_chain/eth1/eth1_monitor.nim
tersec: First step: https://github.com/status-im/nim-web3/pull/78

I realize that this is partly how PHP ended up with names like `mysql_real_escape_string`, but I'm open to suggestions on that point.

The key thing is that going forward, including in Deneb/Cancun, sharing the actual call signatures will be counterproductive, so the least inelegant solution will be to have separate `GetPayload` call signatures targeted at `nimbus-eth1` and `nimbus-eth2`, which can share most supporting infrastructure (`ExecutionPayloadV1`/`ExecutionPayloadV2`/`ExecutionPayloadV3` types, et cetera).

AdamSpitz: It makes sense to me that nimbus-eth1 and nimbus-eth2 might need different APIs, or that you might need wrappers or something to give you the more-specific types.

But I'm still a bit confused (though maybe this is a naive question - I'm fuzzy on what all the possible use cases are). Do you not also need a version of GetPayloadV2ResponseExact (called GetPayloadV2ResponseExactContainingPayloadV1ForCryingOutLoud) that contains an ExecutionPayloadV1? The [spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#response-2) currently says: "ExecutionPayloadV1 MUST be returned if the payload timestamp is lower than the Shanghai timestamp". Is nimbus-eth2 never going to be calling engine_getPayloadV2 with a payload whose timestamp is lower than the Shanghai timestamp? (e.g. Can you always count on engine_getPayloadV1 also being available? Because [this](https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md#engine_exchangecapabilities) seems to imply that the EL client is allowed to deprecate the V1 methods and only provide V2, or whatever.)

tersec: > It makes sense to me that nimbus-eth1 and nimbus-eth2 might need different APIs, or that you might need wrappers or something to give you the more-specific types.

Well, don't see the point of a wrapper. The call sigs aren't that big, and might as well leverage the serialization API checking of correct format, rather than getting some maybe-invalid-JSON and having to separately check/reject/convert it after the fact.

> Do you not also need a version of GetPayloadV2ResponseExact (called GetPayloadV2ResponseExactContainingPayloadV1ForCryingOutLoud) that contains an ExecutionPayloadV1?

`nimbus-eth2` might, yeah. After enough networks (in particular, probably worth waiting for EF mainnet) have moved to Shapella, then can use the Engine API v2 to get the value of each bid from the engine API endpoints and compare them against each other as well as the builder API endpoints.

I don't think that's currently done routinely, because it depends on basically Shapella-ready ELs, which `nimbus-eth2` does not assume are deployed.

> e.g. Can you always count on engine_getPayloadV1 also being available?

Long enough to unconditionally switch to `engine_getPayloadV2`, probably yes.

But in general your points are well-taken. Once `nimbus-eth2` can switch to always relying on `engine_getPayloadV2` or newer, then it will make sense to replace (and probably not keep) `GetPayloadV2ResponseExact` with `GetPayloadV2ResponseExactContainingPayloadV1ForCryingOutLoud` for the reasons you note.

The invariant here is that regardless of which engine API version `nimbus-eth2` calls, it will know exactly which `ExecutionPayload` return type it wants, and anything else is an error. For `nimbus-eth1`, this appears not to be true, as you've observed. These are simply different API needs.

tersec: One potential alternate approach though which might align things on the `nimbus-eth1` side: when building blocks, it can only build to the current timestamp, more or less, I believe. Maybe some slots ahead in theory -- haven't close-read that portion of the spec in a while to check if it's technically legal to ask an EL to build 10 slots ahead, and, if it is, probably someone's doing it and will complain if their use case is broken -- but for the most part, the EL too knows which return type it wants ahead of time also.

beacon_chain/eth1/eth1_monitor.nim
tersec: Also addressed by https://github.com/status-im/nimbus-eth2/pull/4696

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,254 +611,745 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+          randomData,
+          suggestedFeeRecipient,
+          withdrawals))
+
+      if response.payloadStatus.status != PayloadExecutionStatus.valid or
+         response.payloadId.isNone:
+        raise newException(CatchableError, "Head block is not a valid payload")
+
+      # Give the EL some time to assemble the block
+      await sleepAsync(chronos.milliseconds 500)
```
etan-status, line 797: `engine_api.getPayload` further down here may take another 1000 ms, so if we hit this path, the outer `await requestsCompleted or deadline` may timeout on spec compliant ELs.

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -445,254 +611,745 @@ func toVoteData(blk: Eth1Block): Eth1Data =
...
+  let
+    engineApiWithdrawals = toEngineWithdrawals withdrawals
+  let isFcUpToDate = m.nextExpectedPayloadParams.areSameAs(
+    headBlock, safeBlock, finalizedBlock, timestamp,
+    randomData, suggestedFeeRecipient, engineApiWithdrawals)
+
+  let
+    timeout = when PayloadType is deneb.ExecutionPayloadForSigning:
+      # TODO We should follow the spec and track the timeouts of
+      #      the individual engine API calls inside `getPayloadFromSingleEL`.
```
etan-status, line 876: This should be done even in the pre-Deneb case. Currently, the 1s timeout includes all of `getPayloadFromSingleEL` and doesn't just wrap the `getPayload` message.

Furthermore, we should probably start `deadline` only after all requests have been sent, so that all ELs have the full 1s available for `getPayload`.

etan-status (commented):
Missed a block with `Failed to obtain execution payload from EL`, despite Geth logging `Stopping work on payload` only 0.679s after `Starting work on payload`.

But, Geth also logged `Chain head was updated` 0.858s before finishing the payload, which is probably the `await rpcClient.forkchoiceUpdated` within `getPayloadFromSingleEL`. 

The 1s timeout is probably too aggressive with the current implementation, as it spans the entire block building process and not just the actual `getPayload` call.

@zah

etan-status:
Also, @tersec, when using an external VC, we don't seem to eagerly trigger the `runProposalForkchoiceUpdated` even though `nextProposalSlot` is known by `ActionTracker`. `checkNextProposer` seems to only take into account the local DAG attached validators. Is this intended, or an oversight?

tersec:
> Also, @tersec, when using an external VC, we don't seem to eagerly trigger the `runProposalForkchoiceUpdated` even though `nextProposalSlot` is known by `ActionTracker`. `checkNextProposer` seems to only take into account the local DAG attached validators. Is this intended, or an oversight?

Well, it's not intended. But it seems like it at least should be set up to work, so would have to look at why it might not.

https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/validators/action_tracker.nim#L236-L239

https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/consensus_object_pools/consensus_manager.nim#L289-L299

https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/consensus_object_pools/consensus_manager.nim#L342-L349

https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/gossip_processing/block_processor.nim#L630-L631

So this should only be failing to handle VC validators if `getNextValidatorAction` isn't picking them up either, and it's the same core function used by other parts of Nimbus for this purpose. `runProposalForkchoiceUpdated` isn't particularly unique in this regard.

etan-status:
That's correct, so probably something else going on :-/ It's indeed calling the same source as is used in the log

etan-status:
@tersec Found the culprit:

```nim
  # In Capella and later, computing correct withdrawals would mean creating a
  # proposal state. Instead, only do that at proposal time.
  if nextWallSlot.is_epoch:
    debug "runProposalForkchoiceUpdated: not running early fcU for epoch-aligned proposal slot",
      nextWallSlot, validatorIndex, nextProposer
    return
```

Actually it has to be this one, as I don't see any fcU sent to Geth in the slot prior to my missed block.

Which is consistent with the behaviour when `self.consensusManager.checkNextProposer(wallSlot).isSome`, and then `runProposalForkchoiceUpdated` returns. In that situation, no fcU is sent at all, until we try to actually build the next block.

Should we simply fall back to a regular fcu in that case, so that at least the attestations are correct?

tersec:
> @tersec Found the culprit:
> 
> ```nim
>   # In Capella and later, computing correct withdrawals would mean creating a
>   # proposal state. Instead, only do that at proposal time.
>   if nextWallSlot.is_epoch:
>     debug "runProposalForkchoiceUpdated: not running early fcU for epoch-aligned proposal slot",
>       nextWallSlot, validatorIndex, nextProposer
>     return
> ```
> 
> Actually it has to be this one, as I don't see any fcU sent to Geth in the slot prior to my missed block.
> 
> Which is consistent with the behaviour when `self.consensusManager.checkNextProposer(wallSlot).isSome`, and then `runProposalForkchoiceUpdated` returns. In that situation, no fcU is sent at all, until we try to actually build the next block.
> 
> Should we simply fall back to a regular fcu in that case, so that at least the attestations are correct?

https://github.com/status-im/nimbus-eth2/pull/5195

etan-status:
Another problem was that `Slot end` is triggered immediately after `Slot start` if no in-BN validators have duties. This leads to possibly heavy processing being done early into the slot, even when VCs are processing duties.

I added an additional delay to the slot processing, to wait until shortly before end of slot before calling `onSlotEnd` if there is a potential for pending duties: https://github.com/status-im/nimbus-eth2/pull/5196

PR #4470, opened 2023-01-06
Bump libp2p

Menduist:
This bump includes:
- https://github.com/status-im/nim-libp2p/pull/800 &
- https://github.com/status-im/nim-libp2p/pull/808: gossipsub was capped to 4mb
- https://github.com/status-im/nim-libp2p/pull/811: ddos protection, now each peer can only open 10 concurrent streams per protocol. **Please check that this is enough for nimbus protocols**, otherwise that's overidable per-protocol
- https://github.com/status-im/nim-libp2p/pull/816: Fix cancellation with mplex, thanks @etan-status!


arnetheduck:
> Please check that this is enough for nimbus protocols

I think this should be fine - a peer should really not be doing more concurrency than .. 2:ish per request type so 10 is ample maneuvering space - if anything, we might experiment with lowering it with the appropriate experimentation - I do also wonder if this should lead to a scoring event of some sort - ie we might need to add a notification mechanism so the application can track violations.

arnetheduck (approved):

PR #4480, opened 2023-01-09
stricter check for context bytes when recv LC data

etan-status:
Bellatrix and Altair light client data share same body, but have other fork digests. Validate that the peer's sent fork digest matches the one that we expect (derived from `attested_header.slot`).


arnetheduck:
I'll assume you carefully thought about the off-by-one of the slots involved here? In sync committee messages, this is a mess...

PR #4482, opened 2023-01-10
Wire up engine_newPayloadV3

henridf:
Currently as draft as the nim-web3 submodule is pointing to https://github.com/henridf/nim-web3/tree/eip4844; I will update and un-draft once https://github.com/status-im/nim-web3/pull/74 is merged.


tersec:
Merged https://github.com/status-im/nim-web3/pull/74

tersec:
Missing tests, along the lines of https://github.com/status-im/nimbus-eth2/blob/unstable/tests/test_eth1_monitor.nim

Basically some sanity checks that bidirectional conversions of `asConsensusExecutionPayload` and `asEngineExecutionPayload` convert back and forth without internal inconsistencies or dropping fields.

I've been making these by generating random values -- a first step, and probably enough to immediate merging, is to copy and paste the capella tests to eip4844 in that test module, and leaving the additional field empty. If one wants, as a separate PR then, one can add a more realistic distribution (I've tended to bias towards zero and max/boundary value things somewhat for testing purposes, and otherwise using uniform random, in a way which applies to both minimal and mainnet presets where list lengths differ between them) of values recursively through each value.

tersec:
It'd be obviously possible to do these more randomly, but we tend to avoid running PRNGs as part of the testing process itself when not necessary, to keep things consistent. But also, these serve as kind of API/function documentation, so it's fine for test cost to be relatively verbose and generated code in this way.

henridf:
@tersec thanks for the pointer to this test. Add eip4844 coverage in 90b4091109fe2d2b2968d85bf56965f692d9a600. For the new `excess_data_gas` field, I just re-used the  value  from `base_fee_per_gas`; if there's a reason to use a separate random value let me know and I can adjust.

tersec:
Reusing the random value is fine -- there's no reason to think there'd be any particular failure correlation between them to be concerned about. They aren't functions of each other.

tests/test_eth1_monitor.nim
```diff
@@ -1021,3 +1022,547 @@ suite "Eth1 monitor":
       check:
         executionPayload == asConsensusExecutionPayload(
           asEngineExecutionPayload(executionPayload))
+
+  test "Roundtrip engine RPC V2 and eip4844 ExecutionPayload representations":
```
tersec:
```suggestion
  test "Roundtrip engine RPC V3 and eip4844 ExecutionPayload representations":
```

tersec:
The CI will reject this as-is, because you haven't run `make test` locally then `git commit`ed the results. It adds a new test item, and the CI will notice that the test summaries aren't consistent with PR.

henridf:
> you haven't run make test locally then git commited the results

oops, fixed.

tersec (approved):

PR #4487, opened 2023-01-10
compile-time flag to test SSZ `txs_root` / `withdrawals_root`

etan-status:
There is a current discussion to transition certain EL block header parts to match their CL counterparts.

This includes:
1. Transitioning EL `Withdrawal` object to match CL `Withdrawal` object, i.e., use SSZ and Gwei amounts.
2. Transitioning EL `withdrawals_root` from hexary root to SSZ root.
3. Transitioning EL `txs_root` from hexary root to SSZ root. EL `Transaction` object is still exchanged in RLP format.

To support such tests, the engine API is updated for compatibility.

Use this command to build a compatible Nimbus beacon node:
```sh
make -j nimbus_beacon_node DEBUG_CAPELLA_USE_SSZ=1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,10 @@ ifeq ($(USE_LIBBACKTRACE), 0)
 NIM_PARAMS += -d:disable_libbacktrace
 endif
 
+ifeq ($(DEBUG_CAPELLA_USE_SSZ), 1)
+NIM_PARAMS += -d:DEBUG_CAPELLA_USE_SSZ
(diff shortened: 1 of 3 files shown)


tersec:
Would prefer this stay out of `unstable`, because it creates a maze of invasive `when`/`else`s.

etan-status:
Superseded by https://github.com/status-im/nimbus-eth2/pull/4509

PR #4491, opened 2023-01-11
Refactor block/blobs types

henridf:
Use type system to enforce invariant that a pre-4844 block cannot have a sidecar.

This is a first step along the lines of what we discussed in https://github.com/status-im/nimbus-eth2/pull/4454#issuecomment-1375505802. This could go be extended further by changing more/all places that take a `Opt[blob]` to drop the `Opt` and take a `ForkySignedBeaconBlockMaybeBlobs` instead (better/shorter name suggestions welcome). 

@tersec what do you think?


tersec:
If you update this to current `unstable`, it should make the unrelated GitHub Actions `windows-amd64` CI failure (https://github.com/status-im/nimbus-eth2/actions/runs/3894352819/jobs/6648236396) less likely:
```
2023-01-11T18:08:25.2828522Z gcc.exe -c  -w -fmax-errors=3 -mno-ms-bitfields -DWIN32_LEAN_AND_MEAN -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-libbacktrace -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-libbacktrace/install/usr/include -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-secp256k1/secp256k1/../secp256k1_wrapper -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-secp256k1/secp256k1/../secp256k1_wrapper/secp256k1 -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-secp256k1/secp256k1/../secp256k1_wrapper/secp256k1/src -DHAVE_CONFIG_H -DUSE_ASM_X86_64 -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-bearssl/bearssl/abi/../csources/src/ -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-bearssl/bearssl/abi/../csources/inc/ -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-bearssl/bearssl/abi/../csources/tools/ -DBR_USE_WIN32_TIME=1 -DBR_USE_WIN32_RAND=1 -DBR_LE_UNALIGNED=1 -DBR_64=1  -DBR_amd64=1 -DBR_INT128=1 -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nim-zlib/zlib/csources -DHAVE_UNISTD_H -march=native -mno-avx512f -fno-omit-frame-pointer -g3 -Og -gdwarf-3 -O3 -fno-strict-aliasing -fno-ident -fno-lto   -ID:/a/nimbus-eth2/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib -ID:/a/nimbus-eth2/nimbus-eth2/research -o D:/a/nimbus-eth2/nimbus-eth2/nimcache/release/state_sim/sqlite3.c.o D:/a/nimbus-eth2/nimbus-eth2/vendor/nim-sqlite3-abi/sqlite3.c
2023-01-11T18:08:40.9149973Z 
2023-01-11T18:08:40.9152404Z cc1.exe: out of memory allocating 82368 bytes
2023-01-11T18:08:40.9579338Z mingw32-make[1]: *** [nimcache/release/state_sim/state_sim.makefile:911: D:/a/nimbus-eth2/nimbus-eth2/nimcache/release/state_sim/sqlite3.c.o] Error 1
2023-01-11T18:08:41.0566196Z mingw32-make: *** [Makefile:308: state_sim] Error 2
2023-01-11T18:08:41.0566815Z mingw32-make: *** Waiting for unfinished jobs....
```
via https://github.com/status-im/nimbus-eth2/pull/4495

henridf:
> If you update this to current unstable

done

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1460,11 +1460,11 @@ proc installMessageValidators(node: BeaconNode) =
           MsgSource.gossip, signedBlock)))
 
   if node.dag.cfg.EIP4844_FORK_EPOCH != FAR_FUTURE_EPOCH:  
-    node.network.addValidator(
+   node.network.addValidator(
```
```suggestion
    node.network.addValidator(
```

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -802,7 +802,17 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
 
         withBlck(forked):
           blck.root = hash_tree_root(blck.message)
-          await node.router.routeSignedBeaconBlock(blck)
+          let signedBlockAndBlobs =
+            when blck is eip4844.SignedBeaconBlock:
+              # TODO: Fetch blobs from EE
+              eip4844.SignedBeaconBlockAndBlobsSidecar(
```
tersec, line 808: Similar post-EIP4844 issue as in the other PR -- this will fall back to just `blck`, without the sidecar, unless someone notices specifically by auditing all the usages of `eip4844.SignedBeaconBlock` or otherwise. 

The exact fallout of this depends on what uses which types.

https://github.com/henridf/nimbus-eth2/blob/6feb0bb34fc3b28898826c13dcc07971c683a4d9/beacon_chain/validators/message_router.nim#L85-L87
```nim
proc routeSignedBeaconBlock*(
    router: ref MessageRouter, blckAndBlobs: ForkySignedBeaconBlockMaybeBlobs):
    Future[RouteBlockResult] {.async.} =
```

Currently this PR has
```nim
  ForkySignedBeaconBlockMaybeBlobs* =
    phase0.SignedBeaconBlock |
    altair.SignedBeaconBlock |
    bellatrix.SignedBeaconBlock |
    capella.SignedBeaconBlock |
    eip4844.SignedBeaconBlockAndBlobsSidecar
```
which means that it will result in a compile-time error, which is good -- it'll alert people that this is maybe not the intended use of that function.

However, the `ForkySignedBeaconBlockMaybeBlobs` type looks potentially usable as a more general-purpose function.

Maybe making it clearer, via its name or some in-line comments/documentation, that it's meant for gossip-routable and similar blocks, and it would be incorrect to add plain `eip4844.SignedBeaconBlock` (or, probably, future, forks' `SignedBeaconBlock`s) to it.

Then, one can safely use this `when foo is eip4844.SignedBeaconBlock` construction safely.

henridf: > Maybe making it clearer, via its name or some in-line comments/documentation

Will add a comment. Re name, I've had a hard time coming up with a more descriptive (not to mention shorter) name. Will noodle on it some more.

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -875,7 +885,16 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
 
       let res = withBlck(forked):
         blck.root = hash_tree_root(blck.message)
-        await node.router.routeSignedBeaconBlock(blck)
+        let signedBlockAndBlobs =
+          when blck is eip4844.SignedBeaconBlock:
```
tersec, line 889: One abstraction we've used for this, to avoid it landing in modules which architecturally shouldn't care that much about this level of details, is the `init(object of type from object of other type)` construction:
https://github.com/status-im/nimbus-eth2/blob/2324136552d2788b5e535bbd22b93c2da1268718/beacon_chain/spec/forks.nim#L384-L405

Those examples tend to construct new objects, but it should be fine to have an identity-function `init` (or could consider other nomenclatures, but that's a reasonable one, and the closest `forks.nim` currently has to this use case -- it's usually used to constructed a `Forked` from `Forky` object) which uses a `template`, so in that degenerate case it's guaranteed to becomes a no-op at runtime.

The trouble is there's no compile-time equivalent to `case`, as there is for `if` (which has `when`). So it's otherwise a bit clumsy. That approach uses template/function overloading to achieve a similar goal of compile-time-checked safety, while centralizing these type games in one place, rather than spread them around the codebase.

beacon_chain/rpc/rest_beacon_api.nim
henridf: > However, the ForkySignedBeaconBlockMaybeBlobs type looks potentially usable as a more general-purpose function.

I don't think I follow - it's a type, but could be a function - maybe you could sketch the signature of the function you have in mind?

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -327,6 +377,11 @@ proc validateBeaconBlock*(
         # validation.
         return errReject("BeaconBlock: rejected, parent from unviable fork")
 
+    let blobs =
+      when signed_beacon_block is eip4844.SignedBeaconBlockAndBlobsSidecar:
+        Opt.some(signed_beacon_block.blobs_sidecar)
+      else:
+        Opt.none(eip4844.BlobsSidecar)
```
tersec, line 384: This is acting as an interface to the `Opt`-world in the block quarantine:
https://github.com/status-im/nimbus-eth2/blob/2324136552d2788b5e535bbd22b93c2da1268718/beacon_chain/consensus_object_pools/block_quarantine.nim#L45-L50
It is reasonable to scope this PR not to change that. It leaves the block quarantine codepaths with the same issue this fixes in the block validation codepaths, but still an improvement, and addressing the block quarantine would involve a whole new set of types. Useful to partition the changes required.

beacon_chain/rpc/rest_beacon_api.nim
tersec: Oops, you're right, mis-typed. Not "function", but "type".

Yeah, a more general-purpose type. That is, per above, as already discussed, because it doesn't specify that it's specifically crafted around the requirements of block gossip, it's potentially tempting to reuse or extend that type for other purposes in a way which creates potential bugs for its original purpose.

henridf: > Similar post-EIP4844 issue as in the other PR -- this will fall back to just blck, without the sidecar, unless someone notices specifically by auditing all the usages of eip4844.SignedBeaconBlock or otherwise.

I tried addressing this with the `init*` constructor approach you outlined below but... the fundamental issue is still that we do have to explicitly check here (either literally here, at L806, or in some nearby function call) if the incoming block is `eip4844.SignedBeaconBlock`, in order to know that we need to fetch the corresponding blobs. To call the `init*` templates below, we'd need to have already determined if this is a eip4844 block.

So I guess the solution here (at least for the specific issue of avoiding silent future failure) would be   `static: doAssert high(BeaconStateFork) == BeaconStateFork.EIP4844` as you indicated in the other PR. 

Unless I'm missing something?


```nim
template init*(blck: phase0.SignedBeaconBlock): ForkySignedBeaconBlockMaybeBlobs =
  blck
template init*(blck: altair.SignedBeaconBlock): ForkySignedBeaconBlockMaybeBlobs =
  blck
template init*(blck: bellatrix.SignedBeaconBlock): ForkySignedBeaconBlockMaybeBlobs =
  blck
template init*(blck: capella.SignedBeaconBlock): ForkySignedBeaconBlockMaybeBlobs =
  blck
template init*(blck: eip4844.SignedBeaconBlock, blobs: eip4844.BlobsSidecar): ForkySignedBeaconBlockMaybeBlobs =
  eip4844.SignedBeaconBlockAndBlobsSidecar(
    beacon_block: blck,
    blobs_sidecar: blobs
  )
```

tersec: Sure, then I agree that `static: doAssert high(BeaconStateFork) == BeaconStateFork.EIP4844` is the best available approach.

henridf: Actually - the `high(BeaconStateFork) == BeaconStateFork.EIP4844` pattern is for enums, not generic types. 

And returning to the root issue:

> Similar post-EIP4844 issue as in the other PR -- this will fall back to just blck, without the sidecar

I think this isn't an issue here? If a new `SignedBeaconBlock` type is added, the `ForkySignedBeaconBlockMaybeBlobs` generic isn't extended to included it, that should be result compilation error at any place (for example, a function call) where that new `SignedBeaconBlock` type is used in lieu of a `ForkySignedBeaconBlockMaybeBlobs`.  Conversely, if I remove one of the types from `ForkySignedBeaconBlockMaybeBlobs`, I get a "type mismatch" compilation error.

beacon_chain/rpc/rest_beacon_api.nim
henridf: same issue at https://github.com/status-im/nimbus-eth2/pull/4491#pullrequestreview-1248064440, conversation there

beacon_chain/gossip_processing/gossip_validation.nim
henridf: Yes, agreed.

beacon_chain/rpc/rest_beacon_api.nim
tersec: The `high(BeaconStateFork) == BeaconStateFork.EIP4844` pattern doesn't really depend on what it's guarding -- it's a generic compile-time mechanism to flag that a particular section of code depends in some way not otherwise found in a compile-time or maybe runtime way on the particular forks known. It can be used just as readily for code based on generic types as enums, though typically when there are generic types, there are better approaches. For most specific cases, indeed, there are better methods, but it's a fallback approach.

And, yes, what you note, that if "a new `SignedBeaconBlock type is added, the ForkySignedBeaconBlockMaybeBlobs generic isn't extended to included it" is true by default, if people remain aware that it's supposed to match the gossip types, not some other potentially interesting use of a type called `ForkySignedBeaconBlockMaybeBlobs`

To be a bit more explicit, the sequence of events I'm concerned about, arrayed over months/years, is:
1) this PR gets merged, without better in-line documentation (comments, naming of type, some `static: doAssert not default(eip4844.SignedBeaconBlock) is ForkySignedBeaconBlockMaybeBlobs` thing, etc)

2) someone else, unaware of the very specific purpose of the shape of `ForkySignedBeaconBlockMaybeBlobs`, either (unlikely) for EIP4844, or (more likely) more-or-less mechanically adding on to this type later, doesn't know that they shouldn't add the bare `SignedBeaconBlock` for EIP4844 or later, and only the with-blobs version. This will not trigger any immediate error or compilation warning -- it'll just be one more possible type that type class can match.

There's still no full bug here, but it's primed and one step away:

3) then, later yet, someone changes the routing or other functions which take a `ForkySignedBeaconBlockMaybeBlobs`, and now they also (accidentally, from step 2) take a bare `SignedBeaconBlock` type from an EIP48844-or-later fork, and, e.g., try to broadcast it. That's a bug. Currently, the type system will catch this category of bugs, but after this PR, it might not.

(2) and (3) are more likely to happen specifically in the next couple of hardforks after EIP4844 -- enough of them, and the pattern of only the with-blobs version would become obvious, but after the first one, it might look like just an oversight or omission.

It's not a current bug, and the code is sort of fine as it is in a narrow way, but it's a potential future maintenance hazard by weakening the type checking guarantees if someone without full understanding of a the reason/context for the code to via steps (2)/(3), maybe different people even for each, induce a bug.

henridf:
At a high-level my takeaway given all this discussion is that this thing isn't (yet?) structured properly. I guess the question is, is it an improvement over status quo (https://github.com/status-im/nimbus-eth2/pull/4454#issuecomment-1375505802) ? Doesn't seem like an obvious yes.

beacon_chain/rpc/rest_beacon_api.nim
henridf: > high(BeaconStateFork) == BeaconStateFork.EIP4844 [..] can be used just as readily for code based on generic types as enums

What's the equivalent of "high" for generic types? it must be staring at me in the eyes, but I don't see one.

> (comments, naming of type, some static: doAssert not default(eip4844.SignedBeaconBlock) is ForkySignedBeaconBlockMaybeBlobs thing, etc)

Added a comment and assert in c9232d0717abdf0935b79c3357f8b13aa45933bd. 

Re name, the best I'm coming up with is GossipSignedBeaconBlockMaybeBlobs. That's a tad more descriptive and emphasises the gossip-oriented motivation for this type, but on the other hand there's a strong discipline here (as commented at the top of forks.nim) that generic types for dealing with forks are `Forky*`, so I'm not sure why this particular case should deviate from that. Happy to change if you think it's worthwhile.

beacon_chain/spec/forks.nim
```diff
@@ -183,12 +183,18 @@ type
...
+  # gossip-originating blocks, which are eventually separated into the
+  # constituent parts before passing along into core functions.
   ForkySignedBeaconBlockMaybeBlobs* =
     phase0.SignedBeaconBlock |
     altair.SignedBeaconBlock |
     bellatrix.SignedBeaconBlock |
     capella.SignedBeaconBlock |
     eip4844.SignedBeaconBlockAndBlobsSidecar
+  # This type should only contain types that are gossiped.
+  static: doAssert not default(eip4844.SignedBeaconBlock) is ForkySignedBeaconBlockMaybeBlobs
```
tersec: This can't be put in a `type` section -- everything here is implicitly `type Foo`, e.g., `ForkySignedBeaconBlockMaybeBlobs* = ...`

tersec:
> At a high-level my takeaway given all this discussion is that this thing isn't (yet?) structured properly. I guess the question is, is it an improvement over status quo (#4454 (comment)) ? Doesn't seem like an obvious yes.

With your latest commit (fixed by moving it outside the `type` section so it's not trying to parse it as a type definition or declaration), I do view it as an improvement over the status quo, yeah.

beacon_chain/rpc/rest_beacon_api.nim
tersec: > Re name, the best I'm coming up with is GossipSignedBeaconBlockMaybeBlobs. That's a tad more descriptive and emphasises the gossip-oriented motivation for this type, but on the other hand there's a strong discipline here (as commented at the top of forks.nim) that generic types for dealing with forks are Forky*, so I'm not sure why this particular case should deviate from that. Happy to change if you think it's worthwhile.

Agree. With the other changes, the existing name is good.

tersec (approved):

beacon_chain/spec/forks.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4491/commits/c9232d0717abdf0935b79c3357f8b13aa45933bd addresses this

PR #4494, opened 2023-01-12
Working Makefile targets for Capella devnet2

zah:
make capella-devnet-2
make clean-capella-devnet-2

You'll need to have https://github.com/tmuxinator/tmuxinator installed. It's available as a regular package in most Linux distributions or through Nix or Brew on macOS.

This commit also fixes the initial hang in the Eth1 monitor in the "find TTD block" procedure through a fix to the network metadata files which hasn't been upstreamed yet.


tersec:
How does this fix the TTD detection spin loop?

zah:
It adds the hash of the deployment block of the deposit contract (which happens to be the genesis block in these devnets). Without this, even when `findTTDBlock` is removed, the monitor is not initialized properly and deposit monitoring doesn't actually start (which is certainly not fatal in these network, because post-genesis deposits are not expected).

tersec:
> It adds the hash of the deployment block of the deposit contract (which happens to be the genesis block in these devnets). Without this, even when `findTTDBlock` is removed, the monitor is not initialized properly and deposit monitoring doesn't actually start (which is certainly not fatal in these network, because post-genesis deposits are not expected).

Is this a valid deposit contract configuration?

zah:
Yes, in these devnets, the deployment of the deposit contract is effectively baked into the genesis execution state. Knowing the hash is important because you need it in order to initialize the EIP-4881-compatible deposit contract snapshot object.

arnetheduck:
> How does this fix the TTD detection spin loop?

A follow-up question is obviously why the other clients don't need this in their ttd detection loops - ie why do we need the extra data?

zah:
The requirement is related to EIP-4881 which is not yet supported in all clients. The only difference between the 4881 snapshots and our original snapshots is that the new snapshots feature both block height and block hash, so when we bootstrap the client we need to know both. Technically, it's possible to obtain one of these values from an EL client by supplying the other, but this was exactly the problem you highlighted in other PRs - it would make bootstrapping reliant on a synced EL client. Including the values in the metadata is a simple way to sidestep the problem and other clients are likely to discover this too at some point.

I've contributed an enhancement to the tooling that the EF uses for setting up the devnets, so all future devnets will have the full metadata that we need:

https://github.com/ethpandaops/ethereum-genesis-generator/pull/21

PR #4498, opened 2023-01-12
Fix eip4844 gossip transition

henridf:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -805,19 +805,26 @@ proc updateBlocksGossipStatus*(
     # Individual forks added / removed
     discard
 
+  template blocksTopic(fork: BeaconStateFork , forkDigest: ForkDigest): auto =
```
```suggestion
  template blocksTopic(fork: BeaconStateFork, forkDigest: ForkDigest): auto =
```

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -805,19 +805,26 @@ proc updateBlocksGossipStatus*(
...
+    if fork == BeaconStateFork.EIP4844:
```
tersec: Presumably post-EIP-4844 forks will also use more EIP4844-like topics than they will pre-EIP4844 topics. This logic will silently revert, post-EIP4844, to `getBeaconBlocksTopic` rather than `getBeaconBlockAndBlobsSidecarTopic`.

We've dealt with this a few ways:

1) if confident that the next fork will also use `getBeaconBlockAndBlobsSidecarTopic`, replace `fork == BeaconStateFork.EIP4844` with `fork >= BeaconStateFork.EIP4844`

2) if not confident, and want a runtime error but not an immediate compile-time error when updating the forks, use something like
```nim
if fork == BeaconStateFork.EIP4844:
  ...
elif fork < BeaconStateFork.EIP4844:
  ...
else:
  doAssert false, "Unknown fork in FOO"    # or similar -- can be helpful to show the fork, or etc. An informative assert message is useful because it allows immediate pinpointing of where to fix
```
where there are several viable variations here, but the general idea.

3) if one wants an immediate compile-time error, the actual `if` can be kept as-is, but next to it also something like
```nim
  static: doAssert high(BeaconStateFork) == BeaconStateFork.EIP4844
```
so the next person updating this will know immediately this is code which needs checking.

(4) use a `case` expression:
```nim
case fork
of BeaconStateFork.Phase0, BeaconStateFork.Altair, BeaconStateFork.Bellatrix, BeaconStateFork.Capella: ...
of BeaconStateFork.EIP4844: ...
```
The idea here is that Nim checks case expressions and statements to be exhaustive, so if there's a new `BeaconStateFork` enum value, any `case` expression on that not including said value, as you discovered in your first couple of PRs, will throw errors at compile time, automatically.

The choice between these is contextual -- I usually find (1) makes too many assumptions, and it's easy to change the `==` to an `>=` or other more inclusive condition upon seeing the next hardfork beyond EIP4844, but tracking down all the places which have this sort of logic is a factory for bugs.

So, typically, I'd prefer (2), (3), or (4).

(3) and (4) are cleaner, and easy to transform to (2) if, when the person doing the updates sees it, decides not to immediately just include the post-EIP4844 hardfork in the EIP4844 `blocksTopic` logic. It's also more obvious than (2) (which is more obvious than (1)).

So, among these, I tend to prefer (4): use a `case` expression. It doesn't include any extra boilerplate, really (and one can use range expressions if one wants, to elude `BeaconStateFork.Altair` and `BeaconStateFork.Bellatrix`, but that particular list will only ever be 4 forks long, so...).

henridf: Thanks for the feedback - agreed that compile-time check is preferable. Done in a595bb2.

tersec (approved):

PR #4493, opened 2023-01-11
make LC data fork aware

etan-status:
In a future fork, light client data will be extended with execution info to support more use cases. To anticipate such an upgrade, introduce `Forky` and `Forked` types, and ready the database schema. Because the mapping of sync committee periods to fork versions is not necessarily unique (fork schedule not in sync with period boundaries), an additional column is added to `period` -> `LightClientUpdate` table.

PR #4499, opened 2023-01-12
cache LC headers and sync committees for bootstrap

etan-status:
To allow LC data retention longer than the one for historic states, introduce persistent DB caches for `current_sync_committee` and `LightClientHeader` for finalized epoch boundary blocks. This way, historic `LightClientBootstrap` requests may still be honored even after pruning. Note that historic `LightClientUpdate` requests are already answered using fully persisted objects, so don't need changes.

Sync committees and headers are cached on finalization of new data. For existing data, info is lazily cached on first access.


etan-status:
REST endpoint for testing:

```
http://127.0.0.1:5052/eth/v1/beacon/light_client/bootstrap/0xBLOCKROOT
```

If accessed or imported before pruning, data will be cached and will remain available even after pruning.

beacon_chain/beacon_chain_db_light_client.nim
```diff
@@ -22,11 +22,23 @@ import
 
 logScope: topics = "lcdata"
 
+# `lc_xxxxx_headers` contains a copy of historic `LightClientHeader`.
```
arnetheduck, line 22: can we document the expected db growth from this here? ie bytes/epoch

beacon_chain/beacon_chain_db_light_client.nim
```diff
@@ -84,6 +115,68 @@ type
       ## Tracks the finalized sync committee periods for which complete data
       ## has been imported (from `dag.tail.slot`).
 
+proc initHeadersStore(
+    backend: SqStoreRef,
+    name, typeName: string): KvResult[LightClientHeaderStore] =
+  ? backend.exec("""
```
arnetheduck: why not a kvstore? it seems that lookups are made by hash?

beacon_chain/beacon_chain_db_light_client.nim
```diff
@@ -139,18 +232,18 @@ func hasCurrentSyncCommitteeBranch*(
   false
 
 proc getCurrentSyncCommitteeBranch*(
-    db: LightClientDataDB, slot: Slot): altair.CurrentSyncCommitteeBranch =
+    db: LightClientDataDB, slot: Slot): Opt[altair.CurrentSyncCommitteeBranch] =
   if not slot.isSupportedBySQLite:
-    return default(altair.CurrentSyncCommitteeBranch)
+    return err()
```
arnetheduck: `Opt.none` reads better for `Opt`, in general

beacon_chain/beacon_chain_db_light_client.nim
```diff
@@ -160,6 +253,81 @@ func putCurrentSyncCommitteeBranch*(
   let res = db.currentBranches.putStmt.exec((slot.int64, SSZ.encode(branch)))
   res.expect("SQL query OK")
 
+proc initSyncCommitteesStore(
+    backend: SqStoreRef,
+    name: string): KvResult[SyncCommitteeStore] =
+  ? backend.exec("""
+    CREATE TABLE IF NOT EXISTS `""" & name & """` (
```
arnetheduck: ditto - I wonder if we should make a sparse int-keyed kvstore that serves use cases like this - finalizedblocks is essentially the same.. technically, a `kvstore` with the integer converted to blob is pretty much the same..

beacon_chain/beacon_chain_db_light_client.nim
etan-status: The headers store here also needs a `slot` column to support pruning.
The access is by root though.

image

arnetheduck: soo .. if the slot is available on query, there's also a trick in the current kvstore: we put the slot in the "key" just before the hash - this allows the db to sort the entries by slot "more or less" which improves performance (because new entries get written at the tail of the btree) _and_ you can remove ranges via `find` (prefix query essentially)..

beacon_chain/beacon_chain_db_light_client.nim
etan-status: For this one, a generic int --> value kvstore would indeed work (the other one is 3 coumns).
Went with the same style as the surrounding tables here for consistency. 
Using kvstore for some but not others feels off.

beacon_chain/beacon_chain_db_light_client.nim
etan-status: in this table, the queries are always by root (`/eth/v1/beacon/light_client/bootstrap/0xBLOCKROOT`).
the only use for the slot is the pruning, which essentially deletes all entries with slot < a threshold.

the trick is neat, and would speed up the pruning at the cost of slower root query.
was also debating an index on slot, but given the size of this table (1 entry per epoch) seems rather overkill in a constantly pruning setup (and in one without pruning, it doesn't matter because the pruning query is never run).

beacon_chain/beacon_chain_db_light_client.nim
etan-status: Added bytes per period (as some data is stored per period, makes it easier to compare)

beacon_chain/beacon_chain_db_light_client.nim
etan-status: Converted them in this file

beacon_chain/beacon_chain_db_light_client.nim
etan-status: Month computations done as 28 periods

beacon_chain/beacon_chain_db_light_client.nim
arnetheduck: > cost of slower root query

the root query is not really affected, ie the sorting helps lookup also by ensuring that we mostly look in the same "region" of the tree and thus have better cache locality on "recent" data queries - that said, performance does indeed not matter too much in a table like this - I'm thinking about it mainly from a code surface and sqlite feature set area point of view..

etan-status: The use case here is that each light client needs to obtain a single bootstrap value during its lifetime.
After that, it only needs by-period indexed data, or RAM data.

Ah, yes, if there is such caching of recent data, that's indeed true (although, at the cost of slightly more expensive cache misses).

So, would you prefer if I combine the `block_root` and `slot` columns here? 
(i.e., use kvStore for this table, to share the logic for the other tables that also need by-slot access?)

Current one is a bit easier to browse with generic tools (but guess the user base who interact with the DB like that could also deal with combined column, especially as this is used that way for our main tables already).

arnetheduck: yeah I'm torn .. I can understand both world views as they both have advantages, but I'm getting to the point where I think we're ending up with the worst of both worlds: we don't get the semantic simplicity of  a dumb kvstore and we don't get to play freely with useful sqlite features.

Looking at it another way, this would be the final nail in the coffin for kvstore, I think - I think at that point we might get rid of it completely .. the other point is that @zah has been wanting to create "generic" persistent data structures backed by the db, like `DbSeq` etc, favoring this approach over "specific" tables for each kind of data.. going down the route of "each table is special" means we should perhaps invest more in the sqlite abi wrapper and indeed add some more convenience stuff to it because we'll be seeing many more ad-hoc queries - in @zah's world view, we create persistent seqs, deques etc that are generic in similar ways to the language and use those - but then we're back to where we don't have access to multiple columns, like this pr.

arnetheduck:
what happens if a user downgrades?

etan-status:
They will contribute to the old table once more, and when they upgrade again that data is imported.

They won't have access to the already-upgraded data while downgraded though (they can regenerate with `--light-client-data-import-mode=full`)

etan-status:
Actually, for this PR here, nothing will happen. historic headers and sync committees will simply be fetched from DAG once more (i.e., unavailable after pruning)

etan-status:
Summary of LC related DB schema updates:

https://github.com/status-im/nimbus-eth2/pull/4493
- Here, updates table is migrated to become fork aware. Old data is copied into the new one (but won't overwrite existing entries). Old data is then deleted. If user downgrades, they can re-import update data with `--light-client-data-import-mode=full` or `=on-demand`. Bootstrap data is unaffected and historic bootstrap don't need re-import (so the really slow import when starting from scratch is not needed). User data imported into the legacy table is once more moved into the new ones on upgrade.

https://github.com/status-im/nimbus-eth2/pull/4499 (this PR)
- Only new tables are introduced, and they serve as a cache. If cache is hit, use info from the new table. Otherwise, fallback to DAG and then put a copy into the cache tables. On downgrade, it is the same as before where it always goes through DAG. New table only provide a benefit when pruning.

https://github.com/status-im/nimbus-eth2/pull/4512
- Once again, user state is completely migrated to become fork aware. Same scheme, old data is copied into a new table and then deleted. If user downgrades, they need to specify a fresh `--trusted-block-root`. There is only a tiny amount of info in the tables here (latest finalized header).

arnetheduck:
> Old data is then deleted. 

We generally don't do this - instead, we strive to write both formats and read the new format preferentially, but also read the old one - then, once a HF is past, we remove the writing of the old format (and potentially remove the old data) - this is likely what should happen in #4493 as well - ie a downgrade should not lead to a degradation or user action requirement in general.

beacon_chain/beacon_chain_db_light_client.nim
arnetheduck: re pruning, a mostly-equivalent option (specially since there's no index) is to iterate and decode the items - the slot column duplicates existing data, afaict (and decoding is "mostly-equivalent" to issuing the slot query since the data is flat, uncompressed and likely colocated with the rest of the row)

etan-status:
OK, have updated #4512 to re-introduce legacy table writing (also with behaviour for #4493).

etan-status:
For #4499, there is no legacy table, so it should be alright as is (regarding downgrades)?

beacon_chain/beacon_chain_db_light_client.nim
etan-status: It would still mean returning _all_ rows, SSZ decoding each of them, figuring out if it is possible to delete data during an enumeration (or build a seq with pending deletions), and deleting individual rows instead of all at once.

We could also register SSZ support with sqlite and have it in-place decode the slot (depending on `kind`).

Overall, I think either we should try to make it straight forward, or when we try to be tricky, use the same trick globally. So, from comment history here, I think using a combined `slot || root` column for this table is the way to go. The table is not large enough to have it matter whether the slot is prepended to the root. And maybe having the slot in there could allow your pruning logic to more effectively ensure that data is in the cache before ultimately deleting it (if you have access to what blocks are epoch boundary blocks — I don't have that access at upgrade time)

Will change `headers` to `slot || root` here.

arnetheduck:
> For https://github.com/status-im/nimbus-eth2/pull/4499, there is no legacy table, so it should be alright as is (regarding downgrades)?

sure, as long as a downgrading user doesn't lose anything significant, we're fine..

this also points to another important strategic piece of advice: write the "current" version with "healing" built in (for example block sync tops up missing blocks etc) - then we can assume that the downgrading user will be automatically healed and we can be more liberal in removing stuff

etan-status:
Changed headers table to `slot || root` key, and added capella support

etan-status:
Updated DB growth numbers for Capella and EIP4844 as well.

arnetheduck (approved):

PR #4504, opened 2023-01-13
show number of cores and total physical memory on Windows GitHub Actions

tersec:
Followup to https://github.com/status-im/nimbus-eth2/pull/4495 and https://github.com/status-im/nimbus-eth2/pull/4502

4495 was too aggressive, but it remains possible that GitHub Actions is supplying varyingly configured CI VMs, some of which are identifiable/predictably more likely to result in OOMs and can be detected as such, while others might have fewer cores or more RAM. This PR aims to allow discovery of such.


tersec:
The two GitHub Actions `windows-amd64` CIs ran this, and it works fine -- this PR doesn't change anything else, so just merge.

PR #4506, opened 2023-01-14
LC data fork cleanup

etan-status:
Distinguish between those code locations that need to be updated on each light client data format change, and those others that should generally be fine, as long as a valid light client object is processed.

The former are tagged with static assert for `LightClientDataFork.high`.

The latter are changed to `lcDataFork > LightClientDataFork.None` to indicate that they depend only on presence of any valid object. Also bundled a few minor cleanups and fixes.

Also add `Forky` type for `LightClientStore` and minor fixes / cleanups.

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -128,22 +128,21 @@ proc readChunkPayload*(
     peer.network.forkDigests[].stateForkForDigest(contextBytes).valueOr:
       return neterr InvalidContextBytes
 
-  if contextFork >= BeaconStateFork.Altair:
-    const lcDataFork = LightClientDataFork.Altair
-    let res = await eth2_network.readChunkPayload(
-      conn, peer, MsgType.forky(lcDataFork))
-    if res.isOk:
-      if contextFork != peer.network.cfg.stateForkAtEpoch(res.get.contextEpoch):
-        return neterr InvalidContextBytes
-      var obj = ok MsgType(kind: lcDataFork)
-      template forkyObj: untyped = obj.get.forky(lcDataFork)
-      forkyObj = res.get
-      return obj
+  withLcDataFork(lcDataForkAtStateFork(contextFork)):
+    when lcDataFork > LightClientDataFork.None:
+      let res = await eth2_network.readChunkPayload(
+        conn, peer, MsgType.Forky(lcDataFork))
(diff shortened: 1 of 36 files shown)


beacon_chain/validators/validator_monitor.nim
```diff
@@ -8,6 +8,7 @@
 import
   std/tables,
   metrics, chronicles,
+  ../spec/datatypes/phase0,
```
tersec, line 11: Why would this need a new import, with no other changes?

beacon_chain/sync/sync_protocol.nim
```diff
@@ -128,22 +128,21 @@ proc readChunkPayload*(
...
-      conn, peer, MsgType.forky(lcDataFork))
-    if res.isOk:
-      if contextFork != peer.network.cfg.stateForkAtEpoch(res.get.contextEpoch):
-        return neterr InvalidContextBytes
-      var obj = ok MsgType(kind: lcDataFork)
-      template forkyObj: untyped = obj.get.forky(lcDataFork)
-      forkyObj = res.get
-      return obj
+  withLcDataFork(lcDataForkAtStateFork(contextFork)):
+    when lcDataFork > LightClientDataFork.None:
```
tersec, line 132: Why switch from `contextFork >= BeaconStateFork.Altair` to `lcDataFork > LightClientDataFork.None`?

beacon_chain/validators/validator_monitor.nim
etan-status: nim 1.2 compat, there is a mention to `phase0.BeaconState` in this file. not sure why it worked before, but it is a common issue with nim 1.2 that if types are explicitly referred to by their module, that it also needs an explicit import

beacon_chain/sync/sync_protocol.nim
etan-status: Either works here. Switching on `lcDataForkAtStateFork(contextFork)` avoids having to manually maintain this each time that a new `LightClientDataFork` is introduced, by automatically deriving the `const lcDataFork` in the old code.

etan-status: as in, the old code requires re-visiting this location and update it, for example, if `LightClientDataFork.Capella` is introduced. the new code does not require maintenance like that.

tersec (approved):

PR #4512, opened 2023-01-15
use `ForkedLightClientStore` internally

etan-status:
When running `nimbus_light_client`, we persist the latest header from `LightClientStore.finalized_header` in a database across restarts. Because the data format is derived from the latest `LightClientStore`, this could lead to data being persisted in pre-release formats.

To enable us to test later `LightClientStore` versions on devnets, transition to a `ForkedLightClientStore` internally that is only migrated to newer forks on-demand (instead of starting at latest).

### DIFF
--- a/beacon_chain/gossip_processing/light_client_processor.nim
+++ b/beacon_chain/gossip_processing/light_client_processor.nim
@@ -29,8 +29,6 @@ logScope: topics = "gossip_lc"
 declareHistogram light_client_store_object_duration_seconds,
   "storeObject() duration", buckets = [0.25, 0.5, 1, 2, 4, 8, Inf]
 
-template storeDataFork: LightClientDataFork = LightClientDataFork.high
-
 type
   Nothing = object
 
@@ -101,7 +99,7 @@ type
 
     # Consumer
     # ----------------------------------------------------------------
-    store: ref Option[storeDataFork.LightClientStore]
+    store: ref ForkedLightClientStore
     getBeaconTime: GetBeaconTimeFn
     getTrustedBlockRoot: GetTrustedBlockRootCallback
     onStoreInitialized, onFinalizedHeader, onOptimisticHeader: VoidCallback
@@ -119,7 +117,7 @@ type
     of LightClientFinalizationMode.Optimistic:
       lastProgressTick: BeaconTime # Moment when last update made progress
       lastDuplicateTick: BeaconTime # Moment when last duplicate update received
-      numDuplicatesSinceProgress: int # Number of duplicates since last progress
+      numDupsSinceProgress: int # Number of duplicates since last progress
 
     latestFinalityUpdate: ForkedLightClientOptimisticUpdate
 
@@ -129,9 +127,6 @@ const
   duplicateCountDelay = chronos.minutes(15) # Delay to start counting duplicates
   minForceUpdateDelay = chronos.minutes(30) # Minimum delay until forced-update
   minForceUpdateDuplicates = 100 # Minimum duplicates until forced-update
-
-func storeDataFork*(x: typedesc[LightClientProcessor]): LightClientDataFork =
-  storeDataFork
 
 # Initialization
 # ------------------------------------------------------------------------------
@@ -143,7 +138,7 @@ proc new*(
     cfg: RuntimeConfig,
     genesis_validators_root: Eth2Digest,
     finalizationMode: LightClientFinalizationMode,
-    store: ref Option[storeDataFork.LightClientStore],
+    store: ref ForkedLightClientStore,
     getBeaconTime: GetBeaconTimeFn,
     getTrustedBlockRoot: GetTrustedBlockRootCallback,
     onStoreInitialized: VoidCallback = nil,
@@ -198,23 +193,22 @@ proc tryForceUpdate(
     self: var LightClientProcessor,
     wallTime: BeaconTime) =
   ## Try to force-update to the next sync committee period.
-  let
-    wallSlot = wallTime.slotOrZero()
-    store = self.store
+  let wallSlot = wallTime.slotOrZero()
+  doAssert self.finalizationMode == LightClientFinalizationMode.Optimistic
 
-  if store[].isSome:
-    doAssert self.finalizationMode == LightClientFinalizationMode.Optimistic
-    case store[].get.process_light_client_store_force_update(wallSlot)
-    of NoUpdate:
-      discard
-    of DidUpdateWithoutSupermajority:
-      warn "Light client force-updated without supermajority",
-        finalizedSlot = store[].get.finalized_header.beacon.slot,
-        optimisticSlot = store[].get.optimistic_header.beacon.slot
-    of DidUpdateWithoutFinality:
-      warn "Light client force-updated without finality proof",
-        finalizedSlot = store[].get.finalized_header.beacon.slot,
-        optimisticSlot = store[].get.optimistic_header.beacon.slot
+  withForkyStore(self.store[]):
+    when lcDataFork > LightClientDataFork.None:
+      case forkyStore.process_light_client_store_force_update(wallSlot)
+      of NoUpdate:
+        discard
+      of DidUpdateWithoutSupermajority:
+        warn "Light client force-updated without supermajority",
+          finalizedSlot = forkyStore.finalized_header.beacon.slot,
+          optimisticSlot = forkyStore.optimistic_header.beacon.slot
+      of DidUpdateWithoutFinality:
+        warn "Light client force-updated without finality proof",
+          finalizedSlot = forkyStore.finalized_header.beacon.slot,
+          optimisticSlot = forkyStore.optimistic_header.beacon.slot
 
 proc processObject(
     self: var LightClientProcessor,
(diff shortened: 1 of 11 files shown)


beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -198,61 +193,62 @@ proc tryForceUpdate(
     self: var LightClientProcessor,
     wallTime: BeaconTime) =
   ## Try to force-update to the next sync committee period.
-  let
-    wallSlot = wallTime.slotOrZero()
-    store = self.store
+  let wallSlot = wallTime.slotOrZero()
+  doAssert self.finalizationMode == LightClientFinalizationMode.Optimistic
```
etan-status, line 197: this one should be alright to move out even in the `None` case. it is a guard so that this function can only be called in optimistic mode. strict mode means no heuristics-based force-update

beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -198,61 +193,62 @@ proc tryForceUpdate(
...
+                self.store[].forky(lcDataFork) = initRes.get
+                ok()
+        elif forkyObject is SomeForkyLightClientUpdate:
+          if self.store[].kind == LightClientDataFork.None:
+            err(VerifierError.MissingParent)
+          else:
+            if lcDataFork > self.store[].kind:
+              info "Upgrading light client",
+                oldFork = self.store[].kind, newFork = lcDataFork
+              self.store[].migrateToDataFork(lcDataFork)
```
etan-status, line 244: This is the new demand-based upgrade logic. And also determines what ends up being written to `light_client_db.nim`, as the `self.store[].finalizedHeader` that is written shares `LightClientDataFork` of `self.store[]`.

beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -263,89 +259,102 @@ proc processObject(
...
+          default(ForkedLightClientHeader)
 
     body
 
     var
       didProgress = false
       didSignificantProgress = false
 
-    if store[].isSome != previousWasInitialized:
+    if self.store[].kind > oldKind:
```
etan-status: this check is a bit relaxed now and will also trigger callbacks on `LightClientDataFork` changes. previously latest format was always used so that transition never happened.

etan-status: Actually, probably safer to not report those. the migration to later fork could also be triggered by data that is ultimately rejected

etan-status: (as in, `didProgress` determines what ends up being pushed into fluffy portal network)

beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -401,29 +408,35 @@ proc storeObject*(
...
+        forkyStore = lcDataFork.LightClientStore(
+          finalized_header: forkyHeader,
+          current_sync_committee: current_sync_committee,
+          optimistic_header: forkyHeader)
+        debug "LC reset to finalized header",
+          finalizedSlot = forkyStore.finalized_header.beacon.slot,
+          optimisticSlot = forkyStore.optimistic_header.beacon.slot
+      else:
+        self.store[].reset()
+        debug "LC reset"
```
etan-status, line 440: this is a new possibility, as before resetting to `none` was not possible by signature. changed the reporting further up here to use `newIsInitialized > oldIsInitialized` to handle the case of initialized going from 1 to 0.

beacon_chain/light_client.nim
```diff
@@ -288,6 +291,9 @@ proc installMessageValidators*(
     msg.logReceived()
 
     if contextFork != lightClient.cfg.stateForkAtEpoch(msg.contextEpoch):
+      info "Received obj on wrong fork",
+        contextFork, expectedFork = lightClient.cfg.stateForkAtEpoch(msg.contextEpoch),
+        msg
```
etan-status: oh, a debug artifact. for some reason, `contextFork` was `EIP4844`.
worked around by capturing `let contextFork = stateFork` further down (and commented nim 1.6 version).
not sure if compiler bug, or why the loop variable itself gets captured instead of the current value.

removing the extra log here

beacon_chain/light_client_db.nim
```diff
@@ -46,35 +47,51 @@ type
...
+    info "Importing Altair light client data"
+    # LightClientHeaderKey -> altair.LightClientHeader
+    const legacyKind = Base10.toString(ord(LightClientDataFork.Altair).uint)
+    ? backend.exec("""
+      INSERT OR IGNORE INTO `""" & name & """` (
+        `key`, `kind`, `header`
+      )
+      SELECT `kind` AS `key`, """ & legacyKind & """ AS `kind`, `header`
+      FROM `""" & legacyAltairName & """`;
+    """)
```
etan-status, line 105: tested by starting a fresh db on `unstable` and then running the branch version

beacon_chain/light_client_db.nim
```diff
@@ -85,29 +102,41 @@ func close(store: LightClientHeadersStore) =
...
+      block:
+        const key = LightClientHeaderKey.Finalized
+        let res = db.headers.putStmt.exec(
+          (key.int64, lcDataFork.int64, SSZ.encode(forkyHeader)))
+        res.expect("SQL query OK")
+      block:
+        let period = forkyHeader.beacon.slot.sync_committee_period
+        doAssert period.isSupportedBySQLite
+        let res = db.syncCommittees.keepFromStmt.exec(period.int64)
+        res.expect("SQL query OK")
```
etan-status, line 168: wonder if the `None` case should also be handled here, but is not needed for now. could be used to reset the db when needed. in practice, only non-none is expected to be passed here

etan-status: put an assert for now

beacon_chain/nimbus_light_client.nim
```diff
@@ -52,9 +52,9 @@ programMain:
     quit 1
   let backend = SqStoreRef.init(dbDir, "nlc").expect("Database OK")
   defer: backend.close()
-  static: doAssert LightClientDataFork.high == LightClientDataFork.Altair
```
etan-status, line 55: this one no longer needed as the same table can be reused across new forks, so this section doesn't need continuous maintenance

tests/test_light_client.nim
```diff
@@ -9,230 +9,246 @@
...
+  const  # Test config, should be long enough to cover interesting transitions
+    headPeriod = 3.SyncCommitteePeriod
+  let
+    cfg = block:  # Fork schedule so that each `LightClientDataFork` is covered
+      static: doAssert BeaconStateFork.high == BeaconStateFork.EIP4844
+      var res = defaultRuntimeConfig
+      res.ALTAIR_FORK_EPOCH = 1.Epoch
+      res.BELLATRIX_FORK_EPOCH = 2.Epoch
+      # $capellaImplementationMissing res.CAPELLA_FORK_EPOCH = (EPOCHS_PER_SYNC_COMMITTEE_PERIOD * 1).Epoch
+      # $eip4844ImplementationMissing res.EIP4844_FORK_EPOCH = (EPOCHS_PER_SYNC_COMMITTEE_PERIOD * 2).Epoch
```
etan-status, line 30: these currently fail with `Attempt to use Bellatrix payload with post-Bellatrix state` when set to non-`FAR_FUTURE_EPOCH`.

there is another test suite to check store upgrades, so not a blocker.
Attached `$capellaImplementationMissing` comment so that it shows up in search for that symbol
cannot assert here because then the rest of the test would fail

tests/test_light_client.nim
```diff
@@ -9,230 +9,246 @@
...
     [block_clearance, block_quarantine, blockchain_dag],
   ../beacon_chain/spec/[forks, helpers, light_client_sync, state_transition],
   # Test utilities
   ./testutil, ./testdbutil
 
-proc runTest(storeDataFork: static LightClientDataFork) =
-  suite "Light client - " & $storeDataFork & preset():
+suite "Light client " & preset():
+  const  # Test config, should be long enough to cover interesting transitions
+    headPeriod = 3.SyncCommitteePeriod
```
etan-status, line 22: put an extra period here compared to previous code to have a long enough test to also do eip4844 format tests in the future

tests/test_light_client.nim
```diff
@@ -9,230 +9,246 @@
...
-      # Advance by a couple epochs
-      for i in 1'u64 .. 10:
-        let headSlot = (finalizedSlot.epoch + i).start_slot
-        cpDag.advanceToSlot(headSlot, verifier, quarantine[])
-
-      check true
-
-withAll(LightClientDataFork):
-  when lcDataFork > LightClientDataFork.None:
-    runTest(lcDataFork)
```
etan-status, line 238: changed this test to only run once, but instead perform store upgrades throughout the test.
this was not possible with previous design not using forked store.

tests/test_light_client_processor.nim
```diff
@@ -98,108 +102,124 @@ suite "Light client processor" & preset():
...
-        check:
-          res.isOk
-          store[].isSome
-          if forkyUpdate.finalized_header.beacon.slot >
-              forkyBootstrap.header.beacon.slot:
-            store[].get.finalized_header == forkyUpdate.finalized_header
-          else:
-            store[].get.finalized_header == forkyBootstrap.header
-          store[].get.optimistic_header == forkyUpdate.attested_header
+        check update.kind <= store[].kind
```
etan-status, line 134: this check needs to happen after `storeObject` to handle case where `update.kind` actually triggers store upgrade

etan-status: cannot test store upgrade here because `CAPELLA_FORK_EPOCH` non-`FAR_FUTURE_EPOCH` is broken in this test as well. there is a third test suite in `consensus_spec` that tests upgrades as well, so not a blocker.

tersec:
It's an improvement over the status quo

tersec (approved):

arnetheduck:
Do note that this PR is likely affected by the general db upgrade policy: https://github.com/status-im/nimbus-eth2/pull/4499#issuecomment-1383572412

etan-status:
Yeah, writing to both legacy and new tables now, and reading only from new one. On startup, import legacy into new.

etan-status:
BTW, would be great to have the DB policy in the style guide @arnetheduck

etan-status:
Putting this one on automerge for now to unblock Capella/EIP4844 implementation.
I think the DB is now fine once more. If there is more, will followup in separate PR.

PR #4518, opened 2023-01-17
support `readOnly` backend in LC client/data dbs

etan-status:
When accessing DB in `readOnly` mode that does not already have latest schema, initial writes trigger `attempt to write a readonly database`. Avoid that by only writing schema when DB is not `readOnly`, and provide data from legacy tables if such are present.

### DIFF
--- a/beacon_chain/light_client_db.nim
+++ b/beacon_chain/light_client_db.nim
@@ -32,6 +32,7 @@ type
     Finalized = 1  # Latest finalized header
 
   LegacyLightClientHeadersStore = object
+    getStmt: SqliteStmt[int64, (int64, seq[byte])]
     putStmt: SqliteStmt[(int64, seq[byte]), void]
 
   LightClientHeadersStore = object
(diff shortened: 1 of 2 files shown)


beacon_chain/light_client_db.nim
```diff
@@ -221,6 +261,8 @@ func putSyncCommittee*(
     db: LightClientDB, period: SyncCommitteePeriod,
     syncCommittee: altair.SyncCommittee) =
   doAssert period.isSupportedBySQLite
+  if distinctBase(db.syncCommittees.putStmt) == nil:
```
arnetheduck: this should probably be an assert, at least that's what we've done elsewhere - ie the higher-level code might be assuming writing happened and get into a messy state

arnetheduck (approved):

beacon_chain/light_client_db.nim
etan-status: higher-level code can't assume that, as there may be concurrent writes from different process, or other DB corruption.

But agree, having an assert to highlight API misuse doesn't hurt.

etan-status: would go even further and assert on `not readOnly` - the current condition would hold in readonly mode (where tables already exist)

arnetheduck: > higher-level code can't assume that, as there may be concurrent writes from different process, or other DB corruption.

not really, de facto we require exclusive writing access to the sqlite db while the bn is running - the readonly mode is an expression of that: the code (in chaindag in particular), as written, doesn't work with multiple writers.

arnetheduck: ..and we assert on every other form of corruption we detect

etan-status: Got it!

arnetheduck:
btw, we do have tests for the readonly mode - clearly, they don't cover the light client stuff though, probably a good idea to add some so we don't regress (since the readonly mode is used in utilities mainly, not the main app)

PR #4519, opened 2023-01-18
exit/validatorchange pool includes BLS to execution messages; REST support for new pool

tersec:
- `ExitPool` -> `ValidatorChangePool`
- Add BLS to execution changes to this validator change pool
- in `block_processor`, clean up some Nim warnings about duplicate imports because it doesn't like `from a import b` and `from a import c` as two separate statements and wants them combined into `from a import b, c`
- rename most variables and function paramaters of type `ValidatorChangePool` from `exitPool` to `validatorChangePool` and `exits` to `validatorChanges`
- implement gossip broadcast and validation of BLS to execution changes
- implement BLS to execution change-related Beacon API calls in REST server
- include BLS to execution changes in proposed blocks
- REST error message grammar tweaks
- remove some vestigial passing of the BLS changes as a separate parameter during block construction, since it's now part of the `ValidatorChangePool`

The exits and BLS changes are structurally near-isomorphic:
- they happen either 0 or 1 times per validator lifetime
- they aren't functions of blocks or forks, unlike, e.g, the attestation pool, which needs to care about the DAG
- they're surfaced in exactly two ways: being packed into blocks and via a pair of `GET`/`POST` REST Beacon APIs to introspect the pool
- they're initiated by either by the `POST` Beacon API or a gossip message on a dedicated gossip topic

### DIFF
--- a/beacon_chain/consensus_object_pools/exit_pool.nim
+++ b/beacon_chain/consensus_object_pools/exit_pool.nim
@@ -30,7 +30,7 @@ const
   VOLUNTARY_EXITS_BOUND = MAX_VOLUNTARY_EXITS * 4
 
   # For Capella launch; scale back later
-  BLS_TO_EXECUTION_CHANGES_BOUND = 20000'u64
+  BLS_TO_EXECUTION_CHANGES_BOUND = 32768'u64
 
 type
   OnVoluntaryExitCallback =
@@ -85,7 +85,7 @@ func init*(T: type ValidatorChangePool, dag: ChainDAGRef,
       # TODO scale-back to BLS_TO_EXECUTION_CHANGES_BOUND post-capella, but
       # given large bound, allow to grow dynamically rather than statically
       # allocate all at once
-      initDeque[SignedBLSToExecutionChange](initialSize = 1000),
+      initDeque[SignedBLSToExecutionChange](initialSize = 1024),
     dag: dag,
     attestationPool: attestationPool,
     onVoluntaryExitReceived: onVoluntaryExit
--- a/beacon_chain/consensus_object_pools/exit_pool.nim
+++ b/beacon_chain/consensus_object_pools/exit_pool.nim
@@ -29,8 +29,8 @@ const
   PROPOSER_SLASHINGS_BOUND = MAX_PROPOSER_SLASHINGS * 4
   VOLUNTARY_EXITS_BOUND = MAX_VOLUNTARY_EXITS * 4
 
-  # For Capella launch; can scale back later
-  BLS_TO_EXECUTION_CHANGES_BOUND = MAX_BLS_TO_EXECUTION_CHANGES * 8'u64
+  # For Capella launch; scale back later
+  BLS_TO_EXECUTION_CHANGES_BOUND = 20000'u64
 
 type
   OnVoluntaryExitCallback =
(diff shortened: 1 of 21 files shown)


beacon_chain/consensus_object_pools/exit_pool.nim
```diff
@@ -29,13 +29,17 @@ const
   PROPOSER_SLASHINGS_BOUND = MAX_PROPOSER_SLASHINGS * 4
   VOLUNTARY_EXITS_BOUND = MAX_VOLUNTARY_EXITS * 4
 
+  # For Capella launch; can scale back later
```
arnetheduck: can't we compute a bound here based on the number of unmigrated validators? there should be no more than 1 message per validator, ever, right?

tersec: Approximately 310k

tersec: Any given node will never need to pack nearly that many blocks though, at 16 per block

`sizeof(SignedBLSToExecutionChange) = 176` so even optimally, that's around ~25MB (310k * 176), which is significant for not-very-useful storage

arnetheduck: though that number will go down as time passes - my understanding is that everyone will want to get these done asap so there's going to be a burst in the beginning then slowly it'll plane out - the point being that an aggressive limit will cause drops and people will have to resubmit which is tricky due to libp2p gossipsub message cache retention / deduplication etc

arnetheduck: actually, due to the "seen" flag which is unbounded, we'll drop resubmits

arnetheduck: hm, looking through the logic, I can see that we skip messages that are no longer valid for the given state, but it doesn't look like we purge them from the queue.. so I guess of someone includes bls messages in a different order than us, we'd potentially be throwing out messages that haven't been added to the state and retaining such that have?

Also, it seems that if we're building blocks, it would make sense to prioritize those messages that pertain to "our own" validators, right? Though that's probably the case for exits too, and unrelated to this PR

tersec: > actually, due to the "seen" flag which is unbounded, we'll drop resubmits

This aspect/policy of the logic is tweakable enough -- if one wants to drop the `seen` flag too, that can keep them consistent at least.

arnetheduck: > can keep them consistent at least. 

yeah, that's probably important.. I'd still go for a dynamic limit I think, and not prealloc it but rather just let it grow naturally - 23 mb doesn't sound terrible for the ux gain of not having to lose them. it would however mean that we should likely prune on block construction at least so we don't keep the already-executed ones in the pool. 

anyway, with the seen-clearing, I think this PR can be GTG but I'd consider the other bits seriously

arnetheduck (approved):

arnetheduck:
post-release perhaps though, a bit risky to make pool changes this late

beacon_chain/consensus_object_pools/exit_pool.nim
tersec: > Also, it seems that if we're building blocks, it would make sense to prioritize those messages that pertain to "our own" validators, right? Though that's probably the case for exits too, and unrelated to this PR

Agree on all points.

tersec: https://github.com/status-im/nimbus-eth2/pull/4519/commits/16e61642d9a48f55acf0212f3bda88786e9a8a66 increases the pool to 20k -- enough to avoid most acute issues, but also enough to avoid worst of dropping, and keeps things consistent

tersec: Teku's approach: https://github.com/ConsenSys/teku/commit/1e73f7c122124fe9a9f8efcdcfdb093682b529a2

PR #4527, opened 2023-01-19
Wire up blob fetching with request manager

henridf:
(Putting this up while I work on using BlobsSidecarsbyRange in the sync manager, which is proving to be a challenge).


beacon_chain/sync/request_manager.nim
```diff
@@ -141,6 +167,82 @@ proc fetchAncestorBlocksFromNetwork(rman: RequestManager,
     if not(isNil(peer)):
       rman.network.peerPool.release(peer)
 
+proc fetchAncestorBlocksAndBlobsFromNetwork(rman: RequestManager,
```
henridf, line 168: (This has a lot of duplication with `fetchAncestorBlocksFromNetwork`, but there isn't a shared chunk that can be cleanly factored out.)

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -333,6 +333,16 @@ proc initFullNode(
       blockProcessor[].addBlock(MsgSource.gossip, signedBlock,
                                 Opt.none(eip4844.BlobsSidecar), resfut)
       resfut
+    blockBlobsVerifier = proc(signedBlock: ForkedSignedBeaconBlock, blobs: eip4844.BlobsSidecar):
```
henridf, line 334: Initially, rather than have a separate blockBlobsVerifier, I added a Opt[eip4844.BlobsSidecar] parameter to the existing blockVerifier, but ran into  issues (as pointed out by @tersec, likely https://github.com/status-im/nim-stew/pull/167). 

This seems fine too, so leaving, but can switch to Opt if there is a strong preference toward that (will require a bit more gymnastics, with a separate PR to update nim-stew, etc).

beacon_chain/sync/request_manager.nim
```diff
@@ -48,13 +54,33 @@ func shortLog*(x: seq[FetchRecord]): string =
...
+  var checks = @roots
+  if len(blocks) > len(roots):
+    return false
+  for item in blocks:
+    let res = checks.find(item[].beacon_block.root)
+    if res == -1:
+      return false
+    else:
+      checks.del(res)
+  return true
```
```suggestion
  true
```

henridf:
Not sure why just one build is failing, but this message makes me wonder if it's some infra issue

> Error: The log was not found. It may have been deleted based on retention settings.

henridf:
CI is green.

beacon_chain/nimbus_beacon_node.nim
tersec: Am going to merge https://github.com/status-im/nimbus-eth2/pull/4541 when it passes CI, so maybe `Opt` will work here

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -376,7 +386,9 @@ proc initFullNode(
   node.processor = processor
   node.blockProcessor = blockProcessor
   node.consensusManager = consensusManager
-  node.requestManager = RequestManager.init(node.network, blockVerifier)
+  node.requestManager = RequestManager.init(node.network, dag,
+                                            blockVerifier, blockBlobsVerifier,
+                                            dag.cfg.EIP4844_FORK_EPOCH)
```
tersec: Why pass both `dag` and `dag.cfg.EIP4844_FORK_EPOCH`? The former provides access to the latter.

beacon_chain/sync/request_manager.nim
```diff
@@ -154,8 +256,13 @@ proc requestManagerLoop(rman: RequestManager) {.async.} =
 
       let start = SyncMoment.now(0)
 
+      let getBlobs = rman.isSlotWithBlobs(rman.dag.head.slot)
```
tersec: Is this the only reason `RequestManager` needs a reference to the DAG?

beacon_chain/nimbus_beacon_node.nim
henridf: No good reason (step-by-step implementation without properly cleaning up). Fixed in eff1fb33df204f2233d6134f25cfedaa6d9f187b.

beacon_chain/sync/request_manager.nim
henridf: Yes. I don't really like bringing in that reference, but the RequestManager needs to know where the head is at.

beacon_chain/sync/request_manager.nim
```diff
@@ -45,13 +50,32 @@ func shortLog*(x: seq[FetchRecord]): string =
...
+    blockBlobsVerifier: blockBlobsVerifier,
   )
 
+proc checkResponse(roots: openArray[Eth2Digest],
+                   blocks: openArray[ref SignedBeaconBlockAndBlobsSidecar]): bool =
+  ## This procedure checks peer's response.
+  var checks = @roots
+  if len(blocks) > len(roots):
+    return false
+  for item in blocks:
```
tersec, line 74: https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.1/specs/phase0/p2p-interface.md#configuration defines `MAX_REQUEST_BLOCKS` to be to be 1024, so this `O(n^2)` search (`n` blocks, each `checks.find` is `O(n)`, and then each `checks.del` is another `O(n)`) is a bit much, probably.

There's always a risk of asymptotically faster algorithms being regardless slow, and maybe it turns out that even the pathological case here just isn't that bad, or that in practice Nimbus might tend to request fewer blocks (e.g., 32 at a time). But if the reasonableness of this depends on that latter, that should be documented, and if it doesn't create too much constant factor, it's potentially better to use one of the many `O(n log n)` approaches (some supported better by Nim's stdlib than others, and there's of course value in the simplicity of avoiding involved search algorithm implementations which'd need their own testing, etc).

henridf: Hmm. Good point, but the fact that we haven't run into issues with the existing checkResponse (https://github.com/status-im/nimbus-eth2/blob/ba7c0bc091161f261148dbfcb27b21cc48f1fdf8/beacon_chain/sync/request_manager.nim#L58), which is identical, makes me think we should be ok for now (and if/when we address this, we should address the pre-existing checkResponse and this one together).

If I had to guess, I suppose this hasn't been an issue because these requests are always small in practice. Which isn't a guarantee of course, and could even be a dos vector.

tersec: Fair enough, and you have the `len` sanity check, so Nimbus can immediately throw out responses which might trigger this, so it can avoid the issue if it wants by just requesting fewer roots at a time.

So I don't think this is remotely exploitable, with that guard you put in, at least. It's not great, but yes, probably for the moment better to use an already tested-in-Nimbus/known-working approach to get something functioning, then optimize.

beacon_chain/sync/request_manager.nim
tersec: I'm not sure I understand why though. My read of https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#beaconblockandblobssidecarbyroot-v1 is that `/eth2/beacon_chain/req/beacon_block_and_blobs_sidecar_by_root/1/` will only work for EIP4844 blocks, so even if the head has `s.epoch >= rman.dag.cfg.EIP4844_FORK_EPOCH` (`isSlotWithBlobs`), Nimbus might request a root which is pre-EIP4844.

This could happen a few ways, e.g., a nonfinalizing network where one fork advances through `EIP4844_FORK_EPOCH`, and another has been a side chain, but then the network as a whole reorgs to a new chain -- but this node doesn't have those blocks or blobs, so it has to request them. If it looks like

```
slot 600: ancestor of current head
slot 610: ancestor of another sidechain

-------------------
EIP-4844 fork epoch
-------------------

slot 700: current head, based on 600
slot 710: future reorg target based on 610 (which this node rejected, for any of several reasons. either way, needs to request it)
```
But the current logic requests from slot 610 via an EIP4844 req/resp protocol endpoint which won't support it, from what I can tell, because once its head reaches EIP4844, it will become unable to request pre-EIP4844 slots in the event of such reorgs.

This isn't a major hazard on a finalizing network, but I'm not sure it's a reliable heuristic on a nonfinalizing network.

One can also construct failure scenarios the opposite direction -- this node's head is stuck just before EIP4844 fork epoch, while a side chain it's not for whatever reason following reaches/passes the EIP4844 fork epoch.

Rather, is this something that needs to be know specifically `root` by `root`, or am I missing something, and it's somehow possible to extrapolate from the DAG head to all the roots one might request?

beacon_chain/sync/request_manager.nim
```diff
@@ -45,13 +50,32 @@ func shortLog*(x: seq[FetchRecord]): string =
   "[" & x.mapIt(shortLog(it.root)).join(", ") & "]"
 
 proc init*(T: type RequestManager, network: Eth2Node,
-           blockVerifier: BlockVerifier): RequestManager =
+              dag: ChainDAGRef,
+              blockVerifier: BlockVerifier,
+              blockBlobsVerifier: BlockBlobsVerifier)
```
tersec: Nim doesn't seem to like the `:` being put on the next line. I think either
```nim
            blockBlobsVerifier: BlockBlobsVerifier): RequestManager
```
or
```nim
             blockBlobsVerifier: BlockBlobsVerifier):
             RequestManager
```
should work

beacon_chain/sync/request_manager.nim
henridf: Hmm, I had come up with simpler scenarios, and had sort of convinced myself they would work, but I realize now that's not the case. Even in this simpler example without forks:

```text
------------------
slot 599: current head
slot 600: EIP-4844 fork epoch
slot 601
slot 602: we received this block and need to fill in 600,601
------------------
```

where we would request  600 and 601 using the pre-EIP488 RPC, which would give no results. (from https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/p2p-interface.md#beaconblocksbyroot-v2: "After EIP4844_FORK_EPOCH, BeaconBlocksByRootV2 is replaced by BeaconBlockAndBlobsSidecarByRootV1").

And so we get stuck, I think.

It seems like there's a fundamental issue here, since we have no idea of knowing based on a root whether it's pre or post a particular epoch. 

The only thing I can think of is some sort of fallback where we try the other endpoint if one doesn't return results. Not great but I'm not seeing any other way.

beacon_chain/sync/request_manager.nim
henridf: fixed f05b7503ce9d34522f886a39cf6c7c28044d9ea0

tersec:
```nim
proc f()
     : int = 5
```
for example -- https://play.nim-lang.org/#ix=4lRc

beacon_chain/sync/request_manager.nim
tersec: Yeah, what are other CLs doing, I guess? Seems ugly.

henridf: Thinking the same thing. Will be a good discussion topic soon.

beacon_chain/nimbus_beacon_node.nim
tersec: Merged -- should be able to use `Opt` now

beacon_chain/sync/request_manager.nim
henridf: @tersec: in 26b8cc94f81a5839043896cb7fccc9a73c696fc5, I modified the "fetch blocks vs blobs" logic. Per the comment: 

```
      # As soon as EIP4844_FORK_EPOCH comes around in wall time, we
      # switch to requesting blocks and blobs. In the vicinity of the
      # transition, that means that we *may* request blobs for a
      # pre-eip4844. In that case, we get ResourceUnavailable from the
      # peer and fall back to requesting blocks only.
```

beacon_chain/sync/request_manager.nim
```diff
@@ -154,8 +267,18 @@ proc requestManagerLoop(rman: RequestManager) {.async.} =
 
       let start = SyncMoment.now(0)
 
+      # As soon as EIP4844_FORK_EPOCH comes around in wall time, we
+      # switch to requesting blocks and blobs. In the vicinity of the
+      # transition, that means that we *may* request blobs for a
+      # pre-eip4844. In that case, we get ResourceUnavailable from the
+      # peer and fall back to requesting blocks only.
+      let getBlobs = rman.isBlobsTime(rman.dag.head.slot)
```
tersec: This isn't wall time -- the comment's logic I agree with, but that's not what this does. As a bonus, by following wall time rather than head, no need to pass the DAG anymore.

henridf: Removed the `slot` parameter to `isBlobsTime` (which _was_ looking at wall clock, the slot parameter was leftover from before).

d78ec80f5f7558bf49c4952a1e02cb697199ca09

Re the dag, we do at least need `EIP4844_FORK_EPOCH`. I could just pass the `RuntimeConfig` (`dag.cfg`) to the request manager.

beacon_chain/sync/request_manager.nim
```diff
@@ -138,6 +165,92 @@ proc fetchAncestorBlocksFromNetwork(rman: RequestManager,
...
+    debug "Error while fetching ancestor blocks", exc = exc.msg,
+          items = shortLog(items), peer = peer, peer_score = peer.getScore()
+    raise exc
+  finally:
+    if not(isNil(peer)):
+      rman.network.peerPool.release(peer)
+
+
+proc isBlobsTime(rman: RequestManager): bool =
+  return rman.getBeaconTime().slotOrZero.epoch >= rman.dag.cfg.EIP4844_FORK_EPOCH
```
```suggestion
  rman.getBeaconTime().slotOrZero.epoch >= rman.dag.cfg.EIP4844_FORK_EPOCH
```
Can use implicit return here.

But also, yes, since now `EIP4844_FORK_EPOCH` is a constant (per network), it would be better to just pass `EIP4844_FORK_EPOCH` rather than the whole DAG.

henridf: Done in 2bbf147

tersec (approved):

beacon_chain/sync/request_manager.nim
```diff
@@ -154,8 +267,18 @@ proc requestManagerLoop(rman: RequestManager) {.async.} =
...
+      # As soon as EIP4844_FORK_EPOCH comes around in wall time, we
+      # switch to requesting blocks and blobs. In the vicinity of the
+      # transition, that means that we *may* request blobs for a
+      # pre-eip4844. In that case, we get ResourceUnavailable from the
+      # peer and fall back to requesting blocks only.
+      let getBlobs = rman.isBlobsTime()
       for i in 0 ..< PARALLEL_REQUESTS:
-        workers[i] = rman.fetchAncestorBlocksFromNetwork(rootList)
+        if getBlobs:
+          workers[i] = rman.fetchAncestorBlocksAndBlobsFromNetwork(rootList)
```
tersec: One slight simplification is to write this using an `if` expression rather than statement:
```nim
        workers[i] = block:
          if getBlobs:
            rman.fetchAncestorBlocksAndBlobsFromNetwork(rootList)
          else:
            rman.fetchAncestorBlocksFromNetwork(rootList)
```
either what you have or that is fine, but that does factor out a little redundancy.

The `block` isn't strictly necessary in this simple case (because the `if` is just one expression, not a sequence of statements or expressions of which only the value of the last of which is used), but it makes the indentation a bit less ambiguous. One can write it the way you write it here, the `block:` version, and the just `foo = if: ... else: ...` versions around Nimbus.

henridf: i'm all for expressions! ccfc99d

PR #4537, opened 2023-01-20
make message_router.routeSignedBeaconBlock blob-aware

henridf:
The new code paths from this commit should not be exercised until block proposal is made eip4844-capable.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -2653,22 +2653,12 @@ proc broadcastBeaconBlock*(
   let topic = getBeaconBlocksTopic(node.forkDigests.capella)
   node.broadcast(topic, blck)
 
+proc broadcastBeaconBlockAndBlobsSidecar*(
+    node: Eth2Node, blck: eip4844.SignedBeaconBlockAndBlobsSidecar): Future[SendResult] =
+  let topic = getBeaconBlockAndBlobsSidecarTopic(node.forkDigests.eip4844)
+  node.broadcast(topic, blck)
+
 from ../spec/datatypes/eip4844 import SignedBeaconBlock
-
-proc broadcastBeaconBlock*(
-    node: Eth2Node, blck: eip4844.SignedBeaconBlock): Future[SendResult] =
-  debugRaiseAssert $eip4844ImplementationMissing & ": eth2_network.nim:broadcastBeaconBlock EIP4844 uses different approach (1)"
-
-proc broadcastBeaconBlock*(
-    node: Eth2Node, forked: ForkedSignedBeaconBlock): Future[SendResult] =
-  withBlck(forked):
-    when stateFork == BeaconStateFork.EIP4844:
-      debugRaiseAssert $eip4844ImplementationMissing & ": eth2_network.nim:broadcastBeaconBlock EIP4844 uses different approach (2)"
-      let f = newFuture[SendResult]()
-      f.fail(new CatchableError)
-      f
-    else:
-      node.broadcastBeaconBlock(blck)
 
 proc broadcastSyncCommitteeMessage*(
     node: Eth2Node, msg: SyncCommitteeMessage,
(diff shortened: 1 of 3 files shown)


beacon_chain/spec/forks.nim
```diff
@@ -272,6 +273,15 @@ template toSignedBeaconBlock*(b: ForkySignedBeaconBlockMaybeBlobs): ForkySignedB
...
 
+func optBlobs*(b: ForkySignedBeaconBlockMaybeBlobs):
+         Opt[eip4844.BlobsSidecar] =
+  when b is phase0.SignedBeaconBlock or b is altair.SignedBeaconBlock or
+     b is bellatrix.SignedBeaconBlock or b is capella.SignedBeaconBlock:
+    Opt.none(eip4844.BlobsSidecar)
+  elif b is eip4844.SignedBeaconBlockAndBlobsSidecar:
+    Opt.some(b.blobs_sidecar)
+
+
```
tersec: extra newline

henridf:
```suggestion

```

henridf: fixed

beacon_chain/networking/eth2_network.nim
```diff
@@ -2653,22 +2653,12 @@ proc broadcastBeaconBlock*(
...
-
-proc broadcastBeaconBlock*(
-    node: Eth2Node, blck: eip4844.SignedBeaconBlock): Future[SendResult] =
-  debugRaiseAssert $eip4844ImplementationMissing & ": eth2_network.nim:broadcastBeaconBlock EIP4844 uses different approach (1)"
+proc broadcastBeaconBlockAndBlobsSidecar*(
+    node: Eth2Node, blck: eip4844.SignedBeaconBlockAndBlobsSidecar): Future[SendResult] =
+  let topic = getBeaconBlockAndBlobsSidecarTopic(node.forkDigests.eip4844)
+  node.broadcast(topic, blck)
 
-proc broadcastBeaconBlock*(
```
tersec, line 2662: Mildly surprising this isn't necessary/used, but evidently.

beacon_chain/spec/forks.nim
```diff
@@ -272,6 +273,13 @@ template toSignedBeaconBlock*(b: ForkySignedBeaconBlockMaybeBlobs): ForkySignedB
   else:
     b
 
+func optBlobs*(b: ForkySignedBeaconBlockMaybeBlobs):
+         Opt[eip4844.BlobsSidecar] =
+  when b is phase0.SignedBeaconBlock or b is altair.SignedBeaconBlock or
+     b is bellatrix.SignedBeaconBlock or b is capella.SignedBeaconBlock:
+    Opt.none(eip4844.BlobsSidecar)
+  elif b is eip4844.SignedBeaconBlockAndBlobsSidecar:
+    Opt.some(b.blobs_sidecar)
```
henridf, line 282:
```suggestion
    Opt.some(b.blobs_sidecar)
    
```

beacon_chain/networking/eth2_network.nim
tersec: One way to check, in general (it's probably safe here, but... maybe worth checking in case) is to first run a full build cycle:
```
make LOG_LEVEL=TRACE test && make LOG_LEVEL=TRACE
```
on a version where this is marked `deprecated`:
```nim
proc broadcastBeaconBlock*(
    node: Eth2Node, forked: ForkedSignedBeaconBlock): Future[SendResult] {.deprecated.} =
```
(redirect stdout/stderr as necessary to capture output for later searching)

The point isn't to commit this to the repo, but to see if, anywhere Nim thinks it should be calling this function.

Sometimes there are overloads it manages to find, if one is removed, but which aren't correct. This case is probably unproblematic, because how many `broadcastBeaconBlock`s which take a `ForkedSignedBeaconBlock` can there be.

tersec (approved):
Assuming
```nim
proc broadcastBeaconBlock*(
    node: Eth2Node, forked: ForkedSignedBeaconBlock): Future[SendResult] {.deprecated.} =
```
is unused, as it probably is, then this looks good.

beacon_chain/networking/eth2_network.nim
henridf: Thanks for the info - didn't know about this pragma.

PR #4540, opened 2023-01-21
eip4844 beacon block proposals

henridf:
Still need to add nim-kzg library and use it to compute the aggregated proof.


beacon_chain/validators/validator_duties.nim
```diff
@@ -437,6 +440,32 @@ proc getExecutionPayload[T](
...
+  let payload = try:
+    awaitWithTimeout(
+      node.consensusManager.eth1Monitor.getBlobsBundleV1(payload_id),
+      GETBLOBS_TIMEOUT):
+        beacon_block_payload_errors.inc()
+        warn "Getting blobs sidecar from Engine API timed out", payload_id
+        default(BlobsBundleV1)
+  except CatchableError as err:
+    beacon_block_payload_errors.inc()
+    warn "Getting execution payload from Engine API failed",
```
tersec: Mismatched warning message? This is about blobs, not execution payloads

beacon_chain/validators/validator_duties.nim
```diff
@@ -437,6 +440,32 @@ proc getExecutionPayload[T](
       msg = err.msg
     return Opt.some empty_execution_payload
 
+proc getBlobsBundle(
+    node: BeaconNode, epoch: Epoch, validator_index: ValidatorIndex,
+    payload_id: PayloadID): Future[BlobsBundleV1] {.async.} =
+  # https://github.com/ethereum/consensus-specs/blob/v1.1.10/specs/bellatrix/validator.md#executionpayload
```
tersec: Not relevant spec URL?

beacon_chain/spec/state_transition.nim
```diff
@@ -482,10 +486,11 @@ template partialBeaconBlock*(
       voluntary_exits: validator_changes.voluntary_exits,
       sync_aggregate: sync_aggregate,
       execution_payload: execution_payload,
-      bls_to_execution_changes: validator_changes.bls_to_execution_changes
+      bls_to_execution_changes: validator_changes.bls_to_execution_changes,
+      blob_kzg_commitments: kzg_commitments
       ))
 
-proc makeBeaconBlock*[T: bellatrix.ExecutionPayload | capella.ExecutionPayload](
+proc makeBeaconBlock*[T: bellatrix.ExecutionPayload | capella.ExecutionPayload | eip4844.ExecutionPayload](
```
tersec: Long line

tersec:
failing on `make wss_sim` -- needs the KZG commitment parameters added there too

tersec:
failing on `make wss_sim` -- needs the KZG commitment parameters added there too

beacon_chain/spec/state_transition.nim
henridf: fixed

beacon_chain/validators/validator_duties.nim
henridf: oops, fixed

henridf:
> needs the KZG commitment parameters added there too

done

henridf:
> Still need to add nim-kzg library and use it to compute the aggregated proof.

My preference would be to do that in a follow-on PR (presumably tomorrow), fwiw. I think in the current state, we'd propose invalid eip4844 blocks, but no behavior should change for pre-eip4844.

tersec:
> > Still need to add nim-kzg library and use it to compute the aggregated proof.
> 
> My preference would be to do that in a follow-on PR (presumably tomorrow), fwiw. I think in the current state, we'd propose invalid eip4844 blocks, but no behavior should change for pre-eip4844.

My concern is that I suspect this won't pass Jenkins testnet finalization at all because of the `minimal` preset issues with EIP4844 you discovered. It hasn't gotten that far yet, and I'm not really sure why. I'm fine with merging this as long as it does something sensible with EIP4844 minimal presets and engine API calls.

(To summarize: `BlobsBundleV1` in https://github.com/ethereum/execution-apis/blob/main/src/engine/experimental/blob-extension.md#blobsbundlev1  assumes a `mainnet` preset blob size, where https://github.com/ethereum/consensus-specs/blob/dev/presets/minimal/eip4844.yaml defines `FIELD_ELEMENTS_PER_BLOB: 4`, which leads to a Blob according to https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/polynomial-commitments.md#custom-types having a length of `4 * 32`, not `4096 * 32` which `BlobsBundleV1` requires.)

tersec:
I agree, `nim-kzg` should be in another PR though, not this one.

henridf:
> something sensible with EIP4844 minimal presets and engine API calls.

Following your suggestion, made a change 2420d69011879610ae012b4a078091bc7fe99c5f to skip blob fetching when the preset is minimal. This means that all block proposals will be invalid for EIP4844-minimal, but as we've agreed elsewhere, EIP4844 is mainnet-preset only for now.

tersec:
> > something sensible with EIP4844 minimal presets and engine API calls.
> 
> Following your suggestion, made a change [2420d69](https://github.com/status-im/nimbus-eth2/commit/2420d69011879610ae012b4a078091bc7fe99c5f) to skip blob fetching when the preset is minimal. This means that all block proposals will be invalid for EIP4844-minimal, but as we've agreed elsewhere, EIP4844 is mainnet-preset only for now.

Seems fine, yeah.

Will wait to see CI a bit, but inclined to merge at this point, and can build on it in future PRs.

scripts/launch_local_testnet.sh
```diff
@@ -785,6 +785,8 @@ DEPOSIT_CONTRACT_ADDRESS: ${DEPOSIT_CONTRACT_ADDRESS}
 ETH1_FOLLOW_DISTANCE: 1
 ALTAIR_FORK_EPOCH: 1
 BELLATRIX_FORK_EPOCH: 2
+CAPELLA_FORK_EPOCH: 3
```
tersec: This works, Even without separate VC support?

tersec: I did this and it worked fine a while ago but only after disabling the separate VC validators and only using the built-in BN validators.

henridf: Oops, this is unintentional. Juggling between branches and my own tests. Thanks for the catch.

henridf: 311ddd19fd0de3fd57837e5d2b23d60f6554c2de

tersec (approved):

PR #4545, opened 2023-01-24
allow compile-time changing of SECONDS_PER_SLOT

tersec:
It's one approach -- kind of speicalized/special-cased to this particular change, but not sure it's worth a general solution


arnetheduck:
I suspect at some point we have to bite the bullet and use a runtime parameter for this - ie going against the stream is likely not worth the workarounds

PR #4549, opened 2023-01-24
Some EIP4844 fixes

henridf:
Fix some bugs that were exposed by using the local testnet.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -403,13 +403,15 @@ proc storeBlock*(
   # Establish blob viability before calling addHeadBlock to avoid
   # writing the block in case of blob error.
   when typeof(signedBlock).toFork() >= BeaconBlockFork.EIP4844:
-    if blobs.isSome() and not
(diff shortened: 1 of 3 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -403,13 +403,15 @@ proc storeBlock*(
...
-    if blobs.isSome() and not
-         validate_blobs_sidecar(signedBlock.message.slot,
-                                hash_tree_root(signedBlock.message),
-                                signedBlock.message
-                                .body.blob_kzg_commitments.asSeq,
-                                blobs.get()).isOk():
-       return err((VerifierError.Invalid, ProcessingStatus.completed))
+    if blobs.isSome():
+      let res = validate_blobs_sidecar(signedBlock.message.slot,
+                                       hash_tree_root(signedBlock.message),
```
arnetheduck: throughout, the root is cached in the block, ie https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/spec/datatypes/phase0.nim#L227

henridf: 69a455a

PR #4550, opened 2023-01-24
WIP: decoupled sidecars

arnetheduck:
Research implementation of blob sidecars


arnetheduck:
obsoleted by other efforts

PR #4551, opened 2023-01-24
Local sim impovements

zah:
* Added support for running Capella and EIP-4844 simulations by downloading the correct version of Geth.

* Added support for using Nimbus remote signer and Web3Signer. Use 2 out of 3 threshold signing configuration in the mainnet configuration and regular remote signing in the minimal one.

* The local testnet simulation can now use a payload builder. This is currently not activated in CI due to lack of automated procedures for installing third-party relays or builders.

  You are adviced to use mergemock for now, but for most realistic results, we can create a simple builder based on the nimbus-eth1 codebase that will be able to propose transactions from the regular network mempool.

* Start the simulation from a merged state. This would allow us to start removing pre-merge functionality such as the gossip subsciption logic. The commit also removes the merge-forcing hack installed after the TTD removal.

* Consolidate all the tools used in the local simulation into a single `local_testnet_simulation` binary.


beacon_chain/spec/datatypes/bellatrix.nim
```diff
@@ -151,6 +151,7 @@ type
   BeaconStateRef* = ref BeaconState not nil
   NilableBeaconStateRef* = ref BeaconState
 
+  # TODO: There should be only a single generic HashedBeaconState definition
```
tersec, line 155: As long as this is done in a way which allows for importing of only consensus spec fork definitions through, say, Altair or Bellatrix, rather than creating a basically artificial requirement for doing so through whatever the hard fork happens to be.

beacon_chain/spec/forks.nim
```diff
@@ -934,6 +983,14 @@ func nextForkEpochAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): Epoch =
   of BeaconStateFork.Altair:    cfg.BELLATRIX_FORK_EPOCH
   of BeaconStateFork.Phase0:    cfg.ALTAIR_FORK_EPOCH
 
+func forkVersion*(cfg: RuntimeConfig, consensusFork: ConsensusFork): Version =
```
tersec: At some point, why not remove these separate fork setups? There's no substantive distinction between `ConsensusFork`, `BeaconStateFork`, and `BeaconBlockFork`. Even the latter two being distinct still hasn't proven useful that I can discern, never mind adding a third equivalent enum.

ncli/local_testnet_simulation.nim
```diff
@@ -0,0 +1,653 @@
+# beacon_chain
+# Copyright (c) 2018-2022 Status Research & Development GmbH
```
tersec: Copyright years?

ncli/local_testnet_simulation.nim
```diff
@@ -0,0 +1,653 @@
...
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+when (NimMajor, NimMinor) < (1, 4):
```
tersec: Doesn't need Nim 1.2 support anymore, can just use `{.push raises: [].}`

tersec:
Might it be better to bump `nim-eth` and `nim-web3` via a separate PR?

beacon_chain/spec/beaconstate.nim
```diff
@@ -891,6 +788,249 @@ func get_next_sync_committee*(
   res.aggregate_pubkey = finish(attestersAgg).toPubKey()
   res
 
+# Phase0 genesis is used only in tests nowadays
+when defined(withPhase0Genesis):
```
tersec: Is this worth creating more compile-time `when`s just to avoid compiling a few functions, which then risk bitrotting, as did/have genesis detection?

tests/nim.cfg
```diff
@@ -1,6 +1,6 @@
 # Use only `secp256k1` public key cryptography as an identity in LibP2P.
 -d:"libp2p_pki_schemes=secp256k1"
-
+-d:withPhase0Genesis
```
tersec: All these special compile variables are fallout of the existence of `withPhase0Genesis` (it's not obvious why some of the `.cfg` files need it regardless, but they definitely wouldn't without said compile-time define)

ncli/logtrace.nim
```diff
@@ -22,55 +24,55 @@ const
                     LogTraceCopyright & "\r\n"
 
 type
-  StartUpCommand {.pure.} = enum
+  StartUpCommand* {.pure.} = enum
```
tersec, line 27: Regarding adding exporting markers for `StartupCommand`, `LogTraceConf`, `simDir`, `netDir`, ` `ignoreSerializationErrors`, et cetera, what imports `logtrace`? There shouldn't be a need to export from a module which isn't imported.

beacon_chain/spec/forks.nim
```diff
@@ -297,19 +315,47 @@ template toFork*[T: eip4844.BeaconState | eip4844.HashedBeaconState](
...
+
+  func `as`*(fork: ForkType2, T: type ForkType1): T =
+    case fork
+    of ForkType2.Phase0: ForkType1.Phase0
+    of ForkType2.Altair: ForkType1.Altair
+    of ForkType2.Bellatrix: ForkType1.Bellatrix
+    of ForkType2.Capella: ForkType1.Capella
+    of ForkType2.EIP4844: ForkType1.EIP4844
+
+bidirectionalForkTypeMap BeaconStateFork, BeaconBlockFork
```
tersec: These are all equivalent

Makefile
```diff
@@ -202,59 +204,64 @@ restapi-test:
...
+		--nimbus-signer-nodes 1 \
+		--threshold 1 \
 		--enable-logtrace \
 		--base-port $$(( 6001 + EXECUTOR_NUMBER * 500 )) \
 		--base-rest-port $$(( 6031 + EXECUTOR_NUMBER * 500 )) \
 		--base-metrics-port $$(( 6061 + EXECUTOR_NUMBER * 500 )) \
 		--base-vc-keymanager-port $$(( 6131 + EXECUTOR_NUMBER * 500 )) \
 		--base-vc-metrics-port $$(( 6161 + EXECUTOR_NUMBER * 500 )) \
 		--base-remote-signer-port $$(( 6201 + EXECUTOR_NUMBER * 500 )) \
+		--base-remote-signer-metrics-port $$(( 6251 + EXECUTOR_NUMBER * 500 )) \
```
arnetheduck, line 227: seems like it's time for the script to perhaps simply have a single parameter to offset all these ports?

beacon_chain/spec/datatypes/bellatrix.nim
zah: Just like we don't need the ssz `List` type to be defined in every hard-fork datatypes module, we can define this wrapper type in a module that's imported similarly to the SSZ types.

beacon_chain/spec/beaconstate.nim
zah: I wanted to remove this function from the code altogether, but it's currently used in the test suite in few places. It proved to be too difficult to clean up the test suite as part of this PR, so I added this compile-time flag instead.

ncli/logtrace.nim
zah: This module is now imported and compiled as part of the `local_network_simulation` binary. A single binary for all simulation related functionality proved to lower the overall compilation time when running the simulation.

tersec:
```
2023-01-24T23:17:02.2538422Z /home/runner/work/nimbus-eth2/nimbus-eth2/ncli/local_testnet_simulation.nim: In function ‘doCreateTestnet__local95testnet95simulation_633’:
2023-01-24T23:17:02.2539347Z /home/runner/work/nimbus-eth2/nimbus-eth2/ncli/local_testnet_simulation.nim:324:15: error: stack usage might be 16981248 bytes [-Werror=stack-usage=]
2023-01-24T23:17:02.2539753Z   324 | proc doCreateTestnet*(config: CliConfig,
2023-01-24T23:17:02.2539974Z       |               ^
2023-01-24T23:17:02.2540420Z lto1: some warnings being treated as errors
```

beacon_chain/spec/datatypes/eip4844.nim
```diff
@@ -474,6 +475,10 @@ type
     parentHash*: string
     timestamp*: string
 
+# TODO: There should be only a single generic HashedBeaconState definition
+func initHashedBeaconState*(s: BeaconState): HashedBeaconState =
+  HashedBeaconState(data: s)
```
tersec, line 507: Do these fill in the cached htr?

beacon_chain/spec/datatypes/capella.nim
```diff
@@ -488,6 +489,10 @@ type
     bls_to_execution_changes*:
       List[SignedBLSToExecutionChange, Limit MAX_BLS_TO_EXECUTION_CHANGES]
 
+# TODO: There should be only a single generic HashedBeaconState definition
+func initHashedBeaconState*(s: BeaconState): HashedBeaconState =
```
tersec, line 493: Do these fill in the cached htr (also related to all the rest)?

tersec:
I agree with the `BeaconStateFork`/`BeaconBlockFork` to `ConsensusFork` changes, but they're also this big fragile rebase-requiring-inviting target right now -- might be good to split that off into its own PR and merge that on its own, so this PR doesn't keep getting hit by unrelated changes.

beacon_chain/conf.nim
```diff
@@ -221,12 +220,6 @@ type
       desc: "The slashing DB flavour to use"
       name: "slashing-db-kind" .}: SlashingDbKind
 
-    mergeAtEpoch* {.
```
tersec, line 230: Removal of this and all the related code to handle this means that non-merged post-Bellatrix simulations aren't supported at all, and thus Geth must be supported on all platforms -- this creates requirements to bootstrap and debug partial, potential consensus-only local testnets.

It is valid to run Bellatrix/Capella/Deneb networks, from a protocol standpoint, and currently the macos/aarch64 would/will need special Geth builds with their on infrastructure just to keep the local testnets working at all.

zah: I don't see high utility of running non-merged networks, especially in CI context (where the main complications of maintaining per-platform Geth builds are). For local testing, you should get comfortable with running Geth (the scripts are supposed to make this trivial). Not requiring it seems to me as a formula for establishing sloppy testing practices within the team.

arnetheduck: as long as we keep the ability to forward-sync from genesis...

beacon_chain/spec/beaconstate.nim
arnetheduck: I don't think we _should_ be removing it given it's part of the spec - ie I'd rather not see any `when` for this either, to avoid bitrot and unnecessary build complexity

beacon_chain/spec/datatypes/capella.nim
```diff
@@ -488,6 +489,10 @@ type
     bls_to_execution_changes*:
       List[SignedBLSToExecutionChange, Limit MAX_BLS_TO_EXECUTION_CHANGES]
 
+# TODO: There should be only a single generic HashedBeaconState definition
```
arnetheduck, line 492: unless we concretely plan to take action on these TODO's sometime in the near future, I think we're better off without them - otherwise, they're just noise since they don't represent a piece of work that anyone believes in strong enough to actually get it done

beacon_chain/spec/datatypes/capella.nim
arnetheduck: we've had a few bugs here and there where the cached htr is _not_ filled in correctly at various stages - if we have a function like this, it should probably do so (otherwise, what's the point of it vs simply initializing it directly?)

ncli/local_testnet_simulation.nim
```diff
@@ -0,0 +1,649 @@
+# beacon_chain
```
arnetheduck: shall we call this file `ncli_testnet`? ostensibly it should be possible to use it to create non-local testnets as well

beacon_chain/spec/beaconstate.nim
zah: We have been removing various phase0 spec bits from the codebase. If I have to spell out our policy, it could be "any code that's no longer used should be removed from the codebase".

arnetheduck: we implement the spec -> this function is part of the spec and tested -> we keep it, and not behind when flags (ie the compiler should do the dead-code-elimination for us, no need to do it manually

arnetheduck: the phase0 bits we've removed are those that have been deprecated / removed upstream (ie p2p protocol leftovers and rest)

scripts/launch_local_testnet.sh
```diff
@@ -634,28 +717,47 @@ for BINARY in ${BINARIES}; do
...
 
   for proc in "${PROCS_TO_KILL[@]}"
   do
-    pkill -f -P $$ "${proc}" || true
+    # TODO Previously, the code here used the '-P $$' option to limit
+    #      the kill command only to children of this shell process.
+    #      Unfortunately, this doesn't seem to work at the moment.
+    #      Perhaps the child processes are not direct children of
+    #      the current shell process?
+    pkill -f "$(basename "$proc")" || true
```
jakubgs, line 748: This change can potentially have big ramifications for CI runs. Are we sure the `-P` flag doesn't work?
Or does it not work just on certain platforms?

PR #4555, opened 2023-01-25
Fix epoch boundary calculation in BeaconBlockAndBlobsSidecarByRoot

henridf:
Previously, it wasnt taking into account the finalized epoch. Fix that to follow the [spec](https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.1/specs/eip4844/p2p-interface.md#beaconblockandblobssidecarbyroot-v1):

```
minimum_request_epoch = max(finalized_epoch, current_epoch - MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS, EIP4844_FORK_EPOCH)
```

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -439,11 +439,20 @@ p2pProtocol BeaconSync(version = 1,
     let
       dag = peer.networkState.dag
       count = blockRoots.len
+      # https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.1/specs/eip4844/p2p-interface.md#beaconblockandblobssidecarbyroot-v1
+      min_epochs =
+        if MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS > dag.head.slot.epoch:
+           dag.head.slot.epoch
+        else:
+          Epoch(MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS)
+
       epochBoundary =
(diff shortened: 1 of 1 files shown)


beacon_chain/sync/sync_protocol.nim
```diff
@@ -439,11 +439,20 @@ p2pProtocol BeaconSync(version = 1,
...
+      min_epochs =
+        if MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS > dag.head.slot.epoch:
+           dag.head.slot.epoch
+        else:
+          Epoch(MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS)
+
       epochBoundary =
-        if MIN_EPOCHS_FOR_BLOBS_SIDECARS_REQUESTS >= dag.head.slot.epoch:
-          GENESIS_EPOCH
+        if dag.head.slot.epoch - min_epochs < dag.cfg.EIP4844_FORK_EPOCH:
```
tersec, line 450: If `dag.cfg.EIP4844_FORK_EPOCH` is 0 or sufficiently small, this underflows -- these are unsigned ints, so it'll rather wrap.

henridf: I'm not seeing it - what could underflow? (`dag.head.slot.epoch - min_epochs` doesn't due to the check above)

tersec: Ah, indeed.

tersec (approved):

PR #4560, opened 2023-01-27
`BeaconStateFork/BeaconBlockFork` -> `ConsensusFork`

tersec:
Extracting mechanical changes from https://github.com/status-im/nimbus-eth2/pull/4551

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -508,6 +508,11 @@ OK: 4/4 Fail: 0/4 Skip: 0/4
 + [SyncQueue] hasEndGap() test                                                               OK
 ```
 OK: 23/23 Fail: 0/23 Skip: 0/23
+## Type helpers
+```diff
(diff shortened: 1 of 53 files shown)


etan-status (approved):
Looks good, and allows updating `stateFork` constant to `consensusFork` (comparable length to `lcDataFork`, which worked well, length-wise, re 80 col limit).

Probably good to merge before monday to avoid breaking things when everyone is back on poking this repo.

tersec:
I agree, but GitHub Actions CI is down at the moment. Yesterday, it ran fine, but the test summaries weren't updated, so GA CI was green and Jenkins red. Now the other way around. The only change was those summaries, so probably safe.

But would be nice to get GA CI actually-green. Not sure how feasible, since it's a fake-billing issue apparently, and restricted to Status or Nimbus repos.

Certainly, will aim for that though.

PR #4564, opened 2023-01-29
use v1.3.0-rc.2 consensus spec test vectors

tersec:
https://github.com/ethereum/consensus-spec-tests/releases/tag/v1.3.0-rc.2-hotfix are already out, but because there have been `MAX_PATH` issues on the Windows GitHub Actions CI runner before with those, and GitHub Actions was at last check refusing to run on this repository, only update to the non-hotfixed `rc.2` and exclude the hotfixed tests, to keep this reliably testable by just Linux and macOS Jenkins.


etan-status (approved):
LC related changes seem complete. Thanks!

PR #4565, opened 2023-01-30
track slot as part of fork choice debug API

etan-status:
Extends fork choice state to also track slot numbers to improve accuracy of `/eth/v1/debug/fork_choice` endpoint. Autoenable this API on devnet, and disable some extra checks on devnet to aid focused testing efforts. Align fork choice pruning logic with API based on checkpoints vs root.


tersec (approved):

tersec:
macos-amd64 failing due to same issue as in https://github.com/status-im/nimbus-eth2/pull/4569 -- seems to be/have been a Homebrew outage. It doesn't look like the GitHub Actions CI actually requires `gnu-getopt` brew, as that PR passes even without it, but perhaps it is required.

PR #4577, opened 2023-02-01
Eip4844 sync fixes

henridf:
With these fixes, local testnet finalizes when run as 
`./scripts/launch_local_testnet.sh --verbose --disable-vc --nodes 4 --disable-htop --stop-at-epoch 6 --light-clients 0 -- --verify-finalization --discv5:no`

with 
```
+CAPELLA_FORK_EPOCH: 3
+EIP4844_FORK_EPOCH: 4
``` 
added to the config.

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -201,6 +201,7 @@ proc processSignedBeaconBlock*(
     blockRoot = shortLog(signedBlock.root)
     blck = shortLog(signedBlock.message)
     signature = shortLog(signedBlock.signature)
+    hasBlobs = optBlobs(signedBlockAndBlobs).isSome
     wallSlot
 
   if not afterGenesis:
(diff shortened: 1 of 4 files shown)


henridf:
CI failure seems to be a backend/infra issue.

tersec (approved):

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -201,6 +201,7 @@ proc processSignedBeaconBlock*(
     blockRoot = shortLog(signedBlock.root)
     blck = shortLog(signedBlock.message)
     signature = shortLog(signedBlock.signature)
+    hasBlobs = optBlobs(signedBlockAndBlobs).isSome
```
tersec: This seems to construct an entire `Opt` to get a boolean out?

henridf: Yes, that was gratuitous. Addressed in 99a12ea

PR #4584, opened 2023-02-02
Update hardware.md

andytudhope:
Due to increased requirements when building beacon nodes.

Though this is annoying, I still think you all are wonderful and magical beings.

### DIFF
--- a/docs/the_nimbus_book/src/hardware.md
+++ b/docs/the_nimbus_book/src/hardware.md
@@ -5,7 +5,7 @@ The recommended system requirements for running the Nimbus beacon node are:
 | What       |  Recommended |
 | ---------- | ------ |
 | Operating system | Linux [64-bit](https://en.wikipedia.org/wiki/64-bit_computing), Windows 64-bit, macOS X 10.15+ |
-| Memory     | 4GB |
+| Memory     | 8GB |
 | Disk space | 200GB |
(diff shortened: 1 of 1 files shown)


tersec (approved):
Thanks for the PR

arnetheduck:
Hm, our requirements to *run* Nimbus haven't really changed this much - what's changed is that if you also want to *compile* on the same machine, it takes more space (ie swap space for example)

tersec:
Then this needs to be documented instead, but documenting 8GB is better than documenting 4GB and saying nothing about the build issues, when that's a supported workflow.

Also, I'm personally skeptical that running the compile using swap is going to be time-efficient, since it's a single process actively consuming up to 5.5GB RSS, not an issue of multiple processes some of which could be put aside.

It's probably possible, but since many cloud services also just don't provide swap by default, it likely makes more sense to document the actual memory requirements of building (~8GB, in powers of 2) than lean on swap.

It might make sense indeed to explicitly document distinction between build/run requirements.

arnetheduck:
> Then this needs to be documented instead, but documenting 8GB is better than documenting 4GB and saying nothing about the build issues, when that's a supported workflow.

sure - makes sense, though likely in a note - this is likely no longer the most common form of installation method. the alternative to a swap file is to build on a separate machine (which has its ups and downs)

PR #4590, opened 2023-02-03
fix slow checking of unknown validators

arnetheduck:
We do a linear scan of all pubkeys for each validator and slot - this becomes expensive with large validator counts.

* normalise BN/VC validator startup logging
* fix crash when host cannot be resolved while adding remote validator
* silence repeated log spam for unknown validators
* print pubkey/index/activation mapping on startup/validator identification

### DIFF
--- a/beacon_chain/validator_client/duties_service.nim
+++ b/beacon_chain/validator_client/duties_service.nim
@@ -89,9 +89,12 @@ proc pollForValidatorIndices*(vc: ValidatorClientRef) {.async.} =
   for item in validators:
     var validator = vc.attachedValidators[].getValidator(item.validator.pubkey)
     if isNil(validator):
+      validator.updateValidator(Opt.none ValidatorAndIndex)
       missing.add(validatorLog(item.validator.pubkey, item.index))
     else:
-      validator.updateValidator(item.index, item.validator.activation_epoch)
+      validator.updateValidator(Opt.some ValidatorAndIndex(
+        index: item.index,
+        validator: item.validator))
       updated.add(validatorLog(item.validator.pubkey, item.index))
       list.add(validator)
 
@@ -103,13 +106,6 @@ proc pollForValidatorIndices*(vc: ValidatorClientRef) {.async.} =
     trace "Validator indices update dump", missing_validators = missing,
           updated_validators = updated
     vc.indicesAvailable.fire()
-
-  if not vc.polledIndices:
-    vc.polledIndices = true
-    for validator in vc.attachedValidators[].items():
-      if validator.needsUpdate():
-        notice "Validator deposit not yet processed, monitoring",
-          pubkey = validator.pubkey
 
 proc pollForAttesterDuties*(vc: ValidatorClientRef,
                             epoch: Epoch): Future[int] {.async.} =
--- a/beacon_chain/validator_client/duties_service.nim
+++ b/beacon_chain/validator_client/duties_service.nim
@@ -96,11 +96,20 @@ proc pollForValidatorIndices*(vc: ValidatorClientRef) {.async.} =
       list.add(validator)
 
   if len(updated) > 0:
-    info "Validator indices updated", missing_validators = len(missing),
-         updated_validators = len(updated)
+    info "Validator indices updated",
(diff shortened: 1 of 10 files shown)


beacon_chain/validator_client/duties_service.nim
```diff
@@ -96,12 +96,21 @@ proc pollForValidatorIndices*(vc: ValidatorClientRef) {.async.} =
...
+      updated = len(updated)
     trace "Validator indices update dump", missing_validators = missing,
           updated_validators = updated
     vc.indicesAvailable.fire()
 
+  if not vc.polledIndices:
+    vc.polledIndices = true
+    for validator in vc.attachedValidators[].items():
+      if validator.needsUpdate():
+        notice "Validator deposit not yet processed, monitoring",
```
cheatfate: I thought we already trying to avoid logging every validator in logs, so maybe we should gather information about all non-processed validators and log all of them in one log line?

beacon_chain/validators/validator_duties.nim
```diff
@@ -121,18 +120,13 @@ proc addValidators*(node: BeaconNode) =
...
 
     if data.isSome:
       v.updateValidator(data.get().index, data.get().validator.activation_epoch)
-
-proc getAttachedValidator(node: BeaconNode,
-                          pubkey: ValidatorPubKey): AttachedValidator =
-  node.attachedValidators[].getValidator(pubkey)
+    else:
+      notice "Validator deposit not yet processed, monitoring",
+        pubkey = keystore.pubkey
```
cheatfate: One more log spam location found.

arnetheduck: this is fine since it happens once on startup - the objective here is to log the full pubkey of every validator once, so that the user can know exactly which validators are not yet activated (and check them if need be)

beacon_chain/validator_client/duties_service.nim
arnetheduck: yeah, agree but I couldn't find a good place to log the validators once per startup - we want to keep this as one validator per line so that the pubkey is clearly visible, but ideally it would happen once per startup.

Fixing this properly for the VC require more extensive changes (for example, we'd have to keep track of which validators have been "checked" or do a special request on startup, not sure what the best architecture is) - I didn't do it in this PR because the BN is much much worse in terms of spam: it logs every slot whereas the VC only logged once per epoch, so this PR fixes the most critical problem.

arnetheduck:
https://github.com/status-im/nimbus-eth2/pull/4590/commits/5348a6ff64b586a13c9374bd4af247da90969c03 ensures consistent logging between vc & bn, and avoids the log spam

beacon_chain/validators/keystore_management.nim
```diff
@@ -1324,15 +1304,15 @@ proc getSuggestedFeeRecipient*(
...
   let
-    data = host.getValidatorData(keystore.pubkey)
     feeRecipient = host.getSuggestedFeeRecipient(keystore.pubkey).valueOr(
       host.defaultFeeRecipient)
+    v = host.validatorPool[].addValidator(keystore, feeRecipient)
 
-  let v = host.validatorPool[].addLocalValidator(keystore, feeRecipient)
-  if data.isSome():
-    v.updateValidator(data.get().index, data.get().validator.activation_epoch)
+  if host.getValidatorAndIdxFn != nil:
```
cheatfate: You should avoid direct comparison with `nil`, you can use `not(isNil(host.getValidatorAndIdxFn))` instead.

tersec: What's the difference?

arnetheduck: 3d0ba684c64cad3ce773c12d16d8847d20b9151c - I don't care either way really - looking at the source code, `isNil` is a magic which gets used in some "not nil" analysis but it doesn't seem to practically matter which one we use, specially given that said analysis is likely to be buggy / broken - I'd probably move all our code to `== nil` and not use `isNil` ever until there's a clear and well-understood benefit (to avoid discussions like this)

zah (approved):

PR #4596, opened 2023-02-06
ignore systemctl failure in after_install scripts

jakubgs:
Otherwise installation in Docker containers fails with:
```
...
Adding new user `nimbus' (UID 101) with group `nimbus' ...
Not creating home directory `/home/nimbus'.
/var/lib/dpkg/info/nimbus-beacon-node.postinst: 39: systemctl: not found
dpkg: error processing package nimbus-beacon-node (--configure):
 installed nimbus-beacon-node package post-installation script subprocess returned error exit status 127
Errors were encountered while processing:
 nimbus-beacon-node
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

### DIFF
--- a/scripts/package_src/nimbus_beacon_node/after_install
+++ b/scripts/package_src/nimbus_beacon_node/after_install
@@ -19,4 +19,5 @@ fi
 mkdir -p /var/lib/nimbus
 chown nimbus:nimbus /var/lib/nimbus
 
-systemctl daemon-reload
+# Systems like docker containers do not have systemd.
+systemctl daemon-reload || echo "notice: systemd daemon not reloaded" >&2
--- a/scripts/package_src/nimbus_validator_client/after_install
+++ b/scripts/package_src/nimbus_validator_client/after_install
@@ -19,4 +19,5 @@ fi
 mkdir -p /var/lib/nimbus
 chown nimbus:nimbus /var/lib/nimbus
 
-systemctl daemon-reload
+# Systems like docker containers do not have systemd.
+systemctl daemon-reload || echo "notice: systemd daemon not reloaded" >&2


scripts/package_src/nimbus_beacon_node/after_install
```diff
@@ -19,4 +19,5 @@ fi
 mkdir -p /var/lib/nimbus
 chown nimbus:nimbus /var/lib/nimbus
 
-systemctl daemon-reload
+# Systems like docker containers do not have systemd.
+systemctl daemon-reload || true
```
arnetheduck: in theory we could check if we're running under systemd and only do the reload then - not sure if it's worth the bother?

arnetheduck (approved):

scripts/package_src/nimbus_beacon_node/after_install
jakubgs: We could just check if `systemctl` tool exists, that's one way, but just because it exists doesn't mean `systemctl daemon-reload` wouldn't fail, since even if we install it in a Docker container it still isn't running as PID 1, so it would fail with:
```
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```
Which means we'd also need to add an extra check to see if there PID 1 is a `systemd` process.
I guess it's more explicit, but it's a bunch of extra work for no real gain.

I guess I can improve it by using an `|| echo "Systemd reload failed!" >&2` for clarity.

scripts/package_src/nimbus_beacon_node/after_install
```diff
@@ -19,4 +19,5 @@ fi
 mkdir -p /var/lib/nimbus
 chown nimbus:nimbus /var/lib/nimbus
 
-systemctl daemon-reload
+# Systems like docker containers do not have systemd.
+systemctl daemon-reload || echo "warning: Systemd reload failed!" >&2
```
arnetheduck:
```suggestion
systemctl daemon-reload || echo "notice: systemd daemon not reloaded" >&2
```
this is harmless when it happens, `failed` = support requests

jakubgs: Sure.

PR #4597, opened 2023-02-06
drop dependency on lsb-release script

jakubgs:
The `/etc/os-release` file exists in most distributions and can be easily read in Bash by sourcing it:
```sh
 > docker run --rm -it debian:bullseye
root@2f5d6e038738:/# grep '^ID=' /etc/os-release
ID=debian
```
```sh
 > docker run --rm -it ubuntu:22.04
root@316b572b6e4d:/# grep '^ID=' /etc/os-release
ID=ubuntu
```
The dependency on `lsb-release` tool is unnecessary, and pulls in additional big dependencies like `python3`:
```sh
 # apt show lsb-release | grep Depends
Depends: python3:any, distro-info-data
```
Which if used in a Docker container would make it unnecessarily big.

### DIFF
--- a/scripts/package_src/nimbus_validator_client/after_install
+++ b/scripts/package_src/nimbus_validator_client/after_install
@@ -2,10 +2,15 @@
 
 set -e
 
-DISTRO=$(lsb_release -si)
+DISTRO="UNKNOWN"
+if [[ -r /etc/os-release ]]; then
+    . /etc/os-release
+    DISTRO="${ID}"
+fi
+
 if ! id -u nimbus > /dev/null 2>&1; then
   case $DISTRO in
-    Ubuntu|Debian)
+    Ubuntu|ubuntu|Debian|debian)
       # Debian uses `adduser` to create user...
       adduser --system --no-create-home --group nimbus
       ;;
--- a/scripts/make_packages.sh
+++ b/scripts/make_packages.sh
@@ -142,7 +142,6 @@ fpm -s dir -t deb -n "${PKG_NAME}" \
   -C "${PKG_IMG_DIR}" \
   -p "${PKG_PATH_DEB}" \
   -a "${PKG_ARCH_DEB}" \
-  --depends lsb-release \
   --after-install "${PKG_SRC_DIR}/after_install" \
   --before-remove "${PKG_SRC_DIR}/before_remove" \
   --after-remove "${PKG_SRC_DIR}/after_remove" \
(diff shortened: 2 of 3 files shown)


scripts/package_src/nimbus_validator_client/after_install
```diff
@@ -2,7 +2,12 @@
 
 set -e
 
-DISTRO=$(lsb_release -si)
+DISTRO="UNKNOWN"
+if [[ -r /etc/os-release ]]; then
+    . /etc/os-release
+    DISTRO="${DISTRIB_ID}"
```
arnetheduck:
```
[arnetheduck@praeceps nimbus-eth2]$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="37 (Workstation Edition)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

doesn't fly on fedora

jakubgs: Oh, I forgot to change the variable in both files, since I was trying `/etc/lsb-release` first.

jakubgs: Nice.
```sh
 > docker run --rm -it fedora:37
[root@5149d5e90f25 /]# [[ -r /etc/os-release ]] && source /etc/os-release; echo $ID
fedora
```

jakubgs: So easy to make a mistake when duplicating changes between identical files...
Maybe we could just symlink one to the other?

arnetheduck: I don't think that works since the symlink would be included in the package and thus become dangling

arnetheduck: the only way I imagine it working if there's a file that's copied to the "package folder" during creation

PR #4608, opened 2023-02-10
deprecate req/resp GetMetaData v1

tersec:
Should have been returning `InvalidRequest` already: https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.2/specs/altair/p2p-interface.md#transitioning-from-v1-to-v2

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -281,9 +281,10 @@ p2pProtocol BeaconSync(version = 1,
     {.libp2pProtocol("ping", 1, isRequired = true).} =
     return peer.network.metadata.seq_number
 
-  proc getMetaData(peer: Peer): phase0.MetaData
+  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.2/specs/altair/p2p-interface.md#transitioning-from-v1-to-v2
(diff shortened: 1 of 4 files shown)


beacon_chain/sync/sync_protocol.nim
```diff
@@ -281,9 +281,10 @@ p2pProtocol BeaconSync(version = 1,
     {.libp2pProtocol("ping", 1, isRequired = true).} =
     return peer.network.metadata.seq_number
 
-  proc getMetaData(peer: Peer): phase0.MetaData
+  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.2/specs/altair/p2p-interface.md#transitioning-from-v1-to-v2
+  proc getMetaData(peer: Peer): uint64
```
tersec, line 285: Not sure what the most elegant way to create an always-fails handler is. There is also the possibility of using the `response` pragma, but it seems basically the same: create a fake return type which won't be used. The macro very much prefers/requires having the user call handler return something non-`void`.

tersec: Another option is just to remove `GetMetaData v1` entirely, but the spec does say to return the `InvalidRequest`.

PR #4609, opened 2023-02-10
block on forkchoiceUpdated EL calls due to doing fewer of them

tersec:
In a steady-state synced situation, the CL needs to wait anyway for the EL to finish, so it was not truly asynchronous except during syncing.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -512,7 +512,13 @@ proc storeBlock*(
   #
   # Three general scenarios: (1) pre-merge; (2) merge, already `VALID` by way
   # of `newPayload`; (3) optimistically imported, need to call fcU before DAG
-  # updateHead. Handle each with as little async latency as feasible.
+  # updateHead. Because in a non-finalizing network, completing sync isn't as
+  # useful because regular reorgs likely still occur, and when finalizing the
+  # EL is only called every SLOTS_PER_PAYLOAD slots regardless, await, rather
+  # than asyncSpawn forkchoiceUpdated calls.
+  #
+  # This reduces in-flight fcU spam, which both reduces EL load and decreases
+  # otherwise somewhat unpredictable CL head movement.
 
   if payloadValid:
     self.consensusManager.dag.markBlockVerified(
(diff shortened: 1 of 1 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -543,7 +549,7 @@ proc storeBlock*(
       # - "Beacon chain gapped" from DAG head to optimistic head,
       # - followed by "Beacon chain reorged" from optimistic head back to DAG.
       self.consensusManager[].updateHead(newHead.get.blck)
-      asyncSpawn eth1Monitor.runForkchoiceUpdatedDiscardResult(
+      await eth1Monitor.runForkchoiceUpdatedDiscardResult(
```
arnetheduck, line 552: refactor to put the discard here?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -572,10 +578,10 @@ proc storeBlock*(
           # Some attached validator is next proposer, so prepare payload. As
           # updateHead() updated the DAG head, runProposalForkchoiceUpdated,
           # which needs the state corresponding to that head block, can run.
-          asyncSpawn self.consensusManager.runProposalForkchoiceUpdated(
+          await self.consensusManager.runProposalForkchoiceUpdated(
```
arnetheduck, line 581: it's kind of weird that we sometimes run fcu through consensusmanager and sometimes through eth1monitor - can't we collect all FCU calls in consensus manager? also, can't we move the proposer-or-not logic there?

arnetheduck (approved):
soo... this PR looks like it might solve the immediate problem of concurrent / stale head updates, but it also feels like the code could be simplified further with a round of refactoring (which we can leave for another PR) - ie it's starting to get quite hairy, to understand what sets which head where and why

arnetheduck:
merging for now since sync is completely broken by the concurrent head updates when spawned tasks get reordered

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4465 already does this in `elManager.forkchoiceUpdated`

tersec:
> soo... this PR looks like it might solve the immediate problem of concurrent / stale head updates, but it also feels like the code could be simplified further with a round of refactoring (which we can leave for another PR) - ie it's starting to get quite hairy, to understand what sets which head where and why

There is potentially other cleanup from cases to consolidate -- previously, one distinction was exactly whether to `asyncSpawn` or `await`. If it always `awaits`, there is less reason to delineate cases so finely.

beacon_chain/gossip_processing/block_processor.nim
tersec: `runForkchoiceUpdatedDiscardResult` is one such example of a now-too-fine delineation, yes which was there only to support some cases being `asyncSpawn`, indeed, for example.

tersec: https://github.com/status-im/nimbus-eth2/pull/4614

Issue #4574, opened 2023-01-31
Remote keymanager CI flaky


tersec:
Usually on Windows GitHub Actions.

One example,
```
023-01-30T19:07:05.3408292Z [Suite] ImportRemoteKeys/ListRemoteKeys/DeleteRemoteKeys [Beacon Node] [Preset: mainnet]
2023-01-30T19:07:05.3679038Z     D:\a\nimbus-eth2\nimbus-eth2\tests\test_keymanager_api.nim(1099, 54): Check failed: responseJson1["data"][i]["status"].getStr() == expectedImportStatus(i)
2023-01-30T19:07:05.3680097Z     responseJson1["data"][i]["status"].getStr() was error
2023-01-30T19:07:05.3680546Z     expectedImportStatus(i) was imported
2023-01-30T19:07:05.3681358Z     D:\a\nimbus-eth2\nimbus-eth2\tests\test_keymanager_api.nim(1100, 55): Check failed: responseJson1["data"][i]["message"].getStr() == ""
2023-01-30T19:07:05.3681998Z     responseJson1["data"][i]["message"].getStr() was (kind: FailedToCreateKeystoreDir, error: "1336")
2023-01-30T19:07:05.3682587Z     D:\a\nimbus-eth2\nimbus-eth2\tests\test_keymanager_api.nim(1122, 20): Check failed: found == true
2023-01-30T19:07:05.3683310Z     found was false
2023-01-30T19:07:05.3683713Z     true was true
2023-01-30T19:07:05.3684230Z     D:\a\nimbus-eth2\nimbus-eth2\tests\test_keymanager_api.nim(1132, 52): Check failed: responseJson2["data"][0]["status"].getStr() == "deleted"
2023-01-30T19:07:05.3684737Z     responseJson2["data"][0]["status"].getStr() was not_found
2023-01-30T19:07:05.3685142Z   [FAILED] Importing list of remote keys [Beacon Node] [Preset: mainnet]
2023-01-30T19:07:05.3685676Z   [OK] Missing Authorization header [Beacon Node] [Preset: mainnet]
2023-01-30T19:07:05.3686149Z   [OK] Invalid Authorization Header [Beacon Node] [Preset: mainnet]
2023-01-30T19:07:05.3686625Z   [OK] Invalid Authorization Token [Beacon Node] [Preset: mainnet]
```
https://github.com/status-im/nimbus-eth2/actions/runs/4042382726/jobs/6957498698

[...]

PR #4612, opened 2023-02-13
Implement the set of gas_limit end-points in the Keymanager API (Fixes #3946)

### LINKED ISSUE #3946: Feature: Implement keymanager API `gas_limit`

Implement the `gas_limit` API as per https://github.com/ethereum/keymanager-APIs/blob/master/apis/gas_limit.yaml and https://ethereum.github.io/keymanager-APIs/#/Gas%20Limit

zah:


beacon_chain/validators/keystore_management.nim
```diff
@@ -738,7 +746,34 @@ proc getSuggestedFeeRecipient*(
...
+    return ok defaultGasLimit
+  try:
+    ok parseBiggestUInt(strutils.strip(
+      readFile(gasLimitPath), leading = false, trailing = true))
+  except SerializationError as e:
+    warn "Invalid local gas limit file", gasLimitPath,
+      err= e.formatMsg(gasLimitPath)
+    err malformedConfigFile
+  except CatchableError as exc:
+    warn "getSuggestedGasLimit: failed loading gas limit file; falling back to default gas limit",
```
arnetheduck: no weird log prefixes on user-targeted logs

arnetheduck: could log the default limit too btw

zah: Both suggestions were applied here and in `getFeeRecipient` where the same pattern existed:

3ec87f861e6bd1f48f99decf1df5379714cf13f9

tests/test_keymanager_api.nim
```diff
@@ -106,6 +107,9 @@ const
 func specifiedFeeRecipient(x: int): Eth1Address =
   copyMem(addr result, unsafeAddr x, sizeof x)
 
+proc specifiedGasLimit(): uint64 =
```
tersec: Also, regarding the `random` import.

(1) Even if using randomness, using the random import means one has to ensure the PRNG setup is however one desires it not just for the CSPRNG but for the Nim PRNG. It is cleaner to only have to maintain one such environment.

(2) In general, it is nicer to have these tests as deterministic as possible. While this would in general usually mitigate against a CSPRNG and in favor of `std/random`, which in theory is deterministic depending on seed -- again, this depends on the environment setup having been correct, in essentially a global variable way, difficult to detect across a test runner and dependent on details of which tests end up aggregated in the same binary or process.

(3) So, better would be to generate some set or sequence of random numbers to use as gas limits to test with (or not-so-random, for testing boundary conditions, et cetera, either way) and have those be deterministic every time.

The keymanager tests already have more stochasticity in them them than ideal (https://github.com/status-im/nimbus-eth2/issues/4574, https://github.com/status-im/nimbus-eth2/issues/4619); adding more will not necessarily help.

beacon_chain/rpc/rest_key_management_api.nim
```diff
@@ -390,6 +390,73 @@ proc installKeymanagerHandlers*(router: var RestRouter, host: KeymanagerHost) =
...
+        pubkey: pubkey,
+        gas_limit: gasLimit.get))
+    else:
+      case gasLimit.error
+      of noSuchValidator:
+        keymanagerApiError(Http404, "No matching validator found")
+      of malformedConfigFile:
+        keymanagerApiError(Http500, "Error reading gas limit file")
+
+  # https://ethereum.github.io/keymanager-APIs/#/Gas%20Limit/setGasLimit
```
tersec, line 416: The REST server code has this issue as well, but it would be nice not to spread it further: because it mostly links to the live/evergreen/permalink/user-friendly form of the spec URLs, we have several times missed that the spec got updated, some new field or parameter, got added, et cetera.

Using a link such as https://github.com/ethereum/keymanager-APIs/blob/v1.0.0/apis/gas_limit.yaml here in addition to the evergreen link would not magically prevent this, but it would clarify that this was implemented against a specific version of the spec, and if there is an updated spec version, then it might be worth checking for updates.

I am aware this creates code churn which sometimes seems a bit superfluous, but it also makes very clear what has and has not been checked for the latest specs, and the alternative is probably worse.

(Same comment applies to https://ethereum.github.io/keymanager-APIs/#/Gas%20Limit/getGasLimit link)

zah: Since the rest of the REST API modules are not following the proposed standard, this sounds like a separate stand-alone PR that fixes all modules at once.

tests/test_keymanager_api.nim
zah: The import of `random` was removed during the rebase.

beacon_chain/rpc/rest_key_management_api.nim
tersec: The premise of those is that one has verified these at some particular time -- it was feasible exactly because I introduced those as I was writing them, with knowledge of the specs handy. Going through all the APIs is feasible, but also inefficient, when probably, at any given point, most are fine.

One purpose of these is to provide a kind of shortcut, exactly not to require verification of the mapping of code <-> spec for thousands of lines of code at once.

Right now, you are implementing what presumably is meant to be the current version, which satisfies the condition above. It therefore does not require extra effort to later, separately, verify what should be verified as simply part of this PR process. I do not see the need for consistency among all the REST server functions in this regard, and requiring that renders achieving this quite a bit less worthwhile.

So, do this or do not do this, but it will (probably) not happen by your proposed alternate mechanism.

zah: What I don't understand is why you are placing the Keymanager API is a separate category. The Keymanager API has the same backward/forward compatibility policy as the BeaconAPI. It has been extended with new end-points, but the probability of changes in existing end-points should be similar to the Beacon API.

tersec: I'm not putting it in a separate category. The Beacon APIs should be changed too, and when I have updated them, I have added these per-version URLs. However, for the reasons outlined, I have not (and do not intend to) gone back and retroactively check any of them I'm not actively updating.

PR #4616, opened 2023-02-13
restore doppelganger check on connectivity loss

arnetheduck:
https://github.com/status-im/nimbus-eth2/pull/4398 introduced a regression in functionality where doppelganger detection would not be rerun during connectivity loss. This PR reintroduces this check and makes some adjustments to the implementation to simplify the code flow for both BN and VC.

* track when check was last performed for each validator (to deal with late-added validators)
* track when we performed a doppel-detectable activity (attesting) so as to avoid false positives
* remove nodeStart special case (this should be treated the same as adding a validator dynamically just after startup)
* skip doppelganger check for sync committee messages


beacon_chain/validator_client/common.nim
```diff
@@ -482,8 +482,10 @@ proc getDelay*(vc: ValidatorClientRef, deadline: BeaconTime): TimeDiff =
   vc.beaconClock.now() - deadline
 
 proc getValidatorForDuties*(vc: ValidatorClientRef,
-                            key: ValidatorPubKey, slot: Slot): Opt[AttachedValidator] =
-  vc.attachedValidators[].getValidatorForDuties(key, slot)
+                            key: ValidatorPubKey, slot: Slot,
+                            doppelActivity = false,
+                            syncCommitteeDuty = false): Opt[AttachedValidator] =
```
zah: `syncCommitteeDuty` is an unused parameter?

arnetheduck: ca7d323a6cd36540925eed477bdf0a1c3d9950b2 - good catch!

PR #4629, opened 2023-02-15
Rename eip4844->deneb

henridf:
still a draft as i need to wait for a [consensus-spec-tests](https://github.com/ethereum/consensus-spec-tests) release with the rename.


tersec:
This PR mixes several things, which are in principle separable:

1) semi-descriptive function name changes such as `maybeUpgradeStateToEIP4844` -> `maybeUpgradeStateToDeneb` and `upgrade_to_eip4844` to `upgrade_to_deneb`. Presumably the latter rename will, indeed, turn out to be correct (or maybe already has, in the `master` branch of `consensus-specs`).

These are nice, but are not tied to the rest in any particular way.

2) Fallout from renaming the module `beacon_chain/spec/datatypes/eip4844` to `beacon_chain/spec/datatypes/deneb`. This will have to be done, but it does not have to be done all at once. A good first step is to rename/move `beacon_chain/spec/datatypes/eip4844.nim` to `beacon_chain/spec/datatypes/deneb.nim`, and then set up a new `beacon_chain/spec/datatypes/eip4844.nim` as a forwarding module:
```nim
import ./deneb
export deneb
```

And... stop there, at first. If it does not work, seamlessly, something is already broken and needs to be investigated. Once that works, can change things in a more ongoing way. Can still do a mass replace if appropriate, but do not have to. Regardless of the specific logistics and timing, and no point is there any rush, or critical period or flag day. Eventually, the forwarding module can be removed, after the last reference `eip4844` module reference is gone, but until then it functions fine.

3) `ConsensusFork.EIP4844` to `ConsensusFork.Deneb`: this, alas, is basically necessary to do in one single commit, but also, does not have to happen in the same commit or PR as any of these others.

4) Other smaller module renames, e.g., `eip4844_preset` to `deneb_preset` for minimal/mainnet: small enough it probably does not make sense to stage them across multiple commits, but also, do not have to be done at the same time as any of the rest.

5) Some just regressions:
```nim
# https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.2/specs/eip4844/p2p-interface.md#topics-and-messages
# https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#topics-and-messages
```
and similar.

`dev` says nothing about when it was last checked. If the spec is not ready for these URLs to use `deneb`, it is not ready for these URLs to use `deneb`. That's fine. The URLs are yet another factor to migrate separately (and not necessarily all at once, though there are few enough `/eip4844/` URLs it might be the best option).

Realistically, this one makes sense only after the next tagged release, presumably `v1.3.0-rc.3`, comes out. In the meantime, it adds noise to this PR.

6) `eip4844Data` to `denebData`: another probably-has-to-happen-in-one-shot one, but does not have to happen with any of the rest of these.

7) Random comments, e.g.,
```
- EIP4844: ~230 KB per `SyncCommitteePeriod` (~6.3 MB per month)
# - Deneb: ~230 KB per `SyncCommitteePeriod` (~6.3 MB per month)
```
Yes, good to keep comments/documentation updated, and likewise separable.

The handy thing about EIP4844 in all its capitalizations (and EIP-4844 and some other variations) is that they are all easy to search and find stray ones later. No need to get them all at once.

There are probably more subdivisions, but in general, would not want to merge this PR as-is and it would likely require some otherwise unnecessary coordination to ensure other PRs do not land in the meantime, since it touches so much.

tersec:
https://github.com/status-im/nimbus-eth2/pull/4628 is (timing was unintentional, but convenient for this purpose), an example of a less disruptive and more carefully scoped approach.

PR #4630, opened 2023-02-15
indicate Capella / Shanghai readiness on startup

etan-status:
In startup "Launching beacon node" message, log readiness for Capella / Shanghai as part of config to aid with support requests.


etan-status:
(Delay this one until mainnet is ready)

PR #4635, opened 2023-02-16
VC: Use not-synced/opt-synced BNs.

cheatfate:
* Optimize checkNode to check only specific BN statuses.
* Add ability to use specific BN status nodes for specific API calls.
* Allow usage of opt-synced and not-synced nodes for all API calls.

### DIFF
--- a/beacon_chain/validator_client/block_service.nim
+++ b/beacon_chain/validator_client/block_service.nim
@@ -39,7 +39,7 @@ proc produceBlock(
         await vc.produceBlockV2(slot, randao_reveal, graffiti,
                                 ApiStrategyKind.Best)
       except ValidatorApiError:
-        error "Unable to retrieve block data"
+        error "Unable to retrieve block data", reason = vc.getFailureReason()
         return Opt.none(PreparedBeaconBlock)
       except CancelledError as exc:
         error "Block data production has been interrupted"
@@ -69,7 +69,8 @@ proc produceBlindedBlock(
         await vc.produceBlindedBlock(slot, randao_reveal, graffiti,
                                      ApiStrategyKind.Best)
       except ValidatorApiError as exc:
-        error "Unable to retrieve blinded block data", error_msg = exc.msg
+        error "Unable to retrieve blinded block data", error_msg = exc.msg,
+              reason = vc.getFailureReason()
         return Opt.none(PreparedBlindedBeaconBlock)
       except CancelledError as exc:
         error "Blinded block data production has been interrupted"
@@ -215,7 +216,8 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
             debug "Sending blinded block"
             await vc.publishBlindedBlock(signedBlock, ApiStrategyKind.First)
           except ValidatorApiError:
-            error "Unable to publish blinded block"
+            error "Unable to publish blinded block",
+                  reason = vc.getFailureReason()
             return
           except CancelledError as exc:
             debug "Blinded block publication has been interrupted"
@@ -276,7 +278,7 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
             debug "Sending block"
             await vc.publishBlock(signedBlock, ApiStrategyKind.First)
           except ValidatorApiError:
-            error "Unable to publish block"
+            error "Unable to publish block", reason = vc.getFailureReason()
             return
           except CancelledError as exc:
             debug "Block publication has been interrupted"
--- a/beacon_chain/validator_client/block_service.nim
+++ b/beacon_chain/validator_client/block_service.nim
@@ -38,8 +38,8 @@ proc produceBlock(
       try:
         await vc.produceBlockV2(slot, randao_reveal, graffiti,
                                 ApiStrategyKind.Best)
-      except ValidatorApiError:
-        error "Unable to retrieve block data", reason = vc.getFailureReason()
+      except ValidatorApiError as exc:
+        error "Unable to retrieve block data", reason = vc.getFailureReason(exc)
         return Opt.none(PreparedBeaconBlock)
       except CancelledError as exc:
         error "Block data production has been interrupted"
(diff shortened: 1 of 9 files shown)


arnetheduck:
Tried running this, but there are still several concerning logs being output:

```
ERR 2023-02-20 09:34:36.083+01:00 Unable to get attester duties              epoch=157224 service=duties_service
INF 2023-02-20 09:34:38.166+01:00 Beacon node has compatible configuration   service=fallback_service node=localhost:5152[Nimbus/v23.2.0-4b9614-stateofus] node_index=0 node_roles=AGBSD
WRN 2023-02-20 09:34:38.208+01:00 Execution client not in sync (beacon node optimistically synced) sync_distance=1 head_slot=5031172 is_optimistic=true service=fallback_service node=localhost:5152[Nimbus/v23.2.0-4b9614-stateofus] node_index=0 node_roles=AGBSD
ERR 2023-02-20 09:34:38.291+01:00 Unable to get attester duties              epoch=157225 service=duties_service
INF 2023-02-20 09:34:40.293+01:00 Beacon node has compatible configuration   service=fallback_service node=localhost:5152[Nimbus/v23.2.0-4b9614-stateofus] node_index=0 node_roles=AGBSD
WRN 2023-02-20 09:34:40.333+01:00 Execution client not in sync (beacon node optimistically synced) sync_distance=1 head_slot=5031172 is_optimistic=true service=fallback_service node=localhost:5152[Nimbus/v23.2.0-4b9614-stateofus] node_index=0 node_roles=AGBSD
ERR 2023-02-20 09:34:40.377+01:00 Could not obtain head block's root because beacon node only optimistically synced slot=5031173 service=sync_committee_service
INF 2023-02-20 09:34:48.000+01:00 Slot start                                 slot=5031174 attestationIn=<unknown> blockIn=<unknown> validators=32 delay=865us925ns
ERR 2023-02-20 09:34:48.068+01:00 Unable to get attester duties              epoch=157224 service=duties_service
```

* `Unable to get attester duties` - we should be able to get duties from optimistically synced node, ie this should not be an error
* `Beacon node has compatible configuration` - this keeps repeating although it didn't change from the last call
* `Execution client not in sync` - we don't really need to print this repeatedly (ie once is enough per state "change", specially not at the warning level - instead, it should be part of the periodic slot start information
* `Could not obtain head block's root because beacon node only optimistically synced` - this should be a `NOT` saying `Execution client not in sync; skipping validator duties for now`, same as when the validator is running inside beacon node

arnetheduck:
Much nicer! 

```
INF 2023-02-21 14:54:12.000+01:00 Slot start                                 slot=5039971 attestationIn=<unknown> blockIn=<unknown> validators=32 node_status=el-unsynced delay=765us370ns
NOT 2023-02-21 14:54:12.063+01:00 Unable to get proposer duties              slot=5039971 epoch=157499 service=duties_service
NOT 2023-02-21 14:54:12.064+01:00 Unable to get attester duties              epoch=157499 service=duties_service
INF 2023-02-21 14:54:13.781+01:00 Execution client not in sync (beacon node optimistically synced) sync_distance=0 head_slot=5039971 is_optimistic=true service=fallback_service node=localhost:5152 node_index=0 node_roles=AGBSD
NOT 2023-02-21 14:54:13.863+01:00 Unable to get attester duties              epoch=157500 service=duties_service
INF 2023-02-21 14:54:15.906+01:00 Execution client not in sync (beacon node optimistically synced) sync_distance=0 head_slot=5039971 is_optimistic=true service=fallback_service node=localhost:5152 node_index=0 node_roles=AGBSD
NOT 2023-02-21 14:54:16.002+01:00 Execution client not in sync; skipping validator duties for now slot=5039971 service=sync_committee_service

```

Some issues remain:

* the beacon node is not responding to duties requests when optimistically synced (this is a problem in the REST server) - it responds with 503 instead of setting `execution_optimistic`
* the validator client is logging `Execution client not in sync; skipping validator duties for now slot=5039971 service=sync_committee_service` even though there are no sync committee duties - this is redundant and caused by trying to fetch the head root _before_ checking if there are duties: https://github.com/status-im/nimbus-eth2/blob/79eddcde4030f28bc4c80d10bbd46a637068e907/beacon_chain/validator_client/sync_committee_service.nim#L410
* logging for the failed duties attempts do not hint at error - debug logs show the underlying root cause (`DBG 2023-02-21 15:13:00.089+01:00 Failed to deserialize REST JSON data       err="<data>(1, 23) Not all required fields were specified when reading 'GetAttesterDutiesResponse'" data="{\"code\":503,\"message\":\"Beacon node is currently syncing and not serving request on that endpoint\"}"`) - ideally some of this information would be transported up to the `Unable to get attester duties` log - ie we know that the request failed because all nodes are out of sync (503) - in the case of multiple beacon nodes, we can still so we can show the "best" error

These issues can be addressed here or separately

arnetheduck (approved):

beacon_chain/validator_client/block_service.nim
```diff
@@ -276,7 +278,7 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
             debug "Sending block"
             await vc.publishBlock(signedBlock, ApiStrategyKind.First)
           except ValidatorApiError:
-            error "Unable to publish block"
+            error "Unable to publish block", reason = vc.getFailureReason()
```
arnetheduck: rather than using state in the vc here, it would be better to have this information in the `ValidatorApiError` (as a custom field)

arnetheduck: ie by the time the exception is handled by the async pipeline, the state in the vc will have changed - this style of logging introduces a race condition

beacon_chain/validator_client/common.nim
```diff
@@ -223,6 +240,50 @@ proc `$`*(roles: set[BeaconNodeRole]): string =
...
+  of ApiFailure.NotSynced: "Beacon node not in sync with network"
+  of ApiFailure.Internal: "Beacon node reports internal failure"
+  of ApiFailure.Unexpected: "Beacon node reports unexpected status"
+
+proc getNodeCounts*(vc: ValidatorClientRef): BeaconNodesCounters =
+  var res = BeaconNodesCounters()
+  for node in vc.beaconNodes: inc(res.data[int(node.status)])
+  res
+
+proc getFailureReason*(vc: ValidatorClientRef,
```
arnetheduck: `vc` unusued

arnetheduck (approved):

beacon_chain/validator_client/common.nim
cheatfate: Its for the future requests, i thought you will not like my new approach.

arnetheduck: approach looks ok, but why do we need to pass in vc here? the exception should be enough?

cheatfate: `vc` argument has been removed.

PR #4636, opened 2023-02-16
datatypes/eip4844.nim becomes datatypes/deneb.nim forwarding module

tersec:
Not sure why Git/GitHub isn't properly parsing this out, but literally:
```
$ git mv eip4844.nim deneb.nim
$ cat > eip4844.nim
import ./deneb
export deneb
$ git add eip4844.nim
```
This renames `eip4844.nim` to `deneb.nim`, with no changes at all to its contents.


arnetheduck:
> Not sure why Git/GitHub isn't properly parsing this out, but literally:

because git doesn't store actual renames, just takes a guess.. for a "clean" rename, one would likely have to perform the rename fully and remove the old module (which seems reasonable anyway, ie this eip4844 import/export looks overengineered.

tersec:
`eip4844` is imported now in at least a couple dozen modules -- this enables the next steps to be less all-or-nothing.

In particular, when imported, they regularly also contain a lot of `eip4844.FooSymbol` in them.

This is not happening all at once.

arnetheduck:
> be less all-or-nothing.

sure .. or one can just do an `all` since in this case it's a nice atomic change _until_ we have a deneb module

tersec:
Not sure I understand. Of course, one could just copy and not move `eip4844.nim` to `deneb.nim` and start building on that -- but then any changes, indefinitely, need to go synchronized into both. That seems worse than what this PR does.

It's not about the `deneb` module which indeed doesn't exist. It's about avoiding either this complete duplication (which creates a basically pointless pressure to finish things ASAP to remove that duplication/maintenance), or a the requirement to create a single PR of hundreds of lines which is admittedly atomic but cannot be adjusted for circumstances at all (e.g., other PRs to avoid conflicting with).

I've done things before where there's basically no PR possible until everything is in place, where no usable intermediate result exists, and if at all possible, find those worth avoiding.

But, sure. It is possible. Just, not, to my mind, advisable, and since it's in no way urgent, I won't prioritize that kind of semi-predictable, thus-far-mostly-cosmetic time sink.

PR #4637, opened 2023-02-17
batch-verify BLS to execution change messages

tersec:


beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -456,3 +456,39 @@ proc scheduleContributionChecks*(
         contribution.beacon_block_root, contributionKey, contributionSig)
 
   ok((aggregatorFut, proofFut, contributionFut, contributionSig))
+
+proc scheduleBlsToExecutionChangeCheck*(
+      batchCrypto: ref BatchCrypto,
+      genesisFork: Fork, genesis_validators_root: Eth2Digest,
```
arnetheduck, line 462: this PR makes me think we should move gf and gvr to BatchCrypto.init..

arnetheduck (approved):

beacon_chain/gossip_processing/batch_validation.nim
tersec: Not sure gf is broadly used enough to warrant it. It is used here, but a single use might not be enough for that to be optimal. It's quite a special case in the spec explicitly there to avoid fork-dependency, whereas typically, the messages are not meant to be timeless.

gvr, I would agree, but it would also be more a cross-cutting PR that should be done separately, exactly because it's used by `scheduleSyncCommitteeMessageCheck`, `scheduleAttestationCheck`, et cetera.

tersec: https://github.com/status-im/nimbus-eth2/pull/4640

PR #4644, opened 2023-02-20
docs: add apt repo

arnetheduck:


docs/the_nimbus_book/src/keep-updated.md
```diff
@@ -5,44 +5,50 @@ Make sure you stay on the lookout for any critical updates to Nimbus. The best w
...
-rm nimbus-eth2_Linux_arm64v8*.tar.gz
-```
+    ```sh
+    wget <insert download link here>
+    tar -xzf nimbus-eth2_Linux_arm64v8*.tar.gz --strip-components 1 -C nimbus-eth2
+    rm nimbus-eth2_Linux_arm64v8*.tar.gz
+    ```
 
-## Build from source
+=== "Debain / Ubuntu"
```
tersec:
```suggestion
=== "Debian / Ubuntu"
```

docs/the_nimbus_book/src/keep-updated.md
```diff
@@ -59,16 +65,32 @@ As of `v1.4.0`, releases are marked with the following tags:
 
 *Occassionally you may need to either upgrade or downgrade to a specific version of Nimbus.*
 
-To pull a specific version of Nimbus (e.g. `v22.9.1`), run:
-```sh
-git checkout v22.9.1 && make update
-```
+Nimbus can safely be downgraded to any version targeting the current hard fork of the chain, unless otherwise noted in the release notes between the version you're currently running and the version you're installing.
```
tersec: This is a little ambiguous -- if there are multiple releases across which people are updating, there are multiple sets of release notes. One can, I suppose, say that they, collectively, are release notes as this suggests, but one also sees "release notes" used in a singular way, to refer to a specific set of release notes for a specific release.

Which is this committing to, that people only have to look at the release notes of the version they're updating to, or all the versions in between?

docs/the_nimbus_book/src/keep-updated.md
```diff
@@ -59,16 +65,32 @@ As of `v1.4.0`, releases are marked with the following tags:
...
+
+    Use the package manager to install a specific version:
+
+    ```sh
+    sudo apt-get install nimbus-beacon-node=23.2.0
+    ```
+
+=== "Build from source"
+
+    To pull a specific version of Nimbus (e.g. `v22.9.1`), run:
```
tersec, line 84: Of course any individual version number set here will become outdated, and there's not much reason to necessarily update it each release cycle, but maybe at least use the current (`v23.2.0`) release to start with?

docs/the_nimbus_book/src/keep-updated.md
```diff
@@ -5,44 +5,50 @@ Make sure you stay on the lookout for any critical updates to Nimbus. The best w
...
+    ```
 
-## Build from source
+=== "Debain / Ubuntu"
 
-Upgrading Nimbus when built from source is similar to the installation process.
+    Update Nimbus via the package manager as usual
+    ```sh
+    sudo apt-get update
+    sudo apt-get upgrade
```
tersec: It might be slightly nicer to use a single `sudo apt-get update && sudo apt-get upgrade` here, because of the early exit if the `update` fails, in which case the `upgrade` isn't useful to run, and incurs slight (but still, in that situation pointless) risk.

docs/the_nimbus_book/src/keep-updated.md
zah: Well, the answer is "all versions in between", but how would you word that?

docs/the_nimbus_book/src/keep-updated.md
zah: A general problem of the `mkbook` tool is there is no notion of variables. We can potentially solve this by introducing some kind of template engine (e.g. [python's jinja](https://jinja.palletsprojects.com/en/3.1.x/)) as a preprocessing step in the book-generation scripts.

tersec: Sure, but also, regardless of that more automated ability to insert a current version, when there's manual editing going on affecting said version numbers directly, seems worth updating them.

docs/the_nimbus_book/src/keep-updated.md
tersec: > Nimbus can safely be downgraded to any version targeting the current hard fork of the chain, unless otherwise noted among the release notes between the version you're currently running and the version you're installing.

One simple change -- "in" -> "among". Works alright regardless of number of release notes.

docs/the_nimbus_book/src/keep-updated.md
arnetheduck: 6bde6ac5034705d816e45d9cad48036bef058dab

docs/the_nimbus_book/src/keep-updated.md
arnetheduck: I don't see that it really matters which version we use - ie typically the user will be downgrading to the _previous_ version they were running before their upgrade, which we don't know.

docs/the_nimbus_book/src/keep-updated.md
arnetheduck: 6bde6ac5034705d816e45d9cad48036bef058dab - removed the text altogether since it's complex

PR #4645, opened 2023-02-21
use consensus spec v1.3.0-rc.3 test vectors

tersec:
https://github.com/ethereum/consensus-specs/releases/tag/v1.3.0-rc.3
https://github.com/ethereum/consensus-spec-tests/releases/tag/v1.3.0-rc.3

Mostly minimal changes included, though it doesn't implement https://github.com/ethereum/consensus-specs/pull/3244 in any meaningful way, just adds the SSZ objects to pass the tests.

### DIFF
--- a/beacon_chain/spec/datatypes/eip4844.nim
+++ b/beacon_chain/spec/datatypes/eip4844.nim
@@ -50,8 +50,13 @@ type
   # current spec doesn't ever SSZ-serialize it or hash_tree_root it
   VersionedHash* = array[32, byte]
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.3/specs/deneb/beacon-chain.md#custom-types
+  BlobIndex* = uint64
+
   Blob* = array[BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB, byte]
 
+  # TODO remove BlobsSidecar and SignedBeaconBlockAndBlobsSidecar; they'e not
+  # in rc.3 anymore
   BlobsSidecar* = object
     beacon_block_root*: Eth2Digest
     beacon_block_slot*: Slot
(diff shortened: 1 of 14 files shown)


beacon_chain/spec/datatypes/eip4844.nim
```diff
@@ -76,12 +102,12 @@ type
     timestamp*: uint64
     extra_data*: List[byte, MAX_EXTRA_DATA_BYTES]
     base_fee_per_gas*: UInt256
-    excess_data_gas*: UInt256  # [New in EIP-4844]
 
     # Extra payload fields
     block_hash*: Eth2Digest # Hash of execution block
     transactions*: List[Transaction, MAX_TRANSACTIONS_PER_PAYLOAD]
     withdrawals*: List[Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD]
+    excess_data_gas*: UInt256  # [New in EIP-4844]
```
henridf, line 110: new in Deneb

beacon_chain/spec/datatypes/eip4844.nim
```diff
@@ -97,12 +123,12 @@ type
     timestamp*: uint64
     extra_data*: List[byte, MAX_EXTRA_DATA_BYTES]
     base_fee_per_gas*: UInt256
-    excess_data_gas*: UInt256  # [New in EIP-4844]
 
     # Extra payload fields
     block_hash*: Eth2Digest  # Hash of execution block
     transactions_root*: Eth2Digest
     withdrawals_root*: Eth2Digest
+    excess_data_gas*: UInt256  # [New in EIP-4844]
```
henridf, line 131: Deneb

beacon_chain/spec/datatypes/eip4844.nim
tersec: Will fix in next PR

beacon_chain/spec/datatypes/eip4844.nim
tersec: Will fix in next PR

beacon_chain/spec/datatypes/eip4844.nim
tersec: Also, deliberately didn't update the https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.2/specs/eip4844/beacon-chain.md#executionpayload etc URLs for these. So they mark it as not-yet-reviewed for actually being 100% consistent with rc.3, just enough to pass tests. So would not have updated them to rc.3 without changing that.

Deliberately didn't update the rest of the URLs either, even the easy ones. That was always going to be a separate PR, to keep the noisy/invasive changes also completely semantics-change-free, i.e. why all those PRs have descriptions along the lines of comments/markdown only. This makes enough changes I wanted people to be able to see what it did and not be buried in hundreds of lines of comment changes.

That's part of the purpose/value of these -- it's okay to make these partial updates, too. In this case, it's true that I'd have preferred to update the wording to `Deneb` at the same time since it's no more invasive once one is moving the line as a whole.

beacon_chain/spec/datatypes/eip4844.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4646

beacon_chain/spec/datatypes/eip4844.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4646

Issue #4573, opened 2023-01-31
`make NIMFLAGS="-d:debug"` builds in release mode


tersec:
This is given as an example in https://nimbus.guide/developers.html for
```
make NIMFLAGS="-d:release"
```
but the obvious extrapolation would be to apply that to `-d:debug`.

This does not necessarily work:
```
$ rm nimcache -rv && time make NIMFLAGS="-d:debug"
... builds ...
```
meanwhile, as this is running:
```
nimbus-eth2$ ls nimcache/
release
```
but no `debug`, and
```
nim c --compileOnly -o:build/ncli_db -d:debug --verbosity:0 --hints:off -d:chronicles_log_level=TRACE -d:release --parallelBuild:1 -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,lodestar,prysm,teku -d:nimCachePathOverride=nimcache/release/ncli_db ncli/ncli_db.nim
```
clarifies what's happening: it passes `NIMFLAGS` before some hardcoded `-d:release` in Nim command line.

tersec:
This isn't going to work, because of Nim's behavior around these flags:
```
$ nim c -d:release -d:debug empty.nim
Hint: used config file '/etc/nim/nim.cfg' [Conf]
Hint: used config file '/etc/nim/config.nims' [Conf]
.........................................................
Hint:  [Link]
Hint: gc: refc; opt: speed; options: -d:release
26642 lines; 0.131s; 31.707MiB peakmem; proj: /home/user/empty.nim; out: /home/user/empty [SuccessX]
$ nim c -d:debug -d:release empty.nim
Hint: used config file '/etc/nim/nim.cfg' [Conf]
Hint: used config file '/etc/nim/config.nims' [Conf]
.........................................................
Hint:  [Link]
Hint: gc: refc; opt: speed; options: -d:release
26642 lines; 0.125s; 31.641MiB peakmem; proj: /home/user/empty.nim; out: /home/user/empty [SuccessX]
```
https://github.com/status-im/nimbus-eth2/pull/4655 does enable proper overriding from command-line or environmental `NIMFLAGS` over the `Makefile` set, so one can, e.g., use `-u:release --opt:none` to achieve a similar result.

tersec:
https://github.com/status-im/nimbus-eth2/pull/4655 resolves this as much as feasible.

Issue #4571, opened 2023-01-31
`registerProtocols` in LTO in gcc 12.x incorrectly tries to modify read-only link segment


tersec:
Tracking https://github.com/nim-lang/Nim/issues/20604 / https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107769

tersec:
Opened as https://github.com/status-im/nimbus-eth2/issues/4001 but that keeps attracting people who have other segfaults

tersec:
It is a sigsegv, but not all sigsegvs. It looks approximately like:
```
~/.opt/status-im/nimbus-eth2 stable
❯ ./run-ropsten-beacon-node.sh     
Traceback (most recent call last, using override)
../sysdeps/x86_64/start.S(115) _start
/home/user/.opt/status-im/nimbus-eth2/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(456) main
/home/user/.opt/status-im/nimbus-eth2/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(446) NimMain
/home/user/.opt/status-im/nimbus-eth2/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(437) PreMain
/home/user/.opt/status-im/nimbus-eth2/vendor/nim-libp2p/libp2p/stream/bufferstream.nim(147) PreMainInner
/home/user/.opt/status-im/nimbus-eth2/beacon_chain/sync/sync_protocol.nim(220) beacon_chain_sync_protocolInit000
/home/user/.opt/status-im/nimbus-eth2/beacon_chain/networking/eth2_network.nim(928) registerProtocol
/home/user/.opt/status-im/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system.nim(1266) insert
/home/user/.opt/status-im/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/gc.nim(260) unsureAsgnRef
/home/user/.opt/status-im/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/excpt.nim(610) signalHandler
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

If it looks too different, and especially if it lacks `unsureAsgnRef`, then it is likely not this bug.

It happens on startup, not shutdown, as well.

tersec:
[...]

PR #4655, opened 2023-02-22
build make all targets in debug mode on GitHub Actions CI

tersec:
This reduces the time for an equivalent, starting sans `nimcache`s, build locally for `make -j2 LOG_LEVEL=TRACE` from 25 to 15 minutes. A recent `unstable` commit required 1h40m for Windows; if this is similarly faster, that's worth the tradeoff in building slightly different binaries which are never used.

Jenkins still builds release `make LOG_LEVEL=TRACE`, and while it's important not to run Nim in `--compileOnly` mode because it does sometimes generate invalid C code which it's useful to run `gcc` against, and for build artifacts which are run/released/tested in CI or otherwise, `-O3` with LTO has definitely triggered issues, those would not be visible in this context regardless.

`make all_tests` regardless effectively builds a copy of `nimbus_beacon_node`.

This additional release mode build time doesn't much help catch bugs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ DEPOSITS_DELAY := 0
 #- "--define:release" cannot be added to "config.nims"
 #- disable Nim's default parallelisation because it starts too many processes for too little gain
 #- https://github.com/status-im/nim-libp2p#use-identify-metrics
-NIM_PARAMS += -d:release --parallelBuild:1 -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,lodestar,prysm,teku
+NIM_PARAMS := -d:release --parallelBuild:1 -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,lodestar,prysm,teku $(NIM_PARAMS)
 
 ifeq ($(USE_LIBBACKTRACE), 0)
 NIM_PARAMS += -d:disable_libbacktrace
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
 
       - name: Build binaries (with trace logging enabled)
         run: |
-          ${make_cmd} -j ${ncpu} V=1 NIM_COMMIT=${{ matrix.branch }} LOG_LEVEL=TRACE
+          ${make_cmd} -j ${ncpu} V=1 NIM_COMMIT=${{ matrix.branch }} LOG_LEVEL=TRACE NIMFLAGS="-u:release --opt:none"
           # The Windows image runs out of disk space, so make some room
           rm -rf build nimcache
 


Makefile
```diff
@@ -143,7 +143,7 @@ DEPOSITS_DELAY := 0
 #- "--define:release" cannot be added to "config.nims"
 #- disable Nim's default parallelisation because it starts too many processes for too little gain
 #- https://github.com/status-im/nim-libp2p#use-identify-metrics
-NIM_PARAMS += -d:release --parallelBuild:1 -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,lodestar,prysm,teku
+NIM_PARAMS := -d:release --parallelBuild:1 -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,lodestar,prysm,teku $(NIM_PARAMS)
```
zah, line 146: If `-d:release` is still here, how is #4573 addressed?

To clarify, I don't think passing `-d:release ... -d:debug` on the command line is a solution, because Nim will routinely use checks such as `when defined(release): ...` which will still trigger on such invocation.

tersec: Yes, I discovered this. It does not address 4573, and that's why it uses `-u:release --opt:none`. The main goal here is the CI time savings though. 4573 was incidental.

tersec:
Example durations for "building binaries" for (Linux, macOS, Windows) before are (1h4m, 41m, 1h42m) and (46m, 48m, 1h23m), (46m, 41m, 1h52m), (57m, 42m, 1h38m), and (50m, 48m, 1h20m).

This PR took (31m, 23m, 1h7m).

zah:
For debug builds, we may be able to greatly reduce the build times by disabling LTO and using the mold linker:
https://github.com/rui314/mold

arnetheduck:
looking at that repo, `lld` seems like a better option: mold is faster because it does parallel computation which the CI's barely have (and we link many things in parallel anyway so..) - the difference vs lld is minimal really, and then there's this gem: 

`mold/macOS is commercial software. For mold/macOS, please visit https://github.com/bluewhalesystems/sold.`

This brings us back to the old issue about using clang on windows as well

tersec:
Probably -- this was meant to be a minimally invasive as I could manage (it doesn't change `config.nims` at all, doesn't introduce any new dependencies, et cetera).

Depending on how we'd have to deploy the `mold` linker, it might mean an additional download, which can also fail. I don't usually post those to the CI issue tracking bugs, because they're not especially interesting, but they're recurring.

Edit after seeing @arnetheduck's comment -- same applies to `lld`.

zah:
I'm certainly not insisting that `mold` should be included in this PR, but I think it's valuable direction to pursue for faster compile-test cycles in local builds. Another potential downsides of the debug builds is that they used to be too slow to handle the load on a real network and this can be potentially addressed by enabling the optimisations in specific modules (e.g. BLS, HTR, etc). Perhaps we are already there due to the usage of assembly in BLST.

tersec:
> Another potential downsides of the debug builds is that they used to be too slow to handle the load on a real network

This is exactly why it doesn't enable these on any code which runs. There's another risk associated with doing that, that e.g., that https://github.com/status-im/nimbus-eth2/issues/4571 only shows up with `-O3` and LTO. If people just use debug builds all the time, those will become release-surprises.

arnetheduck:
> If people just use debug builds all the time, those will become release-surprises.

this is one of the original reasons we went with release builds to begin with

tersec:
If people want to change linkers, or make further changes to default local build settings, or apply this to Jenkins, or any other potential changes, that can be another PR.

PR #4656, opened 2023-02-23
Fix IndexError for reasons.

cheatfate:


beacon_chain/validator_client/common.nim
```diff
@@ -282,8 +286,10 @@ proc getFailureReason*(exc: ref ValidatorApiError): string =
             res = item.failure
         res
     $maxFailure
-  else:
+  elif errorsCount == 1:
     $errors[0].failure
+  else:
+    $ApiFailure.Timeout
```
arnetheduck: shouldn't timeout failures set this explicitly in errors? and this function return "unknown" when errorsCount == 0?

PR #4657, opened 2023-02-23
Fix attester and proposer duties optimistic execution behavior.

cheatfate:


beacon_chain/rpc/rest_utils.nim
```diff
@@ -42,6 +42,19 @@ proc getSyncedHead*(node: BeaconNode, slot: Slot): Result[BlockRef, cstring] =
 
   ok(head)
 
+proc getOptimisticSyncedHead*(node: BeaconNode,
```
arnetheduck: maybe we should return `Result[(head, optimistic)...` instead? that would avoid the separate / duplicate query for the flag

arnetheduck (approved):

beacon_chain/rpc/rest_utils.nim
```diff
@@ -30,23 +30,12 @@ func match(data: openArray[char], charset: set[char]): int =
       return 1
   0
 
-proc getSyncedHead*(node: BeaconNode, slot: Slot): Result[BlockRef, cstring] =
+proc getSyncedHead*(node: BeaconNode, slot: Slot,
```
arnetheduck: sure - this works - I was thinking about this in the other direction, that `getSyncedHead` returns whether the head was optimistic or not - this makes it easy to translate into "execution_optimistic" in the caller - otherwise, the caller has to make the same check again as getSyncedHead is doing (`isSynced` calls `is_optimistic` which costs a hash-table lookup)

cheatfate: This change will require a lot of changes in places where `optimistic` is not supposed.

arnetheduck: sure - that said, I think *most* uses of `getSyncedHead` actually require this knowledge anyway, so overall I believe it would be a simplification (since most functions using it also should be returning `execution_optimistic`).

Issue #4661, opened 2023-02-24
Thousands of `BareExcept` warnings


tersec:
Came along with `nimbus-build-system` bump. Real warnings, should be fixed, but currently creates problems finding anything else i compiler output. From `make LOG_LEVEL=TRACE test && make LOG_LEVEL=TRACE` run, as the CIs effectively do:
```
$ grep BareExcept warnings.txt | sort | uniq -c | sort -rn | head -n15
   2176 nimbus-eth2/vendor/nim-chronicles/chronicles.nim(350, 7) Warning: The bare except clause is deprecated; use `except CatchableError:` instead [BareExcept]
    663 nimbus-eth2/vendor/nim-chronos/chronos/asyncfutures2.nim(389, 3) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
    621 nimbus-eth2/vendor/nim-unittest2/unittest2.nim(821, 5) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
    115 nimbus-eth2/vendor/nim-confutils/confutils.nim(885, 5) Warning: The bare except clause is deprecated; use `except CatchableError:` instead [BareExcept]
     37 nimbus-eth2/vendor/NimYAML/yaml/serialization.nim(1353, 3) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
     33 nimbus-eth2/vendor/nim-unittest2/unittest2.nim(981, 5) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
     19 nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/threads.nim(171, 5) Warning: The bare except clause is deprecated; use `except CatchableError:` instead [BareExcept]
     17 nimbus-eth2/ncli/e2store.nim(53, 3) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
     16 nimbus-eth2/beacon_chain/spec/signatures_batch.nim(438, 3) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
     14 nimbus-eth2/vendor/nim-metrics/metrics.nim(711, 5) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
     14 nimbus-eth2/nfuzz/libnfuzz.nim(88, 5) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
[...]

PR #4664, opened 2023-02-24
fix 2 BareExcept warning sources

tersec:
https://github.com/status-im/nimbus-eth2/issues/4661

in particular
```
     17 nimbus-eth2/ncli/e2store.nim(53, 3) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
     14 nimbus-eth2/nfuzz/libnfuzz.nim(88, 5) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
      9 nimbus-eth2/beacon_chain/statusbar.nim(92, 3) Warning: catch a more precise Exception deriving from CatchableError or Defect. [BareExcept]
```

The semantics are slightly different, between `except:` and `except CatchableError`, but for these three, it shouldn't be important.

`nfuzz` isn't currently in active use. No loss there -- if someone tries to use it and discovers that it's inadequate to the purpose, it can be addressed then.

`ncli/e2store.nim`  is used to address what `ioErrorMsg` does, and (a) that's `nim-stew` code under Status control, and fixable even if it throws something where this difference is important; and (b) it doesn't intentionally throw any exceptions. It calls `formatMessageW` and `localFree`. Both of these are just direct calls to Win32 API functions, which don't per se participate in the Nim exception system.

` nimbus-eth2/beacon_chain/statusbar.nim` uses it in some generic way, as part of an `ignoreExceptions` wrapper. This is used for `terminal.nim` functions, which uniformly call [raiseOSError](https://nim-lang.org/docs/os.html#raiseOSError%2COSErrorCode%2Cstring), which raises an [OSError](https://nim-lang.org/docs/system.html#OSError), which:
```nim
  OSError = object of CatchableError
```

### DIFF
--- a/nfuzz/libnfuzz.nim
+++ b/nfuzz/libnfuzz.nim
@@ -85,7 +85,7 @@ template decodeAndProcess(typ, process: untyped): bool =
       raise newException(
         FuzzCrashError,
         "Unexpected  (logging?) IOError in state transition", e)
-    except Exception as e:
+    except CatchableError as e:
       # TODO why an Exception?
       # Lots of vendor code looks like it might raise a bare exception type
       raise newException(FuzzCrashError, "Unexpected Exception in state transition", e)
--- a/nfuzz/libnfuzz.nim
+++ b/nfuzz/libnfuzz.nim
@@ -85,7 +85,7 @@ template decodeAndProcess(typ, process: untyped): bool =
       raise newException(
         FuzzCrashError,
         "Unexpected  (logging?) IOError in state transition", e)
-    except CatchableError as e:
+    except Exception as e:
       # TODO why an Exception?
       # Lots of vendor code looks like it might raise a bare exception type
       raise newException(FuzzCrashError, "Unexpected Exception in state transition", e)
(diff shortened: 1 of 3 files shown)


nfuzz/libnfuzz.nim
```diff
@@ -85,7 +85,7 @@ template decodeAndProcess(typ, process: untyped): bool =
       raise newException(
         FuzzCrashError,
         "Unexpected  (logging?) IOError in state transition", e)
-    except Exception as e:
+    except CatchableError as e:
```
zah: I'm not sure this is appropriate. The `FuzzCrashError` exception is intended to trigger a crash report from the fuzzer and this should happen if the fuzzer discovers a `Defect` in the code.

tersec: https://github.com/status-im/nimbus-eth2/pull/4664/commits/f152dea56e77ef771898e61fa4584589a6953607

tersec:
Warning appears to be broken while targeting 1.6; goal will be just to disable it.

PR #4673, opened 2023-02-26
Various fixes for VC and BN REST server.

cheatfate:
1. Fix issue when VC i unable to detect error codes properly.
2. Force VC to use `RestPlainResponse` which will allow to print BN response error messages in VC logs.
3. Fix issue when prepareBeaconCommitteeSubnet() unable to perform action when BN is optimistically synced.

### DIFF
--- a/beacon_chain/validator_client/api.nim
+++ b/beacon_chain/validator_client/api.nim
@@ -670,7 +670,10 @@ proc getProposerDuties*(
             node.updateStatus(RestBeaconNodeStatus.Unexpected)
             ApiResponse[GetProposerDutiesResponse].err($res.error)
           else:
-            ApiResponse[GetProposerDutiesResponse].ok(res.get())
+            let data = res.get()
+            if data.execution_optimistic.get(false):
+              node.updateStatus(RestBeaconNodeStatus.OptSynced)
+            ApiResponse[GetProposerDutiesResponse].ok(data)
         of 400:
           debug ResponseInvalidError, response_code = response.status,
                 endpoint = node, reason = response.getErrorMessage()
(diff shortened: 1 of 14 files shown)


beacon_chain/validator_client/api.nim
```diff
@@ -1366,46 +1451,43 @@ proc submitPoolAttestations*(
...
+                endpoint = node, reason = response.getIndexedErrorMessage()
           node.status = RestBeaconNodeStatus.Offline
           failures.add(ApiNodeFailure.init(node, ApiFailure.Unexpected))
-          false
+          ApiResponse[bool].err(ResponseUnexpectedError)
 
     if res.isErr():
       raise (ref ValidatorApiError)(msg: res.error, data: failures)
-    return true
+    return false
```
arnetheduck: why false here?

cheatfate: This is impossible case, but because i can't prove it there is no reason to return successful result.

arnetheduck: why not `raiseAssert` then?

arnetheduck: or `doAssert res.isErr`?

arnetheduck (approved):

beacon_chain/validator_client/common.nim
```diff
@@ -344,6 +348,84 @@ chronicles.expandIt(SyncCommitteeDuty):
...
+      warn "Beacon node has incompatible configuration"
+      node.status = status
+  of RestBeaconNodeStatus.Compatible:
+    if node.status != status:
+      notice "Beacon node is compatible"
+      node.status = status
+  of RestBeaconNodeStatus.NotSynced:
+    if node.status notin {RestBeaconNodeStatus.NotSynced,
+                          RestBeaconNodeStatus.OptSynced}:
+      # When BN returns 503 it doesn't clarify if its opt-synced or not-synced,
```
arnetheduck: 503 means not-synced - opt-synced is a subset of not-synced: we can either be not-synced (which means both EL and CL are out of sync) or opt-synced (only EL is out of sync). The EL can never be in sync if the CL is out of sync.

PR #4682, opened 2023-02-28
Remove uses of beacon_block_and_blobs_sidecar topic

henridf:
The topic goes away with decoupled blocks and blobs.

The first commit of this PR is from #4681; I will rebase after that PR goes in.


henridf:
> The first commit of this PR is from https://github.com/status-im/nimbus-eth2/pull/4681; I will rebase after that PR goes in.

done.

tersec (approved):

PR #4683, opened 2023-02-28
Remove SignedBeaconBlockAndBlobsSidecar

henridf:
This commit removes SignedBeaconBlockAndBlobsSidecar and all remaining
references.

Only the last commit is net-new; the first three are from #4682. I will rebase  after that PR goes in.


henridf:
> Only the last commit is net-new; the first three are from https://github.com/status-im/nimbus-eth2/pull/4682. I will rebase after that PR goes in.

Done.

tersec (approved):

PR #4688, opened 2023-03-02
vc: fix missing attestations due to doppelganger in epoch 1

arnetheduck:
* log which validator triggers doppelganger
* move activity detection closer to where it's performed, record
aggregation as activity (since it's part of the liveness endpoint)
* vc: check for doppelgangers in epoch 0 also (so that activity in epoch
1 can happen)
* avoid some looping when processing activities

### DIFF
--- a/beacon_chain/validator_client/doppelganger_service.nim
+++ b/beacon_chain/validator_client/doppelganger_service.nim
@@ -40,6 +40,9 @@ proc processActivities(service: DoppelgangerServiceRef, epoch: Epoch,
           validator.doppelgangerChecked(epoch)
 
           if item.is_live and validator.triggersDoppelganger(epoch):
+            warn "Doppelganger detection triggered",
+              validator = shortLog(validator), epoch
+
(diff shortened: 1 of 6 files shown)


beacon_chain/validator_client/doppelganger_service.nim
```diff
@@ -40,6 +40,9 @@ proc processActivities(service: DoppelgangerServiceRef, epoch: Epoch,
           validator.doppelgangerChecked(epoch)
 
           if item.is_live and validator.triggersDoppelganger(epoch):
```
cheatfate, line 42: i think `triggersDoppelganger` is ugly name, because its very hard to see `s` inside `trigger(s)` from first view, and it could be considered as `trigger doppelganger` instead of `triggers doppelganger`.

arnetheduck: have a better suggestion?

tersec (approved):

PR #4697, opened 2023-03-06
Read messages before applying quota to avoid mplex backpressure issues

Menduist:
With mplex, the backpressure is at the connection level, and not the stream level.
It means that if a received message is not read, the entire connection will get blocked

This PR modifies the quota to be applied after reading the message, and not before, since doing so will pause the entire connection
The number of simultaneous requests is still limited by libp2p

cc @arnetheduck


beacon_chain/networking/eth2_network.nim
```diff
@@ -1137,6 +1124,19 @@ proc handleIncomingStream(network: Eth2Node,
...
+    # regardless if the request succeeds or fails - we don't count waiting
+    # for this quota against timeouts so as not to prematurely disconnect
+    # clients that are on the edge - nonetheless, the client will count it.
+    #
+    # When a client exceeds their quota, they will be slowed down without
+    # notification - as long as they don't make parallel requests (which is
+    # limited by libp2p), this will naturally adapt them to the available
+    # quota.
+
+    awaitQuota(peer, libp2pRequestCost, shortProtocolId(protocolId))
```
arnetheduck: hm, it's ok to read before quota, but we must count failures also - ie receiving an invalid message should also count / be slowed down - this protects against buggy clients tight-looping an invalid request

arnetheduck: ping @Menduist  what shall we do with this PR?

Menduist: Sorry forgot about it, I'll find a way to take your comment into account and we should merge it

Menduist:
A possible issue with this PR is that we may hold 10 "request object" in memory while we wait for quota, whereas before, in this scenario, the data would have been queued in the kernel instead (so not decompressed etc)
Not sure if this can cause trouble, I would guess most request objects are quite small, but I'll let someone with better knowledge of the req/response domain weigh in

arnetheduck:
>  Not sure if this can cause trouble, I would guess most request objects are quite small, but I'll let someone with better knowledge of the req/response domain weigh in

this is fine since the number of concurrent requests is reasonably capped - that said, there's an [old issue](https://github.com/status-im/nim-ssz-serialization/issues/20) open about computing the maximum size of an SSZ message based on its limits which would further cap the damage - it's already possible for fixed-size messages (which most requests are) so it would be good to check that as well.

cc @zah

arnetheduck:
> A possible issue with this PR

@Menduist in the meantime, this deserves an explanatory comment in the code

arnetheduck (approved):

beacon_chain/networking/eth2_network.nim
```diff
@@ -1152,6 +1140,19 @@ proc handleIncomingStream(network: Eth2Node,
...
+        # regardless if the request succeeds or fails - we don't count waiting
+        # for this quota against timeouts so as not to prematurely disconnect
+        # clients that are on the edge - nonetheless, the client will count it.
+        #
+        # When a client exceeds their quota, they will be slowed down without
+        # notification - as long as they don't make parallel requests (which is
+        # limited by libp2p), this will naturally adapt them to the available
+        # quota.
+
+        awaitQuota(peer, libp2pRequestCost, shortProtocolId(protocolId))
```
arnetheduck, line 1162: also, we need an explanation here why it's done in finally, which is .. unusual

Menduist: Done

PR #4701, opened 2023-03-07
Update sync to use post-decoupling RPC

henridf:


beacon_chain/sync/sync_manager.nim
```diff
@@ -237,6 +238,23 @@ proc remainingSlots(man: SyncManager): uint64 =
     else:
       0'u64
 
+func groupBlobs*[T](req: SyncRequest[T], blobs: seq[ref BlobSidecar]):
+     Result[seq[BlobSidecars], string] =
+  var grouped = newSeq[BlobSidecars](req.count)
+  var rawCur = 0
+
+  for groupedCur in 0 ..< len(grouped):
```
tersec, line 246: Depending on whether it's easier to use `int` or `uint64` as the main type of the loop iterator, there's also `lenu64` (e.g., `for groupedCur in 0 ..< lenu64(grouped):`). Whether it's an improvement here isn't clear to me though, since `groupedCur` is also used as an array index.

It'd be nice to have this be maximally typesafe, but at some point there's just some tension between Nim's requirement to index arrays by `int` and the general usage of `uint64` by Ethereum, which requires case-by-case judgement.

tests/test_sync_manager.nim
```diff
@@ -70,6 +70,13 @@ suite "SyncManager test suite":
       curslot = curslot + 1'u64
     res
 
+  proc createBlobs(slots: seq[Slot]): seq[ref BlobSidecar] =
```
tersec: can probably be `func`

henridf: f01cfad

beacon_chain/sync/sync_manager.nim
henridf: My preference is to leave as is (I don't think there's much typesafety benefit in this function), but I don't mind changing if you feel strongly about it.

tersec: Yeah, as is is fine.

tersec (approved):

PR #4709, opened 2023-03-09
Don't use the wrong SECONDS_PER_ETH1_BLOCK constant

zah:


beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -393,8 +393,8 @@ func voting_period_start_time(state: ForkedHashedBeaconState): uint64 =
 func is_candidate_block(cfg: RuntimeConfig,
                         blk: Eth1Block,
                         period_start: uint64): bool =
-  (blk.timestamp + cfg.SECONDS_PER_ETH1_BLOCK * cfg.ETH1_FOLLOW_DISTANCE <= period_start) and
-  (blk.timestamp + cfg.SECONDS_PER_ETH1_BLOCK * cfg.ETH1_FOLLOW_DISTANCE * 2 >= period_start)
+  (blk.timestamp + SECONDS_PER_SLOT * cfg.ETH1_FOLLOW_DISTANCE <= period_start) and
+  (blk.timestamp + SECONDS_PER_SLOT * cfg.ETH1_FOLLOW_DISTANCE * 2 >= period_start)
```
etan-status, line 397: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#get_eth1_data

https://eips.ethereum.org/EIPS/eip-6110

zah: The point is that the spec accidentally got wrong as a result of an oversight. The change here preserves the spirit of the original spec. Arguably, since this is the "honest validator" spec, some deviations are reasonable. It's unlikely that the proposed fix will cause any issues in practice because it effectively increases the size of the candidate set which should only prevent potential situations where Nimbus decides to disagree with the establishing majority in the `Eth1Data` voting.

zah:
Since, we've merged some alternative mitigations, I think we can wait for https://eips.ethereum.org/EIPS/eip-6110 as the eventual resolution for this.

PR #4720, opened 2023-03-10
rename Deneb database tables (not in testnets atm)

etan-status:


beacon_chain/beacon_chain_db.nim
```diff
@@ -518,7 +518,7 @@ proc new*(T: type BeaconChainDB,
       kvStore db.openKvStore("altair_state_no_validators").expectDb(),
       kvStore db.openKvStore("bellatrix_state_no_validators").expectDb(),
       kvStore db.openKvStore("capella_state_no_validator_pubkeys").expectDb(),
-      kvStore db.openKvStore("eip4844_state_no_validator_pubkeys").expectDb()]
+      kvStore db.openKvStore("deneb_state_no_validator_pubkeys").expectDb()]
```
etan-status, line 521: note, this one and `deneb_blocks` seem to get created even when DENEB_FORK_EPOCH == FAR_FUTURE_EPOCH. I guess they are empty, but a bit ugly, still.

tersec: It would seem reasonable enough to gate creating those on `DENEB_FORK_EPOCH != FAR_FUTURE_EPOCH`, I agree

PR #4736, opened 2023-03-15
allow trusted node sync based on LC trusted block root

etan-status:
Extends `trustedNodeSync` with a new `--trusted-block-root` option that allows initializing a light client. No `--state-id` must be provided. The beacon node will then use this light client to obtain the latest finalized state from the remote server in a trust-minimized fashion. Note that the provided `--trusted-block-root` should be somewhat recent, and that security precautions such as comparing the state root against block explorers is still recommended.


etan-status:
~~Requires https://github.com/status-im/nim-stew/pull/174~~

etan-status:
```
build/nimbus_beacon_node trustedNodeSync --data-dir=$HOME/nimbus/data/goerli-trusted --network:goerli \
    --trusted-node-url=http://127.0.0.1:5052 --backfill=false \
    --trusted-block-root=0x24ad5b7d941e147b80edb0aa34c1c454e6e467e68e83142e51e9d21b9226eb79
```

image

beacon_chain/conf.nim
```diff
@@ -780,15 +780,19 @@ type
...
+      .}: Option[string]
 
       blockId* {.
         hidden
         desc: "Block id to sync to - this can be a block root, slot number, \"finalized\" or \"head\" (deprecated)"
       .}: Option[string]
 
+      lcTrustedBlockRoot* {.
+        hidden
+        desc: "Recent trusted finalized block root to initialize light client from"
```
tersec, line 792: What compatibility guarantees are you aiming for on this? That is, whether it's sort of temporarily hidden until it's more stable, or whether it's intended to be a kind of debug option. If it's something which might plausibly disappear (e.g., it's a workaround for some temporary condition elsewhere), then, maybe ensure either in the option name, description, or both it's clear that's the case. Otherwise `hidden` is nice, but potentially not enough to keep it from becoming part of the required backwards compatibility surface.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -28,6 +28,20 @@ export
...
+    return err("Missing or incorrect Content-Type")
+  ok contentType.get.mediaType
+
+func decodeEthConsensusVersion*(
+    value: string): Result[ConsensusFork, cstring] =
+  let normalizedValue = value.toLowerAscii()
+  for consensusFork in ConsensusFork:
+    if normalizedValue == ($consensusFork).toLowerAscii():
+      return ok consensusFork
+  err(cstring("Unsupported Eth-Consensus-Version: " & value))
```
tersec: Not sure why this even compiles (is there a warning?), but it's not really valid -- `cstring` should be from a static set. It's literally a `NULL`-terminated C string, without any additional memory management.

beacon_chain/conf.nim
etan-status: > temporarily hidden until it's more stable

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
etan-status: Don't think so, we have other `cstring` conversions as well, e.g., `cstring(Base10.toString(uint16(port))` in Chronos, or `logLoggingFailure(cstring(msg), e)` in `nimbus_beacon_node`.

beacon_chain/spec/eth2_apis/rest_light_client_calls.nim
```diff
@@ -0,0 +1,311 @@
...
+  template mediaType: auto = mediaTypeRes.get
+
+  if mediaType == OctetStreamMediaType:
+    let l = data.len
+    var
+      res: S
+      o = 0
+    while l - o != 0:
+      # response_chunk_len
+      if l - o < 8:
```
tersec: Maybe worth doing something like `type foo = uint64; const bar = sizeof(foo)`, not sure.

As it is, there's an implicit relationship here between the two which modifying some, but not all, will cause decoding errors for.

etan-status: Hmm, yeah, it's a bit tricky. Two tricky portions further down where reviewing becomes more annoying by reducing magic number:

- `int.high.foo` only guaranteed to work if `foo.high >= int.high`.
- The `ForkDigest` conversion explicitly going from 0 .. 3

Have replaced the magic numbers for now (while keeping them as a comment to be able to reason about it)

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
tersec:
```nim
func yxf(x: int): cstring = cstring($x)
echo yxf(3)
```
results in (cleaned up slightly)
```c
static N_INLINE(NCSTRING, nimToCStringConv)(NimStringDesc* s) {
    NCSTRING result;
    result = (NCSTRING)0;
    {
        NIM_BOOL T3_;
        T3_ = (s == ((NimStringDesc*) NIM_NIL));
        if (T3_) goto LA4_;
        T3_ = ((*s).Sup.len == ((NI) 0));
        LA4_: ;
        if (!T3_) goto LA5_;
        result = "";
    }
    goto LA1_;
    LA5_: ;
    {
        result = ((NCSTRING) ((*s).data));
    }
    LA1_: ;
    return result;
}

N_LIB_PRIVATE N_NIMCALL(NCSTRING, yxf__a_3)(NI x) {
    NCSTRING result;
    NimStringDesc* T1_;
    T1_ = dollar___systemZdollars_3(x); /* allocate here */
    result = nimToCStringConv(T1_);     /* lose track of allocation here */
    popFrame();
    return result;
}
```

tersec: Or, it looks like this might be okay, not certain. It's true it's used elsewhere.

etan-status: So, makes a copy, essentially, which is expected to convert from Pascal encoding to C encoding.
Which, seems alright, to be slightly inefficient in case of an error, no?

`raiseRestDecodingBytesError` only takes a `cstring` for some reason

tersec: To summarize some investigation, this kind of thing reproducibly can be made to fail:
```nim
type X = object
  s: string

proc foo(x: int): cstring = cstring((ref X)(s: $x)[].s)

let c = foo(3)
GC_fullCollect()
discard (ref X)(s: $4)
GC_fullCollect()
discard (ref X)(s: $5)
echo "Printing result of foo(3): ", c
doAssert c == "3"
```

zah: Converting a Nim `string` to a `cstring` is only safe when the `cstring` is stored in a stack location (e.g. it's being used as a parameter). This is not the case in Dustin's example above which is storing a `cstring` in `c`, which happens to be a global variable.

Nevertheless, such code is a bit fragile as it is relying on intricate GC implementation details.

etan-status: Nim should not compile code that doesn't work as expected though? As in, it should detect that a `cstring` ends up on the heap.

tersec: Is the original code guaranteed safe? It's not heap-allocated, in any obvious way, but it also escapes the lifetime of the dynamically constructed string from which it was converted.

The heap-allocated version was how I created a reproducible example, but I don't know whether the original was reliable either.

tersec: For example,
```nim
proc foo(x: int): cstring =
  var y = $x
  cstring(y)

proc baz() =
  let c = foo(3)
  echo "Printing result of foo(3): ", c
  doAssert c == "3"

baz()
```
with `--mm:arc` and `-mm:orc` also fails. No (Nim-level) heap allocation involved.

etan-status: PR is updated to not return `cstring` anymore. The `cstring` conversion now happens as late as possible, while passing argument to function that takes `cstring`.

beacon_chain/spec/eth2_apis/rest_light_client_calls.nim
```diff
@@ -0,0 +1,316 @@
...
+            obj.checkForkConsistency(cfg, consensusFork)
+            res.add obj
+          else:
+            raiseRestDecodingBytesError(
+              cstring("Unsupported fork: " & $consensusFork))
+      except SszError as exc:
+        raiseRestDecodingBytesError(cstring("Malformed data: " & $exc.msg))
+    return res
+
+  if mediaType == ApplicationJsonMediaType:
```
tersec: Minor, but `elif` here would clarify the mutually exclusive nature of these in a way beyond the `return res`es at the end of each, and without requiring verifying that there aren't any other early returns within those conditional blocks.

tersec (approved):

beacon_chain/spec/eth2_apis/rest_light_client_calls.nim
etan-status: Moved the return to the top to prevent accidents, I think that's what you meant.

tersec: Partly that, but also, it's structured as
```nim
  if mutually_exclusive_condition_1:
    a couple dozen lines of code with multiple indentation levels and control flow
    return

  if mutually_exclusive_condition_2:
    some more code
    return
```

Both `if` blocks can't possibly run, but to infer that needs more than a cursory glance for no particular reason that I can discern. It's a case of is-this-switch-statement-meant-to-fall-through-or-not.

etan-status: you mean, use `elif` for the second condition in these?

downsides are: 
- it's not consistently usable, e.g., when there is a `let` in-between the two `if`.
- it suggests some sort of relations between the two conditions, or more generally, an exhaustion of multiple options
- if just doing `if` / `elif` / `elif` ..., analyzers like clang-tidy like to point out that an `else` is missing to achieve said exhaustion

ultimately, very minor though.

in the context of this PR, there is only a single `return`. are you referring to the `raise`?

for example:

```nim
      if responseChunkLen > int.high.chunkLenType:
          raiseRestDecodingBytesError("Malformed data: Unsupported length")
        if l - o < responseChunkLen.int:
          raiseRestDecodingBytesError("Malformed data: Incomplete chunk")   
```

or, the `try` / `except` bloks such as this one:

```nim
        try:
          withLcDataFork(lcDataForkAtConsensusFork(consensusFork)):
            when lcDataFork > LightClientDataFork.None:
              type T = typeof(res[0])
              var obj = T(kind: lcDataFork)
              obj.forky(lcDataFork) = SSZ.decode(
                data.toOpenArray(begin + contextLen, after - 1),
                T.Forky(lcDataFork))
              obj.checkForkConsistency(cfg, consensusFork)
              res.add obj
            else:
              raiseRestDecodingBytesError(
                cstring("Unsupported fork: " & $consensusFork))
        except SszError as exc:
          raiseRestDecodingBytesError(cstring("Malformed data: " & $exc.msg))
```

I can see the "multiple indentation levels and control flow" argument, but nim's `try` / `except` kinda pushes that.

Maybe I'm misunderstanding the point.

PR #4743, opened 2023-03-17
VC: Hardening and optimizing time handling.

cheatfate:


beacon_chain/beacon_clock.nim
```diff
@@ -69,19 +73,39 @@ proc fromNow*(c: BeaconClock, t: BeaconTime): tuple[inFuture: bool, offset: Dura
 proc fromNow*(c: BeaconClock, slot: Slot): tuple[inFuture: bool, offset: Duration] =
   c.fromNow(slot.start_beacon_time())
 
+proc toDuration*(c: BeaconTime): Duration =
```
arnetheduck: `toDuration` doesn't really make sense - `BeaconTime` is a point in time, you cannot logically convert it to a duration without a reference point - we don't really need `toDuration` since we always have a reference point available when working with time (most of the time, the slot but if need be, we can introduce `GENESIS_BEACON_TIME` just like we have `GENESIS_SLOT`)

beacon_chain/beacon_clock.nim
```diff
@@ -69,19 +73,39 @@ proc fromNow*(c: BeaconClock, t: BeaconTime): tuple[inFuture: bool, offset: Dura
...
+  if c.ns_since_genesis >= 0:
+    chronos.nanoseconds(int64(c.ns_since_genesis))
+  else:
+    chronos.nanoseconds(int64(-c.ns_since_genesis))
+
 proc durationToNextSlot*(c: BeaconClock): Duration =
```
arnetheduck, line 72: Ideally we would remove `durationToNextSlot` and `fromNow` - both of these are compound operations that mix "current time" from the global clock with "time arithmetic" which leads to problems and confusion in code that uses them, since getting the "current time" generally should only be done in a controlled way:

* once per "block" of arithmetic (ie we should not get the current time twice when performing arithmetic because there's an inherent race condition when the clock between the calls)
* not reused between `await` boundaries

therefore:

* remove `fromNow`, `durationToNext*` etc
* use `now` consistently when the current time is needed as the only function that accesses the global clock
* introduce a time difference operation like `durationFrom(a, b: BeaconTime): tuple[future: bool, dur: Duration]` as is done in `fromNow`
* replace all code in both `bn` and `vc` to use the above functions
  * `fromNow` becomes `slot.start_beacon_time().durationFrom(clock.now())`
  * `durationToNextSlot` becomes `let now = clock.now(); saturate((now.slotOrZero() + 1).start_beacon_time().durationFrom(now))` more or less (`slotOrZero` is incorrect for pre-genesis - it needs to return duration to genesis for any time before genesis)
  * the above can be added as helpers that take a `BeaconTime` and do the saturation if need be, but these helpers should never access the clock itself
  * more `durationFrom` can be added for slot vs time etc to make it less verbose

`durationFrom` can also be `durationTo` or some other name or shape - doesn't matter which direction it goes in, but since `Duration` is undefined for negative values, we need to return a tuple - we also _sometimes_ need to know that genesis hasn't happened yet, so we don't always want to saturate.

beacon_chain/nimbus_validator_client.nim
```diff
@@ -93,16 +96,22 @@ proc initValidators(vc: ValidatorClientRef): Future[bool] {.async.} =
...
-  if genesisTime.inFuture:
-    notice "Waiting for genesis", genesisIn = genesisTime.offset
-    await sleepAsync(genesisTime.offset)
+  let
+    res = BeaconClock.init(vc.beaconGenesis.genesis_time)
+    currentTime = res.now()
+    currentSlot = currentTime.slotOrZero()
+    currentEpoch = currentSlot.epoch()
+
+  if currentTime.ns_since_genesis >= 0:
```
arnetheduck: with the above removal of `fromNow`, this code can still use `inFuture` etc by calling `durationFrom`

beacon_chain/validator_client/common.nim
```diff
@@ -883,3 +886,75 @@ proc prepareRegistrationList*(
...
            failure: ApiFailure): ApiNodeFailure =
   ApiNodeFailure(node: node, failure: failure)
+
+proc checkedWaitForSlot*(vc: ValidatorClientRef, destinationSlot: Slot,
+                         offset: TimeDiff,
+                         showLogs: bool): Future[Opt[Slot]] {.async.} =
+  let
+    currentTime = vc.beaconClock.now()
+    currentSlot = currentTime.slotOrZero()
+    chronosOffset = chronos.nanoseconds(
```
arnetheduck, line 928: ditto - `durationFrom` gives `inFuture` so there is no need to revert to `nanoseconds` or other internal implementation details here

beacon_chain/validator_client/attestation_service.nim
```diff
@@ -405,34 +380,67 @@ proc spawnAttestationTasks(service: AttestationServiceRef,
...
-          validators = dutiesSkipped
+  var tasks: seq[Future[void]]
+  try:
+    for index, duties in dutiesByCommittee:
+      tasks.add(service.publishAttestationsAndAggregates(slot, index, duties))
+    let timeout = vc.beaconClock.durationToNextSlot()
+    await allFutures(tasks).wait(timeout)
+  except AsyncTimeoutError:
+    # Cancelling all the pending tasks.
+    let pending = tasks.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
```
arnetheduck, line 391: I see it's functional programming month :)

beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -406,36 +406,69 @@ proc publishSyncMessagesAndContributions(service: SyncCommitteeServiceRef,
     debug "Producing contribution and proofs", delay = delay
   await service.produceAndPublishContributions(slot, beaconBlockRoot, duties)
 
-proc spawnSyncCommitteeTasks(service: SyncCommitteeServiceRef, slot: Slot) =
+proc spawnSyncCommitteeTasks(service: SyncCommitteeServiceRef,
```
arnetheduck:
```suggestion
proc processSyncCommitteeTasks(service: SyncCommitteeServiceRef,
```

beacon_chain/beacon_clock.nim
cheatfate: You think that `BeaconTime` is a point at time, but it is not, because `BeaconTime` is a point in time with adjusted start. So when you convert it to `Duration` it means that you want to calculate distance to the `start`.
This procedure is used in this scenario:
```nim
info "Initializing beacon clock",
     genesis_time = vc.beaconGenesis.genesis_time,
     current_slot = "<n/a>", current_epoch = "<n/a>",
     time_to_genesis = currentTime.toDuration()

notice "Waiting for genesis",
       genesis_time = vc.beaconGenesis.genesis_time,
       time_to_genesis = currentTime.toDuration()
```
So do you want me to but this calculation everywhere?

beacon_chain/nimbus_validator_client.nim
cheatfate: This code do use only `now()` as you asked to do, nothing else. So why should i apply any arithmetic in here if i just trying to obtain current time and trying to understand is this time before genesis or after genesis? Why should i call to arithmetic operation?

beacon_chain/validator_client/common.nim
cheatfate: `chronosOffset` do not related to any `BeaconClock` arithmetic, it just converts `TimeDiff` object. Otherwise it will be 2 calls to `durationTo/durationFrom`.

beacon_chain/beacon_clock.nim
cheatfate: Proposed solution do not support negative Slot values which is happened when you are running before genesis. So it become impossible to provide such UX experience like it was done now. Because its impossible to supply negative Slot values it is very hard to find proper argument value for `durationFrom/durationTo`. Its why `durationToNextSlot` looks the most appropriate name for the procedure which is able to calculate duration to next slot which is either before or after genesis.

beacon_chain/beacon_clock.nim
arnetheduck: well, what this function returns is the "time-since-genesis-or-zero" - it's not a conversion (as `toDuration` implies in its name) which is not meaningful, but rather a time one-sided time-arithmetic-function - its correct signature would be `fromGenesis(t: BeaconTime): tuple[future: bool, dur: Duration]` - then the caller can choose whether to saturate it or not, but at least the name correctly represents what the function does.

Throwing away the sign here will lead to misuse because it relies on the caller to check "future-or-not" before calling this function which is easy to forget - it's better to return the sign and have the caller decide about discarding it.

beacon_chain/nimbus_validator_client.nim
arnetheduck: we've generally avoided the use of specific time granularity in the code (`ns_since_genesis` exposes "nanoseconds") - ie we've strived to not expose the granularity of the clock in code that does time computations, instead using "abstract" types which keep this information as an internal implementation detail.

In fact, I'm not sure why `ns_since_genesis` is public at all - I would generally not have considered it something that callers should use / rely on.

That said, _iff_ we want to use `ns_since_genesis`, it's important to verify that we use it consistently everywhere across the codebase (both bn and vc) - using "slightly different" ways of dealing with time in different parts of the codebase becomes technical debt.

beacon_chain/beacon_clock.nim
```diff
@@ -16,6 +16,10 @@ from times import Time, getTime, fromUnix, `<`, `-`, inNanoseconds
 
 export chronos.Duration, Moment, now
 
+const
+  NANOSECONDS_PER_SLOT* = SECONDS_PER_SLOT * 1_000_000_000'u64
```
arnetheduck: unused now?

cheatfate: https://github.com/status-im/nimbus-eth2/pull/4743/commits/447332fe80dd78e07791231d26b6d9aa284ab382

beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -406,36 +406,68 @@ proc publishSyncMessagesAndContributions(service: SyncCommitteeServiceRef,
...
+    warn "Service crashed with unexpected error", err_name = exc.name,
+         err_msg = exc.msg
+    return
+
+  doAssert(len(vc.forks) > 0, "Fork schedule must not be empty at this point")
 
+  var currentSlot: Opt[Slot]
   while true:
     # This loop could look much more nicer/better, when
     # https://github.com/nim-lang/Nim/issues/19911 will be fixed, so it could
```
arnetheduck, line 455: this bug is fixed now btw

beacon_chain/validator_client/common.nim
```diff
@@ -915,3 +918,75 @@ proc prepareRegistrationList*(
...
+        timeToSlot = destinationSlot.start_beacon_time() - wallTime +
+                     chronosOffset
+        if showLogs:
+          warn "System time adjusted backwards, rescheduling slot actions"
+        continue
+
+    elif wallSlot > destinationSlot + SLOTS_PER_EPOCH:
+      if showLogs:
+        warn "Time moved forwards by more than an epoch, skipping ahead"
+      return Opt.some(wallSlot - SLOTS_PER_EPOCH)
```
arnetheduck: why does this return an "old" slot? the beacon node does this because it "replays" attestations for the old slots but I don't think the VC does? It's also inconsistent with the `elif wallSlot > destination` return

beacon_chain/validator_client/sync_committee_service.nim
cheatfate: Maybe, but i have never seen extended test suite about loops and exceptions in Nim language, so i wish to be safe here.

arnetheduck (approved):

PR #4749, opened 2023-03-19
query EL and builder relay for bids in parallel

tersec:
The VC implementation has tradeoffs, but among them is that it doesn't have the tight coupling this PR moves towards breaking up a bit.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -85,7 +85,8 @@ declarePublicGauge(attached_validator_balance_total,
 logScope: topics = "beacval"
 
 type
-  ForkedBlockResult* = Result[ForkedBeaconBlock, string]
+  # could keep bid value here
+  ForkedBlockResult = Result[ForkedBeaconBlock, string]
 
   SyncStatus* {.pure.} = enum
     synced
@@ -675,12 +676,12 @@ proc getBlindedBlockParts[EPH: ForkyExecutionPayloadHeader](
 
   return ok((executionPayloadHeader.get, forkedBlck))
 
-proc proposeBlockMEV[
+proc proposeBlockMEV1[
     SBBB: bellatrix_mev.SignedBlindedBeaconBlock |
           capella_mev.SignedBlindedBeaconBlock](
     node: BeaconNode, head: BlockRef, validator: AttachedValidator, slot: Slot,
     randao: ValidatorSig, validator_index: ValidatorIndex):
-    Future[Opt[BlockRef]] {.async.} =
+    Future[Result[SBBB, string]] {.async.} =
   # Used by the BN's own validators, but not the REST server
   when SBBB is bellatrix_mev.SignedBlindedBeaconBlock:
     type EPH = bellatrix.ExecutionPayloadHeader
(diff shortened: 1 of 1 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -820,85 +811,106 @@ proc makeBlindedBeaconBlockForHeadAndSlot*[
...
+        info "Payload builder error",
+          slot, head = shortLog(head), validator = shortLog(validator),
+          err = payloadBuilderBidFut.read().error()
+        false
+      else:
+        # Effectively the same case, but without the log message
+        false
+    else:
+      info "Payload builder bid future failed",
+        slot, head = shortLog(head), validator = shortLog(validator)
```
zah:
```suggestion
        slot, head = shortLog(head), validator = shortLog(validator),
        err = payloadBuilderBidFut.error.msg
```

beacon_chain/validators/validator_duties.nim
```diff
@@ -820,85 +811,106 @@ proc makeBlindedBeaconBlockForHeadAndSlot*[
...
+      if engineBlockFut.read().isOk:
+        true
+      else:
+        info "Engine block building error",
+          slot, head = shortLog(head), validator = shortLog(validator),
+          err = payloadBuilderBidFut.read().error()
+        false
+    else:
+      info "Engine block building failed",
+        slot, head = shortLog(head), validator = shortLog(validator)
```
zah:
```suggestion
        slot, head = shortLog(head), validator = shortLog(validator),
        err = engineBlockFut.error.msg
```

beacon_chain/validators/validator_duties.nim
```diff
@@ -820,85 +811,106 @@ proc makeBlindedBeaconBlockForHeadAndSlot*[
...
+      false
+
+  let useBuilderBlock =
+    if builderBidAvailable:
+      if engineBidAvailable:
```
zah, line 886: Both branches of this `if` statement return `true`?

beacon_chain/validators/validator_duties.nim
```diff
@@ -974,6 +986,49 @@ proc proposeBlock(node: BeaconNode,
...
+      let res = await validator.getEpochSignature(
+        fork, genesis_validators_root, slot.epoch)
+      if res.isErr():
+        warn "Unable to generate randao reveal",
+             validator = shortLog(validator), error_msg = res.error()
+        return head
+      res.get()
+
+  template proposeBlockContinuation(type1, type2: untyped): auto =
+    await proposeBlockAux[type1, type2](
```
zah: I'm usually migrating all explicit generic parameters to `typedesc` parameters, because the former have number of shortcomings in the compiler (e.g. limited ability to overload, lack of default values, more bugs).

beacon_chain/validators/validator_duties.nim
tersec: Yes, one of them's the point where the bid comparison will be placed. The other will remain unconditionally `true`, because there's no alternative, because the engine bid didn't arrive.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4749/commits/5d4d13a974090b850036e16354eb03e550b30f0a

PR #4751, opened 2023-03-20
no timeout for purely local async slashing protection and signing

tersec:
alternative to https://github.com/status-im/nimbus-eth2/pull/4750

Removes, for locally held keys, the race entirely


zah:
This looks like the kind of code that somebody will wonder about in the future. A comment with an explanation could help.

tersec:
> This looks like the kind of code that somebody will wonder about in the future. A comment with an explanation could help.

https://github.com/status-im/nimbus-eth2/pull/4764

PR #4765, opened 2023-03-24
VC: Fix beacon-node invalid URL crash.

cheatfate:
Make `beacon-node` URL argument less strict and more UX friendly.

### DIFF
--- a/beacon_chain/validator_client/common.nim
+++ b/beacon_chain/validator_client/common.nim
@@ -492,26 +492,58 @@ proc parseRoles*(data: string): Result[set[BeaconNodeRole], cstring] =
       return err("Invalid beacon node role string found")
   ok(res)
 
+proc normalizeUri*(remoteUri: Uri): Uri =
+  var r = remoteUri
+  if (len(r.scheme) == 0) and (len(r.username) == 0) and
+     (len(r.password) == 0) and (len(r.hostname) == 0) and
+     (len(r.port) == 0) and (len(r.path) > 0):
+    # When `scheme` is not specified and `port` is not specified - whole
+    # hostname is stored in `path`.`query` and `anchor` could still be present.
+    # test.com
+    # test.com?q=query
+    # test.com?q=query#anchor=anchor
+    parseUri("http://" & $remoteUri & ":" & $defaultEth2RestPort)
+  elif (len(r.scheme) > 0) and (len(r.username) == 0) and
+       (len(r.password) == 0) and (len(r.hostname) == 0) and
+       (len(r.port) == 0) and (len(r.path) > 0):
+    # When `scheme` is not specified but `port` is specified - whole
+    # hostname is stored in `scheme` and `port` is in `path`.
+    # 192.168.0.1:5052
+    # test.com:5052
+    # test.com:5052?q=query
+    # test.com:5052?q=query#anchor=anchor
+    parseUri("http://" & $remoteUri)
+  elif (len(r.scheme) > 0) and (len(r.hostname) > 0):
+    # When `scheme` is specified, but `port` is not we use default.
+    # http://192.168.0.1
+    # http://test.com
+    # http://test.com?q=query
+    # http://test.com?q=query#anchor=anchor
+    if len(r.port) == 0: r.port = $defaultEth2RestPort
+    r
+  else:
+    remoteUri
+
 proc init*(t: typedesc[BeaconNodeServerRef], remote: Uri,
            index: int): Result[BeaconNodeServerRef, string] =
   doAssert(index >= 0)
   let
     flags = {RestClientFlag.CommaSeparatedArray}
+    remoteUri = normalizeUri(remote)
     client =
       block:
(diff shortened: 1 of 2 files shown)


beacon_chain/validator_client/common.nim
```diff
@@ -492,26 +492,58 @@ proc parseRoles*(data: string): Result[set[BeaconNodeRole], cstring] =
...
+  var r = remoteUri
+  if (len(r.scheme) == 0) and (len(r.username) == 0) and
+     (len(r.password) == 0) and (len(r.hostname) == 0) and
+     (len(r.port) == 0) and (len(r.path) > 0):
+    # When `scheme` is not specified and `port` is not specified - whole
+    # hostname is stored in `path`.`query` and `anchor` could still be present.
+    # test.com
+    # test.com?q=query
+    # test.com?q=query#anchor=anchor
+    parseUri("http://" & $remoteUri & ":" & $defaultEth2RestPort)
```
tersec, line 505: Regarding `defaultEth2RestPort` 
```nim
  # This is not part of the spec! But it's port which Lighthouse uses
  defaultEth2RestPort* = 5052
  defaultEth2RestPortDesc* = $defaultEth2RestPort
```
Not sure it's entirely worthwhile to enshrine it even more.

tersec:
Is there a specific issue this fixes (e.g., people reporting issues using configurations they expected to work but did not, or consistency with other VCs)?

beacon_chain/validator_client/common.nim
```diff
@@ -492,26 +492,58 @@ proc parseRoles*(data: string): Result[set[BeaconNodeRole], cstring] =
...
+  elif (len(r.scheme) > 0) and (len(r.username) == 0) and
+       (len(r.password) == 0) and (len(r.hostname) == 0) and
+       (len(r.port) == 0) and (len(r.path) > 0):
+    # When `scheme` is not specified but `port` is specified - whole
+    # hostname is stored in `scheme` and `port` is in `path`.
+    # 192.168.0.1:5052
+    # test.com:5052
+    # test.com:5052?q=query
+    # test.com:5052?q=query#anchor=anchor
+    parseUri("http://" & $remoteUri)
```
tersec, line 515: It's possible that BNs' REST ports are almost always unencrypted HTTP, practically. Not sure if this is the case, but wouldn't be surprised.

However, applying this `normalizeUri` approach to the other URLs a VC might use -- e.g., a relay URL -- seems to be maybe assume relatively more. https://ethstaker.cc/mev-relay-list/ for example shows all HTTPS URLs.

cheatfate: If you use fully-qualified URL `normalizeUri` will not apply any changes... And if you going to use HTTPS i think its better to use "https://" notation.

cheatfate:
> Is there a specific issue this fixes (e.g., people reporting issues using configurations they expected to work but did not, or consistency with other VCs)?

Because VC supports multiple BNs it should ignore invalid or dead hosts in BNs list, right now VC will crash it will find such type of BN, for example:
```nimbus_validator_client--beacon-node=http://127.0.0.1:2025 --beacon-node=localhost:2025```
will crash.

beacon_chain/validator_client/common.nim
cheatfate: This is our default, and i'm going to use it if user did not specified it explicitly, or you think `80` is better?

tersec: I'm unsure if it's on balance beneficial to have any default port, given that the specs don't have one. The ecosystem is built to be interoperable between different clients, which have no obligation to have BNs defaulting to 5052.

tersec:
> > Is there a specific issue this fixes (e.g., people reporting issues using configurations they expected to work but did not, or consistency with other VCs)?
> 
> Because VC supports multiple BNs it should ignore invalid or dead hosts in BNs list, right now VC will crash it will find such type of BN, for example: `nimbus_validator_client--beacon-node=http://127.0.0.1:2025 --beacon-node=localhost:2025` will crash.

Which results in
> Error: unhandled exception: Trying to access value with err Result: Unable to resolve remote hostname [ResultDefect]

That seems distinct from assuming any particular default protocol or port, though. Of course the VC shouldn't trigger a `Defect` in this situation, but that doesn't address, e.g. default ports.

cheatfate:
@tersec i though you asking what type of issue this PR fixes, no?

beacon_chain/validator_client/common.nim
arnetheduck: > have BNs defaulting to 5052.

the aim of the default is to make nimbus bn and nimbus vc work together with minimal ceremony - anyone else that happens to follow it is a bonus

tersec: Okay, in which case having a default port is reasonable and having 5052 be that default port is reasonable.

tersec:
> @tersec i though you asking what type of issue this PR fixes, no?

Sure, and that was answered. So with that addressed, no particular objections.

tersec (approved):

PR #4775, opened 2023-03-30
web3signer refactoring and test suite.

cheatfate:


beacon_chain/nimbus_signing_node.nim
```diff
@@ -231,143 +192,301 @@ proc installApiHandlers*(node: SigningNode) =
...
-    1
+proc asyncInit(sn: SigningNodeRef) {.async.} =
+  notice "Launching signing node", version = fullVersionStr,
+         cmdParams = commandLineParams(), config = sn.config
 
-proc getRouter*(): RestRouter =
-  RestRouter.init(validate)
+  info "Initializaing validators", path = sn.config.validatorsDir()
+  var keysList: seq[string]
+  for keystore in listLoadableKeystores(sn.config, sn.keystoreCache):
```
zah: Isn't this the same logic as the `reload` operation above? You can introduce a proc for this.

beacon_chain/nimbus_signing_node.nim
```diff
@@ -231,143 +192,301 @@ proc installApiHandlers*(node: SigningNode) =
...
+
+          let
+            proof = request.proofs.get()[0]
+            feeRecipientRoot =
+              hash_tree_root(distinctBase(
+                node.config.expectedFeeRecipient.get()))
+
+          if not(is_valid_merkle_branch(feeRecipientRoot, proof.merkleProofs,
+                                        log2trunc(proof.index),
+                                        get_subtree_index(proof.index),
```
zah: You also need to check that `proof.index` has the expected value.
Otherwise, I'll be able to send a merkle proof for another field (e.g. the `graffitti` of the block) where I can put the expected fee recipient value in order to trick you.

beacon_chain/spec/eth2_apis/rest_remote_signer_calls.nim
```diff
@@ -70,132 +84,216 @@ proc signDataPlain*(identifier: ValidatorPubKey,
...
+
+  let
+    startSignMoment = Moment.now()
+    response =
+      try:
+        let
+          res = await client.signDataPlain(identifier, body,
+                                           restAcceptType = "application/json")
+          duration = Moment.now() - startSignMoment
+        nbc_remote_signer_time.observe(float(milliseconds(duration)) / 1000.0)
```
zah: We have a naming convention where duration metrics use "_duration" as a suffix (although I can see this is not a new metric).

beacon_chain/validators/validator_pool.nim
```diff
@@ -435,6 +443,55 @@ proc signData(v: AttachedValidator,
   else:
     v.signWithDistributedKey(request)
 
+proc getFeeRecipientProof(blck: ForkedBeaconBlock | ForkedBlindedBeaconBlock |
```
zah, line 446: The plan was to propose a more generalized solution where a list of generalized indices is specified in the remote keystore configuration, but we can deliver this in a follow-up PR.

beacon_chain/validators/validator_pool.nim
```diff
@@ -413,17 +419,19 @@ proc signWithDistributedKey(v: AttachedValidator,
...
-                             WEB3_SIGNER_DELAY_TOLERANCE):
-    return SignatureResult.err "Timeout"
-  if res.isErr:
-    return SignatureResult.err res.error
+  let
+    deadline = sleepAsync(WEB3_SIGNER_DELAY_TOLERANCE)
+    (client, info) = v.clients[0]
+    res = await client.signData(info.pubkey, deadline, 2, request)
+
+  if not(deadline.finished()): await cancelAndWait(deadline)
```
zah, line 430: Why is it useful to cancel simple timers?
That's a genuine question: I'm not currently cancelling timers in my own code.

beacon_chain/spec/eth2_apis/rest_remote_signer_calls.nim
```diff
@@ -70,132 +84,216 @@ proc signDataPlain*(identifier: ValidatorPubKey,
...
+      operationFut = signData(client, identifier, body)
+      if isNil(timerFut):
+        await allFutures(operationFut)
+      else:
+        discard await race(timerFut, operationFut)
+    except CancelledError as exc:
+      var pending: seq[Future[void]]
+      if not(operationFut.finished()):
+        pending.add(operationFut.cancelAndWait())
+      await allFutures(pending)
```
zah: `allFutures` can be called here with an empty sequence here. Is this a reasonable thing to do? Isn't this whole block unnecessarily complicated? It seems to me that you can just `await` the `operationFut` future when it's not finished.

beacon_chain/spec/eth2_apis/rest_remote_signer_calls.nim
```diff
@@ -70,132 +84,216 @@ proc signDataPlain*(identifier: ValidatorPubKey,
...
+          # Number of attempts exceeded, so we return result we have.
+          return resp
+        else:
+          # We have some attempts left, so we show debug log about current
+          # attempt
+          debug "Unable to get signature using remote signer",
+                kind = resp.error.kind, reason = resp.error.message,
+                attempts_count = attemptsCount, attempt = attempt
+          lastError = Opt.some(resp.error)
+          inc(attempt)
```
zah, line 295: A little bit of back-off would be appropriate here perhaps.

cheatfate: I'm not sure i understand.

beacon_chain/validators/validator_pool.nim
cheatfate: Timers are not cheap right now, so it should be cancelled as well as other futures.

beacon_chain/validators/validator_pool.nim
cheatfate: Right now the only change we do in remote keystore configuration is is type of remote keystore.

beacon_chain/spec/eth2_apis/rest_remote_signer_calls.nim
zah: I meant waiting for an increasing number of miliseconds before each retry:

https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/implement-retries-exponential-backoff

beacon_chain/validators/validator_pool.nim
zah: Can you clarify what's being saved by cancellation? I've taken a look at how `cancel` is implemented and it merely clears the `completion` function associated with the timer future (which in turn just calls `timerFut.complete()`).

cheatfate: Yep, right now its not very effective, but as soon as we will have more effective implementation your code will start leaking timers and this leaks will be pretty hidden inside the code.

arnetheduck: > clears the completion function

this btw has a tendency to save memory over time in general - closures have a tendency to keep futures alive longer than they need be, so being hygienic about clearing "obsolete" stuff also helps the GC

cheatfate:
@zah whole test suite missing distributive signing tests, because i dont know how to create distributed keys set.

zah:
There is an example here:

https://github.com/status-im/nimbus-eth2/blob/6c0d756d54998cf30edd1678cbf121c8052a378a/beacon_chain/validators/keystore_management.nim#L1458

Check out the use of the variable named `threshold` throughout this file in general.

PR #4779, opened 2023-03-31
Adjust remote signer idle connection timeout defaults.

cheatfate:
We adjust our defaults to defaults of original Web3Signer software https://docs.web3signer.consensys.net/Reference/CLI/CLI-Syntax#idle-connection-timeout-seconds

### DIFF
--- a/beacon_chain/validators/validator_pool.nim
+++ b/beacon_chain/validators/validator_pool.nim
@@ -27,10 +27,10 @@ export
 
 const
   WEB3_SIGNER_DELAY_TOLERANCE = 3.seconds
-  WEB3_SIGNER_DEFAULT_TIMEOUT = 29.seconds
-    # We make it compatible with original Web3Signer defaults
+  WEB3_SIGNER_DEFAULT_TIMEOUT = (int64(SECONDS_PER_SLOT) + 1).seconds
+    # This timeout value should not be greater than default value specified at:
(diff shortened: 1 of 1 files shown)


tersec (approved):

beacon_chain/validators/validator_pool.nim
```diff
@@ -27,6 +27,12 @@ export
 
 const
   WEB3_SIGNER_DELAY_TOLERANCE = 3.seconds
+  WEB3_SIGNER_DEFAULT_TIMEOUT = 29.seconds
```
arnetheduck: why not make this ... SECONDS_PER_SLOT + 1? ie it's not _that_ expensive to open a new connection, but trying to use a stale one causes a lot of problems - I can see even a 2s timeout making more sense for http connection pooling by default, ie if the connection is not immediately useful, it could simply be dropped - the problems caused are typically very annoying..

cheatfate: Its not that expensive when connection is HTTP one, but pretty expensive when HTTPS.

cheatfate: And you forgot that servers do not like a lot of connections from one IP too...

arnetheduck: well - if we close connections more aggressively, we will have fewer connections to deal with - and yes, I know ssl takes some time, but .. I see connection recycling as a way to handle bursts of requests - after that, the returns significantly diminish while the number of problems we see go up..

PR #4782, opened 2023-04-03
Quarantine.cleanupOrphans(): Delete unviable orphan

henridf:
@arnetheduck: in cleanupOrphans, we add unviable orphans to the unviable table. Shouldn't we also delete them from the orphans table?


arnetheduck:
seems reasonable :)

PR #4785, opened 2023-04-03
Add blob sidecar gossip validation

henridf:
In this PR, blob sidecars are validated but not yet sent further up the stack.

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -288,8 +288,6 @@ proc processSignedBlobSidecar*(
 
   v
 
-
-
 proc setupDoppelgangerDetection*(self: var Eth2Processor, slot: Slot) =
   # When another client's already running, this is very likely to detect
   # potential duplicate validators, which can trigger slashing.
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -41,6 +41,10 @@ declareCounter beacon_blocks_received,
   "Number of valid blocks processed by this node"
 declareCounter beacon_blocks_dropped,
   "Number of invalid blocks dropped by this node", labels = ["reason"]
+declareCounter blob_sidecars_received,
+  "Number of valid blobs processed by this node"
+declareCounter blob_sidecars_dropped,
+  "Number of invalid blobs dropped by this node", labels = ["reason"]
 declareCounter beacon_attester_slashings_received,
   "Number of valid attester slashings processed by this node"
 declareCounter beacon_attester_slashings_dropped,
@@ -84,6 +88,9 @@ declareHistogram beacon_aggregate_delay,
 
 declareHistogram beacon_block_delay,
   "Time(s) between slot start and beacon block reception", buckets = delayBuckets
+
+declareHistogram blob_sidecar_delay,
+  "Time(s) between slot start and blob sidecar reception", buckets = delayBuckets
 
 type
   DoppelgangerProtection = object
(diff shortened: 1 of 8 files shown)


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -246,6 +253,43 @@ proc processSignedBeaconBlock*(
...
+
+    blob_sidecars_received.inc()
+    blob_sidecar_delay.observe(delay.toFloatSeconds())
+  else:
+    debug "Dropping blob", error = v.error()
+
+    blob_sidecars_dropped.inc(1, [$v.error[0]])
+
+  v
+
```
tersec, line 290: extra newlines

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1478,6 +1478,13 @@ proc installMessageValidators(node: BeaconNode) =
           toValidationResult(node.processor[].processSignedBeaconBlock(
             MsgSource.gossip, signedBlock)))
 
+    for i in 1..MAX_BLOBS_PER_BLOCK:
+      node.network.addValidator(
+        getBlobSidecarTopic(forkDigests.deneb, i),
+        proc (signedBlobSidecar: deneb.SignedBlobSidecar): ValidationResult =
+            toValidationResult(node.processor[].processSignedBlobSidecar(
+              MsgSource.gossip, signedBlobSidecar, i)))
```
tersec: This will always call `processSignedBlobSidecar` with `i` being `MAX_BLOBS_PER_BLOCK` because of how closures capture variables in Nim.

```nim
var vals1: seq[proc() {.gcsafe.}]
var vals2: seq[proc() {.gcsafe.}]
var vals3: seq[proc() {.gcsafe.}]

proc fakeAddValidator(s: var auto, x: auto) =
  s.add x

for i in 1..10:
  vals1.fakeAddValidator(
    proc() = echo i)

echo "Without closureScope"

for closure in vals1:
  closure()  # always prints 10

for i in 1..10:
  closureScope:
    vals2.fakeAddValidator(
      proc() = echo i)

echo "With closureScope but no variable reassignment within closureScope"

for closure in vals2:
  closure()  # still always prints 10

for i in 1..10:
  closureScope:
    let closureCaptured = i   # inside closureScope
    vals3.fakeAddValidator(
      proc() = echo closureCaptured)

# This has finally captured the loop index in a variable per iteration

for closure in vals3:
  closure()  # now it prints 1 .. 10
```

https://github.com/status-im/nimbus-eth2/blob/6a0a5c03ed0531f9fe9fc906a5839acc73693033/beacon_chain/nimbus_beacon_node.nim#L924-L927
https://github.com/status-im/nimbus-eth2/blob/6a0a5c03ed0531f9fe9fc906a5839acc73693033/beacon_chain/nimbus_beacon_node.nim#L1389-L1399
https://github.com/status-im/nimbus-eth2/blob/6a0a5c03ed0531f9fe9fc906a5839acc73693033/beacon_chain/nimbus_beacon_node.nim#L1481-L1491
show other examples of this.

https://nim-lang.org/docs/system.html#closureScope.t%2Cuntyped explains/documents this more.

beacon_chain/spec/network.nim
```diff
@@ -97,6 +97,11 @@ func getSyncCommitteeContributionAndProofTopic*(forkDigest: ForkDigest): string
   ## For subscribing and unsubscribing to/from a subnet.
   eth2Prefix(forkDigest) & "sync_committee_contribution_and_proof/ssz_snappy"
 
+# https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.5/specs/deneb/p2p-interface.md#blob_sidecar_index
+func getBlobSidecarTopic*(forkDigest: ForkDigest,
```
tersec, line 101: Missing a test in
https://github.com/status-im/nimbus-eth2/blob/6a0a5c03ed0531f9fe9fc906a5839acc73693033/tests/test_honest_validator.nim#L19-L29

tersec: Or, rather, it's missing one of the tests which includes testing the index, examples of which one can find also in `test_honest_validators`.

henridf: 👍 [e0c861f](https://github.com/status-im/nimbus-eth2/pull/4785/commits/e0c861fc0243a44e96faa59d12dbfea40481a0eb)

beacon_chain/nimbus_beacon_node.nim
henridf: ah thanks! 
ca15151

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1478,6 +1478,15 @@ proc installMessageValidators(node: BeaconNode) =
           toValidationResult(node.processor[].processSignedBeaconBlock(
             MsgSource.gossip, signedBlock)))
 
+    for i in 1..MAX_BLOBS_PER_BLOCK:
```
```suggestion
    for i in 0 ..< MAX_BLOBS_PER_BLOCK:
```

According to https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.5/specs/deneb/validator.md#constructing-the-signedblobsidecars:
```python
def get_blob_sidecars(block: BeaconBlock,
                      blobs: Sequence[Blob],
                      blob_kzg_proofs: Sequence[KZGProof]) -> Sequence[BlobSidecar]:
    return [
        BlobSidecar(
            block_root=hash_tree_root(block),
            index=index,
            slot=block.slot,
            block_parent_root=block.parent_root,
            blob=blob,
            kzg_commitment=block.body.blob_kzg_commitments[index],
            kzg_proof=blob_kzg_proofs[index],
        )
        for index, blob in enumerate(blobs)
    ]
```
which are 0-based, not 1-based.

henridf: thanks, good catch

tersec (approved):

Issue #4787, opened 2023-04-04
Slashed validators are counted in the required stability subnets at startup


Menduist:
**Describe the bug**
At startup, we add every validators, including slashed ones in the `knownValidators` validators set.
This causes the nodes to subscribe to a lot of stability subnets
When the stability subnets get refreshed, they slowly decrease since the slashed validators are not longer counted

**Screenshots**

https://metrics.status.im/d/pgeNfj2Wz23/nimbus-fleet-testnets?orgId=1&var-instance=linux-05.he-eu-hel1.nimbus.prater&var-container=beacon-node-prater-libp2p&from=1680278988225&to=1680550202130
![image](https://user-images.githubusercontent.com/13471753/229764899-31a908cc-62af-4bef-8960-a8916ce3a114.png)

Node with 43 slashed validators, 17 active.

tersec:
https://github.com/status-im/nimbus-eth2/pull/4793

PR #4793, opened 2023-04-06
only add active validators to knownValidators at startup

tersec:
Fix https://github.com/status-im/nimbus-eth2/issues/4787

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -414,8 +414,13 @@ proc initFullNode(
         node.validatorMonitor[].addMonitor(validator.pubkey, validator.index)
 
       if validator.index.isSome():
-        node.consensusManager[].actionTracker.knownValidators[
-          validator.index.get()] = wallSlot
+        withState(dag.headState):
+          let idx = validator.index.get()
+          if distinctBase(idx) <= forkyState.data.validators.lenu64:
+            template v: auto = forkyState.data.validators.item(idx)
+            if  is_active_validator(v, wallSlot.epoch) or
+                is_active_validator(v, wallSlot.epoch + 1):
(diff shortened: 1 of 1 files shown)


arnetheduck (approved):

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -414,8 +414,13 @@ proc initFullNode(
...
       if validator.index.isSome():
-        node.consensusManager[].actionTracker.knownValidators[
-          validator.index.get()] = wallSlot
+        withState(dag.headState):
+          let idx = validator.index.get()
+          if distinctBase(idx) <= forkyState.data.validators.lenu64:
+            template v: auto = forkyState.data.validators.item(idx)
+            if  is_active_validator(v, wallSlot.epoch) or
+                is_active_validator(v, wallSlot.epoch + 1):
+              node.consensusManager[].actionTracker.knownValidators[idx] = wallSlot
```
arnetheduck, line 423: we should log a notice / warning here that the validator is exited

tersec: Might not be exited but not yet active, too. The window there is smaller because most of the time it's not active on that side of the timeline, it won't have a validator index yet, but there can be some window after the deposit contract picks it up but before the churn rate limits let it become active.

The check here is deliberately broad. The actual Nimbus mechanism cares about active-or-not, not why it might not be active, which is a bit irrelevant in determining whether it should be asking the action tracker to schedule actions.

arnetheduck: >  The actual Nimbus mechanism cares about active-or-not,

the user cares about knowing about the state of their validator - ie the logging can differentiate between reasons, once it has figured out it's not active

tersec: https://github.com/status-im/nimbus-eth2/pull/4812

Issue #4084, opened 2022-09-07
I'm wondering if nimbus plans to add this for Mev Boost? 


YikesMcSurfFace:
https://bloxroutelabs.medium.com/eth-2-0-validators-capturing-mev-cecdf01f6be6

Lighthouse/prysm have already added it, wondering if nimbus plans to do the same for Mevboost. Pardon me I am not a dev.

<img width="521" alt="image" src="https://user-images.githubusercontent.com/113007714/188802435-4aceeee1-dcdc-4f5a-8d74-f446d8a739be.png">

tersec:
Yes, this is intended at some point.

tersec:
https://writings.flashbots.net/the-cost-of-resilience/ documents some reasonable default thresholds

tersec:
https://github.com/status-im/nimbus-eth2/pull/4847

tersec:
https://github.com/status-im/nimbus-eth2/pull/4847

PR #4795, opened 2023-04-07
use higher of available engine and builder API bids

tersec:


zah:
IIRC, Lighthouse has a special configuration parameter allowing the user to specify preference for local blocks by requiring the blinded block value to exceed the local block value by a specific amount before it's being selected. Something like this can be implemented in a follow-up PR.

tersec:
> IIRC, Lighthouse has a special configuration parameter allowing the user to specify preference for local blocks by requiring the blinded block value to exceed the local block value by a specific amount before it's being selected. Something like this can be implemented in a follow-up PR.

Yes, we have an issue https://github.com/status-im/nimbus-eth2/issues/4084 on that and I deliberately kept that out of this PR, to avoid bikeshedding delay/risk. It's a good idea, and now trivial to implement: if using a builder API relay would only net negligible increased profit, but might fail and lose the entire block/payload, when one already has a perfectly good engine API block/payload, then, it's not a good idea to use the builder block and risk proposal failure.

PR #4808, opened 2023-04-11
Quarantine and reassembly of gossiped blobs and blocks

henridf:
This PR adds support for reassembly of blobs and blocks received through gossip:

- A new component BlobQuarantine, that is used to store blobs pending
  receiving all blobs for a block.
- The existing block quarantine is augmented to store blocks in a
  "blobless" bucket - these are blocks that have blob transactions,
  and are awaiting blobs before being sent up the stack.

- Upon receiving a new blob from gossip, it is stored in blob
  quarantine. If the corresponding block is already in "blobless"
  block quarantine, and if all blobs for that block are present in
  blob quarnatine, then they are pulled out of quarantine and sent
  along for local processing.

- Upon receiving a new block from gossip, if it has blob transactions
  and all blobs are in blob quarantine, they are pulled out and the
  block+blobs is sent along for local processing. Otherwise, the block
  is added to the "blobless" block quarantine bucket.

- When an orphan becomes unorphaned, rather than immediately sending
  it to local processing, we first check if it has blob
  transactions. If so, and if the blobs are not present in quarantine,
  it is moved to the blobless bucket. If the blobs are present, they
  are pulled out, and the block+blobs can be sent to local processing.

Still remaining:
  - Periodically initiate blob requests for "blobless" blocks (analogous to the existing "handleMissingBlocks")
  - Storing blobs
  - Proposals
  - Sync

### DIFF
--- a/beacon_chain/consensus_object_pools/blob_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/blob_quarantine.nim
@@ -1,0 +1,70 @@
+# beacon_chain
+# Copyright (c) 2018-2023 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+import
+  std/tables,
+  ../spec/datatypes/deneb
+
+
+const
+  MaxBlobs = SLOTS_PER_EPOCH * MAX_BLOBS_PER_BLOCK
+
+
+type
+  BlobQuarantine* = object
+    blobs*: Table[(Eth2Digest, BlobIndex), ref BlobSidecar]
+
+
+func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
+  if quarantine.blobs.lenu64 > MaxBlobs:
+    return
+  discard quarantine.blobs.hasKeyOrPut((blobSidecar.block_root,
+                                        blobSidecar.index), blobSidecar)
+
+func blobIndices*(quarantine: BlobQuarantine, digest: Eth2Digest):
+     seq[BlobIndex] =
+  var r: seq[BlobIndex] = @[]
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    if quarantine.blobs.hasKey((digest, i)):
+      r.add(i)
+  r
+
+func hasBlob*(quarantine: BlobQuarantine, blobSidecar: BlobSidecar) : bool =
+  quarantine.blobs.hasKey((blobSidecar.block_root, blobSidecar.index))
+
+func popBlobs*(quarantine: var BlobQuarantine, digest: Eth2Digest):
+     seq[ref BlobSidecar] =
+  var r: seq[ref BlobSidecar] = @[]
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    var b: ref BlobSidecar
+    if quarantine.blobs.pop((digest, i), b):
+      r.add(b)
+  r
+
+func peekBlobs*(quarantine: var BlobQuarantine, digest: Eth2Digest):
+     seq[ref BlobSidecar] =
+  var r: seq[ref BlobSidecar] = @[]
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    quarantine.blobs.withValue((digest, i), value):
+      r.add(value[])
+  r
+
+func removeBlobs*(quarantine: var BlobQuarantine, digest: Eth2Digest) =
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    quarantine.blobs.del((digest, i))
+
+func hasBlobs*(quarantine: BlobQuarantine, blck: deneb.SignedBeaconBlock):
+     bool =
+  let idxs = quarantine.blobIndices(blck.root)
+  if len(blck.message.body.blob_kzg_commitments) != len(idxs):
+    return false
+  for i in 0..len(idxs):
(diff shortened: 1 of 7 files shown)


beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -0,0 +1,67 @@
...
+const
+  MaxBlobs = SLOTS_PER_EPOCH * MAX_BLOBS_PER_BLOCK
+
+
+type
+  BlobQuarantine* = object
+    blobs*: Table[(Eth2Digest, BlobIndex), ref BlobSidecar]
+
+
+func init*(T: type BlobQuarantine): T =
```
tersec: Completely empty/default `init` functions probably aren't that useful, since Nim has well-defined semantics for default object construction, and no real risk of uninitialized memory or anything similar, because it's compiler-enforced unless one uses the `.noInit.` pramga.

It's how, e.g., `StateCache` is used -- there's no explicit `init` for it, it's just created as `StateCache()` or `default(StateCache)`.

If this ever grows a parameter, it's easy enough to refactor its one or two callers. So my preference would be to leave out `BlobQuarantine.init()` as it exists now.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -0,0 +1,67 @@
...
+
+func hasBlob*(quarantine: BlobQuarantine, blobSidecar: BlobSidecar) : bool =
+  return quarantine.blobs.hasKey((blobSidecar.block_root, blobSidecar.index))
+
+func popBlobs*(quarantine: var BlobQuarantine, digest: Eth2Digest):
+     seq[ref BlobSidecar] =
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    var b: ref BlobSidecar
+    if quarantine.blobs.pop((digest, i), b):
+      result.add(b)
```
tersec: https://status-im.github.io/nim-style-guide/language.result.html
>
> Use expression-based return or explicit `return` keyword with a value

With some explanation of why.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -0,0 +1,67 @@
...
+
+func peekBlobs*(quarantine: var BlobQuarantine, digest: Eth2Digest):
+     seq[ref BlobSidecar] =
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    quarantine.blobs.withValue((digest, i), value):
+      result.add(value[])
```
tersec: Also https://status-im.github.io/nim-style-guide/language.result.html

beacon_chain/consensus_object_pools/blob_quarantine.nim
henridf: This comes straight from block_quarantine.nim. I'll change that too, then.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -0,0 +1,67 @@
...
+type
+  BlobQuarantine* = object
+    blobs*: Table[(Eth2Digest, BlobIndex), ref BlobSidecar]
+
+
+func init*(T: type BlobQuarantine): T =
+  T()
+
+func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
+  if quarantine.blobs.len.uint64 > MaxBlobs:
```
tersec:
```suggestion
  if quarantine.blobs.lenu64 > MaxBlobs:
```

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -0,0 +1,67 @@
...
+                                        blobSidecar.index), blobSidecar)
+
+func blobIndices*(quarantine: BlobQuarantine, digest: Eth2Digest):
+     seq[BlobIndex] =
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    if quarantine.blobs.hasKey((digest, i)):
+      result.add(i)
+
+func hasBlob*(quarantine: BlobQuarantine, blobSidecar: BlobSidecar) : bool =
+  return quarantine.blobs.hasKey((blobSidecar.block_root, blobSidecar.index))
```
tersec: Much less problematic than `result`, but since the control flow is so simple here, the note from https://status-im.github.io/nim-style-guide/language.result.html#practical-notes that
> "expression" returns guarantee that all code branches produce one (and only one) value to be returned

means that it's typically the best default, when straightforward enough, as here, versus explicit `return`.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -0,0 +1,67 @@
...
+  if quarantine.blobs.len.uint64 > MaxBlobs:
+    return
+  discard quarantine.blobs.hasKeyOrPut((blobSidecar.block_root,
+                                        blobSidecar.index), blobSidecar)
+
+func blobIndices*(quarantine: BlobQuarantine, digest: Eth2Digest):
+     seq[BlobIndex] =
+  for i in 0..MAX_BLOBS_PER_BLOCK-1:
+    if quarantine.blobs.hasKey((digest, i)):
+      result.add(i)
```
tersec: Another https://status-im.github.io/nim-style-guide/language.result.html

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: Sure, sounds good.

henridf: > This comes straight from block_quarantine.nim. I'll change that too, then.

Actually not doing that here as it requires touching a bunch of unrelated test files.

tersec: Makes sense, also fine.

tersec (approved):

PR #4814, opened 2023-04-12
include small dedup in block processor to handle blockByRoot blocks

tersec:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -42,6 +43,10 @@ const
     ## syncing the finalized part of the chain
   PAYLOAD_PRE_WALL_SLOTS = SLOTS_PER_EPOCH * 2
     ## Number of slots from wall time that we start processing every payload
+  MAX_DEDUP_QUEUE_LEN = 16
```
arnetheduck, line 50: why not simply an (unbounded) hash set? ie this is only used for "in-flight" stuff anyway which is structurally bounded by other mechanisms - if we had a leak there, we'd have significantly more problems than this check

arnetheduck (approved):

beacon_chain/gossip_processing/block_processor.nim
tersec: Because I want to avoid unbounded resource usage in anything, if there's not a strong argument for it, and if this doesn't prove sufficient, can/will pursue something with more unbounded potential downside too.

One specific architectural constraint it adds is that it means ensure all the exits of "in-flight" are covered, which, while in theory should be already feasible, is an unnecessary risk.

PR #4816, opened 2023-04-13
add `formatIt` for `ForkedLightClientXyz`

etan-status:
Even though the `ForkyLightClientXyz` have `formatIt`, they do not apply when logging `ForkedLightClientXyz`, leading to large logs at times. Defining `formatIt` for `ForkedLightClientXyz` fixes this.


etan-status:
Example of log that is now much more compact:
largelog

tersec (approved):

PR #4817, opened 2023-04-13
always use fcUV2 in shapella even for non-proposer fcUs

tersec:


beacon_chain/beacon_node_light_client.nim
```diff
@@ -60,15 +60,32 @@ proc initLightClient*(
...
-                payloadAttributes = NoPayloadAttributes)
+
+              template callForkchoiceUpdated(attributes: untyped) =
+                discard await node.elManager.forkchoiceUpdated(
+                  headBlockHash = blckPayload.block_hash,
+                  safeBlockHash = beaconHead.safeExecutionPayloadHash,
+                  finalizedBlockHash = beaconHead.finalizedExecutionPayloadHash,
+                  payloadAttributes = none attributes)
+
+              case node.dag.cfg.consensusForkAtEpoch(beaconHead.blck.bid.slot.epoch)
```
etan-status, line 76: `beaconHead.blck` is `nil` so this fails

etan-status: Should this be `blck.message.slot`, or simply based on wallSlot? @tersec

tersec: It's based on the timestamp of the referenced block

tersec: https://github.com/status-im/nimbus-eth2/pull/4835

PR #4821, opened 2023-04-14
remove empty block fallback now that capella's on mainnnet

tersec:


tersec:
jenkins/nimbus-eth2/macos/aarch64 runner is out of disk.

tests/testblockutil.nim
```diff
@@ -108,6 +108,42 @@ proc build_empty_merge_execution_payload(state: bellatrix.BeaconState):
   bellatrix.ExecutionPayloadForSigning(executionPayload: payload,
                                        blockValue: Wei.zero)
 
+from eth/common/eth_types import GasInt
+from eth/eip1559 import calcEip1599BaseFee
+from stew/saturating_arith import saturate
+
+proc build_empty_execution_payload(
```
arnetheduck, line 115: do we need this still then?

arnetheduck (approved):

tests/testblockutil.nim
tersec: https://github.com/status-im/nimbus-eth2/blob/8f2f40f0afafaae4c87620e3ad4c15ed84ae2cfe/tests/test_blockchain_dag.nim#L1066-L1111
uses this capability, and deliberately only relies on Bellatrix (not Capella or Deneb) blocks as a result.

PR #4822, opened 2023-04-15
Fix incorrect handling of blobless blocks

henridf:
We were passing all blocks to local processing, including those for which we were missing blobs. This commit fixes that.

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -249,6 +249,7 @@ proc processSignedBeaconBlock*(
            blockRoot = shortLog(signedBlock.root),
            blck = shortLog(signedBlock.message),
            signature = shortLog(signedBlock.signature)
+        return v
 
     self.blockProcessor[].addBlock(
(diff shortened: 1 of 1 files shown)


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -242,6 +242,13 @@ proc processSignedBeaconBlock*(
     when typeof(signedBlock).toFork() >= ConsensusFork.Deneb:
       if self.blobQuarantine[].hasBlobs(signedBlock):
         blobs = self.blobQuarantine[].popBlobs(signedBlock.root)
+
+        self.blockProcessor[].addBlock(
```
tersec: This appears to only call `self.blockProcessor[].addBlock(...)` at all `when typeof(signedBlock).toFork() >= ConsensusFork.Deneb`?

henridf: indeed, some crossed wires there 🤪 ... should be fixed.

tersec (approved):

PR #4823, opened 2023-04-15
Remove unnecessary field derefs in BlockProcessor.storeBlock

henridf:


henridf:
Seems like spurious test failures.

tersec:
> Seems like spurious test failures.

Yeah, checked both of them. One was related to some known (now fixed) issues with that Jenkins host and the other was another example of https://github.com/status-im/nimbus-eth2/issues/4619#issuecomment-1509984138 which is not related to this PR and which has been seen for months now.

PR #4830, opened 2023-04-17
low attestations during epoch should instafail in CI; dbg -> warn level in `newPayload` log

tersec:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -293,7 +293,8 @@ proc newExecutionPayload*(
     Future[Opt[PayloadExecutionStatus]] {.async.} =
 
   if not elManager.hasProperlyConfiguredConnection:
-    debug "No EL connection for newPayload"
+    warn "No EL connection for newPayload",
```
arnetheduck: this needs a more descriptive message since users are likely to see it - ie something like `No execution client connected, cannot create payload for block proposal`

tersec: https://github.com/status-im/nimbus-eth2/commit/b1161b7546ec83434a613e6728ce7f7448e379bf

It's not primarily used as part of the block production workflow, though it's part of that too -- most of the time, it'd be seen when trying to process incoming blocks from gossip (e.g., the logs in https://github.com/status-im/nimbus-eth2/pull/4831 show this scenario).

So there's a limit to how interpretively descriptive this message can be. But sure, it can be moreso than now.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -293,7 +293,8 @@ proc newExecutionPayload*(
     Future[Opt[PayloadExecutionStatus]] {.async.} =
 
   if not elManager.hasProperlyConfiguredConnection:
-    debug "No EL connection for newPayload"
+    warn "No execution client connected; cannot process block payloads",
+      executionPayload = shortLog(executionPayload)
```
etan-status: On `warn`, could this become spammy during sync with `--no-el`?

For user support, logging a sanitized EL config as well could be useful, as the startup line is long scrolled away when these issues happen unexpectedly (but possibly a separate issue)

tersec: Within-same-screenfull-of-lines is a metric I don't regard as particularly important.

But yeah, it could become spammy, and if the user specifies that they know there's no EL, it should not be a warning.

beacon_chain/validators/validator_duties.nim
```diff
@@ -236,7 +236,7 @@ proc createAndSendAttestation(node: BeaconNode,
           fork, genesis_validators_root, data)
         if res.isErr():
           warn "Unable to sign attestation", validator = shortLog(validator),
-                attestatingData = shortLog(data), error_msg = res.error()
```
etan-status, line 239: or `attestationData` to match the type name. fine with either.

etan-status (approved):
(pending the new log verbosity not leading to spam during sync)

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -332,6 +332,7 @@ proc weigh_justification_and_finalization(
       previous_epoch_target_balance,
       current_epoch_target_balance,
       epoch = get_current_epoch(state)
+    doAssert false
```
zah: Why don't we use `fatal` and `quit(1)` here? The `assert` mostly contributes noise in the error message.

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4830/commits/57c9deaa47a19a4244ef4ab2c82f23db31e74f32

beacon_chain/spec/state_transition_epoch.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4830/commits/57c9deaa47a19a4244ef4ab2c82f23db31e74f32

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4830/commits/57c9deaa47a19a4244ef4ab2c82f23db31e74f32

PR #4833, opened 2023-04-18
Request manager: Handle blobless blocks

henridf:
Post-Deneb, when the request manager receives a missing block from a peer, it needs to check if the corresponding blobs are available, and if so pass them along. If they aren't available, the newly-fetched block must be put in blobless quarantine (while the blobs are retrieved, coming in next commit).


henridf:
Not sure why `jenkins/nimbus-eth2/linux/x86_64 ` is marked as "pending", but appears to have completed successfully when looking at the jenkins ui.

PR #4844, opened 2023-04-20
RequestManager: add support for fetching Blobs

henridf:


tersec:
`jenkins/nimbus-eth2/macos/aarch64` is `macm1-01`/Jenkins issue:
```
Running on macm1-01 in /Users/jenkins/workspace/_platforms_macos_aarch64_PR-4844
...
[2023-04-21T10:51:24.009Z] Submodule path 'vendor/nimbus-security-resources': checked out 'ce9945b1b159d4c9b628f8c4cd2d262964692810'
[2023-04-21T10:51:24.009Z] Submodule path 'vendor/nimcrypto': checked out '24e006df85927f64916e60511620583b11403178'
[2023-04-21T10:51:24.009Z] Submodule path 'vendor/sepolia': checked out '23dbdd7077fa0abdd7c8f29862a2f8883151d144'
[2023-04-21T10:51:24.941Z] 
[2023-04-21T10:51:31.882Z] [92mBuilding:[39m Nim compiler
[2023-04-21T10:51:58.658Z] Cannot contact macm1-01: java.lang.InterruptedException
[Pipeline] sh
[2023-04-21T10:57:37.837Z] + make deps
[2023-04-21T10:57:37.837Z] Building: install/usr/lib/libunwind.a
[2023-04-21T10:57:42.249Z] Building: install/usr/lib/libbacktrace.a
[2023-04-21T10:57:44.427Z] sh: /Users/jenkins/workspace/_platforms_macos_aarch64_PR-4844@tmp/durable-297b3ad4/jenkins-result.txt.tmp: No such file or directory
[2023-04-21T10:57:44.431Z] mv: rename /Users/jenkins/workspace/_platforms_macos_aarch64_PR-4844@tmp/durable-297b3ad4/jenkins-result.txt.tmp to /Users/jenkins/workspace/_platforms_macos_aarch64_PR-4844@tmp/durable-297b3ad4/jenkins-result.txt: No such file or directory
```
https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/aarch64/job/PR-4844/2/consoleText

beacon_chain/sync/request_manager.nim
```diff
@@ -72,6 +85,21 @@ proc checkResponse(roots: openArray[Eth2Digest],
...
+  for blob in blobs:
+    var found = false
+    for id in idList:
+      if id.block_root == blob.block_root and
+         id.index == blob.index:
+          found = true
+          break
+    if not found:
+        return false
+  return true
```
```suggestion
  true
```
would be slightly nicer, but it's okay this way too

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -68,3 +76,12 @@ func hasBlobs*(quarantine: BlobQuarantine, blck: deneb.SignedBeaconBlock):
     if idxs[i] != uint64(i):
       return false
   true
+
+func blobFetchRecord*(quarantine: BlobQuarantine, blck: deneb.SignedBeaconBlock):
+     BlobFetchRecord =
+  var indices: seq[BlobIndex]
+  for i in 0..len(blck.message.body.blob_kzg_commitments)-1:
```
tersec: In general, it's a little more idiomatic within the Status codebases to write these
`for i in 0 .. len(foo) - 1` loops as `for i in 0 ..< len(foo)`.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -68,3 +76,12 @@ func hasBlobs*(quarantine: BlobQuarantine, blck: deneb.SignedBeaconBlock):
...
+    let idx = BlobIndex(i)
+    if not quarantine.blobs.hasKey((blck.root, idx)):
+      indices.add(idx)
+  return BlobFetchRecord(block_root: blck.root, indices: indices)
```
tersec: prefer implicit/expression return

beacon_chain/sync/request_manager.nim
henridf: indeed! I dunno why in Nim, the expression-based return doesn't come to me.

beacon_chain/consensus_object_pools/blob_quarantine.nim
henridf: adjusted this and other occurences in this file.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1345,6 +1348,36 @@ proc onSlotStart(node: BeaconNode, wallTime: BeaconTime,
...
 
+proc handleMissingBlobs(node: BeaconNode) =
+  let
+    wallTime = node.beaconClock.now()
+    wallSlot = wallTime.slotOrZero()
+    delay = wallTime - wallSlot.start_beacon_time()
+    waitDur = TimeDiff(nanoseconds: BLOB_GOSSIP_WAIT_TIME_NS)
+
+  var fetches: seq[BlobFetchRecord]
+  for blobless in node.quarantine[].peekSortedBlobless():
```
tersec: How important is it that these be sorted? Currently, every second, this not-exactly-bounded (but to be fair, likely not that large) list is sorted.

Is it so that `node.blockProcessor[].addBlock(...)` delivers these blocks in-order to avoid quarantine, and that it's basically a tradeoff of sort on the front-end or handle the orphan-resolution on the back-end, or are there other reasons?

Apropos `fetchMissingBlobs()`, nothing going through asynchronous network paths can safely be assumed to stay in the same order as called.

beacon_chain/consensus_object_pools/block_quarantine.nim
```diff
@@ -314,3 +314,11 @@ func popBlobless*(quarantine: var Quarantine, root: Eth2Digest):
     return some(blck)
   else:
     return none(deneb.SignedBeaconBlock)
+
+iterator peekSortedBlobless*(quarantine: var Quarantine): deneb.SignedBeaconBlock =
+  var blobless: seq[deneb.SignedBeaconBlock]
+  for k, v in quarantine.blobless.mpairs():
+    blobless.add(v)
+  let sorted = blobless.sortedByIt(it.message.slot)
```
tersec: `peekSortedBloblesss` already takes a `var quarantine`, so it's not presenting itself as some pure and immutable function of `Quarantine`, so can use an [in-place sort](https://nim-lang.org/docs/algorithm.html#sort%2CopenArray%5BT%5D%2Cproc%28T%2CT%29) and avoid some copying and allocation.

`sortedByIt` doesn't even avoid the useless potential closure created by the `cmp` function: https://github.com/nim-lang/Nim/blob/1aa9273640c0c51486cf3a7b67282fe58f360e91/lib/pure/algorithm.nim#L467-L501 but rather just hides it behind an abstraction.

beacon_chain/nimbus_beacon_node.nim
henridf: Hmm, good question. Yes, my intent as I did this was to deliver in-order blocks to `addBlock(..)`. No other reasons. 

That said, this is all somewhat comparable to the call to `block_quarantine.checkMissing()` for missing blocks, which aren't sorted, and that presumably works out fine, so maybe I'm missing something here? 🤔

beacon_chain/consensus_object_pools/block_quarantine.nim
henridf: Ah! I somehow missed in-place sort when looking around for a stdlib sort function. Yes, that's a better choice. Before I change this around, let's first see how we resolve the above question of whether sortedness is even necessary.

henridf:
A side note: this is PR #4844. 😎

beacon_chain/nimbus_beacon_node.nim
tersec: `block_quarantine.checkMissing()` is input to asynchronous network functionality via the request manager which it can't really predict the output ordering of well regardless, so it wouldn't gain as much to sort them.

I think sorting is probably reasonable, but (see elsewhere) can be better done in-order. The `block_processor` absolutely wants blocks in order.

It's a little odd because the way this is designed, it does mix the two output data paths, one of which cares about order, and the other of which really doesn't much:
```nim
    if not node.blobQuarantine[].hasBlobs(blobless):
      let missing = node.blobQuarantine[].blobFetchRecord(blobless)
      doAssert not len(missing.indices) == 0
      fetches.add(missing)
    else:
      # this is a programming error should it occur.
      warn "missing blob handler found blobless block with all blobs"
      node.blockProcessor[].addBlock(
        MsgSource.gossip,
        ForkedSignedBeaconBlock.init(blobless),
        node.blobQuarantine[].popBlobs(
          blobless.root)
      )
      node.quarantine[].removeBlobless(blobless)
```
the `not node.blobQuarantine[].hasBlobs(blobless)` branch doesn't much benefit from being sorted, because it'll get shuffled by network latency and request latency from different peers regardless, while the `else` branch definitely would want to try to use a sorted order if convenient.

But they're mixed in the same iteration through the blocks. However, because the cares-about-sorted path shouldn't happen, probably prefer not to sort.

henridf: Oh! you put your finger on it. So the second branch (commented "this is a programming error should it occur") really shouldn't occur, under the invariant that if a block is in "blobless" quarantine, then we don't have all of its blobs in blob quarantine. Maybe I should `doAssert` here rather than just `warn`/`addBlock`.

And as you rightly point out, sortedness isn't preserved by async network queries. 

With all this in mind, I removed sorting in b05032c

beacon_chain/consensus_object_pools/block_quarantine.nim
henridf: (Resolved: sortedness not necessary)

beacon_chain/nimbus_beacon_node.nim
tersec: To the `doAssert`, that'd be an improvement if and only if it can be guaranteed. 

https://github.com/status-im/nimbus-eth2/blob/ea8ef5799e6eb494be4345624412a426077ecc9d/beacon_chain/consensus_object_pools/blob_quarantine.nim#L24-L28 doesn't itself guarantee this, and
https://github.com/status-im/nimbus-eth2/blob/ea8ef5799e6eb494be4345624412a426077ecc9d/beacon_chain/gossip_processing/eth2_processor.nim#L303-L322
doesn't check `hasBlobs` until after it calls `put` on the blob quarantine.

So in general, the extent to which this is guaranteed depends on blocks with blobs being removed as soon as all the blobs show up. Haven't yet audited all this that far, but would be a criteria, I guess.

Also, I don't see how
https://github.com/status-im/nimbus-eth2/blob/ea8ef5799e6eb494be4345624412a426077ecc9d/beacon_chain/consensus_object_pools/blob_quarantine.nim#L62-L69
doesn't try to request an out-of-bounds `seq` element.

But anyway, at this point this PR is clearly an improvement over status quo, so, merging.

henridf: Thanks for calling out my nonsense re the invariant. 🤣 Getting mixed up with earlier revs etc.

PR #4846, opened 2023-04-21
VC: Remote BN clock offset monitoring.

cheatfate:


tersec:
`jenkins/nimbus-eth2/macos/aarch64` failure is out of disk on `macm1-01-ms`

beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -427,3 +427,27 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+            return RestApiResponse.jsonError(Http400,
+                                             InvalidTimestampValue,
+                                             $dres.error())
+          dres.get().timestamp1
+    let
+      delay = node.processingDelay.valueOr: ZeroDuration
+      response = RestNimbusTimestamp2(
+        timestamp1: timestamp1,
+        timestamp2: timestamp2,
+        timestamp3: getTimestamp(),
```
arnetheduck, line 451: the difference between 2 and 3 is the time to parse the json - is it really .. meaningful to include

cheatfate: I wish to keep it to be compatible with original protocol - NTP.

beacon_chain/nimbus_validator_client.nim
```diff
@@ -198,7 +200,8 @@ proc runVCSlotLoop(vc: ValidatorClientRef) {.async.} =
         blockIn = vc.getDurationToNextBlock(wallSlot),
         validators = vc.attachedValidators[].count(),
         good_nodes = goodNodes, viable_nodes = viableNodes,
-        bad_nodes = badNodes, delay = shortLog(delay)
+        bad_nodes = badNodes, average_time_offset = vc.getAverageOffset(),
```
arnetheduck: the time offset is something that we would put in a metric rather than in a recurring log: it's highly unusual for two computers not to have the same time, it indicates an installation error where the user forgot to enable ntp on both (ntp or equivalent is a hard requirement for using nimbus, per manual).

the log is only interesting if the deviations exceed a threshold at which point things stop working - these thresholds should likely be `MAXIMUM_GOSSIP_CLOCK_DISPARITY div 2` to relate them to the actual threshold in the protocol.

arnetheduck: in other words, we don't want to take up valuable log space for a piece of information that in a correctly configured setup is noise

beacon_chain/validator_client/fallback_service.nim
```diff
@@ -9,6 +9,9 @@ import common
 
 const
   ServiceName = "fallback_service"
+  WARNING_TIME_OFFSET = TimeOffset.init(2.seconds)
```
arnetheduck: per comments elsewhere, the vc stops working if offset gets near `MAXIMUM_GOSSIP_CLOCK_DISPARITY`, so half or 2/3 of that is a reasonable warning level already

beacon_chain/validator_client/fallback_service.nim
```diff
@@ -256,15 +273,94 @@ proc checkNodes*(service: FallbackServiceRef): Future[bool] {.async.} =
...
+        raise exc
+      except CatchableError as exc:
+        warn "An unexpected error occurred while asking for time offset",
+             reason = $exc.msg, error = $exc.name
+        Opt.none(int64)
+
+    if tres.isSome():
+      let timeOffset = TimeOffset.init(tres.get())
+      node.timeOffset = Opt.some(timeOffset)
+      debug "Remote beacon time offset updated", time_offset = timeOffset
```
arnetheduck: metric instead here

beacon_chain/validator_client/fallback_service.nim
```diff
@@ -256,15 +273,94 @@ proc checkNodes*(service: FallbackServiceRef): Future[bool] {.async.} =
...
+        let
+          delay = vc.processingDelay.valueOr: ZeroDuration
+          res = await node.client.getTimeOffset(delay)
+        Opt.some(res)
+      except RestError as exc:
+        debug "Unable to obtain remote beacon node time offset",
+              reason = $exc.msg
+        Opt.none(int64)
+      except RestResponseError as exc:
+        debug "Remote beacon node responds with unexpected status code",
```
arnetheduck: if we receive 404 or 503 here, seems like we should exit the loop really - consider that the nimbus vc regularly is used with other BN:s that don't support the nimbus-only interface

beacon_chain/validator_client/fallback_service.nim
```diff
@@ -256,15 +273,94 @@ proc checkNodes*(service: FallbackServiceRef): Future[bool] {.async.} =
     raise exc
   return res
 
+proc runTimeMonitor(service: FallbackServiceRef,
+                    node: BeaconNodeServerRef) {.async.} =
+  let
+    vc = service.client
+    roles = AllBeaconNodeRoles
+    statuses = AllBeaconNodeStatuses - {RestBeaconNodeStatus.Offline}
```
arnetheduck, line 347: hm, consider the node we only use for publishing blocks - we probably don't want to divulge our VC to it - maybe the time checking should be a "role" as well - that also solves how to disable it entirely (users may not want it to be active: on a correctly configured setup, it just wastes cycles for no utility)

cheatfate: Believe me, it is not waste of cycles, requests are done once per slot and from my experience of using VC this information is very important, because sometimes configuration of NTPD is not super trivial and time difference between two hosts could introduce an issue which is very hard to investigate.

arnetheduck: regardless, if I'm using a "remote" bn to publish stuff, I don't want it to be pinged (I might not want to reveal that I'm using a nimbus VC for example)

arnetheduck: in fact, we could disable the feature entirely on localhost connections (there really provides no value)

beacon_chain/validator_client/fallback_service.nim
cheatfate: How i can create dynamic metrics?
When single BN is used your will see proper metric's chart, but as soon as you introduce more than on BN you will see dinosaur spine chart which is pretty hard to read.

beacon_chain/validator_client/fallback_service.nim
cheatfate: And its actually why i want to see it in `Slot start` messages too, because when this value is in metrics and people will start asking for help - why their VC has poor performance they are not providing metrics to you, but they can copy-paste some logs for you.

arnetheduck: > why their VC has poor performance they are not providing metrics to you, but they can copy-paste some logs for you.

this is already covered by the other higher-priority logging that is conditioned on the difference in time (ie notices / warnings) so if it's happening, users will copy-paste _those_ logs instead of this cryptic value.

it's important to understand that these kinds of errors are exotic edge conditions: they don't happen on localhost, they don't happen on a correctly setup machine.

they will however happen if the BN is stuck on processing (state replays) so we need to be a bit careful about overdoing the reaction to a failure.

beacon_chain/validator_client/fallback_service.nim
arnetheduck: via labels, like so: https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/validators/validator_monitor.nim#L42

the label should ideally be some "short" name of the bn (it gets sent to the metrics server repeatedly so long names add to overhead)

beacon_chain/validator_client/fallback_service.nim
cheatfate: Right now it will ignore all the delays which could happen on BN side or VC side, the only problem which it monitors is OS clock difference, nothing more.

And its not a rare condition, i can of course remove this monitoring when VC and BN are on the same host and connected via localhost, but all other setups are affected...

tersec: I'd like to just add my support for slightly less useless `INFO` logs.

cheatfate: > I'd like to just add my support for slightly less useless `INFO` logs.

Could you be more specific?

I just want to add `time_offset = <time_offset>` property with BN (if its present).

tersec: Yes, I agree with this change.

cheatfate:
I want to describe the problem in example, here is the case BN is running in local network on Linux device, while VC is running on Windows. I will show you 3 attestations timings. Please ignore Windows log timestamps because this timestamps in UTC+2 timezone with daylight saving, so right now it actually has UTC+3. Both PCs are synced with single NTP server which is also placed in local network. I dont know the reason but Windows system always has 300ms offset and all my attempts to fix it failed.

#### Windows running VC
```
INF 2023-04-26 04:28:00.001+03:00 Slot start                                 slot=5497040 epoch=171782 attestationIn=4s blockIn=<unknown> validators=2 node_status=synced time_offset=309ms623us692ns delay=1ms878us
NOT 2023-04-26 04:28:00.093+03:00 Time offset data                           timestamp1=1682472480001878000 timestamp2=1682472480356551065 timestamp3=1682472480356557355 timestamp4=1682472480093385000 delay14=7221600 delay23=849191 rtt=91ms507us
NOT 2023-04-26 04:28:04.172+03:00 Attestation published                      delay=172ms307us300ns service=attestation_service validator=819354e6@212553 attestation="(aggregation_bits: 0b00000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 5497040, index: 36, beacon_block_root: \"4bea56a1\", source: \"171781:c85e170e\", target: \"171782:a7844ad8\"), signature: \"858cb8ca\")"
```

#### Linux running BN
```
INF 2023-04-26 01:28:00.000+00:00 Slot start                                 topics="beacnde" slot=5497040 epoch=171782 sync=synced peers=159 head=076777ae:5497039 finalized=171780:b7250fb0 delay=849us191ns
INF 2023-04-26 01:28:00.059+00:00 Slot end                                   topics="beacnde" slot=5497040 nextActionWait=2m47s940ms766us608ns nextAttestationSlot=5497054 nextProposalSlot=-1 syncCommitteeDuties=none head=076777ae:5497039
INF 2023-04-26 01:28:04.435+00:00 Attestation sent                           attestation="(aggregation_bits: 0b00000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 5497040, index: 36, beacon_block_root: \"4bea56a1\", source: \"171781:c85e170e\", target: \"171782:a7844ad8\"), signature: \"858cb8ca\")" delay=435ms9us714ns subnet_id=36
```
First i want to explain some data in Windows logs:

1. `Start slot` log statement has field `time_offset=309ms623us692ns` which indicates that there present offset in clocks between VC and BN. Also in this statement you can see that processing delay is insignificant and equal to `delay=1ms878us`.
2. `Time offset` log statement is related to time monitoring request (it will not going to be seen in release). The most important value here is `rtt = 91ms507us` which means round trip time of single request.
3. `Attestation published` statement shows us that it published attestation with delay `delay=172ms307us300ns`

Now we should check BN side :

1. We should note that processing delay on BN side is also insignificant and shown in `Slot start` statement as `delay=849us191ns`.
2. But `Attestation sent` log statement has pretty big delay of `delay=435ms9us714ns`. 

So how we get such big delay on BN side while VC thinks that delay should be not bigger than `172ms307us300ns`. This time also includes RTT because it get calculated after attestation has been published, so we could expect that BN will receive initial attestation publish request in `172ms - RTT / 2 ~ 127ms`. Now we can add `time_offset` value and we will obtain BN delay value - `127ms + 309ms = 436ms ~ 435ms9us714ns`.

So formula for BN delay would be 
`BN DELAY = (VC DELAY - RTT / 2) + TIME_OFFSET`

To prove my calculations i will add one more attestation

#### Windows VC
```
INF 2023-04-26 04:32:36.003+03:00 Slot start                                 slot=5497063 epoch=171783 attestationIn=4s blockIn=<unknown> validators=2 node_status=synced time_offset=312ms142us853ns delay=3ms954us900ns
NOT 2023-04-26 04:32:36.094+03:00 Time offset data                           timestamp1=1682472756002438400 timestamp2=1682472756359660092 timestamp3=1682472756359665070 timestamp4=1682472756094869600 delay14=12816100 delay23=965073 rtt=92ms431us200ns
NOT 2023-04-26 04:32:40.142+03:00 Attestation published                      delay=142ms425us300ns service=attestation_service validator=8107ff6a@212539 attestation="(aggregation_bits: 0b00000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 5497063, index: 22, beacon_block_root: \"4880500a\", source: \"171782:a7844ad8\", target: \"171783:3be258a7\"), signature: \"8cf849eb\")"
```

#### Linux BN
```
INF 2023-04-26 01:32:36.000+00:00 Slot start                                 topics="beacnde" slot=5497063 epoch=171783 sync=synced peers=159 head=74a99b47:5497062 finalized=171781:c85e170e delay=965us73ns
INF 2023-04-26 01:32:36.029+00:00 Slot end                                   topics="beacnde" slot=5497063 nextActionWait=1m11s970ms125us27ns nextAttestationSlot=5497069 nextProposalSlot=-1 syncCommitteeDuties=none head=74a99b47:5497062
INF 2023-04-26 01:32:40.410+00:00 Attestation sent                           attestation="(aggregation_bits: 0b00000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 5497063, index: 22, beacon_block_root: \"4880500a\", source: \"171782:a7844ad8\", target: \"171783:3be258a7\"), signature: \"8cf849eb\")" delay=410ms398us844ns subnet_id=22
```

*TIME_OFFSET* = 312ms142us853ns 
*RTT* = 92ms431us200ns
*VC ATTESTATION DELAY* = 142ms425us300ns 
*BN ATTESTATION DELAY* = 410ms398us844ns 
*CALCULATED* = 142 - 92 / 2 + 312 ~ 408ms (and you could not that we have VC processing delay `delay=3ms954us900ns`).

#### Windows VC
```
INF 2023-04-26 05:10:12.009+03:00 Slot start                                 slot=5497251 epoch=171789 attestationIn=4s blockIn=<unknown> validators=2 node_status=synced time_offset=325ms451us603ns delay=9ms471us700ns
NOT 2023-04-26 05:10:12.102+03:00 Time offset data                           timestamp1=1682475012012461400 timestamp2=1682475012400128122 timestamp3=1682475012400165377 timestamp4=1682475012102661400 delay14=9471700 delay23=719739 rtt=90ms200us
NOT 2023-04-26 05:10:16.158+03:00 Attestation published                      delay=158ms799us700ns service=attestation_service validator=8107ff6a@212539 attestation="(aggregation_bits: 0b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 5497251, index: 5, beacon_block_root: \"e8bda37d\", source: \"171788:bdd9aeb7\", target: \"171789:0bdfbb89\"), signature: \"a18318bf\")"
```

#### Linux BN
```
INF 2023-04-26 02:10:12.000+00:00 Slot start                                 topics="beacnde" slot=5497251 epoch=171789 sync=synced peers=159 head=76da0f77:5497250 finalized=171787:b4bf42ea delay=719us739ns
INF 2023-04-26 02:10:12.069+00:00 Slot end                                   topics="beacnde" slot=5497251 nextActionWait=7m35s930ms206us645ns nextAttestationSlot=5497289 nextProposalSlot=-1 syncCommitteeDuties=none head=76da0f77:5497250
INF 2023-04-26 02:10:16.452+00:00 Attestation sent                           attestation="(aggregation_bits: 0b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, data: (slot: 5497251, index: 5, beacon_block_root: \"e8bda37d\", source: \"171788:bdd9aeb7\", target: \"171789:0bdfbb89\"), signature: \"a18318bf\")" delay=452ms425us474ns subnet_id=5
```

*TIME_OFFSET* = 325ms451us603ns
*RTT* = 90ms200us
*VC ATTESTATION DELAY* = 158ms799us700ns 
*BN ATTESTATION DELAY* = 452ms425us474ns 
*CALCULATED* = 158 - 90 / 2 + 325 ~ 438ms (and you could not that we have VC processing delay `delay=9ms471us700ns`)

There of course present some error ~ 10-20ms, but it because we measure RTT for time-monitoring call and not for attestation publish request call, which is also fast but could provide another RTT value.

The core of the problem here is of course Windows time handling or some weird Windows behavior which could not properly correct their clocks according to NTP server and i'm not sure why we should start warn users when offset in time is bigger than 
`MAXIMUM_GOSSIP_CLOCK_DISPARITY` which is `500ms`. Even with such big clock difference VC could continue working.

arnetheduck:
>  I dont know the reason but Windows system always has 300ms offset and all my attempts to fix it failed.

nagle maybe?

>  MAXIMUM_GOSSIP_CLOCK_DISPARITY which is 500ms. Even with such big clock difference VC could continue working.

What happens here is that if a clock is off by more than `MAXIMUM_GOSSIP_CLOCK_DISPARITY`, in particular when it's early and performing the slot transition earlier than everyone else on the network, the network will simply reject it per the gossip propagation rules because the slot has not yet happened. 

There is nuance here: 

* this effect happens above all for blocks which are sent on the slot transition edge (ideally) - if we send a block more than 500ms before the slot started (according to your neighbour nodes above all because of gossip), it will simply be dropped - this is the main scenario that can break a user setup
* it applies to being early - if you're late, there is more tolerance (up to 4s)
* there is also more tolerance for attestations being early - when the VC is early on sending the attestation, the risk of voting for an old head increases. A late attestation might not be included if blocks are full.

So, we have several paths forward:

* be more lenient in warnings - decreases detection rate of problematic setup, but maybe it's better than having lots of false positives
  * if we consider a clock that's been set "by hand", it might be accurate within a second which is not accurate enough
* we can be more aggressive when the VC clock is _earlier_ than the BN - this is the most problematic case - ie we can use different limits for late and early
* we can get to the bottom of why there is clock difference - is it lag? granularity? something else?

beacon_chain/validator_client/fallback_service.nim
arnetheduck: > And its not a rare condition, 

well, it is a condition that doesn't exist on a correctly setup machine - it is also "useless" information when it's small (because then everything is alright and you don't really need to know).

The point here is that by adding one piece of information, all the other information becomes more obscure - for the general user, being presented with text that repeatedly doesn't contain useful information is a cost.

Notably, the exact same logging is done already, conditionally, when this is a problem, per `NOTICE_TIME_OFFSET`, with the same frequency as the slot start message - if this log doesn't appear, we know it's fine and don't have to waste the attention of the user.
 
> slightly less useless

there is no new relevant information in this log, that isn't covered by a higher-level (notice) log message already - how does this make the info log more useful?

beacon_chain/validator_client/fallback_service.nim
```diff
@@ -10,6 +10,25 @@ import common
...
+  WARN_TIME_OFFSETS = [
+    TimeOffset.init(-(MAXIMUM_GOSSIP_CLOCK_DISPARITY.nanoseconds div 2)),
+    TimeOffset.init(MAXIMUM_GOSSIP_CLOCK_DISPARITY.nanoseconds * 2),
+  ]
+  NOTE_TIME_OFFSETS = [
+    TimeOffset.init(-(MAXIMUM_GOSSIP_CLOCK_DISPARITY.nanoseconds div 4)),
+    TimeOffset.init(MAXIMUM_GOSSIP_CLOCK_DISPARITY.nanoseconds),
+  ]
+
+declareHistogram nvc_time_offset,
```
arnetheduck: I think a gauge would be enough here, histograms are expensive (they result in a line for every bucket)

arnetheduck (approved):

PR #4854, opened 2023-04-24
BlockProcessor.storeBlock: write blobs to DB

henridf:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -493,7 +494,9 @@ proc storeBlock*(
     # If the EL responded at all, we don't need to try again for a while
     self[].lastPayload = signedBlock.message.slot
 
-  # TODO: store blobs in db
+  # write blobs now that block has been written.
+  for b in blobs:
+    self.consensusManager.dag.db.putBlobSidecar(b[])
```
zah, line 499: Should the block and the blobs be written as a single transaction (atomically)?

henridf: Good question. My thinking is that this coupling isn't needed. If we write the block, we know that the block and its blobs have been validated(*). If we then crash between after writing the block, but before we've written blobs, we end up with a chain that is still valid - we're "just" (?) missing a blob, and won't be able to answer subsequent RPC requests for it.

I can definitely see an argument for coupling the writes though, and may also be missing some consideration. But this feels different than the other instances of multiple atomic writes where we would end up with an inconsistent db.

(*) blob validation not yet implemented

PR #4856, opened 2023-04-25
VC: Remote BN received block monitoring.

cheatfate:


beacon_chain/validator_client/common.nim
```diff
@@ -990,3 +1011,151 @@ proc checkedWaitForNextSlot*(vc: ValidatorClientRef, curSlot: Opt[Slot],
...
+  # have not yet seen the block), we'll impose a minimum delay of
+  # 2000ms. The delay is enforced only when we're not hitting the
+  # "normal" cutoff time for sending out attestations. An earlier delay
+  # of 250ms has proven to be not enough, increasing the risk of losing
+  # attestations, and with growing block sizes, 1000ms started to be
+  # risky as well. Regardless, because we "just" received the block,
+  # we'll impose the delay.
+
+  # Take into consideration chains with a different slot time
+  const afterBlockDelay = nanos(attestationSlotOffset.nanoseconds div 2)
```
arnetheduck: it would be best if this was part of a function in `validator_duties` so that we don't duplicate this logic - it is likely to change

beacon_chain/conf.nim
```diff
@@ -937,6 +941,11 @@ type
       defaultValueDesc: $defaultBeaconNodeUri
       name: "beacon-node" .}: seq[Uri]
 
+    monitoringType* {.
+      desc: "Method used for block monitoring which are seen by beacon node"
+      defaultValue: BlockMonitoringType.Event
+      name: "monitoring-type".}: BlockMonitoringType
```
```suggestion
      name: "block-monitor-type".}: BlockMonitoringType
```

also need a `--block-monitor` bool to enabled/disable it - I would probably start disabled and ask the community to test

tersec: I suspect it might work better to pick one and ensure that works, rather than have to indefinitely support both when both, in principle, rely on standardized and supported Beacon API.

Already starting out by introducing an option seems premature.

arnetheduck: > I suspect it might work better to pick one and ensure that works, rather than have to indefinitely support both when both, in principle, rely on standardized and supported Beacon API.

what we can do is have the on/off option and keep the type hidden - if the events API turns out to be "stable enough" across all bn:s and middlewares, we can eventually remove the polling one

tersec: As long as it's truly hidden in a way that will allow for actual removal, not mere "deprecation", later.

This is why I typically add some `--debug-foo` option naming as well, for example.

beacon_chain/conf.nim
```diff
@@ -944,6 +949,11 @@ type
       defaultValueDesc: $defaultBeaconNodeUri
       name: "beacon-node" .}: seq[Uri]
 
+    monitoringType* {.
+      desc: "Enable block monitoring which are seen by beacon node"
```
```suggestion
      desc: "Enable block monitoring which are seen by beacon node (BETA)"
```

arnetheduck (approved):

PR #4858, opened 2023-04-26
check signatures in CP sync backfill with LC

etan-status:
When using trusted node sync with light client (`--trusted-block-root`), the trust assumption on the server is reduced to solely be responsible for data availability, but not data correctness. This means that we must check block proposer signatures against the downloaded checkpoint, as they are not covered by the block root.

Note that this lowers the backfill speed when using LC based CP sync due to the extra checks, by about 60% for me.


tersec (approved):

etan-status:
@arnetheduck suggested to augment this with batch signature verification for a potential 2x speedup, and also update libp2p based backfill with batching accordingly.

PR #4860, opened 2023-04-26
add `el_offline` to `/eth/v1/node/syncing`

etan-status:
Add compatibility with https://github.com/ethereum/beacon-APIs/pull/290 to the beacon node. Behaviour when configured with multiple ELs is not specified; intention suggests to indicate whether all ELs are offline.

### DIFF
--- a/beacon_chain/eth1/eth1_monitor.nim
+++ b/beacon_chain/eth1/eth1_monitor.nim
@@ -1965,6 +1965,9 @@ func isNewLastBlock(m: ELManager, blk: Eth1BlockHeader|BlockObject): bool =
 func hasConnection*(m: ELManager): bool =
   m.elConnections.len > 0
 
+func hasAnyWorkingConnection*(m: ELManager): bool =
+  m.elConnections.anyIt(it.state == Working)
+
 func hasProperlyConfiguredConnection*(m: ELManager): bool =
   for connection in m.elConnections:
     if connection.etcStatus == EtcStatus.match:
--- a/beacon_chain/rpc/rest_node_api.nim
+++ b/beacon_chain/rpc/rest_node_api.nim
@@ -268,12 +268,10 @@ proc installNodeApiHandlers*(router: var RestRouter, node: BeaconNode) =
         else:
           none[bool]()
       elOffline =
-        if node.currentSlot().epoch() < node.dag.cfg.CAPELLA_FORK_EPOCH:
-          none[bool]()  # Added with ethereum/beacon-APIs v2.4.0
-        elif isNil(node.elManager):
-          some(true)
-        else:
+        if node.currentSlot().epoch() >= node.dag.cfg.CAPELLA_FORK_EPOCH:
           some(not node.elManager.hasAnyWorkingConnection)
+        else:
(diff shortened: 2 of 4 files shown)


beacon_chain/rpc/rest_node_api.nim
```diff
@@ -267,10 +267,18 @@ proc installNodeApiHandlers*(router: var RestRouter, node: BeaconNode) =
           some(node.dag.is_optimistic(node.dag.head.root))
         else:
           none[bool]()
+      elOffline =
+        if node.currentSlot().epoch() < node.dag.cfg.CAPELLA_FORK_EPOCH:
+          none[bool]()  # Added with ethereum/beacon-APIs v2.4.0
+        elif isNil(node.elManager):
```
zah: The `elManager` reference cannot be `nil` anymore.

beacon_chain/rpc/rest_node_api.nim
```diff
@@ -267,10 +267,16 @@ proc installNodeApiHandlers*(router: var RestRouter, node: BeaconNode) =
           some(node.dag.is_optimistic(node.dag.head.root))
         else:
           none[bool]()
+      elOffline =
+        if node.currentSlot().epoch() >= node.dag.cfg.CAPELLA_FORK_EPOCH:
+          some(not node.elManager.hasAnyWorkingConnection)
```
etan-status, line 272: I _think_ the concern here is that an old client may be strict about inclusion of this extra key.

Mostly theoretical, only gnosis / research, as capella shipped already

beacon_chain/eth1/eth1_monitor.nim
```diff
@@ -1965,6 +1965,9 @@ func isNewLastBlock(m: ELManager, blk: Eth1BlockHeader|BlockObject): bool =
 func hasConnection*(m: ELManager): bool =
   m.elConnections.len > 0
 
+func hasAnyWorkingConnection*(m: ELManager): bool =
+  m.elConnections.anyIt(it.state == Working)
```
etan-status, line 1969: this treats `--no-el` same as a config with ELs that are all unreachable. essentially the same: stuck with the local block hash validation fallback and optimistic sync

zah: Sounds reasonable to me

beacon_chain/rpc/rest_node_api.nim
zah: The API spec requires from the client to ignore unfamiliar fields and this field has some value even before the capella fork epoch, so I think it would be better to introduce the small risk of incompatibility here.

etan-status: Well, out own VC has a history of not ignoring them. But I guess, those versions are incompatible with Capella based BN anyway.

Anyway, I'm fine with either, but for `execution_optimistic` there is a similar guard for bellatrix fork epoch, which is enforced across 3 different modules. I'd personally rather err on the consistent side of having the guard (and then remove them altogether if we decide that's the way to go), over having it sometimes but not other times.

If you strongly feel that the guard should be removed, I can remove it; I don't think that there are practical consequences from having it or not having it.

(followup PR, depending on your preference)

PR #4865, opened 2023-04-27
track `el_offline` status in VC

etan-status:
Updates the validator client to track the new `el_offline` field in the `/eth/v1/node/syncing` response.
https://github.com/ethereum/beacon-APIs/pull/290

If `el_offline` is present, the BN will be treated same as if it was optimistically synced. Note that a BN could report that it is in sync, but may still have intermittent connection issues to the EL, making it impossible to `getPayload` from them since Capella.

### DIFF
--- a/beacon_chain/nimbus_validator_client.nim
+++ b/beacon_chain/nimbus_validator_client.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2018-2021 Status Research & Development GmbH
+# Copyright (c) 2018-2023 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
@@ -184,8 +184,11 @@ proc runVCSlotLoop(vc: ValidatorClientRef) {.async.} =
         # Good nodes are nodes which can be used for ALL the requests.
         goodNodes = counts.data[int(RestBeaconNodeStatus.Synced)]
         # Viable nodes are nodes which can be used only SOME of the requests.
-        viableNodes = counts.data[int(RestBeaconNodeStatus.OptSynced)] +
+        viableNodes = counts.data[int(RestBeaconNodeStatus.SyncedELOff)] +
+                      counts.data[int(RestBeaconNodeStatus.OptSynced)] +
+                      counts.data[int(RestBeaconNodeStatus.OptSyncedELOff)] +
                       counts.data[int(RestBeaconNodeStatus.NotSynced)] +
+                      counts.data[int(RestBeaconNodeStatus.NotSyncedELOff)] +
                       counts.data[int(RestBeaconNodeStatus.Compatible)]
         # Bad nodes are nodes which can't be used at all.
         badNodes = counts.data[int(RestBeaconNodeStatus.Offline)] +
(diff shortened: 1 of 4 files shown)


zah (approved):

cheatfate:
I'm not sure i understand states in this PR.
If EL is offline -> BN is optimistically synced, but if BN is optimistically synced EL could be offline/online
If BN is fully synced -> EL could not be offline...

Also this change is not enough and should not be applied in such way. There we have only REST server changes. Proper VC handling is missing.

etan-status:
EL can also go offline if BN is synced. For example, if some blocks are missed (as happened at the beginning of Capella), BN is still considered in sync, and head is still considered VALID (not optimistic), but EL connection may still be offline.

Same case for pre-Merge research networks, those do not have an optimistic head but still need an EL connection for deposits importing, which can go offline. This case is not considered relevant for Nimbus though.

Not sure what proper VC handling is missing, but as this is incomplete, will discard it.

cheatfate:
With VC handling i mean what should VC do with this status? Can it go further? What requests could be used and what requests not? What request's responses could be considered as success? As for me `opt-synced` and `synced` status is totally enough to indicate that BN is ok or not. With this optional status even more complexities (states) can be added and all this 
states needs some documentation, because it unclear and confusing.

So right now map of decisions could be considered as
* BN `synced` EL `synced` - VC marks this nodes as `Synced` and so it can be used for validator duties.
* BN `synced` EL `not-in-sync` - VC marks this nodes as `OptSynced` and it can be partially used for validator duties.
* BN `not-in-sync` EL `synced` - I think this is degenerate case (but maybe i'm wrong).
* BN `not-in-sync` EL `not-in-sync` - VC marks this nodes as `NotSynced` and so it can't be used for validator duties.

And now map of decisions is:
* BN `synced` EL `synced` - VC marks this nodes as `Synced` and so it can be used for validator duties.
* BN `synced` EL `not-in-sync` - VC marks this nodes as `OptSynced` and it can be partially used for validator duties.
* BN `synced` EL `offline` - ???
* BN `not-in-sync` EL `synced` - degenerate case??? unsynced
* BN `not-in-sync` EL `not-in-sync` - VC marks this nodes as `NotSynced` and so it can't be used for validator duties.
* BN `not-in-sync` EL `offline` - unsynced?

Or maybe this new status is overhead and it is actually not needed at all?

I'm curious, how BN could detect that EL is offline using networking? Because until we are not speaking with EL using OS pipes its very hard to prove that EL is offline and it is not link between BN and EL got broken.

etan-status:
Generally, I don't personally think that the new `el_offline` field is too useful, beyond maybe informing the user that they should check the connection between BN and EL.

BN `synced` EL `offline`: I think that this can happen in these cases:

- BN is offline as well, as it doesn't receive new data, it thinks it is synced, for about ~12 minutes.
- A couple blocks are missed in a row, and EL is actually the only component that's offline. Note: If a block were to be received, this would transition the BN to become optimistically synced, as it cannot validate the new block against the EL.

In first situation, BN can't really be used for duties. In second situation, BN can be used for everything except block production. However, this is so rare that it's likely not worth to try and treat this differently.

For BN `not-in-sync` EL `offline`: Here, the additional `offline` bit allows user to understand that there is a network or configuration problem that they may have to address. Otherwise, this will never go beyond `optimistically-synced`, until the EL is back online.

Overall, not sure what we should do with `el_offline`. We can ignore it, we can log it, we can treat it same as an optimistically synced BN, or we can use it as a tiebreaker (if no fully online and synced is available, consider a synced one with an offline EL).

PR #4864, opened 2023-04-26
Wire in blob validation

henridf:

PR #4870, opened 2023-04-27
Load trusted setup

henridf:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1877,6 +1881,9 @@ proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref HmacDrbgContext) {.rai
 
   let node = BeaconNode.init(rng, config, metadata)
 
+  if node.dag.cfg.DENEB_FORK_EPOCH != FAR_FUTURE_EPOCH:
+    discard Kzg.loadTrustedSetupFromString(trustedSetup)
```
```
nimbus-eth2$ rg loadTrustedSetupFromString
vendor/nim-kzg4844/kzg4844/csources/bindings/nim/kzg_ex.nim
63:proc loadTrustedSetupFromString*(_: type Kzg,
66:    kzg.loadTrustedSetupFromString(input)

vendor/nim-kzg4844/kzg4844/csources/bindings/nim/tests/test_yaml.nim
73:    let res = loadTrustedSetupFromString(trustedSetup)

vendor/nim-kzg4844/kzg4844/csources/bindings/nim/tests/test_kzg_ex.nim
25:    let res = Kzg.loadTrustedSetupFromString(trustedSetup)

vendor/nim-kzg4844/kzg4844/csources/bindings/nim/tests/test_kzg.nim
27:    let res = loadTrustedSetupFromString(trustedSetup)

vendor/nim-kzg4844/kzg4844/csources/bindings/nim/kzg.nim
86:proc loadTrustedSetupFromString*(input: string): Result[KzgCtx, string] =
```
if it's okay to `discard` here, why is it not `discarded` (or just not returned as non-`void` value) elsewhere?

Or if it is important elsewhere to check the return value, why is it safe here to discard it?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -450,8 +451,11 @@ proc initFullNode(
 
   node.updateValidatorMetrics()
 
-const SlashingDbName = "slashing_protection"
+const
+  SlashingDbName = "slashing_protection"
   # changing this requires physical file rename as well or history is lost.
+  trustedSetup* =
```
tersec: It doesn't look like there could be any non-`nimbus_beacon_node` users of this `trustedSetup` constant which yet exist, because it didn't exist previously to this PR, so not sure why this has to be exported.

beacon_chain/nimbus_beacon_node.nim
henridf: Not ok to discard! I pushed an in-progress rev with uncommitted changes in my tree. Have force-pushed a new rev that should address this and your other comment below.

beacon_chain/nimbus_beacon_node.nim
henridf: Yep, fixed.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1877,6 +1881,11 @@ proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref HmacDrbgContext) {.rai
 
   let node = BeaconNode.init(rng, config, metadata)
 
+  if node.dag.cfg.DENEB_FORK_EPOCH != FAR_FUTURE_EPOCH:
+    let res = Kzg.loadTrustedSetupFromString(trustedSetup)
+    if res.isErr():
+      raiseAssert res.error()
```
tersec, line 1915: Sure, alright, this at least keeps things consistent.

My previous question still stands to some extent, but is resolved enough for this PR.

tersec (approved):

beacon_chain/nimbus_beacon_node.nim
henridf: I must be missing something. Your previous question was "why is it ok to discard the return value", which is no longer done. What part of that still stands?

tersec: Yes, it's in the most precise sense resolved. The return value of `loadTrustedSetupFromString` isn't discarded.

So now it shifts towards a pair of questions:

1) API design question, which is kind of out of scope of this PR, of how/why this function critical to proper Deneb/Cancun functioning can fail in an interesting runtime way, and if it's possible to create a (-n honest) API for this which doesn't have that issue. Maybe it isn't.

2) How possible this is to show up as a real runtime error to end-users. If this truly "cannot" happen, that it can only happen with some build errors, or Nim compiler bugs, or filesystem corruption, et cetera, then asserting is fine. If there's some way it might legitimately fire for end-users, it'd be better to use `fatal` logs with `quit 1` afterwards, because asserts show up as indistinguishable from crashes to end-users.

The root question, sort of, then is what is the underlying failure-model of this function, when/how/why is it allowed to fail, or should it simply never fail except for logic error. That's what I'm still not sure of.

tersec: e.g. one strawman design for (1); will check source presently to see what actual options are: use compile-time loading so the actual `loadTrustedSetupFromString` runs unambiguously at compile-time, and by the time Nimbus ships, it's just one more resource within the binary, similar to how the mainnet/Prater/Sepolia network definitions are handled.

tersec: Because
```nim
  trustedSetup =
    staticRead"../vendor/nim-kzg4844/kzg4844/csources/src/trusted_setup_4.txt"
```
so the input is known at compile-time, so this should in principle be possible, but `loadTrustedSetupFromString` calls `loadTrustedSetup` which is via C bindings, so Nim might not be able to achieve this. It does look like an assert is fine though, since the input at least is baked into the binary image and shouldn't ever fail.

henridf:
Same build issue as here: https://github.com/status-im/nimbus-eth2/pull/4864#issuecomment-1525076901

(waiting on https://github.com/status-im/nim-kzg4844/tree/master/kzg4844, cc @jangko )

scripts/compile_nim_program.sh
```diff
@@ -31,7 +39,7 @@ fi
 # parallel.
 # We can't use '--nimcache:...' here, because the same path is being used by
 # LTO on macOS, in "config.nims"
-"$NIMC" c --compileOnly -o:build/${BINARY} "$@" -d:nimCachePathOverride=nimcache/release/${BINARY} "${SOURCE}"
+"$NIMC" c -f --compileOnly -o:build/${BINARY} "$@" -d:nimCachePathOverride=nimcache/release/${BINARY} "${SOURCE}"
```
zah, line 48: This is a pretty significant compromise. Please add a TODO comment explaining why this is needed.

henridf: Agreed, having this `-f` sucks. Comment added.

PR #4878, opened 2023-04-29
VC: Fix issue when optimistically synced node could block sync committee service functions.

cheatfate:
Issue was reported in discord chat and it affects scenario when multiple BN get used and some of them are optimistically synced.

### DIFF
--- a/beacon_chain/validator_client/sync_committee_service.nim
+++ b/beacon_chain/validator_client/sync_committee_service.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2022 Status Research & Development GmbH
+# Copyright (c) 2022-2023 Status Research & Development GmbH
 # Licensed and distributed under either of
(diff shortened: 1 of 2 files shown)


beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -360,8 +360,7 @@ proc publishSyncMessagesAndContributions(service: SyncCommitteeServiceRef,
           res.data.root
         else:
           if res.execution_optimistic.get():
-            notice "Execution client not in sync; skipping validator duties " &
-                   "for now", slot = slot
+            notice "Execution client not in sync", slot = slot
```
arnetheduck, line 363: per changes to `getHeadBlockRoot` we can no longer end up in this branch of the logic?

cheatfate: yup

PR #4880, opened 2023-04-30
harden gossip `strictVerification` in local testnets

etan-status:
Fail local testnets on any gossip REJECT, instead of just asserting some of the attestation related checks. This now also ensures that blocks, BLS to Execution changes, blob sidecars and LC messages are checked when running in a local testnet environment (`--verify-finalization`).

https://github.com/status-im/nimbus-eth2/pull/2904#discussion_r719603935

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -132,7 +132,6 @@ func check_beacon_and_target_block(
 
 func check_aggregation_count(
     attestation: Attestation, singular: bool): Result[void, ValidationError] =
-
   let ones = attestation.aggregation_bits.countOnes()
   if singular and ones != 1:
     return errReject("Attestation must have a single attestation bit set")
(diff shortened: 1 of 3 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -156,33 +155,86 @@ func check_attestation_subnet(
...
+# ValidatorChangePool
+template checkedReject(
+    pool: ValidatorChangePool, msg: cstring): untyped =
+  pool.dag.checkedReject(msg)
+
+template checkedReject(
+    pool: ValidatorChangePool, error: ValidationError): untyped =
+  pool.dag.checkedReject(error)
+
+template checkedResult(
```
arnetheduck, line 222: meh, helpers like this are a bit weird in that they appear as a false dependency in the calling code (vs pool.dag) - but whatever

arnetheduck (approved):

beacon_chain/gossip_processing/gossip_validation.nim
etan-status: Yeah, I see the problem. The alternative would be a `when compiles` jungle, or using different template names for each case.

etan-status: Or, clobbering the calling site with extracting the `strictVerification` bit. looks quite distracting though, tried it.

PR #4882, opened 2023-05-01
only trigger liveness failsafe for mainnet

tersec:
Sepolia has a closed validator set and Prater/Goerli have too many missed blocks, leading to the liveness failsafe typically being in effect, and preventing testing MEV before mainnet deployment.

### DIFF
--- a/beacon_chain/validators/validator_duties.nim
+++ b/beacon_chain/validators/validator_duties.nim
@@ -619,7 +619,7 @@ proc getBlindedBlockParts[EPH: ForkyExecutionPayloadHeader](
           "getBlindedExecutionPayload error: " & exc.msg)
 
   if executionPayloadHeader.isErr:
-    warn "getBlindedBlockParts: getBlindedExecutionPayload failed",
+    warn "Could not obtain blinded execution payload header",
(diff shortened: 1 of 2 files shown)


beacon_chain/validators/validator_duties.nim
```diff
@@ -619,7 +619,7 @@ proc getBlindedBlockParts[EPH: ForkyExecutionPayloadHeader](
           "getBlindedExecutionPayload error: " & exc.msg)
 
   if executionPayloadHeader.isErr:
-    debug "getBlindedBlockParts: getBlindedExecutionPayload failed",
+    warn "getBlindedBlockParts: getBlindedExecutionPayload failed",
```
arnetheduck: a more user-friendly log here?

arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4882/commits/a10334fb74a80f818f7d8f1f80913f34dd0a79cd

PR #4886, opened 2023-05-03
Github Runner: test use of self hosted runner for build

apentori:
**Work in progress**

Branch to test the deployment of self hosted linux github runner

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
             nimflags-extra: --mm:refc
           - target:
               os: linux
-            builder: ubuntu-22.04
+            builder: ['self-hosted','ubuntu-22.04']
             shell: bash
           - target:
               os: macos


jakubgs (approved):

apentori:
The runner are hosted on Ubuntu 22.04 server

etan-status (approved):

.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         include:
           - target:
               os: linux
-            builder: ubuntu-20.04
+            builder: [self-hosted]
```
arnetheduck: let's maybe keep both around for a bit until things are stable

arnetheduck:
> The runner are hosted on Ubuntu 22.04 server

per discussion, we should use the oldest supported ubuntu LTS release here, meaning 20.04 in this case

apentori:
The runner are being reinstalled on ubuntu 20.04.

First test detected a problem in step `Build Nim and Nimbus dependencies` at getting git submodudule for `vendor/nimbus-build-system/vendor/Nim`.

Currently investigating

.github/workflows/ci.yml
apentori: Tag `ubuntu-20.04` added in the list of builder and as a label for the runner

apentori:
Runner updated to ubuntu 22.04

jakubgs (approved):

jakubgs:
Seems like we have some builds already running:

image

I thought that maybe I could find some indication what host it's running on, but I can't find anything for this nightly:
https://github.com/status-im/nimbus-eth2/actions/runs/6491357946/job/17628544161

But the runners are definitely there:

image

PR #4887, opened 2023-05-03
Incremental pruning

arnetheduck:
When turning on pruning the first time the current pruning algorithm will prune the full database at startup. This delays restart unnecessarily, since all of the pruned space is not needed at once.

This PR introduces incremental pruning such that we will never prune more than 32 blocks or the sync speed, whichever is higher.

This mode is expected to become default in a follow-up release.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1976,8 +1976,10 @@ proc pruneHistory*(dag: ChainDAGRef, startup = false) =
           break
 
         if not dag.db.delBlock(fork, bid.root):
-          # Stop at the first gap - an inconsistent DB might have more blocks
-          # but we have no efficient way of detecting that
+          # Stop at the first gap - this is typically the pruning point of the
+          # previous call to pruneHistory. An inconsistent DB might have more
+          # blocks beyond that point but we have no efficient way of detecting
+          # that.
           break
 
         cur = dag.parent(bid)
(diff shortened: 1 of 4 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -50,10 +50,16 @@ declareGauge beacon_processed_deposits_total, "Number of total deposits included
 logScope: topics = "chaindag"
 
 const
-  # When finality happens, we prune historical states from the database except
-  # for a snapshort every 32 epochs from which replays can happen - there's a
-  # balance here between making long replays and saving on disk space
   EPOCHS_PER_STATE_SNAPSHOT* = 32
+    ## When finality happens, we prune historical states from the database except
+    ## for a snapshort every 32 epochs from which replays can happen - there's a
```
zah:
```suggestion
    ## for a snapshot every 32 epochs from which replays can happen - there's a
```

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1941,19 +1976,20 @@ proc pruneHistory*(dag: ChainDAGRef, startup = false) =
           break
 
         if not dag.db.delBlock(fork, bid.root):
-          # Stop at the first gap - a buggy DB might have more blocks but we
-          # have no efficient way of detecting that
+          # Stop at the first gap - an inconsistent DB might have more blocks
```
zah: Shouldn't we have some kind of logging on such "soft assertions"?
We can potentially define a catch all metric that will be increased every time something fishy is detected.

arnetheduck: As the comment says, this cannot be detected, so we cannot log it - this break in particular is the "normal" exit condition in the loop, ie the first block that is missing is the last block that we pruned in the previous pruning call - I've updated the comment which failed to mention this point and thus made this look worse than it is.

Which block exactly we end up pruning depends on how often we call prunehistory which in turn depends on many things like clocks, processing times, which slot it is and gaps on the chain so it's hard to write an exact condition that differentiates an unexpected gap from an expected one.

PR #4894, opened 2023-05-05
Sign the blinded blocks only if they provide better value than the EL block

zah:


arnetheduck (approved):

beacon_chain/validators/validator_duties.nim
```diff
@@ -698,38 +696,14 @@ proc getBuilderBid[
...
-            node, slot, validator, validator_index, forkedBlck,
-            executionPayloadHeader),
-          1.seconds):
-        Result[SBBB, string].err("getBlindedBlock timed out")
-
-  if blindedBlock.isErr:
-    return err blindedBlock.error()
-
-  return ok (blindedBlock.get, bidValue)
+  let unsignedBlindedBlock = getUnsignedBlindedBeaconBlock[SBBB](
```
arnetheduck, line 699: this construction can be done after the value comparison, assuming the forks add up - fine to leave this for another day though

tersec (approved):

PR #4905, opened 2023-05-08
embed genesis states using incbin

arnetheduck:
Specially in LTO, these are massively slow to compile

### DIFF
--- a/beacon_chain/networking/network_metadata_mainnet.s
+++ b/beacon_chain/networking/network_metadata_mainnet.s
@@ -1,3 +1,10 @@
+# beacon_chain
+# Copyright (c) 2023 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
     .section .rodata
 
 # name_data = start of data
--- a/beacon_chain/networking/network_metadata_mainnet.s
+++ b/beacon_chain/networking/network_metadata_mainnet.s
@@ -5,7 +5,7 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-    .section .rodata
+    .data
 
 # name_data = start of data
 # name_end = end of data (without alignment)
@@ -17,7 +17,7 @@ eth2_mainnet_genesis_data:
 eth2_mainnet_genesis_end:
     .global eth2_mainnet_genesis
     .type   eth2_mainnet_genesis, @object
-    .balign  8
+    .p2align 3
 eth2_mainnet_genesis:
     .quad eth2_mainnet_genesis_data
     .global eth2_mainnet_genesis_size
@@ -31,7 +31,7 @@ eth2_goerli_genesis_data:
 eth2_goerli_genesis_end:
     .global eth2_goerli_genesis
     .type   eth2_goerli_genesis, @object
-    .balign  8
+    .p2align 3
 eth2_goerli_genesis:
     .quad eth2_goerli_genesis_data
     .global eth2_goerli_genesis_size
@@ -45,7 +45,7 @@ eth2_sepolia_genesis_data:
 eth2_sepolia_genesis_end:
     .global eth2_sepolia_genesis
     .type   eth2_sepolia_genesis, @object
-    .balign  8
+    .p2align 3
 eth2_sepolia_genesis:
     .quad eth2_sepolia_genesis_data
     .global eth2_sepolia_genesis_size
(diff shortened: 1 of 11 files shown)


beacon_chain/networking/network_metadata_mainnet.s
```diff
@@ -0,0 +1,47 @@
+    .section .rodata
```
etan-status:
```suggestion
# beacon_chain
# Copyright (c) 2023 Status Research & Development GmbH
# Licensed and distributed under either of
#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
# at your option. This file may not be copied, modified, or distributed except according to those terms.

    .section .rodata
```

tests/test_network_metadata.nim
```diff
@@ -0,0 +1,32 @@
+import
```
etan-status, line 8:
```suggestion
# beacon_chain
# Copyright (c) 2023 Status Research & Development GmbH
# Licensed and distributed under either of
#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
# at your option. This file may not be copied, modified, or distributed except according to those terms.

import
```

tersec:
How portable is this?

edit: also, build time difference benchmarks?

arnetheduck:
> How portable is this? 

not at all, but that's fine because it can be selectively enabled where available - it's completely isolated that way

> benchmark


compiling the newly added test:

```
 ../env.sh nim c -d:release --passc:-flto --passl:-flto -f test_network_metadata.nim
```

pre:

```
-rw-r--r-- 1 arnetheduck arnetheduck 116M May  8 22:33 ./nimcache/release/tnm/@m..@sbeacon_chain@snetworking@snetwork_metadata.nim.c

212033 lines; 36.328s; 1.136GiB peakmem; proj: /home/arnetheduck/status/nimbus-eth2/tests/tnm.nim; out: /home/arnetheduck/status/nimbus-eth2/tests/tnm [SuccessX]
```

post:
```
-rw-r--r-- 1 arnetheduck arnetheduck 145K May  8 22:35 ./nimcache/release/tnm/@m..@sbeacon_chain@snetworking@snetwork_metadata.nim.c

212064 lines; 29.575s; 674.828MiB peakmem; proj: /home/arnetheduck/status/nimbus-eth2/tests/tnm.nim; out: /home/arnetheduck/status/nimbus-eth2/tests/tnm [SuccessX]`
```

~6-7s, 500mb peak mem usage

tersec:
The memory usage improvements are, if anything, more compelling than the compilation time improvements.

tests/test_network_metadata.nim
```diff
@@ -0,0 +1,39 @@
...
+    metadata = getMetadataForNetwork(name)
+    cfg = metadata.cfg
+    state = newClone(readSszForkedHashedBeaconState(
+      metadata.cfg, metadata.genesisData))
+
+  check:
+    $getStateRoot(state[]) == root
+
+suite "Network metadata":
+  test "Mainnet":
```
etan-status:
```suggestion
  test "mainnet":
```

etan-status, line 26:
```suggestion
  test "mainnet":
```

nfuzz/libnfuzz.nim
```diff
@@ -137,17 +144,17 @@ proc nfuzz_block_header(input: openArray[byte], xoutput: ptr byte,
 proc nfuzz_deposit(input: openArray[byte], xoutput: ptr byte,
     xoutput_size: ptr uint, disable_bls: bool): bool {.exportc, raises: [FuzzCrashError, Defect].} =
   decodeAndProcess(DepositInput):
-    process_deposit(mainnetMetadata.cfg, data.state, data.deposit, flags).isOk
+    process_deposit(getRuntimeConfig(some "mainnet"), data.state, data.deposit, flags).isOk
```
zah, line 147: What's the advantage of not referencing the constant here? 

I see several implications of the change:

* When you use the constant, the constant folding pass in the Nim compiler will make it so that the specific value is inlined here. There won't be a function call, but admittedly the overall size of the binary may increase.

* You are replacing a semantically type checked expression with a "stringy" interface that may get accidentally broken.

arnetheduck: the point of the change is to avoid accessing the incomplete "constants" (ie those without genesis) and enforce this via a function by making the constants private - they can no longer be constants because they are reconstituted at the linking stage rather than nim stage - so it's a tradeoff basically - one _could_ introduce named functions corresponding to the constants (or make separate constants for the `cfg` alone, but it feels like overkill).

I don't mind removing this change, though it introduces another kind of failure (that of using a metadata without genesis even though we have one)

arnetheduck: also, this module is practically unused, which weighed into the tradeoff of whether keeping the compile-timeness of it is valuable

zah: OK, it's possible to restore the semantic checking, but this can be done in another PR.

PR #4906, opened 2023-05-08
make invalid execution payloads more visible

tersec:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -341,16 +341,24 @@ proc getExecutionValidity(
...
+          executionPayload = shortLog(blck.message.body.execution_payload),
           blck = shortLog(blck)
         return NewPayloadStatus.invalid
       of PayloadExecutionStatus.syncing, PayloadExecutionStatus.accepted:
         return NewPayloadStatus.notValid
       of PayloadExecutionStatus.valid:
         return NewPayloadStatus.valid
   except CatchableError as err:
-    error "getExecutionValidity: newPayload failed", err = err.msg
+    error "getExecutionValidity: newPayload failed",
```
arnetheduck, line 358:
```suggestion
    error "newPayload failed",
```

arnetheduck (approved):

beacon_chain/gossip_processing/block_processor.nim
tersec: The issue here is that there is another instance of `newPayload failed`"
https://github.com/status-im/nimbus-eth2/blob/f4c215c5e6d181a0cc75a2d52cbcbf8c61c86e84/beacon_chain/gossip_processing/block_processor.nim#L321

It's important to be able to unambiguously distinguish between which is which without depending on more random details of which other fields they log as well (which sometimes works distinguishing other similar-looking logs).

So, am open to changing it in another PR, but would want it to be unambiguous, which, yes, means that it'd be not quite as minimal/succinct as otherwise possible.

arnetheduck: well, when changing, it should be done in such a way that a user reasonably can take action based on it - ie how can the message be framed so it makes sense from that perspective?

tersec: https://github.com/status-im/nimbus-eth2/pull/4930

The action is that when they report some issue, it's clearer if this particular failure mode is the root cause. Otherwise, it doesn't really show up in Nimbus logs that much except sometimes unrelated-seeming downstream logs.

Ultimately the action to take is to ensure their EL is working properly.

arnetheduck: that message is better in that it links "newPayload" to the execution client - ie that's one key piece of information the user needs - the fact that `newPayload` has something to do with the EL is not general knowledge.

One option is `Error during newPayload call from X to execution client` or something along those lines.

tersec: It's a bit more nuanced, and see https://github.com/status-im/nimbus-eth2/pull/4930 for one suggestion -- `newPayload` shouldn't be leaking any exception that `getExecutionValidity` catches, so it's kind of a bug in its own way.

PR #4911, opened 2023-05-08
accelerate `getShufflingRef`

etan-status:
When an uncached `ShufflingRef` is requested, we currently replay state which can take several seconds. Acceleration is possible by:

1. Start from any state with locked-in `get_active_validator_indices`. Any blocks / slots applied to such a state can only affect that result for future epochs, so are viable for querying target epoch. `compute_activation_exit_epoch(state.slot.epoch) > target.epoch`

2. Determine highest common ancestor among `state` and `target.blck`. At the ancestor slot, same rules re `get_active_validator_indices`. `compute_activation_exit_epoch(ancestorSlot.epoch) > target.epoch`

3. We now have a `state` that shares history with `target.blck` up through a common ancestor slot. Any blocks / slots that the `state` contains, which are not part of the `target.blck` history, affect `get_active_validator_indices` at epochs _after_ `target.epoch`.

4. Select `state.randao_mixes[N]` that is closest to common ancestor. Either direction is fine (above / below ancestor).

5. From that RANDAO mix, mix in / out all RANDAO reveals from blocks in-between. This is just an XOR operation, so fully reversible. `mix = mix xor SHA256(blck.message.body.randao_reveal)`

6. Compute the attester dependent slot from `target.epoch`. `if epoch >= 2: (target.epoch - 1).start_slot - 1 else: GENESIS_SLOT`

7. Trace back from `target.blck` to the attester dependent slot. We now have the destination for which we want to obtain RANDAO.

8. Mix in all RANDAO reveals from blocks up through the `dependentBlck`. Same method, no special handling necessary for epoch transitions.

9. Combine `get_active_validator_indices` from `state` at `target.epoch` with the recovered RANDAO value at `dependentBlck` to obtain the requested shuffling, and construct the `ShufflingRef` without replay.


etan-status:
Measurement on 2019 MacBook Pro (Mainnet):

```sh
ulimit -n 1024 && make update && make -j nimbus_beacon_node && build/nimbus_beacon_node --data-dir="$HOME/Downloads/nimbus/data/mainnet" --rest --tcp-port=9010 --udp-port=9010 --history=prune --no-el
```

> State replayed                             topics="chaindag" blocks=30 slots=64 current=1819b537:6397279@6397280 ancestor=c8043e44:6397215@6397216 target=73bd8e90:6397245@6397280 ancestorStateRoot=2c2f3a62 targetStateRoot=e9a50a6d found=false assignDur=126ms487us120ns replayDur=3s464ms169us744ns

- Old logic: 3s464ms169us744ns (state replay)
- New logic:
    - from dag.headState: 629us890ns
    - from dag.epochRefState: 570us984ns
    - from dag.clearanceState: 551us462ns

etan-status:
Another one:

> INF 2023-05-09 09:44:02.112+02:00 State replayed                             topics="chaindag" blocks=28 slots=64 current=2a896cd9:6399487@6399488 ancestor=c88d045f:6399423@6399424 target=82287782:6399451@6399488 ancestorStateRoot=f3a91737 targetStateRoot=96923c73 found=false assignDur=126ms544us527ns replayDur=3s427ms537us618ns

- Old logic: 3s427ms537us618ns (state replay)
- New logic:
    - from dag.headState: 1ms819us443ns
    - from dag.epochRefState: 1ms440us682ns
    - from dag.clearanceState: 2ms143us15ns

etan-status:
Depends on https://github.com/status-im/nimbus-eth2/pull/4910

etan-status:
Depends on https://github.com/status-im/nimbus-eth2/pull/4932

etan-status:
Tests passing locally, also should be green on GH after the two dependencies are in.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -149,6 +149,30 @@ func get_ancestor*(blck: BlockRef, slot: Slot,
 
     blck = blck.parent
 
+func commonAncestor*(a, b: BlockRef, lowSlot: Slot): Opt[BlockRef] =
```
arnetheduck, line 152: this function should have its own tests

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -490,6 +492,31 @@ func putShufflingRef*(dag: ChainDAGRef, shufflingRef: ShufflingRef) =
...
   dag.shufflingRefs.put shufflingRef
 
+func findShufflingRef*(
+    dag: ChainDAGRef, bid: BlockId, epoch: Epoch): Opt[ShufflingRef] =
+  ## Lookup a shuffling in the cache, returning `none` if it's not present - see
+  ## `getShufflingRef` for a version that creates a new instance if it's missing
+  let
+    dependent_slot = epoch.shufflingDependentSlot
+    dependent_bsi = dag.atSlot(bid, dependent_slot).valueOr:
+      return err()
```
arnetheduck:
```suggestion
    dependent_bsi = ? dag.atSlot(bid, dependent_slot)
```

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1299,23 +1325,272 @@ proc getFinalizedEpochRef*(dag: ChainDAGRef): EpochRef =
...
+  const numDelayEpochs = compute_activation_exit_epoch(GENESIS_EPOCH).uint64
+  let
+    lowEpoch = max(epoch, (numDelayEpochs - 1).Epoch) - (numDelayEpochs - 1)
+    lowSlot = lowEpoch.start_slot
+  if state.data.slot < lowSlot or blck.slot < lowSlot:
+    return err()
+
+  # Check that state is related to the information stored in the DAG,
+  # and determine the corresponding `BlockRef`, or `finalizedHead` if finalized
+  let
```
zah, line 1361: Can you add a clarifying comment explaining why the finalized block in particular is a good starting point for re-calculating the RANDAO. 

If a shuffling is requested for a very old state, wouldn't it still be better to find the nearest state snapshot in the database and start the optimised RANDAO computation from there? Is there an assumption that we are never computing the shuffling for such old epochs?

etan-status: Shuffling consists of two parts:

1. `get_active_validator_indices` at target epoch
2. RANDAO at dependent slot

--

1. can be queried from _any_ state that includes history up to ~5 epochs before target. If it's finalized, that means, any state can be used. you just go through validators and check which ones were active at that particular epoch.

2. can be recomputed from any history that has a common ancestor. if it is a finalized part, just start at the historic `randao_mixes` for the requested epoch (or the one before / after, if closer) and apply blocks from there.

etan-status: Here, I just want a `BlockRef`, and they don't exist for the finalized portion of the chain (that part only has `BlockId`)

etan-status: If you are compatible, the work to recover RANDAO is ~constant, as there is a checkpoint stored in `BeaconState` each epoch.

You don't have to replay _all_ the epochs to recover it, can start from closest checkpoint.

etan-status: > Is there an assumption that we are never computing the shuffling for such old epochs?

Yes, gossip validation ignores very old attestations before loading the shuffling.

zah (approved):

etan-status:
@arnetheduck wanted to do a more thorough review on this as well, so please wait with merge until that's done.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -149,6 +149,30 @@ func get_ancestor*(blck: BlockRef, slot: Slot,
...
+
+  var
+    aa = a
+    bb = b
+  while aa != bb:
+    if aa.slot >= bb.slot:
+      aa = aa.parent
+      doAssert aa != nil, "All `BlockRef` lead to `finalizedHead`"
+      if aa.slot < lowSlot:
+        return err()
```
arnetheduck, line 168: `Opt.none()` looks better with Opt

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1314,23 +1340,272 @@ proc getFinalizedEpochRef*(dag: ChainDAGRef): EpochRef =
...
+    stateBid = state.latest_block_id
+    stateBlck =
+      if dag.finalizedHead.blck == nil:
```
arnetheduck, line 1364: finalizedHead is always not nil

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1314,23 +1340,272 @@ proc getFinalizedEpochRef*(dag: ChainDAGRef): EpochRef =
...
+        return err()
+      elif stateBid.slot > dag.finalizedHead.blck.slot:
+        ? dag.getBlockRef(stateBid.root)
+      elif stateBid.slot == dag.finalizedHead.blck.slot:
+        if stateBid.root != dag.finalizedHead.blck.root:
```
arnetheduck, line 1369: this would be a bug

arnetheduck: in fact, all of this can be replaced byt `stateBlck = getBlockRef(stateBit.root) or dag.finalizedHead.blck` (`getBlockRef` returns the finalized blck in applicable cases)

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1314,23 +1340,272 @@ proc getFinalizedEpochRef*(dag: ChainDAGRef): EpochRef =
...
+  withBlck(bdata):  # See `process_randao` / `process_randao_mixes_reset`
+    mix.data.mxor eth2digest(blck.message.body.randao_reveal.toRaw()).data
+  ok()
+
+proc computeRandaoMix*(
+    dag: ChainDAGRef, state: ForkyHashedBeaconState,
+    blck: BlockRef, epoch: Epoch
+): Opt[tuple[dependentBid: BlockId, mix: Eth2Digest]] =
+  ## Compute the requested RANDAO mix for `blck@epoch` based on `state`.
+  ## `state` must have the correct `get_active_validator_indices` for `epoch`.
```
arnetheduck, line 1416: this seems like an unnecessary constraint _for this function_ - ie the requirement here is that the state shares an ancestor within `EPOCHS_PER_HISTORICAL_VECTOR` slots (in a non-finalizing history, this will not always be true, even for a blckref).

ie randao recovery / mixing is orthogonal to shuffling and the logic would ideally reflect this (the shuffling is more strongly constrained)

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: minus the unnecessary `getBlockRef` scan if we already know that it is finalized. Good to know that this doesn't need as much defense as have been put here. So, current logic should be fine, it can just be rewritten to be more concise, as I understand.

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: that could be slow to process though, if we relax the precondition, and start from a very old state, it would mean adding many thousands of blocks. Agree though, the result should still be correct, but the function is not intended to be used that way.

arnetheduck: so, performance is not really the question here, but rather clarity of the implementation: mixing orthogonal constraints creates a false dependency which makes the code confusing.

The randao mixing time is essentially linear in the number of blocks that need to be replayed no matter the depth, including the special case of shuffling.

In particular, this function could be used in the REST interface that returns randao: that API would be a lot more efficient if it didn't do a full state replay and it would benefit from the full range.

arnetheduck: The other reason why this is important is to keep the constraints in tune with respect to the limits that the state itself provides: `EPOCHS_PER_HISTORICAL_VECTOR` is also tied to the distance at which `get_block_root_at_slot` return correct results etc, so in analyzing the correctness of this function, it's better that its constraints are directly expressed in the concepts that cause the constraint.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1314,23 +1340,272 @@ proc getFinalizedEpochRef*(dag: ChainDAGRef): EpochRef =
...
+      while highSlot > lowSlot and
+          state.data.get_block_root_at_slot(highSlot - 1) == highRoot:
+        dec highSlot
+      if highSlot + SLOTS_PER_HISTORICAL_ROOT > state.data.slot:
+        BlockId(slot: highSlot, root: highRoot)
+      else:
+        let bsi = ? dag.getBlockIdAtSlot(highSlot)
+        doAssert bsi.bid.root == highRoot
+        bsi.bid
+  while bid.slot >= slotsToMix.a:
```
arnetheduck, line 1488: do we have a test for when `slotsToMix.a` is an empty slot?

etan-status: In the way how it is called, don't think that's possible, as `slotsToMix` is based on the common ancestor block. But sure, can add such a test as well.

beacon_chain/consensus_object_pools/blockchain_dag.nim
arnetheduck: per discussion, this algorithm could be simplified in the following manner:

* because `xor` is commutative, we can apply the `commonAncestor` logic directly by walking blocks via the parent root if the block data loaded together with randao instead of going via blockref/getblockatslot/etc - this reduces the components involved in computing the randao and thus increases its robustness
  * key insight here is that we can walk state and desired mixes in any order, rather than first "undoing" one then "doing" the other
  * this should reduce the number of moving parts in the code, reduce off-by-one in slot logic, handle empty slots more gracefully etc
* minimal pre-validation still needs to be done so that we don't end up with overlong walks that end up failing in case of long periods of non-finality
  * the code could further be generalised by taking a `BlockId` as input, though this would complicate pre-checking slightly
* it "should" be possible to load the randao/parent from the block without decoding all of the block - this is completely orthogonal to this PR but an interesting idea for the future

Issue #3198, opened 2021-12-15
Provide pre-compiled binary support for Broadwell (ADX, BMI2) binaries


tersec:
This is phrased to encompass any of several approaches to this (e.g., wholly distinct monolithic binary images, runtime-selected shared libraries, et cetera).

All `block_sim` runs used 2000 validators and 160 slots. They were randomly interleaved to avoid one set being coincidentally aligned with something external potentially affecting performance.

With `-march=haswell`, the Intel x86 microarchitecture immediately preceding Broadwell, 15 sampled times for this `block_sim` configuration:
```
81.23s
79.44s
80.08s
79.53s
77.11s
78.15s
77.02s
78.33s
77.67s
76.66s
77.34s
77.44s
79.65s
80.05s
80.14s
```

With `-march=broadwell`, 15 sampled times for this `block_sim` configuration:
```
62.25s
61.81s
61.86s
61.21s
60.70s
61.16s
61.66s
62.06s
60.22s
61.70s
60.54s
60.79s
63.36s
62.83s
61.26s
```

https://github.com/status-im/nimbus-eth2/blob/stable/docs/cpu_features.md#bmi2--adx explains this effect and provides further microbenchmarks.

tersec:
https://lighthouse-book.sigmaprime.io/installation-binaries.html documents Lighthouse's approach.

arnetheduck:
my preferred solution would involve dynamic runtime selection rather than separate binaries - one approach is to fork blst such that the different platform options each get a fully separate symbol namespace (at the C level), avoiding the need to do dynamic loading (which is messy for all kinds of reasons) - we'd probably have to do a bit of C surgery to get that done, but hopefully it can also be upstreamed - there's a lot of demand for this.

arnetheduck:
their objections so far have been that the optimized operations are too "low-level" to warrant dynamic runtime selection - instead, I'd aim to get a separate symbol namespace all the way to the "black box primitives", ie `sign_adx`, `sign_generic` etc and a `sign` that either uses [ifunc](https://sourceware.org/glibc/wiki/GNU_IFUNC) or simple pointers based on cpuid

tersec:
[...]

PR #4914, opened 2023-05-09
add optimized Linux amd64 flavor

tersec:
https://github.com/status-im/nimbus-eth2/issues/3198

This does not do dynamic runtime selection, but does demonstrate another approach.

Have confirmed that per https://github.com/status-im/nimbus-eth2/blob/stable/docs/cpu_features.md#bmi2--adx the target it adds statically includes all of the `mulx`, `adox`, and `adcx` instructions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -802,6 +802,10 @@ dist-amd64:
 	+ MAKE="$(MAKE)" \
 		scripts/make_dist.sh amd64
 
+dist-amd64-opt:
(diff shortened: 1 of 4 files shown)


arnetheduck:
geth is looking into runtime detection in blst, ergo this PR is likely to be obsolete soon - this is not something we want to support over time either, ie releasing is complex enough as-is

tersec:
Sure.

PR #4922, opened 2023-05-09
cleanup `state_transition_epoch` and bump to v1.3.0

etan-status:
More v1.3.0 consensus-specs bumps, focused on `state_transition_epoch`. Also fixed `current_epoch` spurious style check warning, and cleanup.

### DIFF
--- a/beacon_chain/spec/state_transition_epoch.nim
+++ b/beacon_chain/spec/state_transition_epoch.nim
@@ -6,7 +6,7 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 # State transition - epoch processing, as described in
-# https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.5/specs/phase0/beacon-chain.md#epoch-processing
+# https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/phase0/beacon-chain.md#epoch-processing
 # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/altair/beacon-chain.md#epoch-processing
 # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/bellatrix/beacon-chain.md#epoch-processing
 # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/capella/beacon-chain.md#epoch-processing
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -175,11 +175,12 @@ func is_eligible_validator*(validator: ParticipationInfo): bool =
...
 from ./datatypes/deneb import BeaconState
 
-# https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.3/specs/altair/beacon-chain.md#get_unslashed_participating_indices
-# https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.5/specs/phase0/beacon-chain.md#get_total_balance
+# https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/altair/beacon-chain.md#get_unslashed_participating_indices
 func get_unslashed_participating_balances*(
     state: altair.BeaconState | bellatrix.BeaconState | capella.BeaconState |
            deneb.BeaconState): UnslashedParticipatingBalances =
+  ## Return the set of validator indices that are both active and unslashed
+  ## for the given ``flag_index`` and ``epoch``.
```
etan-status:
```suggestion
```

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -1074,28 +1074,27 @@ func process_historical_summaries_update*(
...
-# https://github.com/ethereum/consensus-specs/blob/v1.3.0-rc.3/specs/phase0/beacon-chain.md#epoch-processing
+# https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/phase0/beacon-chain.md#epoch-processing
 proc process_epoch*(
     cfg: RuntimeConfig, state: var phase0.BeaconState, flags: UpdateFlags,
     cache: var StateCache, info: var phase0.EpochInfo): Result[void, cstring] =
-  let currentEpoch = get_current_epoch(state)
-  trace "process_epoch",
-    current_epoch = currentEpoch
+  let epoch = get_current_epoch(state)
+  trace "process_epoch", epoch
```
etan-status, line 1082: if it's called `currentEpoch` it glitches style check, because there is a `current_epoch` template in use as well.

tersec: Yeah, I had noticed this as well.

It's funny because `epoch` is also an overloaded symbol but it happens not to trip the style checker in the same way, so at least it's an improvement.

tersec: As long as there were/are other style checking symbol names more difficult to reasonably reconcile (e.g., there ware some where there's both an EL version with one case/underscore convention and a CL version with another such convention, and they collide in `nimbus-eth2`), I hadn't much pursued the remainder of these, but it's nice to whittle this down at least to be able to focus on the more difficult cases by resolving incidental ones.

etan-status: The renamed one was already called `epoch` in a different log, so it actually improves consistency as well. It's also only used for CI strict verification.

tersec (approved):

PR #4923, opened 2023-05-09
update weak subjectivity calculations

etan-status:
Weak subjectivity logic (`--weak-subjectivity-checkpoint`) was outdated. Updated to latest specs, also taking into account total active balance.

See https://github.com/ethereum/consensus-specs/pull/2190

### DIFF
--- a/beacon_chain/spec/weak_subjectivity.nim
+++ b/beacon_chain/spec/weak_subjectivity.nim
@@ -8,35 +8,59 @@
 {.push raises: [].}
 
 import
-  ./datatypes/base, ./forks, ./helpers
+  ./datatypes/base, ./beaconstate, ./forks, ./helpers
 
-const
-  SAFETY_DECAY* = 10'u64
+# https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/phase0/weak-subjectivity.md#configuration
+const SAFETY_DECAY* = 10'u64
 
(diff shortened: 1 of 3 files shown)


beacon_chain/spec/weak_subjectivity.nim
```diff
@@ -8,35 +8,59 @@
...
+  var
+    cache: StateCache
+    ws_period = cfg.MIN_VALIDATOR_WITHDRAWABILITY_DELAY
+  let
+    N = get_active_validator_indices_len(state, get_current_epoch(state))
+    t = (get_total_active_balance(state, cache) div N).toEther
+  const T = MAX_EFFECTIVE_BALANCE.toEther
+  let delta = cfg.get_validator_churn_limit(state, cache)
+  const
+    Delta = MAX_DEPOSITS * SLOTS_PER_EPOCH
```
tersec, line 36: Interesting that at least similar code in isolation doesn't set up the style checker and does appear to work fine:
```nim
$ cat a.nim
let delta = 3
const Delta = 6
echo "delta = ", delta
echo "Delta = ", Delta
$ nim c --styleCheck:usages --styleCheck:error --hints:off -r a
delta = 3
Delta = 6
```

tersec (approved):

beacon_chain/spec/weak_subjectivity.nim
etan-status: In Nim, the first letter case has semantic meaning, it's only subsequent letters and _ that are equivalent semantically

PR #4925, opened 2023-05-09
add Capella constants to REST config endpoint

etan-status:
Capella constants were missing; if we want to support BLS changes from VC, we need the new domain constant to check compatibility.


beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -278,25 +278,31 @@ type
...
     extra_data*: List[byte, MAX_EXTRA_DATA_BYTES]
     base_fee_per_gas*: UInt256
 
     # Extra payload fields
-    block_hash*: Eth2Digest # Hash of execution block
+    block_hash*: Eth2Digest
+      ## Hash of execution block
     transactions*: List[Transaction, MAX_TRANSACTIONS_PER_PAYLOAD]
-    withdrawals*: Option[List[Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD]]  # [New in Capella]
+    withdrawals*: Option[List[Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD]]
```
cheatfate, line 304: I think this should not be `Option` anymore.

etan-status: Hmm, maybe better to combine that with a Deneb compatibility PR (adding `excess_data_gas`).

Would it still work on local testnets without the `Option`?

PR #4927, opened 2023-05-09
final portion of non-trivial v1.3.0 bumps

etan-status:
Updates unchanged logic to latest v1.3.0 consensus-specs refs, and cleans up surrounding sections / syncs comments, and so on.

```
https://github.com/ethereum/consensus-specs/(blob|tree)/(?!v1\.3\.0/)
```

### DIFF
--- a/beacon_chain/spec/datatypes/capella.nim
+++ b/beacon_chain/spec/datatypes/capella.nim
@@ -65,13 +65,17 @@ type
 
   # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/capella/beacon-chain.md#executionpayload
   ExecutionPayload* = object
+    # Execution block header fields
     parent_hash*: Eth2Digest
-    fee_recipient*: ExecutionAddress  # 'beneficiary' in the yellow paper
+    fee_recipient*: ExecutionAddress
+      ## 'beneficiary' in the yellow paper
     state_root*: Eth2Digest
-    receipts_root*: Eth2Digest # 'receipts root' in the yellow paper
+    receipts_root*: Eth2Digest
     logs_bloom*: BloomLogs
-    prev_randao*: Eth2Digest  # 'difficulty' in the yellow paper
-    block_number*: uint64  # 'number' in the yellow paper
+    prev_randao*: Eth2Digest
+      ## 'difficulty' in the yellow paper
+    block_number*: uint64
+      ## 'number' in the yellow paper
     gas_limit*: uint64
     gas_used*: uint64
     timestamp*: uint64
@@ -79,9 +83,11 @@ type
     base_fee_per_gas*: UInt256
 
     # Extra payload fields
-    block_hash*: Eth2Digest # Hash of execution block
+    block_hash*: Eth2Digest
+      ## Hash of execution block
     transactions*: List[Transaction, MAX_TRANSACTIONS_PER_PAYLOAD]
-    withdrawals*: List[Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD]  # [New in Capella]
+    withdrawals*: List[Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD]
+      ## [New in Capella]
 
   ExecutionPayloadForSigning* = object
     executionPayload*: ExecutionPayload
@@ -89,6 +95,7 @@ type
 
   # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/capella/beacon-chain.md#executionpayloadheader
   ExecutionPayloadHeader* = object
+    # Execution block header fields
     parent_hash*: Eth2Digest
     fee_recipient*: ExecutionAddress
     state_root*: Eth2Digest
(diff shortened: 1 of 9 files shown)


beacon_chain/sync/sync_protocol.nim
```diff
@@ -534,8 +537,10 @@ p2pProtocol BeaconSync(version = 1,
             continue
 
           # TODO extract from libp2pProtocol
-          peer.awaitQuota(blockResponseCost, "blobs_sidecars_by_range/1")
```
tersec: Not sure what the motivation for splitting this line is? It's 73 characters.

I guess there's some point for consistency in splitting every line, but that as policy creates its own fragility and isn't really rooted in the Status Nim style guide that I can tell, which only mentions in this regard the 80 character limit from NEP-1 of which I'm aware.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -465,8 +465,10 @@ p2pProtocol BeaconSync(version = 1,
             bytes = bytes.len(), blck = shortLog(blockRef), blobindex = index
           continue
 
-        peer.awaitQuota(blockResponseCost, "blob_sidecars_by_root/1")
```
tersec: 68 characters

https://github.com/status-im/nimbus-eth2/pull/4927#discussion_r1189342315

beacon_chain/sync/sync_protocol.nim
```diff
@@ -465,8 +465,10 @@ p2pProtocol BeaconSync(version = 1,
...
-        peer.network.awaitQuota(blockResponseCost, "blob_sidecars_by_root/1")
```
tersec: 78 characters

https://github.com/status-im/nimbus-eth2/pull/4927#discussion_r1189342315

beacon_chain/spec/datatypes/capella.nim
```diff
@@ -65,30 +65,37 @@ type
...
+    withdrawals*: List[Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD]
+      ## [New in Capella]
 
   ExecutionPayloadForSigning* = object
     executionPayload*: ExecutionPayload
     blockValue*: Wei
 
   # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/capella/beacon-chain.md#executionpayloadheader
   ExecutionPayloadHeader* = object
+    # Execution block header fields
```
tersec, line 98: Alright, sure, this does appear to be from the spec. Strange inclusion there though since it's not their usual approach. All of the consensus spec objects are the `${FOO}` fields.

etan-status: Yeah, I guess either have both "section headers", or none of them in our copy. As the other one for the extra fields is already there, added this one as well for consistency.

tersec: Sure, agree.

beacon_chain/sync/sync_protocol.nim
etan-status: image
it exceeds the 80 column line for me. but can remove them from this pr, has nothing to do with the spec bumps

tersec:
```
>>> len('          peer.awaitQuota(blockResponseCost, "blobs_sidecars_by_range/1")')
73
>>> len('          peer.network.awaitQuota(blockResponseCost, "blobs_sidecars_by_range/1")')
81
```
But the actual removed line is the 73 byte string there, not the 81 byte string.

etan-status: yeah, did all 4 for consistency. now is back to the previous one (with the one line at 81), at least consistent :)

tersec: This is the case with all of https://github.com/status-im/nimbus-eth2/pull/4927/commits/63dcfe3b6a7f08c6423789916a7436ae28a249af -- there are two parts of each pair, and the `await peer.network.awaitQuota` one can overflow the 80 character limit even when the `await peer.awaitQuota` version does not.

tersec: For one such pair neither does though:
```
>>> len('        peer.network.awaitQuota(blockResponseCost, "blob_sidecars_by_root/1")')
77
```

tersec: > yeah, did all 4 for consistency. now is back to the previous one (with the one line at 81), at least consistent :)

Sure, reasonable enough. If nothing else, it doesn't have to be part of this PR, even if it's worth changing.

beacon_chain/sync/sync_protocol.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4927/commits/63dcfe3b6a7f08c6423789916a7436ae28a249af

beacon_chain/sync/sync_protocol.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4927/commits/63dcfe3b6a7f08c6423789916a7436ae28a249af

tersec (approved):

PR #4930, opened 2023-05-10
modify newPayload failure logging

tersec:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -355,7 +355,7 @@ proc getExecutionValidity(
       of PayloadExecutionStatus.valid:
         return NewPayloadStatus.valid
   except CatchableError as err:
-    error "getExecutionValidity: newPayload failed",
+    error "newPayload failed and leaked exception",
```
zah, line 358: Can you clarify what "leaked exception" means in this context? If you are referring to Nim exceptions, the message is not very well suited for our users, who are not expected to be programmers.

tersec: In theory, the only exceptions should be appearing from
https://github.com/status-im/nimbus-eth2/blob/4873f8bdc443a6cd4e5b53a52fd6ef92190e4191/beacon_chain/gossip_processing/block_processor.nim#L336-L343

which already contains
https://github.com/status-im/nimbus-eth2/blob/4873f8bdc443a6cd4e5b53a52fd6ef92190e4191/beacon_chain/gossip_processing/block_processor.nim#L309-L326

that is, there shouldn't be any exceptions catchable here which the `newExecutionPayload` function doesn't already catch and log.

So another option would be to remove this `try/catch` altogether.

I've tended not to do that because there's no compile-time enforcement of this invariant continuing to hold true, and debugging these async exceptions is unfortunate enough to want to handle the possibility ahead of time.

But it also means that it's theoretically no more likely to appear than some `doAssert`, so the exact message just isn't as important, because it's closer to a `panic` in that case.

PR #4935, opened 2023-05-10
don't run flaky CI in Windows

tersec:
9 times in the last day alone: https://github.com/status-im/nimbus-eth2/issues/4619#issuecomment-1540583630


tests/all_tests.nim
```diff
@@ -54,8 +54,9 @@ when not defined(i386):
   # Avoids "Out of memory" CI failures
   import
     ./test_blockchain_dag,
-    ./test_keystore,
-    ./test_keystore_management,
-    ./test_keymanager_api
```
zah, line 60: The flaky tests are `test_keymanager_api` and `test_signing_node`. `test_keystore_management` is mostly doing file I/O and `test_keystore` is mostly about the serialization and cryptography and these have been running smoothly.

etan-status (approved):

tests/all_tests.nim
etan-status: @zah, let's please prioritize fixing the tests then. Flaky CI is super time consuming for everyone (not just the maintainers of these flaky tests); the tests should be disabled in CI until they are capable of running reliably. When they run reliably again, just add them back to run on CI.

PR #4938, opened 2023-05-11
Update "Getting started" part of the book

narimiran:
The "Getting started" quickstart guide should contain information on how to, well, _quickly start_ with Nimbus (and not more) — the metrics and migration how-to guides are moved to the how-to section of the book, as that is more suited.

`philosophy.md` is removed and its contents are part of the index page. This way the design goals (and Nimbus strengths) are immediately seen, and (IMO even more importantly) now when you click on the 'next' on the landing page, you're directly taken to the getting started with beacon node guide.

Build prerequisites are now part of the "Prepare your machine" (`install.md`), as this is cleaner, avoids duplication and unnecessary information (this is linked both from the 'getting started' guide and the 'build from source' how-to).

The quickstart guide now offers both 'download binaries' (link to the binaries guide, instead of duplication) and 'build from source' options for installing Nimbus.

Other than those, several minor (cosmetic, if you will) fixes and improvements.

Here's a quick before/after screenshot:


narimiran:
Can I get a review?

docs/the_nimbus_book/src/install.md
```diff
@@ -8,8 +8,68 @@ Check that your machine matches the [minimal system requirements](./hardware.md)
...
+    On common Linux distributions the dependencies can be installed with:
+
+    ```sh
+    # Debian and Ubuntu
+    sudo apt-get install build-essential git cmake
+
+    # Fedora
+    dnf install @development-tools cmake
+
+    # Archlinux, using an AUR manager
```
tersec:
```suggestion
    # Arch Linux, using an AUR manager
```
Yes, there's a question of whether contextually it should be just "Arch", (is it really called "Arch Linux", or is it a distribution of Linux named "Arch"), but e.g., https://en.wikipedia.org/wiki/Arch_Linux https://en.wikipedia.org/wiki/Debian https://en.wikipedia.org/wiki/Ubuntu https://en.wikipedia.org/wiki/Fedora_(operating_system) suggest that specifically Arch Linux has decided to keep "Linux" as part of its name.

docs/the_nimbus_book/src/eth1.md
```diff
@@ -1,16 +1,15 @@
...
+Relying on third-party services such as Infura, Alchemy and Pocket is no longer possible.
+Sharing the same execution client between multiple beacon nodes is not supported.
 
-Nimbus has been tested all major execution clients, see the [execution client comparison](https://ethereum.org/en/developers/docs/nodes-and-clients/#execution-clients) for more information.
-
-!!! warning
-    You need to run your own execution client.
-    Relying on third-party services such as Infura, Alchemy and Pocket is no longer possible.
-    Sharing the same execution client between multiple beacon nodes is not supported.
+Nimbus has been tested all major execution clients.
```
tersec:
```suggestion
Nimbus has been tested with all major execution clients.
```

docs/the_nimbus_book/src/el-light-client.md
```diff
@@ -12,13 +12,13 @@ Since the merge 🐼, execution clients can no longer run standalone.
...
+| Feature | Beacon Node | Light Client |
 | -- | -- | -- |
-| Disk usage | **<1MB** | ~70GB |
-| Bandwidth | **TBD (low)** | *TBD* |
-| Sync time | **Seconds** | Days |
-| Head delay | 4/3 slot (15 s) | **None** |
-| Security | Light | **Full** |
+| Disk usage | ~70GB | **<1MB** |
+| Bandwidth | *TBD* | **TBD (low)** |
+| Sync time | Days | **Seconds** |
```
tersec: Is it worth addressing usage of https://nimbus.guide/trusted-node-sync.html in the context of sync time?

It's true that even backfill (I hesitate to call it "backwards sync", because at some level the node is synced enough to perform validation duties immediately) to [MIN_EPOCHS_FOR_BLOCK_REQUESTS](https://github.com/ethereum/consensus-specs/blob/v1.4.0-alpha.1/specs/phase0/p2p-interface.md#configuration) might require days, though it should typically be more hours, since it's not rerunning actual state transitions and not going back to genesis, so this is not inaccurate as such.

Indeed, at some point the EF might start more strongly and explicitly recommending against (forwards) genesis sync, since it's before the [weak subjectivity period](https://github.com/ethereum/consensus-specs/blob/v1.4.0-alpha.1/specs/phase0/weak-subjectivity.md#weak-subjectivity-period), so I'm not sure how relevant this sort of "sync time" remains or will remain.

https://nimbus.guide/trusted-node-sync.html#delay-block-history-backfill though allows
> to get started more quickly by delaying the backfill of the block history using the --backfill=false parameter. In this case, the beacon node will first sync to the current head so that it can start performing its duties, then backfill the blocks from the network.

So this summary is a bit incomplete.

docs/the_nimbus_book/src/hardware.md
```diff
@@ -9,11 +9,15 @@ The recommended system requirements for running the Nimbus beacon node are:
...
 
-    Broadly, to run both an execution and a consensus client on the same machine, we recommend a **2 TB** SSD drive and **8 GB RAM**.
+### Execution client
 
-!!! note "Minimal requirements"
-    Nimbus has been optimized to also run well on hardware significantly less powerful than the recommended system requirements — the more validators you run on the same node, the more hardware resources and network bandwidth will it will use.
+In addition to the beacon node, you will need to run an [execution client](./eth1.md).
+Check the documentation of the client of choice and add them to the above requirements.
+
+Broadly, to run both an execution and a consensus client on the same machine, we recommend a **2 TB** SSD drive and **8 GB RAM**.
```
tersec:
```suggestion
Broadly, to run both an execution and a consensus client on the same machine, we recommend a **2 TB** SSD and **16 GB RAM**.
```
It might be possible with 8GB, and carefully configured Geth or some other EL (though probably not Erigon; Nethermind says they've reduced RAM requirements with 1.18 and 1.19, so maybe it's feasible now with them), but 16GB is a safer recommendation.

If someone asks whether it's, rather, possible, that's a more nuanced discussion, the answer to which is "maybe", but to set expectations, better to start with 16GB.

docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -1,30 +1,25 @@
...
-!!! tip "No validator client needed"
-    Unlike other beacon chain clients, Nimbus does not require setting up a separate validator client process — the beacon node can itself perform validator duties.
+Unlike other beacon chain clients, Nimbus **doesn't require** setting up a separate validator client process — the beacon node can itself perform validator duties.
+This is a simple, safe and efficient way to get started.
 
-    This is a simple, safe and efficient way to get started.
+!!! tip "Separate validator client"
 
-    Advanced users may want to use a separate [validator client](./validator-client.md) process instead.
+    While not needed, advanced users may want to use a [separate validator client](./validator-client.md) process instead.
```
tersec:
```suggestion
    While not needed, advanced users may want to use a [separate validator client](./validator-client.md) instead.
```
It's not false that there's a separate process, but also a bit beside the point.

docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -1,30 +1,25 @@
-# Run a validator
+# Validating
 
 Once your beacon node is [running](./quick-start.md), the next step is to set up a validator.
 
-!!! tip "No validator client needed"
-    Unlike other beacon chain clients, Nimbus does not require setting up a separate validator client process — the beacon node can itself perform validator duties.
+Unlike other beacon chain clients, Nimbus **doesn't require** setting up a separate validator client process — the beacon node can itself perform validator duties.
```
tersec: Teku also supports this mode: https://docs.teku.consensys.net/get-started/start-teku#start-the-clients-in-a-single-process

Erigon is also in the process of implementing this: https://erigon.gitbook.io/erigon/advanced-usage/consensus-layer https://erigon.substack.com/p/announcing-caplin-erigon-full-consensus

There's no need to enumerate this in the documentation, just ensure the phrasing implicitly admits for the existence of these.

docs/the_nimbus_book/src/pi-guide.md
```diff
@@ -41,12 +41,13 @@ As such, we try our best to explain things from first-principles.
 
 ### 1. Download Raspberry Pi Imager
 
-[Raspberry Pi Imager](https://www.raspberrypi.org/blog/raspberry-pi-imager-imaging-utility/) is an imaging utility that makes it simple to manage your microSD card with Raspbian (the free Pi operating system based on Debian).
+[Raspberry Pi Imager](https://www.raspberrypi.org/blog/raspberry-pi-imager-imaging-utility/) is an imaging utility that makes it simple to manage your microSD card with Raspberry Pi OS (the free Pi operating system based on Debian, previously called Raspbian).
```
tersec, line 44: This is not a rhetorical but real question, the answer to which I'm not suggesting but genuinely asking: is it worth calling out the Raspbian to Raspberry Pi OS name change? Presumably people who have it installed already will have realized it (or not?) more to the point, anyone following these instructions from the guide might not find it maximally salient.

docs/the_nimbus_book/src/pi-guide.md
```diff
@@ -2,7 +2,7 @@
 
 <blockquote class="twitter-tweet"><p lang="en" dir="ltr">I expect the new Raspberry Pi 4 (4GB RAM option, external SSD) to handle an Eth2 validator node without breaking a sweat. That&#39;s $100 of hardware running at 10 Watts to support a 32 ETH node (currently ~$10K stake).</p>&mdash; Justin Ðrake (@drakefjustin) <a href="https://twitter.com/drakefjustin/status/1143091047058366465?ref_src=twsrc%5Etfw">June 24, 2019</a></blockquote> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
```
tersec, line 3: Is the "Ð" in
> Justin Ðrake
intentional?

docs/the_nimbus_book/src/hardware.md
tersec: Also, for example, Joe Clapis has been testing exclusively with 16GB for a while so far as I know.

docs/the_nimbus_book/src/eth1.md
```diff
@@ -1,16 +1,15 @@
...
+See the [execution client comparison](https://ethereum.org/en/developers/docs/nodes-and-clients/#execution-clients) for more information.
 
 !!! info
-    Syncing an execution client may take hours or even days, depending on your hardware!
+    Syncing an execution client **may take hours or even days**, depending on your hardware!
```
tersec, line 11: One aspect that might be worth mentioning, or not -- it seems to be kind of folklore-ish knowledge not widely documented that I've found, though it might just be that I've not seen such documentation: there's a threshold IOPS (roughly) below which each different EL client simply won't finish syncing because it has no surplus I/O for the purpose.

Various people have tested/benchmarked this, and typical SSDs, both SATA and NVMe, are fine, but every so often someone comes to the Nimbus Discord server reporting a never-ending EL client sync, and it's typically with such low-end hardware or low-quality VPS hosting.

However, it's kind of a caveat that hopefully will be decreasingly relevant, and maybe a bit discouraging to present up-front given that it's both uncommon and not catastrophic. So I'm not sure here is the best place to document this.

docs/the_nimbus_book/src/hardware.md
```diff
@@ -9,11 +9,15 @@ The recommended system requirements for running the Nimbus beacon node are:
...
+
+
+### Minimal requirements
+
+Nimbus has been optimized to also run well on hardware significantly less powerful than the recommended system requirements — the more validators you run on the same node, the more hardware resources and network bandwidth will it will use.
```
tersec, line 23: https://github.com/ethereum/consensus-specs/pull/3312 has removed network bandwidth being a function of number of validators. It's not implemented, yet, by Nimbus, but we're not waiting for any particular additional event, so probably not worth mentioning something which presently will become (mostly: sending an additional attestation per epoch per validator is mostly in the noise) false.

tersec: Actually -- will retract this -- there's less additional bandwidth usage, but still some, because this affects only the stability subnets, which are substantial but not the only additional usage. So keep unchanged.

docs/the_nimbus_book/src/quick-start.md
```diff
@@ -27,32 +27,28 @@ Prepare your machine by installing [Nimbus' dependencies](./install.md).
...
+    Binary releases are available from [GitHub](https://github.com/status-im/nimbus-eth2/releases/latest) and our [APT repository](https://apt.status.im/) (Debian/Ubuntu).
 
-# Download the latest release - replace the link with the latest release on the download page!
-wget https://github.com/status-im/nimbus-eth2/releases/download/v22.10.1/nimbus-eth2_Linux_amd64_22.10.1_97a1cdc4.tar.gz
+    We currently have binaries available for Linux `AMD64`, `ARM` and `ARM64`, Windows `AMD64` and macOS (`AMD64` and `ARM64`).
 
-# Unpack the archive into the `nimbus-eth2` directory you just created
-tar xvf nimbus-eth2_Linux_amd64_22.10.1_97a1cdc4.tar.gz --strip-components 1 -C nimbus-eth2
-```
+    See [this guide](./binaries.md) on how to install them.
```
tersec: Because the Nimbus guide as a whole is a "guide", there's ambiguity in terms of the referent of "this": "this" guide might be read as self-referential to `quick-start.md` or only generally referential to the Nimbus guide as a whole. Of course, the link is fine, and disambiguates, but maybe better to use some other word than "this" here.

narimiran: Very valid remark! All of "this ..." links should be more concrete! I'll fix it!

docs/the_nimbus_book/src/pi-guide.md
narimiran: Yep, that's how it is: https://twitter.com/drakefjustin

docs/the_nimbus_book/src/pi-guide.md
narimiran: My (maybe wrong) opinion: While the OS is now officially called "Raspberry Pi OS", most of people have at some point at least heard about "Raspbian".
By mentioning both names, you let those people know that it is the same thing. (And maybe it helps if people search for 'raspbian'?)

I'm fine either way, I can remove that remark.

tersec: Alright, I agree, seems reasonable to keep it.

docs/the_nimbus_book/src/el-light-client.md
narimiran: > Is it worth addressing usage of https://nimbus.guide/trusted-node-sync.html in the context of sync time?

This is a page about the light client, so I'm not sure if we should go into details about "offtopic" stuff.

Would just changing 'Days' to 'Hours' be enough?
But we mention 'several days' in other places too (e.g. https://nimbus.guide/trusted-node-sync.html), so this needs to be unified.

docs/the_nimbus_book/src/eth1.md
narimiran: Is this something that should maybe go in https://nimbus.guide/hardware.html, where we mention minimal/recommended requirements?

tersec: Yeah, makes sense.

docs/the_nimbus_book/src/el-light-client.md
tersec: Well, to the extent any of this is off-topic for the light client discussion, having a summary table making that particular comparison already is.

But sure, changing "days" to "hours" consistently for this would be fine to start with.

tersec: https://github.com/status-im/nimbus-eth2/pull/4938/commits/cd8253a9269ce1a904a94c9b3bd6f408b1252fe6

docs/the_nimbus_book/src/eth1.md
tersec: Addressing this (or not) in https://nimbus.guide/hardware.html (or not) can be another PR; no need to get everything in this one.

narimiran: > Addressing this (or not) in https://nimbus.guide/hardware.html (or not) can be another PR; no need to get everything in this one.

Yes, I was/am planning to do that in a separate PR, as I wanted to keep this one concentrated only on the 'getting started' part of the guide.

----

Thanks for merging, this now unblocks my future PRs, built on top of this.

narimiran: > Addressing this (or not) in https://nimbus.guide/hardware.html (or not) can be another PR; no need to get everything in this one.

Now addressed in https://github.com/status-im/nimbus-eth2/pull/5059

PR #4943, opened 2023-05-12
fix SSZ response for `produceBlindedBlock`

etan-status:
In `produceBlindedBlock`, we sent the `ForkedBlindedBeaconBlock` when requested to reply in SSZ format. However, expected result is just the inner `ForkyBlindedBeaconBlock` together with `eth-consensus-version`.

Note: We do not use SSZ format in our VC for this endpoint at this time, which explains why we haven't noticed earlier.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -1214,7 +1214,7 @@ proc writeValue*[
   writer.endRecord()
 
 proc writeValue*[
-    BlockType: ForkedBeaconBlock|ForkedBlindedBeaconBlock](
+    BlockType: ForkedBeaconBlock](
     writer: var JsonWriter[RestJson],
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -1214,7 +1214,7 @@ proc writeValue*[
   writer.endRecord()
 
 proc writeValue*[
-    BlockType: ForkedBeaconBlock|ForkedBlindedBeaconBlock](
+    BlockType: ForkedBeaconBlock](
```
etan-status, line 1217: Verified that this is not needed anymore, by compiling with this overload

```nim
proc writeValue*[
    BlockType: ForkedBlindedBeaconBlock](
    writer: var JsonWriter[RestJson],
    value: BlockType) {.raises: [IOError, Defect].} =

  static: raiseAssert "error"
```

PR #4947, opened 2023-05-12
ignore attestations that miss shuffling cache

etan-status:
Replace `getShufflingRef` with `findShufflingRef`.


etan-status:
This doesn't work, local testnets don't finalize. Let's stick with the proper fix instead: https://github.com/status-im/nimbus-eth2/pull/4911

arnetheduck:
this shouldn't lead to finality problems though - if it does, something is wrong with the testnet :/

PR #4948, opened 2023-05-12
align sync horizons between VC and BN

etan-status:
Sometimes, more than 3 blocks are missed in a row, and VC falls back to `Execution client not in sync (beacon node optimistically synced). Aligning the VC `SYNC_TOLERANCE` with the BN `defaultSyncHorizon` (50) should continue with validator duties at that time.

### DIFF
--- a/beacon_chain/consensus_object_pools/consensus_manager.nim
+++ b/beacon_chain/consensus_object_pools/consensus_manager.nim
@@ -252,8 +252,6 @@ func isSynced(dag: ChainDAGRef, wallSlot: Slot): bool =
   # the defaultSyncHorizon, it will start triggering in time so that potential
   # discrepancies between the head here, and the head the DAG has (which might
   # not yet be updated) won't be visible.
-  const defaultSyncHorizon = 50
-
   if dag.head.slot + defaultSyncHorizon < wallSlot:
     false
   else:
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -252,8 +252,6 @@ func isSynced(dag: ChainDAGRef, wallSlot: Slot): bool =
   # the defaultSyncHorizon, it will start triggering in time so that potential
   # discrepancies between the head here, and the head the DAG has (which might
   # not yet be updated) won't be visible.
-  const defaultSyncHorizon = 50
```
tersec, line 255: This exists only for the purpose of `checkNextProposer`, so it's an intentionally nonstrict heuristic

tersec: https://github.com/status-im/nimbus-eth2/blob/40253a76ddcfd2aa29ef662bb8ee61306093fd1c/beacon_chain/validators/validator_duties.nim#L142-L170
is the more authoritative version

tersec: https://github.com/status-im/nimbus-eth2/blob/40253a76ddcfd2aa29ef662bb8ee61306093fd1c/beacon_chain/conf.nim#L531-L536

is configurable, so the VC to match would have to match said configuration, in general.

etan-status: oh, right. that should be cleaned then. no need to rush this PR

PR #4952, opened 2023-05-14
avoid polluting Jenkins CI with geth binaries

etan-status:
When `scripts/geth_binaries.sh` was updated, as a side effect of dumping downloaded Geth tarballs / zips into the repo root directory, artifacts for Jenkins jobs started preserving those downloads with every job, repeatedly leading to full disks. Exclude those downloads from Jenkins.


tersec (approved):

etan-status:
Verified geth binaries no longer being cached on Jenkins artifacts.

jakubgs (commented):
Thanks. I'm surprised we didn't notice this earlier.

PR #4954, opened 2023-05-14
fix `getSyncCommitteeDuties` for `ALTAIR_FORK_EPOCH`.period

etan-status:
When fetching historical `getSyncCommitteeDuties` for the very first sync committee period, the case must be handled where Altair may not have been scheduled on a sync committee period boundary.

### DIFF
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -164,6 +164,7 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
     return RestApiResponse.jsonResponseWRoot(
       duties, epochRef.proposer_dependent_root, optimistic)
 
+  # https://ethereum.github.io/beacon-APIs/#/Validator/getSyncCommitteeDuties
   router.api(MethodPost, "/eth/v1/validator/duties/sync/{epoch}") do (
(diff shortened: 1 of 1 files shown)


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -164,6 +164,7 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
     return RestApiResponse.jsonResponseWRoot(
       duties, epochRef.proposer_dependent_root, optimistic)
 
+  # https://ethereum.github.io/beacon-APIs/#/Validator/getSyncCommitteeDuties
```
tersec, line 167: Potentially useful to also link to https://github.com/ethereum/beacon-APIs/blob/v2.4.0/apis/validator/duties/sync.yaml which is a versioned URL.

tersec (approved):

beacon_chain/rpc/rest_validator_api.nim
etan-status: The proper link would be https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.4.0#/Validator/getSyncCommitteeDuties  — but none of the other URLs in these files have the version, so I just fixed the "missing link" problem here, while leaving the other issue of head vs locked version for later.

PR #4955, opened 2023-05-14
use dependent root as `execution_optimistic` basis for duties

etan-status:
The validator beacon APIs `getAttesterDuties`, `getProposerDuties`, and `getSyncCommitteeDuties`, have reported the `execution_optimistic` state for the current head block. This can lead to a race if duties are requested around the slot start, if a new head block is currently being processed by the EL, during which the BN head may be briefly optimistic.

`execution_optimistic` is documented in beacon APIs as:

> True if the response references an unverified execution payload.
> Optimistic information may be invalidated at a later time.
> If the field is not present, assume the False value.

As the duty endpoints reference the shuffling dependent root instead of the currently selected head block, `execution_optimistic` is now fetched based on that shuffling dependent block root. As this dependent block is in the past it doesn't usually become optimistic when adding new blocks.

Note that the endpoints requested 4/8 seconds into the slot that perform the actual duties instead of just querying for duty schedule, still report `execution_optimistic` based on the BN head block.

### DIFF
--- a/beacon_chain/rpc/rest_utils.nim
+++ b/beacon_chain/rpc/rest_utils.nim
@@ -272,6 +272,17 @@ func keysToIndices*(cacheTable: var Table[ValidatorPubKey, ValidatorIndex],
         indices[listIndex[]] = some(ValidatorIndex(validatorIndex))
   indices
 
+proc getShufflingOptimistic*(node: BeaconNode,
+                             dependentSlot: Slot,
+                             dependentRoot: Eth2Digest): Option[bool] =
+  if node.currentSlot().epoch() >= node.dag.cfg.BELLATRIX_FORK_EPOCH:
+    if dependentSlot <= node.dag.finalizedHead.slot:
+      some[bool](false)
+    else:
+      some[bool](node.dag.is_optimistic(dependentRoot))
+  else:
+    none[bool]()
+
 proc getStateOptimistic*(node: BeaconNode,
                          state: ForkedHashedBeaconState): Option[bool] =
   if node.currentSlot().epoch() >= node.dag.cfg.BELLATRIX_FORK_EPOCH:
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -65,11 +65,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
           return RestApiResponse.jsonError(Http400, InvalidEpochValueError,
                                         "Cannot request duties past next epoch")
         res
-    let (qhead, qoptimistic) =
+    let (qhead, _) =
       block:
         let res = node.getSyncedHead(qepoch)
         if res.isErr():
-          return RestApiResponse.jsonError(Http503, BeaconNodeInSyncError)
+          return RestApiResponse.jsonError(Http503, BeaconNodeInSyncError,
+                                           $res.error())
         res.get()
     let shufflingRef = node.dag.getShufflingRef(qhead, qepoch, true).valueOr:
       return RestApiResponse.jsonError(Http400, PrunedStateError)
@@ -102,11 +103,9 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
                   )
         res
 
-    let optimistic =
-      if node.currentSlot().epoch() >= node.dag.cfg.BELLATRIX_FORK_EPOCH:
-        some(qoptimistic)
-      else:
-        none[bool]()
+    let optimistic = node.getShufflingOptimistic(
+      qepoch.shufflingDependentSlot,
+      shufflingRef.attester_dependent_root)
 
     return RestApiResponse.jsonResponseWRoot(
       duties, shufflingRef.attester_dependent_root, optimistic)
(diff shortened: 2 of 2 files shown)


etan-status:
Note: There will be a followup cleanup that simplifies the way how `is_optimistic` works.

New usage code:
https://github.com/status-im/nimbus-eth2/pull/4956 

@tersec will ensure that `executionValid` is correctly populated for all `BlockRef`.

Note: `executionValid` flag in `BlockRef` does not consume any additional memory, as it can fit into the C struct padding before `parent`.

arnetheduck:
> if a new head block is currently being processed by the EL, during which the BN head may be briefly optimistic.

why is it optimistic here?  this seems like an issue: we should not have updated the head until the el is done - until then, we should keep using the old head in _all_ interactions, rest or internal..

tersec:
> > if a new head block is currently being processed by the EL, during which the BN head may be briefly optimistic.
> 
> why is it optimistic here? this seems like an issue: we should not have updated the head until the el is done - until then, we should keep using the old head in _all_ interactions, rest or internal..

There's no well-defined time period one can guarantee "until the el is done". `--no-el` or optimistic syncing demonstrate this.

The specific gap in a typical slot which might occur between `newPayload` and `forkchoiceUpdated` is only one specific special case of this.

arnetheduck:
> There's no well-defined time period one can guarantee "until the el is done". --no-el or optimistic syncing demonstrate this.

well, sort of: we can define that period as being either 32 slots (based on max attestation validity period) or the dependent root boundary for shuffling (proposer or attestation would have the be decided) - the point remains that it makes sense to delay the edge from non-optimistic to optimistic because otherwise we will systemically be losing attestation profits whenever there's an EL hiccup.

tersec:
> > There's no well-defined time period one can guarantee "until the el is done". --no-el or optimistic syncing demonstrate this.
> 
> well, sort of: we can define that period as being either 32 slots (based on max attestation validity period) or the dependent root boundary for shuffling (proposer or attestation would have the be decided) - the point remains that it makes sense to delay the edge from non-optimistic to optimistic because otherwise we will systemically be losing attestation profits whenever there's an EL hiccup.

`ATTESTATION_PROPAGATION_SLOT_RANGE == 32` is
> The maximum number of slots during which an attestation can be propagated.

It's not how old a block it can point to. That's still an arbitrary limit.

More to the point, optimistic sync requires a head which is updated, including running fork choice on it, continuously regardless of EL input, not just placing some 32-slot, or dependent root-based boundary on it, because the new blocks building the chain apply to some such head, not some random block 32  slots ago or a dependent root.

And it doesn't suffice to just remember the block 32 slots ago or similar -- attestations might have affected such a "valid" head fork choice too.

The main engine API notion here which addresses this is the [latest valid hash](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#payloadstatusv1):
> the hash of the most recent valid block in the branch defined by payload and its ancestors

which avoids these issues, because it only requires one tracking one fork choice structure, and ensures there's an ancestor/descent connection between the LVH and the (potentially optimistic) head.

During optimistic syncing paired with an EL not already synced, there still typically will be no non-optimistic head at all, because ELs won't sync at all until they can begin snap syncing (Erigon aside, which has its own quirks/tradeoffs).

Beyond that, inferring beacon API intent, the existence of the optimistic flags on various calls suggest that it's meant for these to be plausibly optimistic at times.

arnetheduck:
> That's still an arbitrary limit.

This is the limit at which attestations lose their reward completely - therefore, it makes sense to use it as cutoff in places where attestation performance is in question. This also makes it not arbitrary.

>  that it's meant for these to be plausibly optimistic at times.

No, it merely means that nobody thought about the attestation performance consequences when designing an API which gives too little information - thus, in order to accommodate more important goals (such as maximizing rewards), we need to make a tradeoff.

I'm not sure what the resistance here is about - should we not attempt to attest when clearly there is a reward for doing so _and_ a way to do it, because an API along the way is incomplete?

etan-status:
Regardless of the head being optimistic or valid, we currently report `execution_optimistic` based on an incorrect block. It should be reported based on `dependent_root` for the duty schedule endpoints, which is what's being addressed here.

Yes, there may be other aspects about determining whether the head should be optimistic to begin with in some scenarios, but they are out of scope here. Furthermore, the VC may also just ignore `execution_optimistic` in the duties endpoints and also be fine. But also not the point here, this is just a fix to return `execution_optimistic` for the correct block.

zah:
I agree with @etan-status. This fix seems correct and indisputable. 
I also don't see any direct objections from @tersec regarding merging it, but is this the case?

beacon_chain/rpc/rest_utils.nim
```diff
@@ -272,6 +272,17 @@ func keysToIndices*(cacheTable: var Table[ValidatorPubKey, ValidatorIndex],
         indices[listIndex[]] = some(ValidatorIndex(validatorIndex))
   indices
 
+proc getShufflingOptimistic*(node: BeaconNode,
```
zah, line 275: `isShufflingOptimistic` would have been slightly more clear as a name.

zah (approved):

beacon_chain/rpc/rest_utils.nim
etan-status: there are other functions below this, `getStateOptimistic`, `getBlockOptimistic`, and `getBlockRefOptimistic`. This follows the same convention.

arnetheduck:
lgtm too - ie this fix is fine regardless of what we do to the more general case - the same thing might happen during generation of attestation data for example - this shouldn't block this PR (but it might lead to its rollback once we've solved the more general case).

Also, perhaps more notably: our VC doesn't actually care about the duties optimistic flag - it simply registers the duties and tries anyway: it doesn't matter that the client is optimistic _when fethcing duties_: it only matters when actually performing the duties that ensue.

tersec:
> I agree with @etan-status. This fix seems correct and indisputable. I also don't see any direct objections from @tersec regarding merging it, but is this the case?

Indeed, I'm in favor of this PR.

tersec (approved):

PR #4956, opened 2023-05-14
use `executionValid` bit in `BlockRef`

etan-status:
Update `beacon_node/rpc` usage for the new `executionValid` field.


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1969,6 +1969,23 @@ template is_optimistic*(dag: ChainDAGRef, root: Eth2Digest): bool =
...
 
+template is_optimistic*(dag: ChainDAGRef, bid: BlockId): bool =
+  let blck =
+    if bid.slot <= dag.finalizedHead.slot:
+      dag.finalizedHead.blck
+    else:
+      dag.getBlockRef(bid.root).expect("Non-finalized block is known")
+  not blck.executionValid
+
+template is_optimistic*(
```
tersec: Does the combination of these leave any usages of the pure-`Eth2Digest`-root `is_optimistic`? That one has the least information, so ideal to remove if feasible.

etan-status: Indeed, that's possible now.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1961,13 +1961,22 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+    dag: ChainDAGRef, signedBlock: SomeForkySignedBeaconBlock): bool =
+  let blck =
+    if signedBlock.message.slot <= dag.finalizedHead.slot:
+      dag.finalizedHead.blck
+    else:
+      dag.getBlockRef(signedBlock.root).expect("Non-finalized block is known")
```
tersec: Is the assumption here that this version of `is_optimistic` will only be called on `SomeForkySignedBeaconBlock` which have become part of the `ChainDAG` at some point (and then potentially finalized)? Because in principle, this is a function one might call on an arbitrary `SomeForkySignedBeaconBlock`.

This makes sense for the `BlockId` version, because `BlockId`/`BlockRef` are effectively part of set of assumptions, but doesn't necessarily share those.

etan-status: That's a good point and describes the assumption correctly. To avoid accidents in the future, have moved the conversion to `BlockId` to the caller instead. In case of `rest_utils`, that seems sufficiently isolated as is; conceptually, REST historical beacon queries are already limited to previously added blocks as a whole.

tersec (approved):

PR #4969, opened 2023-05-17
replace `optimisticRoots` table with field in `BlockRef`

tersec:
`optimisticRoots` dates to when `BlockRef` was used for finalized blocks, i.e. all known blocks, and adding fields to it not strictly necessary was less than optimal, so the cost of using `BlockRef` is less now than it was. Meanwhile, Nimbus now supports `--no-el` and some other ways of running without an EL client, in a useful way, makes `optimisticRoots` more expensive than it was at the time, when the design assumptions it made were more reasonable.

On usual 64-bit, and likely 32-bit architectures, this doesn't even increase `sizeof(BlockRef)`, because since then, the `Opt[Eth2Digest]` field got added, which created some padding where a `bool` can live.

Some usage of `is_optimistic` still exists -- it's still okay to ask whether a root is optimistic -- but this PR replaces examples where the root was accessed in a `BlockRef` one already had, only to be passed to a function which then looked up that root in another hash table, when the answer to that query is now in the `BlockRef` to begin with. So `is_optimistic` now just is a lookup in a different hash table (for `BlockRef`s) rather than the `optimisticRoots` hash table.

Overlaps somewhat with https://github.com/status-im/nimbus-eth2/pull/4956 but that depends on this, rather than vice versa.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -67,7 +67,7 @@ func init*(
           phase0.TrustedBeaconBlock | altair.TrustedBeaconBlock): BlockRef =
   # Use same formal parameters for simplicity, but it's impossible for these
   # blocks to be optimistic.
-  BlockRef.init(root, Opt.some ZERO_HASH, true, blck.slot)
+  BlockRef.init(root, Opt.some ZERO_HASH, executionValid = true, blck.slot)
 
 func init*(
     T: type BlockRef, root: Eth2Digest, executionValid: bool,
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -76,7 +76,9 @@ func init*(
           deneb.SomeBeaconBlock | deneb.TrustedBeaconBlock): BlockRef =
   BlockRef.init(
     root, Opt.some Eth2Digest(blck.body.execution_payload.block_hash),
-    executionValid, blck.slot)
+    executionValid =
+      blck.body.execution_payload.block_hash == ZERO_HASH or executionValid,
+    blck.slot)
 
 func parent*(bs: BlockSlot): BlockSlot =
   ## Return a blockslot representing the previous slot, using the parent block
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -35,6 +35,7 @@ type
       ## Root that can be used to retrieve block data from database
 
     executionBlockHash*: Opt[Eth2Digest]
+    executionValid*: bool
 
     parent*: BlockRef ##\
       ## Not nil, except for the finalized head
@@ -54,25 +55,28 @@ template slot*(blck: BlockRef): Slot = blck.bid.slot
 
 func init*(
     T: type BlockRef, root: Eth2Digest,
-    executionBlockHash: Opt[Eth2Digest], slot: Slot): BlockRef =
+    executionBlockHash: Opt[Eth2Digest], executionValid: bool, slot: Slot):
+    BlockRef =
   BlockRef(
     bid: BlockId(root: root, slot: slot),
-    executionBlockHash: executionBlockHash)
+    executionBlockHash: executionBlockHash, executionValid: executionValid)
 
 func init*(
-    T: type BlockRef, root: Eth2Digest,
+    T: type BlockRef, root: Eth2Digest, executionValid: bool,
     blck: phase0.SomeBeaconBlock | altair.SomeBeaconBlock |
           phase0.TrustedBeaconBlock | altair.TrustedBeaconBlock): BlockRef =
-  BlockRef.init(root, Opt.some ZERO_HASH, blck.slot)
+  # Use same formal parameters for simplicity, but it's impossible for these
+  # blocks to be optimistic.
+  BlockRef.init(root, Opt.some ZERO_HASH, true, blck.slot)
 
 func init*(
-    T: type BlockRef, root: Eth2Digest,
+    T: type BlockRef, root: Eth2Digest, executionValid: bool,
     blck: bellatrix.SomeBeaconBlock | bellatrix.TrustedBeaconBlock |
           capella.SomeBeaconBlock | capella.TrustedBeaconBlock |
           deneb.SomeBeaconBlock | deneb.TrustedBeaconBlock): BlockRef =
   BlockRef.init(
     root, Opt.some Eth2Digest(blck.body.execution_payload.block_hash),
-    blck.slot)
+    executionValid, blck.slot)
 
 func parent*(bs: BlockSlot): BlockSlot =
   ## Return a blockslot representing the previous slot, using the parent block
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1234,6 +1234,8 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
 
   dag.initLightClientDataCache()
 
+  dag.markBlockVerified(dag.finalizedHead.blck.root)
+
   dag
 
 template genesis_validators_root*(dag: ChainDAGRef): Eth2Digest =
@@ -1697,8 +1699,7 @@ template is_optimistic*(dag: ChainDAGRef, root: Eth2Digest): bool =
     # Either it doesn't exist at all, or it's finalized
     false
 
-proc markBlockVerified*(
-    dag: ChainDAGRef, quarantine: var Quarantine, root: Eth2Digest) =
+proc markBlockVerified*(dag: ChainDAGRef, root: Eth2Digest) =
   var cur = dag.getBlockRef(root).valueOr:
     return
   logScope: blck = shortLog(cur)
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1234,8 +1234,6 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
 
   dag.initLightClientDataCache()
 
-  dag.markBlockVerified(dag.finalizedHead.blck)
-
   dag
 
 template genesis_validators_root*(dag: ChainDAGRef): Eth2Digest =
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -992,9 +992,13 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   # Load head -> finalized, or all summaries in case the finalized block table
   # hasn't been written yet
   for blck in db.getAncestorSummaries(head.root):
-    # The execution block root gets filled in as needed
-    let newRef =
-      BlockRef.init(blck.root, Opt.none Eth2Digest, blck.summary.slot)
+    # The execution block root gets filled in as needed. Nonfinalized Bellatrix
+    # and later blocks are loaded as optimistic, which gets adjusted that first
+    # `VALID` fcU from an EL plus markBlockVerified. Pre-Bellatrix blocks still
+    # get marked as `VALID` by that `BlockRef.init` overload.
+    let newRef = BlockRef.init(
+      blck.root, Opt.none Eth2Digest, executionValid = false,
+      blck.summary.slot)
     if headRef == nil:
       headRef = newRef
 
@@ -1229,17 +1233,6 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
     keysDur = Moment.now() - frontfillTick
 
   dag.initLightClientDataCache()
-
-  # If these aren't actually optimistic, the first fcU will resolve that
-  withState(dag.headState):
-    when consensusFork >= ConsensusFork.Bellatrix:
-      template executionPayloadHeader(): auto =
-        forkyState().data.latest_execution_payload_header
-      const emptyExecutionPayloadHeader =
-        default(type(executionPayloadHeader))
-      if executionPayloadHeader != emptyExecutionPayloadHeader:
-        dag.optimisticRoots.incl dag.head.root
-        dag.optimisticRoots.incl dag.finalizedHead.blck.root
 
   dag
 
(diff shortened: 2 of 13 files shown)


etan-status (commented):
should we mark finalizedHead as valid on startup as well?

tersec:
> should we mark finalizedHead as valid on startup as well?

https://github.com/status-im/nimbus-eth2/pull/4969/commits/dac4684347460d9ab5e863f6e874a83de218a988

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1697,26 +1692,30 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
-    dag: ChainDAGRef, quarantine: var Quarantine, root: Eth2Digest) =
-  # Might be called when block was not optimistic to begin with, or had been
-  # but already had been marked verified.
-  if not dag.is_optimistic(root):
-    return
+  let blck = dag.getBlockRef(root)
+  if blck.isSome:
+    not blck.get.executionValid
+  else:
+    # Either it doesn't exist at all, or it's finalized
```
etan-status, line 1697: Matches current behaviour, but ideally this call would only be made on known-to-exist blocks. Out of scope for this PR though.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1697,26 +1692,30 @@ proc pruneBlocksDAG(dag: ChainDAGRef) =
...
+
   debug "markBlockVerified"
 
   while true:
-    if not dag.is_optimistic(cur.bid.root):
+    if cur.executionValid:
       return
 
-    dag.optimisticRoots.excl cur.bid.root
+    cur.executionValid = true
```
etan-status, line 1704: If we choose to mark finalized as valid always, the `cur.parent.isNil` down there can't really happen anymore. But I don't mind the defensive nature of it either (maybe with a assert unreachable on strictVerification mode?). But, see other comment, maybe we shouldn't treat the finalized one special at all, as it is tricky to track correctly.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1230,16 +1234,7 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
-  withState(dag.headState):
-    when consensusFork >= ConsensusFork.Bellatrix:
-      template executionPayloadHeader(): auto =
-        forkyState().data.latest_execution_payload_header
-      const emptyExecutionPayloadHeader =
-        default(type(executionPayloadHeader))
-      if executionPayloadHeader != emptyExecutionPayloadHeader:
-        dag.optimisticRoots.incl dag.head.root
-        dag.optimisticRoots.incl dag.finalizedHead.blck.root
+  dag.markBlockVerified(dag.finalizedHead.blck.root)
```
etan-status: I think this needs to be done each time we reach a new finalized head. Otherwise, in `--no-el` mode, when finality advances, the new finalized head will be `executionValid: false` once more.

Seems a bit finnicky, in retro maybe we shouldn't treat finalizedHead specially, and only treat pre-finalized as valid implicitly.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -54,25 +55,28 @@ template slot*(blck: BlockRef): Slot = blck.bid.slot
...
 
 func init*(
-    T: type BlockRef, root: Eth2Digest,
+    T: type BlockRef, root: Eth2Digest, executionValid: bool,
     blck: phase0.SomeBeaconBlock | altair.SomeBeaconBlock |
           phase0.TrustedBeaconBlock | altair.TrustedBeaconBlock): BlockRef =
-  BlockRef.init(root, Opt.some ZERO_HASH, blck.slot)
+  # Use same formal parameters for simplicity, but it's impossible for these
+  # blocks to be optimistic.
+  BlockRef.init(root, Opt.some ZERO_HASH, true, blck.slot)
```
etan-status:
```suggestion
  BlockRef.init(root, Opt.some ZERO_HASH, executionValid = true, blck.slot)
```

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -54,25 +55,28 @@ template slot*(blck: BlockRef): Slot = blck.bid.slot
 
 func init*(
     T: type BlockRef, root: Eth2Digest,
-    executionBlockHash: Opt[Eth2Digest], slot: Slot): BlockRef =
+    executionBlockHash: Opt[Eth2Digest], executionValid: bool, slot: Slot):
+    BlockRef =
   BlockRef(
     bid: BlockId(root: root, slot: slot),
-    executionBlockHash: executionBlockHash)
+    executionBlockHash: executionBlockHash, executionValid: executionValid)
```
etan-status, line 62: How is propagation of `executionValid: true` achieved (`markBlockVerified`)? 
Example: `newPayload` returns SYNCING for a block, then `VALID` for the next block, how does the original block become marked `VALID`?

Should we always initialize with `executionValid: false`, and have the caller mark it (+ implicitly, ancestors) immediately valid using `markBlockVerified` after linking the `BlockRef` with its parent, if payload is known to be valid (< merge, or engine response). This way, `markBlockVerified` would be the only way that can set `executionValid`

Or, should the `BlockRef` link step implicitly propagate the bit to all parents?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -992,9 +992,13 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
   # Load head -> finalized, or all summaries in case the finalized block table
   # hasn't been written yet
   for blck in db.getAncestorSummaries(head.root):
-    # The execution block root gets filled in as needed
-    let newRef =
-      BlockRef.init(blck.root, Opt.none Eth2Digest, blck.summary.slot)
+    # The execution block root gets filled in as needed. Nonfinalized Bellatrix
+    # and later blocks are loaded as optimistic, which gets adjusted that first
+    # `VALID` fcU from an EL plus markBlockVerified. Pre-Bellatrix blocks still
+    # get marked as `VALID` by that `BlockRef.init` overload.
```
etan-status: Should it be pre-Bellatrix blocks, or pre-Merge blocks?

Pre-Merge blocks, as in, `block_hash == Eth2Digest()`, technically even for Capella/Deneb blocks.

tersec: Both are true.

One (pre-Bellatrix) is statically verifiable by the type system, which is what this gets at:
https://github.com/status-im/nimbus-eth2/blob/884296ffaf9e1e659e07589c804ce0911357304f/beacon_chain/consensus_object_pools/block_dag.nim#L64-L70

There's no such special treatment in
https://github.com/status-im/nimbus-eth2/blob/884296ffaf9e1e659e07589c804ce0911357304f/beacon_chain/consensus_object_pools/block_dag.nim#L72-L79

Either way it's an optimization rather than corrrectness-determining as such, but a useful enough one. This is primarily for head states, which will need the fcU in current networks both EF and Gnosis, testnets and mainnets, regardless.

It probably makes sense to compare with `blck.body.execution_payload` in the latter though and treat default as nonoptimistic, then this comment could be updated too.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: In particular, an unreachable assert is too aggressive here, since the callers were built with this assumption in mind, that they could call `markBlockVerified` as as an always-safe function without any (external) preconditions to check.

It might be worth looking at the callers and adjusting that, but in general, if the result is that they have to have logic checking validity of the call itself, it's not worth it, better to centralize here. In particular, it should not depend on anything as subtle (from the caller's perspective) as whether the finalized block is treated specially. To the extent it does, that's a good argument for at least ensuring this function's callers don't have to care, to avoid that design coupling.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: It's an optimization. It's not proposing or undertaking other validating actions in `--no-el` mode, so it's not critical what the exact state is for the one boundary block: the claim Nimbus is going to make is that finalized blocks aren't optimistic, which is reasonable, but for blocks which haven't actually been returned as nonoptimistic by the EL, the other argument is reasonable too.

Regardless, for the optimistic sync mechanism, it's not that brittle -- when it matters, the point is that an fcU comes in and resets things as needed, so any combination of (a) mark-finalized-BlockRef-nonoptimistic (or not) and (b) when-finality-advances-mark-that-also-optimistic-or-not function correctly, just satisfying different preferences and with different optimization tradeoffs.

Because `BlockRef`s don't exist pre-finalization, in this case there's not much optimization to be gained, because the newPayload/fcU-`markBlockVerified` has at most one additional block to mark verified, if the initialization finalized state's not treated specially, but similarly, its effect regardless is limited, either way.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/4969/commits/1f3d9d0fa87c9eef6b6608cff3eddd7c795f92d0 does this comparison and updates the comment.

beacon_chain/consensus_object_pools/block_dag.nim
tersec: All parents is too many -- that in nonfinalizing scenarios, that becomes O(n^2)ish as it iterates over the same  ancestor path repeatedly

https://github.com/status-im/nimbus-eth2/pull/4969/commits/1f3d9d0fa87c9eef6b6608cff3eddd7c795f92d0 does just enough of this to inductively get all the relevant blocks regardless of whether new `addResolvedHeadBlock` adds a block with `executionValid` or not. If it wasn't already valid, it links everything to the previous valid, no gaps, and if it does get added as valid, it still does so because it ignores the new block's `executionValid` status when doing the check.

So this way it incrementally sets `executionValid` only to blocks it needs.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: Because it was a potential correctness change, it was worth looking at this more closely, and the trouble was that if `markBlockVerified` was called as no-op, that left `newPayload` as not actually setting `executionValid` correctly. It was, in fact, working fine (because the relevant `BlockRef` had been created by then), but the resulting removal of code which either did nothing or was redundant meant all `markBlockVerified` callers operated directly off a `BlockRef`'s `root` field, at which point it was at least as safe to avoid the additional potential `BlockRef` lookup failure mode.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/1f3d9d0fa87c9eef6b6608cff3eddd7c795f92d0

Because it was a potential correctness change, it was worth looking at this more closely, and the trouble was that if `markBlockVerified` was called as no-op, that left `newPayload` as not actually setting `executionValid` correctly. It was, in fact, working fine (because the relevant `BlockRef` had been created by then), but the resulting removal of code which either did nothing or was redundant meant all `markBlockVerified` callers operated directly off a `BlockRef`'s `root` field, at which point it was at least as safe to avoid the additional potential `BlockRef` lookup failure mode.

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: Sure, should we just, not mark finalized as valid then (neither at startup nor on new finalized checkpoint)? Seems like the more straight forward thing to do.

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -184,13 +184,13 @@ proc updateExecutionClientHead(self: ref ConsensusManager,
...
   of PayloadExecutionStatus.valid:
-    self.dag.markBlockVerified(self.quarantine[], newHead.blck.root)
+    self.dag.markBlockVerified(newHead.blck)
   of PayloadExecutionStatus.invalid, PayloadExecutionStatus.invalid_block_hash:
     self.attestationPool[].forkChoice.mark_root_invalid(newHead.blck.root)
     self.quarantine[].addUnviable(newHead.blck.root)
     return Opt.none(void)
   of PayloadExecutionStatus.accepted, PayloadExecutionStatus.syncing:
-    self.dag.optimisticRoots.incl newHead.blck.root
+    newHead.blck.executionValid = false
```
etan-status: Hmm. I know it's a mechanical change, but is it actually correct to reset `executionValid` on syncing/accepted, even if we already received a valid/invalid before? looking at besu, it tends to do these valid > syncing/accepted transitions sometimes

beacon_chain/validators/keystore_management.nim
```diff
@@ -1416,7 +1416,7 @@ func getPerValidatorDefaultFeeRecipient*(
 proc getSuggestedFeeRecipient*(
     host: KeymanagerHost, pubkey: ValidatorPubKey,
     defaultFeeRecipient: Eth1Address):
-    Result[Eth1Address, ValidatorConfigFileStatus] {.deprecated.} =
+    Result[Eth1Address, ValidatorConfigFileStatus] =
```
etan-status, line 1419: 😅

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -557,9 +557,6 @@ proc storeBlock*(
   # This reduces in-flight fcU spam, which both reduces EL load and decreases
   # otherwise somewhat unpredictable CL head movement.
 
-  if payloadValid:
-    dag.markBlockVerified(self.consensusManager.quarantine[], signedBlock.root)
-
```
etan-status, line 562: This is no longer needed, as `executionValid = blockVerified` in `BlockRef.init` already sets it correctly in `dag.addHeadBlock`?

etan-status (approved):

beacon_chain/gossip_processing/block_processor.nim
tersec: Exactly, yes.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: Correctness is ambiguous here. The [section of optimistic sync specs](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#transitioning-from-valid---invalidated-or-invalidated---valid) I usually quote only addresses `VALID`/`INVALID` changes.

These `VALID`/`SYNCING-or-ACCEPTED` changes are arguably more defensible, and as far as I know, not addressed explicitly by any relevant spec. One explanation is that it's the EL reporting that it doesn't have the state anymore from which to verify either way, so, e.g., if it were to be asked to produce a block, which is one purpose of `forkchoiceUpdated`, it'd be unable to efficiently or in a timely way do so.

But that's not the purpose of the particular `forkchoiceUpdated` affected by this change, so regardless, I'd be okay to change this one.

tersec: https://github.com/status-im/nimbus-eth2/pull/4969/commits/bcb0acf3bdab766be7e437b0eaad6860006d915f

tersec:
> Sure, should we just, not mark finalized as valid then (neither at startup nor on new finalized checkpoint)? Seems like the more straight forward thing to do.

https://github.com/status-im/nimbus-eth2/pull/4969/commits/bcb0acf3bdab766be7e437b0eaad6860006d915f

PR #4970, opened 2023-05-17
remove removed Debug/getDebugChainHeads V1

tersec:
https://github.com/ethereum/beacon-APIs/pull/319

### DIFF
--- a/beacon_chain/rpc/rest_debug_api.nim
+++ b/beacon_chain/rpc/rest_debug_api.nim
@@ -61,13 +61,6 @@ proc installDebugApiHandlers*(router: var RestRouter, node: BeaconNode) =
           RestApiResponse.jsonError(Http500, InvalidAcceptError)
     return RestApiResponse.jsonError(Http404, StateNotFoundError)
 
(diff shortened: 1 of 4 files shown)


zah:
Previously, we've discussed that the best possible way to remove deprecated endpoints is to replace their implementation with an informative error message.

tersec:
> Previously, we've discussed that the best possible way to remove deprecated endpoints is to replace their implementation with an informative error message.

Considered that. This is in the debug namespace. While that's been abused for useful things such as `getState`, which probably shouldn't have been in this namespace to begin with, in general, trying to comply with more usual the meaning of that as, actually being a commitment to something "less" than the beacon/validator/etc namespaces (otherwise, what's the point).

That's kind of the deal with these namespaces: easier to add, easier to change, easier to remove.

So don't strongly oppose it in this case, but it's not the default I'd go with for the debug namespace.

zah:
It's very cheap to add the suggested error message and there aren't any downsides as far as I can tell.

tersec:
This was accidentally merged, rather than updated/rebased to current `unstable` -- thus also did revert commit in `unstable`.

PR #4971, opened 2023-05-17
VC: Sync committee duties fixes and Obol selections implementation.

cheatfate:


tersec:
Only CI failure is related to issue noted in https://github.com/status-im/nimbus-eth2/pull/4976 -- it's trying to build with Nim 1.0.x

PR #4986, opened 2023-05-23
refactor state diffs not to require two states in memory

tersec:
It wasn't feasible to integrate before into the rest of Nimbus without substantially increasing memory consumption due to holding a previous state in memory. However, the diff didn't use the vast majority of the previous state, so this uses a small, intermediate object which can be safely buffered.

With 600k validators, this object is still not tiny, but it's not part of the serialized compatibility surface, so can be optimized as warranted by either using a bitmap (8x less memory, less than 100KiB, constant regardless of what kind of validators exist) or optimizing by omitting certain kinds of withdrawal credentials (needs more than 1 bit of memory, because then some sort of either sparse or offset representation has to be used; a `PackedSet` would be reasonable).

Either way, this is simple, reasonably efficient in both time and space, and uses only built-in data structures.

### DIFF
--- a/beacon_chain/statediff.nim
+++ b/beacon_chain/statediff.nim
@@ -35,7 +35,7 @@ func applyValidatorIdentities(
         withdrawal_credentials: item.withdrawal_credentials):
       raiseAssert "cannot readd"
 
-func setValidatorStatuses(
+func setValidatorStatusesNoWithdrawals(
     validators: var HashList[Validator, Limit VALIDATOR_REGISTRY_LIMIT],
     hl: List[ValidatorStatus, Limit VALIDATOR_REGISTRY_LIMIT]) =
   doAssert validators.len == hl.len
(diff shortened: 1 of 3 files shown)


beacon_chain/statediff.nim
```diff
@@ -156,11 +173,27 @@ func diffStates*(state0, state1: capella.BeaconState): BeaconStateDiff =
...
   )
 
+from std/sequtils import mapIt
+
+func getBeaconStateDiffSummary*(state0: capella.BeaconState):
+    BeaconStateDiffPreSnapshot =
+  BeaconStateDiffPreSnapshot(
+    eth1_data_votes_recent:
+      if state0.eth1_data_votes.len > 0:
+        state0.eth1_data_votes[^1 .. ^1]
```
zah, line 188: The logic here is quite non-obvious. It would be useful to add a reference to `replaceOrAddEncodeEth1Votes`, explaining what the plan is.

zah (approved):

beacon_chain/statediff.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5023

PR #4988, opened 2023-05-24
reenable incbin on 64-bit platforms

arnetheduck:
turns out we're still making releases for 32-bit platforms 😲


zah:
The next nightly build will demonstrate whether this works as intended

PR #4989, opened 2023-05-24
Add support for using custom remote signers in local sim

zah:
Other changes:

* Enable the Nimbus remote signer in the minimal simulation
* Move all log files into the `logs` folder of the simulation
* Create PID files for all processes and use them during the clean-up phase instead of the previous more fragile methods for killing the remaining processes.

You can run the local testnet simulation with the custom signer by launching it like this:

```bash
make local-testnet-minimal SIGNER_TYPE=custom
```

Prior to this, you'll need to adapt the [`scripts/signers/custom.sh`](https://github.com/status-im/nimbus-eth2/blob/df71125754fa151fb1f6004d1f3cc4537b5fb4c6/scripts/signers/custom.sh) script to properly launch your signer with the right command-line parameters (the comments in the file provide more detail).


tersec:
When initially introduced, the remote signers were one of the glitchier parts of the local testnet simulation.

The rest looks innocuous enough, including `SIGNER_TYPE=custom`, but would be preferable if the remote signer being default in the minimal testnet were a separate PR so it can be easily reverted if that ends up useful. It took a little while to figure out the exact right set of changes for https://github.com/status-im/nimbus-eth2/pull/4667 to work and it's cleaner if that's built-in to the PR construction.

zah:
@cheatfate have delivered a number of fixes for the HTTP client and the performance looks more stable now, at least locally for me. It would be very easy to disable the remote signers once again if they prove to be problematic - we just need to edit the flags supplied in the Makefile target.

zah:
2 successful runs so far. I've rebased and I'm enabling auto-merge which will mean 3 successful runs in a row.

PR #4993, opened 2023-05-25
Sync: fix backfill blob iteration order

henridf:


beacon_chain/sync/sync_queue.nim
```diff
@@ -538,15 +538,21 @@ proc getRewindPoint*[T](sq: SyncQueue[T], failSlot: Slot,
            safe_slot = safeSlot, fail_slot = failSlot
     safeSlot
 
-iterator blocks*[T](sq: SyncQueue[T],
-                    sr: SyncResult[T]): ref ForkedSignedBeaconBlock =
+iterator blocks[T](sq: SyncQueue[T],
+                    sr: SyncResult[T]): (ref ForkedSignedBeaconBlock, Opt[BlobSidecars]) =
```
tersec:
```suggestion
                   sr: SyncResult[T]): (ref ForkedSignedBeaconBlock, Opt[BlobSidecars]) =
```

beacon_chain/sync/sync_queue.nim
```diff
@@ -538,15 +538,21 @@ proc getRewindPoint*[T](sq: SyncQueue[T], failSlot: Slot,
...
   case sq.kind
   of SyncQueueKind.Forward:
     for i in countup(0, len(sr.data) - 1):
-      yield sr.data[i]
+      if sr.blobs.isSome:
```
tersec: For each of these, it might be feasible to move the `if` into the `yield`, e.g.,
```nim
  yield (
    sr.data[i],
    if sr.blobs.isSome: Opt.some(sr.blobs.get()[i]) else: Opt.none(BlobSidecars))
```

which reduces duplication, and more to the point, perhaps, points to a cleaner refactoring the transformation from `i` to the `yield` argument.

My instinct is that the intermediate refactoring shown above is probably worse than what you wrote, but allowing pulling it out completely might be better (`yield  sr.data[i], foo(i))`)

henridf: done in ffaa9aa08e29521d49052e928d8c03eba2d43fe8

beacon_chain/sync/sync_queue.nim
henridf: done in ffaa9aa08e29521d49052e928d8c03eba2d43fe8

tersec (approved):

PR #4999, opened 2023-05-26
VC: Use SSZ encoding while processing blocks data

cheatfate:
This PR depends on https://github.com/status-im/nimbus-eth2/pull/4994

### DIFF
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -325,6 +325,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
       slot: Slot, randao_reveal: Option[ValidatorSig],
       graffiti: Option[GraffitiBytes],
       skip_randao_verification: Option[string]) -> RestApiResponse:
+    let contentType =
+      block:
+        let res = preferredContentType(jsonMediaType, sszMediaType)
+        if res.isErr():
+          return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
+        res.get()
     let message =
       block:
         let qslot = block:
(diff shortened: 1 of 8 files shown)


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -325,6 +325,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
       slot: Slot, randao_reveal: Option[ValidatorSig],
       graffiti: Option[GraffitiBytes],
       skip_randao_verification: Option[string]) -> RestApiResponse:
+    let contentType =
+      block:
+        let res = preferredContentType(jsonMediaType, sszMediaType)
+        if res.isErr():
+          return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
+        res.get()
```
arnetheduck:
```suggestion
    let contentType = preferredContentType(jsonMediaType, sszMediaType).valueOr:
      return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
```

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -416,7 +422,16 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+    return
+      if contentType == sszMediaType:
+        let headers = [("eth-consensus-version", message.kind.toString())]
+        withBlck(message):
+          RestApiResponse.sszResponse(blck, headers)
+      elif contentType == jsonMediaType:
+        withBlck(message):
+          RestApiResponse.jsonResponseWVersion(blck, message.kind)
+      else:
+        RestApiResponse.jsonError(Http500, InvalidAcceptError)
```
arnetheduck: shouldn't we return json here? this would happen with `*/*` for example, no? don't think it's good to be too strict here, specially for backwards compat etc

cheatfate: `preferredContentType()` returns only `application/json` or `application/octet-stream` and it is able to handle `*/*` too. I think i can add raiseAssert in here.

beacon_chain/rpc/rest_validator_api.nim
cheatfate: Fixed.

beacon_chain/rpc/rest_validator_api.nim
cheatfate: Fixed.

arnetheduck (approved):

PR #5000, opened 2023-05-27
bump chronos

arnetheduck:
depends on https://github.com/status-im/nim-chronos/pull/395


etan-status:
Note that chronos bumps mess with nim-eth tests due to cancellation asserts

arnetheduck:
right, thanks - turns out it's a bug in utp, so eth2 is not affected

etan-status:
https://github.com/status-im/nim-eth/pull/614
tracked here

PR #5002, opened 2023-05-30
also pack attestations where LMD vote is orphaned

etan-status:
When `attestation.data.beacon_block_root` gets orphaned, attestations with a good `attestation.data.target.root` may still be valuable. The LMD GHOST vote is not relevant for attestation rewards.

Switch to use the FFG vote (`attestation.data.target.root`) instead, gossip validation ensures it is an ancestor of `beacon_block_root`.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -537,26 +537,21 @@ func score(
 
 proc check_attestation_compatible*(
     dag: ChainDAGRef,
-    state: ForkyBeaconState,
+    state: ForkyHashedBeaconState,
     attestation: SomeAttestation): Result[void, cstring] =
-  let targetEpoch = attestation.data.target.epoch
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -537,26 +537,21 @@ func score(
...
   let
-    attestedBlck =
-      dag.getBlockRef(attestation.data.beacon_block_root).valueOr:
-        return err("Unknown `beacon_block_root`")
+    targetEpoch = attestation.data.target.epoch
+    compatibleRoot = state.dependent_root(targetEpoch.get_previous_epoch)
 
-    dependentSlot = (targetEpoch - MIN_SEED_LOOKAHEAD).start_slot - 1
+    attestedBlck = dag.getBlockRef(attestation.data.target.root).valueOr:
+      return err("Unknown `target.root`")
```
arnetheduck, line 547: this can also be:
```
? dag.getBlockRef(...).orErr("Unknown...")
```

PR #5009, opened 2023-05-31
use correct exception in `parseCmdArg(enr.Record)`

etan-status:
`parseCmdArg` is expected to raise `ValueError` but for `enr.Record` we currently raise `ConfigurationError`. Change to `ValueError` instead.


etan-status:
`nim-eth` equivalent: https://github.com/status-im/nim-eth/pull/614

PR #5011, opened 2023-05-31
ensure `quit` on config error with `IOError`

etan-status:
When there is an `IOError` while logging a configuration error, we don't actually `quit` the program. Catch `IOError` to always `quit`.

### DIFF
--- a/beacon_chain/nimbus_binary_common.nim
+++ b/beacon_chain/nimbus_binary_common.nim
@@ -206,19 +206,22 @@ template makeBannerAndConfig*(clientId: string, ConfType: type): untyped =
     )
   except CatchableError as err:
     # We need to log to stderr here, because logging hasn't been configured yet
-    stderr.write "Failure while loading the configuration:\n"
-    stderr.write err.msg
-    stderr.write "\n"
+    try:
+      stderr.write "Failure while loading the configuration:\n"
+      stderr.write err.msg
+      stderr.write "\n"
 
-    if err[] of ConfigurationError and
-       err.parent != nil and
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_binary_common.nim
```diff
@@ -206,19 +206,22 @@ template makeBannerAndConfig*(clientId: string, ConfType: type): untyped =
...
+      if err[] of ConfigurationError and
+        err.parent != nil and
+        err.parent[] of TomlFieldReadingError:
+        let fieldName = ((ref TomlFieldReadingError)(err.parent)).field
+        if fieldName in ["web3-url", "bootstrap-node",
+                        "direct-peer", "validator-monitor-pubkey"]:
+          stderr.write "Since the '" & fieldName & "' option is allowed to " &
+                       "have more than one value, please make sure to supply " &
+                       "a properly formatted TOML array\n"
+    except IOError:
```
tersec, line 223: What kind of `IOError`s can appear here, and is it useful to log them as well, in addition to the already-logged `except CatchableError as err`?

etan-status: Well, when `stderr.write` fails, we can't log :-) Chronicles isn't set up at this spot, so can't use that one either.

Example to see `IOError`

test.nim
```nim
try:
  for i in 0 ..< 1000:
    stdout.write "asdf\p"
except IOError:
  stderr.write "ghjk\p"

stderr.write "ijkl\p"
```

Then: `nim c test`, and `./test >&-` to close stdout

tersec (approved):

PR #5013, opened 2023-05-31
handle one of the `ProveField` warnings

etan-status:
When assigning between `ForkyHashedBeaconState`, suppress `ProveField` warning, as `tgt.kind == src.kind` was already checked, but compiler doesn't understand that (as we only `case tgt.kind`).

### DIFF
--- a/beacon_chain/spec/forks.nim
+++ b/beacon_chain/spec/forks.nim
@@ -578,7 +578,7 @@ template withState*(x: ForkedHashedBeaconState, body: untyped): untyped =
     template forkyState: untyped {.inject, used.} = x.phase0Data
     body
 
-template forky*(
+template forky(
     x: ForkedHashedBeaconState, kind: static ConsensusFork): untyped =
   when kind == ConsensusFork.Deneb:
     x.denebData
--- a/beacon_chain/spec/forks.nim
+++ b/beacon_chain/spec/forks.nim
@@ -627,6 +627,7 @@ func assign*(tgt: var ForkedHashedBeaconState, src: ForkedHashedBeaconState) =
     withState(tgt):
       {.push warning[ProveField]: off.}
       assign(forkyState, src.forky(consensusFork))
+      {.pop.}
   else:
     # Ensure case object and discriminator get updated simultaneously, even
     # with nimOldCaseObjects. This is infrequent.
--- a/beacon_chain/spec/forks.nim
+++ b/beacon_chain/spec/forks.nim
@@ -578,6 +578,21 @@ template withState*(x: ForkedHashedBeaconState, body: untyped): untyped =
     template forkyState: untyped {.inject, used.} = x.phase0Data
     body
 
+template forky*(
+    x: ForkedHashedBeaconState, kind: static ConsensusFork): untyped =
+  when kind == ConsensusFork.Deneb:
+    x.denebData
+  elif kind == ConsensusFork.Capella:
+    x.capellaData
+  elif kind == ConsensusFork.Bellatrix:
+    x.bellatrixData
+  elif kind == ConsensusFork.Altair:
+    x.altairData
+  elif kind == ConsensusFork.Phase0:
+    x.phase0Data
+  else:
+    static: raiseAssert "Unreachable"
+
 template withEpochInfo*(x: ForkedEpochInfo, body: untyped): untyped =
   case x.kind
   of EpochInfoFork.Phase0:
@@ -609,17 +624,9 @@ template withEpochInfo*(
 
 func assign*(tgt: var ForkedHashedBeaconState, src: ForkedHashedBeaconState) =
   if tgt.kind == src.kind:
-    case tgt.kind
-    of ConsensusFork.Deneb:
-      assign(tgt.denebData,     src.denebData):
-    of ConsensusFork.Capella:
-      assign(tgt.capellaData,   src.capellaData):
-    of ConsensusFork.Bellatrix:
-      assign(tgt.bellatrixData, src.bellatrixData):
-    of ConsensusFork.Altair:
-      assign(tgt.altairData,    src.altairData):
-    of ConsensusFork.Phase0:
-      assign(tgt.phase0Data,    src.phase0Data):
+    withState(tgt):
+      {.push warning[ProveField]: off.}
+      assign(forkyState, src.forky(consensusFork))
   else:
     # Ensure case object and discriminator get updated simultaneously, even
     # with nimOldCaseObjects. This is infrequent.


beacon_chain/spec/forks.nim
```diff
@@ -578,6 +578,21 @@ template withState*(x: ForkedHashedBeaconState, body: untyped): untyped =
     template forkyState: untyped {.inject, used.} = x.phase0Data
     body
 
+template forky*(
```
tersec: Is it useful to export this?

beacon_chain/spec/forks.nim
```diff
@@ -609,17 +624,9 @@ template withEpochInfo*(
...
-    of ConsensusFork.Capella:
-      assign(tgt.capellaData,   src.capellaData):
-    of ConsensusFork.Bellatrix:
-      assign(tgt.bellatrixData, src.bellatrixData):
-    of ConsensusFork.Altair:
-      assign(tgt.altairData,    src.altairData):
-    of ConsensusFork.Phase0:
-      assign(tgt.phase0Data,    src.phase0Data):
+    withState(tgt):
+      {.push warning[ProveField]: off.}
```
tersec, line 628: Is this `.pop.`ed anywhere?

etan-status: automatically, at end of `block` / scope, no?

beacon_chain/spec/forks.nim
etan-status: possibly yes, the equivalent in `forks_light_client.nim` was useful at times.

tersec: Marking things as `*` means that Nim won't pick them up as unused even if/when they are, and there has been sizable dead code eliminated not detected on that basis, so it's arguably preferable to avoid creating false dependencies (exporting) until a specific use case appears.

etan-status:
```suggestion
template forky(
```

etan-status: 19005e8a532d82abe1cf0337f695862d46bd97cf

beacon_chain/spec/forks.nim
tersec: `{.push.}` and `{.pop.}` aren't lexically scoped in Nim.

```nim
proc f() =
  if true:
    block:
      if false:
        {.push warning[BareExcept]:off.}  # this takes effect anyway
        discard  # despite scope ending here, both block and if
    try:
      discard
    except:  # warning disabled here
      discard
    if false:
      {.pop.}  # scopes in which {.push.} and {.pop.} appear aren't lexically related
               # and if one comments out `{.push.}` one gets
               # Error: {.pop.} without a corresponding {.push.}
    try:
      discard  # warning not disabled here
    except:
      discard

f()
```

tersec: To get this to work in the `nimbus-eth2` environment:
```
diff --git a/config.nims b/config.nims
index 5d74da20..9ede5460 100644
--- a/config.nims
+++ b/config.nims
@@ -189,7 +189,7 @@ switch("warning", "LockLevel:off")

 # Too many right now to read compiler output. Warnings are legitimate, but
 # should be fixed out-of-band of `unstable` branch.
-switch("warning", "BareExcept:off")
+#switch("warning", "BareExcept:off")

 # Too many of these because of Defect compat in 1.2
 switch("hint", "XCannotRaiseY:off")
```
because otherwise that warning is just always disabled, regardless.

etan-status: Interesting. So, how to do it in a template then?! 
Because the `{.pop.}` after the implicit result assignment don't go well.
And `return` also doesn't work inside a template as it's copy pasted into the caller code.
And the `{.push.}` / `{.pop.}` around the template (on the outside) doesn't go well with generics.
Luckily, here I can just put an explicit `{.pop.}`, but would still be good to know what's the recommended approach for disabling on just one line / remainder of scope

beacon_chain/spec/forks.nim
```diff
@@ -609,17 +624,9 @@ template withEpochInfo*(
...
+      assign(forkyState, src.forky(consensusFork))
```
etan-status, line 629:
```suggestion
      assign(forkyState, src.forky(consensusFork))
      {.pop.}
```

beacon_chain/spec/forks.nim
etan-status: In your simple example with a `proc`, the `{.push.}` actually survives even outside the function scope, if the `{.pop.}` is removed. So, definitely not the case that they end after a scope, at least not in general.

tersec: > Interesting. So, how to do it in a template then?! Because the `{.pop.}` after the implicit result assignment don't go well. And `return` also doesn't work inside a template as it's copy pasted into the caller code. And the `{.push.}` / `{.pop.}` around the template (on the outside) doesn't go well with generics. Luckily, here I can just put an explicit `{.pop.}`, but would still be good to know what's the recommended approach for disabling on just one line / remainder of scope

https://nim-lang.org/docs/manual.html#effect-system-side-effects and https://nim-lang.org/docs/manual.html#effect-system-gc-safety-effect respectively document that
```nim
func f() =
  {.cast(noSideEffect).}:
    echo "test"
```
and
```nim
proc setPerThread() =
  {.cast(gcsafe).}:
    deepCopy(perThread, someGlobal)
```
works in some cases, i.e. explicitly setting up a pragma as a block.

So one might be able to usually do
```nim
template f() =
  {.warning[ProveInit]:off.}:
    try:
      discard
    except:
      discard

f()
expr
```

tersec: > In your simple example with a `proc`, the `{.push.}` actually survives even outside the function scope, if the `{.pop.}` is removed. So, definitely not the case that they end after a scope, at least not in general.

Indeed, see also https://github.com/nim-lang/Nim/issues/19603 and https://github.com/nim-lang/Nim/issues/19609

etan-status: Example where it's broken. Here, the `{.warning.}` leaks its scope to the second instance as well

```nim
{.warning[BareExcept]:off.}:
  try:
    echo "X"
  except:  # warning disabled here
    discard

try:
  echo "Y"
except:  # warning disabled here
  discard
```

tersec: https://github.com/nim-lang/Nim/issues/21975

etan-status:
Jenkins CI failure: https://github.com/status-im/nimbus-eth2/pull/5018

PR #5014, opened 2023-05-31
remove web3signer block signing V1 (phase0 only) support

tersec:


tests/test_signing_node.nim
```diff
@@ -425,55 +425,6 @@ suite "Nimbus remote signer/signing test (web3signer)":
     finally:
       await client.closeWait()
 
-  asyncTest "Signing phase0 block":
```
arnetheduck, line 428: this looks like the only block signing test we have - can we turn it into a more recent-version-test instead?

tersec: https://github.com/status-im/nimbus-eth2/blob/5add10b53b5f4c35e5fb2db0d3757f58f1a2b72b/tests/test_signing_node.nim#L750-L923

have tests using the `BlockV2` interface for all supported forks.

arnetheduck (approved):

PR #5015, opened 2023-05-31
allow payload builder client to be function of validator/proposer

tersec:


beacon_chain/validators/validator_duties.nim
```diff
@@ -1399,7 +1400,7 @@ proc registerValidators*(node: BeaconNode, epoch: Epoch) {.async.} =
 
     const HttpOk = 200
 
-    let payloadBuilderClient = node.getPayloadBuilderClient().valueOr:
+    let payloadBuilderClient = node.getPayloadBuilderClient(0).valueOr:
```
tersec, line 1403: This is a dummy/stub because the entire logic of validator registration will need reworking, not computing these one validator index at a time. In the meantime, this is harmless.

PR #5017, opened 2023-05-31
bump `nim-chronicles` to `533022be6afee15800148a30b502b8ca9e327377`

etan-status:
- Fix registry locking when threading disabled
- Fix for styleCheck:usages (v0.10.2)
- Setup for Nimble 1.4.0


etan-status:
This way, the only change to the regular `nim-chronicles` master branch is `Don't log non-printable characters`. The rest is now aligned with the 8 Apr 2022 version that `nbc-1.5.3` which we are using is based on.

PR #5025, opened 2023-06-03
do not use untrusted `genesis.ssz` with `--trusted-block-root`

etan-status:
When using trusted node sync with `--trusted-block-root`, the remote server is only trusted for data availability, not for correctness. As a downloaded genesis state cannot be validated for correctness, require it to be passed via the network metadata `genesis.ssz` file for `--trusted-block-root` mode. Network metadata is considered trusted as it is provided by the user and not by the remote server. Further adds a check for consistent `genesis_time` when using `StateId` based trusted node sync. This is just a sanity check to avoid spreading blatantly incorrect data, similar to existing `genesis_validators_root` checks.


beacon_chain/trusted_node_sync.nim
```diff
@@ -124,17 +124,23 @@ proc doTrustedNodeSync*(
...
         genesisState
       else:
-        notice "Downloading genesis state", restUrl
-        try:
-          awaitWithTimeout(
-              client.getStateV2(StateIdent.init(StateIdentType.Genesis), cfg),
-              largeRequestsTimeout):
-            info "Attempt to download genesis state timed out"
+        case syncTarget.kind
+        of TrustedNodeSyncKind.TrustedBlockRoot:
```
arnetheduck, line 128: doesn't the genesis validator root determine the genesis state to a sufficient degree? if so, comparing the two should be sufficient to check that the genesis state is ok

etan-status: How to obtain the correct genesis validators root without a trusted state?

etan-status: Note that for known networks, a `genesis.ssz` file is provided with the network metadata.

arnetheduck: look at the genesis creation function - afair it's mostly-zeroes except for genesis vals - ie the only dynamic portion of genesis initialization is the validator set which is represented by the genesis_validators_root in the checkpoint state (ie should be possible to substitute in a merkle hash or simply check that the rest of the bytes are zero or whatever they should be  instead of hashing)

etan-status: It's not possible to fully validate it due to eth1 data.

without `genesis.ssz` we don't have `genesis_validators_root` nor `genesis_time` in network metadata.

arnetheduck: ok - well, in that case, we can simply not write genesis - it's not really needed for anything except rebuilding deep history

> we don't have genesis_validators_root nor genesis_time in network metadata.

these are available from the checkpoint state (which eventually gets trusted)

etan-status: `genesis_validators_root` and `genesis_time` are used as basis to validate all the signatures.

etan-status: As in, if we don't have them, we can't validate signatures. and without a `genesis.ssz` file, there is no trust-minimized way to obtain them.

etan-status: > these are available from the checkpoint state (which eventually gets trusted)

We only have a trusted block root, that may be 1 year back. Its state may no longer be hosted.

To advance from there and check sync committee signatures, we need `genesis_validators_root` and `genesis_time`.

etan-status: The proper solution would be to pull off something similar to what they do in execution with "known to be merged", and put `genesis_time` and `genesis_validators_root` directly into the `config.yaml`. That file is already trusted.

The `genesis.ssz` file right next to it also works, but is bigger. If someone has a custom network, they can simply put the file there to mark it trusted. None of this affects EF networks as they provide `genesis.ssz`.

arnetheduck: ooh. that's a good point. hm. bummer. but if we trust a provided block root, and that block points to a state, and that state contains them, there's no sig needed to trust _that_ state?

etan-status: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/light-client.md
this, for list of requirements btw

etan-status: > but if we trust a provided block root, and that block points to a state, and that state contains them, there's no sig needed to trust that state?

Yes, that could work, _if_ the state is still available. We still couldn't use it to validate the real genesis state, because we don't know the historical eth1data and deposit history. But, we could go on without requiring a `genesis.ssz` file in network configuration.

Also, keep in mind that states tend to become quite big after genesis, so downloading an extra state that most likely doesn't work as it is old, just to obtain two parameters that are present in network metadata in sane cases anyway, is probably not worth the complexity. Up to you to decide though. I would be interested in a practical use case though.

arnetheduck: I don't have one, just checking the understanding - ie sounds to me like a reasonable outcome here is simply to document the above in the code so that if ever it becomes a real question, we have the answer ready

etan-status: bf0924ebb60961d8a9524501122e5f72ad66e4ac

arnetheduck: and historical_roots are also of no use?

etan-status: need a way to obtain proof. at this time, no API.

etan-status: Also, the proof changes every slot

etan-status: bf0924ebb60961d8a9524501122e5f72ad66e4ac

etan-status: a39ee4c8a270e71d6c4a0c5f69e8af6e31096836

arnetheduck (approved):

PR #5031, opened 2023-06-05
consensus spec v1.4.0 attestation stability subnets

tersec:


beacon_chain/rpc/rest_config_api.nim
```diff
@@ -253,8 +253,7 @@ proc installConfigApiHandlers*(router: var RestRouter, node: BeaconNode) =
           # https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/phase0/validator.md#constants
           TARGET_AGGREGATORS_PER_COMMITTEE:
             Base10.toString(TARGET_AGGREGATORS_PER_COMMITTEE),
-          RANDOM_SUBNETS_PER_VALIDATOR:
-            Base10.toString(RANDOM_SUBNETS_PER_VALIDATOR),
+          RANDOM_SUBNETS_PER_VALIDATOR: "1",
```
tersec: This is a waypoint to removal of `RANDOM_SUBNETS_PER_VALIDATOR` from the REST output entirely, because it's indeed not in the v1.4.0 version.

But that's a separate PR, because it's the only part of this which would change enforceable and externally visible behavior.

One of the issues with the current setup, and one reason for changing it, is that it's not enforceable because it's a function of things (validator count and per-node local randomness) which isn't externally verifiable.

While currently the new approach isn't enforced either, presumably in the upcoming hardfork or the one after that it will be. Until then, this config endpoint output is worth treating separately while also extricating the `RANDOM_SUBNETS_PER_VALIDATOR` fully from the rest of the codebase.

arnetheduck:
do we want to put this behind a runtime toggle for a few versions, to have a plan b if the network becomes unstable?

tersec:
> do we want to put this behind a runtime toggle for a few versions, to have a plan b if the network becomes unstable?

Not sure if that's the best approach. In the meantime, https://github.com/status-im/nimbus-eth2/pull/5092 contains the parts of this both necessary and independent of such a concern.

arnetheduck:
> Not sure if that's the best approach. 

An escape hatch is needed if this turns out to be a bad idea - reverting to current behavior is the most simple one, but if you have others in mind?

tersec:
> do we want to put this behind a runtime toggle for a few versions, to have a plan b if the network becomes unstable?

https://github.com/status-im/nimbus-eth2/pull/5128

tersec:
Superceded by https://github.com/status-im/nimbus-eth2/pull/5128

PR #5032, opened 2023-06-05
bump chronos

arnetheduck:
* ratelimiter crash fix
* implicit returns now available!


arnetheduck:
crash introduced in #5024

PR #5034, opened 2023-06-06
initial version of mock builder relay

tersec:
Depends on https://github.com/status-im/nimbus-eth2/pull/5021


tersec:
It's harmless enough in-tree, further fixes can go there. It won't work yet because of the noted dependent PR, though.

PR #5040, opened 2023-06-08
bump stew

arnetheduck:
`results` updates, brings better error messages and convenient helpers


arnetheduck:
turns out this hits a pretty serious nim bug: https://github.com/nim-lang/Nim/issues/22049 (!)

PR #5043, opened 2023-06-08
bump `nim-serialization` to `845bed4478c8f0ac893494083798348e9e7ffb58`

etan-status:
- fix nim devel error: 'result' requires explicit initialization
- Use Nim 2.0 in CI
- test both refc and ORC in post-1.6 Nim versions
- fix makeFieldReadersTable crash when orc enabled.
- remove appveyor and travis scripts


etan-status:
Can't bump to a commit after this one, because `nim-ssz-serialization` isn't compatible in `{.raises.}`. @jangko 

Also, while this is necessary for `--mm:orc`, it is not sufficient alone.
See https://github.com/status-im/nimbus-eth2/blob/dev/etan/mm-orcbug/test.nim

etan-status:
This commit in `nim-serialization` seems to make the macros in Yaml misbehave:
https://github.com/status-im/nim-serialization/commit/bc46b4c1c1730cc25bf5fb5f3d64bd708a6ad89e

```sh
nim c tests/consensus_spec/test_fixture_light_client_sync.nim
```

```nim
      type TestMetaYaml {.sparse.} = object
        genesis_validators_root: string
        trusted_block_root: string
        bootstrap_fork_digest: Option[string]
        store_fork_digest: Option[string]
      let
        meta = block:
          var s = openFileStream(path/"meta.yaml")
          defer: close(s)
          var res: TestMetaYaml
          yaml.load(s, res)
          res
```

etan-status:
@jangko Turns out that `FieldReader` type somehow leaks into completely other contexts.

For example, in NimYAML when I add `echo $repr(getTypeImpl(o))` to the start of `checkMissing`, it gets dumped as 

```
object
  genesis_validators_root: string
  trusted_block_root: string
  bootstrap_fork_digest: FieldReader`gensym689
  store_fork_digest: FieldReader`gensym689
```

In `nim-serialization`, if I change

```nim
template readFieldIMPL[Reader](field: type FieldTag,
                               reader: var Reader): auto =
  mixin readValue
  type FieldType = GetFieldType(field)
  {.gcsafe.}: # needed by Nim-1.6
    # TODO: The `FieldType` coercion below is required to deal
    # with a nim bug caused by the distinct `ssz.List` type.
    # It seems to break the generics cache mechanism, which
    # leads to an incorrect return type being reported from
    # the `readFieldIMPL` function.

    # additional notes: putting the FieldType coercion in
    # `makeFieldReadersTable` will cause problems when orc enabled
    # hence, move it here
    FieldType reader.readValue(FieldType)
```

to

```nim
template readFieldIMPL[Reader](field: type FieldTag,
                               reader: var Reader): auto =
  mixin readValue
  {.gcsafe.}: # needed by Nim-1.6
    # TODO: The `GetFieldType(field)` coercion below is required to deal
    # with a nim bug caused by the distinct `ssz.List` type.
    # It seems to break the generics cache mechanism, which
    # leads to an incorrect return type being reported from
    # the `readFieldIMPL` function.

    # additional notes: putting the GetFieldType(field) coercion in
    # `makeFieldReadersTable` will cause problems when orc enabled
    # hence, move it here
    GetFieldType(field) reader.readValue(GetFieldType(field))
```

fixes the problem. Renaming `FieldType` to something else confirms that the `NimYAML` macros actually somehow get confused and use this type instead of the expected `Option[string]`.

etan-status:
Let's do a partial bump for now with the stuff that works. 

The commit that breaks YAML: https://github.com/status-im/nim-serialization/commit/bc46b4c1c1730cc25bf5fb5f3d64bd708a6ad89e

The commit that breaks SSZ: https://github.com/status-im/nim-serialization/commit/9f56a0738c616061382928b9f60e1c5721622f51

We can do a new bump once the problems are addressed.

Note, the current version is also not good, so we need to eventually do a bump.
Current version has memory corruption in ORC, see https://github.com/nim-lang/Nim/issues/22058

PR #5049, opened 2023-06-09
`ProveField` cleanups in `forks`

etan-status:
Some more cleanup for `ProveField` warnings in `forks` module. Note that `ProveField` is disabled by default in makefile, but sometimes these pop up when doing a regular `nim c`, and cleaning these may allow enabling the warning in some future.


etan-status:
```nim
import beacon_chain/spec/forks

var state: ForkedHashedBeaconState

for index, validator in getStateField(state, validators):
  echo "here"
  echo $index
```

no memcpy

```c
T2_ = (&state__test_4.denebData.data.validators);
```

tersec (approved):

PR #5061, opened 2023-06-12
avoid implicit `CaseTransition` in keystore Json parsing

etan-status:
Override default Json parsing for keystore case objects to avoid `CaseTransition` logic to be emitted. When parsing the discriminator, reinitialize the entire object instead of implicitly changing it, to avoid UB and also possible oversights when extending the object. See https://github.com/status-im/nim-serialization/pull/59

### DIFF
--- a/beacon_chain/spec/keystore.nim
+++ b/beacon_chain/spec/keystore.nim
@@ -556,8 +556,94 @@ proc readValue*[T: SimpleHexEncodedTypes](r: var JsonReader, value: var T) {.
   if len(seq[byte](value)) == 0:
     r.raiseUnexpectedValue("Valid hex string expected")
 
-proc readValue*(r: var JsonReader, value: var Kdf)
-               {.raises: [SerializationError, IOError, Defect].} =
+template readValueImpl(r: var JsonReader, value: var Checksum) =
+  var
+    functionSpecified = false
+    paramsSpecified = false
+    messageSpecified = false
+
+  for fieldName in readObjectFields(r):
+    case fieldName
+    of "function":
+      value = Checksum(function: r.readValue(ChecksumFunctionKind))
+      functionSpecified = true
+
+    of "params":
+      if functionSpecified:
+        case value.function
+        of sha256Checksum:
+          r.readValue(value.params)
+      else:
+        r.raiseUnexpectedValue(
+          "The 'params' field must be specified after the 'function' field")
+      paramsSpecified = true
+
+    of "message":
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/keystore.nim
```diff
@@ -556,16 +556,102 @@ proc readValue*[T: SimpleHexEncodedTypes](r: var JsonReader, value: var T) {.
...
+      functionSpecified = true
+
+    of "params":
+      if functionSpecified:
+        case value.function
+        of sha256Checksum:
+          r.readValue(value.params)
+      else:
+        r.raiseUnexpectedValue(
+          "The 'params' field must be specified after the 'function' field")
```
tersec, line 578: Is this constraint consistent with the current spec? Typically JSON doesn't have these ordering requirements.

beacon_chain/spec/keystore.nim
```diff
@@ -556,16 +556,102 @@ proc readValue*[T: SimpleHexEncodedTypes](r: var JsonReader, value: var T) {.
...
+      paramsSpecified = true
+
+    of "message":
+      if functionSpecified:
+        case value.function
+        of sha256Checksum:
+          r.readValue(value.message)
+      else:
+        r.raiseUnexpectedValue(
+          "The 'message' field must be specified after the 'function' field")
```
tersec, line 588: Similarly, this ordering requirement is unusual in JSON parsing

beacon_chain/spec/keystore.nim
etan-status: The same requirement was already present for `Kdf` type. I extended the logic for the other two types as well.

Right now, it only works by accident when the `params` would actually appear early, because there is only a different `function`. As soon as a second `function` is introduced, the existing logic results in crash due to generic object parser not being case object aware. Probably better to find out early, to push clients into ordering the fields in a sane way, similar to how was done for `Kdf` already.

beacon_chain/spec/keystore.nim
etan-status: I doubt that the spec enforces it, but there are not that many tools that use keystore API, and they can be adjusted to send the fields in the order that works, while still being valid JSON.

etan-status: Argument being, if it was fine for `Kdf`, it seems logical/consistant to me that it should also be fine for the other two case objects in the same api.

tersec (approved):

PR #5070, opened 2023-06-13
ensure that submodule bumps are reachable

etan-status:
It occurs sometimes that a submodule is bumped to a PR commit instead of the corresponding canonical branch (as registered in `.gitmodules`). Because we typically use `squash`, that PR commit can subsequently become unreachable, randomly breaking the build of `nimbus-eth2`. Prevent these accidents by only allowing submodule bumps to commits on the branch registered in `.gitmodules`. On private branches, simply update `.gitmodules` to match the personal dev branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,42 +82,6 @@ jobs:
             exit 2
           fi
 
-      - name: Check submodules (Linux)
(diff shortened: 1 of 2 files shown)


etan-status:
Followup from https://github.com/status-im/nimbus-eth2/pull/5068

tersec (approved):

PR #5059, opened 2023-06-12
various small docs improvements

narimiran:

PR #5076, opened 2023-06-14
Make the validating guide self-contained

narimiran:
This PR is split in two commits for easier reviewing — the second commit is just changing the links, the first commit is the major one.

----

Until now, the validator guide was basically just a page with three links to other guides, not very useful (nor pretty) on its own.
Now, those three separate how-to guides are put under one umbrella, and the validator guide is a regular guide, comparable to the other getting-started guides we have.

Here's a before-after comparison screenshot:


docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -10,16 +10,211 @@ This is a simple, safe and efficient way to get started.
...
+Once created, the keys are used to create a deposit transaction on the Ethereum execution chain.
+Follow the instructions [here](https://launchpad.ethereum.org/en/upload-deposit-data) to upload the deposit data.
+
+!!! warning
+    If you are making a mainnet deposit make sure you verify that the deposit contract you are interacting with is the correct one.
+
+    You should verify that the address is indeed: [0x00000000219ab540356cBB839Cbe05303d7705Fa](https://etherscan.io/address/0x00000000219ab540356cBB839Cbe05303d7705Fa)
+
+!!! info
+    Once you send off your transaction(s), your validator will be put in a queue based on deposit time.
```
tersec: Not sure if/how it's worth elaborating on this, because what you write here is correct, but it's slightly misleading in that it's missing a step which for several hours immediately after sending off the transactions might confuse people.

In particular, there are to main steps here,
- waiting [`ETH1_FOLLOW_DISTANCE` ](https://github.com/ethereum/consensus-specs/blob/v1.4.0-alpha.3/specs/phase0/beacon-chain.md#time-parameters-1) blocks, typically about 8 hours, for consensus clients to pick up the deposit transaction, then

- waiting through the pending/activation queue, as described already here.

The first part is effectively constant-time, and does not depend on how many people are trying to join, while the second does. When, as, now, the pending queue is 10k+ validators, the different isn't that significant for overall time-to-active, but it does mean that for those 8 hours, pure consensus layer tooling will show nothing about those deposits (which aren't yet pending transactions).

It might be worth noting that this is expected, and correct, and not a problem.

docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -10,16 +10,211 @@ This is a simple, safe and efficient way to get started.
...
+
+# Unpack the archive
+tar xvf staking_deposit-cli-9ab0b05-linux-amd64.tar.gz --strip-components 2
+```
+
+### 2. Generate keys
+
+!!! tip "Live image"
+    You can increase the security of this process by downloading a [Live linux image](https://ubuntu.com/tutorials/try-ubuntu-before-you-install). To do so, copy `deposit` to a USB stick, boot into the live image, and run the tool from inside the image.
+    Make sure you **don't** enable Wifi and unplug any Ethernet cables when using this process.
```
tersec:
```suggestion
    Make sure you **don't** enable Wi-Fi and unplug any Ethernet cables when using this process.
```

From https://en.wikipedia.org/wiki/Wi-Fi and similar sources, it looks like the most correct, official spelling and capitalization is "Wi-Fi".

docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -10,16 +10,211 @@ This is a simple, safe and efficient way to get started.
...
+wget https://github.com/ethereum/staking-deposit-cli/releases/download/v2.2.0/staking_deposit-cli-9ab0b05-linux-amd64.tar.gz
+
+# Unpack the archive
+tar xvf staking_deposit-cli-9ab0b05-linux-amd64.tar.gz --strip-components 2
+```
+
+### 2. Generate keys
+
+!!! tip "Live image"
+    You can increase the security of this process by downloading a [Live linux image](https://ubuntu.com/tutorials/try-ubuntu-before-you-install). To do so, copy `deposit` to a USB stick, boot into the live image, and run the tool from inside the image.
```
tersec:
```suggestion
    You can increase the security of this process by downloading a [Live Linux image](https://ubuntu.com/tutorials/try-ubuntu-before-you-install). To do so, copy `deposit` to a USB stick, boot into the live image, and run the tool from inside the image.
```

If one wants to avoid two capitalized words in a row, "Linux live image" works too.

docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -10,16 +10,211 @@ This is a simple, safe and efficient way to get started.
...
+To make a deposit, you will need to generate keys then submit a deposit transaction to the execution chain.
+
+!!! tip "Launchpad"
+    The process of setting up a validator is also documented at the Ethereum launchpad site:
+
+    * [Mainnet](https://launchpad.ethereum.org/)
+    * [Goerli/Prater EthStaker Launchpad](https://goerli.launchpad.ethstaker.cc/en/) or [Goerli/Prater EF Launchpad](https://prater.launchpad.ethereum.org/)
+
+!!! tip
+    Use Prater to stress test and future proof your set up against peak mainnet load.
```
tersec:
```suggestion
    Use Prater to stress test and future-proof your setup against peak mainnet load.
```

There's not currently a great answer/approach to this, unfortunately, because https://prater.beaconcha.in/ shows Prater has 462k active validators, while mainnet has 614k validators with a pending queue of 93k validators, so it'll reach 700k within a month or two.

So the spirit of this is correct, and in some ways, Prater is more demanding -- it's less stable, there are more reorgs, etc -- but also, its base load is significantly less than mainnet and that gap will continue to grow.

I think the actionable thing here is along the lines of rephrasing to keep the claims more verifiably/unambiguously true, without making it 3x longer to explain mostly non-salient background detail.

docs/the_nimbus_book/src/run-a-validator.md
```diff
@@ -10,16 +10,211 @@ This is a simple, safe and efficient way to get started.
     While not needed, advanced users may want to use a [separate validator client](./validator-client.md) instead.
 
 
-## Steps
 
-1. [Make a deposit](./deposit.md) for your validator.
+## Overview
 
-2. [Import your validator keys](./keys.md) into Nimbus.
+To start validating, you need to do these three steps, which will be explained in more detail below:
```
tersec:
```suggestion
To start validating, you need to do these three steps, explained in more detail below:
```

docs/the_nimbus_book/src/run-a-validator.md
tersec: Also, a general note (this applies to https://github.com/status-im/nimbus-eth2/pull/5059#discussion_r1230841876 too), https://eips.ethereum.org/EIPS/eip-6110 will change how this works, but it's not in Shaplla or Dencun, so it's not going to be the deployed mainnet or testnet public setup for several months.

docs/the_nimbus_book/src/run-a-validator.md
tersec: Later this year, https://github.com/eth-clients/holesky will launch, and it should be larger again than mainnet, but that doesn't help until September or October.

docs/the_nimbus_book/src/run-a-validator.md
narimiran: I've made the same change as in #5059.

docs/the_nimbus_book/src/run-a-validator.md
narimiran: I rephrased this part, so it doesn't mention any stress testing.

tersec (approved):

PR #5080, opened 2023-06-15
Add Yamux experimental support

Menduist:
Mplex is starting to get slowly kicked out of libp2p, and it's probably for the best

We've had experimental yamux in nim-libp2p for a while, but it had limited testing on actual networks
Adding this flag, so that we can test on a few fleet nodes, and in some future, make it default to true

Also see https://github.com/libp2p/specs/issues/553

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -305,6 +305,11 @@ type
         defaultValue: false
         name: "enr-auto-update" .}: bool
 
+      enableYamux* {.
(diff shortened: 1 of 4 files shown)


tersec:
Maybe mark this hidden, and/or otherwise signal that it's not a supported option?

This seems ideally ephemeral/temporary and to be removed at some point.

Menduist:
Thanks, I wasn't aware that this was possible

arnetheduck (approved):

PR #5081, opened 2023-06-15
broaden `SszError` catches to `SerializationError`

etan-status:
`nim-serialization` is tagged with `{.raises:[SerializationError].}` so it is no longer sufficient to catch `SszError` in some situations. `SszError` inherits from `SerializationError`, so broadening the caught exception types can be done now, to enable bumping `nim-serialization`.

https://github.com/status-im/nimbus-eth2/pull/5043#issuecomment-1584227993

#5061 is also needed to bump `nim-serialization`.


etan-status:
`nim-serialization` bump furthermore needs https://github.com/status-im/nim-serialization/pull/61 for `NimYAML` compatibility.

tersec (approved):

PR #5084, opened 2023-06-15
ensure sync duties for next epoch are registered in time

etan-status:
For attestations, VC queries duties for current and next epoch. For sync messages, VC queries for current and next period (if soon). This means that for sync messages we don't actually have the duties for next epoch in all situations, leading to situation where VC may miss sync duties in the final slot of an epoch when using. As duties remain same within a sync committee period, simply copy them over to next epoch to avoid this situation without adding network latency.

### DIFF
--- a/beacon_chain/validator_client/duties_service.nim
+++ b/beacon_chain/validator_client/duties_service.nim
@@ -306,7 +306,8 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef,
         res
 
     nextEpoch = epoch + 1
-    nextEpochHasSameDuties = not nextEpoch.is_sync_committee_period
+    nextEpochHasSameDuties = 
+      epoch.sync_committee_period == nextEpoch.sync_committee_period
 
   let addOrReplaceItems =
     block:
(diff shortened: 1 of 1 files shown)


beacon_chain/validator_client/duties_service.nim
```diff
@@ -305,7 +305,8 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef,
               validator_sync_committee_index: validatorSyncCommitteeIndex))
         res
 
-    fork = vc.forkAtEpoch(epoch)
+    nextEpoch = epoch + 1
+    nextEpochHasSameDuties = not nextEpoch.is_sync_committee_period
```
arnetheduck:
```suggestion
    nextEpochHasSameDuties = 
      epoch.sync_committee_period == nextEpoch.sync_committee_period
```
feels easier to understand

etan-status: 2 divisions though, instead of 1. still want it?

arnetheduck: they're shifts. I know they can be tough on the CPU, but I *think* we can afford it in 2023.

arnetheduck:
It does feel like the duties should really be keyed by sync committee period + dependent root here here rather than epoch - this would resolve this problem and avoid the extreme strain on the CPU due to the extra division potentially introduced in the future if period ratios become non-constant.

arnetheduck (approved):

PR #5085, opened 2023-06-15
require sync committee supermajority in CI

etan-status:
To better catch problems with sync committee messages in CI, extend local testnet simulation to also verify that each block is signed by a supermajority of the sync committee.

Requires #5083 and #5084


etan-status:
There seems to be another issue, possibly around fork boundaries, as now slot 24 is failing, but not due to the VC anymore.

etan-status:
Requires #5089

PR #5086, opened 2023-06-16
bump `nim-serialization` to `384eb2561ee755446cff512a8e057325848b86a7`

etan-status:
- readFieldIMPL workaround when orc enabled
- remove nimble.lock
- remove nim 1.2 and nim 1.4 from ci
- remove Defect raises from makeFieldReadersTable
- another attempt to reduce compiler warnings
- More fix to exception tracking and generic_suite
- bump version to 0.2.0
- avoid implicit synthesis of `CaseTransition`
- make nimble file consistent with test matrix
- workaround for `NimYAML` issue


etan-status:
Requires #5061

PR #5099, opened 2023-06-19
reduce `nim-eth` dependencies just for RNG

etan-status:
We have several modules that import `nim-eth` for the sole purpose of its `keys.newRng` function. This function is meanwhile a simple wrapper around `nim-bearssl`'s `HmacDrbgContext.new()`, so the import doesn't really serve a use anymore. Replace `keys.newRng` with the direct call to reduce `nim-eth` imports.


arnetheduck (approved):

etan-status:
For `nim-websock`, there's https://github.com/status-im/nim-websock/pull/148

Apparently, those `utils` there don't really make sense. It's the most convoluted wrapper around `generate` that we have.

PR #5101, opened 2023-06-20
VC: Use scoring function to select best attestation data when using multiple BNs.

cheatfate:
With this PR our validator client (VC) will be able to select the best scored attestation data when using configurations with multiple beacon nodes.
To achieve maximum performance you should enable VC's block monitoring feature using `--block-monitor-type=poll` (polling BN's for latest head 3 times per slot) or `--block-monitor-type=event` (using BN's `/eth/v1/events` interface).

When block monitoring is enabled - VC will be able to obtain "perfect" score and continue attesting without waiting for responses from all the beacon nodes. Without this feature VC will wait for all BN responses and selects the best response using scoring algorithm.

"perfect score" - is score when `attestation data slot equal to block's slot` and `target epoch = source epoch + 1`.

### DIFF
--- a/beacon_chain/validator_client/api.nim
+++ b/beacon_chain/validator_client/api.nim
@@ -55,7 +55,7 @@ proc `$`*(s: ApiScore): string =
   var res = Base10.toString(uint64(s.index))
   res.add(": ")
   if s.score.isSome():
-    res.add(formatFloat(s.score.get(), ffDecimal, 4))
+    res.add(shortScore(s.score.get()))
   else:
     res.add("<n/a>")
   res
@@ -321,7 +321,7 @@ template bestSuccess*(
         retRes = ApiResponse[handlerType].err("No online beacon node(s)")
         break mainLoop
       else:
-        let
+        var
           (pendingRequests, pendingNodes) =
             block:
               var requests: seq[FutureBase]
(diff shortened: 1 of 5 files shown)


beacon_chain/validator_client/api.nim
```diff
@@ -5,10 +5,11 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-import chronicles
+import std/strutils
+import chronicles, stew/base10
 import ../spec/eth2_apis/eth2_rest_serialization,
        ../spec/datatypes/[phase0, altair]
-import common, fallback_service
+import common, fallback_service, scoring
```
arnetheduck:
```suggestion
import "."/[common, fallback_service, scoring]
```

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +265,180 @@ template firstSuccessParallel*(
...
-            requests.add(fut)
-            nodes.add(node)
-          (requests, nodes)
-
-      status =
-        try:
           if isNil(timerFut):
-            await allFutures(pendingRequests)
-            ApiOperation.Success
+            await allFut or timerFut
```
arnetheduck: if `isNil`?

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +265,180 @@ template firstSuccessParallel*(
...
-                debug "One of operation requests has been interrupted",
-                      node = beaconNode
+          for index, future in pendingRequests.pairs():
+            let
+              node {.inject.} = pendingNodes[index]
+              apiResponse {.inject.} =
+                if future.finished():
+                  if future.failed():
+                    ApiResponse[responseType].err($future.error.msg)
+                  else:
```
arnetheduck, line 391: future can't be cancelled here?

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +265,180 @@ template firstSuccessParallel*(
...
+    retRes: ApiResponse[handlerType]
+    scores: seq[ApiScore]
+    bestScore: Opt[float64]
+    bestNode: Opt[BeaconNodeServerRef]
+
+  while true:
+    var resultReady = false
+    let onlineNodes =
+      try:
+        if iterations == 0:
```
arnetheduck: what's the idea with multiple iterations?

cheatfate: We have a period of time which is regulated by `timeFut` value, so if we not received any successful responses in one iteration we are trying to request again with same/new set of BNs until `timeFut` will not expire.

beacon_chain/validator_client/api.nim
cheatfate: `future` is part of `pendingRequests` which is awaited using `allFutures` and `or`, both combinator operations did not cancel argument Futures, so `future` in this case will never be cancelled, with one exception - only if it cancels itself and leak CancelledError.

beacon_chain/validator_client/api.nim
cheatfate: Fixed.

beacon_chain/validator_client/api.nim
cheatfate: doAssert added.

beacon_chain/validator_client/scoring.nim
```diff
@@ -0,0 +1,26 @@
...
+                              adata: ProduceAttestationDataResponse): float64 =
+  let
+    slot = vc.rootsSeen.getOrDefault(
+      adata.data.beacon_block_root, FAR_FUTURE_SLOT)
+    score = float64(adata.data.source.epoch) + float64(adata.data.target.epoch)
+
+  if slot == FAR_FUTURE_SLOT:
+    score
+  else:
+    if adata.data.slot + 1 == slot:
```
arnetheduck: there's a trick we can do: if the slot of the attestation matches that of the block, this is the best possible score, therefore we can stop waiting for "the other" bn:s

etan-status:
This needs to merge from `unstable` to take account of the new CI job - as of #5140 we build both the pinned nim version as well as the latest upstream 1.6 version to catch regressions in nim more easily.

`beacon_chain/validator_client/common.nim` has conflict so not touching this PR

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +272,193 @@ template firstSuccessParallel*(
...
-        except CatchableError:
-          # This should not be happened, because allFutures() and race() did not
-          # raise any exceptions.
-          ApiOperation.Failure
+        except CatchableError as exc:
+          # This case could not be happened.
+          error "Unexpected exception while waiting for beacon nodes",
+                err_name = $exc.name, err_msg = $exc.msg
+          var default: seq[BeaconNodeServerRef]
+          default
```
arnetheduck:
```suggestion
          default(seq[BeaconNodeServerRef])
```
I think this should work

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +272,193 @@ template firstSuccessParallel*(
...
-            if not(waitFut.finished()):
-              await waitFut.cancelAndWait()
-              ApiOperation.Timeout
-            else:
-              if not(timerFut.finished()):
-                await timerFut.cancelAndWait()
-              ApiOperation.Success
+            # We get here only, if all the requests are failed. To avoid requests
+            # spam we going to wait for changes in BNs statuses.
+            await vc.waitNodes(timerFut, statuses, roles, true)
```
arnetheduck, line 311: would it make sense to make `waitNodes` return the nodes that were activated according to the filter? this would remove the need for a separate call to `filterNodes`, avoiding some stateful:ness

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +272,193 @@ template firstSuccessParallel*(
...
+            try:
+              raceFut = race(pendingRequests)
+
+              if isNil(timerFut):
+                await raceFut or timerFut
+              else:
+                await allFutures(raceFut)
+
+              for index, future in pendingRequests.pairs():
+                if (future != timerFut) and
```
arnetheduck: is timerFut part of `pendingRequests`? I can't see where it's added

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +272,193 @@ template firstSuccessParallel*(
...
+                    handlerResponse =
+                      try:
+                        bodyHandler
+                      except CancelledError as exc:
+                        raise exc
+                      except CatchableError:
+                        raiseAssert(
+                          "Response handler must not raise exceptions")
+
+                  if apiResponse.isOk() and handlerResponse.isOk():
```
arnetheduck: `apiResponse` is not used below - shouldn't it be enough that the handler is happy?

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +272,193 @@ template firstSuccessParallel*(
...
+                        except CatchableError:
+                          raiseAssert("Score handler must not raise exceptions")
+                    scores.add(ApiScore.init(node, score))
+                    if bestResponse.isNone() or
+                      (score > bestResponse.get().score):
+                      bestResponse = Opt.some(
+                        BestNodeResponse.init(node, handlerResponse, score))
+                      if perfectScore(score):
+                        perfectScoreFound = true
+                        break
```
arnetheduck, line 390:
```suggestion
                        break mainLoop
```

?

beacon_chain/validator_client/api.nim
cheatfate: It is not part of `pendingRequests`. Because `lazyWait` procedure do not support `timerFut` as part of `pendingRequests` without appropriate `pendingNodes` value.

beacon_chain/validator_client/api.nim
cheatfate: There some cleanup should be done in inner loop before exiting to mainLoop.

beacon_chain/validator_client/api.nim
cheatfate: I wish not to touch it right now and in this PR, it requires much more work to change current behavior.

beacon_chain/validator_client/api.nim
arnetheduck: in that case, `future != timerFut` is redundant? it looks confusing

beacon_chain/validator_client/api.nim
arnetheduck: `break innerLoop` then?

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +271,192 @@ template firstSuccessParallel*(
...
+                        doAssert(not(future.cancelled()))
+                        if future.failed():
+                          ApiResponse[responseType].err($future.error.msg)
+                        else:
+                          ApiResponse[responseType].ok(
+                            Future[responseType](future).read())
+                      else:
+                        doAssert(timerFut.finished())
+                        ApiResponse[responseType].err(
+                          "Timeout exceeded while awaiting for the response")
```
arnetheduck:
```suggestion
                    apiResponse {.inject.} = future.toApiResponseOr("Timeout exceeded while awaiting for the response")
```

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +271,192 @@ template firstSuccessParallel*(
...
+                # If timeout is exceeded we need to cancel all the tasks which
+                # are still running.
+                var pendingCancel: seq[Future[void]]
+                for future in pendingRequests.items():
+                  if not(future.finished()):
+                    pendingCancel.add(future.cancelAndWait())
+                await allFutures(pendingCancel)
+                break innerLoop
+              else:
+                if perfectScoreFound:
```
arnetheduck: shouldn't the perfect score check come before `if timerfut.finished`?

beacon_chain/validator_client/api.nim
```diff
@@ -236,128 +271,192 @@ template firstSuccessParallel*(
...
+              if not(isNil(raceFut)) and not(raceFut.finished()):
+                pendingCancel.add(raceFut.cancelAndWait())
+              if not(isNil(timerFut)) and not(timerFut.finished()):
+                pendingCancel.add(timerFut.cancelAndWait())
+              # We should cancel all the requests which are still pending.
+              for future in pendingRequests.items():
+                if not(future.finished()):
+                  pendingCancel.add(future.cancelAndWait())
+              # Awaiting cancellations.
+              await allFutures(pendingCancel)
```
arnetheduck, line 425: when calling cancelAndWait, is it possible that some of these futures are completed instead of cancelled?

cheatfate: Its only possible to get into this situation
So in this case:
 ```<future not finished> -> <future not finished> -> <future finished>``` result `future` will be finished, not cancelled, but it sill requires number of async steps to get into this.
And in this case
```<future not finished> -> <future not finished> -> <future cancelled>``` result `future` will be cancelled, but only if some of the intermediate futures will not decide to swallow cancellation or replace it with an error.

File: beacon_chain/validator_client/scoring.nim [hunk 14]
```diff
@@ -0,0 +1,44 @@
...
+func perfectScore*(score: float64): bool =
+  if score == Inf: true else: false
+
+proc shortScore*(score: float64): string =
+  if score == Inf: "<perfect>" else: formatFloat(score, ffDecimal, 4)
+
+proc getAttestationDataScore*(vc: ValidatorClientRef,
+                              adata: ProduceAttestationDataResponse): float64 =
+  let
+    slot = vc.rootsSeen.getOrDefault(
```
arnetheduck: we can avoid constructing `rootsSeen` by doing `anyIt(vc.blocksSeen[adata.data.slot], it.root == adata.data.beacon_block_root` here

beacon_chain/validator_client/scoring.nim
```diff
@@ -0,0 +1,44 @@
...
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+import std/strutils
+import "."/common
+
+{.push raises: [].}
+
+func perfectScore*(score: float64): bool =
+  if score == Inf: true else: false
```
arnetheduck:
```suggestion
  score == Inf
```

beacon_chain/validator_client/scoring.nim
cheatfate: No we can't, because `adata.data.slot` is usually `currentSlot` while `adata.data.beacon_block_root` could point to block_root which slot is not equal to `adata.data.slot`.

beacon_chain/validator_client/api.nim
cheatfate: It should not be part of this branch, but timerFut check should be first.

beacon_chain/validator_client/scoring.nim
arnetheduck: yes, so my suggested expression captures this logic .. `look in adata.data.slot and see if at that slot, there exists a block that is adata.data.root` - this is only true when the block is from the same slot

beacon_chain/validator_client/api.nim
cheatfate: I dont know how to build such construction, just because all my attempts are failed to compile.

beacon_chain/validator_client/api.nim
cheatfate: there is no innerLoop.

arnetheduck: https://github.com/status-im/nimbus-eth2/pull/5101/files#diff-0efa923f42e20f0ff6915fd060c9eb33ac1d997af5bd003871f22da721886a17R335

it says `block innerLoop` there?

arnetheduck (approved):

beacon_chain/validator_client/scoring.nim
arnetheduck: ah, you're right, the imperfect scoring still needs it

beacon_chain/validator_client/api.nim
cheatfate: Sorry, yep there is, but as i said there should be some cleanup made right after `pendingRequests` loop, so it breaks out of `pendingRequests` loop and after cleanup it breaks to `innerLoop`.

beacon_chain/validator_client/api.nim
joao1963: ok

PR #5109, opened 2023-06-21
VC: Some block monitor fixes

cheatfate:
* Fix `poll` block monitor issue when it deadlocks VC.
* Fix `events` block monitor issue when it doesn't recover connection when connection has been lost.
* Add `node` field to "Block received" log statement.

### DIFF
--- a/beacon_chain/validator_client/block_service.nim
+++ b/beacon_chain/validator_client/block_service.nim
@@ -480,7 +480,7 @@ proc pollForEvents(service: BlockServiceRef, node: BeaconNodeServerRef,
         let blck = EventBeaconBlockObject.decodeString(event.data).valueOr:
           debug "Got invalid block event format", reason = error
           return
-        vc.registerBlock(blck)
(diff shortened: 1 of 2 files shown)


beacon_chain/validator_client/block_service.nim
```diff
@@ -503,15 +503,15 @@ proc runBlockEventMonitor(service: BlockServiceRef,
 
   while true:
     while node.status notin statuses:
-      await vc.waitNodes(nil, statuses, roles, false)
+      await vc.waitNodes(nil, statuses, roles, true)
```
arnetheduck, line 506: shouldn't this be waiting for the specific node to become connected?

cheatfate: `waitNodes` using event which indicates that some of the changes in nodes status happened, so we  wait for node status to be changed, checks status, if status is not in {RestBeaconNodeStatus.Synced} it continues to waiting changes.

arnetheduck (approved):

PR #5113, opened 2023-06-21
Bump stint to v2.0: new array backend

jangko:


arnetheduck:
let's hold off on merging this one into eth2 until we've increased general testing efforts in stint itself - ie fuzzing for example - we don't use a lot of stint, but the parts we use must be solid

mratsim:
I strongly suggest adding the fuzzing technique from libsecp256k1 (and libopus), see https://github.com/mratsim/constantine/issues/53

Nim version: https://github.com/mratsim/constantine/blob/151f284/helpers/prng_unsafe.nim#L230-L244
```Nim
func random_long01Seq(rng: var RngState, a: var openArray[byte]) =
  ## Initialize a bytearray
  ## It is skewed towards producing strings of 1111... and 0000
  ## to trigger edge cases
  # See libsecp256k1: https://github.com/bitcoin-core/secp256k1/blob/dbd41db1/src/testrand_impl.h#L90-L104
  let Bits = a.len * 8
  var bit = 0
  zeroMem(a[0].addr, a.len)
  while bit < Bits :
    var now = 1 + (rng.random_unsafe(1 shl 6) * rng.random_unsafe(1 shl 5) + 16) div 31
    let val = rng.sample_unsafe([0, 1])
    while now > 0 and bit < Bits:
      a[bit shr 3] = a[bit shr 3] or byte(val shl (bit and 7))
      dec now
      inc bit
```
This helps catch all low probability carry issues, full zeros or full ones issues. (even in cosntant-time code that traditional fuzzers have issue with compared to branches)

Then use that RNG for random testing vs GMP.

It allowed me to easily find a division bug in nim-bigints as well: https://github.com/nim-lang/bigints/issues/123#issuecomment-1568822653

jangko:
don't know what's wrong with jenkins, I have no access to read it.

PR #5122, opened 2023-06-24
add `libnimbus_lc.a` C library

etan-status:
Add a new C library for processing light client data based on the Nimbus implementation. This can be used from other, non-Nimbus components.


etan-status:
macOS requires #5132

tersec (approved):

tersec:
Whatever nitpicks one might have about the C implementation, this PR is almost pure addition rather than modification of existing functionality, and presents no new default-running attack surface, so worth merging and adjusting from there.

arnetheduck:
I might have a few but I'll need a few days to review - no opinion on merge order

beacon_chain/libnimbus_lc/libnimbus_lc.h
```diff
@@ -0,0 +1,1034 @@
...
+#define _Nonnull
+#define _Nullable
+#endif
+
+/**
+ * Initializes Nim & Garbage Collector. Must be called before anything else
+ * of the API. Also, all following calls must come from the same thread as from
+ * which this call was done.
+ */
+void NimMain(void);
```
arnetheduck, line 48: See https://github.com/status-im/nim-style-guide/blob/main/interop/async/asynclib.nim#L16 - `NimMain` should not be exposed to users of the library

beacon_chain/libnimbus_lc/libnimbus_lc.h
```diff
@@ -0,0 +1,1034 @@
...
+ *
+ * - The cryptographically secure random number generator must be destroyed
+ *   with `ETHRandomNumberDestroy` once no longer needed, to release memory.
+ *
+ * @return Pointer to an initialized cryptographically secure random number
+ *         generator context - If successful.
+ * @return `NULL` - If an error occurred.
+ */
+ETH_RESULT_USE_CHECK
+ETHRandomNumber *ETHRandomNumberCreate(void);
```
arnetheduck, line 66: RNG feels like something we should hide behind a generic light client context - ie no need for library consumer to know / care

beacon_chain/libnimbus_lc/libnimbus_lc.h
```diff
@@ -0,0 +1,1034 @@
...
+ * @see https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/phase0/beacon-chain.md#beaconstate
+ * @see https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/altair/beacon-chain.md#beaconstate
+ * @see https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/bellatrix/beacon-chain.md#beaconstate
+ * @see https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/capella/beacon-chain.md#beaconstate
+ * @see https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/configs/README.md
+ */
+ETH_RESULT_USE_CHECK
+ETHBeaconState *ETHBeaconStateCreateFromSsz(
+    const ETHConsensusConfig *cfg,
+    const char *consensusVersion,
```
arnetheduck, line 161: we don't need the version (we can deduce it from the header)

beacon_chain/libnimbus_lc/libnimbus_lc.h
```diff
@@ -0,0 +1,1034 @@
...
+/**
+ * Destroys a Merkle root.
+ *
+ * - The Merkle root must no longer be used after destruction.
+ *
+ * @param      root                 Merkle root.
+ *
+ * @see https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/phase0/beacon-chain.md#custom-types
+ */
+void ETHRootDestroy(ETHRoot *root);
```
arnetheduck, line 203: it might be easier to pass these around by value

beacon_chain/libnimbus_lc/libnimbus_lc.nim
```diff
@@ -0,0 +1,1168 @@
...
+  ## Obtains the data gas used of a given execution payload header.
+  ##
+  ## Parameters:
+  ## * `execution` - Execution payload header.
+  ##
+  ## Returns:
+  ## * Data gas used.
+  execution[].data_gas_used.cint
+
+func ETHExecutionPayloadHeaderGetExcessDataGas(
```
arnetheduck, line 1159: isn't `cint` too small for this? ie max 2**31?

arnetheduck:
In terms of C API, I'm not necessarily convinced that we need this great level of detail to access the internal details of the Nimbus implementation (the stores, the RNG, etc etc) - ie the REST API has a an appropriate abstraction level in that it represents the data and the events necessary to build a consumer of the light client API, and it has the benefit of being familiar to the ecosystem and well documented.

As such, when thinking about a light client data consumer, I envision they launch the light client via some `startLightClient(LCContext* context)` function and receive events when things happen (ie `onLightClientHeader` or whatever the events are called.

One can also plausibly think that someone might not want the entire async stack but rather drive the application using well-known data: in this case, one would feed it the messages as they come off the libp2p stack instead but again forego the internal implementation details and rather focus on a high-level context that collects everything needed under an opaque type.


Re tests, why the genesis.ssz and other binary files? these can trivially be generated and thus not bloat the git history (which unfortunately will have to stay due to this being merged already)...

arnetheduck:
```
+ rm -f build/libnimbus_lc.a
+ /home/arnetheduck/status/nimbus-eth2/vendor/nimbus-build-system/scripts/env.sh nim c -d:disable_libbacktrace -d:release --app:staticlib --noMain --nimcache:nimcache/libnimbus_lc_static -o:build/libnimbus_lc.a -d:release --parallelBuild:1 -d:libp2p_agents_metrics -d:KnownLibP2PAgents=nimbus,lighthouse,lodestar,prysm,teku --verbosity:0 --hints:off -d:chronicles_log_level=DEBUG beacon_chain/libnimbus_lc/libnimbus_lc.nim
```

debugging leftovers in `make test`

beacon_chain/libnimbus_lc/libnimbus_lc.h
etan-status: What if someone wants to run multiple light client engines in the same thread? e.g., for a dashboard that tracks both Goerli and Mainnet? Sharing the RNG would be great in that case. Or should it simply be set up as part of a library wide `ETHContextCreate`?

beacon_chain/libnimbus_lc/libnimbus_lc.h
etan-status: That's only possible due to a coincidence and not generally applicable. The header is not necessarily at the same location forever.

beacon_chain/libnimbus_lc/libnimbus_lc.nim
etan-status: Yeah, same for `Slot`, in theory would have to return it as a pair of high/low `cint`, or 4x / 8x (or allocate + matching `Destroy`). 
But for practical ranges, `cint` is good enough. If higher values are possible in practice, needs to be extended.

beacon_chain/libnimbus_lc/libnimbus_lc.h
etan-status: Maybe. In the WASM case, can only pass `cint`/`pointer`, so there would still be the need for memory management in the client.

PR #5128, opened 2023-06-27
opt-out support for v1.4.0 attestation stability subnets

tersec:
If/when this works well enough to become the default, the runtime switch machinery, including the command-line option, might disappear without notice.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -238,6 +238,11 @@ type
       desc: "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)"
       name: "num-threads" .}: int
 
+    useOldStabilitySubnets* {.
+      hidden
+      defaultValue: true
+      name: "temporary-debug-use-old-attestation-stability-subnets" .}: bool
+
     # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.3/src/engine/authentication.md#key-distribution
     jwtSecret* {.
       desc: "A file containing the hex-encoded 256 bit secret key to be used for verifying/generating JWT tokens"
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -241,7 +241,7 @@ type
     useOldStabilitySubnets* {.
       hidden
       defaultValue: true
-      name: "temporary-debug-use-old-attestation-stability-subnets" .}: bool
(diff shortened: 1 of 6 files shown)


beacon_chain/conf.nim
```diff
@@ -238,6 +238,12 @@ type
       desc: "Number of worker threads (\"0\" = use as many threads as there are CPU cores available)"
       name: "num-threads" .}: int
 
+    useOldStabilitySubnets* {.
+      hidden
+      defaultValue: true
+      desc: "deprecated; might be removed at any time, with no warning"
+      name: "temporary-debug-use-old-attestation-stability-subnets" .}: bool
```
```suggestion
      desc: "use pre-1.4.0 stability subnet subscription model (this option will be removed in a future release)"
      name: "debug-use-old-attestation-stability-subnets" .}: bool
```

the option is hidden so users won't see it - for those that look in the source code, it's better to describe it accurately - we don't provide stability guarantees for debug stuff, no need to go over the top with the name either

tersec: The goal is to avoid, if at all possible, adding to
https://github.com/status-im/nimbus-eth2/blob/970f5dfc4e2df53d12ed71fc591e23ecc2bafb60/beacon_chain/nimbus_beacon_node.nim#L1857-L1862
and keeping any kind of stub config option in `conf.nim`, et cetera after all this.

And, backing off from wording about can be removed at any time isn't exactly encouraging in that regard, while adding a description on top of that makes it look even less likely that it could be removed.

https://github.com/status-im/nimbus-eth2/pull/5128/commits/01601726245f4711d4e103a900cc5ab776aef5b1 might be a reasonable approach. Descriptions aren't particularly functional for `hidden` options regardless, so in some sense it was dead code anyway.

arnetheduck:
so, leave out `temporary` from option name, doc that `--debug` options are generally ephemeral and fix PR title (it's an opt-out opition...) and we're gtg

tersec:
> so, leave out `temporary` from option name, doc that `--debug` options are generally ephemeral and fix PR title (it's an opt-out opition...) and we're gtg

https://github.com/status-im/nimbus-eth2/pull/5128/commits/a8f6433ee5c40c1edaa80920578147bdab53cdfe

arnetheduck (approved):

PR #5130, opened 2023-06-27
VC: Runtime config

cheatfate:
With this PR VC will be able to understand any type of response on [/eth/v1/config/spec](https://ethereum.github.io/beacon-APIs/#/Config/getSpec) call without any changes in types and/or source code.
VC will perform runtime check of values and now be able to use all the values from [/eth/v1/config/spec](https://ethereum.github.io/beacon-APIs/#/Config/getSpec) response.

### DIFF
--- a/beacon_chain/spec/eth2_apis/rest_types.nim
+++ b/beacon_chain/spec/eth2_apis/rest_types.nim
@@ -479,39 +479,6 @@ type
     TARGET_AGGREGATORS_PER_SYNC_SUBCOMMITTEE*: uint64
     SYNC_COMMITTEE_SUBNET_COUNT*: uint64
 
-  # The `RestSpec` is a dynamic dictionary that includes version-specific spec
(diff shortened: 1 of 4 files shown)


beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -683,7 +686,7 @@ type
   GetPoolVoluntaryExitsResponse* = DataEnclosedObject[seq[SignedVoluntaryExit]]
   GetProposerDutiesResponse* = DataRootEnclosedObject[seq[RestProposerDuty]]
   GetSpecResponse* = DataEnclosedObject[RestSpec]
-  GetSpecVCResponse* = DataEnclosedObject[RestSpecVC]
```
arnetheduck, line 686: restspecvc can be removed?

arnetheduck (approved):

PR #5131, opened 2023-06-27
fix code for Nim `v1.6.14` compatibility

etan-status:
With `v1.6.14` there is compilation issue in `trusted_node_sync` where a type is not inferred automatically anymore for a `nil` instance. Fix it so we can bump the compiler.

See https://github.com/status-im/nimbus-build-system/pull/63

### DIFF
--- a/beacon_chain/trusted_node_sync.nim
+++ b/beacon_chain/trusted_node_sync.nim
@@ -150,7 +150,7 @@ proc doTrustedNodeSync*(
                 client.getStateV2(StateIdent.init(StateIdentType.Genesis), cfg),
                 largeRequestsTimeout):
               info "Attempt to download genesis state timed out"
-              nil
+              (ref ForkedHashedBeaconState)(nil)
           except CatchableError as exc:
             info "Unable to download genesis state",
               error = exc.msg, restUrl
--- a/beacon_chain/trusted_node_sync.nim
+++ b/beacon_chain/trusted_node_sync.nim
@@ -150,6 +150,7 @@ proc doTrustedNodeSync*(
                 client.getStateV2(StateIdent.init(StateIdentType.Genesis), cfg),
                 largeRequestsTimeout):
               info "Attempt to download genesis state timed out"
+              # https://github.com/nim-lang/Nim/issues/22180
               (ref ForkedHashedBeaconState)(nil)
           except CatchableError as exc:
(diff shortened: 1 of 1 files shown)


arnetheduck:
while I don't mind the fix, this looks like a regression that should be reported - the compiler has enough information to infer the type here.

etan-status:
@narimiran any idea? :)

narimiran:
> @narimiran any idea? :)

How was this not caught before by the CIs testing the `version-1-6` branch!?

arnetheduck:
nimbus-eth2 I think builds with the actual release version that we use, not the `version-1-6` branch (because it's more important that we check actual release.

This fix should be fairly easy to reduce to a smaller example so it can be upstreamed as a bug.

narimiran:
> nimbus-eth2 I think builds with the actual release version that we use, not the `version-1-6` branch

Then I guess I misunderstood these:
https://github.com/status-im/nimbus-eth2/blob/unstable/.github/workflows/ci.yml#L36 + https://github.com/status-im/nimbus-eth2/blob/unstable/.github/workflows/ci.yml#L215

arnetheduck:
hm, ok - I'm not sure which repo is used there to interpret the version-1-6 tag though: the upstream nim repo or our status-im/Nim fork - if it's the latter, it will only update when someone updates it (I think - this is something to explore / understand / document)

tersec:
One issue is that code such as
```nim
type
  A = object

let a =
  if false:
    (ref A)(nil)
  else:
    #nil  # it can infer across if/else or branches of a single cases statement, but not from if branch => else branch, then into nested `if`
    if false:
      nil
    else:
      nil
``` 

This isn't a regression, though, from 1.6.12,

tersec:
Small repro for regression -- this one keeps the error messages almost verbatim.
```nim
import chronos/asyncloop

type
  A = object

proc doTrustedNodeSync*() {.async.} =
  let x =
    if false:
      (ref A)(nil)
    else:
      if false:
        quit 1
      else:
        if true:
          discard cancelAndWait(default(FutureBase))
          nil
          # (ref A)(nil)
        else:
          (ref A)(nil)
```

tersec:
Reported upstream: https://github.com/nim-lang/Nim/issues/22180

arnetheduck (approved):

PR #5135, opened 2023-06-28
update the RPi guide regarding overclocking

narimiran:
As per discussion with @arnetheduck on Discord.


docs/the_nimbus_book/src/pi-guide.md
```diff
@@ -260,7 +260,14 @@ For more on `raspi-config`, see [here](https://www.raspberrypi.org/documentation
 
 Once you're done, `ssh` back into your Pi.
 
-### 11. Install the beacon node
+
+### 11. Overclocking
+
+Nimbus requires Raspberry Pi to be overclocked.
```
tersec:
```suggestion
Nimbus requires the Raspberry Pi to be overclocked.
```
or "a", but the guide in general seems to aim for a kind of conceptual specificity

arnetheduck:
A few more changes needed for the RPI guide:

* prefer the APT repo as means of installation - the "manual" installation can be moved to a footnote
* APT installation does not have the `run_...` scripts - spell out the commands as if the APT installation was used
* spell out the shell commands to add the apt repo instead of linking to page
* include a step to sync using https://nimbus.guide/trusted-node-sync.html and show the commands to do so with the `nimbus` user (`sudo -u nimbus ...`)
* move the images to this repo instead of outside hosting

narimiran:
> A few more changes needed for the RPI guide:
> 
> * prefer the APT repo as means of installation - the "manual" installation can be moved to a footnote
> * APT installation does not have the `run_...` scripts - spell out the commands as if the APT installation was used
> * [...] show the commands to do so with the `nimbus` user (`sudo -u nimbus ...`)
> * spell out the shell commands to add the apt repo instead of linking to page
> * move the images to this repo instead of outside hosting


These were addressed.

Please check if the commands for points (2) and (3) are correct. I've written them by heart, I don't have such setup nearby so I'm not sure what exactly needs to be specified.
If those are not correct, feel free to use the "Add a suggestion" feature to correct the mistakes.


> * include a step to sync using https://nimbus.guide/trusted-node-sync.html

This was not addressed.
Earlier you said that even the main guide should be updated to use trusted node sync as a primary method of syncing. I plan to do that in a separate PR and then I would also update the RPi guide, as I expect it'll require some coordinated changes in both guides (+ the trusted node sync page itself).

docs/the_nimbus_book/src/pi-guide.md
```diff
@@ -336,11 +361,17 @@ We're finally ready to connect to the Prater testnet!
     For instructions on how to do so, see the [eth1 page](./eth1.md).
 
 To connect to Prater, run:
+```sh
```
arnetheduck: let's put these under `=== "Using package manager"` etc - I *think* all tabs with the same title switch in sync when you press on it as well

docs/the_nimbus_book/src/pi-guide.md
```diff
@@ -260,7 +260,33 @@ For more on `raspi-config`, see [here](https://www.raspberrypi.org/documentation
 
 Once you're done, `ssh` back into your Pi.
 
-### 11. Install the beacon node
+
+### 11. Overclocking
+
+Nimbus requires the Raspberry Pi to be overclocked.
+To overclock your Raspberry Pi, follow the [RPi overclocking guide](https://docs.rocketpool.net/guides/node/local/prepare-pi.html#overclocking-the-pi) by Joe Clapis.
```
arnetheduck: We should put direct instructions here for how to do it instead of linking

docs/the_nimbus_book/src/pi-guide.md
narimiran: > I think all tabs with the same title switch in sync when you press on it as well

Correct, that's how it works.

I've changed it to use this syntax.

docs/the_nimbus_book/src/pi-guide.md
narimiran: The overclocking instructions added, and link is now there for those who want more information.

arnetheduck (approved):

PR #5146, opened 2023-06-29
Complete Deneb bn side for /eth/v2/validator/blocks/{slot}

henridf:


henridf:
Not sure why the build is failing; it passes locally.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,38 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-      if contentType == sszMediaType:
-        let headers = [("eth-consensus-version", message.kind.toString())]
-        withBlck(message):
+      withBlck(message.blck):
+        let data =
+          when blck is deneb.BeaconBlock:
+            var sidecars : seq[BlobSidecar] = @[]
+            let bundle = message.blobsBundleOpt.get()
+            let blockRoot = hash_tree_root(blck)
+            let (blobs, kzgs, proofs) = (bundle.blobs, bundle.kzgs, bundle.proofs)
```
tersec: All three of these are variations on `seq`s (or `List`s, which are just `seq`s with some sugar).

Nim generates less than great code for this. For example,
```nim
let a = default(seq[int])
let copy_of_a = a
```
becomes
```c
  /* ... initialize `a` */
  genericSeqAssign((&copy_of_a__a_9), a__a_8, (&NTIseqLintT__qwqHTkRvwhrRyENtudHQ7g_));
```
where `genericSeqAssign` isn't great., as it calls in turn `genericAssign`:
```c
N_LIB_PRIVATE N_NIMCALL(void, genericSeqAssign)(void* dest, void* src, TNimType* mt) {
  void* src_2;
  src_2 = src;
  genericAssign(dest, ((void*) ((&src_2))), mt);
  popFrame();
}
```
This code will create three of these `genericSeqAssign` calls and therefore 3 `genericAssign` calls, the latter of whic have had a tendency to pop up in profiling results especially when they're of larger structures.

This probably is a case where the admittedly kind of hacky `template blobs: untyped = bundle.blobs` pattern would be better, or even, if it doesn't make things too long, directly using `bundle.blobs`, `bundle.kzgs`, and `bundle.proofs`.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,38 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-        res.get.blck
+        res.get
     return
-      if contentType == sszMediaType:
-        let headers = [("eth-consensus-version", message.kind.toString())]
-        withBlck(message):
+      withBlck(message.blck):
+        let data =
+          when blck is deneb.BeaconBlock:
+            var sidecars : seq[BlobSidecar] = @[]
```
tersec:
```suggestion
            var sidecars: seq[BlobSidecar]
```

tersec: But even better would be to use [newSeqOfCap](https://nim-lang.org/docs/system.html#newSeqOfCap%2CNatural) to exploit that in the `for i in 0..<blobs.len:` loop, the exact ultimate size of `sidecars` will be known, so can be allocated up front, once, rather than either risking over/under-allocating and maybe re-allocating/copying along the way to grow the `seq`.

henridf: 3744ada0

beacon_chain/rpc/rest_validator_api.nim
henridf: Oops - my mechanical sympathy for nim is still off. Thanks for the explanation!

3744ada0

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,37 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-        let headers = [("eth-consensus-version", message.kind.toString())]
-        withBlck(message):
+      withBlck(message.blck):
+        let data =
+          when blck is deneb.BeaconBlock:
+            let bundle = message.blobsBundleOpt.get()
+            let blockRoot = hash_tree_root(blck)
+            var sidecars = newSeqOfCap[BlobSidecar](bundle.blobs.len)
+            for i in 0..<bundle.blobs.len:
+              var sidecar = deneb.BlobSidecar(
```
tersec:
```suggestion
              let sidecar = deneb.BlobSidecar(
```

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,37 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+                slot: blck.slot,
+                block_parent_root: blck.parent_root,
+                proposer_index: blck.proposer_index,
+                blob: bundle.blobs[i],
+                kzg_commitment: bundle.kzgs[i],
+                kzg_proof: bundle.proofs[i]
+              )
+              sidecars.add(sidecar)
+
+            DenebBlockContents(`block`: blck, blob_sidecars: List[BlobSidecar, Limit MAX_BLOBS_PER_BLOCK].init(sidecars))
```
tersec: 121 character line

beacon_chain/rpc/rest_validator_api.nim
henridf: 92cacf807a095137afc13d09f5e74a2e7a586733

beacon_chain/rpc/rest_validator_api.nim
henridf: 92cacf807a095137afc13d09f5e74a2e7a586733

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,40 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
           return RestApiResponse.jsonError(Http400, res.error())
-        res.get.blck
+        res.get
     return
-      if contentType == sszMediaType:
-        let headers = [("eth-consensus-version", message.kind.toString())]
-        withBlck(message):
+      withBlck(message.blck):
+        let data =
+          when blck is deneb.BeaconBlock:
```
tersec, line 417: This condition in isolation is fine -- just as well not to assume what the E-fork `produceBeaconBlockV2` will look like. But right now, it will silently fall back to the pre-Deneb case.

This is another use case for using something similar to a `static: doAssert` or `static: raiseAssert` in an `else` branch of the `when` to simulate the kind of exhaustiveness checking one can get from `case` at runtime. While right now, this indeed covers all forks, it becomes a bit of a maintenance time-hazard because it'll have to be found manually as-is, rather than failing at compile time visibly/obviously when the next fork support gets added.

Otherwise, looks good.

henridf: 9df97f8d70fa69f3635404b594d4881df60044ac

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,43 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
           RestApiResponse.sszResponse(blck, headers)
-      elif contentType == jsonMediaType:
-        withBlck(message):
-          RestApiResponse.jsonResponseWVersion(blck, message.kind)
-      else:
-        raiseAssert "preferredContentType() returns invalid content type"
+        elif contentType == jsonMediaType:
+          RestApiResponse.jsonResponseWVersion(blck, message.blck.kind)
+        else:
+          raiseAssert "preferredContentType() returns invalid content type"
```
tersec, line 449:
```suggestion
          static: raiseAssert "preferredContentType() returns invalid content type"
```
See also
https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/spec/forks.nim#L281-L303
https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/spec/forks.nim#L581-L594

Runtime asserts are slightly risky, but the compile-time version increases safety here

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -418,17 +410,43 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+            DenebBlockContents(
+              `block`: blck,
+              blob_sidecars: List[BlobSidecar,
+                                  Limit MAX_BLOBS_PER_BLOCK].init(sidecars))
+          elif blck is phase0.BeaconBlock or blck is altair.BeaconBlock or
+               blck is bellatrix.BeaconBlock or blck is capella.BeaconBlock:
+            blck
+          else:
+            raiseAssert "produceBlockV2 received unexpected version"
```
tersec:
```suggestion
            static: raiseAssert "produceBlockV2 received unexpected version"
```

Runtime asserts are slightly risky, but the compile-time version increases safety. See also
https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/spec/forks.nim#L581-L594

https://github.com/status-im/nimbus-eth2/blob/187e1a06335e36bbc508fff38729833d154edbaa/beacon_chain/spec/forks.nim#L281-L303

tersec (approved):

PR #5154, opened 2023-07-03
trusted node sync as a primary way of syncing

narimiran:
As per discussion on Discord.


docs/the_nimbus_book/mkdocs.yml
```diff
@@ -57,7 +57,6 @@ nav:
       - 'install.md'
       - 'build.md'
       - 'start-syncing.md'
-      - 'trusted-node-sync.md'
```
arnetheduck, line 60: we generally don't remove files because this breaks links to them - ie I think in general the pages about the two different sync methods should continue to exist, but we simply rephrase any content that links to them giving preference to TNS.

In particular, it's still possible to sync from genesis - just that it takes very long (it's mainly useful on testnets where a trusted node might not be available)

arnetheduck: ah, nevermind, I see the page remains in the system so links don't break :+1: 

that said, I still think the genesis sync remains as a page / supported feature, so here we go a little bit too far in eliminating it

narimiran: > that said, I still think the genesis sync remains as a page / supported feature, so here we go a little bit too far in eliminating it

Ah, ok. I misunderstood what you wanted.

I'll rework this.

narimiran:
Replaced by #5161. Closing.

PR #5155, opened 2023-07-03
rest: allow state root lookups up to 8k slots from head

arnetheduck:


tersec (approved):

beacon_chain/rpc/rest_utils.nim
```diff
@@ -79,6 +79,15 @@ func getBlockSlotId*(node: BeaconNode,
     if stateIdent.root == getStateRoot(node.dag.headState):
       ok(node.dag.head.bid.atSlot())
     else:
+      let headSlot = getStateField(node.dag.headState, slot)
+      for i in 1..<SLOTS_PER_HISTORICAL_ROOT:
+        if i > headSlot:
+          break
+        if getStateField(node.dag.headState, state_roots).item((int headSlot - i) mod int SLOTS_PER_HISTORICAL_ROOT) ==
```
tersec: It's not ideal to have these `uint64` to `int` conversions around, even if practically it's not a likely issue. But it already happens in various other places, so I'm not sure it makes sense to hold up this PR in particular for it.

It's more one of the architectural issues in Nimbus which might (or might not) need addressing more systematically at some point, e.g., all the places where these protocol-`uint64`s get turned into `int` for Nim array indexing purposes with varying levels of validation.

arnetheduck: 7b0d7bf15f833a607e2ed790a661aee24fdc4c08

PR #5167, opened 2023-07-07
Fetch by-root request directly from quarantine

arnetheduck:
When the requestmanager is busy fetching blocks, the queue might get filled with multiple entries of the same root - since there is no deduplication, requests containing the same root multiple times will be sent out.

Also, because the items sit in the queue for a long time potentially, the request might be stale by the time that the manager is ready with the previous request.

This PR removes the queue and directly fetches the blocks to download from the quarantine which solves both problems (the quarantine already de-duplicates and is clean of stale information).

Removing the queue for blobs is left for a future PR.

### DIFF
--- a/beacon_chain/sync/request_manager.nim
+++ b/beacon_chain/sync/request_manager.nim
@@ -34,18 +34,23 @@ const
   BLOB_GOSSIP_WAIT_TIME_NS* = 2 * 1_000_000_000
     ## How long to wait for blobs to arrive over gossip before fetching.
 
+  POLL_FREQUENCY = 1.seconds
+
 type
-  BlockVerifier* =
+  BlockVerifierFn* =
     proc(signedBlock: ForkedSignedBeaconBlock, maybeFinalized: bool):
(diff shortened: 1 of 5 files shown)


beacon_chain/sync/request_manager.nim
```diff
@@ -189,6 +199,8 @@ proc fetchBlobsFromNetwork(self: RequestManager,
     if blobs.isOk:
       let ublobs = blobs.get()
       if not checkResponse(idList, ublobs.asSeq()):
+        debug "Mismatching response to blobs by root",
```
```suggestion
        debug "Mismatched response to blobs by root",
```

tersec (approved):

beacon_chain/sync/request_manager.nim
```diff
@@ -34,18 +34,23 @@ const
   BLOB_GOSSIP_WAIT_TIME_NS* = 2 * 1_000_000_000
     ## How long to wait for blobs to arrive over gossip before fetching.
 
+  POLL_FREQUENCY = 1.seconds
```
henridf, line 37: `POLL_INTERVAL`

PR #5169, opened 2023-07-07
Perform block pre-check before validating execution

arnetheduck:
When syncing, blocks have not been gossip-validated and are therefore prone to trivial faults like being known-unviable, duplicate or missing their parent.

In addition, the duplicate-block check in BlockProcessor was not considering the quarantine flow and would therefore cause recently-quarantined blocks to be silenty dropped when their parent appears delaying the sync end-game and thus causing longer startup resync time.

This PR verifies trivial conditions before performing execution validation thus avoiding duplicates and missing parents alike.

It also ensures that the fast-sync EL mode is used for finalized blocks even if the EL is timing out / slow to respond - this allows the CL to complete its sync faster and switch to "normal" lock-step at the head of the chain more quickly, thus also allowing the EL to access the latest consensensus information earlier.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -51,10 +51,6 @@ const
     ## syncing the finalized part of the chain
   PAYLOAD_PRE_WALL_SLOTS = SLOTS_PER_EPOCH * 2
     ## Number of slots from wall time that we start processing every payload
-  MAX_DEDUP_QUEUE_LEN = 16
-    ## Number of blocks, with FIFO discipline, against which to check queued
-    ## blocks before being processed to avoid spamming ELs. This should stay
-    ## small enough that even O(n) algorithms are reasonable.
 
 type
   BlobSidecars* = seq[ref BlobSidecar]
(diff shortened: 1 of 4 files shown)


tersec (approved):

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -734,19 +754,6 @@ proc addBlock*(
...
     raiseAssert "unbounded queue"
 
-# Dedup
-# ------------------------------------------------------------------------------
-
-func checkDuplicateBlocks(self: ref BlockProcessor, entry: BlockEntry): bool =
-  let key = (entry.blck.root, entry.blck.signature)
-  if self.dupBlckBuf.contains key:
-    return true
-  doAssert self.dupBlckBuf.len <= MAX_DEDUP_QUEUE_LEN
```
tersec, line 744: It looks like it doesn't remove the definition of `MAX_DEDUP_QUEUE_LEN` itself, which should now be unused?

arnetheduck: 2ad3f553d0cf7816a18144671b862241a3e48165

PR #5152, opened 2023-07-02
tweak accelerated shuffling reference state viability check

tersec:
https://github.com/status-im/nimbus-eth2/pull/5137#issuecomment-1614160578

PR #5174, opened 2023-07-10
early exit `commonAncestor` when comparing with `finalizedHead`

etan-status:
As all `BlockRef` lead to `finalizedHead` (`parent == nil`), can shortcut in that situation and immediately return `finalizedHead` if passed as one of the arguments.

### DIFF
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -163,9 +163,9 @@ func commonAncestor*(a, b: BlockRef, lowSlot: Slot): Opt[BlockRef] =
   if a.slot < lowSlot or b.slot < lowSlot:
     return err()
   if a.parent == nil:
-    return ok a  # All `BlockRef` lead to `finalizedHead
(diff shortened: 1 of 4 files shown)


tersec:
Test case? Along the lines of the blockchain DAG test in https://github.com/status-im/nimbus-eth2/pull/5152

tersec (approved):

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -162,6 +162,10 @@ func commonAncestor*(a, b: BlockRef, lowSlot: Slot): Opt[BlockRef] =
   doAssert b != nil
   if a.slot < lowSlot or b.slot < lowSlot:
     return err()
+  if a.parent == nil:
```
arnetheduck, line 165: this is not entirely safe: if someone holds on to a blockref across `await` for example, this condition may be violated

etan-status: Not a new assumption, there are other asserts further down forcing the two blockrefs to be from a consistent view.

PR #5176, opened 2023-07-10
async batch verification (+40% sig verification throughput)

arnetheduck:
When batch verification is done, the main thread is blocked reducing concurrency.

With this PR, the new thread signalling primitive in chronos is used to offload the full batch verification process to a separate thread allowing the main threads to continue async operations while the other threads verify signatures.

Similar to previous behavior, the number of ongoing batch verifications is capped to prevent runaway resource usage.

In addition to the asynchronous processing, 3 addition changes help drive throughput:

* A loop is used for batch accumulation: this prevents a stampede of small batches in eager mode where both the eager and the scheduled batch runner would pick batches off the queue, prematurely picking "fresh" batches off the queue
* An additional small wait is introduced for small batches - this helps create slightly larger batches which make better used of the increased concurrency
* Up to 2 batches are scheduled to the threadpool during high pressure, reducing startup latency for the threads

Together, these changes increase attestation verification throughput under load up to 40%.

More CPU cores used (110 vs 140% usage - the cores here are running other nimbus' - on my node, CPU hovers around 180% usage after this PR):
image

Better attestation throughput (7k -> ~10k/slot)
image

Increased batch size (leading to more efficient verification):
image

Depends on:
https://github.com/status-im/nim-chronos/pull/406
https://github.com/status-im/nim-blscurve/pull/154
https://github.com/status-im/nim-taskpools/pull/33
https://github.com/status-im/nimbus-eth2/pull/5169

### DIFF
--- a/beacon_chain/gossip_processing/batch_validation.nim
+++ b/beacon_chain/gossip_processing/batch_validation.nim
@@ -8,17 +8,18 @@
 {.push raises: [].}
 
 import
-  std/[deques, sequtils],
+  std/[atomics, deques, sequtils],
+  stew/ptrops,
   metrics,
   # Status
-  chronicles, chronos,
+  chronicles, chronos, chronos/threadsync,
   ../spec/signatures_batch,
   ../consensus_object_pools/[blockchain_dag, spec_cache]
 
 export signatures_batch, blockchain_dag
 
 logScope:
-  topics = "gossip_checks"
+  topics = "batch_validation"
 
 declareCounter batch_verification_batches,
   "Total number of batches processed"
@@ -26,6 +27,8 @@ declareCounter batch_verification_signatures,
   "Total number of verified signatures before aggregation"
 declareCounter batch_verification_aggregates,
   "Total number of verified signatures after aggregation"
+declareCounter batch_verification_batches_skipped,
+  "Total number of batches skipped"
 
 # Batched gossip validation
 # ----------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -177,135 +216,158 @@ func combine(a: var PublicKey, b: PublicKey) =
...
 
   let
-    batch = batchCrypto[].batches.popFirst()
-    batchSize = batch[].sigsets.len()
+    startTick = Moment.now()
+
+  # If the hardware is too slow to keep up or an event caused a temporary
+  # buildup of signature verification tasks, the batch will be dropped so as to
+  # recover and not cause even further buildup - this puts an (elastic) upper
+  # bound on the amount of queued-up work
```
mratsim, line 267: What happens to the dropped work?

mratsim (commented):
Excellent.

In that case we can probably go a bit further, create the Taskpool directly in another thread.

Because in this case, assuming you have a Rpi with 4 cores, only 3 will work on cryptography and there is 1 which will handle the rest but I think the rest might not be heavy enough to fully utilize it.

In that case we have a slight oversubscription but the OS routinely handles that fine.

Main issue is changing taskpools to handle work submitted from a thread that is not part of the threadpool.

beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -177,135 +216,158 @@ func combine(a: var PublicKey, b: PublicKey) =
   tmp.aggregate(b)
   a = PublicKey(tmp.finish())
 
-proc processBatch(batchCrypto: ref BatchCrypto) =
-  ## Process one batch, if there is any
-
-  # Pruning the queue here makes sure we catch up with processing if need be
-  batchCrypto.pruneBatchQueue() # Skip old batches
+proc batchVerifyTask(task: ptr BatchTask) {.nimcall.} =
+  # Task suitable for running in taskpools - look, no GC!
```
elcritch, line 220: Ah, I see you're just skipping the GC.

beacon_chain/gossip_processing/batch_validation.nim
arnetheduck: the attestations are dropped instead of being processed -> no gossip etc

arnetheduck:
> In that case we can probably go a bit further, create the Taskpool directly in another thread.

with a Flowvar in hand, one can wait for specific tasks instead of the whole pool so which thread "owns" the pool should ideally be less critical - it's already an mpmc:ish kind of setup where tasks can spawn more tasks and wait for tasks from anywhere, no? This, along with a preference to post tasks to not-this-thread.

potentially one should allow `Flowvar[void]` for synchronization purposes when the data is transferred out-of-band as often happens.

alokeutpal (approved):

PR #5178, opened 2023-07-10
avoid unnecessary buffer zero:ing

arnetheduck:


tersec:
Safe because
```nim
    if dataLen > 0:
      try:
        await conn.readExactly(addr frameData[0], dataLen)
      except LPStreamEOFError, LPStreamIncompleteError:
        return err "Incomplete snappy frame"
```

tersec (approved):

PR #5184, opened 2023-07-12
VC: Refactor sync committee duties and selection proofs calculation.

cheatfate:


beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+proc cmp(x, y: AttestationSlotRequest|SyncCommitteeSlotRequest): int =
+  if x.slot == y.slot: 0 elif x.slot < y.slot: -1 else: 1
+
+func getSyncSelectionOffset*(proof: SyncCommitteeSelectionProof,
+                             inindex: IndexInSyncCommittee,
+                             slot: Slot): int =
+  let index = proof.validator_sync_committee_indices.find(inindex)
+  if index == -1:
+    return -1
+  int(Epoch(uint64(index)).start_slot() + (slot - slot.epoch().start_slot()))
```
etan-status:
```suggestion
  int(Epoch(uint64(index)).start_slot() + slot.since_epoch_start())
```

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+              map[].duties.withValue(request.slot.epoch(), dap):
+                dap[].slotSig = signature
+        (res1, res2)
+  sigres
+
+proc updateRuntimeConfig*(vc: ValidatorClientRef,
+                          info: VCRuntimeConfig): Result[void, string] =
+  let res = info.getOrDefault("ALTAIR_FORK_EPOCH", FAR_FUTURE_EPOCH)
+  if res == FAR_FUTURE_EPOCH:
+    return err("Missing or invalid ALTAIR_FORK_EPOCH value")
```
etan-status: Why invalid? It could simply be a private network that doesn't have `ALTAIR_FORK_EPOCH` scheduled yet.

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+      return err("Different ALTAIR_FORK_EPOCH value")
+  else:
+    vc.runtimeConfig.altairEpoch = Opt.some(res)
+  ok()
+
+proc `+`*(slot: Slot, epochs: Epoch): Slot =
+  let
+    slotEpoch = slot.epoch()
+    slotInEpoch = slot - slotEpoch.start_slot()
+  (slotEpoch + uint64(epochs)).start_slot() + slotInEpoch
```
etan-status:
```suggestion
  slot + epochs * SLOTS_PER_EPOCH
```

beacon_chain/validator_client/sync_committee_service.nim
```diff
@@ -211,128 +209,132 @@ proc produceAndPublishContributions(service: SyncCommitteeServiceRef,
...
+      var
+        resItems: seq[ContributionItem]
+        resFutures: seq[FutureBase]
+        resMap: array[SYNC_COMMITTEE_SUBNET_COUNT,
+                      Future[SyncCommitteeContribution]]
+      for duty in duties:
+        for inindex in duty.validator_sync_committee_indices:
+          let
+            validator = vc.getValidatorForDuties(duty.pubkey, slot).valueOr:
+              continue
```
etan-status:
```suggestion
        let validator = vc.getValidatorForDuties(duty.pubkey, slot).valueOr:
          continue
        for inindex in duty.validator_sync_committee_indices:
          let
```

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+      for index, item in asorted.pairs():
+        if item != bsorted[index]:
+          return false
+      true
+
+func `==`*(a, b: SyncCommitteeDuty): bool =
+  (a.pubkey == b.pubkey) and
+  (a.validator_index == b.validator_index) and
+  compareUnsorted(a.validator_sync_committee_indices,
+                  b.validator_sync_committee_indices)
```
etan-status, line 1416: Alternatively, could just sort the `validator_sync_committee_indices` when creating a `SyncCommitteeDuty`. It is extremely unlikely for a validator to be assigned multiple indices, so the extra sort step could be skipped in the vast majority of real cases.

On mainnet, each sync committee subnet has 128 validators, and there are over 700k validators total. Maybe, once every couple years, a single validator will end up twice in a subnet, if at all, where the sorting would actually be required.

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+
+proc cmp(x, y: AttestationSlotRequest|SyncCommitteeSlotRequest): int =
+  if x.slot == y.slot: 0 elif x.slot < y.slot: -1 else: 1
```
etan-status:
```suggestion
  cmp(x.slot, y.slot)
```

beacon_chain/validator_client/common.nim
```diff
@@ -130,11 +130,18 @@ type
     index*: int
     timeOffset*: Opt[TimeOffset]
 
+  SyncCommitteeSelectionProof* = object
+    signatures*: seq[Opt[ValidatorSig]]
+    validator_sync_committee_indices*: seq[IndexInSyncCommittee]
```
etan-status: As this stores all signatures for a full `Epoch`, wondering whether that should be reflected in the name for clarity, e.g., `EpochSyncCommitteeSelectionProofs` or smth similar.

Or, encapsulating each epoch into an object, could simplify the relatively hard to read offset computation logic (`getSyncSelectionOffset` etc)

```nim
EpochSelectionProof* = object
  signatures*: array[SLOTS_PER_EPOCH, Opt[ValidatorSig]]
  validator_sync_committee_index: IndexInSyncCommittee

SyncCommitteeSelectionProof* = seq[EpochSelectionProof]
```

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+       vc: ValidatorClientRef,
+       pubkey: ValidatorPubKey,
+       slot: Slot,
+       inindex: IndexInSyncCommittee
+     ): Opt[ValidatorSig] =
+  vc.syncCommitteeProofs.withValue(slot.epoch(), epochProofs):
+    epochProofs[].proofs.withValue(pubkey, validatorProofs):
+      let offset = getSyncSelectionOffset(validatorProofs[], inindex, slot)
+      return
+        if offset < len(validatorProofs[].signatures):
```
etan-status:
```suggestion
        if offset >= 0 and offset < len(validatorProofs[].signatures):
```

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+  (a.validator_index == b.validator_index) and
+  compareUnsorted(a.validator_sync_committee_indices,
+                  b.validator_sync_committee_indices)
+
+proc cmp(x, y: AttestationSlotRequest|SyncCommitteeSlotRequest): int =
+  if x.slot == y.slot: 0 elif x.slot < y.slot: -1 else: 1
+
+func getSyncSelectionOffset*(proof: SyncCommitteeSelectionProof,
+                             inindex: IndexInSyncCommittee,
+                             slot: Slot): int =
```
etan-status: Should this return an `Opt[int]` instead? Could allow dropping `isSyncSelectionSignature`, and simplify the range checks that are currently present in all callers of `getSyncSelectionOffset`, by allowing `valueOr` syntax.

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+      block:
+        var res: seq[SyncCommitteeSlotRequest]
+        for epoch in start.epoch() .. finish.epoch():
+          let
+            fork = vc.forkAtEpoch(epoch)
+            period = epoch.sync_committee_period()
+          for duty in vc.syncDutiesForPeriod(period):
+            let validator = vc.attachedValidators[].
+              getValidator(duty.pubkey).valueOr:
+                # Ignore all the validators which are not here anymore
```
etan-status, line 1507: Hmm. What happens to validators that have exited during the period? Probably alright to just continue to try and having errors for the remaining slots.

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+                if not(proof.isSyncSelectionSignature(inindex, slot)):
+                  res.add(
+                    SyncCommitteeSlotRequest(
+                      validator: validator,
+                      fork: fork,
+                      slot: slot,
+                      duty: duty,
+                      sync_committee_index: inindex))
+        # We make requests sorted by slot number.
+        sorted(res, cmp, order = SortOrder.Ascending)
```
etan-status, line 1527: Considering the requests are all fired off concurrently, does the sorting actually make a difference?

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+        getSubcommitteeIndex(it.sync_committee_index))))
+
+  while len(pendingRequests) > 0:
+    try:
+      discard await race(pendingRequests)
+    except CancelledError as exc:
+      var pending: seq[Future[void]]
+      for future in pendingRequests:
+        if not(future.finished()): pending.add(future.cancelAndWait())
+      await allFutures(pending)
```
etan-status, line 1541:
```suggestion
      while true:
        try:
          await allFutures(pending)
          break
        except CancelledError:
          discard
```
      
```

File: beacon_chain/validator_client/common.nim [hunk 13]
```diff
@@ -1347,3 +1390,289 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+            if validator.index.isNone():
+              # Ignore all the valididators which do not have index yet.
+              continue
+            let proof = vc.getSyncCommitteeSelectionProof(duty.pubkey, epoch).
+                          get(default(SyncCommitteeSelectionProof))
+            for inindex in duty.validator_sync_committee_indices:
+              for slot in epoch.slots():
+                if slot < start: continue
+                if slot > finish: break
+                if not(proof.isSyncSelectionSignature(inindex, slot)):
```
etan-status: Wondering if this should also retry if `isSyncSelectionSignature` is `true`, but the inner `Opt[ValidatorSig]` is still `none`, similar to the logic for attestations.

beacon_chain/validator_client/duties_service.nim
```diff
@@ -99,237 +100,165 @@ proc pollForValidatorIndices*(service: DutiesServiceRef) {.async.} =
...
-    remainingItems -= arraySize
-
+proc pollForSyncCommitteeDuties*(
+       service: DutiesServiceRef,
+       period: SyncCommitteePeriod
+     ): Future[int] {.async.} =
   let
+    vc = service.client
+    indices = toSeq(vc.attachedValidators[].indices())
+    epoch = period.start_epoch()
```
etan-status:
```suggestion
    epoch = max(period.start_epoch(), service.runtimeConfig.altairEpoch)
```

`ALTAIR_FORK_EPOCH` is not necessarily aligned with a sync committee boundary

beacon_chain/validator_client/duties_service.nim
```diff
@@ -369,6 +298,16 @@ proc pollForAttesterDuties*(service: DutiesServiceRef) {.async.} =
     if (counts[0].count == 0) and (counts[1].count == 0):
       debug "No new attester's duties received", slot = currentSlot
 
+    block:
+      let
+        moment = Moment.now()
+        sigres = await service.fillAttestationSelectionProofs(
+          currentSlot, currentSlot + Epoch(1))
+      debug "Slot signatures has been received",
```
etan-status:
```suggestion
      debug "Slot signatures have been received",
```

beacon_chain/validator_client/duties_service.nim
```diff
@@ -405,72 +344,79 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef) {.async.} =
...
     currentSlot = vc.getCurrentSlot().get(Slot(0))
-    # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/altair/validator.md#sync-committee
-    dutySlot = currentSlot + 1
-    dutyEpoch = dutySlot.epoch()
+    currentEpoch = currentSlot.epoch()
+    altairEpoch = vc.runtimeConfig.altairEpoch.valueOr:
+      return
+
+  if currentEpoch < altairEpoch:
+    # We are not going to poll for sync committee duties until `altairEpoch`.
```
etan-status, line 362: Hmm. Or maybe start polling it one epoch early, to ensure that the data is ready by the time duties start?

Fine also with the current implementation, the relevant networks are all post-Altair. Would have to adjust the local testnet simulation to start in phase0 for this to be testable.

beacon_chain/validator_client/duties_service.nim
```diff
@@ -405,72 +344,79 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef) {.async.} =
...
+
+    if (counts[0].count == 0) and (counts[1].count == 0):
+      debug "No new sync committee duties received", slot = currentSlot
+
+    block:
+      let
+        moment = Moment.now()
+        sigres = await service.fillSyncCommitteeSelectionProofs(
+          currentSlot, currentSlot + Epoch(AGGREGATION_PRE_COMPUTE_EPOCHS))
+      debug "Sync committee selection proofs has been received",
```
etan-status:
```suggestion
      debug "Sync committee selection proofs have been received",
```
(plural has --> have)

beacon_chain/validator_client/duties_service.nim
```diff
@@ -405,72 +344,79 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef) {.async.} =
...
-            (epoch: dutyEpoch,
-             period: dutyPeriod)
-          )
-          if lookaheadPeriod > dutyPeriod:
-            res.add(
-              (epoch: lookaheadPeriod.start_epoch(),
-               period: lookaheadPeriod)
-            )
+            lookaheadSlot = currentSlot +
+                            SUBSCRIPTION_LOOKAHEAD_EPOCHS * SLOTS_PER_EPOCH
```
etan-status, line 398: Technically, this should be randomized according to spec:

https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#sync-committee-subnet-stability

> To join a sync committee subnet, select a random number of epochs before the end of the current sync committee period between 1 and SYNC_COMMITTEE_SUBNET_COUNT, inclusive. Validators should join their member subnet at the beginning of the epoch they have randomly selected. For example, if the next sync committee period starts at epoch 853,248 and the validator randomly selects an offset of 3, they should join the subnet at the beginning of epoch 853,245. Validators should leverage the lookahead period on sync committee assignments so that they can join the appropriate subnets ahead of their assigned sync committee period.

Out of scope for this PR / it was broken before (and likely is theoretically incorrect in the BN as well).

beacon_chain/validator_client/common.nim
cheatfate: We sending signature requests ordering by slot, so in ideal world we could expect first sent - first served, so FIFO (first input first output) approach. And in ideal world we would get first slots to be filled with signatures first, it means that concurrent tasks could access and obtain selection proofs for slots. Also we are executing this filling procedure every slot, so we could fill empty spots which we unable to fill at first attempt.

beacon_chain/validator_client/common.nim
cheatfate: If you check `cmpUnsorted` implementation you will see that only if array size is 3 - sorting will be performed, with 2 it will perform 2 comparisons. What is probability of 3?

beacon_chain/validator_client/common.nim
cheatfate: I'm trying to avoid `beacon_time` constants usage, something telling me it is gonna change soon.

beacon_chain/validator_client/duties_service.nim
```diff
@@ -405,72 +344,79 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef) {.async.} =
...
+  let
+    currentPeriod = currentEpoch.sync_committee_period()
+    nextPeriod = currentPeriod + 1
 
   if vc.attachedValidators[].count() != 0:
+    var counts: array[2, tuple[period: SyncCommitteePeriod, count: int]]
+    counts[0] = (currentPeriod,
+                 await service.pollForSyncCommitteeDuties(currentPeriod))
+    counts[1] = (nextPeriod,
+                 await service.pollForSyncCommitteeDuties(nextPeriod))
```
etan-status, line 374: `nextPeriod` is probably good enough to fetch shortly before the end of current period. Just in case new validators get added to the VC / exited. Maybe bundle together with the subscriptions lookahead, or just do 1 epoch in advance (similar to attestations)

beacon_chain/validator_client/common.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/sync_committee_service.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: This function was removed.
https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: I think at some point BN will return an error.

beacon_chain/validator_client/common.nim
cheatfate: This function was removed.
https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/duties_service.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/duties_service.nim
cheatfate: https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/common.nim
cheatfate: This code was removed.
https://github.com/status-im/nimbus-eth2/pull/5184/commits/387016a8c9ad17fc30514a90ac9bfb7f14a52aae

beacon_chain/validator_client/duties_service.nim
cheatfate: Duties are requested every slot, so if validators will be added/exited this will change duties appropriately.

beacon_chain/validator_client/duties_service.nim
cheatfate: But according to spec in non-altair epoch this calls should not be available, so [/eth/v1/validator/duties/sync/{epoch}](https://ethereum.github.io/beacon-APIs/#/Validator/getSyncCommitteeDuties) could be not available before Altair.

beacon_chain/validator_client/duties_service.nim
cheatfate: This procedure will be called only for post-altair epochs.

beacon_chain/validator_client/common.nim
cheatfate: I dont think this approach is needed here, we are not using multiple cancellation usually and if it really happens all futures in `pending` will not be cancelled...

beacon_chain/validator_client/duties_service.nim
cheatfate: Between 1 and 128 for `mainnet`?

beacon_chain/validator_client/common.nim
etan-status: Probably 0 on mainnet 😅

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1388,284 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+                   inindex: IndexInSyncCommittee,
+                   slot: Slot): bool =
+  let index = proof.getIndex(inindex).valueOr: return false
+  proof[index].signatures[int(slot.since_epoch_start())].isSome()
+
+proc setSignature*(proof: var SyncCommitteeSelectionProof,
+                   inindex: IndexInSyncCommittee, slot: Slot,
+                   signature: Opt[ValidatorSig]) =
+  let index = proof.getIndex(inindex).valueOr:
+    raiseAssert "EpochSelectionProof should be present at this moment"
```
etan-status:
```suggestion
  let index = proof.getIndex(inindex).expect(
    "EpochSelectionProof should be present at this moment")
```

beacon_chain/validator_client/duties_service.nim
etan-status: `ALTAIR_FORK_EPOCH.sync_committee_period.start_epoch` may be < `ALTAIR_FORK_EPOCH`, meaning that ultimately what is being sent in the request could be a pre-altair epoch when this function is called with `period = ALTAIR_FORK_EPOCH.sync_committee_period`.

etan-status: Actually, the problem could be avoided by requesting `(period + 1).start_epoch - 1` :) That one is always `>= ALTAIR_FORK_EPOCH`. This would also allow tracking validators that exited during a period.

Not sure how well the non-Nimbus BNs like that though. It is fine according to beacon-APIs specs, imo, but implementations may still be glitchy

beacon_chain/validator_client/duties_service.nim
```diff
@@ -99,237 +100,165 @@ proc pollForValidatorIndices*(service: DutiesServiceRef) {.async.} =
...
+          map[].duties.withValue(period, periodDuty):
+            dutyFound = true
+            if checkReorg(periodDuty[], duty):
+              if not(alreadyWarned):
+                info "Sync committee duties re-organization"
+                alreadyWarned = true
+              res.add((period, duty))
+        if not(dutyFound):
+          res.add((period, duty))
+          info "Received new sync committee duty", duty, period = period
```
etan-status:
```suggestion
          info "Received new sync committee duty", duty, period
```

beacon_chain/validator_client/duties_service.nim
etan-status: That's a good point, so it may or may not work if queried earlier. Anyhow, this probably would only affect the initial altair slot, if at all, and there is no practical relevance to optimize for that case anymore, as Altair is well in the past on relevant networks. Fine with keeping it as is.

beacon_chain/validator_client/duties_service.nim
etan-status: Got it. So, this is all good then, with the caveat of `nextPeriod` requested well before it is useful, but tha't s not hurting correctness.

beacon_chain/validator_client/duties_service.nim
etan-status: `SYNC_COMMITTEE_SUBNET_COUNT` is `4` on mainnet.

https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/altair/validator.md#misc

Actually doesn't even seem to be network specific, but instead a pure constant

etan-status: I actually don't really understand the rationale of this weird randomized slow joining of sync meshes - it seems to work well enough without doing it, and I don t think our BN randomizes the joining either.

beacon_chain/validator_client/common.nim
```diff
@@ -1347,3 +1388,284 @@ proc waitForNextSlot*(service: ClientServiceRef) {.async.} =
...
+              map[].duties.withValue(request.slot.epoch(), dap):
+                dap[].slotSig = signature
+        (res1, res2)
+  sigres
+
+proc updateRuntimeConfig*(vc: ValidatorClientRef,
+                          info: VCRuntimeConfig): Result[void, string] =
+  let res = info.getOrDefault("ALTAIR_FORK_EPOCH", FAR_FUTURE_EPOCH)
+  if res == FAR_FUTURE_EPOCH:
+    return err("Missing ALTAIR_FORK_EPOCH value")
```
etan-status:
```suggestion
    return ok()
```

It's not necessarily missing, it could actually be set to `FAR_FUTURE_EPOCH`, e.g., on a pure phase0 testnet.

I think ignoring the value even if it is explicitly set to `FAR_FUTURE_EPOCH` is fine, to support the case where the hypothetical pure phase0 testnet BN gets restarted with a new config that now has set `ALTAIR_FORK_EPOCH`, while other BNs may not have restarted yet and are still reporting `FAR_FUTURE_EPOCH`. Not sure if we actually aim to support that scenario at all, though; if we require the entire cluster to schedule new forks simultaneously, I guess could `return err` if the value is actually missing in the `info`, but otherwise just treat `FAR_FUTURE_EPOCH` same as any other value.

etan-status (approved):

beacon_chain/validator_client/duties_service.nim
etan-status: re "This would also allow tracking validators that exited during a period.", no guarantees that this actually works that way :) I suspect that BNs just return the same info regardless of what epoch is actually being queried.

To return a result that takes into account validators that may have exited during a period, the BN would have to apply the epoch transition function across all the epochs to determine whether a validator would leak into an exit if all slots were empty 😅

In case someone actually implements the BN like that, we would actually have to be careful with the request, and request:

- `period.start_epoch` for future periods (and ensure we don't call it too early)
- `wallClock.slotOrZero.epoch` for the current period
- `(period + 1).start_epoch - 1` for periods in the past (don't think there is a path that does this

Too many open questions, the beacon-API spec is not clear what is expected regarding exits, and implementations may interpret it differently. So, needs a bit of empirical testing.

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/5184/commits/429267892cd9a808e27d4fbc6f59a91096c1e32c

beacon_chain/validator_client/common.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/5184/commits/429267892cd9a808e27d4fbc6f59a91096c1e32c

beacon_chain/validator_client/duties_service.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/5184/commits/429267892cd9a808e27d4fbc6f59a91096c1e32c

beacon_chain/validator_client/common.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/5184/commits/429267892cd9a808e27d4fbc6f59a91096c1e32c

PR #5186, opened 2023-07-12
allow initializing test DBs from later forks

etan-status:
Running `makeTestDB` in tests currently always initializes DB with a `phase0` state, preventing tests that configure a fork schedule that starts in a different fork from working properly. Fix that by upgrading the genesis state to whatever fork the fork schedule starts with.

### DIFF
--- a/tests/testdbutil.nim
+++ b/tests/testdbutil.nim
@@ -39,7 +39,7 @@ proc makeTestDB*(
   # Upgrade genesis state to later fork, if required by fork schedule
   cfg.maybeUpgradeState(genState[])
   withState(genState[]):
-    when consensusFork >= ConsensusFork.Phase0:
+    when consensusFork > ConsensusFork.Phase0:
       forkyState.data.fork.previous_version =
         forkyState.data.fork.current_version
       forkyState.data.latest_block_header.body_root =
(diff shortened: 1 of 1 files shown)


tests/testdbutil.nim
```diff
@@ -21,22 +21,31 @@ proc makeTestDB*(
...
+      makeInitialDeposits(validators.uint64, flags),
+      flags))
 
   # Override Eth1Data on request, skipping the lengthy Eth1 voting process
   if eth1Data.isOk:
     withState(genState[]):
       forkyState.data.eth1_data = eth1Data.get
       forkyState.root = hash_tree_root(forkyState.data)
 
+  # Upgrade genesis state to later fork, if not starting at Phase0
```
tersec: Is the intent here indeed just to update the state to newer forks, not to have this happen naturally by running through enough slots to trigger the `maybeUpgradeState` organically?

https://github.com/ethereum/consensus-specs/issues/2902 means this can produce odd effects in general, though for this purpose, it should be fine -- if there are no attestations or similar from before it upgrades forks, then there's nothing to get rejected by the fork-dependent signature checks.

etan-status: This is for the genesis state (slot == 0), e.g., starting at bellatrix

etan-status: Adjusted it so that `fork.previous_version == fork.current_version` in that case, e.g., both set to Bellatrix.

PR #5188, opened 2023-07-13
Integrate `hashtree` for better hashing performance

arnetheduck:
https://github.com/status-im/nim-ssz-serialization/pull/35 brings in https://github.com/prysmaticlabs/hashtree as a supported backend for SHA256, giving a nice little performance boost to all `hash_tree_root` calls on supported platforms / compilers.

Expected gains are on the order of 30% which in the case of a replayed state nets us 0.2-0.3s improvement.

More about this design here: https://hackmd.io/@potuz/BJyrx9DOF - kudos to @potuz for this excellent library!


tersec:
```
fatal: remote error: upload-pack: not our ref dc1992a73be5643cf4b06bd29a22478f8d32513b
Errors during submodule fetch:
	vendor/nim-ssz-serialization
```
probably that's not available in the git repo, but this error triggers regardless on `git pull` in such cases, even when not having checked out this branch specifically

.github/workflows/ci.yml
```diff
@@ -136,6 +136,10 @@ jobs:
             fi
           fi
 
+          if [[ '${{ runner.os }}' == 'Windows' ]]; then
+            export NIMFLAGS="${NIMFLAGS} --cc:clang"
+          fi
```
tersec, line 141: Does this imply new build requisites for https://nimbus.guide/install.html#build-prerequisites ?

arnetheduck: not really, this is something our ci builds are set up to do explicitly..

arnetheduck: the problem before this line was that clang pretends to be gcc (via executable name) but isn't compatible enough

tersec (approved):

PR #5192, opened 2023-07-14
generalize `commonAncestor` function to `BlockId`

etan-status:
To enable additional use cases, e.g., `/states/###/randao` beacon API, `ShufflingRef` acceleration logic needs to be able to operate on parts of the DAG that do not have `BlockRef`. Changing `commonAncestor` to act on `BlockId` instead of `BlockRef` is a step toward that and also simplifies the logic some more.


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1330,7 +1330,44 @@ proc getFinalizedEpochRef*(dag: ChainDAGRef): EpochRef =
...
+  if stateBid.slot < lowSlot:
+    return Opt.none(Slot)
+
+  var blockBid = (? dag.atSlot(bid, stateBid.slot)).bid
+  if blockBid.slot < lowSlot:
+    return Opt.none(Slot)
+
+  while stateBid != blockBid:
+    if stateBid.slot >= blockBid.slot:
+      stateBid = ? dag.getBlockIdAtSlot(state, stateBid.slot - 1)
```
```suggestion
      stateBid = ? dag.getBlockIdAtSlot(
        state, min(blockBid.slot, stateBid.slot - 1))
```
slight optimization in case of empty slots.

PR #5194, opened 2023-07-17
increase `getPayloadFromSingleEL` timeout

etan-status:
When producing a local block and `runProposalForkchoiceUpdated` was missed, `getPayloadFromSingleEL` adds additional ~500ms of latency. Quick fix to avoid missing blocks to that.


etan-status:
See https://github.com/status-im/nimbus-eth2/pull/4465

PR #5195, opened 2023-07-17
fall back to non-fcu fork choice on epoch boundaries

tersec:


beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -340,11 +340,11 @@ proc getGasLimit*(
 from ../spec/datatypes/bellatrix import PayloadID
 
 proc runProposalForkchoiceUpdated*(
-    self: ref ConsensusManager, wallSlot: Slot) {.async.} =
+    self: ref ConsensusManager, wallSlot: Slot): Future[bool] {.async.} =
```
etan-status:
```suggestion
    self: ref ConsensusManager, wallSlot: Slot): Future[Opt[void]] {.async.} =
```

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -340,11 +340,11 @@ proc getGasLimit*(
...
   let
     nextWallSlot = wallSlot + 1
     (validatorIndex, nextProposer) = self.checkNextProposer(wallSlot).valueOr:
-      return
+      return false
```
etan-status: `? self.checkNextProposer(wallSlot)`, if going with `Opt[void]`

etan-status (approved):

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: Yeah, will update these to be consistent

etan-status: also, Future[Opt[void]], accidentally forgot the Future wrapper on first attempt

tersec: https://github.com/status-im/nimbus-eth2/pull/5195/commits/88ea3e23303d1d4fa833a5012a48583bdb5aad09

I did try `? self.checkNextProposer` and it didn't seem to work, but can investigate later. The `valueOr` approach still works.

PR #5202, opened 2023-07-19
avoid Nim 2.0 issue when loading config

etan-status:
To avoid Nim 2.0 issue https://github.com/nim-lang/Nim/issues/22284, explicitly specify `ref SecondarySources` instead of using `auto`, and add `{.raises.}` annotation

### DIFF
--- a/ncli/deposit_downloader.nim
+++ b/ncli/deposit_downloader.nim
@@ -82,6 +82,8 @@ proc main(flags: CliFlags) {.async.} =
 
 waitFor main(
   load(CliFlags,
-       secondarySources = proc (config: CliFlags, sources: auto) =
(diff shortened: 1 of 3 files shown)


ncli/deposit_downloader.nim
```diff
@@ -82,6 +82,8 @@ proc main(flags: CliFlags) {.async.} =
 
 waitFor main(
   load(CliFlags,
-       secondarySources = proc (config: CliFlags, sources: auto) =
+       secondarySources = proc (
+          config: CliFlags, sources: ref SecondarySources
```
etan-status:
```suggestion
           config: CliFlags, sources: ref SecondarySources
```

PR #5204, opened 2023-07-20
generalize `ShufflingRef` acceleration logic (#5197)

etan-status:
Split up the `ShufflingRef` acceleration logic into generically usable parts and attester shuffling specific parts. The generic parts could be used to accelerate other purposes, e.g., REST `/states/xxx/randao` API.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -749,6 +749,17 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         # the state will be obtained.
         bslot.slot.epoch()
 
+    # Try to obtain RANDAO in an accelerated way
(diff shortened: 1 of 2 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -749,6 +749,17 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         # the state will be obtained.
         bslot.slot.epoch()
 
+    # Try to obtain RANDAO in an accelerated way
```
arnetheduck, line 752: this feels like it should be a function in chaindag rather than here (so that "anywhere" that needs a randao as fast as possible can just call it - ie the rest api should just be a thin validation-and-translation layer

etan-status: REST uses a specific method for caching states queried through it, which is why the access to the REST specific cache is separate.

arnetheduck (approved):

PR #5207, opened 2023-07-21
speed up state/block loading

arnetheduck:
When loading blocks and states from db/era, we currently redundantly check their CRC32 - for a state, this costs 50ms of loading time presently (110mb uncompressed size) on a decent laptop.

* remove `maxDecompressedDbRecordSize` - not actually used on recent data since we store the framed format - also, we're in luck: we blew past the limit quite some time ago
* fix obsolete exception-based error checking
* avoid `zeroMem` when reading from era store

see https://github.com/status-im/nim-snappy/pull/22 for benchmarks

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -203,13 +203,6 @@ type
     slot*: Slot
     parent_root*: Eth2Digest
 
-const
-  # The largest object we're saving is the BeaconState, and by far, the largest
-  # part of it is the validator - each validator takes up at least 129 bytes
-  # in phase0,  which means 100k validators is >12mb - in addition to this,
-  # there are several MB of hashes.
-  maxDecompressedDbRecordSize = 64*1024*1024
-
 # Subkeys essentially create "tables" within the key-value store by prefixing
 # each entry with a table id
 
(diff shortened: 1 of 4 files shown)


zah:
Can you clarify the claim why you find these checks to be redundant? Is it because the code computes `hash_tree_root` of the loaded state? Does this happen on every code path loading a state? Or is this a claim that Sqlite3 itself verifies the integrity of the data, which doesn't seem to be the case:
https://www.sqlite.org/cksumvfs.html

arnetheduck:
>  Is it because the code computes hash_tree_root of the loaded state?
Yes

> Does this happen on every code path loading a state?

Sort of - it happens as soon as a slot transition or block application happens - that said, this PR does increase risk slightly: in case of the wrong kind of data corruption, decompression could plausibly create a bogus state that gets read from in some code path where   no HTR is called - however, the crc32 check was never really an intended part of the database design, ie a data corruption can happen when loading a public key or any other uncompressed data from disk (which does not pass through snappy) - if we want consistent crc32 checking, this needs to be an explicit feature on top of all database data.

Notably, our crc32 implementation is also a ridiculously slow one - >10x faster implementations exist, like https://github.com/google/crc32c - if we go down that path, we should switch to a proper implementation.

PR #5208, opened 2023-07-22
feat: add chiado

filoozom:
Hi, this adds the Chiado (Gnosis testnet) config. It's a fairly trivial change as we already have the preset and config folder.

I've tried running Chiado with the `--network` flag but it doesn't seem to work. More information in [this Discord thread](https://discord.com/channels/613988663034118151/812268053718237196/1128994962156028047).

I'm not entirely sure what I need to change in the `Makefile` regarding `GNOSIS_WEB3_URLS`.


etan-status (approved):

etan-status:
Thanks for the contribution :) Having easy access to the primary public testnet for Gnosis is useful, and also in line with how EF networks are handled (also re extra binary size).

PR #5211, opened 2023-07-24
update FAQ

narimiran:


docs/the_nimbus_book/src/faq.md
```diff
@@ -35,20 +49,18 @@ Enable it by passing `--rest` to the run command:
 build/nimbus_beacon_node --rest ...
 ```
 
-### Why does my validator miss two epochs of attestations after restarting?
-
-When a validator is started (or restarted) it prudently listens for 2 epochs for attestations from a validator with the same index (a doppelganger), before sending an attestation itself.
+### Why does my validator miss two epochs of attestations after (re)starting?
 
-In sum, it's a simple way of handling the case where one validator comes online with the same key as another validator that's already online (i.e. one device was started without switching the other off).
+When a validator is started (or restarted), it prudently listens for 2 epochs for attestations from a validator with the same index (a doppelganger), before sending an attestation itself.
```
tersec: Prudently reads slightly oddly here -- it's something which typically applies to human-like agents. I'm not sure if software can do anything "prudently", really.

Maybe more importantly, it's characterizing tonally what the rest of your text describes factually already, so it's a bit redundant.

tersec: Also, for validators, there's an mapping of index <-> public key (though not-yet-active validators might not have indices, but only public keys, so it's not 1:1), and so it depends which is more intuitively understood by Nimbus users. Ethereum presents both abstractions, but the public key of a validator is sort of what Nimbus controls directly -- the index is a function of that, not vice versa. Seems like maybe using public key, or some other way of phrasing validator identity, might be better here.

docs/the_nimbus_book/src/faq.md
```diff
@@ -159,7 +199,7 @@ The answer to this question very much depends on how much ETH you have at your d
 
 You should certainly top up if your balance is close to 16 ETH: this is to ensure you don't get removed from the validator set (which automatically happens if your balance falls below 16 ETH).
 
-At the other end of the spectrum, if your balance is closer to 31 ETH, it's probably not worth your while adding the extra ETH required to get back to 32.
+At the other end of the spectrum, if your balance is closer to 31 ETH, it's probably not worth your while adding the extra ETH to get back to the required 32.
```
tersec: Only 16 is required, per `EJECTION_BALANCE`: https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#validator-cycle so this change makes this sentence incorrect.

docs/the_nimbus_book/src/faq.md
narimiran: > Prudently reads slightly oddly here

Yeah, this was already existing and when I read it, it also struck me as a bit weird. I'm glad that you feel the same. Now that are two of us — I'll remove it.


> Seems like maybe using public key, or some other way of phrasing validator identity, might be better here.

Ok, I'll change it.

docs/the_nimbus_book/src/faq.md
narimiran: Now reverted. Thanks!

docs/the_nimbus_book/src/faq.md
```diff
@@ -35,20 +49,18 @@ Enable it by passing `--rest` to the run command:
...
+When a validator is started (or restarted), it listens for 2 epochs for attestations from a validator with the same public key (a doppelganger), before sending an attestation itself.
+This is a simple way of handling the case where one validator comes online with the same key as another validator that's already online, e.g. one device was started without switching the other off.
 
 While this strategy requires the client to wait two whole epochs on restart before attesting, a couple of missed attestations is a very minor price to pay in exchange for significantly reducing the risk of an accidental slashing.
-
 You can think of it as a small penalty that you pay only on first launch and restarts.
 When you take into account the total runtime of your validator, the impact should be minimal.
 
-While we strongly recommend it, you can disable it with an explicit flag (`--doppelganger-detection=false`) if you don't plan on moving your setup.
+While we strongly recommend against it, you can disable doppelganger detection with an explicit flag (`--doppelganger-detection=false`) if you don't plan on moving your setup.
```
tersec, line 61: The other hazard here is if the same machine ends up via `systemd` or similar ends up launching multiple copies of the Nimbus BN or VC. Shouldn't happen, but I'm not sure how strongly that's prohibited, and the risks are high enough that I'd not run without doppelganger detection and still depend on `systemctl restart ...`, but rather `kill` the Nimbus processes directly and let systemd detect it needs to start a new instance, just to be more certain.

docs/the_nimbus_book/src/faq.md
```diff
@@ -168,8 +208,7 @@ Please see our dedicated [guide for withdrawals](./withdrawals.md) for further i
 
 ### How are validators incentivized to stay active and honest?
 
-In addition to being penalized for being offline, validators are penalized for behaving maliciously.
-For example, attesting to invalid or contradicting blocks.
+In addition to being penalized for being offline, validators are penalized for behaving maliciously (for example, attesting to invalid or contradicting blocks).
```
tersec, line 211: I know this was already in the text, and not something you added (vs reformatting), but there are at least three different levels of "penalized" happening, in order of increasing severity/importance:
- attesting to invalid blocks -- gossip `REJECT`, might lose a peer, eventually, out of hundreds. no direct on-chain effect with regard to rewards, etc.
- being offline: on-chain penalties, albeit modest ones, by missing attestation (typically; proposer/sync committee rarer) rewards. neither of the first two cases are slashable.
- attesting to contradicting blocks (for a certain definition of "contradicting"): slashable, 1 ETH penalty or more, plus forced validator exit.

Not sure what the best approach to communicating these three levels is, but at least, it's a bit misleading to lump them all together. Losing a peer occasionally happens regardless, slightly suboptimal on-chain performance isn't great but isn't catastrophic, and slashing "should" never happen.

narimiran: In the following paragraph, it is said:

> - Minor penalties are given for inadvertent actions (or inactions) that hinder consensus.
> - And major penalties — or **slashings** — are given for malicious actions.

And then we have two sections that answer the following questions:

> How badly will a validator be penalized for being offline?
>
> How much will a validator be penalized for acting maliciously?

I thought those would be enough. Your opinion?

tersec: Oh, indeed, I agree

tersec (approved):

PR #5212, opened 2023-07-25
BN: Use Bloom filter for heavy REST validators requests.

cheatfate:


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -123,7 +129,58 @@ proc toString*(kind: ValidatorFilterKind): string =
...
+        distinctBase(forkyState.data.validators).data[index].pubkey
+      res[pubkey] = ValidatorIndex(index)
+  res
+
 proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
+  let
+    validatorsCount = max(node.getValidatorsCount(),
+                          MINIMAL_VALIDATORS_BF_COUNT)
+    # We expect 133% of validators to be registered while node is
+    # running. If number of validators will exceed this value we going
```
tersec, line 173:
```suggestion
    # running. If number of validators will exceed this value we're going
```

tersec:
In general, "Bloom" is a person's name in this context:
> A Bloom filter is a space-efficient [probabilistic](https://en.wikipedia.org/wiki/Probabilistic) [data structure](https://en.wikipedia.org/wiki/Data_structure), conceived by Burton Howard Bloom in 1970

So all the references to lowercase "bloom filters" are a bit suboptimal -- the proper name of this data structure is the "Bloom filter".

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
...
+
+func getBit*(bf: BloomFilter, pos: Natural): bool {.inline.} =
+  doAssert(pos < bf.length)
+  let index = uint(pos) shr divShift()
+  (bf.words[index] and (1'u shl (uint(pos) and modMask()))) != 0'u
+
+proc getBitsCount*(itemsCount: int): int =
+  ## We are using 8 hashes and we want to get 0.0001 false positive rate, so we
+  ## should use `m/n == 21` according to Table 3 of
+  ## https://pages.cs.wisc.edu/~cao/papers/summary-cache/node8.html
```
tersec, line 382: Alternatively, given a certain number of expected items (defined elsewhere in this PR), both optimal number of hash functions and optimal `m/n` can be found. Not sure how worthwhile -- 8 hash functions typically gets pretty good, but figure 4 ("Probability of false positives (log scale). The top curve is for 4 hash functions. The bottom curve is for the optimum (integral) number of hash functions.") does show potentially useful improvements there.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
...
+  bf.raiseBit(hashes[3])
+  bf.raiseBit(hashes[4])
+  bf.raiseBit(hashes[5])
+  bf.raiseBit(hashes[6])
+  bf.raiseBit(hashes[7])
+  inc(bf.used)
+
+func checkKey*(bf: BloomFilter, pubkey: ValidatorPubKey): bool =
+  let hashes = pubkey.toHashes()
+  if not(bf.getBit(hashes[0])): return false
```
tersec, line 471: This is not better as a loop? If it has compile-time constant bounds, I'd expect suitable unrolling to be something a C compiler might do, if useful for performance, since compilers need to be able to handle this for auto-vectorization.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
...
+              uint32.fromBytesLE(pubkey.blob.toOpenArray(44, 47))) and
+             uint32(bf.mask)
+    pos[7] = (uint32.fromBytesLE(pubkey.blob.toOpenArray(4, 7)) xor
+              uint32.fromBytesLE(pubkey.blob.toOpenArray(40, 43))) and
+             uint32(bf.mask)
+  pos
+
+proc registerKey*(bf: var BloomFilter, pubkey: ValidatorPubKey) =
+  let hashes = pubkey.toHashes()
+  bf.raiseBit(hashes[0])
```
tersec, line 459: Same comment/question about the loop.

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -123,7 +129,58 @@ proc toString*(kind: ValidatorFilterKind): string =
...
+    # to lose bloom filter's efficiency only.
+    itemsCount = validatorsCount + (validatorsCount div 3)
+    bitsCount = getBitsCount(int(itemsCount))
```
tersec, line 176: It's slightly distasteful to have to convert to `int` here in any explicit way -- functionally, this is something like `bitsCount = itemsCount * 21` but refracted through Nim's type system. But the `int(foo)` conversion is a pure artifact here, and either unnecessary to begin with, or if Nim requires it, potentially unsafe in a runtime defect way.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
...
       fork, genesis_validators_root, slot.epoch, proposer_pubkey, randao)
+
+template len*(bf: BloomFilter): int =
+  bf.used
+
+template divShift(): uint =
+  when sizeof(uint) == 8: 6'u else: 5'u
+
+template modMask(): uint =
+  when sizeof(uint) == 8: 63'u else: 31'u
```
tersec, line 367: `divShift` and `modMask` need to be consistent with each other -- wonder if it's better to explicitly define `modMask` as something similar to `(1'u shl divShift()) - 1` to make this both automatic and explicit.

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -19,6 +19,12 @@ import
 
 export rest_utils
 
+const
+  MINIMAL_VALIDATORS_BF_COUNT = 1_000_000
+    ## Minimal size of validators bloom filter. Current mainnet number of
+    ## validators is near 700k, so you could update this number if it exceeds.
```
tersec, line 25:
```suggestion
    ## validators is near 700k; this default could be increased when useful.
```

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
...
+
+func raiseBit*(bf: var BloomFilter, pos: Natural) {.inline.} =
```
tersec, line 369: use `setBit` from `stew/bitops2`?

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
...
+  doAssert(pos < bf.length)
+  let index = uint(pos) shr divShift()
+  bf.words[index] = bf.words[index] or (1'u shl (uint(pos) and modMask()))
+
+func getBit*(bf: BloomFilter, pos: Natural): bool {.inline.} =
```
tersec, line 374: Use `getBit` from `stew/bitops2`?

beacon_chain/rpc/rest_utils.nim
```diff
@@ -24,6 +26,15 @@ type
   ValidatorIndexError* {.pure.} = enum
     UnsupportedValue, TooHighValue
 
+  BloomFilter* = object
+    words: seq[uint]
+    length: int
+    mask: uint
+    used: int
```
tersec, line 33: What is the utility of `used`? It's maintained in a few places, but aside from being able to present this as a more or less normal Nim data structure, it seems to be just overhead not important for the functioning of the Bloom filter itself

beacon_chain/rpc/rest_utils.nim
```diff
@@ -334,3 +356,124 @@ proc verifyRandao*(
 
     verify_epoch_signature(
       fork, genesis_validators_root, slot.epoch, proposer_pubkey, randao)
+
+template len*(bf: BloomFilter): int =
+  bf.used
+
+template divShift(): uint =
+  when sizeof(uint) == 8: 6'u else: 5'u
```
tersec, line 364: Semi-arbitrarily asking here: all of these disparate codepaths based on 32-bit or 64-bit `uint`s seem to be predicated on there being enough of a performance (or memory usage? cache locality? etc) improvement based on not just picking one of `byte`/`uint8`, `uint32` (native on some machines, though none the BN will run well on), or `uint64` (and let 32-bit machines get suboptimal code) and using it consistently.

Debugging these issues is unfortunate, because anything 32-bit only relies on setting up a 32-bit environment first. But there doesn't seem to be a fundamental reason here not to go with `uint64` or `byte` (each has advantages; `uint32` is an awkward middle ground, optimal for nothing) as the leaf size and avoid e.g., `toHashes` doubling or tripling in size.

beacon_chain/rpc/rest_utils.nim
```diff
@@ -24,6 +26,15 @@ type
   ValidatorIndexError* {.pure.} = enum
     UnsupportedValue, TooHighValue
 
+  BloomFilter* = object
+    words: seq[uint]
+    length: int
```
tersec, line 31: `length` is used in a couple of asserts, and there's a `bitsCount` function which returns it. But in general, the `words` `seq` has all the length information the Bloom filter functionally needs, so like `used`, this is more overhead, both code and memory.

PR #5216, opened 2023-07-27
Rename "data gas" to "blob gas"

henridf:
Per https://github.com/ethereum/consensus-specs/pull/3461

Once https://github.com/status-im/nim-web3/pull/91 lands, I will update the pointer here and un-draft the PR.


tersec:
Is this supported by ELs yet, or? These are generally most specifically useful when either they're required for an upcoming devnet or they show up in a consensus-spec/execution-api/etc release.

henridf:
> Is this supported by ELs yet, or? These are generally most specifically useful when either they're required for an upcoming devnet or they show up in a consensus-spec/execution-api/etc release.

EL support is in progress (from a quick look at the spec PR, I see erigon, nim-eth PR links to it). 

Also this is slated for devnet8.

tersec:
This shouldn't intrinsically break/interact with the consensus spec tests, so should be possible to update separately as required, though by then beta.1 probably will be out anyway, with its new tests

PR #5223, opened 2023-07-31
Revert "generalize `ShufflingRef` acceleration logic"

tersec:
Reverts status-im/nimbus-eth2#5197


tersec:
One round of Jenkins across all platforms went well -- restarted to try a second round. If this works, that's enough evidence to merge, because it's blocking all (except https://github.com/status-im/nimbus-eth2/pull/5212 which is interesting, but exception proving the rule, etc) the current PRs from being green.

It's probably the culprit, and if it doesn't fix things, well, revert the revert.

PR #5226, opened 2023-07-31
more log lines when jobs fail

etan-status:
Increase log lines emitted on test failure to 1000.


tersec (approved):

arnetheduck:
This is good for a quickie, but in reality one would need to write output to file and make it part of the artefacts of the test run

PR #5236, opened 2023-08-02
trivial whitespace change

tersec:
From before the known `ListKeys` issues became prominent, from v23.7.0 stable.

Not for merging, but as a way to distinguish code changes from environmental changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+
+
+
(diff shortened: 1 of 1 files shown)


tersec:
https://github.com/status-im/nimbus-eth2/pull/5255 is from an even older commit, and shows the same issue

PR #5238, opened 2023-08-03
switch from `create`/`dealloc` to `new+GC_ref`/`GC_unref`

etan-status:
Using `create` with objects containing managed objects is broken: https://github.com/nim-lang/Nim/issues/22341

Switch to a safer pattern based on `new+GC_ref`/`GC_unref` instead.


etan-status:
`-d:useGcAssert -d:useSysAssert` to highlight the problems with `create`/`dealloc`.

PR #5245, opened 2023-08-03
remove less-than-useful warning

tersec:


beacon_chain/validators/validator_duties.nim
```diff
@@ -818,10 +818,7 @@ proc proposeBlockAux(
 
   let payloadBuilderClientMaybe = node.getPayloadBuilderClient(
     validator_index.distinctBase)
-  if payloadBuilderClientMaybe.isErr:
```
arnetheduck, line 821: the warning seems legit in general though, specially if new failure modes are added in the future .. overall, seems better not to try to fetch a builder client at all when `node.config.payloadBuilderEnable == false`?

tersec: The general intent was/is to centralize both config and error checks in
https://github.com/status-im/nimbus-eth2/blob/60f0a2f6a662abb9c8332d9f0d3eb7cc998fab63/beacon_chain/beacon_node.nim#L125-L149

and not to repeat them elsewhere, in an ad hoc way. The result is a function which has the only typical "error" return values things which aren't worth logging.

This function can return one of
```nim
  if not node.config.payloadBuilderEnable:
    return err "Payload builder globally disabled"
```

```nim
      if validator_index >= forkyState.data.validators.lenu64:
        return err "Validator index too high"
```

```nim
  if payloadBuilderAddress.isNone:
    return err "Payload builder disabled"
```

```nim
  if res.isOk and res.get.isNil:
    err "Got nil payload builder REST client reference"
```

Or, otherwise, the desired REST client reference.

Of those 4 errors, two of them are simply conveying the configuration.

Validator index too high is one of the "should never" happen versions; generally, if it's going to trigger there, it will elsewhere in those codepaths too, so it won't be a lost error. It's here also for typesafety, primarily.

The nil reference I'm not sure how common actually is. It's the only one I'd consider that maybe should be logged, but if so, maybe should be logged here, and still drop logging in callers. It's not really a per-caller property regardless why `let res = RestClientRef.new(payloadBuilderAddress.get)` might have failed.

So, the whole point here is treat the builder-client-fetch returning an error as just another way of communicating the combined results of the config settings, especially since there are multiple (e.g., it might be disabled per-validator or it might be disabled globally) and used in two places, potentially more (validator registration and proposing).

arnetheduck: well - one way or the other then? ie either have a function that does "build-enabled" check and any other relevant checking or do it at call sites, but why both? why recheck builder-enabled here?

tersec: That's already the case. The callers don't check otherwise. It all lives in this function, consistently.

https://github.com/status-im/nimbus-eth2/blob/60f0a2f6a662abb9c8332d9f0d3eb7cc998fab63/beacon_chain/rpc/rest_beacon_api.nim#L917-L927

https://github.com/status-im/nimbus-eth2/blob/60f0a2f6a662abb9c8332d9f0d3eb7cc998fab63/beacon_chain/rpc/rest_beacon_api.nim#L937-L948

https://github.com/status-im/nimbus-eth2/blob/60f0a2f6a662abb9c8332d9f0d3eb7cc998fab63/beacon_chain/rpc/rest_validator_api.nim#L541-L576

arnetheduck: https://github.com/status-im/nimbus-eth2/pull/5245/files#diff-3b66e2206a732f702b9a10ac7b8bc7c67c4ab8cd54aacea0902895a7e53bbecbL828 - there's a check for `node.config.payloadbuilderEnable just below?

tersec: https://github.com/status-im/nimbus-eth2/pull/5245/commits/0db75415d0d3c85c4fb78771ef8b4a4e312b3b57

PR #5247, opened 2023-08-04
note summary data about invalid baked-in states

tersec:
This way, it can be compared with known-good information to detect potential miscompiles or disk corruption.


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -550,7 +550,9 @@ proc init*(T: type BeaconNode,
       try:
         newClone readSszForkedHashedBeaconState(cfg, metadata.genesisBytes)
       except CatchableError as err:
-        raiseAssert "Invalid baked-in state: " & err.msg
+        raiseAssert "Invalid baked-in state of length " &
```
zah, line 553: Have this actually happened for someone? This would be quite concerning.

tersec: yes

PR #5252, opened 2023-08-04
replace copy with symlink

etan-status:
One of the test files is accidentally a copy instead of a symlink.


arnetheduck:
why are we even storing these? ie they could be generated on the fly, no?

etan-status:
Two reasons:

- Using known-good mainnet data avoids the possibility of testing against a buggy generator
- Also adding some execution layer structures in followups for which we don't have nice generators ready

The only really big file was the genesis.ssz files that accidentally became a copy instead of a `ln -s` after too many local force pushes and rebases.

PR #5253, opened 2023-08-04
Revert "correctly align genesis size"

tersec:
Reverts status-im/nimbus-eth2#5244

breaks `unstable`

PR #5255, opened 2023-08-04
`346d05a9`+ unconditionally enable UBSAN

tersec:
based on https://github.com/status-im/nimbus-eth2/commit/346d05a95dbb622a82021c97b5cd407b15c9c867 + unconditionally enable UBSAN

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -129,11 +129,7 @@ jobs:
 
           # Stack usage test and UBSAN on recent enough gcc:
           if [[ '${{ runner.os }}' == 'Linux' && '${{ matrix.target.cpu }}' == 'amd64' ]]; then
-            if [[ '${{ github.sha }}' =~ ^7 ]]; then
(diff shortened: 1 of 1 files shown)


arnetheduck:
what about running ubsan in a separate parallel test?

tersec:
In this case, this was a one-off I'm mostly keeping open because it also demonstrates the earliest tested extent of the `ListKeys` issue: https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/aarch64/job/PR-5255/1/display/redirect

tersec:
The concern I have about running it in a parallel test, more broadly, is that it's too easily ignored, the way it was this time.

https://github.com/status-im/nimbus-eth2/pull/5253#issuecomment-1666165087 explains this reasoning a bit more -- but the nightlies were broken for months, for example, exactly because they didn't really block anything and didn't show up in any obviously wrong way in normal development.

I think one approach might be a combination of more-asynchronous (e.g., `unstable` only, not in PRs where people are waiting) but also more frequent (1/8 probability rather than 1/16 probability).

etan-status:
https://github.com/status-im/nimbus-eth2/pull/5258

PR #5257, opened 2023-08-05
stop building `test_libnimbus_lc` and force UBSAN

tersec:
Mainly, don't merge because it's probably too much to run UBSAN every commit in CI.

But also, there might be a better approach to the `test_libnimbus_lc` removal (e.g., fixing instead). This exists to verify in CI what
```
make NIMFLAGS="--passC:-fsanitize=undefined --passL:-fsanitize=undefined" test_libnimbus_lc
```
shows locally, i.e. that it triggers some UBSAN linking error on that specific target which removing that target removes, that no other target has this issue.


tersec:
Addressed by https://github.com/status-im/nimbus-eth2/pull/5258

PR #5268, opened 2023-08-07
speed up LC signature verification

etan-status:
We know the aggregate publickey of a fully participating sync committee. Because participation is typically very high (>95%), it is faster to start from that aggregate publickey and subtract the individual keys of non-participants, than summing up all the participating pubkeys.


etan-status:
https://github.com/status-im/nim-blscurve/pull/157

etan-status:
https://github.com/status-im/nimbus-eth2/pull/5274

tersec (approved):

PR #5286, opened 2023-08-12
avoid invalidating validators htr cache within epoch

tersec:
In general, `validators` doesn't change between `process_epoch()` calls.

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -1121,6 +1121,23 @@ func toBeaconStateSummary(state: ForkyBeaconState): BeaconStateSummary =
     latest_block_root: state.block_roots[root_idx],
     latest_state_root: state.state_roots[root_idx])
 
+func validatorsPotentiallyChanged(
+    state0Summary: BeaconStateSummary, state0EpochAligned: bool,
+    state1: ForkyBeaconState): bool =
+  # state.block_roots and state.state_roots are set in process_slot(), which
+  # runs before process_epoch() in process_slots(). The epoch boundary slot,
+  # therefore, gets {block,state}_roots entries which do not reflect changes
+  # in the validator set done at epoch boundaries.
+  #
+  # state.latest_block_header gets set, instead, in process_block(), after
+  # process_epoch(), immediately updated each epoch boundary slot. This is
+  # only a partial update, excluding state.latest_block_header.state_root,
+  # so isn't generally consistent or comparable with state.block_roots.
+  #
+  # Therefore, exclude cases where either state is on an epoch-aligned slot.
+  state0EpochAligned or state1.slot.is_epoch or
+    state1.toBeaconStateSummary != state0Summary
+
 proc getStateOnlyMutableValidators(
     immutableValidators: openArray[ImmutableValidatorData2],
     store: KvStoreRef, key: openArray[byte],
@@ -1135,7 +1152,9 @@ proc getStateOnlyMutableValidators(
   # TODO rollback is needed to deal with bug - use `noRollback` to ignore:
   #      https://github.com/nim-lang/Nim/issues/14126
 
-  let prevStateSummary = output.toBeaconStateSummary
+  let
+    prevStateSummary = output.toBeaconStateSummary
+    prevStateEpochAligned = output.slot.is_epoch
 
   case store.getSnappySSZ(key, toBeaconStateNoImmutableValidators(output))
   of GetResult.found:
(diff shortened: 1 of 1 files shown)


beacon_chain/beacon_chain_db.nim
```diff
@@ -1103,6 +1109,16 @@ proc getBlockSZ*(
...
     getBlockSZ(db, key, data, deneb.TrustedSignedBeaconBlock)
 
+func toBeaconStateSummary(state: ForkyBeaconState): BeaconStateSummary =
+  # Two states with identical BeaconStateSummary fields have the same
+  # hash_tree_root(state.validators).
+  let slot = state.slot
+  BeaconStateSummary(
+    epoch: slot.epoch,
+    num_validators: state.validators.len,
+    latest_block_root: state.block_roots[slot mod SLOTS_PER_HISTORICAL_ROOT],
```
arnetheduck: doesn't this change for .. every slot?

arnetheduck: ie I imagine the value to copy here should be that of `epoch.start_slot`?

arnetheduck: also, I think this might have boundary problems on slot 0 of the epoch

tersec: > ie I imagine the value to copy here should be that of `epoch.start_slot`?

Yes, https://github.com/status-im/nimbus-eth2/pull/5286/commits/fbe144e47d0fde8e660a70f3cf8861de4fa82ed6

tersec: > also, I think this might have boundary problems on slot 0 of the epoch

One approach, which I initially used, was to just block out the epoch boundary from consideration.

[`process_slot()` per spec](https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/beacon-chain.md#beacon-chain-state-transition-function) does
```python
def process_slot(state: BeaconState) -> None:
    # Cache state root
    previous_state_root = hash_tree_root(state)
    state.state_roots[state.slot % SLOTS_PER_HISTORICAL_ROOT] = previous_state_root
    # Cache latest block header state root
    if state.latest_block_header.state_root == Bytes32():
        state.latest_block_header.state_root = previous_state_root
    # Cache block root
    previous_block_root = hash_tree_root(state.latest_block_header)
    state.block_roots[state.slot % SLOTS_PER_HISTORICAL_ROOT] = previous_block_root
```

But it's called before `process_block()`, where divergence could occur:
```python
def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, validate_result: bool=True) -> None:
    block = signed_block.message
    # Process slots (including those with no blocks) since block
    process_slots(state, block.slot)
    # Verify signature
    if validate_result:
        assert verify_block_signature(state, signed_block)
    # Process block
    process_block(state, block)
    # Verify state root
    if validate_result:
        assert block.state_root == hash_tree_root(state)
```
in this case, though, that's an acceptable incompleteness, which only has to be internally consistent: the changes which matter here happen in `process_epoch()`.

Unfortunately, the relevant `process_slot()` also happens before `process_epoch()`:
```python
def process_slots(state: BeaconState, slot: Slot) -> None:
    assert state.slot < slot
    while state.slot < slot:
        process_slot(state)
        # Process epoch on the start slot of the next epoch
        if (state.slot + 1) % SLOTS_PER_EPOCH == 0:
            process_epoch(state)
        state.slot = Slot(state.slot + 1)
```

So, yes, there's a boundary issue here. `latest_block_header` gets set after `process_epoch`, so does not have this problem, and can cover that additional case. There might not be a block, but then there can't be a reorg, either.

tersec: > also, I think this might have boundary problems on slot 0 of the epoch

https://github.com/status-im/nimbus-eth2/pull/5286/commits/bf14fc5a23bb5bd2170bb5675b89bdc5deb35184

arnetheduck:
> In general, validators doesn't change

proposer slashings?

edit: actually, any slashings?

tersec:
> > In general, validators doesn't change
> 
> proposer slashings?
> 
> edit: actually, any slashings?

Indeed. `process_slashings()` is part of `process_epoch()`, but `slash_validator()` happens as part of block processing.

Possible to track, in principle (since beginning of epoch, have any reorgs occured + have any slashings occurred), but at that point this design becomes a bit brittle/fragile anyway, as well as requiring some abstraction layer inversions/violations to properly work rather than being a kind of peephole optimization in the database layer.

PR #5302, opened 2023-08-17
refactor BN engine/builder block selection to work as REST block v3 backend

tersec:
`collectBidFutures` knows how to query both engine API and builder API simultaneously and collect the results.

`builderBetterBid` is separately callable and uses the local block value boost properly.


tersec:
kill this one, it's not worth rebasing, if it's worthwhile, it's easier to make anew

PR #5142, opened 2023-06-28
update tested Nim versions

etan-status:
Add `version-2-0` testing, and make upstream version testing Linux only.

PR #5308, opened 2023-08-17
cleanup `{.raises.}` for `writeValue`

etan-status:
- Remove unnecessary `Defect` references
- Remove spurious `SerializationError` references
- Remove duplicate `writeValue` template in `keystore.nim`; same implementation already exists a bit further above in same file.

### DIFF
--- a/beacon_chain/el/el_conf.nim
+++ b/beacon_chain/el/el_conf.nim
@@ -1,3 +1,10 @@
+# beacon_chain
+# Copyright (c) 2023 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at http://opensource.org/licenses/MIT).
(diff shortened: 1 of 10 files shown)


arnetheduck:
same as here: https://github.com/status-im/nim-ssz-serialization/pull/61#discussion_r1297208229

etan-status:
The `SerializationError` discussion in the linked comment applies more to the library instead of Nimbus the application. 

In Nimbus we don't have types that sometimes fail to serialize, and we also don't have usage where we try to send objects around that can't be serialized.

As the presence of `{.raises: [SerializationError].}` on `EngineApiRoles`, `Eth2Digest` and ` KeystoresAndSlashingProtection` ultimately trips off Nim 2.0 compilation, will remove them for now to unblock https://github.com/status-im/nimbus-eth2/pull/5142

If we later decide that `nim-serialization` should be extended with support for serialization failures, that should be implemented more broadly than just on a handful of arbitrary types.

PR #5310, opened 2023-08-17
remove unnecessary `mixin`

etan-status:
Nim 2.0 gets confused with the `mixin self, id` in `validator_monitor`. It looks like this is actually not needed, so removing it.

### DIFF
--- a/beacon_chain/validators/validator_monitor.nim
+++ b/beacon_chain/validators/validator_monitor.nim
@@ -236,7 +236,6 @@ proc addMonitor*(
     self.indices[index.get().uint64] = monitor
 
 template metricId: string =
-  mixin self, id
   if self.totals: total else: id
 
 proc addAutoMonitor*(


etan-status:
```
/home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/consensus_object_pools/blockchain_dag.nim(923, 16) template/generic instantiation of `withState` from here
/home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/consensus_object_pools/blockchain_dag.nim(926, 33) template/generic instantiation of `registerEpochInfo` from here
/home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/validators/validator_monitor.nim(420, 16) template/generic instantiation of `withEpochInfo` from here
/home/runner/work/nimbus-eth2/nimbus-eth2/beacon_chain/validators/validator_monitor.nim(240, 10) Error: undeclared field: 'totals'
make: *** [Makefile:442: mev_mock] Error 1
```

arnetheduck:
I always randomly try to add mixin:s when something doesn't work in Nim to @zah's delight, but removing them is a new one ;)

PR #5313, opened 2023-08-18
attest to known valid block when possible

tersec:


arnetheduck:
Per discussion, let's hold off on this PR until we've investigated the possibility to quarantine blocks like this which has the advantage of not polluting the internal state with optimistic information resulting in a more clean break

tersec:
> Per discussion, let's hold off on this PR until we've investigated the possibility to quarantine blocks like this which has the advantage of not polluting the internal state with optimistic information resulting in a more clean break

I'm not sure a quarantine to avoid polluting the internal state is feasible, because while one can deal with qurantines when only `newPayload` and not `forkchoiceUpdated` has to be called, ELs won't necessarily return `VALID` on just `newPayload` (especially during reorgs, but even in general), so this has to check `forkchoiceUpdated` first. 

There are two main issues with this, one sort of incidental to Nimbus (though requiring a substantial refactor to work around) and one a bit more fundamental to how the engine API works: (1) semi-incidentally, in `storeBlock` (but, obviously, the exact function name, factoring, etc doesn't matter), before it adds the block to the DAG, it can't readily use `forkchoiceUpdated`.

Could be addressed, but because (2) more fundamentally, `forkchoiceUpdated` functions as a quasi-atomic check-and-commit operation to a head/safe/finalized (EL) block, where if the EL says yes, the CL should be prepared to follow along. If the EL responds `VALID`, this is not an issue. If the `EL` responds `INVALID` or `INVALID_BLOCK_HASH`, it's in a different way, not an issue (or, already handled as part of usual optimistic sync). If it's `SYNCING` or `ACCEPTED`, though, the EL will now expect the CL to keep going along with the head it's requested (as commented many times, the engine API and the opt sync spec are built around each other, it's not some coincidence). This means not having the blocks too separated into any quarantine.

arnetheduck:
does this also cover the rest api? ie when node is asked to produce attestation data, does it respond with a latest-valid-based view?

tersec:
> does this also cover the rest api? ie when node is asked to produce attestation data, does it respond with a latest-valid-based view?

https://github.com/status-im/nimbus-eth2/pull/5313/commits/a1887e58e3880dafa78dbfc680d2511ab57d38cc

arnetheduck (approved):

PR #5315, opened 2023-08-18
Single thread minimal

arnetheduck:
This increases the end-to-end testing for single-threaded mode which
otherwise might run into problems like
https://github.com/status-im/nimbus-eth2/pull/5303


tersec (approved):

arnetheduck:
lint issue is false alarm

PR #5333, opened 2023-08-20
`validator_duties`->`beacon_validators`

arnetheduck:
This PR renames the existing `validator_duties` to `beacon_validators` and in doing so, names validators running inside the beacon node process "beacon validators" while those running the VC can be referred to as "client validators" to disambiguate the two.

The existing `validator_duties` instead takes on a new responsibility: as a home for logic shared between beacon and client validators - ie code that provides consistency in implementation and behavior between the two modes of operation.

Not only does this simplify reasoning about where to put code -it also reduces the number of dependencies the validator client has from ~5000 to ~3000 modules (!) according to `nim genDepend` significantly reducing compile times.


zah:
Quite nice, this also addresses the metrics pollution problem that many of our users have noticed (beacon node metrics being served by the validator client).

PR #5337, opened 2023-08-23
Add metadata for the Holesky network

zah:


beacon_chain/networking/network_metadata_downloads.nim
```diff
@@ -0,0 +1,64 @@
...
+      status: response[0])
+
+proc fetchBytes*(metadata: GenesisMetadata): Future[seq[byte]] {.async.} =
+  case metadata.kind
+  of NoGenesis:
+    raiseAssert "fetchBytes should be called only when metadata.hasGenesis is true"
+  of BakedIn:
+    @(metadata.bakedBytes)
+  of BakedInUrl:
+    decodeFramed(await downloadFile(parseUri metadata.url))
```
arnetheduck: when fetching from url, we need to validate the ssz root of the genesis else we can be fed anything - also, when doing TNS, we should be attempting to fetch it from the TNS source

beacon_chain/networking/network_metadata.nim
```diff
@@ -242,38 +277,62 @@ when const_preset == "gnosis":
...
-      vendorDir & "/eth2-networks/shared/prater", some goerli, not incbinEnabled)
+      vendorDir & "/eth2-networks/shared/prater",
+      some goerli,
+      useBakedInGenesis = some "prater")
+
+    holeskyMetadata = loadCompileTimeNetworkMetadata(
+      vendorDir & "/holesky/custom_config_data",
+      some holesky,
+      downloadGenesisFrom = some DownloadInfo(
+        url: "https://github.com/status-im/nimbus-eth2/releases/download/v23.8.0/holesky-genesis.ssz.snappy-framed",
```
arnetheduck, line 323: the extension for framed snappy is generally `.sz`: https://github.com/google/snappy/blob/main/framing_format.txt

PR #5346, opened 2023-08-24
NimYAML: bump to 1.1.0

arnetheduck:
We were using a status-based fork with additional Nim 1.2 support - should not needed anymore so this PR bumps it to a clean upstream version


arnetheduck:
lint failure is bogus

Issue #5362, opened 2023-08-26
Failed to deserialize REST JSON data in `deposits exit` call


heeckhau:
I ran into this error while trying to exit a Goerli  validator :
```
(FAT 2023-08-25 07:21:26.306+00:00 Failed to obtain the config spec of the beacon node reason="Unable to decode REST response, error [Serialization error]")
```

With debug mode enabled, I got the attached output:
`/home/user/nimbus-eth2/build/nimbus_beacon_node deposits exit --validator=546123 --print --log-level=DEBUG`
```
Failed to deserialize REST JSON data       err="<data>(1, 2889) Non-decimal character encountered: 0x00" data="{\"data\":{\...
```
Full log:
```
DBG 2023-08-25 07:21:26.305+00:00 Sending REST request to remote server      remote=127.0.0.1:5052 request=/eth/v1/beacon/genesis http_method=GET
DBG 2023-08-25 07:21:26.305+00:00 Got REST response headers from remote server remote=127.0.0.1:5052 request=/eth/v1/beacon/genesis status=200 http_method=GET
DBG 2023-08-25 07:21:26.305+00:00 Received REST response body from remote server remote=127.0.0.1:5052 request=/eth/v1/beacon/genesis contentType=application/json size=169
DBG 2023-08-25 07:21:26.305+00:00 Sending REST request to remote server      remote=127.0.0.1:5052 request=/eth/v1/beacon/states/head/fork http_method=GET
DBG 2023-08-25 07:21:26.306+00:00 Got REST response headers from remote server remote=127.0.0.1:5052 request=/eth/v1/beacon/states/head/fork status=200 http_method=GET
DBG 2023-08-25 07:21:26.306+00:00 Received REST response body from remote server remote=127.0.0.1:5052 request=/eth/v1/beacon/states/head/fork contentType=application/json size=119
DBG 2023-08-25 07:21:26.306+00:00 Sending REST request to remote server      remote=127.0.0.1:5052 request=/eth/v1/config/spec http_method=GET
DBG 2023-08-25 07:21:26.306+00:00 Got REST response headers from remote server remote=127.0.0.1:5052 request=/eth/v1/config/spec status=200 http_method=GET
DBG 2023-08-25 07:21:26.306+00:00 Received REST response body from remote server remote=127.0.0.1:5052 request=/eth/v1/config/spec contentType=application/json size=3859
[...]

PR #5370, opened 2023-08-29
Address issue #5362.

cheatfate:


beacon_chain/deposits.nim
```diff
@@ -220,15 +220,26 @@ proc restValidatorExit(config: BeaconNodeConf) {.async.} =
     quit 1
 
   let signingFork = try:
-    let response = await client.getSpec()
```
tersec, line 223: Is `getSpec` used anywhere else? If not, might just be better to remove it and the entire `RestSpec` object along with the `GetSpecResponse` wrapper for that. Otherwise there are two semi-duplicative objects which need to track the `GetSpec` REST endpoint. This decode-later approach is probably better given the semi-freeform, or at least more flexible, version there.

beacon_chain/deposits.nim
```diff
@@ -220,15 +220,26 @@ proc restValidatorExit(config: BeaconNodeConf) {.async.} =
...
+            let s = spec.getOrDefault("DENEB_FORK_EPOCH", $FAR_FUTURE_EPOCH)
+            Epoch(Base10.decode(uint64, s).get(uint64(FAR_FUTURE_EPOCH)))
       # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/beacon-chain.md#voluntary-exits
       # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/deneb/beacon-chain.md#modified-process_voluntary_exit
-      if currentEpoch >= Epoch(spec.data.DENEB_FORK_EPOCH):
+      if currentEpoch >= denebForkEpoch:
+        let capellaForkVersion =
+          block:
+            var res: Version
+            let s = spec.getOrDefault("CAPELLA_FORK_VERSION", "")
```
zah, line 240: Using this default value doesn't seem appropriate, as there is no such treatment prescribed in the specs. If the field is missing, this should probably be interpreted as an error.

cheatfate: CAPELLA_FORK_VERSION should be a string with correct format e.g. "0x01000000". So if default is an empty string it will fail in `hexToByteArrayStrict(s, distinctBase(res))` which will only decode 4 bytes sequence and will raise an exception if string size is incorrect.

beacon_chain/deposits.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5370/commits/a96778ed4ac8dc470f76ca25ad98b4a67544399c does this

beacon_chain/deposits.nim
tersec: Maybe worth documenting/commenting the intent as such in the code, then?

cheatfate: Is it enough?

tersec: Regarding https://github.com/status-im/nimbus-eth2/pull/5370/commits/7a7c83c455f0d8c85768dbce078a7348303f0435 I'd say so, yes.

PR #5371, opened 2023-08-30
use correct decimal or hex encoding/decoding for REST GetSpec

tersec:
https://github.com/status-im/nimbus-eth2/issues/5362

https://github.com/ethereum/beacon-APIs/blob/v2.4.2/apis/config/spec.yaml describes that
```
                  Values are returned with following format:
                    - any value starting with 0x in the spec is returned as a hex string
                    - numeric values are returned as a quoted integer
```


tersec:
See also https://github.com/status-im/nimbus-eth2/pull/5370, a slightly different take on the same idea

PR #5387, opened 2023-09-04
add infrastructure to select fork choice version

etan-status:
To allow testing https://github.com/ethereum/consensus-specs/issues/3466 add support for selecting fork choice version at launch. This means we can deploy a different logic when `DENEB_FORK_EPOCH != FAR_FUTURE_EPOCH` that won't be used on Mainnet.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -29,6 +29,8 @@ import
   ./el/el_conf,
   ./filepath
 
+from fork_choice/fork_choice_types
+  import ForkChoiceVersion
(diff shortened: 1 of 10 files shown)


etan-status:
Also adds

```json
  "extra_data": {
    "version": "stable"
  }
```

to `http://127.0.0.1:5052/eth/v1/debug/fork_choice` so that explorer tools can distinguish selected version

`extra_data` is an implementation-specific dictionary.

arnetheduck (approved):

PR #5388, opened 2023-09-04
VC: Fix VC must not crash if beacon node address could not be resolved.

cheatfate:


beacon_chain/nimbus_validator_client.nim
```diff
@@ -19,8 +19,21 @@ proc initGenesis(vc: ValidatorClientRef): Future[RestGenesis] {.async.} =
...
+    if len(offlineNodes) == 0:
+      let sleepDuration = 2.seconds
+      info "Could not resolve beacon nodes, repeating",
+           sleep_time = sleepDuration
+      await sleepAsync(sleepDuration)
+      for node in vc.nonameNodes():
+        let status = checkName(node)
+        node.updateStatus(status, ApiNodeFailure())
+        if status == RestBeaconNodeStatus.Noname:
+          warn "Cannot initialize beacon node", node = node, status = status
```
arnetheduck, line 32: will this be printed every 2 seconds on a node without name?

cheatfate: yup, but only in situation when there no any node with name available, if there will be at least one node with name available this string will not be printed. Its initial check before we obtaining genesis. So it will print this message every 2 seconds for every beacon node in list.

beacon_chain/nimbus_validator_client.nim
```diff
@@ -282,7 +296,10 @@ proc asyncInit(vc: ValidatorClientRef): Future[ValidatorClientRef] {.async.} =
                                        beacon_nodes_count = len(vc.beaconNodes)
 
   for node in vc.beaconNodes:
-    notice "Beacon node initialized", node = node
+    if node.status == RestBeaconNodeStatus.Offline:
```
arnetheduck, line 313: offline == initialized?

arnetheduck (approved):

PR #5389, opened 2023-09-05
Rename --validator-source to --web3-signer-url and document it

zah:
The user can also configure multiple instances now.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -162,14 +162,14 @@ type
       desc: "A directory containing validator keystores"
       name: "validators-dir" .}: Option[InputDir]
 
-    validatorsSource* {.
+    web3signers* {.
       desc: "Remote Web3Signer URL that will be used as a source of validators"
-      name: "validators-source"}: Option[string]
+      name: "web3-signer-url" .}: seq[Uri]
 
-    validatorsSourceInverval* {.
-      desc: "Number of minutes between validator list updates"
-      name: "validators-source-interval"
-      defaultValue: 60 .}: Natural
+    web3signerUpdateInterval* {.
+      desc: "Number of seconds between validator list updates"
+      name: "web3-signer-update-interval"
+      defaultValue: 3600 .}: Natural
 
     secretsDirFlag* {.
       desc: "A directory containing validator keystore passwords"
@@ -885,14 +885,14 @@ type
       desc: "A directory containing validator keystores"
       name: "validators-dir" .}: Option[InputDir]
 
-    validatorsSource* {.
+    web3signers* {.
       desc: "Remote Web3Signer URL that will be used as a source of validators"
-      name: "validators-source"}: Option[string]
+      name: "web3-signer-url" .}: seq[Uri]
 
-    validatorsSourceInverval* {.
-      desc: "Number of minutes between validator list updates"
-      name: "validators-source-interval"
-      defaultValue: 60 .}: Natural
+    web3signerUpdateInterval* {.
+      desc: "Number of seconds between validator list updates"
+      name: "web3-signer-update-interval"
+      defaultValue: 3600 .}: Natural
 
     secretsDirFlag* {.
       desc: "A directory containing validator keystore passwords"
--- a/beacon_chain/nimbus_validator_client.nim
+++ b/beacon_chain/nimbus_validator_client.nim
@@ -90,11 +90,12 @@ proc initValidators(vc: ValidatorClientRef): Future[bool] {.async.} =
   var duplicates: seq[ValidatorPubKey]
   for keystore in listLoadableKeystores(vc.config, vc.keystoreCache):
     vc.addValidator(keystore)
-  let res = await queryValidatorsSource(vc.config)
-  if res.isOk():
-    let dynamicKeystores = res.get()
-    for keystore in dynamicKeystores:
-      vc.addValidator(keystore)
+  for web3signerUrl in vc.config.web3signers:
+    let res = await queryValidatorsSource(web3signerUrl)
+    if res.isOk():
+      let dynamicKeystores = res.get()
+      for keystore in dynamicKeystores:
+        vc.addValidator(keystore)
   true
 
 proc initClock(vc: ValidatorClientRef): Future[BeaconClock] {.async.} =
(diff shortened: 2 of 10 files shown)


beacon_chain/conf.nim
```diff
@@ -162,14 +162,14 @@ type
       desc: "A directory containing validator keystores"
       name: "validators-dir" .}: Option[InputDir]
 
-    validatorsSource* {.
+    web3signers* {.
       desc: "Remote Web3Signer URL that will be used as a source of validators"
-      name: "validators-source"}: Option[string]
+      name: "web3-signer-url" .}: seq[Uri]
```
cheatfate, line 167: This is very confusing option now, `web3-signer-url` could be URL used for signatures only, but now it manages validators you are running...

beacon_chain/nimbus_validator_client.nim
```diff
@@ -90,11 +90,12 @@ proc initValidators(vc: ValidatorClientRef): Future[bool] {.async.} =
   var duplicates: seq[ValidatorPubKey]
   for keystore in listLoadableKeystores(vc.config, vc.keystoreCache):
     vc.addValidator(keystore)
-  let res = await queryValidatorsSource(vc.config)
-  if res.isOk():
-    let dynamicKeystores = res.get()
-    for keystore in dynamicKeystores:
-      vc.addValidator(keystore)
+  for web3signerUrl in vc.config.web3signers:
+    let res = await queryValidatorsSource(web3signerUrl)
```
cheatfate: Is there any reason why we should query web3-signer nodes one by one? Why we should not query all of them concurrently?

cheatfate:
This PR will not work for multiple signer nodes as you expect, you should fix `updateDynamicValidators`, it is not working as aggregator. So if last queried web3-signer returns empty sequence, all the dynamic validators will be deleted and no validators will be working for whole period. So first you should query validators, combine all the responses into one list and only after that you should call `updateDynamicValidators`.

beacon_chain/nimbus_validator_client.nim
arnetheduck: also, for large setups, this seems rather expensive - if it coincides with other activity such as attestation production, it might disrupt it - it should likely be scheduled at a time of low activity, ie in the last seconds of a slot in the middle of an epoch

beacon_chain/validators/beacon_validators.nim
```diff
@@ -166,8 +145,42 @@ proc addValidators*(node: BeaconNode) =
...
+      feeRecipient = node.consensusManager[].getFeeRecipient(
+        keystore.pubkey, index, epoch)
+      gasLimit = node.consensusManager[].getGasLimit(keystore.pubkey)
+
+      v = node.attachedValidators[].addValidator(keystore, feeRecipient,
+                                                 gasLimit)
+    v.updateValidator(data)
+
+  try:
+    waitFor allFutures(mapIt(node.config.web3signers,
```
arnetheduck, line 175: allFutures does not raise on error, you need to investigate each future individually - any exceptions raised will be ignored / discarded

arnetheduck: also, the `waitFor` is potentially problematic if this function runs in a context of another loop - not sure we have a good way of preventing that - it would be better to make the function async and eliminate this risk

beacon_chain/conf.nim
zah: Since the capability to obtain the validators dynamically is part of the Web3Signer standard (and it's not marked as optional functionality), I think it's reasonable for users to expect it when they provide a Web3Signer URL. 

This also makes documenting the functionality simpler as evident by the additions to the Nimbus guide. Users who don't need the automatic behavior are referred to the more manual process of creating remote keystores.

beacon_chain/nimbus_validator_client.nim
zah: This is an one-time operation at start-up. The current code is unlikely to be a serious problem, although indeed running the queries in parallel would be better.

beacon_chain/validators/beacon_validators.nim
zah: This code is a simple refactoring of the previous version which was also using `waitFor` here (querying a single source). The use of `allFutures` was intentional to some extent because `addValidatorsFromWeb3Signer` handles all possible exceptions (having additional sanity checks here doesn't hurt of course).

zah:
@cheatfate, did you notice this change in `updateDynamicValidators`?
https://github.com/status-im/nimbus-eth2/pull/5389/files#diff-29d0ce8e5a9dd42e1b99e189027992a6b432f486fbea761b8193369628f17cb0R403

Basically, I've modified the function to operate on the dynamic set of a single Web3Signer. In your example, when the last signer returns an empty list, the function will delete only the previous validators obtained from this particular signer.

beacon_chain/nimbus_validator_client.nim
```diff
@@ -85,17 +85,28 @@ proc initGenesis(vc: ValidatorClientRef): Future[RestGenesis] {.async.} =
...
-      for keystore in dynamicKeystores:
-        vc.addValidator(keystore)
+
+  let web3signerValidatorsFuts = mapIt(
+    vc.config.web3signers,
+    vc.addValidatorsFromWeb3Signer(it))
+
+  # We use `allFutures` because all failures are already reported as
+  # user-visible warnings in `queryValidatorsSource`.
+  # We don't consider them fatal because the Web3Signer may be experiencing
```
arnetheduck, line 106: will such failures be logged anywhere?

beacon_chain/validators/beacon_validators.nim
arnetheduck: well, the sanity checks in this case also swallow cancellation (and imply that error handling is written in a schematic rather than deliberate way) - I'm not sure that was intended - and because allFutures doesn't cancel its dependent futures, a cancellation here looks like it will leak

arnetheduck: ah, no cancellation either, my bad - perhaps we should log an error though since an exception leaking out of that construct this would imply a bug somewhere

beacon_chain/nimbus_validator_client.nim
zah: As the comment above indicates:

> We use `allFutures` because all failures are already reported as user-visible warnings in `queryValidatorsSource`

PR #5391, opened 2023-09-05
allow archiving blobs

tersec:


arnetheduck (approved):

arnetheduck:
I anticipate we might need a separate option at some point, but this covers the basics well

PR #5392, opened 2023-09-05
Fix /eth/v1/validator/liveness/{epoch} call names and links in code

cheatfate:


tests/test_validator_client.nim
```diff
@@ -335,3 +336,47 @@ suite "Validator Client test suite":
...
+      for index in start ..< (start + count):
+        let visibility = if index mod modv == 0: true else: false
+        res.add(RestLivenessItem(index: ValidatorIndex(uint64(index)),
+                                 is_live: visibility))
+      res
+
+    const Tests = [
+      (0, 100_000, 3),
+      (100_000, 200_000, 7),
+      (1_000_000, 2_000_000, 2)
```
arnetheduck: I'm not sure this test will reveal anything new, ie it's just more of the same - it does take a fair bit of cycles to run though, what's the aim?

PR #5393, opened 2023-09-05
avoid `Taskpool.spawn` within `{.async.}` for Nim 2.0

etan-status:
In Nim 2.0, attempting to use `Taskpool.spawn` inside `{.async.}` `proc` leads to `Error: cannot generate destructor for generic type: Isolated`.

Add an intermediate wrapper `proc` that performs the `spawn` operation to workaround the problem.

### DIFF
--- a/beacon_chain/gossip_processing/batch_validation.nim
+++ b/beacon_chain/gossip_processing/batch_validation.nim
@@ -228,6 +228,12 @@ proc batchVerifyTask(task: ptr BatchTask) {.nimcall.} =
 
   discard task[].signal.fireSync()
 
+proc spawnBatchVerifyTask(tp: Taskpool, task: ptr BatchTask) =
+  # Inlining this `proc` leads to problems on Nim 2.0
+  # - Error: cannot generate destructor for generic type: Isolated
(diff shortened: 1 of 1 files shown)


etan-status:
@mratsim in Nim 2.0 there is weird interaction between `toTask` / `spawn` and `async` macro and the `isolate`. Not sure what it is, have linked a possible reason in the code comment.

```sh
make -j update && make -j deps NIM_COMMIT=version-2-0 && make -j nimbus_beacon_node NIM_COMMIT=version-2-0
```

etan-status:
While the extra wrapper is ugly, it unblocks Nim 2.0 CI in https://github.com/status-im/nimbus-eth2/pull/5142

mratsim:
An upstream bug should be filed. I'd assume it can be reproduced with just std/tasks and std/asyncdispatch or plain closure iterators.

tersec:
> An upstream bug should be filed. I'd assume it can be reproduced with just std/tasks and std/asyncdispatch or plain closure iterators.

https://github.com/nim-lang/Nim/issues/22305 looks likely. If it's not that, when that's fixed, this can be investigated more and a new bug filed.

tersec (approved):

PR #5406, opened 2023-09-07
wait a few slots into an epoch to schedule next epoch's duties

tersec:
This changes the lookahead slot range from `SLOTS_PER_EPOCH` to `2*SLOTS_PER_EPOCH` to 0.75 epochs to 1.75 epochs, to avoid some empty slot overhead when calculating these duties.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1105,8 +1105,10 @@ proc updateGossipStatus(node: BeaconNode, slot: Slot) {.async.} =
           "Getting head EpochRef should never fail")
         node.consensusManager[].actionTracker.updateActions(epochRef)
 
-      if node.consensusManager[].actionTracker.needsUpdate(
-          forkyState, slot.epoch + 1):
+      # Avoid some empty slot transition overhead and reorg risk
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1105,8 +1105,10 @@ proc updateGossipStatus(node: BeaconNode, slot: Slot) {.async.} =
           "Getting head EpochRef should never fail")
         node.consensusManager[].actionTracker.updateActions(epochRef)
 
-      if node.consensusManager[].actionTracker.needsUpdate(
-          forkyState, slot.epoch + 1):
+      # Avoid some empty slot transition overhead and reorg risk
+      if  slot mod SLOTS_PER_EPOCH >= SLOTS_PER_EPOCH div 4 and
```
```suggestion
      if  slot.since_epoch_start >= SLOTS_PER_EPOCH div 4 and
```

arnetheduck:
I'm not convinced about this one - seems like a degradation, ie users use it to not miss a block during restart and narrowing that window seems bad on balance

tersec:
> I'm not convinced about this one - seems like a degradation, ie users use it to not miss a block during restart and narrowing that window seems bad on balance

Alright.

PR #5408, opened 2023-09-07
fix sync contribution validation for first slot per period

etan-status:
Sync committee duties are performed by the sync committee as determined by slot + 1. We did it correctly for individual messages, but selected the incorrect participants for aggregate contributions for the very first slot per period (roughly 1 per ~27 hrs on Mainnet). The faulty participants selection code was originally introduced in #2925.


tersec (approved):

etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5423

That referenced new test is fixed with this PR here :)

PR #5409, opened 2023-09-07
only request `nextPeriod` sync duties close to end of `currentPeriod`

etan-status:
Currently we always request duties for current and next sync period. As sync periods are quite long (~27 hrs on Mainnet), having access to the duties so early doesn't help too much. To avoid running into errors when the BN does not have the duties available around period boundary, delay requesting them until the current period is close to finish.

`SYNC_COMMITTEE_SUBNET_COUNT` epochs are what the spec says should be the lookahead timing of starting to subscribe to sync committee gossip. Reusing the constant here for consistency.

This fixes these warning messages in the first slot of a new period.

```
rocketpool_validator   | WRN 2023-09-07 20:19:35.439+00:00 Beacon node has incompatible configuration reason="Epoch value is far from the future;400;getSyncCommitteeDuties(first);invalid-request" node=http://eth2:5052[Nimbus/v23.8.0-872b19-stateofus] node_index=0 node_roles=AGBSDT
rocketpool_validator   | WRN 2023-09-07 20:19:35.440+00:00 Unable to get sync committee duties        period=889 epoch=227584 reason="Epoch value is far from the future;400;getSyncCommitteeDuties(first);invalid-request" service=duties_service
rocketpool_validator   | NOT 2023-09-07 20:19:35.441+00:00 Beacon node is in sync                     head_slot=7274495 sync_distance=1 is_optimistic=false node=http://eth2:5052[Nimbus/v23.8.0-872b19-stateofus] node_index=0 node_roles=AGBSDT
```

### DIFF
--- a/beacon_chain/validator_client/duties_service.nim
+++ b/beacon_chain/validator_client/duties_service.nim
@@ -372,13 +372,17 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef) {.async.} =
     counts[0] = (currentPeriod,
                  await service.pollForSyncCommitteeDuties(currentPeriod))
 
-    # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/altair/validator.md#sync-committee-subnet-stability
-    const lookaheadEpochs = SYNC_COMMITTEE_SUBNET_COUNT
-    if (currentEpoch + lookaheadEpochs) >= nextPeriod.start_epoch:
+    const
+      numDelayEpochs = 4  # Chosen empirically
+      numLookaheadEpochs =
+        max(EPOCHS_PER_SYNC_COMMITTEE_PERIOD, numDelayEpochs) -
+        numDelayEpochs + 1
+    if (currentEpoch + numLookaheadEpochs) >= nextPeriod.start_epoch:
       counts[1] = (nextPeriod,
                    await service.pollForSyncCommitteeDuties(nextPeriod))
     else:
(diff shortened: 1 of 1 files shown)


cheatfate (approved):

beacon_chain/validator_client/duties_service.nim
```diff
@@ -371,8 +371,15 @@ proc pollForSyncCommitteeDuties*(service: DutiesServiceRef) {.async.} =
     var counts: array[2, tuple[period: SyncCommitteePeriod, count: int]]
     counts[0] = (currentPeriod,
                  await service.pollForSyncCommitteeDuties(currentPeriod))
-    counts[1] = (nextPeriod,
-                 await service.pollForSyncCommitteeDuties(nextPeriod))
+
+    # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/altair/validator.md#sync-committee-subnet-stability
+    const lookaheadEpochs = SYNC_COMMITTEE_SUBNET_COUNT
```
arnetheduck: afair this drives a log which allows users to anticipate the sync committee duties early and therefore avoid restarting / upgrading the node - it would be better to continue making the request early but perhaps delay it by a few epochs around the cutoff

etan-status: The VC doesn't have such a log, and the BN which has it doesn't start logging it until the subscription is active. However, can still change to a scheme similar to what you describe.

PR #5413, opened 2023-09-10
enable Windows GitHub Actions with Nim 2.0

tersec:


etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5419

PR #5418, opened 2023-09-11
Address #5377 part(1/2)

cheatfate:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -940,6 +940,9 @@ template unrecognizedFieldWarning =
   trace "JSON field not recognized by the current version of Nimbus. Consider upgrading",
         fieldName, typeName = typetraits.name(typeof value)
 
+template unrecognizedFieldIgnore =
+  discard readValue(reader, JsonString)
```
arnetheduck, line 944: maybe a debug log should be added here

PR #5421, opened 2023-09-12
rm `state_sim` WASM wrapper/demo

tersec:
`state_sim` was already removed


etan-status (approved):
The `build.sh` information is mostly same as `build_ncli.sh`, so we don't bury any wasm knowledge with this PR.

All in favor of cleanups :)

PR #5425, opened 2023-09-13
provide hint when Git LFS is not installed

etan-status:
Add a hint message during git clone to direct the user to install Git LFS if it is missing. It is required to clone Holesky submodule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,8 @@ ifeq ($(NIM_PARAMS),)
 # The `git reset ...` will try to fix a `make update` that was interrupted
 # with Ctrl+C after deleting the working copy and before getting a chance to
 # restore it in $(BUILD_SYSTEM_DIR).
+
+# `vendor/holesky` requires Git LFS
 ifeq (, $(shell which git-lfs))
 ifeq ($(shell uname), Darwin)
 $(error Git LFS not installed. Run 'brew install git-lfs' to set up)
--- a/Makefile
+++ b/Makefile
@@ -124,10 +124,12 @@ ifeq ($(NIM_PARAMS),)
 # The `git reset ...` will try to fix a `make update` that was interrupted
 # with Ctrl+C after deleting the working copy and before getting a chance to
 # restore it in $(BUILD_SYSTEM_DIR).
+ifeq (, $(shell which git-lfs))
 ifeq ($(shell uname), Darwin)
-	LFS_MISSING_MSG := "Git LFS not installed. Run 'brew install git-lfs' to set up."
+	$(error Git LFS not installed. Run 'brew install git-lfs' to set up.)
 else
-	LFS_MISSING_MSG := "Git LFS not installed."
+	$(error Git LFS not installed.)
+endif
 endif
 
 GIT_SUBMODULE_UPDATE := git submodule update --init --recursive
--- a/Makefile
+++ b/Makefile
@@ -134,8 +134,7 @@ endif
 
 GIT_SUBMODULE_UPDATE := git submodule update --init --recursive
 .DEFAULT:
-	+@ if ! which git-lfs >/dev/null; then echo -e $(LFS_MISSING_MSG)'\n'; exit 1; fi; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -124,9 +124,16 @@ ifeq ($(NIM_PARAMS),)
 # The `git reset ...` will try to fix a `make update` that was interrupted
 # with Ctrl+C after deleting the working copy and before getting a chance to
 # restore it in $(BUILD_SYSTEM_DIR).
+ifeq ($(shell uname), Darwin)
+	LFS_MISSING_MSG := "Git LFS not installed. Run 'brew install git-lfs' to set up."
+else
+	LFS_MISSING_MSG := "Git LFS not installed."
+endif
```
jakubgs: You could just do the check here:
```suggestion
ifeq (, $(shell which git-lfs))
ifeq ($(shell uname), Darwin)
	$(error Git LFS not installed. Run 'brew install git-lfs' to set up.)
else
	$(error Git LFS not installed.)
endif
endif
```

Makefile
```diff
@@ -124,9 +124,18 @@ ifeq ($(NIM_PARAMS),)
...
+else
+	$(error Git LFS not installed.)
+endif
+endif
+
 GIT_SUBMODULE_UPDATE := git submodule update --init --recursive
 .DEFAULT:
-	+@ echo -e "Git submodules not found. Running '$(GIT_SUBMODULE_UPDATE)'.\n"; \
+	+@ if ! which git-lfs >/dev/null; then echo -e $(LFS_MISSING_MSG)'\n'; exit 1; fi; \
+		echo -e "Git submodules not found. Running '$(GIT_SUBMODULE_UPDATE)'.\n"; \
```
etan-status:
```suggestion
	+@ echo -e "Git submodules not found. Running '$(GIT_SUBMODULE_UPDATE)'.\n"; \
```

Makefile
etan-status: Thanks. Had to remove the indentation to avoid `commands commence before first target` error

Makefile
```diff
@@ -124,6 +124,14 @@ ifeq ($(NIM_PARAMS),)
 # The `git reset ...` will try to fix a `make update` that was interrupted
 # with Ctrl+C after deleting the working copy and before getting a chance to
 # restore it in $(BUILD_SYSTEM_DIR).
+ifeq (, $(shell which git-lfs))
+ifeq ($(shell uname), Darwin)
+$(error Git LFS not installed. Run 'brew install git-lfs' to set up.)
+else
+$(error Git LFS not installed.)
```
etan-status:
```suggestion
$(error Git LFS not installed. Run 'brew install git-lfs' to set up)
else
$(error Git LFS not installed)
```

In the error message there is already a trailing `.`

jakubgs (approved):
Looks good. It might just benefit from it's own comment saying what this is necessary for.

Makefile
```diff
@@ -124,6 +124,14 @@ ifeq ($(NIM_PARAMS),)
 # The `git reset ...` will try to fix a `make update` that was interrupted
 # with Ctrl+C after deleting the working copy and before getting a chance to
 # restore it in $(BUILD_SYSTEM_DIR).
+ifeq (, $(shell which git-lfs))
```
etan-status, line 129:
```suggestion

# `vendor/holesky` requires Git LFS
ifeq (, $(shell which git-lfs))
```

arnetheduck:
what about doing https://stackoverflow.com/questions/42019529/how-to-clone-pull-a-git-repository-ignoring-lfs instead?

etan-status:
Skipping the LFS portion sounds like something waiting to blow up when more Ethereum repos adopt LFS. What's your preference? Any downside of simply cloning them same as GA does?

jakubgs:
I have installed `git-lfs` on all 3 Jenkins slave types:

- [`infra-ci#037f1ca4`](https://github.com/status-im/infra-ci/commit/037f1ca4) - ci-slave: install git-lfs for nimbus-eth2

Issue #5278, opened 2023-08-08
Nimbus_beacon_node won't connect to direct peers (Unable to establish outgoing link)


stephanep:
**Describe the bug**
When starting nimbus_beacon_node with a --direct-peer specification, it fails to connect to that peer at start time and never reattempts.
The following event show up in the debug logs: Direct peer error dialing  topics="libp2p gossipsub" msg="Unable to establish outgoing link".
There is no further attempt made, and the direct peer i'm trying to connect to never shows up in the list of peers of the node.

**To Reproduce**
1. Platform details (OS, architecture):
```
nimbus@v-debian:~$ uname -a
Linux v-debian 6.1.0-10-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-2 (2023-07-27) x86_64 GNU/Linux
```

3. Branch/commit used:
Latest stable version to date.
```
nimbus@v-debian:~$ ./nimbus_beacon_node --version
Nimbus beacon node v23.7.0-60f0a2-stateofus
Copyright (c) 2019-2023 Status Research & Development GmbH

eth2 specification v1.4.0-beta.0

Nim Compiler Version 1.6.14 [Linux: amd64]
```

Gave the nightly a try. Same results.

4. Commands being executed:
```
nimbus@v-debian:~$ ./nimbus_beacon_node --direct-peer=/ip4/207.96.125.164/tcp/9000/p2p/16Uiu2HAkwqd8kmmdqLjteD3TyGPz9cVpsZCuW1h8ioBr94FvNqHb --log-level=TRACE > trace.out
```

I took a random peer for the sake of the example, but on my prod setup, where i first hit the problem, i'm correctly using static (non-random) network keys and have configured the direct peer on each side.

6. Relevant log lines:
```
[...]

PR #5427, opened 2023-09-13
Fix direct peers

Menduist:
Ref https://github.com/status-im/nimbus-eth2/issues/5278

The libp2p sides fixes are here: https://github.com/status-im/nim-libp2p/pull/949

I would like to add a test to the CI to makes sure it actually works, but not sure what's the best way to do so

### DIFF
--- a/docs/the_nimbus_book/src/options.md
+++ b/docs/the_nimbus_book/src/options.md
@@ -109,9 +109,10 @@ The following options are available:
      --discv5                  Enable Discovery v5 [=true].
      --dump                    Write SSZ dumps of blocks, attestations and states to data dir [=false].
      --direct-peer             The list of privileged, secure and known peers to connect and maintain the
-                               connection to, this requires a not random netkey-file. In the complete
-                               multiaddress format like: /ip4/<address>/tcp/<port>/p2p/<peerId-public-key>.
-                               Peering agreements are established out of band and must be reciprocal..
+                               connection to. This requires a not random netkey-file. In the multiaddress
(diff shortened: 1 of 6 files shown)


arnetheduck:
> best way to do so

I imagine making one of the peers in the local testnet script become a direct peer and disable all of the other ways it has of getting peers is the easiest thing to do - this "should" cause the local testnet to fail if the connection is not maintained.

Menduist:
Done, though was a bit bloody as we need to know the ENRs of two nodes before-hand.
Will merge the libp2p fix, and then this one will be ready for merge

docs/the_nimbus_book/src/options.md
```diff
@@ -109,9 +109,10 @@ The following options are available:
      --discv5                  Enable Discovery v5 [=true].
      --dump                    Write SSZ dumps of blocks, attestations and states to data dir [=false].
      --direct-peer             The list of privileged, secure and known peers to connect and maintain the
-                               connection to, this requires a not random netkey-file. In the complete
-                               multiaddress format like: /ip4/<address>/tcp/<port>/p2p/<peerId-public-key>.
-                               Peering agreements are established out of band and must be reciprocal..
+                               connection to. This requires a not random netkey-file. In the multiaddress
+                               format like: /ip4/<address>/tcp/<port>/p2p/<peerId-public-key>, or enr format
+                               (enr:-xx). Peering agreements are established out of band and must be
+                               reciprocal..
```
arnetheduck:
```suggestion
                               reciprocal.
```

Menduist: Oh, confutils is adding a dot, funny

PR #5428, opened 2023-09-13
remove 4 more missing constants in preset handling warning

tersec:


beacon_chain/spec/presets.nim
```diff
@@ -23,7 +23,11 @@ const
 
   # Constants from `validator.md` not covered by config/presets in the spec
   TARGET_AGGREGATORS_PER_COMMITTEE*: uint64 = 16
+
+  # Not used anywhere; only for network preset checking
   EPOCHS_PER_RANDOM_SUBNET_SUBSCRIPTION: uint64 = 256
+  MESSAGE_DOMAIN_INVALID_SNAPPY = 0'u64
+  TTFB_TIMEOUT = 5'u64
```
etan-status, line 30: And, I guess, for the REST config endpoint, similar to others

beacon_chain/networking/eth2_network.nim
```diff
@@ -2258,9 +2258,7 @@ func gossipId(
     data: openArray[byte], phase0Prefix, topic: string): seq[byte] =
   # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/p2p-interface.md#topics-and-messages
   # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/altair/p2p-interface.md#topics-and-messages
-  const
-    MESSAGE_DOMAIN_INVALID_SNAPPY = [0x00'u8, 0x00, 0x00, 0x00]
```
etan-status, line 2262: How comes, the invalid snappy domain is nowhere used, even though it is mentioned in the spec?

tersec: Don't know, didn't check. Was dead code and this safely solves an immediate issue without modifying any core BN semantics.

beacon_chain/spec/presets.nim
tersec: I'm not sure how either of these determines VC compatibility with BNs. The VC doesn't use `MESSAGE_DOMAIN_INVALID_SNAPPY` directly, nor does it care about `TTFB_TIMEOUT`.

tersec: And, maybe more to the point, we've had experiences where some VCs sort of blindly checked every single parameter they knew about with what was returned, and reported an incompatibility in irrelevant parameters.

It seems not to be behavior the spec intends, as I recall, though I'd have to check for specific spec verbiage, but regardless adding anything to `getSpec` is a risk of its own. This PR is meant to be a semantic no-op.

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/p2p-interface.md#topics-and-messages
> The `message-id` of a gossipsub message MUST be the following 20 byte value computed from the message data:
> * If `message.data` has a valid snappy decompression, set `message-id` to the first 20 bytes of the `SHA256` hash of
>   the concatenation of `MESSAGE_DOMAIN_VALID_SNAPPY` with the snappy decompressed message data,
>   i.e. `SHA256(MESSAGE_DOMAIN_VALID_SNAPPY + snappy_decompress(message.data))[:20]`.
> * Otherwise, set `message-id` to the first 20 bytes of the `SHA256` hash of
>   the concatenation of `MESSAGE_DOMAIN_INVALID_SNAPPY` with the raw message data,
>   i.e. `SHA256(MESSAGE_DOMAIN_INVALID_SNAPPY + message.data)[:20]`.
>
> *Note*: The above logic handles two exceptional cases:
> (1) multiple snappy `data` can decompress to the same value,
> and (2) some message `data` can fail to snappy decompress altogether.

https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/altair/p2p-interface.md#topics-and-messages
> The derivation of the `message-id` has changed starting with Altair to incorporate the message `topic` along with the > message `data`. These are fields of the `Message` Protobuf, and interpreted as empty byte strings if missing.
> The `message-id` MUST be the following 20 byte value computed from the message:
> * If `message.data` has a valid snappy decompression, set `message-id` to the first 20 bytes of the `SHA256` hash of
>   the concatenation of the following data: `MESSAGE_DOMAIN_VALID_SNAPPY`, the length of the topic byte string (encoded as little-endian `uint64`),
>   the topic byte string, and the snappy decompressed message data:
>   i.e. `SHA256(MESSAGE_DOMAIN_VALID_SNAPPY + uint_to_bytes(uint64(len(message.topic))) + message.topic + snappy_decompress(message.data))[:20]`.
> * Otherwise, set `message-id` to the first 20 bytes of the `SHA256` hash of
>   the concatenation of the following data: `MESSAGE_DOMAIN_INVALID_SNAPPY`, the length of the topic byte string (encoded as little-endian `uint64`),
>   the topic byte string, and the raw message data:
>   i.e. `SHA256(MESSAGE_DOMAIN_INVALID_SNAPPY + uint_to_bytes(uint64(len(message.topic))) + message.topic + message.data)[:20]`.
>
> Implementations may need to carefully handle the function that computes the `message-id`. In particular, messages on topics with the Phase 0
> fork digest should use the `message-id` procedure specified in the Phase 0 document.
> Messages on topics with the Altair fork digest should use the `message-id` procedure defined here.
>  an implementation only supports a single `message-id` function, it can define a switch inline;
> for example, `if topic in phase0_topics: return phase0_msg_id_fn(message) else return altair_msg_id_fn(message)`.

Presumably those exceptional cases are avoided before that gossip id message runs to begin with, and/or it's viewed as acceptable that that it returns an incorrect result because the message will be discarded regardless due to not decompressing.

tersec: https://github.com/status-im/nimbus-eth2/pull/3471

etan-status (approved):

PR #5431, opened 2023-09-15
per payload-builder payload builder validator registration

tersec:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -1678,6 +1680,23 @@ proc registerValidators*(node: BeaconNode, epoch: Epoch) {.async.} =
     warn "registerValidators: exception",
       error = exc.msg
 
+proc registerValidators*(node: BeaconNode, epoch: Epoch) {.async.} =
+  if not node.config.payloadBuilderEnable: return
+
+  var builderKeys: Table[string, seq[ValidatorPubKey]]
+  for pubkey in node.attachedValidators[].validators.keys:
```
arnetheduck, line 1692: where do vc-based validators get registered?

tersec: By the VC, which has to do this signing anyway for validators it controls (and the vast majority of what's happening here is signing some messages):
https://github.com/status-im/nimbus-eth2/blob/568e1fbb8cd401af55a405687a34cca2ff0ff883/beacon_chain/validator_client/duties_service.nim#L508-L546

arnetheduck (approved):

PR #5432, opened 2023-09-15
log engine getPayload requests

tersec:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -409,16 +409,21 @@ proc getExecutionPayload(
           get_expected_withdrawals(forkyState.data)
         else:
           @[]
-      payload = await node.elManager.getPayload(
-        PayloadType, beaconHead.blck.bid.root, executionHead, latestSafe,
-        latestFinalized, timestamp, random, feeRecipient, withdrawals)
 
-    if payload.isNone:
+    info "Requesting engine payload",
```
arnetheduck, line 413: why info? is this of general interest?

tersec: Essentially.

It's a sufficiently valuable debugging tool to detect various otherwise mysterious cases of aborted engine block production, where no explicit error appears, and not particularly spammy: it happens at most once per slot, only when that node is actually proposing.

arnetheduck (approved):

beacon_chain/validators/beacon_validators.nim
arnetheduck: should we have similar ones for builder payloads? or this this the same log?

tersec: > should we have similar ones for builder payloads? or this this the same log?

It would be a different log, and different protocol message, since it's not based on `getPayload` or `getExecutionPayload` (which is specific to the engine API).

But, yes, it would probably be useful to log the equivalent request for the builder API, essentially, the unblind request, and for similar reasons: they can both degenerate into difficult-to-debug-in-hindsight open-ended timeouts where the crucial message appears to have vanished without any visible logging from either protocol participant.

tersec: > should we have similar ones for builder payloads? or this this the same log?

https://github.com/status-im/nimbus-eth2/pull/5434

PR #5436, opened 2023-09-15
Increase default file limit on startup

arnetheduck:
We use file descriptors for validators and sockets and might run out of either on high-validator setups - increasing the limit here is harmless and avoids a common limiting factor in setup


arnetheduck:
lint failure on file that shouldn't have header

etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5440

tersec (approved):

PR #5437, opened 2023-09-18
avoid repeated total active balance computation in fork choice

etan-status:
We currently compute `justified_total_active_balance` inside `calculateProposerBoost`, despite that sum already being known in the `EpochRef` cache. Tracking `justified_total_active_balance` whenever the justified checkpoint updates allows replacing the repeated computation with a lookup, at minimal memory cost.

### DIFF
--- a/beacon_chain/fork_choice/proto_array.nim
+++ b/beacon_chain/fork_choice/proto_array.nim
@@ -126,18 +126,15 @@ iterator realizePendingCheckpoints*(
   self.currentEpochTips.clear()
 
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/fork-choice.md#get_weight
-func calculateProposerBoost(validatorBalances: openArray[Gwei]): uint64 =
-  var total_balance: uint64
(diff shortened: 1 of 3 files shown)


beacon_chain/fork_choice/proto_array.nim
```diff
@@ -126,18 +126,15 @@ iterator realizePendingCheckpoints*(
   self.currentEpochTips.clear()
 
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/fork-choice.md#get_weight
-func calculateProposerBoost(validatorBalances: openArray[Gwei]): uint64 =
-  var total_balance: uint64
-  for balance in validatorBalances:
-    total_balance += balance
```
etan-status, line 132: Note that this excluded slashed validators. 

But in https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/phase0/fork-choice.md#get_weight

it is documented as `get_total_active_balance` which means, slashed validator effective balances are included for the purpose of computing the boost.

PR #5443, opened 2023-09-19
Add KeystoreCache to Keystore Management API.

cheatfate:


beacon_chain/nimbus_validator_client.nim
```diff
@@ -458,8 +459,8 @@ proc asyncRun*(vc: ValidatorClientRef) {.async.} =
   vc.blockService.start()
   vc.syncCommitteeService.start()
 
-  if not isNil(vc.keymanagerServer):
-    doAssert vc.keymanagerHost != nil
+  if not(isNil(vc.keymanagerServer)):
+    doAssert not(isNil(vc.keymanagerHost))
```
tersec, line 463: Both styles seem fine, so this seems like a neutral change, but, is there any particular reason to prefer one or the other?

tersec (approved):

PR #5442, opened 2023-09-18
implement EIP-7514 for Deneb: Add Max Epoch Churn Limit

etan-status:
Cap activations per epoch according to EIP-7514:

- https://eips.ethereum.org/EIPS/eip-7514
- https://github.com/ethereum/consensus-specs/pull/3499

PR #5444, opened 2023-09-19
apply proposer boost to first block in case of equivocation

etan-status:
Implement spec changes to fork choice; this only affects equivocation when multiple blocks are signed for the same slot. Regular operation is not changed.

- https://github.com/ethereum/consensus-specs/pull/3352


etan-status:
This is also necessary to pass 1.4.0-beta.2 tests.

etan-status:
Merging this one into PR #5442 (which contains the bump to 1.4.0-beta.2 tests).

The 1.4.0-beta.1 tests check against the old logic of assigning proposer boost to the last block in case of equivocation

PR #5445, opened 2023-09-19
reject invalid proofs when validating blobs

etan-status:
Currently, passing `0xc00000...` proof seems to pass `verifyProofs`. Unsure why such a check is not necessary in spec, and also unsure whether it is correct to reject proof at infinity, or if it could occur, e.g., for a blob containing all 0 bytes. Weird overall...

### DIFF
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -791,14 +791,10 @@ proc validate_blobs*(expected_kzg_commitments: seq[KzgCommitment],
   if proofs.len != blobs.len:
     return err("validate_blobs: different proof and blob lengths")
 
-  const infinityProof = KzgProof([
-    0xc0.byte, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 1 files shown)


hwwhww:
It is actually a valid case in https://github.com/ethereum/consensus-spec-tests/blob/master/tests/general/deneb/kzg/verify_blob_kzg_proof/kzg-mainnet/verify_blob_kzg_proof_case_correct_proof_0951cfd9ab47a8d3/data.yaml

asn-d6:
Proof being the point at infinity **is a valid case** for when the blob is all zeroes. A blob being all zeroes is basically the zero polynomial -- its commitment will be the point at infinity and the corresponding proof will also be the point at infinity.

A related edge case can happen for a blob full of twos (or any other constant number): A blob full of twos would correspond to the polynomial `p(x) = 2` and it's corresponding KZG proof would be the point at infinity. It's corresponding commitment would be `2G` (not the point at infinity).

If these edge cases are spilling out to clients and causing confusion, we should probably document them somewhere.

etan-status:
Found the problem. There are two output values, one is a return value, and the other is a `bool *`. We only checked the first one but not the second. When I check both, the test works fine :)

etan-status:
Introduction of problem here: https://github.com/status-im/nimbus-eth2/pull/4732

Was always broken. Nice that the tests caught it 😅

PR #5447, opened 2023-09-19
VC: new scoring functions.

cheatfate:
New scoring functions added:

* Aggregated attestations
* Sync committee contributions
* Sync committee messages


beacon_chain/validator_client/scoring.nim
```diff
@@ -47,3 +68,95 @@ proc getAttestationDataScore*(rootsSeen: Table[Eth2Digest, Slot],
...
+      else:
+        if slot != FAR_FUTURE_SLOT:
+          if slot == currentSlot:
+            # Perfect score
+            Inf
+          else:
+            float64(1) / (float64(1) + float64(currentSlot) - float64(slot))
+        else:
+          # Block monitoring is disabled or we missed a block.
+          getLexicographicScore(cdata.data.root)
```
etan-status: So, around as valuable as if there was 1 missed slot (0.5 score), on average.
But more valueable as if there were 2 missed slots (0.33 score).

Would it make sense to have `getLexicographicScore` only give scores within 0.33 .. 0.67, to avoid extreme situation where an unknown block ranks below something known-to-be ancient that missed like 30 slots

etan-status: The other extreme would be based on majority, if the VC is attached to 8 BNs and 7 of them recommend some message with slot 2 before wall clock, but 1 of them recommends a more recent message with slot 1 before wall clock. In that case, maybe the older one should be selected, as the more recent one did not propagate through the network.

In any case, probably not too big of a deal, sync duties span 8192 slots at a time, and missing one of them is not that impactful. If missed slots happen at a regular rate, there's a different problem in the network than the scoring 😅

etan-status (approved):

cheatfate:
I have updated `getSyncCommitteeMessageDataScore()` so now it returns:
- when slot has been found in rootsSeen, its score range is in `(1.000, 2.000]` or `+Inf`.
- when slot is not found or block monitoring disabled, its score range is in `(0.000, 1.000]`.
- when response is sent from optimistically synced node, score is `-Inf`.

beacon_chain/validator_client/scoring.nim
cheatfate: addressed in https://github.com/status-im/nimbus-eth2/pull/5447/commits/2944ac6a209c5eaedf2ca75413af2cc310ab7ca2

beacon_chain/validator_client/scoring.nim
```diff
@@ -47,3 +68,99 @@ proc getAttestationDataScore*(rootsSeen: Table[Eth2Digest, Slot],
...
+          if slot == currentSlot:
+            # Perfect score
+            Inf
+          else:
+            float64(1) +
+              float64(1) / (float64(1) + float64(currentSlot) - float64(slot))
+        else:
+          # Block monitoring is disabled or we missed a block, in this case
+          # score value will be in range of `(0, 1]`
+          getLexicographicScore(cdata.data.root)
```
arnetheduck, line 149: I think it would be better for the implementation to fall back on VC configuration order in case it doesn't have information / ends up with tie.. basically, we then get a hierarchical scoring algorithm where ties at one level are resolved at the next, the last level being the order of declaration of VC (so that if everything else is equal, we pick the first node)

beacon_chain/validator_client/scoring.nim
```diff
@@ -47,3 +68,99 @@ proc getAttestationDataScore*(rootsSeen: Table[Eth2Digest, Slot],
...
+    vc.rootsSeen, vc.beaconClock.now().slotOrZero(), cdata)
+
+proc processVotes(bits: var CommitteeValidatorsBits,
+                  attestation: Attestation): uint64 =
+  var res = 0'u64
+  for index in 0 ..< len(attestation.aggregation_bits):
+    if attestation.aggregation_bits[index]:
+      if not(bits[index]):
+        inc(res)
+        bits[index] = true
```
arnetheduck, line 171: where is the length of `bits` decided?

cheatfate: `CommitteeValidatorsBits* = BitList[Limit MAX_VALIDATORS_PER_COMMITTEE]`

arnetheduck: a bitlist is a seq - this *should* cause a runtime index defect but maybe it doesn't due to some nim bugs

arnetheduck: it certainly causes a buffer overrun

cheatfate: why do you think so? What if `[]` access is overloaded?

arnetheduck: https://github.com/status-im/nim-ssz-serialization/blob/master/ssz_serialization/bitseqs.nim#L184 - see the assert - there's no length extension on out-of-bounds access

cheatfate: I'm sorry, probably i'm too dumb to understand how this code could get out-of-bounds access
```nim
  for index in 0 ..< len(attestation.aggregation_bits):
    if attestation.aggregation_bits[index]:
      if not(bits[index]):
        inc(res)
        bits[index] = true
```
`index` is in range of `[0, len(attestation.aggregation_bits])`, because of `for` statement. So if it will go out-of-bounds it means that `for` ignore its bounds, or `len(attestation.aggregation_bits)` reports un-trusted values.

beacon_chain/validator_client/scoring.nim
```diff
@@ -47,3 +70,91 @@ proc getAttestationDataScore*(rootsSeen: Table[Eth2Digest, Slot],
...
                               adata: ProduceAttestationDataResponse): float64 =
   getAttestationDataScore(vc.rootsSeen, adata)
+
+proc getAggregatedAttestationDataScore*(
+       adata: GetAggregatedAttestationResponse
+     ): float64 =
+  let
+    length = int(MAX_VALIDATORS_PER_COMMITTEE)
+    ones = countOnes(adata.data.aggregation_bits)
+    res = if length == ones: Inf else: float64(ones) / float64(length)
```
arnetheduck: `data.aggregation_bits` cannot practically have this many ones - typical committee size now around 500 validators.

cheatfate: The idea of score is to get higher value when number of bits being set is bigger... I think current algorithm definitely obtains such score value and also it doesn't depends on typical/normal/obvious and any other count of validators.

arnetheduck: yes, but it's not possible to attain a perfect score with this algorithm - even if an attestation has all of its bits set, this will not result in a perfect score since `length` is always larger than the committee size

cheatfate: So what if one BN returns attestation with committee size 500 bits but only 480 bits set (`480 div 500 == 0.9600`), and another BN returns attestation data with committee size 490 bits but only 471 bits set (`471 div 490 == 0.9612`), so your proposal is to use attestation data with 490/471. While my proposal is to use 500/480 instead. So yeah perfect score is very hard to achieve, but better attestations will be selected.

arnetheduck: > perfect score is very hard to achieve

it's not "very hard" - it's simply not possible unless all eth in circulation is staked - `MAX_VALIDATORS_PER_COMMITTEE` is derived from a validator count that cannot practically be achieved (it was chosen so that we will never ever hit it which makes the merkle tree "stable") - we can simply remove the "perfect attestation" score at this point because it doesn't make sense.

PR #5460, opened 2023-09-26
mention Holesky in the book

narimiran:
This replaces mentions of Prater/Goerli with Holesky.

It also introduces `run-holesky-beacon-node.sh` (based on the existing `run-prater-beacon-node.sh`), which is referenced throughout the guide.

Some links currently don't work (which is expected), but I'll re-check them after Thursday.

### DIFF
--- a/docs/the_nimbus_book/src/attestation-performance.md
+++ b/docs/the_nimbus_book/src/attestation-performance.md
@@ -35,9 +35,9 @@ The following options are available:
 
 Where:
 
-- The `network` can be `mainnet`, `prater`, or `sepolia`.
+- The `network` can be `mainnet`, `holesky`, or `sepolia`.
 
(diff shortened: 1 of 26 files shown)


tersec:
> It also introduces `run-holesky-beacon-node.sh` (based on the existing `run-prater-beacon-node.sh`), which is referenced throughout the guide.

There's been discussion of potentially deprecating these -- at least, would be good to hold off on automatically adding a holesky one.

narimiran:
> There's been discussion of potentially deprecating these -- at least, would be good to hold off on automatically adding a holesky one.

Could/should that be a part of a separate PR, if/when we definitely decide if that should be deprecated and what should be used as a replacement?

Also, there are `make prater` and `make goerli` — are those also deprecated or should we introduce `make holesky`?

PR #5474, opened 2023-10-03
REST API produceBlockV3 implementation

cheatfate:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -1885,3 +1885,162 @@ proc registerDuties*(node: BeaconNode, wallSlot: Slot) {.async.} =
...
+          useBuilderBlock =
+            if collectedBids.builderBidAvailable:
+              (not collectedBids.engineBidAvailable) or builderBetterBid(
+                localBlockValueBoost,
+                collectedBids.payloadBuilderBidFut.read.get().blockValue,
+                collectedBids.engineBlockFut.read.get().blockValue)
+            else:
+              if not(collectedBids.engineBidAvailable):
+                return Result[ForkedAndBlindedBeaconBlock, string].err(
+                       "Engine bit is not available")
```
tersec:
```suggestion
                       "Engine bid is not available")
```

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1885,3 +1885,162 @@ proc registerDuties*(node: BeaconNode, wallSlot: Slot) {.async.} =
 
         node.consensusManager[].actionTracker.registerDuty(
           slot, subnet_id, validator_index, isAggregator)
+
+proc makeBeaconBlockForHeadAndSlotV3*(
```
tersec: Well, this entire function has duplication. Some of it might be okay, but some of it might be refactored.

Would have to look at the extent it's because it wants this `ForkedAndBlindedBeaconBlock` type you created.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1885,3 +1885,162 @@ proc registerDuties*(node: BeaconNode, wallSlot: Slot) {.async.} =
...
+            payloadValue = blindedResult.get().blockValue
+          return Result[ForkedAndBlindedBeaconBlock, string].err(
+                       "Implementation is missing")
+          # return Result[ForkedAndBlindedBeaconBlock, string].ok(
+          #   ForkedAndBlindedBeaconBlock.init(
+          #     blindedBlock, Opt.some(payloadValue), Opt.none(UInt256)))
+        else:
+          collectedBids.engineBlockFut.read().get()
+
+      elif slot.epoch >= node.dag.cfg.CAPELLA_FORK_EPOCH:
```
tersec: Repeating this `collectedBids`/`useBuilderBlock` logic separately for each fork is unfortunate, but might be okay because the current intention is that at any given time, only 2 or maybe 3 builder API forks will be supported. Old ones can be removed (they're not part of the consensus protocol as such and are unnecessary for historical recreation).

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1885,3 +1885,162 @@ proc registerDuties*(node: BeaconNode, wallSlot: Slot) {.async.} =
...
+          return Result[ForkedAndBlindedBeaconBlock, string].ok(
+            ForkedAndBlindedBeaconBlock.init(
+              blindedBlock, Opt.some(payloadValue), Opt.none(UInt256)))
+        else:
+          collectedBids.engineBlockFut.read().get()
+
+      elif slot.epoch >= node.dag.cfg.BELLATRIX_FORK_EPOCH:
+        # Bellatrix MEV is not supported; this signals that, because it triggers
+        # intentional SignedBlindedBeaconBlock/ExecutionPayload mismatches.
+        let
```
tersec: There's an odd disconnect here -- as the comment says (copy/pasted from elsewhere, but it's not wrong), the Bellatrix Builder API isn't supported. So this logic is just unnecessary -- either it uses the engine API or nothing. None of this `builderBetterBid`, etc is necessary or useful.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1885,3 +1885,162 @@ proc registerDuties*(node: BeaconNode, wallSlot: Slot) {.async.} =
...
+          useBuilderBlock =
+            if collectedBids.builderBidAvailable:
+              (not collectedBids.engineBidAvailable) or builderBetterBid(
+                localBlockValueBoost,
+                collectedBids.payloadBuilderBidFut.read.get().blockValue,
+                collectedBids.engineBlockFut.read.get().blockValue)
+            else:
+              if not(collectedBids.engineBidAvailable):
+                return Result[ForkedAndBlindedBeaconBlock, string].err(
+                       "Engine bit is not available")
```
tersec:
```suggestion
                       "Engine bid is not available")
```

beacon_chain/validators/beacon_validators.nim
tersec: Indeed, this would automatically mitigate the duplication issue; only Capella and Deneb forks need this logic, no others.

beacon_chain/spec/forks.nim
```diff
@@ -145,6 +155,28 @@ type
...
 
+  ForkedAndBlindedBeaconBlock* = object
+    case kind*: ConsensusBlindedFork
+    of ConsensusBlindedFork.Phase0:
+      phase0Data*: phase0.BeaconBlock
+    of ConsensusBlindedFork.Altair:
+      altairData*: altair.BeaconBlock
+    of ConsensusBlindedFork.BellatrixNormal:
+      bellatrixData*: bellatrix.BeaconBlock
+    of ConsensusBlindedFork.BellatrixBlinded:
```
tersec: Is this `ConsensusBlindedFork.BellatrixBlinded` necessary for some not-in-this-PR calling code which requires this kind of regularity in types? Because, as noted by the comments in this PR's changes, the Bellatrix Builder API is not supported. This case should never occur, and `bellatrix_mev.BlindedBeaconBlock` is a empty stub.

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5474/commits/0d4dcabcc509c502ae4b062e5cbc2522ca88f281

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5474/commits/0d4dcabcc509c502ae4b062e5cbc2522ca88f281

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5474/commits/0d4dcabcc509c502ae4b062e5cbc2522ca88f281

beacon_chain/spec/mev/deneb_mev.nim
```diff
@@ -89,6 +89,11 @@ type
     signed_blinded_blob_sidecars*:
       List[SignedBlindedBlobSidecar, Limit MAX_BLOBS_PER_BLOCK]
 
+  BlindedBlockContents* = object
```
tersec: Would be good to either note as explicitly non-spec, or if it's derived from, e.g., a Beacon API requirement fairly directly (it's the type required by or returned by some REST endpoint), link to a stable, not `master`/`main`/etc URL for that.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3008,6 +3127,197 @@ proc readValue*(reader: var JsonReader[RestJson],
...
+    block:
+      writer.writeFieldName("data")
+      writer.beginRecord()
+      writer.writeField("blinded_block", value.denebBlinded.blinded_block)
+      writer.writeField("blinded_blob_sidecars",
+                        value.denebBlinded.blinded_blob_sidecars)
+      writer.endRecord()
+  writer.endRecord()
+
+proc readValue*(reader: var JsonReader[RestJson],
```
tersec, line 3215: Why would the REST server ever have to read a `BlindedBeaconBlock` it doesn't know the specific fork of already? It's for proposing, in the moment, and if it doesn't match the current slot/epoch it's not relevant to the node.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -572,6 +572,123 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        (await node.makeBeaconBlockForHeadAndSlotV3(qrandao, qgraffiti, qhead,
+                                                    qslot)).valueOr:
+          # HTTP 400 error is only for incorrect parameters.
+          return RestApiResponse.jsonError(Http500, error)
+
+    let headers = message.getHeaders()
+    if contentType == sszMediaType:
+      case message.kind
+      of ConsensusBlindedFork.Phase0:
+        RestApiResponse.sszResponse(message.phase0Data, headers)
```
tersec: Which ultimately uses `ForkedAndBlindedBeaconBlock` `writeValue` added elsewhere in this PR, on the apparently non-`Forked` version of the `BlindedBeaconBlock`, so why is the
```nim
proc writeValue*(writer: var JsonWriter[RestJson],
                 value: ForkedAndBlindedBeaconBlock) {.raises: [IOError].} =
```
function necessary here?

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -572,6 +572,123 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+                return RestApiResponse.jsonError(Http503, BeaconNodeInSyncError)
+              tres
+          proposer = node.dag.getProposer(qhead, qslot).valueOr:
+            return RestApiResponse.jsonError(Http400, ProposerNotFoundError)
+
+        if not node.verifyRandao(
+            qslot, proposer, qrandao, qskip_randao_verification):
+          return RestApiResponse.jsonError(Http400, InvalidRandaoRevealValue)
+
+        (await node.makeBeaconBlockForHeadAndSlotV3(qrandao, qgraffiti, qhead,
```
tersec: It apparently needs `ForkedAndBlindedBeaconBlock` here primarily because it's storing one as `message`, but then it immediately does two things with this `message`:

1. calls `getHeaders` (defined above); and
2. calls `RestApiResponse.sszResponse`, not even on the `Forked` type per se, but on each specific case object field. See question there about why `writeValue` is defined on `ForkedAndBlindedBeaconBlock`, as a result.

But neither really requires storage of a forked/"any fork" version of a `BlindedBeaconBlock` or `BlindedBeaconBlockContents`, because what this `produceBlockV3` endpoint returns is neither, but a bytestring of some sort (e.g., SSZ). So it's still unclear to me why have all this infrastructure for a `ForkedAndBlindedBeaconBlock` which seems not to be necessary to begin with.

The result is hundreds of lines of

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
cheatfate: This is not for REST server it is made for VC, because VC will use this `readValue` procedure to decode `produceBlockV3` response.

tersec: The VC knows too though. It knows the fork schedule and if it's asking for a `produceBlockV3` response during Capella, anything except a Capella response is simply an error.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3008,6 +3127,197 @@ proc readValue*(reader: var JsonReader[RestJson],
...
+                value: var ForkedAndBlindedBeaconBlock) {.
+     raises: [SerializationError, IOError].} =
+  var
+    version: Opt[ConsensusBlindedFork]
+    blinded: Opt[bool]
+    executionValue: Opt[UInt256]
+    consensusValue: Opt[UInt256]
+    data: Opt[JsonString]
+
+  for fieldName in readObjectFields(reader):
```
tersec, line 3225: Is this entire loop, with its specifically handwritten multiple-field detection, etc, not basically equivalent to something like

```nim
type
  RestForkedAndBlindedBeaconBlockSkeleton = object
    version: Opt[string]  # because it otherwise has to do a special conversion
    blinded: Opt[bool]
    executionValue: Opt[UInt256]
    consensusValue: Opt[UInt256]
    data: Opt[JsonString]
```
and then letting the JSON parser handle this, or would it not detect multiple fields with each name?

It seems like recreating logic that exists elsewhere.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -572,6 +572,123 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+                         ConsensusBlindedFork.DenebBlinded}:
+        "true"
+      else:
+        "false"
+    res.add("eth-consensus-version", toString(forked.kind))
+    res.add("eth-execution-payload-blinded", blinded)
+    if forked.executionValue.isSome():
+      res.add("eth-execution-payload-value", $(forked.executionValue.get()))
+    if forked.consensusValue.isSome():
+      res.add("eth-consensus-payload-value", $(forked.consensusValue.get()))
```
arnetheduck:
```suggestion
      res.add("eth-consensus-block-value", $(forked.consensusValue.get()))
```

https://github.com/ethereum/beacon-APIs/blob/c097f1a62c9a12c30e8175a39f205f92d3b931a9/apis/validator/block.v3.yaml#L56

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3050,6 +3169,197 @@ proc readValue*(reader: var JsonReader[RestJson],
...
+    writer.writeField("execution_payload_blinded", "false")
+    writeExecutionValue()
+    writeConsensusValue()
+    writer.writeField("data", value.capellaData)
+  of ConsensusBlindedFork.CapellaBlinded:
+    writeForkVersion()
+    writer.writeField("execution_payload_blinded", "true")
+    writeExecutionValue()
+    writeConsensusValue()
+    writer.writeField("data", value.capellaData)
```
etan-status:
```suggestion
    writer.writeField("data", value.capellaBlinded)
```

beacon_chain/rpc/rest_validator_api.nim
etan-status: I think you need some form of `ForkedBlindedBeaconBlock`, because v3 may return _either_ forked or non-forked depending on which one is better. And that is not known at call-time.

This, unless you want to do that logic inside the REST layer, which doesn't seem correct, because it makes it non-reusable, e.g., for the non-VC local block proposal flow.

Also having a `Forked` type instead of forky `MaybeBlindedBeaconBlock` makes it consistent with the existing v1/v2 `makeBeaconBlockForHeadAndSlot` which also returns a `ForkedBeaconBlock`.

beacon_chain/rpc/rest_validator_api.nim
etan-status: This is for SSZ, not JSON

tersec: Oh, indeed.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
etan-status: The REST library uses a single type for each API, so the `Forked` type is needed here I think.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
etan-status: there's also `strictParse` for the other fields and so on.

etan-status (approved):

tersec (approved):

PR #5490, opened 2023-10-10
VC: Obtain randao signature before slot proposal.

cheatfate:


beacon_chain/validator_client/block_service.nim
```diff
@@ -462,13 +534,14 @@ proc addOrReplaceProposers*(vc: ValidatorClientRef, epoch: Epoch,
           for task in epochDuties.duties:
             if task notin duties:
               # Task is no more relevant, so cancel it.
-              debug "Cancelling running proposal duty task",
+              debug "Cancelling running proposal/randao duty tasks",
```
```suggestion
              debug "Cancelling running proposal/RANDAO duty tasks",
```

beacon_chain/validator_client/block_service.nim
```diff
@@ -462,13 +534,14 @@ proc addOrReplaceProposers*(vc: ValidatorClientRef, epoch: Epoch,
...
                     slot = task.duty.slot,
                     validator = shortLog(task.duty.pubkey)
-              task.future.cancelSoon()
+              task.proposeFut.cancelSoon()
+              task.randaoFut.cancelSoon()
             else:
               # If task is already running for proper slot, we keep it alive.
-              debug "Keep running previous proposal duty task",
+              debug "Keep running previous proposal/randao duty tasks",
```
```suggestion
              debug "Keep running previous proposal/RANDAO duty tasks",
```

beacon_chain/validator_client/block_service.nim
```diff
@@ -125,6 +127,95 @@ proc lazyWait[T](fut: Future[T]) {.async.} =
...
+  let
+    destSlot = slot - 1'u64
+    destOffset = TimeDiff(
+      nanoseconds: NANOSECONDS_PER_SLOT.int64 div INTERVALS_PER_SLOT)
+    deadline = destSlot.start_beacon_time() + destOffset
+    # We going to wait to T - 2 * INTERVALS_PER_SLOT, where T is proposer's
+    # duty slot.
+    currentSlot = (await vc.checkedWaitForSlot(destSlot, destOffset,
+                   false)).valueOr:
+      debug "Unable to perform randao caching because of system time"
```
```suggestion
      debug "Unable to perform RANDAO caching because of system time"
```

tersec: I'm not sure I understand the message though -- in what sense is it "because of system time" -- is it a timeout, or incorrectly set system time, or?

beacon_chain/validator_client/block_service.nim
```diff
@@ -125,6 +127,95 @@ proc lazyWait[T](fut: Future[T]) {.async.} =
...
+      debug LogMessage, epoch = epoch, validator = shortLog(proposerKey),
+                        source = "cache", elapsed_time = timeElapsed,
+                        delay = vc.getDelay(deadline)
+      return rsig.get()
+
+  let
+    validator = vc.getValidatorForDuties(proposerKey, slot).valueOr: return
+    signature =
+      try:
+        let res = await validator.getEpochSignature(fork, genesisRoot, epoch)
```
tersec: Could use `valueOr` here, maybe

beacon_chain/validator_client/block_service.nim
```diff
@@ -125,6 +127,95 @@ proc lazyWait[T](fut: Future[T]) {.async.} =
...
+
+proc getRandao(vc: ValidatorClientRef, slot: Slot,
+               proposerKey: ValidatorPubKey) {.async.} =
+  if slot == Slot(0'u64):
+    return
+
+  let
+    destSlot = slot - 1'u64
+    destOffset = TimeDiff(
+      nanoseconds: NANOSECONDS_PER_SLOT.int64 div INTERVALS_PER_SLOT)
```
tersec: https://github.com/ethereum/consensus-specs/pull/3510 looks poised to remove `INTERVALS_PER_SLOT` from the spec

beacon_chain/validator_client/common.nim
```diff
@@ -52,6 +52,14 @@ type
     Cached, IncorrectTime, MissingIndex, MissingFee, MissingGasLimit,
     ErrorSignature, NoSignature
 
+  RandaoPair* = object
+    pubkey*: ValidatorPubKey
+    signature*: ValidatorSig
+
+  RandaoCacheItem* = object
+    data*: array[SLOTS_PER_EPOCH, RandaoPair]
+    fork*: Fork
```
tersec: Isn't the `Fork` known, because
```nim
randaoCache*: Table[Epoch, RandaoCacheItem]
```
and
```nim
fork = vc.forkAtEpoch(epoch)
```
?

It's not clear to me why it's caching this, when it's a fast, easily computable function of other already-cached information.

beacon_chain/validator_client/block_service.nim
```diff
@@ -125,6 +127,95 @@ proc lazyWait[T](fut: Future[T]) {.async.} =
...
+    signatures[].setSignature(slot, validator.pubkey, signature)
+  let timeElapsed = Moment.now() - start
+  debug LogMessage, epoch = epoch, validator = shortLog(proposerKey),
+                    source = "signature", elapsed_time = timeElapsed,
+                    delay = vc.getDelay(deadline)
+  signature
+
+proc getRandao(vc: ValidatorClientRef, slot: Slot,
+               proposerKey: ValidatorPubKey) {.async.} =
+  if slot == Slot(0'u64):
```
```suggestion
  if slot == GENESIS_SLOT:
```

beacon_chain/validator_client/block_service.nim
```diff
@@ -125,6 +127,95 @@ proc lazyWait[T](fut: Future[T]) {.async.} =
   except CatchableError:
     discard
 
+proc setSignature(item: var RandaoCacheItem, slot: Slot, key: ValidatorPubKey,
+                  signature: ValidatorSig) =
+  item.data[int(slot.since_epoch_start())] =
```
tersec: This doesn't compile with Nim 2.0:
```
2023-10-10T13:37:30.5254048Z /github-runner/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/validator_client/block_service.nim(132, 12) Error: type mismatch
2023-10-10T13:37:30.5254534Z Expression: `[]=`(item.data, int(since_epoch_start(slot)),
2023-10-10T13:37:30.5255223Z       RandaoPair(pubkey: key, signature: signature))
2023-10-10T13:37:30.5255533Z   [1] item.data: array[0..31, RandaoPair]
2023-10-10T13:37:30.5360453Z   [2] int(since_epoch_start(slot)): int
2023-10-10T13:37:30.5367874Z   [3] RandaoPair(pubkey: key, signature: signature): RandaoPair
2023-10-10T13:37:30.5368091Z
2023-10-10T13:37:30.5368245Z Expected one of (first mismatch at [position]):
2023-10-10T13:37:30.5368571Z [0] func `[]=`(a: var HashArray; b: auto; c: auto)
2023-10-10T13:37:30.5368889Z [0] func `[]=`(s: var BitSeq; pos: Natural; value: bool)
2023-10-10T13:37:30.5369225Z [0] func `[]=`(x: var HashList; idx: auto; val: auto)
2023-10-10T13:37:30.5532453Z [0] func `[]=`[T, U, V](r: var openArray[T]; s: HSlice[U, V]; v: openArray[T])
2023-10-10T13:37:30.5532875Z [0] proc `[]=`(s: var string; i: BackwardsIndex; x: char)
2023-10-10T13:37:30.5533329Z [0] proc `[]=`[A, B](t: OrderedTableRef[A, B]; key: A; val: sink B)
2023-10-10T13:37:30.5533732Z [0] proc `[]=`[A, B](t: TableRef[A, B]; key: A; val: sink B)
2023-10-10T13:37:30.5611862Z [0] proc `[]=`[A, B](t: var OrderedTable[A, B]; key: A; val: sink B)
2023-10-10T13:37:30.5612269Z [0] proc `[]=`[A, B](t: var Table[A, B]; key: A; val: sink B)
2023-10-10T13:37:30.5612644Z [0] proc `[]=`[A](t: CountTableRef[A]; key: A; val: int)
2023-10-10T13:37:30.5612993Z [0] proc `[]=`[A](t: var CountTable[A]; key: A; val: int)
2023-10-10T13:37:30.5613323Z [0] proc `[]=`[I: Ordinal; T, S](a: T; i: I; x: sink S)
2023-10-10T13:37:30.5613669Z [0] proc `[]=`[Idx, T; U, V: Ordinal](a: var array[Idx, T]; x: HSlice[U, V];
2023-10-10T13:37:30.5773584Z                                   b: openArray[T])
2023-10-10T13:37:30.5773977Z [0] proc `[]=`[Idx, T](a: var array[Idx, T]; i: BackwardsIndex; x: T)
2023-10-10T13:37:30.5774374Z [0] proc `[]=`[T, U: Ordinal](s: var string; x: HSlice[T, U]; b: string)
2023-10-10T13:37:30.5774795Z [0] proc `[]=`[T; U, V: Ordinal](s: var seq[T]; x: HSlice[U, V]; b: openArray[T])
2023-10-10T13:37:30.5775171Z [0] proc `[]=`[T](aq: AsyncQueue[T]; i: BackwardsIndex; item: T)
2023-10-10T13:37:30.5775517Z [0] proc `[]=`[T](aq: AsyncQueue[T]; i: Natural; item: T)
2023-10-10T13:37:30.5775850Z [0] proc `[]=`[T](s: var openArray[T]; i: BackwardsIndex; x: T)
2023-10-10T13:37:30.5776655Z [0] template `[]=`(a: var BitArray; pos: Natural; value: bool)
2023-10-10T13:37:30.5777077Z [0] template `[]=`(epochFlags: EpochParticipationFlags; idx: ValidatorIndex;
2023-10-10T13:37:30.5777482Z                flags: ParticipationFlags)
2023-10-10T13:37:30.5777822Z [0] template `[]=`(s: string; i: int; val: char)
2023-10-10T13:37:30.5778159Z [0] template `[]=`(x: var BitList; idx: auto; val: bool)
2023-10-10T13:37:30.5778491Z [0] template `[]=`(x: var List; idx: auto; val: auto)
2023-10-10T13:37:30.5778823Z [0] template `[]=`[T](a: var seq[T]; b: ValidatorIndex; c: T)
```

likely because Nim 2.0 is stricter about the indexing type of an array being equivalent at declaration and usage, e.g., if it's defined as a `uint64`, not an `int`:
```
beacon_chain/spec/presets/mainnet/phase0_preset.nim
46:  SLOTS_PER_EPOCH* {.intdefine.}: uint64 = 32

beacon_chain/spec/presets/minimal/phase0_preset.nim
46:  SLOTS_PER_EPOCH* {.intdefine.}: uint64 = 8

beacon_chain/spec/presets/gnosis/phase0_preset.nim
46:  SLOTS_PER_EPOCH* {.intdefine.}: uint64 = 16
```

A small, isolated example which similarly compiles with Nim 1.6 but not Nim 2.0:
```nim
const SLOTS_PER_EPOCH: uint64 = 32
var a: array[SLOTS_PER_EPOCH, string]
a[int(1)] = ""
```

for the same reason, it needs a `uint64` index or something implicitly convertible to one.

beacon_chain/validator_client/block_service.nim
```diff
@@ -125,6 +127,95 @@ proc lazyWait[T](fut: Future[T]) {.async.} =
...
+    RandaoPair(pubkey: key, signature: signature)
+
+proc getSignature(item: RandaoCacheItem, slot: Slot,
+                  key: ValidatorPubKey): Opt[ValidatorSig] =
+  let pitem = item.data[int(slot.since_epoch_start())]
```
tersec: This will have the same Nim 2.0 issue as already noted due to the `int` vs `uint64` mismatch which Nim 1.6 overlooks.

tersec (approved):

beacon_chain/validator_client/common.nim
tersec: since this can potentially change iif/when the VC connects to different BNs, this seems useful enough.

beacon_chain/validator_client/common.nim
```diff
@@ -227,6 +236,7 @@ type
     validatorsRegCache*: Table[ValidatorPubKey, SignedValidatorRegistrationV1]
     blocksSeen*: Table[Slot, BlockDataItem]
     rootsSeen*: Table[Eth2Digest, Slot]
+    randaoCache*: Table[Epoch, RandaoCacheItem]
```
arnetheduck: this looks like the wrong place for this cache - ie we already have a similar mechanism for slot signatures - this code should live in `Validator` just like the slot signature does and be implemented for both beacon and client validators equally - see https://github.com/status-im/nimbus-eth2/blob/ed446b1d5a81a191f20b39cab70cba4e63a1119f/beacon_chain/validators/validator_pool.nim#L764

cheatfate: Such approach allows more easy maintenance: update on reorg and pruning, while keep this data inside Validator requires much more data to be stored to check if specific signature could be actually used or should be recalculated.

Also VC does not use `Validator.slotSignature` because VC obtains slot signatures at the beginning of the epoch (Obol requirement), and so it has at least `SLOTS_PER_EPOCH` signatures - not the single one.

arnetheduck: well, the BN and VC should be behaving the same in this regard and ValidatorPool is the place for such code - clearing here doesn't really matter, ie we can afford to keep a signature per validator without cleanup (just like slot signatures) - there should be no reorg issues since what we're signing is reorg-independent (just like slot sig) meaning that the slot signature code is correct for epoch signatures too - in fact, the implementation is much more simple if we move it into VP.

Re slot sigs, it would be ideal that the obol support uses the same mechanism as well so we don't duplicate them between

cheatfate: Obol mechanism do not use same mechanism because it was already 3rd iteration of this slot signatures approach and it also helps maintenance and could help implement any changes in Obol protocol/behavior. Moving this slot signatures into `Validator` incredibly complicate all the things.

cheatfate: And one more thing to note, BN do not need to support Obol, because Obol is for scheme `VC->OBOL->BN` and never be used in scheme `BN->OBOL-BN`

arnetheduck: ok for obol / slot sigs then - however, the bn validators will also benefit from epoch sig caching, so it seems prudent to unify the code and behavior

arnetheduck (approved):

PR #5510, opened 2023-10-17
Bump nim-eth: Change block timestamp from std.Time to distinct uint64

jangko:
Also change tx.maxFeePerBlobGas from GasInt to UInt256 following Cancun latest spec.

Both BlockHeader.timestamp and Transaction.maxFeePerBlobGas receive extensive tests using Cancun test vectors. The stricter test demand those two fields type match the type in the spec.


cheatfate:
I'm curious why time is represented as `uint64` why not use `int64`? `int64` is totally enough.

jangko:
`int64` is enough until they generate test vectors with block timestamp overflow `int64`. and looks like other EL clients doing the same thing, they use `uint64`.

PR #5525, opened 2023-10-24
RFC: remove phase0 and altair REST JSON block deserialization

tersec:
Keeps writing, because that's required for the REST server to serve historical blocks.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -1048,30 +1048,6 @@ proc readValue*[BlockType: ForkedBeaconBlock](
   prepareForkedBlockReading(reader, version, data, "ForkedBeaconBlock")
 
   case version.get():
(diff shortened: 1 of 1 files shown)


etan-status:
What about a client reading archive data?

tersec:
> What about a client reading archive data?

Existing databases, i.e. local archives to a particular BN, are untouched by this. This is strictly a set of changes to the REST reading support.

Furthermore, because the REST server with these changes can still write phase0 and altair blocks in response to queries, clients looking for archive data from a BN can still get it.

tersec:
In general, the purpose of the reading support in the REST server and validator client was/is to create new blocks, but there's been no need to create new phase0 or altair blocks in years.

arnetheduck:
Using eth2 as a library to consume data is also impacted (like a UI showing blocks, like the GUI I maintain) - the cost is low to non-existent, I don't see the point of actively removing this.

tersec:
Alright. Worth at least making this an explicit/intentional decision.

PR #5538, opened 2023-10-28
rm unused code

tersec:
Everything here was noted by Nim's `XDeclaredButNotUsed` hints.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -723,11 +723,7 @@ proc initLightClientDataCache*(dag: ChainDAGRef) =
     blocks.add bid
 
   # Process blocks (reuses `dag.headState`, but restores it to the current head)
-  var
-    tmpState = assignClone(dag.headState)
-    tmpCache, cache: StateCache
-    oldCheckpoint: Checkpoint
-    cpIndex = 0
+  var cache: StateCache
   for i in countdown(blocks.high, blocks.low):
     bid = blocks[i]
     if not dag.updateExistingState(
--- a/beacon_chain/el/el_conf.nim
+++ b/beacon_chain/el/el_conf.nim
@@ -71,9 +71,6 @@ func unknownRoleMsg(role: string): string =
 template raiseError(reader: var TomlReader, msg: string) =
   raiseTomlErr(reader.lex, msg)
 
-template raiseError(reader: var JsonReader, msg: string) =
-  raiseTomlErr(reader.lex, msg)
-
 proc readValue*(reader: var TomlReader, value: var EngineApiRoles)
                {.raises: [SerializationError, IOError].} =
   let roles = reader.readValue seq[string]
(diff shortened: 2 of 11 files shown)


beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -723,11 +723,7 @@ proc initLightClientDataCache*(dag: ChainDAGRef) =
     blocks.add bid
 
   # Process blocks (reuses `dag.headState`, but restores it to the current head)
-  var
-    tmpState = assignClone(dag.headState)
```
tersec, line 727: This was odd; @etan-status the comment states that it "reuses `dag.headState`, but restores it to the current head", but it does/did not appear to do that.

Is that intentional?

beacon_chain/el/el_manager.nim
```diff
@@ -1810,10 +1791,6 @@ func new*(T: type ELConnection,
     engineUrl: engineUrl,
     depositContractSyncStatus: DepositContractSyncStatus.unknown)
 
-template getOrDefault[T, E](r: Result[T, E]): T =
```
tersec, line 1813: I don't 100%  understand the intent here either -- is this meant as a generic `getOrDefault` override which can exist separately from `Table`, etc?

beacon_chain/el/el_conf.nim
```diff
@@ -71,9 +71,6 @@ func unknownRoleMsg(role: string): string =
 template raiseError(reader: var TomlReader, msg: string) =
   raiseTomlErr(reader.lex, msg)
 
-template raiseError(reader: var JsonReader, msg: string) =
-  raiseTomlErr(reader.lex, msg)
```
tersec, line 75: Sort of funny, this has `raiseTomlErr` for a `JsonReader` function. Looks like incomplete copy/paste.

beacon_chain/el/el_manager.nim
```diff
@@ -2182,7 +2152,7 @@ proc startChainSyncingLoop(m: ELManager) {.async.} =
         continue
 
       await syncEth1Chain(m, syncedConnectionFut.read)
-    except CatchableError as err:
+    except CatchableError:
```
tersec, line 2155: As the comment two lines later notes, "A more detailed error is already logged by trackEngineApiRequest".

beacon_chain/networking/network_metadata.nim
```diff
@@ -474,7 +474,7 @@ when const_preset in ["mainnet", "gnosis"]:
           toOpenArray(metadata.genesis.bakedBytes, 0, sizeof(BeaconStateHeader) - 1),
           BeaconStateHeader)
         Opt.some header.genesis_validators_root
-      except SerializationError as err:
+      except SerializationError:
```
tersec, line 477: Reasonable given that it `raiseAssert`s a line later; this is not some log message where additional information might be useful.

beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
etan-status: Logic was moved from a different place in https://github.com/status-im/nimbus-eth2/pull/3799 that required `tmpState`. The new logic here uses directly `dag.headState` with `dag.updateExistingState` and the loop makes sure it will end up at its old value. So the `tmpState` clone is no longer needed, and the comment is correct.

PR #5549, opened 2023-11-01
libp2p: revert scoring

arnetheduck:
New IWANT duplicate handling contains a down-scoring feature which is suspected of causing gossip mesh problems on certain setups - this PR rolls back that scoring function


diegomrsantos:
In this branch nimbus is using this libp2p branch https://github.com/status-im/nim-libp2p/tree/b2eac7e-and-revert-c6aa085

PR #5550, opened 2023-11-01
fill blindedblockbody KZG commitments from deneb blinded header

tersec:


tersec:
Well, this improves things.

PR #5562, opened 2023-11-04
wrap `kzgs`/`proofs`/`blobs` fields as `BlobsBundle`

etan-status:
Less type conversion / copying by keeping the `BlobsBundle` together.

### DIFF
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -392,21 +392,13 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
             qslot, proposer, qrandao, qskip_randao_verification):
           return RestApiResponse.jsonError(Http400, InvalidRandaoRevealValue)
 
-        let res =
-          case node.dag.cfg.consensusForkAtEpoch(qslot.epoch)
-          of ConsensusFork.Deneb:
+        let res = withConsensusFork(
+            node.dag.cfg.consensusForkAtEpoch(qslot.epoch)):
+          when consensusFork >= ConsensusFork.Bellatrix:
             await makeBeaconBlockForHeadAndSlot(
-              deneb.ExecutionPayloadForSigning,
+              consensusFork.ExecutionPayloadForSigning,
               node, qrandao, proposer, qgraffiti, qhead, qslot)
-          of ConsensusFork.Capella:
-            await makeBeaconBlockForHeadAndSlot(
-              capella.ExecutionPayloadForSigning,
-              node, qrandao, proposer, qgraffiti, qhead, qslot)
-          of ConsensusFork.Bellatrix:
-            await makeBeaconBlockForHeadAndSlot(
-              bellatrix.ExecutionPayloadForSigning,
-              node, qrandao, proposer, qgraffiti, qhead, qslot)
-          of ConsensusFork.Altair, ConsensusFork.Phase0:
+          else:
             return RestApiResponse.jsonError(Http400, InvalidSlotValueError)
         if res.isErr():
           return RestApiResponse.jsonError(Http400, res.error())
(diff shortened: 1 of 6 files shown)


beacon_chain/validators/beacon_validators.nim
```diff
@@ -1188,18 +1183,13 @@ proc proposeBlockAux(
...
           )
           sidecars.add(sidecar)
         Opt.some(sidecars)
-      elif forkyBlck is phase0.BeaconBlock or
-           forkyBlck is altair.BeaconBlock or
-           forkyBlck is bellatrix.BeaconBlock or
-           forkyBlck is capella.BeaconBlock:
-        Opt.none(seq[BlobSidecar])
       else:
-        static: doAssert "Unknown BeaconBlock type"
```
tersec, line 1202: Switching `static: doAssert` to `Opt.none` means that it won't trigger automatically anymore when the E-fork is added, but will incorrectly and silently use the pre-Deneb/sidecar version.

Better to find out about missing functionality at compile-time.

etan-status: It will take the Deneb version, not the `none` version (`consensusFork >= ConsensusFork.Deneb`), and will compile-time fail if no longer compatible

beacon_chain/spec/datatypes/deneb.nim
```diff
@@ -38,6 +38,7 @@ const
 
 type
   KzgCommitments* = List[KzgCommitment, Limit MAX_BLOB_COMMITMENTS_PER_BLOCK]
+  KzgProofs* = List[KzgProof, Limit MAX_BLOB_COMMITMENTS_PER_BLOCK]
```
tersec, line 41: Are these `KzgProofs` ever sent over the wire or `hash_tree_root`ed in this form? The blinded versions are, but I'm not sure when the Merkleization changes here are useful. Individual `KzgProof`s of course are, but not as part of a (non-blinded) `KzgProofs`.

If they're not, I'm not sure it's an improvement to have this `KzgProofs` type rather than the only slightly more verbose but also more explicit `seq[KzgProof]`. With the `List`, sure, but aside from I'm not sure it seems as strongly motivated.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -435,18 +427,13 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
               `block`: forkyBlck,
               blob_sidecars: List[BlobSidecar,
                                   Limit MAX_BLOBS_PER_BLOCK].init(sidecars))
-          elif forkyBlck is phase0.BeaconBlock or
-               forkyBlck is altair.BeaconBlock or
-               forkyBlck is bellatrix.BeaconBlock or
-               forkyBlck is capella.BeaconBlock:
-            forkyBlck
           else:
-            static: raiseAssert "produceBlockV2 received unexpected version"
```
tersec, line 444: Same comment as elsewhere -- this will silently break the E-fork

beacon_chain/spec/datatypes/deneb.nim
etan-status: They will have to be sent, for block production, via beacon-APIs and builder-API.

- https://github.com/ethereum/beacon-APIs/pull/370
- https://github.com/ethereum/builder-specs/pull/90

Right now, they are not, but updating the VC stub implementation for it to have something functional, and requires the limit to be set for successful compile.

beacon_chain/validators/beacon_validators.nim
etan-status: The other one will compile-time break on Electra because `DenebBlockContents` requires a Deneb block, while the generated code will try to put an Electra block into it.

This instance here will be fine with Electra as well (using the blob logic from deneb)

tersec (approved):

Issue #5557, opened 2023-11-03
per-chunk validation of libp2p req/resp chunk


etan-status:
> Before consuming the next response chunk, the response reader SHOULD verify the blob sidecar is well-formatted, has valid inclusion proof, and is correct w.r.t. the expected KZG commitments through `verify_blob_kzg_proof`.

When fetching `BlobSidecar` via libp2p req/resp, `v1.4.0-beta.4` spec now suggests incrementally validating chunks before processing next one.

In Nimbus we currently don't have infrastructure to process chunks while followup parts of the response are still pending download.

PR #5565, opened 2023-11-04
update Deneb for blob sidecar inclusion proofs

etan-status:
`BlobSidecar` is no longer signed, instead use Merkle proof to link blobs with block.

- https://github.com/ethereum/consensus-specs/pull/3531

Associated beacon-API / builder-specs still TBD; minimal changes done to compile in similar style to previous spec, but not standardized yet.

- https://github.com/ethereum/beacon-APIs/pull/369
- https://github.com/ethereum/builder-specs/pull/90


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -302,93 +316,132 @@ template validateBeaconBlockBellatrix(
...
+      dag.forkAtEpoch(block_header.slot.epoch),
+      getStateField(dag.headState, genesis_validators_root),
+      block_header.slot,
+      block_root,
+      dag.validatorKey(proposer).get(),
+      blob_sidecar.signed_block_header.signature):
+    return dag.checkedReject("BlobSidecar: Invalid proposer signature")
+
+  # [REJECT] The sidecar's blob is valid as verified by `verify_blob_kzg_proof(
+  # blob_sidecar.blob, blob_sidecar.kzg_commitment, blob_sidecar.kzg_proof)`.
```
etan-status, line 436: Double-checked with @mratsim , the proposer sig check is the faster one, so doing it before the KZG check: "pairings are 2 levels deep in KZG and 1 deep in BLS signature (just after deserialization and sanity checks that points are on curve and non-infinity)"

Batching of the two verifications would be clunky, likely not worth it.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -407,26 +407,11 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-              )
-              sidecars.add(sidecar)
-
+            let blobsBundle = message.blobsBundleOpt.get()
             DenebBlockContents(
               `block`: forkyBlck,
-              blob_sidecars: List[BlobSidecar,
-                                  Limit MAX_BLOBS_PER_BLOCK].init(sidecars))
+              kzg_proofs: blobsBundle.proofs,
+              blobs: blobsBundle.blobs)
```
etan-status, line 414: Sidecars are no longer sent back and forth, as they no longer need a signature. Instead, just the `kzg_proofs` and `blobs` are sent.

- https://github.com/ethereum/beacon-APIs/pull/369

`kzg_proofs` and `blobs` are downloaded by the VC and then submitted again to the BN to support use case where the block is produced by a different BN than the one who is broadcasting. The BN then converts the `SignedDenebBlockContents` to `SignedBeaconBlock` + `BlobSidecar` list

beacon_chain/spec/datatypes/deneb.nim
```diff
@@ -107,6 +93,16 @@ type
     blob_gas_used*: uint64   # [New in Deneb]
     excess_blob_gas*: uint64 # [New in Deneb]
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.4/specs/deneb/validator.md#blobsbundle
+  KzgProofs* = List[KzgProof, Limit MAX_BLOB_COMMITMENTS_PER_BLOCK]
+  Blobs* = List[Blob, Limit MAX_BLOB_COMMITMENTS_PER_BLOCK]
+  BlobRoots* = List[Eth2Digest, Limit MAX_BLOB_COMMITMENTS_PER_BLOCK]
```
etan-status, line 99: Not sure where to put them and what to put as the correct limit. beacon-API uses 0 .. MAX_BLOBS_PER_BLOCK in informal metadata description, builder-API uses MAX_BLOB_COMMITMENTS_PER_BLOCK in a more explicit specification. In practice, it doesn't matter as the SSZ messages are not hash-tree-rooted in neither of the two protocols. If they were hash_tree_rooted, using the theoretical capacity (MAX_BLOB_COMMITMENTS_PER_BLOCK) makes more sense (that is also what they use for the KzgCommitments inside consensus, where HTR matters).

beacon_chain/spec/mev/deneb_mev.nim
```diff
@@ -69,25 +69,18 @@ type
 
   # https://github.com/ethereum/builder-specs/blob/534e4f81276b8346d785ed9aba12c4c74b927ec6/specs/deneb/builder.md#blindedblobsidecar
   BlindedBlobSidecar* = object
-    block_root*: Eth2Digest
```
etan-status, line 72: `BlindedBlobSidecar` may go away entirely, but the specs are still in flux:

- https://github.com/ethereum/builder-specs/pull/90

For now, have done the minimal changes to it to make the flow work in theory, but the spec may optimize further and get rid of the `BlindedBlobSidecar` altogether.

etan-status:
This should be all that's needed to get us going with the new blob sidecar flavor, as an initial PR upgrading to v1.4.0-beta.4 test vectors.

- Remove `SignedBlobSidecar`, and replace with new `BlobSidecar` conceptually
- Change `BlobSidecar`, and replace with `kzg_proofs` + `blobs` lists conceptually
- BN converts `signed_block` (`kzg_commitments`) + `kzg_proofs` + `blobs` lists to `BlobSidecar` and puts in cryptographic link via Merkle proof, instead of VC putting in a signature
- Beacon-APIs updated for new flow
- Builder-APIs flow updated minimally, it still uses `BlindedBlobSidecar` -> `BlobSidecar` flow; may become simpler in the future

Notably missing:

- Final beacon-API and builder-API specs
- Gossip per-chunk validation: https://github.com/status-im/nimbus-eth2/issues/5557
- I think there could also be more sanity checking, but that's not a new problem. As in, any time one receives via REST a signed block, validate (minimally) that KZG proofs are correct, and that inclusion proofs for sidecars are correct. Helper functions are there. Not high prio as it only affects trusted REST API I think.
- Same for VC, could check that proofs are valid before signing a block.

tersec (approved):

PR #5568, opened 2023-11-04
extend `hasBlob` to all blocks from same proposer and slot

etan-status:
`v1.4.0-beta.4` made the Gossip rules more strict and now requires to ignore blobs from other branches if there are equivocating blocks. Those blobs are only requestable via Req/Resp.


tersec (approved):
This function is also used in `processSignedBlobSidecar` in `gossip_processing/eth2_processor`, but the modified functionality/criteria are fine there -- it's basically a dupe detector for a blob it needs, which is done fine regardless.

etan-status:
> This function is also used in `processSignedBlobSidecar`

not anymore, that usage was incorrect and redundant:

- https://github.com/status-im/nimbus-eth2/pull/5563

PR #5575, opened 2023-11-07
use newer geth testnet binary directory names

tersec:


.github/workflows/ci.yml
```diff
@@ -215,7 +215,7 @@ jobs:
         if: ${{ !cancelled() }} && github.event_name == 'pull_request'
         run: |
           excluded_files="config.yaml"
-          excluded_extensions="ans|json|md|png|service|ssz|txt"
+          excluded_extensions="ans|json|md|png|service|ssz|txt|json.template"
```
etan-status: maybe needs a \ before . as this gets pasted into the regex on line 226.
also, seems to have been alphabetically ordered so far

etan-status: or \\ because it's within " (or change " to '

tersec: https://github.com/status-im/nimbus-eth2/pull/5575/commits/bb940d13b8548b6bb80a2df2ee03bde3f05b5e4a

etan-status (approved):

PR #5576, opened 2023-11-07
enable incremental LTO on macOS

etan-status:
Set `-cache_path_lto` for incremental LTO on macOS.

https://clang.llvm.org/docs/ThinLTO.html


arnetheduck:
this looks like something you may want to add to `~/.config/nim/nim.cfg` instead - ie lto caches are not without their own problems in terms of cold compile time increases, hidden disk usage etc

PR #5590, opened 2023-11-10
allow higher `MIN_EPOCHS_FOR_BLOCK_REQUESTS` than safe minimum

etan-status:
Gnosis uses `MIN_EPOCHS_FOR_BLOCK_REQUESTS` = 33024, but the computed safe minimum (that Nimbus was using) is 2304. Relax the compatibility check to allow `MIN_EPOCHS_FOR_BLOCK_REQUESTS` above the safe minimum and honor `config.yaml` preferences for `MIN_EPOCHS_FOR_BLOCK_REQUESTS`.


tersec:
Why?

PR #5592, opened 2023-11-10
bump `gnosis-chain-configs` to `cc32289688ddc8e2ad539c390b8dcdf785796452`

etan-status:
- Remove prerelease keys


etan-status:
ah, upstream isn't ready yet: 

- https://github.com/gnosischain/configs/pull/23

PR #5602, opened 2023-11-15
send LC finality update on event stream on supermajority

### LINKED ISSUE #5491: beacon rpc sometimes returns LightClientFinalityUpdate with sync_committee length < 2/3
**Describe the bug**
When requesting a `LightClientFinalityUpdate` via the beacon_chain RPC the endpoints sometimes returns an update with only a few entries in the sync committee (as shown in the screenshots). I'm not sure if this is a bug in Nimbus, but I can't think of anything where such a FinalityUpdate would be useful.

According to the consensus specs light clients should:
- check the validity of such updates (see [`process_light_client_update`](https://ethereum.github.io/consensus-specs/specs/altair/light-client/sync-protocol/#process_light_client_update)),
- update `store.best_valid_update`
- update `store.current_max_active_participants` if the update has more participants set (which should never get this low),
- update the optimistic header (if the light client wants this updated asap he'd use `LightClientOptimisticUpdate` and
- apply the update (and thus the finalized header) if the sync committee is large enough (2/3).

Please correct me if I understood something wrong in the consensus specs or if there is another reason why `LightClientFinalityUpdate` with less than 2/3 of the sync committee would be useful. My current best guess is that it isn't useful at all and probably shouldn't be returned by the `/eth/v1/beacon/light_client/finality_update` rpc endpoint. I can easily check for the sync committee size to avoid this issue, instead of just checking the finalized blocknumber, which I mainly do to avoid unnecessary communication internally (I don't care about the optimistic updates/blocks/data).

I don't know if this also affects the subscription (I'll likely change it to use the subscription at some point, but for now I'm polling the rpc endpoint).

It could of course also be that my script is wrong and doesn't decode the sync committee correctly, hence the question of whether I understood something wrong when reading the consensus-specs.

**To Reproduce**
1. Platform details (OS, architecture): Arch Linux, x86 (shouldn't be relevant since this is about the API and I'm using the http://unstable.prater.beacon-api.nimbus.team/ node.
2. Branch/commit used: N/A (I don't know which version that node uses currently, I've tested it today).
3. Commands being executed: Custom code that polls this RPC endpoint every 12 seconds, prints the sync committee size of the update 

I additionally gives these updates to a light client if the finalized block number changed. Sometimes this then didn't apply the update, which is how I found this.

**Screenshots (logs from my application)**
Log output of an application polling the RPC endpoint every 12 seconds and outputting the sync committee size: 
2023-10-10-192557_623x596_scrot

Log output showing the situation I've found: A finality update with <2/3 sync_committee (the first one I received with a new `finalized_header..execution.block_number`), together with debug logs of the light client showing that the sync committee was the reason that the update was not applied.
2023-10-10-195051_1619x304_scrot

**Additional Info**
This seems to happen quite regularly, most Epochs give me a FinalityUpdate with less than 2/3 sync committees.

etan-status:
When new finality is reached without supermajority sync committee support, trigger another event push on beacon-API and libp2p once the finality gains supermajority support.

- https://github.com/ethereum/consensus-specs/pull/3549

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -563,6 +563,11 @@ proc createLightClientUpdates(
     var finalized_slot = attested_data.finalized_slot
     if finalized_slot == forkyLatest.finalized_header.beacon.slot:
       forkyLatest.finality_branch = attested_data.finality_branch
+      let old_num_active_participants =
+        forkyLatest.sync_aggregate.num_active_participants.uint64
+      if not hasSupermajoritySyncParticipation(old_num_active_participants) and
+          hasSupermajoritySyncParticipation(num_active_participants):
+        newFinality = true
     elif finalized_slot < dag.tail.slot or
         not load_finalized_bid(finalized_slot):
       forkyLatest.finalized_header.reset()
--- a/beacon_chain/consensus_object_pools/light_client_pool.nim
+++ b/beacon_chain/consensus_object_pools/light_client_pool.nim
@@ -19,6 +19,10 @@ type
       ## Latest finality update that was forwarded on libp2p gossip.
       ## Tracks `finality_update.finalized_header.beacon.slot`.
 
+    latestForwardedFinalityHasSupermajority*: bool
+      ## Whether or not the latest finality update that was forwarded on
+      ## libp2p gossip had supermajority (> 2/3) sync committee participation.
+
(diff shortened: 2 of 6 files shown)


tersec:
Should this PR depend on the underlying spec PR being merged?

etan-status:
Shouldn't matter that much, on this one. May lead to a tiny increase in sending gossip that gets ignored (not rejected) by other nimbuses, when sync participation is low while at the same time finality is advancing.

tersec (approved):

Issue #5599, opened 2023-11-14
Since 23.10.0, nimbus-eth2 won't send registrations for validators attached to a validator client anymore


stephanep:
**Describe the bug**
Potential regression introduced in 23.10.0 (likely by https://github.com/status-im/nimbus-eth2/pull/5431).
When using validators on a remote nimbus_validator_client, and with no validator directly attached to nimbus_beacon_node, registration for those validators are not sent by nimbus_beacon_node to mev-boost anymore, after upgrading from 23.9.1 to 23.10.0.

**To Reproduce**
I have a solo staking setup comprised of an nimbus_validator_client with a few validators, attached to two nimbus_beacon_node.
Each nimbus_beacon_node connects to a local mevboost instance (latest v1.16 from flashbots).
I've configured MEV as described in https://nimbus.guide/external-block-builder.html.

In particular:
- i DO NOT have any validator directly attached to the beacon node;
- and i'm NOT using multiple per-validator builder preferences.

Said otherwise, i run my nimbus_validator_client with `--payload-builder=true`, my nimbus_beacon_node with `--payload-builder=true --payload-builder-url=http://127.0.0.1:18550` and that's about it.

Everything was working fine when i was using 23.9.1 and earlier releases, the registrations for all validators were successfully sent to mevboost and taken into account by the configured relays.

I then upgraded the beacon nodes and validator client to 23.10.0, and following this, added a couple of new validators to nimbus_validator_client.

After investigating why i failed to use MEV for a block proposed by one of the new validators, i noticed that those had never been registered successfully, contrary to the the older ones (added when i was still running 23.9.1 or earlier).

I am certain (from debug logs and traffic analysis) that nimbus_validator_client correctly builds and sends the registration messages to the /eth/v1/validator/register_validator endpoint of the beacon node API, that the POST request is correctly received by the beacon node and replied with 200 OK.

```
DBG 2023-11-14 02:58:35.033+01:00 Opened connection to remote server         remote=X.X.X.X:XXXX request=/eth/v1/validator/register_validator http_method=POST
DBG 2023-11-14 02:58:35.042+01:00 REST request body has been sent            remote=X.X.X.X:XXXX request=/eth/v1/validator/register_validator size=XXXX http_method=POST
DBG 2023-11-14 02:58:35.104+01:00 Got REST response headers from remote server remote=X.X.X.X:XXXX request=/eth/v1/validator/register_validator status=200 http_method=POST
[...]

PR #5603, opened 2023-11-15
add registration of VC validators back in the new per validator management of builders

stephanep:
Fix handling of VC validators by per validator payload-builder registration (issue #5599).

The proposed change ensures that VC validators are registered with the builder specified by the `--payload-builder-url` argument even if the beacon node has no attached validators. It also prevent such validators from being unintentionally registered with builders configured for specific attached validators by the keymanager api.

### DIFF
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -1575,8 +1575,11 @@ proc registerValidatorsPerBuilder(
         validatorRegistrations.add @[validatorRegistration]
 
     # First, check for VC-added keys; cheaper because provided pre-signed
+    # See issue #5599: currently VC have no way to provide BN with per-validator builders per the specs, so we have to
+    #   resort to use the BN fallback default (--payload-builder-url value, obtained by calling getPayloadBuilderAddress)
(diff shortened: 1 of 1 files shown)


beacon_chain/validators/beacon_validators.nim
```diff
@@ -1576,7 +1576,8 @@ proc registerValidatorsPerBuilder(
 
     # First, check for VC-added keys; cheaper because provided pre-signed
     var nonExitedVcPubkeys: HashSet[ValidatorPubKey]
-    if node.externalBuilderRegistrations.len > 0:
+    if  node.externalBuilderRegistrations.len > 0 and
+        payloadBuilderAddress == node.config.getPayloadBuilderAddress.value:
```
tersec, line 1582: `node.config.getPayloadBuilderAddress` is a BN fallback default, which probably should not be showing up in `registerValidatorsPerBuilder`, except for the discussion at https://github.com/status-im/nimbus-eth2/issues/5599#issuecomment-1812894337 and on its face looks like a bug, rather than something intentional; perhaps worth copying some of the analysis from that issue comment into the source code, where it won't be lost.

spoignant86: ok, added to the commit

PR #5604, opened 2023-11-16
Batch slashing protection registration

arnetheduck:
This PR brings down the time to send 100 attestations from ~1s to ~100ms, making it feasible to run 10k validators on a single node (which regularly send 300 attestations / slot).

This is done by batching the slashing protection database write in a single transaction thus avoiding a slow fsync for every signature - effects will be more pronounced on slow drives.

The benefit applies both to beacon and client validators.

### DIFF
--- a/beacon_chain/validators/slashing_protection.nim
+++ b/beacon_chain/validators/slashing_protection.nim
@@ -154,7 +154,7 @@ proc checkSlashableBlockProposal*(
   ## The error contains the blockroot that was already proposed
   ##
   ## Returns success otherwise
-  checkSlashableBlockProposal(db.db_v2, some(index), validator, slot)
+  checkSlashableBlockProposal(db.db_v2, Opt.some(index), validator, slot)
 
 proc checkSlashableAttestation*(
        db: SlashingProtectionDB,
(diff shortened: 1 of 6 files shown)


beacon_chain/validators/slashing_protection.nim
```diff
@@ -197,9 +197,22 @@ proc registerAttestation*(
   ##
   ## attestation_signing_root is the output of
   ## compute_signing_root(attestation, domain)
-  registerAttestation(db.db_v2, some(index), validator,
+  registerAttestation(db.db_v2, Opt.some(index), validator,
       source, target, attestation_signing_root)
 
+template withContext*(db: SlashingProtectionDB, body: untyped): untyped =
```
tersec, line 203: Doesn't the v1 slashing protection not ever write anymore? My understanding had been that it was there for read-only migration purposes only.

arnetheduck: this is not v1 - it's a leftover of the v1/v2 abstraction scheme (that technically could be cleaned up)

tersec (approved):

PR #5608, opened 2023-11-19
log engine/builder API decisionmaking

tersec:
If this PR is to be useful, it's not clear to me whether it has as much value at `debug` as `info`.

The general goal here is to clarify for for Nimbus end-users and for developers if/when Nimbus is properly comparing engine/builder bids and when it's inadvertently selecting one because the other didn't work. It furthermore shows even when selecting one, if both arrived, the value of the other bid as well.

I don't view this as critical functionality, but might be nice to have, and came up during some discussions/apropos concerns about what happened when Nimbus sometimes did not use a configured builder AP relay but fell back to the engine API

General reasons this can occur are relays being down, or more commonly, returning HTTP 204, indicating nothing interesting to suggest for that slot. It can also happen that they do suggest some block (header), but it's not particularly valuable.

The latter becomes potentially more interesting as the local block value boost is increased, because usually in a direct comparison, the builders and relays will create higher-value blocks than local EL clients, but might not if they have to be 15%+ higher-value.

### DIFF
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -1093,6 +1093,38 @@ proc proposeBlockAux(
         return head   # errors logged in router
       false
 
+  # There should always be an engine bid, and if payloadBuilderClient exists,
+  # not getting a builder bid is also an error. Do not report lack of builder
+  # when that's intentional. Replicate some of the nested if statements here,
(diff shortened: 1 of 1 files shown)


tersec:
Of note, Lighthouse has something like this: https://github.com/sigp/lighthouse/pull/4352

PR #5612, opened 2023-11-21
Dissect Windows specific code from beacon node.

cheatfate:
This PR also adds some new features
1. Switch to Unicode API for Windows, so service will be able to use Unicode paths.
2. Fix Windows service do not recognize command line.
3. Make makeBannerAndConfig to become procedure which returns errors instead of crashing application.


beacon_chain/winservice.nim
```diff
@@ -0,0 +1,186 @@
...
+                                    argCopyrights,
+                                    argNimBanner,
+                                    argServiceName: string,
+                                    argConfigType: untyped,
+                                    argEntryPoint: untyped): untyped =
+
+    proc serviceControlHandler(dwCtrl: DWORD): WINBOOL {.stdcall.} =
+      case dwCtrl
+      of SERVICE_CONTROL_STOP:
+        # We re reporting that we plan stop the service in 10 seconds
```
tersec:
```suggestion
        # We're reporting that we plan to stop the service in 10 seconds
```

beacon_chain/conf.nim
```diff
@@ -1483,3 +1483,50 @@ proc loadKzgTrustedSetup*(trustedSetupPath: string): Result[void, string] =
...
+            config: ConfType, sources: ref SecondarySources
+        ) {.raises: [ConfigurationError].} =
+          if config.configFile.isSome:
+            sources.addConfigFile(Toml, config.configFile.get)
+      )
+    else:
+      ConfType.load(
+        version = version, # but a short version string makes more sense...
+        copyrightBanner = clientId,
+        cmdLine = @environment,
```
tersec: Possible to combine these but with some `cmdLine = if len(environment) == 0: $WHATEVER_DEFAULT else: @environment` construction?

There seems to be no other differences, and 10+ lines of otherwise identical code, or is the default for `cmdLine` tricky enough to handle that it's worthwhile?

beacon_chain/winservice.nim
```diff
@@ -0,0 +1,186 @@
...
+         lpServiceName: LPWSTR,
+         lpHandlerProc: LPHANDLER_FUNCTION
+       ): SERVICE_STATUS_HANDLE {.
+       stdcall, dynlib: "advapi32", importc: "RegisterServiceCtrlHandlerW".}
+
+  proc getCommandLine(dwArgc: DWORD,
+                      lpszArgv: LPWSTR): Result[seq[string], string] =
+    var res: seq[string]
+    let arguments = cast[ptr UncheckedArray[LPWSTR]](lpszArgv)
+    for i in 0 ..< int(dwArgc):
```
tersec, line 80: are `dwArgc`s where this would `Defect` a concern?

beacon_chain/winservice.nim
```diff
@@ -0,0 +1,186 @@
...
+
+      let environment =
+        block:
+          # `valueOr` cannot be used because of we can't use `error`.
+          let res = getCommandLine(dwArgc, lpszArgv)
+          if res.isErr():
+            stderr.write res.error
+            quit QuitFailure
+          res.get()
+
```
tersec: Are the two blank lines here intentional? They don't show up elsewhere in this PR even between functions

beacon_chain/winservice.nim
```diff
@@ -0,0 +1,186 @@
...
+                           lpServiceProc: serviceMainFunction),
+      SERVICE_TABLE_ENTRYW(lpServiceName: nil,
+                           lpServiceProc: nil)
+    ]
+
+    let status =
+      startServiceCtrlDispatcher(LPSERVICE_TABLE_ENTRYW(addr dispatchTable[0]))
+    if status == 0:
+      let errorCode = osLastError()
+      fatal "Failed to start Windows service", error_code = int(errorCode),
```
tersec: Similar question about whether error codes could be out of range of `int`

beacon_chain/conf.nim
cheatfate: The problem here is that `load` is the template declared here https://github.com/status-im/nim-confutils/blob/master/confutils.nim#L1169-L1177. As you can see default value for `cmdLine` is call to private procedure which converts current command line into sequence of strings.

tersec: Ok, reasonable enough.

beacon_chain/winservice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5612/commits/49a8c33df1295a27141e714f3ef73e4584b2c454

beacon_chain/winservice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5612/commits/49a8c33df1295a27141e714f3ef73e4584b2c454

beacon_chain/winservice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5612/commits/49a8c33df1295a27141e714f3ef73e4584b2c454

beacon_chain/winservice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5612/commits/49a8c33df1295a27141e714f3ef73e4584b2c454

kdeme:
Nice!

Would be great if `makeBannerAndConfig` could be pulled out of the winservice code. Then I could use this in Fluffy as is :).

tersec (approved):

beacon_chain/conf_common.nim
```diff
@@ -0,0 +1,53 @@
...
+        if config.configFile.isSome:
+          sources.addConfigFile(Toml, config.configFile.get)
+    )
+  except CatchableError as exc:
+    # We need to log to stderr here, because logging hasn't been configured yet
+    var msg = "Failure while loading the configuration:\p" & exc.msg & "\p"
+    if (exc[] of ConfigurationError) and not(isNil(exc.parent)) and
+       (exc.parent[] of TomlFieldReadingError):
+      let fieldName = ((ref TomlFieldReadingError)(exc.parent)).field
+      if fieldName in ["web3-url", "bootstrap-node",
```
zah:
```suggestion
      if fieldName in ["el", "web3-url", "bootstrap-node",
```

PR #5613, opened 2023-11-21
restrict best LC update collection to canonical blocks

etan-status:
Simplify best `LightClientUpdate` collection by tracking only canonical data instead of tracking the best update across all branches within the sync committee period.

- https://github.com/ethereum/consensus-specs/pull/3553

PR #5623, opened 2023-11-25
rm unused code

tersec:
All found by `XDeclaredButNotUsed`


etan-status (approved):
Thanks for the cleanup. the LC code is indeed obsolete since #5613 ; forgot to remove this piece back then.

PR #5632, opened 2023-11-28
REST: Add implementation of postStateValidators and postStateValidatorBalances.

cheatfate:
This PR implements https://github.com/ethereum/beacon-APIs/pull/367 and tests.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -242,8 +242,8 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
          state: ForkedHashedBeaconState
        ): Result[seq[ValidatorIndex], RestErrorMessage] =
     var
-      keyset: Table[ValidatorPubKey, uint]
-      indexset: Table[ValidatorIndex, uint]
+      keyset: HashSet[ValidatorPubKey]
+      indexset: HashSet[ValidatorIndex]
 
     let validatorsCount = lenu64(getStateField(state, validators))
 
@@ -251,7 +251,7 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
       case item.kind
       of ValidatorQueryKind.Key:
         # Test for uniqueness of value.
-        if keyset.hasKeyOrPut(item.key, 0'u):
+        if keyset.containsOrIncl(item.key):
           return err(RestErrorMessage.init(
             Http400, NonUniqueValidatorIdError, $item.key))
       of ValidatorQueryKind.Index:
@@ -266,20 +266,19 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         if uint64(vindex) < validatorsCount:
           # We're only adding validator indices which are present in
           # validators list at this moment.
-          if indexset.hasKeyOrPut(vindex, 0'u):
+          if indexset.containsOrIncl(vindex):
             return err(RestErrorMessage.init(
               Http400, NonUniqueValidatorIdError,
               Base10.toString(uint64(vindex))))
 
     if len(keyset) > 0:
-      let optIndices = keysToIndices(node.restKeysCache, state,
-                                     keyset.keys().toSeq())
+      let optIndices = keysToIndices(node.restKeysCache, state, keyset.toSeq())
       # Remove all the duplicates.
       for item in optIndices:
         # We ignore missing keys.
         if item.isSome():
-          indexset[item.get()] = 0'u
-    ok(indexset.keys().toSeq())
+          indexset.incl(item.get())
+    ok(indexset.toSeq())
 
   proc getValidators(
          node: BeaconNode,
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -242,15 +242,18 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
          state: ForkedHashedBeaconState
        ): Result[seq[ValidatorIndex], RestErrorMessage] =
     var
-      keyset: HashSet[ValidatorPubKey]
-      indexset: HashSet[ValidatorIndex]
+      keyset: Table[ValidatorPubKey, uint]
+      indexset: Table[ValidatorIndex, uint]
 
     let validatorsCount = lenu64(getStateField(state, validators))
 
     for item in validatorIds:
       case item.kind
       of ValidatorQueryKind.Key:
-        keyset.incl(item.key)
+        # Test for uniqueness of value.
+        if keyset.hasKeyOrPut(item.key, 0'u):
+          return err(RestErrorMessage.init(
+            Http400, NonUniqueValidatorIdError, $item.key))
       of ValidatorQueryKind.Index:
         let vindex = item.index.toValidatorIndex().valueOr:
           case error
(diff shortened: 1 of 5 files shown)


beacon_chain/rpc/rest_constants.nim
```diff
@@ -26,6 +26,8 @@ const
     "Block header/data has not been found"
   EmptyRequestBodyError* =
     "Empty request's body"
+  InvalidRequestBodyError* =
+    "Invalid request's body"
```
tersec:
```suggestion
    "Invalid request body"
```

beacon_chain/rpc/rest_constants.nim
```diff
@@ -26,6 +26,8 @@ const
     "Block header/data has not been found"
   EmptyRequestBodyError* =
     "Empty request's body"
```
tersec:
```suggestion
    "Empty request body"
```

beacon_chain/rpc/rest_constants.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5632/commits/d60caf654dd3b4a0534cb12382884d1dc0b1ffd9

beacon_chain/rpc/rest_constants.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/5632/commits/d60caf654dd3b4a0534cb12382884d1dc0b1ffd9

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3758,6 +3758,45 @@ proc encodeString*(value: set[EventTopic]): Result[string, cstring] =
...
+                 ValidatorFilterKind.ActiveSlashed}
+    exitedSet = {ValidatorFilterKind.ExitedUnslashed,
+                 ValidatorFilterKind.ExitedSlashed}
+    withdrawSet = {ValidatorFilterKind.WithdrawalPossible,
+                   ValidatorFilterKind.WithdrawalDone}
+  var
+    res: seq[string]
+    v = value
+
+  if pendingSet * v == pendingSet:
```
tersec: All four `if` statements here have the same structure:
```nim
  if fooSet * v == fooSet:
    res.add(stringReprOfFoo)
    v.excl(fooSet)
```
but because of the copy/paste structure, potentially error-prone to maintain (there's no automatic error if it doesn't match that template, for example).

tersec: https://github.com/status-im/nimbus-eth2/pull/5632/commits/af09815035d6d4fe81326e663caaa80cc1e7f9b2

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -236,6 +236,129 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+  proc getValidators(
+         node: BeaconNode,
+         bslot: BlockSlotId,
+         validatorsMask: ValidatorFilter,
+         validatorIds: openArray[ValidatorIdent]
+       ): RestApiResponse =
+    node.withStateForBlockSlotId(bslot):
+      let
+        current_epoch = getStateField(state, slot).epoch()
```
tersec: This is defined here as `current_epoch` -- does it have to be current wall epoch, or head state epoch?

Because this isn't really the "current" epoch, it's the epoch of the `bslot`. Maybe that's what the API requires, but it's at least a misleading name.

It seems like it's used to get the validator status, which is another function of the state -- so, it's probably correct that it's based on state and not wall time, but then maybe clarify that via variable name or by some other method. `current_epoch` generally refers to wall time in Nimbus.

tersec: There is a nuance here -- if the validator status would note that by clock/wall time, the validator should be fully exited (or activated, if it's not yet), but it's not yet by the epoch of the `bslot` state, what's the correct answer here?

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -236,6 +236,129 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+          return RestApiResponse.jsonError(error)
+        response =
+          block:
+            var res: seq[RestValidatorBalance]
+            if len(indices) == 0:
+              # Case when `len(indices) == 0 and len(validatorIds) != 0` means
+              # that we can't find validator identifiers in state, so we should
+              # return empty response.
+              if len(validatorIds) == 0:
+                # There is no indices, so we going to return balances of all
```
tersec:
```suggestion
                # There are no indices, so we're going to return balances of all
```

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -236,6 +236,129 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        let vindex = item.index.toValidatorIndex().valueOr:
+          case error
+          of ValidatorIndexError.TooHighValue:
+            return err(RestErrorMessage.init(
+              Http400, TooHighValidatorIndexValueError))
+          of ValidatorIndexError.UnsupportedValue:
+            return err(RestErrorMessage.init(
+              Http500, UnsupportedValidatorIndexValueError))
+        if uint64(vindex) < validatorsCount:
+          # We only adding validator indices which are present in
```
tersec:
```suggestion
          # We're only adding validator indices which are present in
```

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -236,6 +236,129 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+          return RestApiResponse.jsonError(error)
+        response =
+          block:
+            var res: seq[RestValidator]
+            if len(indices) == 0:
+              # Case when `len(indices) == 0 and len(validatorIds) != 0` means
+              # that we can't find validator identifiers in state, so we should
+              # return empty response.
+              if len(validatorIds) == 0:
+                # There is no indices, so we going to filter all the validators.
```
tersec:
```suggestion
                # There are no indices, so we're going to filter all the validators.
```

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3982,3 +4021,82 @@ proc decodeString*(t: typedesc[EventBeaconBlockObject],
...
+        else:
+          Opt.none(seq[ValidatorIdent])
+    filter =
+      block:
+        if statuses.isSome():
+          var res: ValidatorFilter
+          for item in statuses.get():
+            let value = decodeString(ValidatorFilter, item).valueOr:
+              reader.raiseUnexpectedValue($error)
+            if value * res != {}:
```
tersec, line 4080: Should `pending` and `pending_queued` be treated as unique values here for this purpose? It seems like it would trigger this check, because `{pending} * {pending_queued} == {pending_queued} != {}`, but they're not identical.

But this might be the intended semantics of the beacon API.

cheatfate, line 4080: They are not identical but they are not unique, according to specification values should be unique, so there should be no intersection.
https://github.com/ethereum/beacon-APIs/pull/367/files#diff-cc9071f40d45ee8fad4e7513c3351001c1445d206d6726c65a0000421110b1c3R122

tersec: Oops, disregard

beacon_chain/rpc/rest_beacon_api.nim
cheatfate: We return data only according to the `state` which is passed as parameter.

tersec: https://github.com/status-im/nimbus-eth2/pull/5632/commits/47d96ef451a6e9aff5cbe2ea61d8a5b6c7e666cc

beacon_chain/rpc/rest_constants.nim
```diff
@@ -101,6 +103,8 @@ const
     "Invalid block identifier value"
   InvalidValidatorIdValueError* =
     "Invalid validator's identifier value(s)"
```
tersec:
```suggestion
    "Invalid validator identifier value(s)"
```

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -236,6 +236,137 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
     return RestApiResponse.jsonError(Http404, StateNotFoundError)
 
+  proc getIndices(
+         node: BeaconNode,
+         validatorIds: openArray[ValidatorIdent],
+         state: ForkedHashedBeaconState
+       ): Result[seq[ValidatorIndex], RestErrorMessage] =
+    var
+      keyset: Table[ValidatorPubKey, uint]
+      indexset: Table[ValidatorIndex, uint]
```
tersec: The only value it ever seems to assign or store is `0'u`, so I'm not sure how this is an improvement over
```nim
      keyset: HashSet[ValidatorPubKey]
      indexset: HashSet[ValidatorIndex]
```
which earlier PR versions had.

```nim

    if len(keyset) > 0:
      let optIndices = keysToIndices(node.restKeysCache, state,
                                     keyset.keys().toSeq())
```
just ignores all the values entirely.

tersec: https://github.com/status-im/nimbus-eth2/pull/5632/commits/f75a1ed52ee3e1f618904b368fe1815145c9c860

tersec (approved):

PR #5635, opened 2023-12-01
stop checking beacon parent root for fcU caching

tersec:
Discovered on `devnet-12`.

Otherwise, pre-Deneb, it sends two identical fcUs for each proposal around one slot apart from each other, because the caching was comparing a placeholder `ZERO_HASH` `parentBeaconBlockRoot` with the actual `consensusHead.asBlockHash`, and inevitably discovering they differed.

For pre-Deneb cases, one straightforwardly doesn't have to check `parentBeaconBlockRoot` (CL head), and even for Deneb, that differs if and only if `headBlockHash` / `latestHead` differs, so it's either harmful (pre-Deneb) or redundant (Deneb or later) to check both.

https://github.com/paradigmxyz/reth/issues/5623 has some more discussion.

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -17,7 +17,6 @@ import
   eth/async_utils, stew/[assign2, byteutils, objects, results, shims/hashes, endians2],
   # Local modules:
   ../spec/[deposit_snapshots, eth2_merkleization, forks, helpers],
-  ../spec/datatypes/[base, phase0, bellatrix, deneb],
   ../networking/network_metadata,
   ../consensus_object_pools/block_pools_types,
   ".."/[beacon_chain_db, beacon_node_status, beacon_clock, future_combinators],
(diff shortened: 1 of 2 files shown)


beacon_chain/el/el_manager.nim
```diff
@@ -17,7 +17,6 @@ import
   eth/async_utils, stew/[assign2, byteutils, objects, results, shims/hashes, endians2],
   # Local modules:
   ../spec/[deposit_snapshots, eth2_merkleization, forks, helpers],
-  ../spec/datatypes/[base, phase0, bellatrix, deneb],
```
tersec, line 20: Redundant with the `forks` import.

beacon_chain/sync/request_manager.nim
```diff
@@ -342,7 +342,7 @@ proc requestManagerBlobLoop(rman: RequestManager) {.async.} =
              failed = (len(workers) - succeed),
              sync_speed = speed(start, finish)
 
-      except CancelledError as exc:
```
tersec, line 345: `XDeclaredButNotUsed`

PR #5644, opened 2023-12-04
rm spec-deprecated `produceBlindedBlock` REST API support

tersec:
Would need to verify if/that the Nimbus VC can use `produceBlockV3` for this purpose yet.

Currently glitchy in CI because it effectively re-introduces the issue https://github.com/status-im/nimbus-eth2/pull/5628 solved.


tersec:
teku_produceblockv3_2023_12
Ideally, by the time this is removed, the Teku VC would be able to use `produceBlockV3` in place of `produceBlindedBlock` without this flag.

arnetheduck:
isn't this a post-deneb change?

tersec:
> isn't this a post-deneb change?

Sure, this is a draft/tracking PR.

There are already two VCs known which aren't compatible with this, one of which is the Nimbus VC, so I'm certainly not looking for near-term merging.

But I'm absolutely looking for this to actually happen when feasible.

tersec:
https://github.com/status-im/nimbus-eth2/pull/5842 resolves the Nimbus VC side of this. In theory, post Dencun fork, all the VCs will have to support this, but would need to be verified first.

tersec:
Apparently, Prysm's validator client is still gRPC-based, so not in scope here.

tbenr:
Update, now teku's BN is fully compatible with blockV3. Our VC is still using old apis by default but we have a flag to enable V3 usage

nflaig:
Lodestar fully supports produceBlockV3. 

But regarding deprecation vs. removal, the point of deprecating APIs instead of removing them right away is to give everyone time to switch to the latest implementation. There was [some discussion](https://discord.com/channels/595666850260713488/710831784991916072/1208895651648249906) on this already on when to remove an API from the spec and the rough consensus was to have at least one hard fork between deprecation and removal. In any case, we should first remove it from the spec as otherwise a client removing it before that would no longer be spec compliant.

tersec:
Alright, will refrain from removing this (and other similar endpoints) for some time.

PR #5654, opened 2023-12-06
use `PayloadAttributesV3` in `nimbus_light_client` for Deneb

etan-status:
From Deneb onward, `forkchoiceUpdated` requires `PayloadAttributesV3`. In `nimbus_light_client` we still used `PayloadAttributesV2`.

Also clean up two other locations that were already correctly using `PayloadAttributesV3`, to reduce code duplication.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -642,21 +642,11 @@ proc storeBlock(
             finalizedBlockHash = newHead.get.finalizedExecutionPayloadHash,
             payloadAttributes = none attributes)
 
-      case self.consensusManager.dag.cfg.consensusForkAtEpoch(
-          newHead.get.blck.bid.slot.epoch)
-      of ConsensusFork.Deneb:
-        callForkchoiceUpdated(PayloadAttributesV3)
-      of ConsensusFork.Capella:
-        # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.3/src/engine/shanghai.md#specification-1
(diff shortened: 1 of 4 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -642,21 +642,11 @@ proc storeBlock(
             finalizedBlockHash = newHead.get.finalizedExecutionPayloadHash,
             payloadAttributes = none attributes)
 
-      case self.consensusManager.dag.cfg.consensusForkAtEpoch(
-          newHead.get.blck.bid.slot.epoch)
-      of ConsensusFork.Deneb:
-        callForkchoiceUpdated(PayloadAttributesV3)
-      of ConsensusFork.Capella:
-        # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.3/src/engine/shanghai.md#specification-1
```
tersec, line 650: It'd be useful to keep some evidence of these in the resulting code. Essentially, they clarify why have the complexity in the all-but-a-couple-of-cases of doing different types of fcU at all, including the different payload attributes. For example, the light client never produces blocks, so one might imagine that it doesn't have to care, and could always use the same kind of fcU, and same kind of payload attributes.

This spec reference explains why this is not a valid simplification.

tersec (approved):

beacon_chain/gossip_processing/block_processor.nim
etan-status: https://github.com/status-im/nimbus-eth2/pull/5658

PR #5666, opened 2023-12-13
Remove downloaded mentioned twice from keep-updated.md

kdeme:


tersec (approved):

tersec:
The GitHub Actions heuristics notice that this won't change their tested targets and simply won't run -- auto-merge won't work here. It needs to be manually merged.

PR #5673, opened 2023-12-16
rm aspirational reference to EL-unimplemented pull request from docs

tersec:
Followup to https://github.com/status-im/nimbus-eth2/pull/5643#discussion_r1414584638

### DIFF
--- a/docs/the_nimbus_book/src/eth1.md
+++ b/docs/the_nimbus_book/src/eth1.md
@@ -94,10 +94,6 @@ You will need to pass the path to the token file to Nimbus together with the web
       --jwt-secret=/tmp/jwtsecret
     ```
 
-!!! info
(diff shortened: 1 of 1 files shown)


docs/the_nimbus_book/src/eth1.md
```diff
@@ -94,10 +94,6 @@ You will need to pass the path to the token file to Nimbus together with the web
       --jwt-secret=/tmp/jwtsecret
     ```
 
-!!! info
-    When the `--jwt-secret` option is not specified and the execution client is running on the same machine under default setting, Nimbus may be able to connect successfully to it by using the default secret value `0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3`.
```
zah, line 98: This text still describes the actual behavior of the client though. IMO, If we agree that is not realistic for this feature to ever get used, we should remove it from the client and a failure to specify a `--jwt-secret` should become more apparent to the user.

tersec: Well, it's possible an EL will implement it, at which point it's not 100% futile for someone reading the documentation to believe that perhaps it might work.

Is there any remaining use case for ever not using a JWT secret?

PR #5674, opened 2023-12-19
Use the new strict Json flavors mechanism for RestJson

zah:
This requires all object types to be explicitly white-listed for default serialization. The PR makes the minimal changes, although a number of similar mechanisms in `eth2_rest_serialization` can now be removed.

### DIFF
--- a/tests/test_validator_client.nim
+++ b/tests/test_validator_client.nim
@@ -316,6 +316,11 @@ type
   SyncCommitteeBitsObject = object
     data: SyncCommitteeAggregationBits
 
+RestJson.useDefaultSerializationFor(
+  AttestationBitsObject,
+  SyncCommitteeBitsObject
+)
+
 const
   AttestationDataVectors = [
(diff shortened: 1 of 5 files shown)


arnetheduck:
do this also fix bleeding of `int` and other basic types? ie int-as-json-string vs int-as-json-number..

zah:
No, only the object handling has been pushed so far.

tests/test_validator_client.nim
```diff
@@ -316,6 +316,11 @@ type
   SyncCommitteeBitsObject = object
     data: SyncCommitteeAggregationBits
 
+RestJson.useDefaultSerializationFor(
```
arnetheduck, line 319: is it possible to add a pragma to the object instead? ie `type MyObject {.serialization: [RestJson, OtherFormat].}`

zah: Should be possible, yes.

PR #5680, opened 2023-12-23
load Goerli metadata from `goerli` repo

etan-status:
The `eth2-networks` repo often receives metadata updates with a delay. Switch to `goerli` repo to obtain the latest config (Dencun scheduling) when it is updated. This is in line with how Sepolia / Holesky work.

- https://github.com/eth-clients/goerli/pull/178

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2019-2023 Status Research & Development GmbH
+# Copyright (c) 2019-2024 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at http://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at http://www.apache.org/licenses/LICENSE-2.0).
--- a/.gitmodules
+++ b/.gitmodules
@@ -220,6 +220,11 @@
 	url = https://github.com/arnetheduck/nim-results.git
 	ignore = untracked
 	branch = master
+[submodule "vendor/goerli"]
+	path = vendor/goerli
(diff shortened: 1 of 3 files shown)


etan-status:
- https://github.com/eth-clients/goerli/pull/179

tersec:
Assuming the changes are currently synced between `eth2-networks` and `goerli` repos, this seems reasonable. If nothing else, within a few months, probably `nimbus-eth2` will in accordance with https://blog.ethereum.org/2023/11/30/goerli-lts-update drop support for Goerli/Prater altogether, regardless of from which repo.

One somewhat nice aspect now is that it only fetches `eth2-networks`, not `eth2-networks` and `goerli`, on recursive clone/`make update`/`git submodules update`/etc. So functionally, this adds to download time and clone storage space requirements for the sake of consistency with Holesky.

If https://github.com/eth-clients/eth2-networks/pull/92 is merged, that would restore the status quo ante, by making a (shallow) submodule clone of `eth2-networks` less bandwidth and disk-intensive by almost exactly the same amount as adding `goerli` increases it, for no net changes.

Also, it is more consistent with Holesky, but it's less consistent with mainnet. If `mainnet` had also its own updated/maintained repo, that then using both that and the `goerli` repo as here would be ideal. But https://github.com/eth-clients/eth2-mainnet exists and "has been archived by the owner on May 26, 2021. It is now read-only", so that would have to be recreated.

Certainly, not having two out-of-sync copies of the Goerli/Prater network configuration is better, regardless.

etan-status:
Yeah, regardless of deprecation, we just need to pick the repo that has the correct DENCUN_FORK_EPOCH set. Which, right now, is only planned for `goerli` (open PRs)

PR #5685, opened 2024-01-02
rm unused RPC signatures replaced by Plain versions

tersec:
Various updates have switched from the non-`Plain` versions to `Plain` versions to  better recover and report errors, rather than having them swallowed by the REST JSON deserializer and other components.

For example:
- https://github.com/status-im/nimbus-eth2/pull/3318 (`getStateForkPlain`)
- https://github.com/status-im/nimbus-eth2/pull/4673 (`getBlockRootPlain`, `getForkSchedulePlain`, and `getStateValidatorsPlain`, `getSyncCommitteeDutiesPlain`, and others)

The non-`Plain` versions are attractive nuisances at this point, because their further usage means that they'll swallow details of failure modes in ways which hinder reacting to them and informing users about them.

### DIFF
--- a/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim
+++ b/beacon_chain/spec/eth2_apis/rest_beacon_calls.nim
@@ -1,4 +1,4 @@
-# Copyright (c) 2018-2023 Status Research & Development GmbH
+# Copyright (c) 2018-2024 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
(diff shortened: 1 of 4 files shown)


arnetheduck:
Hm, in general, the plain versions feel like there's some gap in the generality of the non-plain version - basically, we should be able to generate a "starting point" from the yaml specification and everything in between, including error handling, should be pluggable in such a way that the dumb automatic part of the translation ends up being correct and useful.

tersec:
I agree, but these `Foo`/`FooPlain` pairs only differ in their return values, so it'd still be better to document/record those keep whole duplicate signatures around. Or, the correct way of decoding things can be found by looking for users of each RPC signature. Either way, it's more correct/reliable than relying on unused code which could contain random garbage so long as it happens to compile, and would not be noticed, for such a purpose.

Ideally, I'd look for the ability of Presto, etc to be able to return the decoded object, if it can be decoded, and information sufficient to decode errors, if it can't, or to be able to signal to Presto that if it gets a response it can't decode, that's an error worth logging at `INFO` or above log levels (though, probably inappropriate to have such a library interface).

Regardless, this PR simply ratifies the existing reality, in a way which removes unused code.

PR #5719, opened 2024-01-09
Avoid global in p2p macro (fixes #4578)

### LINKED ISSUE #4578: Remove use of global `gProtocols` global protocol registry from `nimbus_beacon_node`
Because https://github.com/status-im/nimbus-eth2/issues/4571 will start hitting major stable and LTS distributions, without the underlying gcc 12.x bug having been fixed or Nim (any branch) having implemented some workaround to avoid generating triggering C code, it now lands on `nimbus-eth2` to avoid this issue.

https://github.com/status-im/nimbus-eth2/blob/94837caa2a5fba5ab2777dbd9ff7f0a422cbd4b2/beacon_chain/networking/eth2_network.nim#L1015-L1020
is the code triggering it in this repo. That particular global `gProtocols` comes from https://github.com/status-im/nim-eth/blob/master/eth/p2p/p2p_backends_helpers.nim but it's less clear whether its https://github.com/status-im/nim-eth/blob/master/eth/p2p/rlpx.nim usage might trigger the same issue. As `nimbus-eth2` does not use RLPx, this issue only concerns itself with the `eth2_network` usage.

The key most readily avoidable aspect which triggers the gcc bug is that it's a mutable, global variable.

arnetheduck:
* copy p2p macro to this repo and start de-crufting it
* make protocol registration dynamic, removing light client hacks et al
* split out light client protocol into its own file


beacon_chain/networking/eth2_network.nim
```diff
@@ -2091,16 +1961,17 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
...
               msgNameLit,
               protocolMounterName,
-              codecNameLit,
-              isRequiredLit,
-              isLightClientRequestLit))
+              codecNameLit))
 
   result.implementProtocolInit = proc (p: P2PProtocol): NimNode =
-    return newCall(initProtocol, newLit(p.name), p.peerInit, p.netInit)
+    var id {.global.}: int
```
zah, line 1967: This approach is not building on good foundations. Incrementing a global variable at compile-time like this is fundamentally at odds with incremental compilation (because a recompile that doesn't revisit the whole program will mess up the counting and create duplicate indices). @jangko's approach in the nim-eth repo was better because it was relying on the [macrocache](https://nim-lang.org/docs/macrocache.html) module from the std lib, which has well defined semantics for the purposes of incremental compilation.

You should at least leave my remarks as a comment in the code.

zah: See this for a more in-depth discussion of the problem:
https://github.com/nim-lang/RFCs/issues/46#issuecomment-395686265

arnetheduck: https://github.com/status-im/nimbus-eth2/pull/5745

PR #5729, opened 2024-01-12
Switch to llvm-mingw for faster Windows CI

jangko:


Makefile
```diff
@@ -791,7 +791,7 @@ test_libnimbus_lc: libnimbus_lc.a
 			if (( $${WITH_UBSAN:-0} )); then \
 				echo "MINGW cannot find -lubsan." && exit 1; \
 			fi; \
-			gcc -D__DIR__="\"beacon_chain/libnimbus_lc\"" --std=c17 -Wall -Wextra -pedantic -Werror -pedantic-errors -flto -Wl,--stack,0x0000000000800000 -o build/test_libnimbus_lc -D_CRT_SECURE_NO_WARNINGS beacon_chain/libnimbus_lc/test_libnimbus_lc.c build/libnimbus_lc.a "$${EXTRA_FLAGS[@]}"; \
+			gcc -D__DIR__="\"beacon_chain/libnimbus_lc\"" --std=c17 -Wall -Wextra -pedantic -Werror -pedantic-errors -flto -Wno-nullability-extension -Wl,--stack,0x0000000000800000 -o build/test_libnimbus_lc -D_CRT_SECURE_NO_WARNINGS beacon_chain/libnimbus_lc/test_libnimbus_lc.c build/libnimbus_lc.a "$${EXTRA_FLAGS[@]}"; \
```
arnetheduck, line 793: shouldn't this ubsan run on clang too?

arnetheduck:
switching tests means that we must also switch release binary production (so  that the unit test runs are representative of what ends up getting compiled on release / nightly

jangko:
Release binary is build using cross compilation tools from linux. Don't know if there exists such tool from llvm.

Makefile
jangko: fixed

PR #5736, opened 2024-01-14
Add noSideEffect cast to base64urlEncode in engine_authentication

jangko:


tersec:
Shouldn't be necessary - https://github.com/nim-lang/Nim/pull/22986 was backported to Nim v1.6.18, and that's what `nimbus-eth2` uses.

jangko:
Okay, thanks

kdeme:
@jangko Yeah, this should be fine now. This is the error why I noticed we were a Nim version behind in nimbus-eth1 and bumped it here: https://github.com/status-im/nimbus-eth1/pull/1967

Issue #5738, opened 2024-01-14
`/eth/v3/validator/blocks/{slot}` returning execution value in hex


mcdee:
**Describe the bug**

When requestion `/eth/v3/validator/blocks/{slot}` the header `Eth-Execution-Payload-Value` is returned as hex rather than decimal (at least that's what it looks like; sample value is `a169893a166614`).

**To Reproduce**

Call `/eth/v3/validator/blocks/{slot}` for a suitable slot and check the `Eth-Execution-Payload-Value` header.

**Additional context**
Tested on goerli.

```
nimbus_beacon_node --version
Nimbus beacon node v24.1.1-0e63f8-stateofus
Copyright (c) 2019-2024 Status Research & Development GmbH

eth2 specification v1.4.0-beta.5

Nim Compiler Version 1.6.16 [Linux: amd64]
```

tersec:
I'm not sure I understand the issue here. Are you saying it should return a decimal and does return hex, or vice versa? Because https://ethereum.github.io/beacon-APIs/#/Validator/produceBlockV3 uses as its example:
```
{
  "version": "phase0",
  "execution_payload_blinded": false,
  "execution_payload_value": "12345",
  ...
```

which could be either, and it's documented in the table below as
> Eth-Execution-Payload-Value	
>
> Required in response so client can determine relative value of execution payloads.
> string
> Example: 1

which is also ambiguous, though typically hex is represented with `0x`.

https://github.com/ethereum/beacon-APIs/blob/cfbe6677f6e9e936dfca8708467c7e0382c07b19/apis/validator/block.v3.yaml#L87-L90 states that it should be
```
        Eth-Execution-Payload-Value:
          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Execution-Payload-Value'
        Eth-Consensus-Block-Value:
          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Block-Value'
```

Which are defined by https://github.com/ethereum/beacon-APIs/blob/cfbe6677f6e9e936dfca8708467c7e0382c07b19/beacon-node-oapi.yaml#L422-L431 to be
```
    Eth-Execution-Payload-Value:
      description: Required in response so client can determine relative value of execution payloads.
      required: true
      schema:
        $ref: './types/primitive.yaml#/Wei'
    Eth-Consensus-Block-Value:
      description: Required in response so client can determine relative value of consensus blocks.
      required: true
      schema:
[...]

PR #5740, opened 2024-01-14
report `produceBlockV3` consensus block and execution payload values in decimal

tersec:
https://github.com/status-im/nimbus-eth2/issues/5738


tersec:
terminally broken in CI because https://github.com/status-im/nimbus-eth2/commit/05fb7ffff04ab1013f3156a2747fd4e9199eef58 oops, will recreate as new PR

PR #5747, opened 2024-01-15
near-clone of 7378d55108311e05a437f8bdb011b6ed3207834a

tersec:
check of
```
[2024-01-15T07:55:32.850Z] removeValidatorFiles() multiple keystore types ...... (0.00s)
[2024-01-15T07:55:37.791Z] createValidatorFiles() ........ (5.09s)
[2024-01-15T07:55:37.791Z] saveKeystore() ........ (0.01s)
[2024-01-15T07:55:38.696Z] Wrote test_keymanager_api/bootstrap_node.enr
[2024-01-15T07:55:45.505Z] Serialization/deserialization [Beacon Node] [Preset: mainnet] . (0.00s)
[2024-01-15T07:55:45.505Z] ListKeys requests [Beacon Node] [Preset: mainnet] .... (0.01s)
[2024-01-15T07:55:50.005Z] ImportKeystores requests [Beacon Node] [Preset: mainnet] Traceback (most recent call last, using override)
[2024-01-15T07:55:50.005Z] vendor/nim-libp2p/libp2p/protocols/rendezvous.nim(1026) main
[2024-01-15T07:55:50.005Z] vendor/nim-libp2p/libp2p/protocols/rendezvous.nim(1016) NimMain
[2024-01-15T07:55:50.005Z] SIGSEGV: Illegal storage access. (Attempt to read from nil?)
[2024-01-15T07:55:50.005Z] bash: line 1: 23879 Segmentation fault: 11  NIMBUS_TEST_KEYMANAGER_BASE_PORT=$(( 9960 + EXECUTOR_NUMBER * 6 + 0 )) NIMBUS_TEST_SIGNING_NODE_BASE_PORT=$(( 9960 + EXECUTOR_NUMBER * 6 + 4 )) build/${TEST_BINARY} ${PARAMS}
[2024-01-15T07:55:50.005Z] 
[2024-01-15T07:55:50.005Z] all_tests --xml:build/all_tests.xml --console failed; Last 5000 lines from the log:
[2024-01-15T07:55:50.005Z] {"lvl":"DBG","ts":"2024-01-15 09:54:52.305+02:00","msg":"Stored state","topics":"chaindag","tid":59837754,"file":"blockchain_dag.nim:877","putStateDur":{"value":2417458},"stateRoot":"b19e5a9d","stateSlot":160,"blck":"1680500d:159"}
[...]

PR #5748, opened 2024-01-15
Revert "rm kzg test fixture code duplication and add test vector coverage checks"

tersec:
Reverts status-im/nimbus-eth2#5742

should be harmless, but it is in `all_tests` and the timing lines up, worth a try

plausible because to avoid the https://github.com/status-im/nimbus-eth2/commit/05fb7ffff04ab1013f3156a2747fd4e9199eef58 triggered macOS (both x86 and aarch64) CI finalization script issue, https://github.com/status-im/nimbus-eth2/pull/5740 was deliberately based on https://github.com/status-im/nimbus-eth2/commit/b98f46c04dfe9295393117bb1f40dfba4ac4f97f the last commit before https://github.com/status-im/nimbus-eth2/commit/05fb7ffff04ab1013f3156a2747fd4e9199eef58 so there was a kind of CI test collision possible

### DIFF
--- a/tests/consensus_spec/test_fixture_kzg.nim
+++ b/tests/consensus_spec/test_fixture_kzg.nim
@@ -15,14 +15,27 @@ import
   ../testutil,
   ./fixtures_utils, ./os_ops
 
-from std/sequtils import anyIt, mapIt, toSeq
+from std/sequtils import anyIt, mapIt
 from std/strutils import rsplit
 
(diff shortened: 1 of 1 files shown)


tersec:
Probably (modulo auto-merging-to-current-`unstable` possibilities) not useful/necessary because https://github.com/status-im/nimbus-eth2/pull/5747 (based on parent commit) triggers this

PR #5751, opened 2024-01-15
only show upcoming fork info if one is scheduled

etan-status:
This ensures that information about the next scheduled fork is only displayed if one is actually scheduled. Current fork name is no longer shown.


etan-status:
It's a bit hacky, but maybe aesthetically more pleasing. @arnetheduck

PR #5757, opened 2024-01-16
workaround random `SIGSEGV` on macOS aarch64 CI

etan-status:
Separate a `let` block into multiple `let` statements to reduce probability of hitting random `SIGSEGV` during flaky CI tests.

whatever... 🤯

### DIFF
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -1227,9 +1227,12 @@ proc proposeBlockAux(
     # This change, maybe combined with some macOS specific compiler specifics,
     # could this trigger the `SIGSEGV`? Maybe the extra state adds just enough
     # complexity to the function to disable certain problematic optimizations?
-    # Furthermore, the combination of `(await xyz).valueOr: return` is not very
-    # commonly used together with other `await` in the same `let` block, which
-    # could explain why this is not a more common trigger across Nimbus.
+    # The change in size of the environment changes a number of things such as
+    # alignment and which parts of an environment contain pointers and so on,
+    # which in turn may have surprising behavioural effects, ie most likely this
+    # extra state masks some underlying issue. Furthermore, the combination of
+    # `(await xyz).valueOr: return` is not very commonly used with other `await`
+    # in the same `let` block, which could explain this not being more common.
     #
     # Note that when compiling for Wasm, there are similar bugs with `results`
     # when inlining unwraps, e.g., in `eth2_rest_serialization.nim`.
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -1178,6 +1178,85 @@ proc proposeBlockAux(
             blobsBundle.proofs, blobsBundle.blobs))
         else:
           Opt.none(seq[BlobSidecar])
+
+    # BIG BUG SOURCE: The `let` below cannot be combined with the others above!
+    # If combined, there are sometimes `SIGSEGV` during `test_keymanager_api`.
+    # This has only been observed on macOS (aarch64) in Jenkins, not on GitHub.
+    #
+    # - macOS 14.2.1 (23C71)
+    # - Xcode 15.1 (15C65)
+    # - Nim v1.6.18 (a749a8b742bd0a4272c26a65517275db4720e58a)
+    #
+    # Issue has started occuring around 12 Jan 2024, in a CI run for PR #5731.
+    # The PR did not change anything related to this, suggesting an environment
+    # or hardware change. The issue is flaky; could have been introduced earlier
+    # before surfacing in the aforementioned mentioned PR. About 30% to hit bug.
+    #
+    # [2024-01-12T11:54:21.011Z] Wrote test_keymanager_api/bootstrap_node.enr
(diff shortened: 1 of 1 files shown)


etan-status:
workaround runs here in endless loop until fail: https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/aarch64/job/dev%252Fetan%252Fzz-perpetual/

Seems to be more stable than without it (~10 successful runs so far, none failed)

tersec (approved):

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1178,6 +1178,85 @@ proc proposeBlockAux(
...
+    # +      ok r
+    #      except SerializationError as exc:
+    #        debug "Failed to deserialize REST JSON data",
+    #              err = exc.formatMsg("<data>"),
+    #
+    # At this time we can only speculate about the trigger of these issues.
+    # Until a shared pattern can be identified, it is better to apply
+    # workarounds that at least avoid the known to be reachable triggers.
+    # The solution is hacky and far from desirable; it is what it is.
+    let
```
arnetheduck, line 1262: have you compared the generated C code?

arnetheduck:
> Seems to be more stable than without it (~10 successful runs so far, none failed)

and this is not an artefact of the time of the day (load on server) or something?

etan-status:
There are 50 successful runs in a row with the extra `let` now. While before, it would SIGSEGV every ~3 times.

You can verify on `dev/etan/zz-perpetual` branch, push to it, then go to https://ci.status.im/job/nimbus-eth2/job/platforms/job/macos/job/aarch64/job/dev%252Fetan%252Fzz-perpetual/ and click on left side "Build with Parameters". It will automatically repeat the keymanager test until it fails.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1178,6 +1178,85 @@ proc proposeBlockAux(
...
+    # This has only been observed on macOS (aarch64) in Jenkins, not on GitHub.
+    #
+    # - macOS 14.2.1 (23C71)
+    # - Xcode 15.1 (15C65)
+    # - Nim v1.6.18 (a749a8b742bd0a4272c26a65517275db4720e58a)
+    #
+    # Issue has started occuring around 12 Jan 2024, in a CI run for PR #5731.
+    # The PR did not change anything related to this, suggesting an environment
+    # or hardware change. The issue is flaky; could have been introduced earlier
+    # before surfacing in the aforementioned mentioned PR. About 30% to hit bug.
```
etan-status:
```suggestion
    # before surfacing in the aforementioned PR. About 30% to hit bug.
```

beacon_chain/validators/beacon_validators.nim
etan-status: the extra `let` in this PR introduces an additional state to the closure iterator code. it is still roughly 60 states for this proc, and no I have no clue why it works.

For the other one I refer to here, no, haven't checked, but tried randomly reducing complexity by splitting among statements and using a temporary variable there made it work for Wasm. But as it's only on experimental project, have not followed up on it. But it also points to inline `results` unwrapping being a possible trigger for random bugs.

arnetheduck: > the extra let in this PR introduces an additional state t

this is an important thing to add to the doc comment - the change in size of the environment changes a number of things such as alignment and which parts of an environment contain pointers and so on, which in turn may have surprising behavioural effects, ie most likely this extra state masks some underlying issue.

etan-status: line 1225 already includes that remark

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1178,6 +1178,85 @@ proc proposeBlockAux(
...
+    # [2024-01-12T11:54:34.870Z] SIGSEGV: Illegal storage access. (Attempt to read from nil?)
+    #
+    # The generated `nimcache` differs slightly if the `let` are separated from
+    # a single block; separation introduces an additional state in closure iter.
+    # This change, maybe combined with some macOS specific compiler specifics,
+    # could this trigger the `SIGSEGV`? Maybe the extra state adds just enough
+    # complexity to the function to disable certain problematic optimizations?
+    # Furthermore, the combination of `(await xyz).valueOr: return` is not very
+    # commonly used together with other `await` in the same `let` block, which
+    # could explain why this is not a more common trigger across Nimbus.
```
etan-status:
```suggestion
    # The generated `nimcache` differs slightly if the `let` are separated from
    # a single block; separation introduces an additional state in closure iter.
    # This change, maybe combined with some macOS specific compiler specifics,
    # could this trigger the `SIGSEGV`? Maybe the extra state adds just enough
    # complexity to the function to disable certain problematic optimizations?
    # The change in size of the environment changes a number of things such as
    # alignment and which parts of an environment contain pointers and so on,
    # which in turn may have surprising behavioural effects, ie most likely this
    # extra state masks some underlying issue. Furthermore, the combination of
    # `(await xyz).valueOr: return` is not very commonly used with other `await`
    # in the same `let` block, which could explain this not being more common.
```

etan-status:
I have removed the extra `let` once more on the perpetual testing branch, and it gave SIGSEGV after just two attempts. After more than 60 attempts in a row were successful with the extra `let`.

image

So, yes, this is most likely not time of day dependent.

PR #5761, opened 2024-01-16
add vanity mascot for upcoming fork to status bar

etan-status:
To simplify supporting "am I ready for the fork" requests, add a simple marker to the default status bar that indicates readiness by displaying the fork's corresponding mascot. This is the same one that is also displayed during the actual fork transition, so does not introduce new dependencies. It also only shows in default status bar, not in logs.


etan-status:
image

PR #5763, opened 2024-01-16
fix LC header production for Deneb

etan-status:
The `blob_gas_used` field was not properly populated when constructing Deneb light client data. This is due to #5026 not applying the change to the entire codebase when the new field got introduced, and due to #5350 not catching that oversight in other modules. Also reviewed codebase and discovered that `shortLog` for Deneb execution payloads has same bug.


etan-status:
Thanks for the detailed report: https://notes.ethereum.org/ERz33WNETxeO2Bt1CqST2Q?view

PR #5764, opened 2024-01-16
log `const_preset` on beacon node startup

etan-status:
To understand what binary is being used (regular / minimal / gnosis), extend launch logging.


tersec:
No objection as such, but in general, there seems to be a ~1k/line limit on some Chronicles logging, and `Launching beacon node` already exceeds this.

PR #5765, opened 2024-01-16
simplify eth2_network error handling

arnetheduck:
This PR gets rid of a bunch of redundant exception handling through async raises guarantees.

More can be removed once libp2p gets properly annotated.

Includes nim-eth bump for more accurate timeout handling


beacon_chain/networking/eth2_network.nim
```diff
@@ -540,12 +550,16 @@ proc disconnect*(peer: Peer, reason: DisconnectionReason,
...
       await peer.network.switch.disconnect(peer.peerId)
-  except CatchableError:
-    # We do not care about exceptions in disconnection procedure.
-    trace "Exception while disconnecting peer", peer = peer.peerId,
-                                                reason = reason
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    # switch.disconnect shouldn't raise
+    warn "Unexpected error while disconnecting peer",
```
zah, line 557: Don't we use warning for more actionable information - we are telling the users that there is something wrong with their setup or that we have observed something wrong happening with the network which may explain poor performance, etc. Here, we seem to be telling them about a potential bug in our implementation or alternatively about a very benign networking issue that should not affect their node in any way.

Previously, this was a trace-level message, perhaps `DEBUG` is justified, but warning seems too much.

beacon_chain/networking/eth2_network.nim
```diff
@@ -340,14 +339,22 @@ func shortProtocolId(protocolId: string): string =
...
+  try:
+    ok await dial(node.switch, peer.peerId, protocolId)
+  except LPError as exc:
+    debug "Dialling failed", exc = exc.msg
+    neterr BrokenConnection
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    # TODO remove once libp2p supports `raises`
+    warn "Unknown error when opening stream", exc = exc.msg
```
zah, line 356: Ditto, using `warn` to report implementation issues.

beacon_chain/networking/eth2_network.nim
```diff
@@ -641,22 +655,33 @@ proc sendErrorResponse(peer: Peer,
...
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    debug "Error while writing notification", peer, protocolId, exc = exc.msg
   finally:
-    await stream.close()
+    try:
+      await noCancel stream.close()
+    except CatchableError as exc:
+      warn "Unexpected error while closing notification stream",
```
zah, line 683: Ditto

beacon_chain/networking/eth2_network.nim
```diff
@@ -820,9 +858,15 @@ proc readVarint2(conn: Connection): Future[NetRes[uint64]] {.async.} =
     neterr UnexpectedEOF
   except InvalidVarintError:
     neterr InvalidSizePrefix
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    warn "Unexpected error", exc = exc.msg
```
zah, line 864: Ditto

beacon_chain/networking/eth2_network.nim
```diff
@@ -844,46 +888,49 @@ proc readChunkPayload*(conn: Connection, peer: Peer,
...
-    static: assert ResponseCode.low.ord == 0
-    if responseCodeByte > ResponseCode.high.byte:
-      return neterr InvalidResponseCode
+    await conn.readExactly(addr responseCodeByte, 1)
+  except LPStreamEOFError, LPStreamIncompleteError:
+    return neterr PotentiallyExpectedEOF
+  except CancelledError as exc:
+    raise exc
+  except CatchableError as exc:
+    warn "Unexpected error", exc = exc.msg
```
zah, line 909: Ditto

beacon_chain/networking/eth2_network.nim
```diff
@@ -897,7 +944,7 @@ proc readResponse(conn: Connection, peer: Peer,
       let nextFut = conn.readResponseChunk(peer, E)
       if not await nextFut.withTimeout(timeout):
         return neterr(ReadResponseTimeout)
-      let nextRes = nextFut.read()
+      let nextRes = await nextFut
```
zah, line 947: Why is `await` deemed better than `read()` here? It will introduce an interruption point in the underlying closure iterator which will affect the generated code significantly.

beacon_chain/networking/eth2_network.nim
```diff
@@ -914,15 +961,19 @@ proc readResponse(conn: Connection, peer: Peer,
     let nextFut = conn.readResponseChunk(peer, MsgType)
     if not await nextFut.withTimeout(timeout):
       return neterr(ReadResponseTimeout)
-    return nextFut.read()
+    return await nextFut # Guaranteed to complete without waiting
```
zah, line 964: Ditto (`read` vs `await`)

beacon_chain/networking/eth2_network.nim
```diff
@@ -942,13 +993,18 @@ proc makeEth2Request(peer: Peer, protocolId: string, requestBytes: seq[byte],
...
     raise exc
+  except CatchableError:
+    peer.updateScore(PeerScorePoorRequest)
+    neterr BrokenConnection
   finally:
-    await stream.closeWithEOF()
+    try:
+      await noCancel stream.closeWithEOF()
+    except CatchableError as exc:
+      warn "Unexpected error while closing stream",
```
zah, line 1006: Ditto (inappropriate `warn`)

beacon_chain/networking/eth2_network.nim
```diff
@@ -2022,24 +2082,21 @@ proc peerPingerHeartbeat(node: Eth2Node) {.async.} =
 
       if peer.failedMetadataRequests > MetadataRequestMaxFailures:
         debug "no metadata from peer, kicking it", peer
-        asyncSpawn peer.disconnect(PeerScoreLow)
+        await peer.disconnect(PeerScoreLow)
```
zah, line 2085: Does `disconnect` has an internal time-out mechanism? Could this call stall forever?

beacon_chain/networking/eth2_protocol_dsl.nim
```diff
@@ -652,12 +652,17 @@ proc useStandardBody*(sendProc: SendProc,
   sendProc.setBody quote do:
     mixin init, WriterType, beginRecord, endRecord, getOutput
 
-    var `outputStream` = memoryOutput()
-    `preSerialization`
-    `serialization`
-    `postSerialization`
-    `tracing`
-    let `msgBytes` = getOutput(`outputStream`)
+    let `msgBytes` =
```
zah, line 655: This improvement can be applied to the other copies of the `p2pProtocol` DSL.

beacon_chain/sync/request_manager.nim
```diff
@@ -308,42 +281,28 @@ proc getMissingBlobs(rman: RequestManager): seq[BlobIdentifier] =
...
-        let finish = SyncMoment.now(uint64(len(fetches)))
-
-        var succeed = 0
-        for worker in workers:
-          if worker.finished() and not(worker.failed()):
-            inc(succeed)
-
-        debug "Request manager blob tick",
-             blobs_count = len(fetches),
-             succeed = succeed,
```
zah, line 339: Why were these properties not considered useful?

beacon_chain/sync/sync_manager.nim
```diff
@@ -158,8 +158,9 @@ proc newSyncManager*[A, B](pool: PeerPool[A, B],
   res.initQueue()
   res
 
-proc getBlocks*[A, B](man: SyncManager[A, B], peer: A,
-                      req: SyncRequest): Future[BeaconBlocksRes] {.async.} =
+proc getBlocks[A, B](man: SyncManager[A, B], peer: A,
+                     req: SyncRequest): Future[BeaconBlocksRes] {.
+                     async: (raises: [CancelledError], raw: true).} =
   mixin getScore, `==`
```
zah, line 164: This `mixin` statement doesn't seem to serve any purpose here.

beacon_chain/sync/sync_manager.nim
```diff
@@ -194,8 +182,8 @@ proc shouldGetBlobs[A, B](man: SyncManager[A, B], e: Epoch): bool =
    e >=  wallEpoch - man.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS)
 
 proc getBlobSidecars[A, B](man: SyncManager[A, B], peer: A,
-                           req: SyncRequest
-                          ): Future[BlobSidecarsRes] {.async.} =
+                           req: SyncRequest): Future[BlobSidecarsRes]
+                           {.async: (raises: [CancelledError], raw: true).} =
   mixin getScore, `==`
```
zah, line 187: Ditto (unnecessary `mixin`)

beacon_chain/sync/sync_manager.nim
```diff
@@ -419,110 +388,97 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A) {.async.} =
...
+  let slots = mapIt(blocks, it[].slot)
+  if not(checkResponse(req, slots)):
+    peer.updateScore(PeerScoreBadResponse)
+    man.queue.push(req)
+    warn "Received blocks sequence is not in requested range",
+          blocks_count = len(blocks), blocks_map = blockSmap,
+          request = req
+    return
+
+  func combine(acc: seq[Slot], cur: Slot): seq[Slot] =
```
zah, line 410: I would look deeper into this. The code seems to use a very inefficient quadratic algorithm with a ton of unnecessary sequence copies and memory allocations. Leave a TODO comment perhaps.

beacon_chain/networking/eth2_network.nim
arnetheduck: all these warnings would start showing if libp2p starts raising new unexpected errors - it's a transition measure while libp2p upgrades to chronosv4

beacon_chain/networking/eth2_network.nim
arnetheduck: an already-finished future passed to await does not introduce an interruption point.

`read` raises additional exceptions because there is no guarantee that it is called only on completed futures - `await` does not have this effect because it's guaranteed to perform the read only on finished futures (by waiting for them to be finished, if necessary).

beacon_chain/networking/eth2_network.nim
arnetheduck: ditto (as the comment points out)

beacon_chain/networking/eth2_network.nim
arnetheduck: it should not result in any stalls (if there are, it's a bug in libp2p) - all other calls to it use `await` already, this is just a straggler

beacon_chain/networking/eth2_protocol_dsl.nim
arnetheduck: it is only useful if the DSL is also upgraded to track exceptions, which is a much larger task involving significant changes to rlpx itself - if/when that is done, it would probably make sense to port this together with other changes

beacon_chain/sync/request_manager.nim
arnetheduck: cc @etan-status these changes are part of a blob syncing bugfix

beacon_chain/sync/sync_manager.nim
arnetheduck: indeed, probably reasonable to clean it up in the next round of fixes (mixin wasn't part of the scope of this PR)

beacon_chain/sync/sync_manager.nim
arnetheduck: https://github.com/status-im/nimbus-eth2/issues/5794 might be an opportunity for this - that said, this PR just removed the unnecessary exception handling, the existing logic wasn't looked at

beacon_chain/sync/request_manager.nim
etan-status: If you are referring to #5766 , I did not touch this file at all.

beacon_chain/networking/eth2_network.nim
zah: There is no benefit in using `warn` vs `debug` for these kind of messages. They create a risk that we'll run into an unintended increase in log volume (to potentially dangerous levels depending on the particular issue that triggers the increase).

arnetheduck: #5822 - `warn` would increase chances that we notice it - swallowed exceptions have historically been a source of several security issues - anyway, hopefully we can be rid of them completely soon.

beacon_chain/networking/eth2_network.nim
zah: When I said "interruption point", I meant that it increases the possible states of the [duff's device](https://en.wikipedia.org/wiki/Duff%27s_device) behind the async function. Each `await` introduces an `yield` in the closure iterator and this changes how the function is compiled (local variables can move in the environment, etc). The compiler cannot know at compile-time that the future is already finished.

Right, I've seen the `read` problem before. I guess we need something similar to `expect` for the `Future` type (i.e. an `unsafeRead` API)

arnetheduck: there is `value` and `error` now that work _almost_ like their `Result` counterparts for the cases where you have checked for `completed`, except they don't `Defect` when the future is in the wrong state and instead return the default - this for legacy reasons where `error == nil` means the future didn't fail, in asyncdispatch.

There's [half of a plan](https://github.com/status-im/nim-chronos/pull/471) to make it a defect but it requires a thorough review - enabling it caused some real trouble in some of our libraries.

With that out of the way, `await` is the cleanest and safest option - `value` is the fastest performance and code-genwise (but somewhat compromised) and `read` ... has no place left really.

I guess it's an engineering tradeoff whether the extra dead code is worth it or not.

arnetheduck: fwiw, I also suspect the duff device would be better off as a set of distinct smaller functions with the state being a pointer to the "next" one.. that's a Nim issue though, more than nimbus

arnetheduck: ah - and one more thing: I think in general, this code would have benefitted from getting rid of the futures earlier on - ie instead of collecting completed requests in a seq, it should be extracting the data from the futures and keep that data in the seq directly - that would remove the need for `await`, `read` and so on completely - this doesn't work for re-checking the outcome of `allFutures` but it  does work for the kind of setup that this code uses.

arnetheduck: also, do you know why a duff device is used? it seems rather inefficient compared to simply generating separate functions for each state along with a function pointer to the next state (this would for example make it easier to avoid stack overflows)

zah: The strategy you are suggesting is a good idea and something that I've advocated for myself. It's also what the CPS libraries are doing (too bad their user-land compilation speed is too low).

PR #5766, opened 2024-01-16
bolster `BlobSidecar` syncing on incomplete responses

etan-status:
Avoid marking blocks invalid when corresponding `blobSidecarsByRange` returns an incomplete / incorrect response while syncing. The block itself may still be valid in that scenario.

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -213,7 +213,7 @@ func has_flag*(flags: ParticipationFlags, flag_index: TimelyFlag): bool =
 
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.4/specs/deneb/p2p-interface.md#check_blob_sidecar_inclusion_proof
 func verify_blob_sidecar_inclusion_proof*(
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/helpers.nim
```diff
@@ -211,6 +211,20 @@ func has_flag*(flags: ParticipationFlags, flag_index: TimelyFlag): bool =
   let flag = ParticipationFlags(1'u8 shl ord(flag_index))
   (flags and flag) == flag
 
+# https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.4/specs/deneb/p2p-interface.md#check_blob_sidecar_inclusion_proof
+func verify_blob_sidecar_inclusion_proof*(
+    blob_sidecar: BlobSidecar): Opt[void] =
```
arnetheduck: `Result[void, cstring]` here?

arnetheduck:
meh, would have been nicer to have built this on top of #5765

etan-status:
a4a08e6342a2212161bf749603a140631b0c672f

arnetheduck:
thanks!

Issue #5730, opened 2024-01-12
nimbus_validator_client does not register validators with mev-relay when used with web3signer.


catwith1hat:
# The good case

The mev-boost process must receive validator registrations periodically in order for mev-boost to work. In the logs this looks like this:
```
Jan 12 11:05:22 node3 mev-boost[1638]: time="2024-01-12T11:05:22.539+01:00" level=info msg="http: POST /eth/v1/builder/validators 200" duration=0.333286 method=POST path=/eth/v1/builder/validators status=200 version=v1.6
Jan 12 11:11:46 node3 mev-boost[1638]: time="2024-01-12T11:11:46.363+01:00" level=info msg="http: POST /eth/v1/builder/validators 200" duration=0.291441 method=POST path=/eth/v1/builder/validators status=200 version=v1.6
Jan 12 11:18:13 node3 mev-boost[1638]: time="2024-01-12T11:18:13.806+01:00" level=info msg="http: POST /eth/v1/builder/validators 200" duration=0.540833 method=POST path=/eth/v1/builder/validators status=200 version=v1.6
```
You can see a period refresh via POST request. When I run the in-process validator inside `nimbus_beacon_node` this works fine. I run nimbus_beacon_node with a remote web3signer. Here are the essential command line flags are (stripped for brevity):
```
/nix/store/ij1gdm2rny1wn4mck3rg9pp24b2zk8z2-nimbus-eth2-24.1.1/bin/nimbus_beacon_node \
  '--suggested-fee-recipient=0xd8da6bf26964af9d7eed9e03e53415d37aa96045' \
  '--in-process-validators=true' \
  '--web3-signer-url=http://localhost:9330' \
  '--payload-builder=true' \
  '--payload-builder-url=http://localhost:18910'
```

# The bad case

Switching to an external validator seems to work fine at first glance (attestion, block building works). However, MEV boost does not work. There are no POST requests in the logs, so I suspect that the mev-boost process stopped receiving registrations. The essentials flags are (stripped for brevity):
```
/nix/store/ij1gdm2rny1wn4mck3rg9pp24b2zk8z2-nimbus-eth2-24.1.1/bin/nimbus_beacon_node \
  '--suggested-fee-recipient=0xd8da6bf26964af9d7eed9e03e53415d37aa96045' \
  '--in-process-validators=false' \
  '--payload-builder=true' \
  '--payload-builder-url=http://localhost:18910'
```
```
[...]

PR #5771, opened 2024-01-17
fix validators not registered when VC is used with web3signer (#5730)

stephanep:
This change ensures that suggested-fee-recipient will be used as default even if the validator directory does not exist when a web3-signer-url is specified


tersec:
Should target `unstable` branch

PR #5774, opened 2024-01-17
fix validators not registered when VC is used with web3signer (#5730)

stephanep:
This change ensures that suggested-fee-recipient will be used as default even if the validator directory does not exist when a web3-signer-url is specified.


tersec:
It seems like https://github.com/status-im/nimbus-eth2/pull/5781 is also aiming to address this issue, but I haven't checked both enough regarding how they'd interact.

stephanep:
no longer relevant since refactoring done in #5781
will check if it also addressed #5730, closing this for now

PR #5775, opened 2024-01-17
import finalized head LC bootstrap on launch

etan-status:
If the initial state replays cover the finalized head, import matching `LightClientBootstrap` into database.

This also addresses this error when light client requests bootstrap from the genesis slot on networks that launch with Altair enabled.

```
{"lvl":"DBG","ts":"2023-10-04 11:17:49.665+00:00","msg":"LC bootstrap unavailable: Sync committee branch not cached","topics":"chaindag_lc","slot":0}
```

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -579,7 +579,7 @@ proc assignLightClientData(
       forkyObject.signature_slot = signature_slot
   ok()
 
-proc createLightClientUpdates(
+proc createLightClientUpdate(
     dag: ChainDAGRef,
     state: ForkyHashedBeaconState,
     blck: ForkyTrustedSignedBeaconBlock,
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -665,6 +665,38 @@ proc createLightClientUpdates(
     current_period_best_update = best,
     latest_signature_slot = latest_signature_slot)
 
+proc createLightClientBootstrap(
+    dag: ChainDAGRef, bid: BlockId): Opt[void] =
```
arnetheduck, line 669: nicer in the long run to return a cstring error here too since this operation generally combines several steps

etan-status: noted, right now this particular instance works better as `Opt[void]` because that's used throughout the module.
The other one int he other PR, I changed to `result[void, string]`.

Is there an advantage of using `cstring` over `string`? Sometimes serializing arguments into the error is useful to have at least some resemblance of where things went wrong (especially because the stack trace is erased with results), cstring prevents that to work. We also use it somewhat inconsistently, sometimes it's cstring sometimes regular string.

arnetheduck: > Is there an advantage of using cstring over string?

one memory allocation + the total Result becoming a gc type.. `string` is simply one step up on the error reporting capability ladder

PR #5783, opened 2024-01-18
use consensus spec v1.4.0-beta.6 test vectors

etan-status:


etan-status:
- https://github.com/ethereum/consensus-specs/releases/tag/v1.4.0-beta.6
- https://en.wikipedia.org/wiki/Psychrolutes_marcidus

PR #5800, opened 2024-01-19
bump `nim-json-rpc` to `8d79d528412d7cdd022ff91fb6bacd4f5bb01d58`

etan-status:
- Fix improper yield usage in json-rpc-websocketserver
- Improve both client and server resilience against fields and elements with null value


etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5801

etan-status:
Tested with sync on goerli. still works fine.

PR #5801, opened 2024-01-19
bump `nim-json-serialization` to `d9394dc7286064902d825bbc1203d03d7218633a`

etan-status:
- Resilience against null fields


etan-status:
Tested sync on Goerli, still works well, so having the extra fixes is probably beneficial.

PR #5819, opened 2024-01-23
use `RestPlainResponse` to improve builder API error reporting

tersec:
recreate https://github.com/status-im/nimbus-eth2/pull/5777

Currently, when relays return unexpected responses, Presto just returns that it failed somehow to parse the JSON, which often isn't a JSON at all but some plain text error message which gets swallowed, along with a potentially helpful `Content-Type` and HTTP status.


tersec:
Release just went out, so this now has testing time.

PR #5824, opened 2024-01-24
bump `holesky` to `9d9aabf2d4de51334ee5fed6c79a4d55097d1a43`

etan-status:
- add `MAX_PER_EPOCH_ACTIVATION_CHURN_LIMIT`


etan-status:
Same default is implicitly pulled from preset implicitly, but this makes it explicit if the implicit fallback doesn't work correctly.

Issue #5794, opened 2024-01-19
With low peers (< 10), sync manager may get stuck in with workers map full of 'w'/'Q'


etan-status:
1. Each sync worker gets their job, slots 0 ..< 16, 16 ..< 32, 32 ..< 48 and so on
2. Assume num available peers < `SyncWorkersCount - 1` --> some workers will be stuck in `w` mode
3. All workers except the first one succeed --> 16 ..< 32, 32 ..< 48 etc are now in `Q` mode
4. First worker fails --> descores peer --> retries task or picks a new one, doesn't matter
5. It will select same peer again, if it's a retry, assume request keeps failing, descore peer until disconnect, then the worker is stuck in `w` mode. If it's a new task, assume it succeeds, then the worker is stuck in `Q` mode.

After a while, every peer is stuck in `w` or in `Q` mode, and 0 ..< 16 request is not fulfilled.

- `w` means 'waiting for peer that is not locked by another sync worker'
- `Q` means 'waiting for parent result to become available before we can process our own result'

Some ideas on possible approaches to tackle this:

- Should we release peers while in `Q` mode so that `w` workers can acquire them? We can still de-score them if they are still around by the time we transition to `P`, but we no longer exclusively control them.
- Should we change `SyncWorkersCount` dynamically based on number of available peers?
- Should we relax descoring / disconnects while having < `SyncWorkersCount` peers?

This popped up in a tiny devnet with just 10 peers, where some of the peers provided incomplete responses (blocks without blobs within the spec mandated retention period), leading to gradual descores / disconnects until this situation arised.

etan-status:
Example:

```
 peers: 5 ❯ finalized: 46e8fa2c:9840 ❯ head: 5c3693c5:9842:15 ❯ time: 15278:2 (244450) ❯ sync: 02h08m (64.43%) 11.2702slots/s (wQQ:157487)/opt
```

this is with modification `SyncWorkersCount = 3` and `checkPeerScore(...) = always true`. it still ends up in `wQQ` situation somehow.

etan-status:
Example:

```
 peers: 3 ❯ finalized: 07f3e42a:7342 ❯ head: fe749777:7344:8 ❯ time: 15297:2 (244754) ❯ sync: 01d12h44m (48.41%) 0.9467slots/s (QwwwwQwwwQ:117503)/opt
```

etan-status:
Regarding the point above:

[...]

PR #5831, opened 2024-01-27
BN, add timeout for sync workers which are waiting in queue.

cheatfate:
Add timeout of `1.seconds` for sync workers which are waiting in `Q` state. It means that first worker waiting in queue will have timeout of `12.seconds`, second worker will have `24.seconds` and so on.


alanaladd21 (approved):

cheatfate:
This should fix https://github.com/status-im/nimbus-eth2/issues/5794

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -390,14 +390,15 @@ proc initFullNode(
       dag.cfg.DENEB_FORK_EPOCH, dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS,
       SyncQueueKind.Forward, getLocalHeadSlot,
       getLocalWallSlot, getFirstSlotAtFinalizedEpoch, getBackfillSlot,
-      getFrontfillSlot, dag.tail.slot, blockVerifier)
+      getFrontfillSlot, dag.tail.slot, blockVerifier,
+      workerBlockWaitTimeout = chronos.seconds(1))
```
etan-status: 1s intuitively feels a bit short, I could see false positives if a block is stuck in P for a while (for example, the occasional multi-second state replay). The underlying problem only occurs in low-peer scenario with bad peers failing to provide data, I think it's okay if it takes a bit longer to recover in that edge case, if it means that the happy case is a bit more reliable.

tests/test_sync_manager.nim
```diff
@@ -1207,3 +1205,69 @@ suite "SyncManager test suite":
...
+      r23 = queue.pop(finishSlot, p3)
+      r24 = queue.pop(finishSlot, p4)
+
+    check:
+      r22.slot == r12.slot
+      r22.count == r12.count
+      r23.slot == r13.slot
+      r23.count == r13.count
+      r24.slot == r14.slot
+      r24.count == r14.count
```
etan-status, line 1268: and, also, if `r11` fails (incomplete download, failed validation and so on), and `p1` goes away due to the corresponding descore for failing to provide correct data, a different peer will eventually pick up `r11`.

otherwise, `r12`/`r13`/`r14` would just get stuck again and again (as before).

beacon_chain/sync/sync_queue.nim
```diff
@@ -577,6 +587,24 @@ proc notInRange[T](sq: SyncQueue[T], sr: SyncRequest[T]): bool =
...
+    (sq.outSlot - sr.lastSlot) div sq.chunkSize
+
+func expectedTime[T](sq: SyncQueue[T], sr: SyncRequest[T]): Duration =
+  ## Returns expected wait time of specific pending request.
+  ## SR3 SR2 SR1 SR0 <CURRENT QUEUE POSITION>
+  ## SR3 expected pending time would be
+  ## <CHUNK SIZE> * <SR2 .. SR0 CHUNKS COUNT> * <BLOCK WAIT TIME>
+  nanoseconds(
+    int64(sq.chunkSize * sq.chunksCount(sr)) *
+      sq.pendingWorkerBlockWaitTime.nanoseconds)
```
etan-status, line 607: this is the value if every single request takes the maximum time to complete a single time.

- reality could be longer, e.g., when there are retries of earlier sync requests
- reality could be faster, e.g., if the time is computed while there are many prior requests, but then those prior requests complete quickly

I wonder if a simpler mechanism with a static, e.g., 30sec timeout, could also mitigate the risk of getting stuck. It would take a bit longer to unstuck than the current solution, but is simpler to reason about.

Alternatively, to get it fully correct may involve having to re-schedule the timeouts whenever a prior request completes.

etan-status (approved):
The fix indeed should fix the low-peer scenario, and is relatively tiny.

It's probably okay to push this after the Deneb Mainnet release, just in case a regression gets introduced, as there is no downgrade possibility from the Deneb release back to a Capella release (they cannot connect to Deneb networks).

tests/test_sync_manager.nim
cheatfate: Yep in old version `r12`, `r13` and `r14` will stuck waiting for peer to appear which should provide `r11` again.

etan-status:
On Goerli, a similar situation actually comes up even though the number of available peers is very high.

Sync manager only considers peers viable that report a higher slot progress than the local head. However, because goerli is partitioned into split views and proposals are infrequent, there are long stretches where the local head may be higher than the peer's branches. This leads to a situation where < 10 peers are actually viable for sync manager at a time, and I have observed the situation where the Q status workers temporarily could not proceed for minutes because all the other workers were stuck in U/R stage.

Regarding timings, the D/P stages can take quite long, and are the only way to eventually unstuck Q workers. If we just want to specify a single timeout number, I think Q should wait for at least 30-90 seconds based on manual observations of sync progress. Alternatively, a shorter timeout may be suitable if it only is applied while no other worker is in D/P stage, e.g., 5-15 seconds. While other workers are in D/P, I don't think a timeout is needed, as the situation will resolve itself eventually, but if a single timeout value makes the implementation easier, 30-90 seconds should not do too much harm.

cheatfate:
This PR is outdated because of different approach being implemented and merged.

PR #5840, opened 2024-01-29
mention publicness of fee recipients

tersec:
Most settings are either not public or obviously so (validator index, validator pubkey, graffiti, et cetera).

There have been multiple people confused about (suggested) fee recipients, though.

### DIFF
--- a/docs/the_nimbus_book/src/suggested-fee-recipient.md
+++ b/docs/the_nimbus_book/src/suggested-fee-recipient.md
@@ -28,6 +28,7 @@ For each validator, it selects from the first available, in the following order:
 For example, `nimbus_beacon_node --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` suggests to the execution client that `0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` might be the coinbase.
 If this Nimbus node has two validators, one of which has its own suggested fee recipient via the keymanager API and the other does not, the former would use its own per-validator suggested fee recipient, while the latter would fall back to `0x70E47C843E0F6ab0991A3189c28F2957eb6d3842`.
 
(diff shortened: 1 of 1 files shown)


docs/the_nimbus_book/src/suggested-fee-recipient.md
```diff
@@ -28,6 +28,7 @@ For each validator, it selects from the first available, in the following order:
 For example, `nimbus_beacon_node --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` suggests to the execution client that `0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` might be the coinbase.
 If this Nimbus node has two validators, one of which has its own suggested fee recipient via the keymanager API and the other does not, the former would use its own per-validator suggested fee recipient, while the latter would fall back to `0x70E47C843E0F6ab0991A3189c28F2957eb6d3842`.
 
+Fee recipients are recorded publicly on-chain as part of proposed blocks, so suggested fee recipients should be allow for this.
```
etan-status:
```suggestion
Fee recipients are recorded publicly on-chain as part of proposed blocks, so suggested fee recipients should be allowed for this.
```

tersec: Oops. The suggested change does sort of fix the grammar, but doesn't convey the intent -- https://github.com/status-im/nimbus-eth2/pull/5840/commits/e45e321439f577be8cad66d49f302cecfe45e889 (it's not that fee recipient should _be_ allowed for anything, or not, it's that it's that one should consider this aspect of how they're used, without (a) strictly speaking, passive tenses, or (b) exhorting any "you" or "one" in the documentation and keeping it focused on how Ethereum and Nimbus work, not what "you" or "one" should/might/can/must/etc do)

PR #5829, opened 2024-01-26
implement getBlindedBlock REST API

tersec:
https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.4.2#/Beacon/getBlindedBlock

PR #5842, opened 2024-01-30
VC: Use produceBlockV3 when its available.

cheatfate:


beacon_chain/validator_client/common.nim
```diff
@@ -114,7 +114,8 @@ type
     NoTimeCheck
 
   RestBeaconNodeFeature* {.pure.} = enum
-    NoNimbusExtensions  ## BN do not supports Nimbus Extensions
+    NoNimbusExtensions, ## BN do not supports Nimbus Extensions
```
tersec:
```suggestion
    NoNimbusExtensions, ## BN does not support Nimbus Extensions
```

beacon_chain/validator_client/common.nim
```diff
@@ -114,7 +114,8 @@ type
...
+    NoProduceBlockV3    ## BN do not supports produceBlockV3 call
```
tersec:
```suggestion
    NoProduceBlockV3    ## BN does not support produceBlockV3 call
```

beacon_chain/validator_client/block_service.nim
```diff
@@ -428,6 +590,55 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
...
+        debug "RANDAO reveal production has been interrupted"
+        raise exc
+      except CatchableError as exc:
+        error "An unexpected error occurred while receiving RANDAO data",
+              error_name = exc.name, error_msg = exc.msg
+        return
+
+
+  if vc.isProduceBlockV3Supported():
+    # We call `V3` first, if call is failed and `isProduceBlockV3Supported()`
```
```suggestion
    # We call `V3` first, if call fails and `isProduceBlockV3Supported()`
```

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -1387,98 +1390,139 @@ template prepareForkedBlockReading(reader: var JsonReader[RestJson],
...
-        Opt.some(RestJson.decode(string(data.get()),
-                                 deneb.BeaconBlock,
-                                 requireAllFields = true,
-                                 allowUnknownFields = true))
-      except SerializationError:
-        Opt.none(deneb.BeaconBlock)
-    if res.isNone():
-      reader.raiseUnexpectedValue("Incorrect deneb block format")
-    value = ForkedBeaconBlock.init(res.get()).BlockType
+# proc readValue*[BlockType: ForkedBeaconBlock](
```
tersec: Having an entire commented-out section like this seems like a WIP? At least, maybe using a `when false` would clarify intent, though it seems dubious to add either to `unstable`.

beacon_chain/spec/forks.nim
```diff
@@ -657,6 +657,27 @@ func init*(T: type ForkedSignedBlindedBeaconBlock,
...
+      message: blck, signature: signature))
+
+template init*(T: type ForkedSignedBlindedBeaconBlock,
+               blck: deneb_mev.BlindedBeaconBlock, blockRoot: Eth2Digest,
+               signature: ValidatorSig): T =
+  T(kind: ConsensusFork.Deneb,
+    denebData: deneb_mev.SignedBlindedBeaconBlock(
+      message: blck, signature: signature))
+
+template init*(T: type ForkedSignedBlindedBeaconBlock,
```
tersec: Nimbus doesn't support the Bellatrix builder API.

https://github.com/status-im/nimbus-eth2/pull/5829 re-added the actual (not stub-empty-`object` `bellatrix_mev.BlindedBeaconBlock`, but that's only to support https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.4.2#/Beacon/getBlindedBlock which is part of the beacon API, not builder API.

There should never be a reason to be calling the Bellatrix builder API.

beacon_chain/validator_client/block_service.nim
```diff
@@ -428,6 +590,55 @@ proc publishBlock(vc: ValidatorClientRef, currentSlot, slot: Slot,
...
+      except CancelledError as exc:
+        debug "RANDAO reveal production has been interrupted"
+        raise exc
+      except CatchableError as exc:
+        error "An unexpected error occurred while receiving RANDAO data",
+              error_name = exc.name, error_msg = exc.msg
+        return
+
+
+  if vc.isProduceBlockV3Supported():
```
tersec, line 638: This design seems a little inefficient? `isProduceBlockV3Supported` is tracked on a per-BN basis, so it should be possible to target the V2/V3 preference on a per-BN basis as well.

cheatfate: Why inefficient, it allows us to not introduce any more options to CLI. 
If there is only one BN supplied:
1. First call to `V3` will reveal absence of `V3` and so `V2` will always be used instead of `V3`.
If there is many BNs supplied:
1. In case when ALL BNs missing `V3` or all BNs have `V3` everything will happened like in case with one BN.
2. In case when only some BNs has `V3` and some doesn't - this solution is not optimal, because it will send `V3` call to this nodes again and again. And even if such solution is not efficient, its enough for time frame while we in `V2` deprecation process. Otherwise there should be introduced too many changes in the code and all this changes will be reverted soon as `V2` deprecation happens.

tersec (approved):

beacon_chain/validator_client/block_service.nim
tersec: Okay, this is a good argument, especially (3). That had been my concern, but I agree with you now.

Issue #5843, opened 2024-01-31
beacon API eventstream `blob_sidecar` topic versioned hash incorrect format


Savid:
**Describe the bug**
The `blob_sidecar` topic payload in the [beacon API event stream](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Events/eventstream) has incorrect `versioned_hash` formatted value, its not in hex string format.

Current payload;
```
{"block_root":"0x28f116a750ddbe6d0f66a6bdf2efb5b50605d5c5d9308d48b79033591293aa32","index":"0","slot":"4244117","kzg_commitment":"0x878d73147b9628205180f52bd05d3f8f2a9ad3915d4db05af29a80857e6899f2a2277362fcf961e531e755146dbb95c3","versioned_hash":["1","99","31","214","22","60","236","98","137","10","15","116","245","208","107","65","34","86","160","104","92","59","92","104","94","8","176","35","159","80","54","246"]}
```

Expected payload;

```
{"block_root":"0x28f116a750ddbe6d0f66a6bdf2efb5b50605d5c5d9308d48b79033591293aa32","index":"0","slot":"4244117","kzg_commitment":"0x878d73147b9628205180f52bd05d3f8f2a9ad3915d4db05af29a80857e6899f2a2277362fcf961e531e755146dbb95c3","versioned_hash":"0x01631fd6163cec62890a0f74f5d06b412256a0685c3b5c685e08b0239f5036f6"}
```

Tested on commit [`0638741`](https://github.com/status-im/nimbus-eth2/commit/0638741)

**To Reproduce**

```bash
curl "http://127.0.0.1:5052/eth/v1/events?topics=blob_sidecar" -H "accept: text/event-stream"
```

tersec:
https://github.com/status-im/nimbus-eth2/pull/5844

PR #5844, opened 2024-01-31
fix `blob_sidecar` SSE `versioned_hash` field to be 0x-prefixed hex

tersec:
https://github.com/status-im/nimbus-eth2/issues/5843


samcm:
Thanks @tersec!!

Issue #5675, opened 2023-12-20
VC only resolves BN DNS IPs at initialization and can't follow BN DNS changes after


tersec:
https://github.com/status-im/nimbus-eth2/blob/3a527d622d6ba9c0588c5165f77f7e5d454dc3cb/beacon_chain/validator_client/common.nim#L742-L774
is where the VC resolves DNS addresses to IPs via `getHttpAddress`.

Once it has the `HttpAddress` from `getHttpAddress`, it continues to use this. This means that it never invalidates the cache after the IP it has doesn't correspond to the `getaddrinfo()`-canonical information anymore.

`getHttpAddress` itself unconditionally uses `resolveTAddress()`, `getAddrInfo()`, and `getaddrinfo()` every call to resolve, so there's no additional caching in that layer; this is specificially VC functionality.

tersec:
https://github.com/status-im/nimbus-eth2/pull/5846

PR #5846, opened 2024-02-03
VC: Fix timecheck log statements and DNS resolve cache issue #5675

cheatfate:


tersec (approved):
On balance, sure, this seems like an improvement -- people concerned about blocking DNS lookups (which as far as I know still is the case?) can either use IP addresses or some `/etc/hosts`-equivalent to effectively guarantee that won't become a practical issue, and for one-BN configurations, invoking this recovery codepath essentially implies small latencies won't matter in that moment anyway, if doppelganger protection's going to be active.

Also, reasonably often, at least personally/anecdotally, when I notice "DNS" failing, it's just a canary for Internet connectivity failing as a whole, so even skipping it doesn't help.

One concern I'd have would be, does this risk in multiple-BN configurations where the DNS isn't so dynamic as in a Kubernetes container or similar configuration, so repeatedly lookups wouldn't have been expected to change results, and there's for some reason a plausibility of DNS resolution timeouts long enough to disrupt other, hitherto-working VC/BN pairings by timing out on DNS lookup for an incorrect pairing. Right now, that doesn't happen, but of course, it also never figures out any updated IP addresses, either.

But even as-is, I'd argue using PR reflects a better risk balance than not using this PR.

PR #5860, opened 2024-02-07
consider block availability when initializing LC data collector

etan-status:
When using checkpoint sync, the initial block is missing in the DB. Update the LC data collector initialization to account for that, avoiding a spurious error message when it is incorrectly accessed:

```
ERR 2024-02-07 11:21:55.416+01:00 Block failed to load unexpectedly          topics="chaindag_lc" bid=d30517a7:8257504 tail=8257504
```

Also fixes a regression from #5691 that resulted in similar messages while importing the first few blocks after checkpoint sync.

Thanks to @arnetheduck for reporting this.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
@@ -631,7 +631,11 @@ proc createLightClientUpdate(
   # Check if light client data improved
   let
     finalized_slot = attested_data.finalized_slot
-    finalized_bsi = dag.getExistingBlockIdAtSlot(finalized_slot)
+    finalized_bsi =
+      if finalized_slot >= dag.tail.slot:
+        dag.getExistingBlockIdAtSlot(finalized_slot)
+      else:
+        Opt.none(BlockSlotId)
     has_finality =
       finalized_bsi.isSome and
       finalized_bsi.get.bid.slot >= dag.tail.slot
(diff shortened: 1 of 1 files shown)


beacon_chain/consensus_object_pools/blockchain_dag_light_client.nim
```diff
@@ -702,8 +706,36 @@ proc initLightClientDataCache*(dag: ChainDAGRef) =
...
+  # Initialize tail slot.
+  # Both state and blocks must be available to construct light client data,
+  # see `cacheLightClientData`. If blocks are unavailable, most parts of the
+  # `LightClientHeader` could be reconstructed from the corresponding post-state
+  # but `execution_branch` (since Capella) requires full block availability.
+  # If the assumed-to-be-available range of states / blocks turns out to be
+  # actually unavailable (database inconsistencies), the `tailSlot` will adjust
+  # using `handleUnexpectedLightClientError`. This is unexpected and is logged,
+  # but recoverable by excluding the unavailable range from LC data collection.
+  let targetTailSlot = max(
```
arnetheduck, line 718: not sure if it matters but `backfill` is not updated when pruning so it will sometimes point to a spot where there are no blocks any more

etan-status: Yeah, stumbled upon that one as well. `backfill` goes to `GENESIS_SLOT` anyway on completion (reaching `horizon`).

In general, it seems messy and wrong to require such complex `if` tree just to find out what is the expected range of blocks that can be reasonably assumed to be available in happy case (no db inconsistencies).

Will fix the bug next that we may end up with a gap for the checkpoint block, if the peer gives us an almost complete response for a backfill that solely excludes the checkpoint block itself.

etan-status: As long as `dag.tail` is not getting pruned, it should be fine, btw. And moving dag.tail is also ok.

arnetheduck: dag.tail does get pruned ;)

PR #5863, opened 2024-02-07
fix checkpoint block potentially not getting backfilled into DB

etan-status:
When using checkpoint sync, only checkpoint state is available, block is not downloaded and backfilled later.

`dag.backfill` tracks latest filled `slot`, and latest `parent_root` for which no block has been synced yet.

In checkpoint sync, this assumption is broken, because there, the start `dag.backfill.slot` is set based on checkpoint state slot, and the block is also not available.

However, sync manager in backward mode also requests `dag.backfill.slot` and `block_clearance` then backfills the checkpoint block once it is synced. But, there is no guarantee that a peer ever sends us that block. They could send us all parent blocks and solely omit the checkpoint block itself. In that situation, we would accept the parent blocks and advance `dag.backfill`, and subsequently never request the checkpoint block again, resulting in gap inside blocks DB that is never filled.

To mitigate that, the assumption is restored that `dag.backfill.slot` is the latest filled `slot`, and `dag.backfill.parent_root` is the next block that needs to be synced. By setting `slot` to `tail.slot + 1` and `parent_root` to `tail.root`, we put a fake summary into `dag.backfill` so that `block_clearance` only proceeds once checkpoint block exists.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -23,11 +23,12 @@ OK: 1/1 Fail: 0/1 Skip: 0/1
 OK: 12/12 Fail: 0/12 Skip: 0/12
 ## Backfill
 ```diff
++ Backfill to genesis                                                                        OK
 + Init without genesis / block                                                               OK
-+ backfill to genesis                                                                        OK
-+ reload backfill position                                                                   OK
++ Reload backfill position                                                                   OK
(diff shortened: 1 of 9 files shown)


etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5875

PR #5865, opened 2024-02-08
do not cache zero block hash if block unavailable

etan-status:
With checkpoint sync, the checkpoint block is typically unavailable at the start, and only backfilled later. To avoid treating it as having zero hash, execution disabled in some contexts, wrap the result of `loadExecutionBlockHash` in `Opt` and handle block hash being unknown.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -2226,20 +2226,25 @@ proc pruneHistory*(dag: ChainDAGRef, startup = false) =
           if dag.db.clearBlocks(fork):
             break
 
-proc loadExecutionBlockHash*(dag: ChainDAGRef, bid: BlockId): Eth2Digest =
+proc loadExecutionBlockHash*(
+    dag: ChainDAGRef, bid: BlockId): Opt[Eth2Digest] =
   let blockData = dag.getForkedBlock(bid).valueOr:
-    return ZERO_HASH
+    # Besides database inconsistency issues, this is hit with checkpoint sync.
+    # The initial `BlockRef` is creted before the checkpoint block is loaded.
+    # It is backfilled later, so return `none` and keep retrying.
+    return Opt.none(Eth2Digest)
 
   withBlck(blockData):
     when consensusFork >= ConsensusFork.Bellatrix:
-      forkyBlck.message.body.execution_payload.block_hash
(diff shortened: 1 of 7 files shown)


etan-status:
This also fixes random pandas popping up a couple minutes after a checkpoint sync with `--backfill=false`.

image

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -2226,20 +2226,25 @@ proc pruneHistory*(dag: ChainDAGRef, startup = false) =
...
+      Opt.some forkyBlck.message.body.execution_payload.block_hash
     else:
-      ZERO_HASH
+      Opt.some ZERO_HASH
 
-proc loadExecutionBlockHash*(dag: ChainDAGRef, blck: BlockRef): Eth2Digest =
+proc loadExecutionBlockHash*(
+    dag: ChainDAGRef, blck: BlockRef): Opt[Eth2Digest] =
   if blck.executionBlockHash.isNone:
-    blck.executionBlockHash = Opt.some dag.loadExecutionBlockHash(blck.bid)
```
tersec, line 2293: is it possible to use `valueOr` here to avoid the `unsafeGet`?

etan-status: the `.unsafeGet` is removed in the PR, not added

tersec: oops, indeed

tersec (approved):

beacon_chain/validators/beacon_validators.nim
```diff
@@ -767,7 +767,14 @@ proc getBlindedBlockParts[
     Future[Result[(EPH, UInt256, ForkedBeaconBlock), string]]
     {.async: (raises: [CancelledError]).} =
   let
-    executionBlockHash = node.dag.loadExecutionBlockHash(head)
+    executionBlockHash = node.dag.loadExecutionBlockHash(head).valueOr:
+      # With checkpoint sync, the checkpoint block may be unavailable,
+      # and it could already be the parent of the new block before backfill.
+      # Fallback to EL, hopefully the block is available on the local path.
+      warn "Failed to load parent execution block hash",
```
arnetheduck:
```suggestion
      warn "Failed to load parent execution block hash, skipping block builder",
```

arnetheduck (approved):

PR #5869, opened 2024-02-08
allow `getBlockIdAtSlot` to answer queries from available states

etan-status:
After checkpoint sync, historical block IDs cannot yet be queried. However, they are needed to compute dependent roots of `ShufflingRef`. To allow lookup, enable `getBlockIdAtSlot` to answer from compatible states in memory; as long as they descend from the finalized checkpoint and the requested slot is sufficiently recent, `block_roots` contains everything to recover `BlockSlotId` up to `SLOTS_PER_HISTORICAL_ROOT`. This is similar to how `attester_dependent_root` etc. are computed.

This accelerates the first couple minutes of checkpoint sync on Mainnet, especially the time until finality advances past the synced checkpoint.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -173,6 +173,36 @@ func getBlockRef*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockRef] =
   else:
     err()
 
+func getBlockIdAtSlot*(
+    state: ForkyHashedBeaconState, slot: Slot): Opt[BlockSlotId] =
+  ## Use given state to attempt to find a historical `BlockSlotId`.
+  if slot > state.data.slot:
+    return Opt.none(BlockSlotId)  # State does not know about requested slot
+  if state.data.slot > slot + SLOTS_PER_HISTORICAL_ROOT:
+    return Opt.none(BlockSlotId)  # Cache has expired
+
+  var idx = slot mod SLOTS_PER_HISTORICAL_ROOT
+  let root =
+    if slot == state.data.slot:
+      state.latest_block_root
+    else:
+      state.data.block_roots[idx]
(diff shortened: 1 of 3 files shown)


etan-status:
```
 peers: 16 ❯ finalized: ccb64aa1:261980 ❯ head: f998de42:261981:31 ❯ time: 261982:25 (8383449) ❯ sync: 00h00m (71.11%) 2.6301slots/s (RRRsRssRRs:8383423)/opt
```

about 2 minutes to sync from finalized checkpoint

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -173,6 +173,36 @@ func getBlockRef*(dag: ChainDAGRef, root: Eth2Digest): Opt[BlockRef] =
   else:
     err()
 
+func getBlockIdAtSlot*(
```
arnetheduck, line 176: why not integrate this in the "main" `getBlockIdAtSlot`? it looks preferable to always use the state for this lookup instead of hitting the database..

also, now that we're here, it's may be a good idea to "top up" the `finalizedBlocks` from the state on startup since the latter is used for a bunch of decisions about block validity which otherwise must be answered with "unknown". This change however would have deeper implications on backfill in particular since afair, we don't store an explicit "backfill resume point" in the database.

etan-status: It is already integrated into the main `getBlockIdAtSlot` further down below. 

The reasons why the function is factored out:
- to enable the testing, because we don't use `import {.all.}` it is also exported, but there is no reason to directly call it outside the tests.
- to make the calelr logic that uses the three available memory states more compact / easier to reason about.

Yes, we can't just put it into DB because otherwise we never backfill those blocks. At least not for the Deneb release which should be as minimal risk as possible.

PR #5873, opened 2024-02-09
Consensus block value calculation for produceBlockV3 API call.

cheatfate:


beacon_chain/validators/rewards.nim
```diff
@@ -0,0 +1,187 @@
...
+      doAssert aggregate.sync_committee_bits.len == SYNC_COMMITTEE_SIZE
+      doAssert forkyState.data.current_sync_committee.pubkeys.len ==
+        SYNC_COMMITTEE_SIZE
+
+      for i in 0 ..< SYNC_COMMITTEE_SIZE:
+        if aggregate.sync_committee_bits[i]:
+          proposerOutcome += proposer_reward
+
+proc gweiToWei(eth: Gwei): UInt256 =
+  eth.u256 * 1000000000000000000.u256
```
tersec: This calculation is already done in a few other places.

https://github.com/status-im/nimbus-eth2/blob/1575478b721b6f62d6466a226b91188bdf97e00b/beacon_chain/el/deposit_contract.nim#L139-L139

https://github.com/status-im/nimbus-eth2/blob/1575478b721b6f62d6466a226b91188bdf97e00b/beacon_chain/el/deposit_contract.nim#L146-L147

https://github.com/status-im/nimbus-eth2/blob/1575478b721b6f62d6466a226b91188bdf97e00b/ncli/ncli_testnet.nim#L520-L520

https://github.com/status-im/nimbus-eth2/blob/1575478b721b6f62d6466a226b91188bdf97e00b/ncli/ncli_testnet.nim#L527-L528

It's hard to verify at a glance that numeric constant has exactly the correct number of zeros, et cetera everywhere. It's a prime candidate for being `const`-defined somewhere, which could be used here too.

beacon_chain/validators/rewards.nim
```diff
@@ -0,0 +1,187 @@
...
+      const consensusFork {.inject, used.} = ConsensusFork.Bellatrix
+      template forkyState: untyped {.inject.} = s.bellatrixData
+      template forkyBlck: untyped {.inject.} = b.bellatrixData
+      body
+    of ConsensusFork.Altair:
+      const consensusFork {.inject, used.} = ConsensusFork.Altair
+      template forkyState: untyped {.inject.} = s.altairData
+      template forkyBlck: untyped {.inject.} = b.altairData
+      body
+    of ConsensusFork.Phase0:
```
tersec: Is this ever going to used for historical blocks, or is it only for blocks under proposal?

cheatfate: I just copied and adopted `forks.withStateAndBlck` template.

beacon_chain/validators/rewards.nim
```diff
@@ -0,0 +1,187 @@
...
+      template forkyBlck: untyped {.inject, used.} = b.phase0Data
+      body
+  elif b is capella_mev.BlindedBeaconBlock:
+    const consensusFork {.inject, used.} = ConsensusFork.Capella
+    template forkyState: untyped {.inject.} = s.capellaData
+    template forkyBlck: untyped {.inject.} = b
+  elif b is deneb_mev.BlindedBeaconBlock:
+    const consensusFork {.inject, used.} = ConsensusFork.Deneb
+    template forkyState: untyped {.inject.} = s.denebData
+    template forkyBlck: untyped {.inject.} = b
```
etan-status:
```suggestion
    template forkyBlck: untyped {.inject.} = b
  else:
    {.error: "withStateAndMaybeBlindedBlock does not support " & $typeof(b).}
```
(so that it doesn't silently become a no-op when electra gets added)

beacon_chain/validators/rewards.nim
cheatfate: Fixed.

beacon_chain/spec/beaconstate.nim
```diff
@@ -550,23 +567,41 @@ func get_attestation_participation_flag_indices(
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.6/specs/deneb/beacon-chain.md#modified-get_attestation_participation_flag_indices
 func get_attestation_participation_flag_indices(
     state: deneb.BeaconState,
-    data: AttestationData, inclusion_delay: uint64): set[TimelyFlag] =
+    data: AttestationData, inclusion_delay: uint64,
```
arnetheduck: @tersec's changes should not be needed now that the correct state is used

tersec: https://github.com/status-im/nimbus-eth2/pull/5873/commits/0675cb9810953bc98221363cb23e1c85309a1aff

beacon_chain/spec/helpers.nim
```diff
@@ -523,3 +523,9 @@ proc blockToBlockHeader*(blck: ForkyBeaconBlock): ExecutionBlockHeader =
 
 proc compute_execution_block_hash*(blck: ForkyBeaconBlock): Eth2Digest =
   rlpHash blockToBlockHeader(blck)
+
+func toWei*(eth: SomeInteger): UInt256 =
+  eth.u256 * 1000000000000000000.u256
```
etan-status: problem with the name here is that there is also `Gwei` with a different scaling factor. So, the function used to have `ethToWei` explicit for untyped integer.

With `toWei`, you don't know if it takes `Eth` or `Gwei` from a client perspective.

tersec:
https://github.com/status-im/nimbus-eth2/pull/6022 doesn't replace the "front-end" parts of this, but does provide a new back-end calculation approach

beacon_chain/spec/helpers.nim
cheatfate: I have reverted all the changes, so its not an issue anymore.
https://github.com/status-im/nimbus-eth2/pull/5873/commits/56f0e59f63fc602e20a1b6ba0f3926372069b2ba

beacon_chain/validators/rewards.nim
cheatfate: Removed whole file.

beacon_chain/validator_client/block_service.nim
```diff
@@ -308,6 +330,15 @@ proc publishBlockV3(vc: ValidatorClientRef, currentSlot, slot: Slot,
           else:
             forkyMaybeBlindedBlck.`block`
         )
+
+      debug "Block produced",
+            block_type = "normal",
```
```suggestion
            block_type = "engine",
```

tersec: Or `non-blinded` or similar, but then that defines things in terms of blindedness.

"engine" isn't perfect either, since some BNs are willing to return a blinded block from the engine API here (Nimbus's BN won't, but this is VC code, and it shouldn't assume any particular BN). But it's not inaccurate, since a non-blinded block can only currently be from the engine API.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -104,6 +107,12 @@ proc getValidator*(validators: auto,
     Opt.some ValidatorAndIndex(index: ValidatorIndex(idx),
                                validator: validators[idx])
 
+func blockConsensusValue(r: BlockRewards): UInt256 =
+  # Returns value of `block-consensus-value` in Wei units.
+  (uint64(r.attestations) + uint64(r.sync_aggregate) +
```
tersec: None of these explicit `uint64` conversions is necessary, because `Gwei` is defined as a non-`distinct` alias of `uint64`:
https://github.com/status-im/nimbus-eth2/blob/2d519ac7153eda830e070dbe23e7b1796bd218e3/beacon_chain/spec/datatypes/base.nim#L125

beacon_chain/validators/beacon_validators.nim
```diff
@@ -104,6 +107,12 @@ proc getValidator*(validators: auto,
...
+    uint64(r.proposer_slashings) +
+    uint64(r.attester_slashings)).u256 * 1000000000.u256
```
tersec: This `1000000000.u256` is unnecessarily runtime-constructed, complete with the customary usually pointless `nimZeroMem()`:
```c
N_LIB_PRIVATE N_NIMCALL(void,
                        blockConsensusValue__validatorsZbeacon95validators_112)(
    tyObject_BlockRewards__9c3PGRoMCH9aJ4gpUEVcHWJA *r,
    tyObject_StUint__ExULMdD1zVCTKy89cyuy4Dg *Result) {
  tyObject_StUint__ExULMdD1zVCTKy89cyuy4Dg T1_;
  tyObject_StUint__ExULMdD1zVCTKy89cyuy4Dg T2_;
  nimZeroMem((void *)(&T1_), sizeof(tyObject_StUint__ExULMdD1zVCTKy89cyuy4Dg));
  u256__specZvalidator_12417(
      (NU64)((NU64)((NU64)((NU64)((NU64)((NU64)((*r).attestations) +
                                         (NU64)((*r).sync_aggregate))) +
                           (NU64)((*r).proposer_slashings))) +
             (NU64)((*r).attester_slashings)),
      (&T1_));

  nimZeroMem((void *)(&T2_), sizeof(tyObject_StUint__ExULMdD1zVCTKy89cyuy4Dg));
  u256__OOZvendorZnim45ethZethZcommonZeth95types_308(((NI)1000000000), (&T2_));

  star___OOZvendorZnim45stintZstint_435((&T1_), (&T2_), Result);
}
```

Better to define it as a `const` to avoid the
```c
  nimZeroMem((void *)(&T2_), sizeof(tyObject_StUint__ExULMdD1zVCTKy89cyuy4Dg));
  u256__OOZvendorZnim45ethZethZcommonZeth95types_308(((NI)1000000000), (&T2_));
```
part.

cheatfate: I wish not to change it, because its annoying, please re-read all the comments before and lets end it please. If you think its correct, lets use it as it is, because i wish not to check c-code on this specific thing, there is a lot more `u256` conversions around which was not highlighted.
Also block-proposal is made once per 12 seconds on mainnet, so its CPU intensive operation.

beacon_chain/validators/beacon_validators.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/5873/commits/a034d5c42d111cee9038436a7cd642845b534000

beacon_chain/validator_client/block_service.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/5873/commits/a034d5c42d111cee9038436a7cd642845b534000

PR #5876, opened 2024-02-09
reduce verbosity of `Got request for pre-backfill slot`

etan-status:
When syncing, we log a notice each time someone asks us for a block that we haven't backfilled yet. This is quite verbose and not unexpected, because the status message does not allow indicating backfill progress.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1633,13 +1633,8 @@ proc getBlockRange*(
     head = shortLog(dag.head.root), requestedCount, startSlot, skipStep, headSlot
 
   if startSlot < dag.backfill.slot:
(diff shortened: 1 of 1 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1633,13 +1633,8 @@ proc getBlockRange*(
     head = shortLog(dag.head.root), requestedCount, startSlot, skipStep, headSlot
 
   if startSlot < dag.backfill.slot:
-    if startSlot < dag.horizon:
-      # We will not backfill these
```
etan-status, line 1637: Not really reachable (once backfill completes, backfill.slot == GENESIS_SLOT so the entire early return is no longer hit)

tersec (approved):

PR #5881, opened 2024-02-11
state loading optimizations

arnetheduck:
* compute post-merge randao mix without loading state
* avoid copying state on shuffling computation and compute epochref
* speed up state copy for block production

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -1524,7 +1524,7 @@ proc computeRandaoMix(
   # Fall back to database
   dag.computeRandaoMixFromDatabase(bid, lowSlot)
 
-proc computeRandaoMix*(dag: ChainDAGRef, bid: BlockId): Opt[Eth2Digest] {.deprecated.} =
+proc computeRandaoMix*(dag: ChainDAGRef, bid: BlockId): Opt[Eth2Digest] =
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1514,10 +1514,17 @@ proc computeRandaoMix(
...
+    bdata = ? dag.getForkedBlock(bid)
+    fullMix = computeRandaoMix(bdata)
+  if fullMix.isSome:
+    return fullMix
+
   # Fall back to database
   dag.computeRandaoMixFromDatabase(bid, lowSlot)
 
-proc computeRandaoMix*(dag: ChainDAGRef, bid: BlockId): Opt[Eth2Digest] =
+proc computeRandaoMix*(dag: ChainDAGRef, bid: BlockId): Opt[Eth2Digest] {.deprecated.} =
```
tersec: Well, deprecated, or is it used still?

Typically, one sees this `deprecated`-but-not-removed in e.g., libraries where one can't centrally fix all callers at once.

My read of this is that there are callers remaining that would be outside the scope of this elegantly minimal/low-risk PR to fix?

arnetheduck: oops, debug leftover to see if it was used: af7da7c3a1b8e410ded6cf788fbacc64c793f1d3

tersec (approved):

PR #5882, opened 2024-02-12
blob sidecar era/erb proposal

tersec:
There's some potential value for an Era file equivalent for blob sidecars.

It should intentionally mirror Era file as much as makes sense. Some parts don't, such as emphasizing the states. They're simply not going to be fully verifiable in a standalone way, nor is there much value for their being such.

This version just modified the Era description, as it's easier to work within that regard to start with.

The name is also a stand-in, though, `b` for blobs isn't necessarily bad.

Some main real design decisions are
- TLV or SSZ on the list aspect of the `BlobSidecar`s. Arguments for both, but my preference is for SSZs.
- Whether it's worth adding even a modicum of ability to internally verify the blob sidecar e2s files without immediate reference to external Era files. This could take a couple of forms, for example: cryptographic hashes or similar of _uncompressed_ entries, and some kind of block header storage. But the latter seems not that useful. Geth seems to believe that hashes are a good verification mechanism and if done properly, seem reasonable. Snappy has its own checksums, but they're not designed against adversaries. It's not necessarily required to have either or any such mechanism though; these should be regarded as functionally subsidiary to Era files, such that uses cases exist for Era files without the blob equivalents, but likely not vice versa.

### DIFF
--- a/docs/e2store.md
+++ b/docs/e2store.md
@@ -116,6 +116,14 @@ data: snappyFramed(ssz(BeaconState))
 
 The fork and thus the exact format of the `BeaconState` should be derived from the `slot`.
 
+## BlobCompressedSidecars
+```
+type: [0x02, 0x00]
+data: ssz(List[snappyFramed(BlobSidecar), MAX_BLOBS_PER_BLOCK])
+```
+
+`BlobCompressedSidecars` contain a list of `BlobSidecar` objects encoded using `SSZ` then compressed using the snappy [framing format](https://github.com/google/snappy/blob/master/framing_format.txt).
+
 ## Empty
 
 ```
@@ -169,7 +177,9 @@ def read_slot_index(f):
   return (start_slot, record_start, slot_offsets)
 ```
 
-# Era files
+# Erb files
+
+Stand-in: like .era files, but blobs instead of blocks.
 
 `.era` files are special instances of `.e2s` files that follow a more strict content format optimised for reading and long-term storage and distribution.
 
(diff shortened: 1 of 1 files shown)


rolfyone:
I was hoping something like this rather than adding blobs to era would be the plan, because it seems cleaner.

it sounds like the design is to have 
```
mainnet-00005-4ec633af.era
mainnet-00005-4ec633af.erb
```
so the 2 archives are aligned - this would make it nice to validate because you can just parse both files. I think that's sensible (ignore above names, just grabbed one i was testing with :) )

Is it worth including the zk proofs from the blocks so that they're at least partially self referencing? I don't think it'd remove the need to cross-validate against the block completely, but it might allow a level of verification of the blob file itself... I guess it increases complexity in a way because then you have to check they're the actual commitments from the block anyway...

KolbyML:
I am cordinating with @arnetheduck @advaita-saha and the specs for e2store will be moved to `ethereum/consensus-specs` and the specifications for `era` and `erb` should live here https://github.com/eth-clients/e2store-format-specs/tree/main

docs/e2store.md
```diff
@@ -116,6 +116,14 @@ data: snappyFramed(ssz(BeaconState))
 
 The fork and thus the exact format of the `BeaconState` should be derived from the `slot`.
 
+## BlobCompressedSidecars
+```
+type: [0x02, 0x00]
```
michaelsproul, line 121: This seems to be the same `type` as `CompressedBeaconState`? Is this because `erb` files are assumed to never contain `BeaconState`s?

michaelsproul:
We've implemented blob import/export in Lighthouse (https://github.com/sigp/lighthouse/pull/6656), but using plain SSZ rather than `erb`. We're a bit confused by the current spec, but would be interested in adopting `erb` if a spec is clearly defined for it

michaelsproul:
I'm working on a new spec here and welcome feedback/contributions:

- https://github.com/eth-clients/e2store-format-specs/pull/13

tersec:
- https://github.com/eth-clients/e2store-format-specs/pull/19

PR #5883, opened 2024-02-12
Shared validator pubkey

arnetheduck:
This PR allows sharing the pubkey data between validators by using a thread-local cache for pubkey data, netting about a 400mb mem usage reduction on holesky due to us keeping 3 permanent + several ephemeral state copies in memory at all times and each state copy holding a full validator set.

The PR also introduces a hash cache for the key which gives ~14% speedup for a full state `hash_tree_root` - the key makes up for a large part of the `Validator` htr time.

Finally, the time it takes to copy a state goes down as well from ~80m ms to ~60, for reasons similar to htr.

We use a `ptr` even if a `ref` could in theory have been used - there is not much practical benefit to a `ref` (given it's mutable) while a `ptr` is cheaper and easier to copy (when copying temporary states).

We could go further and cache a cooked pubkey but it turns out this is quite intrusive - in all the relevant places, we're already using a cooked key from the immutable validator data so there are no immediate performance gains of doing so while managing the compressed -> cooked key mapping would become more difficult - something for a future PR perhaps.

```sh
 ./ncli --print-times hashTreeRoot capella_state state-8028127-81e89e6c-a37fb323.ssz 
a37fb32350d0d8ef9ae665e28036ec9843dd9c8a9b520acb6b68c8065bec5191
All time are ms
     Average,       StdDev,          Min,          Max,      Samples,         Test
Validation is turned off meaning that no BLS operations are performed
     118.682,        0.000,      118.682,      118.682,            1, Load file
    1000.351,        0.000,     1000.351,     1000.351,            1, Compute
```

```sh
./ncli --print-times hashTreeRoot capella_state state-8028127-81e89e6c-a37fb323.ssz 
a37fb32350d0d8ef9ae665e28036ec9843dd9c8a9b520acb6b68c8065bec5191
All time are ms
     Average,       StdDev,          Min,          Max,      Samples,         Test
Validation is turned off meaning that no BLS operations are performed
     420.763,        0.000,      420.763,      420.763,            1, Load file
     870.934,        0.000,      870.934,      870.934,            1, Compute
```

### DIFF
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -566,7 +566,11 @@ template pubkey*(v: Validator): ValidatorPubKey =
 
 func hash_tree_root*(v: HashedValidatorPubKey): Eth2Digest =
   if isNil(v.value):
-    Eth2Digest() # Safety net
+    # Safety net - have to use a constant because the general hash_tree_root
+    # function is not yet declared at this point
+    const zeroPubkeyHash = Eth2Digest.fromHex(
+      "fa324a462bcb0f10c24c9e17c326a4e0ebad204feced523eccaf346c686f06ee")
+    zeroPubkeyHash
   else:
     v.value[].root
 
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -553,13 +553,22 @@ type
     flags*: set[RewardFlags]
 
 func pubkey*(v: HashedValidatorPubKey): ValidatorPubKey =
-  v.value[].key
+  if isNil(v.value):
+    # This should never happen but we guard against it in case a
+    # default-initialized Validator instance makes it through the other safety
+    # nets
+    ValidatorPubKey()
+  else:
+    v.value[].key
 
 template pubkey*(v: Validator): ValidatorPubKey =
   v.pubkeyData.pubkey
 
 func hash_tree_root*(v: HashedValidatorPubKey): Eth2Digest =
-  v.value[].root
+  if isNil(v.value):
+    Eth2Digest() # Safety net
+  else:
+    v.value[].root
 
 func getImmutableValidatorData*(validator: Validator): ImmutableValidatorData2 =
   let cookedKey = validator.pubkey.loadValid()  # `Validator` has valid key
--- a/beacon_chain/spec/datatypes/base.nim
+++ b/beacon_chain/spec/datatypes/base.nim
@@ -323,9 +323,16 @@ type
     pubkey*: CookedPubKey
     withdrawal_credentials*: Eth2Digest
 
+  HashedValidatorPubKeyItem* = object
+    key*: ValidatorPubKey
+    root*: Eth2Digest
+
+  HashedValidatorPubKey* = object
+    value*: ptr HashedValidatorPubKeyItem
+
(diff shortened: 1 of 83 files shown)


arnetheduck:
draft until post-deneb

tersec:
https://github.com/status-im/nimbus-eth2/pull/5887 might let this build

beacon_chain/spec/datatypes/base.nim
```diff
@@ -323,9 +323,16 @@ type
...
+    key*: ValidatorPubKey
+    root*: Eth2Digest
+
+  HashedValidatorPubKey* = object
+    value*: ptr HashedValidatorPubKeyItem
+
   # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.6/specs/phase0/beacon-chain.md#validator
   Validator* = object
-    pubkey*: ValidatorPubKey
+    pubkeyData*: HashedValidatorPubKey
```
etan-status: how does the library know to serialize this as `pubkey` when e.g. downloading debug/states via JSON?

beacon_chain/spec/datatypes/base.nim
```diff
@@ -545,6 +552,24 @@ type
 
     flags*: set[RewardFlags]
 
+func pubkey*(v: HashedValidatorPubKey): ValidatorPubKey =
+  if isNil(v.value):
+    # This should never happen but we guard against it in case a
+    # default-initialized Validator instance makes it through the other safety
+    # nets
```
etan-status, line 559: wondering if in such a hypothetical scenario, a log + exit wouldn't be more suitable. it starts injecting nasty bugs into surrounding logic

beacon_chain/spec/eth2_merkleization.nim
```diff
@@ -66,3 +68,42 @@ func toDepositContractState*(merkleizer: DepositsMerkleizer): DepositContractSta
...
+          try:
+            # The interface of HashSet is such that we must construct a full
+            # instance to check if it's in the set - then we can return that
+            # instace and discard the one we just created temporarily
+            keys[tmp]
+          except KeyError:
+            raiseAssert "just checked"
+        else:
+          keys.incl tmp
+          tmp
```
etan-status:
```suggestion
        if keys.containsOrIncl(tmp):
          try:
            # The interface of HashSet is such that we must construct a full
            # instance to check if it's in the set - then we can return that
            # instace and discard the one we just created temporarily
            keys[tmp]
          except KeyError:
            raiseAssert "just checked"
        else:
          tmp
```

beacon_chain/spec/datatypes/base.nim
arnetheduck: using a zero-inited pubkey is the current behavior in this scenario (it's used in test cases)

etan-status: shouldn't it then return the correct `hash_tree_root` of the zero validator instead of zero hash?

etan-status (approved):
zero initialized validator comment still applies, but if we can ensure that those are only existing in tests, and don't get triggered when someone actually makes a deposit to a zero validator, seems alright

beacon_chain/spec/datatypes/base.nim
arnetheduck: good catch

PR #5884, opened 2024-02-13
Schedule Deneb on Gnosis

zah:


zah:
Merged here:
https://github.com/status-im/nimbus-eth2/pull/5885

PR #5886, opened 2024-02-13
allow specifying get_proposer_reward block root at state.slot

tersec:
Intended to support consensus block value calculations for `produceBlockV3`

### DIFF
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -12,7 +12,6 @@ import
   json_serialization/std/sets,
   chronicles,
   ../extras,
-  ./datatypes/[phase0, altair, bellatrix],
   "."/[eth2_merkleization, forks, signatures, validator]
 
 from std/algorithm import fill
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/beaconstate.nim
```diff
@@ -12,7 +12,6 @@ import
   json_serialization/std/sets,
   chronicles,
   ../extras,
-  ./datatypes/[phase0, altair, bellatrix],
```
tersec, line 15: `forks` includes these

beacon_chain/spec/beaconstate.nim
```diff
@@ -550,23 +567,41 @@ func get_attestation_participation_flag_indices(
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.6/specs/deneb/beacon-chain.md#modified-get_attestation_participation_flag_indices
 func get_attestation_participation_flag_indices(
     state: deneb.BeaconState,
-    data: AttestationData, inclusion_delay: uint64): set[TimelyFlag] =
+    data: AttestationData, inclusion_delay: uint64,
+    allow_same_state_data_slot: static bool,
```
arnetheduck, line 571: this looks off conceptually - when creating a new block, the attestations in it are guaranteed to be from earlier slots than the block itself - by construction, we should never be getting the attestation flags of the same-slot-as-state..

tersec: My understanding is that the scenario is, it's using the pre-state of the block being created, not the post-state (which would not require this hack). In that case, to compute the block for slot `n+1`, it's using attestations from slot `n` and the (pre-)state from slot `n`. Actually going through the state transition process is somewhat more expensive, and it's possible it's not required here, haven't checked in a while. If it's not, then this is a reasonable optimization

I agree, I'd prefer to have/need none of this PR, but it depends how @cheatfate has organized his calculations, and he can comment more on why he doesn't have a (pre-block-applied) state from slot `n+1` to calculate a block for slot `n+1`. It might be cleaner, and I suspect it'd be less prone to the consensus value being approximate.

arnetheduck: > My understanding is that the scenario is, it's using the pre-state of the block being created, not the post-state (which would not require this hack). 

during block construction we create a slot-transitioned state (without the block applied yet) - this should be the state we use - this state is created pre-emptively before the slot starts (`dag.clearanceState`), in anticipation of the new block either being proposed or arriving over the wire - ideally, that would the state we would use for proposer reward computations

arnetheduck: that state in particular will have the epoch transition done - the epoch transition has had to be performed in order to compute the prop reward for slot-0 in the epoch - I don't see it working correctly otherwise.

cheatfate: Its not me organizing calculation, i have initially mentioned that i'm not a person who should do it. That was @zah idea to use such approach.

tersec: The question is though, what `BeaconState` are you referencing? `get_proposal_reward()` conceptually works fine for this, it's just a question of having a proper state from which to operate.

tersec:
Is there consensus this is a misguided approach (which I'd tend to believe)? If so, it should be closed.

tersec:
Incorporated into https://github.com/status-im/nimbus-eth2/pull/5873

PR #5890, opened 2024-02-14
test macOS with Nim 2.0 in CI and use non-EOL macOS version

tersec:


mratsim:
You might want to join the beta for ARM hosted runners as well: https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions/

tersec:
> You might want to join the beta for ARM hosted runners as well: https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions/

Agree, that looks interesting indeed. ARM is interesting in a couple regards, as an increasingly important target platform in ARMv8/AArch64, and as basically the only remaining viable/useful 32-bit target architecture, moreso than i386. As https://lists.freebsd.org/archives/freebsd-announce/2024-February/000117.html notes:
> Looking forward, general purpose 32-bit platforms are in a state of decline in the marketplace (some more quickly than others), and we have a shrinking pool of developers dedicated to supporting them.  Of our existing 32-bit platforms today (i386, armv[67], powerpc), only armv7 continues to be used in recent system designs.

So to the extent 32-bit portability is practically useful, it's useful because ARMv7, and better to test that directly rather than an i386/i686 proxy.

PR #5891, opened 2024-02-14
rm support for Capella builder API

tersec:
Intended for merge early in release cycle immediately after the mainnet Dencun fork release.

So, not precisely a draft, but should not be merged until shortly after the mainnet Dencun fork readiness release.


tersec:
This appears to trigger usage of `forkchoiceUpdatedV1` in Jenkins minimal testnet logs, which is (correctly) not in `unstable`. Will recreate and decompose into smaller steps when it comes to doing this.

PR #5895, opened 2024-02-16
log elmanager timeouts

arnetheduck:
Also:

* remove some unused metrics
* simplify execution payload fetching flow

### DIFF
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -59,12 +59,6 @@ declareCounter beacon_blocks_proposed,
 declareCounter beacon_block_production_errors,
   "Number of times we failed to produce a block"
 
-declareCounter beacon_block_payload_errors,
-  "Number of times execution client failed to produce block payload"
-
-declareCounter beacon_blobs_sidecar_payload_errors,
-  "Number of times execution client failed to produce blobs sidecar"
-
 # Metrics for tracking external block builder usage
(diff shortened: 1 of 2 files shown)


beacon_chain/validators/beacon_validators.nim
```diff
@@ -369,12 +363,14 @@ from ../spec/beaconstate import get_expected_withdrawals
 
 proc getExecutionPayload(
     PayloadType: type ForkyExecutionPayloadForSigning,
-    node: BeaconNode, proposalState: ref ForkedHashedBeaconState,
-    epoch: Epoch, validator_index: ValidatorIndex): Future[Opt[PayloadType]]
-    {.async: (raises: [CancelledError]).} =
+    node: BeaconNode, head: BlockRef, proposalState: ref ForkedHashedBeaconState,
+    validator_index: ValidatorIndex): Future[Opt[PayloadType]]
+    {.async: (raises: [CancelledError], raw: true).} =
```
tersec, line 368: what does this `raw: true` do/mean?

arnetheduck: skips the `async` transformation so body becomes sync, but still enforces `raises`

arnetheduck: (it's similar to not having `{.async.}` at all, except for the raises)

tersec (approved):

PR #5897, opened 2024-02-16
add metric for database checkpoint duration

etan-status:
Database checkpointing can take seconds, e.g., while Geth is syncing. Add a debug log + metric for it, and also info log if it takes longer than 250ms, same as for the existing `State replayed` log. If the log shows up for a user while the system is not overloaded, it may point to slow disk speed or thermal issue.

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -15,7 +15,7 @@ import
     state_transition, validator],
   ../spec/forks,
   ../spec/datatypes/[phase0, altair, bellatrix, capella],
-  ".."/[beacon_chain_db, era_db],
+  ".."/[beacon_chain_db, beacon_clock, era_db],
   "."/[block_pools_types, block_quarantine]
 
 from ../spec/datatypes/deneb import shortLog
@@ -23,6 +23,8 @@ from ../spec/datatypes/deneb import shortLog
 export
   eth2_merkleization, eth2_ssz_serialization,
   block_pools_types, results, beacon_chain_db
+
+logScope: topics = "chaindag"
 
 # https://github.com/ethereum/beacon-metrics/blob/master/metrics.md#interop-metrics
 declareGauge beacon_head_root, "Root of the head block of the beacon chain"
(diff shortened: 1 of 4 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -49,7 +51,8 @@ declareGauge ticks_delay,
 declareGauge next_action_wait,
   "Seconds until the next attestation will be sent"
 
-logScope: topics = "beacnde"
+declareGauge db_checkpoint_seconds,
```
arnetheduck: a guage has the downside that it gets overwritten quite quickly (which is why we don't use it for state and other durations), meaning that if there is a slow update, we will likely not see it because the next checkpoint after that will be fast and happen almost immediately (because checkpoints are wall-clock-based)

histograms would likely be more appropriate for this situation

etan-status: I think it's important to pick something here that can be plotted in a timeseries. Should I create a `counter` instead that tracks total overall time spent doing checkpoints, so that the slope can be followed? Gauge cannot be put into timeseries?

The histogram, I guess, would just state, that, sometime back, there was a slow checkpoint, but it would not be possible to correlate it against other data.

arnetheduck: counter of time spent could work better - in general though, this is something of a discrepancy between what the code looks like (as if every sample was captured) and what happens (samples are captured at user-defined intervals - ie the default prometheus polling interval is 1 minute - in the test fleet, we use .. 12s *I think* which still is coarse at best and nyquist tells us that we'd need a sample every 6s to capture slot-granularity events.

etan-status: let's try counter then :)

arnetheduck: wanna add one for state replays also then?

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -47,7 +49,7 @@ declareGauge beacon_current_active_validators, "Number of validators in the acti
 declareGauge beacon_pending_deposits, "Number of pending deposits (state.eth1_data.deposit_count - state.eth1_deposit_index)" # On block
 declareGauge beacon_processed_deposits_total, "Number of total deposits included on chain" # On block
 
-logScope: topics = "chaindag"
+declareCounter total_state_replay_seconds, "Total time spent replaying states"
```
arnetheduck, line 52: oof, missed this: https://prometheus.io/docs/practices/naming/ - don't add total, nim-metrics does that already, as a suffix as the docs suggest.

it should also have a `beacon_dag` prefix or something like this since it competes in a global metric namespace

arnetheduck: (so in grafana, this metric is now called `total_state_replay_seconds_total` if you want to plot it, which is not ideal)

etan-status: oops, also missed it! Correcting it

etan-status: - https://github.com/status-im/nimbus-eth2/pull/5917

PR #5900, opened 2024-02-17
handle `Exception` during `EraFile.verify`

etan-status:
`Taskpool.new()` is marked as `{.raises: [Exception].}`. Catch this.

### DIFF
--- a/beacon_chain/era_db.nim
+++ b/beacon_chain/era_db.nim
@@ -1,8 +1,11 @@
+# beacon_chain
 # Copyright (c) 2018-2024 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
 
 import
   std/os,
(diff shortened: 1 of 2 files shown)


beacon_chain/era_db.nim
```diff
@@ -186,15 +189,19 @@ proc verify*(f: EraFile, cfg: RuntimeConfig): Result[Eth2Digest, string] =
   # We'll load the full state and compute its root - then we'll load the blocks
   # and make sure that they match the state and that their signatures check out
   let
-    startSlot = f.stateIdx.startSlot
-    era = startSlot.era
+    slot = f.stateIdx.startSlot
```
etan-status, line 192: renamed because of spurious `Hint`

```
Hint: 'start_slot' should be: 'startSlot' [let declared in /Users/etan/Documents/Repos/nimbus-eth2/beacon_chain/era_db.nim(192, 5)] [Name]
```

It refers to the `start_slot` function in `beacon_clock` for some reason. Renaming the variable to avoid.

beacon_chain/era_db.nim
```diff
@@ -186,15 +189,19 @@ proc verify*(f: EraFile, cfg: RuntimeConfig): Result[Eth2Digest, string] =
...
+    era = slot.era
 
     rng = HmacDrbgContext.new()
-    taskpool = Taskpool.new()
+    taskpool =
+      try:
+        Taskpool.new()
+      except Exception as exc:
+        return err("Failed to initialize TaskPool: " & exc.msg)
```
etan-status:
```suggestion
        return err("Failed to initialize Taskpool: " & exc.msg)
```

PR #5908, opened 2024-02-17
add `{.push raises: [].}` to recently modified files

etan-status:
Status Nim style mandates `{.push raises: []}.` at start of modules. Ensure that's the case so that exceptions are properly tracked.

- https://status-im.github.io/nim-style-guide/errors.exceptions.html
- https://github.com/status-im/nim-eth/pull/614#discussion_r1220906149

### DIFF
--- a/beacon_chain/validators/slashing_protection_v2.nim
+++ b/beacon_chain/validators/slashing_protection_v2.nim
@@ -240,7 +240,6 @@ func version*(_: type SlashingProtectionDB_v2): static int =
 # Internal
 # -------------------------------------------------------------
 
-{.push raises: [].}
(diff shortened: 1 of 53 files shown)


beacon_chain/validators/slashing_protection_v2.nim
```diff
@@ -240,7 +240,6 @@ func version*(_: type SlashingProtectionDB_v2): static int =
 # Internal
 # -------------------------------------------------------------
 
-{.push raises: [].}
```
etan-status, line 243: This one was a duplicate. There's another one at start of file. No {.pop.} either.

PR #5909, opened 2024-02-17
extend linter to enforce exception tracking is turned on

etan-status:
Status Nim style mandates `{.push raises: []}.` at start of modules. Add a CI task to ensure exceptions keep getting properly tracked.

- https://status-im.github.io/nim-style-guide/errors.exceptions.html
- https://github.com/status-im/nim-eth/pull/614#discussion_r1220906149

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,18 +241,37 @@ jobs:
           excluded_extensions="ans|cfg|json|json\\.template|md|png|service|ssz|txt"
 
           current_year=$(date +"%Y")
-          outdated_files=()
+          problematic_files=()
           while read -r file; do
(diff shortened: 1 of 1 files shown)


etan-status:
Needs these two PRs to merge before this one:

- #5904 
- #5905 

Otherwise there will be issues on release because the lint job triggers on merge to `stable`.

tersec:
Are there use cases for which it's not feasible to apply `{.push raises: [].}? In general, some of the tests used not to use it, apparently because of some library exception usage, for example.

etan-status:
Haven't run into a case where it wasn't possible. But, if there's one, an exclusion mechanism can be added, similar to the one for the copyright year lint which does not apply to files that don't support comments.

PR #5917, opened 2024-02-20
rename new timing metrics, as `_total` suffix is implicit

etan-status:
Change `db_checkpoint_seconds` and `state_replay_seconds` metrics to record the latest duration instead of the total. `nim-metrics` already synthesizes a `_total` metric from these implicitly.


etan-status:
Actually, no, the "_total" is for counters, not for counter 😅 revising once more to just follow naming practice

Issue #212, opened 2019-03-27
[Testnet] Fails to download key and load network file


Swader:
```bash
INF 2019-03-27 02:40:06+01:00 New testnet genesis data received. Starting with a fresh database. tid=12761
Failed to download key
cannot open: /home/swader/.cache/nimbus/BeaconNode/testnet0/network.jsonswader@swader-N76VZ:~/repos/nimbus/vendor/nim-beacon-chain$ 
```

This when restarting testnet join without cache etc. rm -rf ~/.cache/nimbus et al.

PR #5923, opened 2024-02-20
bump `nim-json-rpc` to `9c65df305390ab01a13788af5aea78cd62dceb2a`

etan-status:
- Client also handle error message if id is null
- Client pass meaningful error to newFut when processMessage failed
- Refactoring: extract rpc handler from HTTP and WebSocket server
- Use pragma push/pop pair to disable warning
- http server better exception handling
- Fix CI badge url
- Upgrade github actions to v4
- Revert "Fix CI badge url"
- HttpAuthHook use async raises
- Move CancelledError handling to outer try/except of RpcWebsocketServer
- Implement RPC batch call both in servers and clients
- v0.4.0
- Should compile if chronicles log turned on
- Add framework to support more optional types
- v0.4.2
- test refc in CI in Nim 2.0 and later
- use non-EOL macOS version for GitHub Actions CI
- avoid failing uninitialized `Future`
- Improve batch call example and wrapper comments
- Fix ws and socket client error handling and add test to #212
- Add build test with chronicles to json enabled


etan-status:
Smoke tests done:

- deposit sync
- newPayload
- forkchoiceUpdated

etan-status:
on websocket, sometimes timeouts during eth_getLogs while syncing, but recovers

etan-status:
Probably alright as is, good timing right after release so this gets a lot of testing on `unstable`.

PR #5941, opened 2024-02-21
remove MIRACL Core support

etan-status:
EIP-4844 requires BLST via `nim-kzg4844`; MIRACL Core is not supported. Therefore, remove support for the MIRACL Core backend.

- https://github.com/status-im/nim-blscurve/pull/170

### DIFF
--- a/docs/the_auditors_handbook/src/02.5_foreign_lang_to_from_interop.md
+++ b/docs/the_auditors_handbook/src/02.5_foreign_lang_to_from_interop.md
@@ -13,9 +13,9 @@ Example: secp256k1
 
 ### Compiling directly the C files
 
-Example: zlib
+Example: MIRACL Core
 
-[https://github.com/status-im/nim-zlib/blob/master/zlib/zlib_api.nim](https://github.com/status-im/nim-zlib/blob/master/zlib/zlib_api.nim)
+[https://github.com/status-im/nim-blscurve/blob/master/blscurve/miracl/miracl.nim](https://github.com/status-im/nim-blscurve/blob/master/blscurve/miracl/miracl.nim)
 
 ## Wrapping C++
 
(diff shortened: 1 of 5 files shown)


docs/the_auditors_handbook/src/02.5_foreign_lang_to_from_interop.md
```diff
@@ -13,9 +13,9 @@ Example: secp256k1
 
 ### Compiling directly the C files
 
-Example: MIRACL Core
+Example: zlib
```
tersec: https://github.com/status-im/nimbus-eth2/pull/5823

tersec: In general, this is covered by https://status-im.github.io/nim-style-guide/interop.c.html

etan-status: Yeah, the links were not frozen to a permalink. Besides switching to a different lib, the alternatives would be to keep it at MIRACL Core, but refer to a commit that's not a dead link, or to remove the handbook?

tersec: Well, the point of the decision we made before was to stop maintaining the auditor's guide altogether. That should be complete.

Between that and removing the guide, if it's that bad to have some dead links (this will not be the only one), then yes, I'd suggest just removing the auditors handbook entirely.

mratsim: The auditor handbook can just have a warning to say that it's accurate as of July 2020.

It's still useful to point people new to nim to its security features as it's often something people ask.
I suggest linking to a permalink.

tersec (approved):

PR #5637, opened 2023-12-01
Apply EIP-7044 to block signature batch verification

tersec:
Discovered on `devnet-11`, when Deneb exits caused all the Nimbus nodes to stall/fork off on their own sidechain because they rejected a block https://dora.dencun-devnet-11.ethpandaops.io/slot/208919 containing such a voluntary exit:
```
INF 2023-11-29 11:53:52.513+00:00 Block batch signature verification failed  topics="gossip_blocks" signature=8971136c blockRoot=676f87eb blck="(slot: 208919, proposer_index: 1291, parent_root: \"c06e6069\", state_root: \"dd13a832\", eth1data: (deposit_root: d70a234731285c6804c2a4f56711ddb8c82c99740f207854891028af34e27e5e, deposit_count: 0, block_hash: 9a4512b2be6f9e2ff5093635831a8e3bcadda73319a2ea8349233f1f7d2ccbb6), graffiti: \"lighthouse/reth\", proposer_slashings_len: 0, attester_slashings_len: 0, attestations_len: 1, deposits_len: 0, voluntary_exits_len: 1, sync_committee_participants: 350, block_number: 207122, fee_recipient: \"0xf97e180c050e5ab072211ad2c213eb5aee4df134\", bls_to_execution_changes_len: 0, blob_kzg_commitments_len: 3)"
NOT 2023-11-29 11:53:52.514+00:00 Received invalid block                     topics="requman" peer=16U*1cyViS blocks=[676f87eb] peer_score=200
INF 2023-11-29 11:53:52.750+00:00 Block batch signature verification failed  topics="gossip_blocks" signature=8971136c blockRoot=676f87eb blck="(slot: 208919, proposer_index: 1291, parent_root: \"c06e6069\", state_root: \"dd13a832\", eth1data: (deposit_root: d70a234731285c6804c2a4f56711ddb8c82c99740f207854891028af34e27e5e, deposit_count: 0, block_hash: 9a4512b2be6f9e2ff5093635831a8e3bcadda73319a2ea8349233f1f7d2ccbb6), graffiti: \"lighthouse/reth\", proposer_slashings_len: 0, attester_slashings_len: 0, attestations_len: 1, deposits_len: 0, voluntary_exits_len: 1, sync_committee_participants: 350, block_number: 207122, fee_recipient: \"0xf97e180c050e5ab072211ad2c213eb5aee4df134\", bls_to_execution_changes_len: 0, blob_kzg_commitments_len: 3)"
[...]

PR #5953, opened 2024-02-25
fix EIP-7044 implementation when using batch verification

etan-status:
In #5120, EIP-7044 support got added to the state transition function to force `CAPELLA_FORK_VERSION` to be used when validiting `VoluntaryExit` messages, irrespective of their `epoch`.

In #5637, similar logic was added when batch verifying BLS signatures, which is used during gossip validation (libp2p gossipsub, and req/resp). However, that logic did not match the one introduced in #5120, and only uses `CAPELLA_FORK_VERSION` when a `VoluntaryExit`'s `epoch` was set to a value `>= CAPELLA_FORK_EPOCH`. Otherwise, `BELLATRIX_FORK_VERSION` would still be used when validating `VoluntaryExit`, e.g., with `epoch` set to `0`, as is the case in this Holesky block:

- https://holesky.beaconcha.in/slot/1076985#voluntary-exits

Extracting the correct logic from #5120 into a function, and reusing it when verifying BLS signatures fixes this issue, and also leverages the exhaustive EF test suite that covers the (correct) #5120 logic.

This fix only affects networks that have EIP-7044 applied (post-Deneb).

Without the fix, Deneb blocks with a `VoluntaryExit` with `epoch` set to `< CAPELLA_FORK_EPOCH` incorrectly fail to validate despite being valid.

Incorrect blocks that contain a malicious `VoluntaryExit` with `epoch` set to `< CAPELLA_FORK_EPOCH` and signed using `BELLATRIX_FORK_VERSION` _would_ pass the BLS verification stage, but subsequently fail the state transition logic. Such blocks would still correctly be labeled invalid.

### DIFF
--- a/beacon_chain/spec/signatures.nim
+++ b/beacon_chain/spec/signatures.nim
@@ -216,6 +216,20 @@ func compute_voluntary_exit_signing_root*(
       fork, DOMAIN_VOLUNTARY_EXIT, epoch, genesis_validators_root)
   compute_signing_root(voluntary_exit, domain)
 
+func voluntary_exit_signature_fork*(
+    consensusFork: static ConsensusFork,
+    state_fork: Fork,
+    capella_fork_version: Version): Fork =
+  when consensusFork >= ConsensusFork.Deneb:
+    # Always use Capella fork version, disregarding `VoluntaryExit` epoch
+    # [Modified in Deneb:EIP7044]
+    Fork(
+      previous_version: capella_fork_version,
+      current_version: capella_fork_version,
+      epoch: GENESIS_EPOCH)
+  else:
+    state_fork
+
 func get_voluntary_exit_signature*(
     fork: Fork, genesis_validators_root: Eth2Digest,
     voluntary_exit: VoluntaryExit,
--- a/beacon_chain/spec/signatures.nim
+++ b/beacon_chain/spec/signatures.nim
@@ -226,7 +226,7 @@ func voluntary_exit_signature_fork*(
     Fork(
       previous_version: capella_fork_version,
       current_version: capella_fork_version,
-      epoch: GENESIS_EPOCH)
+      epoch: GENESIS_EPOCH)  # irrelevant when current/previous identical
   else:
     state_fork
 
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -346,19 +346,6 @@ proc process_deposit*(cfg: RuntimeConfig,
 
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.6/specs/phase0/beacon-chain.md#voluntary-exits
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/deneb/beacon-chain.md#modified-process_voluntary_exit
-func voluntary_exit_signature_fork*(
-    cfg: RuntimeConfig,
-    state: ForkyBeaconState): Fork =
-  when typeof(state).kind >= ConsensusFork.Deneb:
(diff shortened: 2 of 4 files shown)


etan-status:
Tested on Holesky, the block that previously failed to apply now applies fine, and can sync past it.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -346,6 +346,19 @@ proc process_deposit*(cfg: RuntimeConfig,
 
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.6/specs/phase0/beacon-chain.md#voluntary-exits
 # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.1/specs/deneb/beacon-chain.md#modified-process_voluntary_exit
+func voluntary_exit_signature_fork*(
+    cfg: RuntimeConfig,
+    state: ForkyBeaconState): Fork =
```
arnetheduck: can't this be a function transforming a given "current" fork into a capella fork? ie
```nim
if fork.epoch >= cfg.CAPELLA_FORK_EPOCH:
  Fork(...)
else:
  fork
```

then it can live in signatures.nim and not take a (forky) state

arnetheduck (approved):

beacon_chain/spec/state_transition_block.nim
etan-status: Not sure if that is generally applicable, e.g., EF Unit Test suite has a `@with_matching_spec_config` decorator because by default it creates states with types that do not fit the config's fork schedule.

Also, it introduces an assumption that the `state.fork` determines `typeof(state).kind` which may introduce subtle bugs if that is ever not the case. that assumption could be tested at call site.

It would probably be safe enough with an additional `is_post_deneb: static bool`, then the assert can be made inside `voluntary_exit_signature_fork`.

Extracting `state.fork` would still require the `withState` in `block_clearance.nim`.

Are you worried about performance here? Then, `func` could become `template`, if necessary.

If the concern is about extra code size from the generic instantiations, that would have also applied before, where the check was inlined, but also replicated across each `ForkyBeaconState`.

For now, seems a bit risky to yolo, given the timing. Wouldn't oppose a subsequent cleanup though.

(and, `>= cfg.DENEB_FORK_EPOCH`).

arnetheduck: this was more about collecting signature-level logic in signatures, rather than this being part of block processing - has got nothing to do with performance

etan-status:
Tested once more that the revised logic also still syncs fine on Holesky.

beacon_chain/spec/state_transition_block.nim
etan-status: OK, moved it down to `signatures.nim`, but kept the aspect that determines `consensusFork` solely from the type instead of using some runtime method.

beacon_chain/spec/signatures.nim
```diff
@@ -216,6 +216,20 @@ func compute_voluntary_exit_signing_root*(
...
+    consensusFork: static ConsensusFork,
+    state_fork: Fork,
+    capella_fork_version: Version): Fork =
+  when consensusFork >= ConsensusFork.Deneb:
+    # Always use Capella fork version, disregarding `VoluntaryExit` epoch
+    # [Modified in Deneb:EIP7044]
+    Fork(
+      previous_version: capella_fork_version,
+      current_version: capella_fork_version,
+      epoch: GENESIS_EPOCH)
```
arnetheduck: genesis?

tersec: irrelevant when `previous_version == current_version`, could be anything. used by `get_domain` to pick which to use.

PR #5956, opened 2024-02-25
in VC, extract all fork versions/epochs instead of just altair epoch

etan-status:
In the past, VC needed to know `ALTAIR_FORK_EPOCH` to determine whether sync committee duties need to be performed. Now, with EIP-7044, it additionally needs to know `CAPELLA_FORK_VERSION` and `DENEB_FORK_EPOCH` to create `VoluntaryExit` signatures post-Deneb.

These specific constants need to be accessed by name. They cannot be derived from the `/eth/v1/config/fork_schedule` endpoint, because that endpoint only provides the information needed for regular signatures. Notably, it does not state what `Fork` belongs to which `ConsensusFork`, and also does not indicate whether the underlying schedule is canonical, or whether it may be from a completely unsupported network, e.g., devnet for Verkle which currently uses a Verkle fork on top of Capella instead of Deneb. So, a heuristic which simply assumes that the fifth fork is Deneb is not reliable.

As it seems to be recurring practice that VC needs to access fork config related config.yaml keys, extend `ValidatorRuntimeConfig` to extract all `ConsensusFork` specific constants. Further validate that the retrieved info is consistent (monotonically increasing `_FORK_EPOCH`) and avoid extracting information about forks that are still subject to change (scheduled at `FAR_FUTURE_EPOCH`).

Note that this is a pre-requisite for implementing EIP-7044 into keymanager. Submitted separately due to the size of this patch.

### DIFF
--- a/beacon_chain/deposits.nim
+++ b/beacon_chain/deposits.nim
@@ -220,33 +220,20 @@ proc restValidatorExit(config: BeaconNodeConf) {.async.} =
            reason = exc.msg
     quit 1
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.6/specs/phase0/beacon-chain.md#voluntary-exits
+  # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.0/specs/deneb/beacon-chain.md#modified-process_voluntary_exit
   let signingFork = try:
     let response = await client.getSpecVC()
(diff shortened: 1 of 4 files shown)


etan-status:
Preferrable to only extract the `CAPELLA_FORK_VERSION` and `DENEB_FORK_EPOCH`.

PR #5959, opened 2024-02-25
add EIP-7044 support to keymanager API

etan-status:
When trying to sign `VoluntaryExit` via keymanager API, the logic is not yet aware of EIP-7044 (part of Deneb). This patch adds missing EIP-7044 support to the keymanager API as well.

As part of this, the VC needs to become aware about:

- `CAPELLA_FORK_VERSION`: To correctly form the EIP-7044 signing domain. The fork schedule does not indicate which of the results, if any, corresponds to Capella.
- `CAPELLA_FORK_EPOCH`: To detect whether Capella was scheduled. If a BN does not have it in its config while other BNs have it, this leads to a log if Capella has not activated yet, or marks the BN as incompatible if Capella already activated.
- `DENEB_FORK_EPOCH`: To check whether EIP-7044 logic should be used.

Related PRs:

- #5120 added support for processing EIP-7044 `VoluntaryExit` messages as part of the state transition functions (tested by EF spec tests).
- #5953 synced the support from #5120 to gossip validation.
- #5954 added support to the `nimbus_beacon_node deposits exit` command.
- #5956 contains an alternative generic version of `VCForkConfig`.

### DIFF
--- a/beacon_chain/nimbus_validator_client.nim
+++ b/beacon_chain/nimbus_validator_client.nim
@@ -348,6 +348,18 @@ proc asyncInit(vc: ValidatorClientRef): Future[ValidatorClientRef] {.async.} =
   let
     keymanagerInitResult = initKeymanagerServer(vc.config, nil)
 
+  func getCapellaForkVersion(): Opt[Version] =
+    if vc.runtimeConfig.forkConfig.isSome():
+      vc.runtimeConfig.forkConfig.get().CAPELLA_FORK_VERSION
+    else:
+      Opt.none(Version)
+
+  func getDenebForkEpoch(): Opt[Epoch] =
+    if vc.runtimeConfig.forkConfig.isSome():
+      Opt.some(vc.runtimeConfig.forkConfig.get().DENEB_FORK_EPOCH)
+    else:
+      Opt.none(Epoch)
+
   proc getForkForEpoch(epoch: Epoch): Opt[Fork] =
     if len(vc.forks) > 0:
       Opt.some(vc.forkAtEpoch(epoch))
@@ -379,6 +391,8 @@ proc asyncInit(vc: ValidatorClientRef): Future[ValidatorClientRef] {.async.} =
         Opt.none(string),
         nil,
         vc.beaconClock.getBeaconTimeFn,
+        getCapellaForkVersion,
+        getDenebForkEpoch,
         getForkForEpoch,
         getGenesisRoot
         )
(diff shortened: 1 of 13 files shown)


beacon_chain/nimbus_validator_client.nim
```diff
@@ -348,6 +348,18 @@ proc asyncInit(vc: ValidatorClientRef): Future[ValidatorClientRef] {.async.} =
   let
     keymanagerInitResult = initKeymanagerServer(vc.config, nil)
 
+  func getCapellaForkVersion(): Opt[Version] =
+    if vc.runtimeConfig.forkConfig.isSome():
+      vc.runtimeConfig.forkConfig.get().CAPELLA_FORK_VERSION
```
cheatfate: I suppose here is mistype.

beacon_chain/rpc/rest_constants.nim
```diff
@@ -241,6 +241,10 @@ const
     "The given Merkle proof is invalid"
   InvalidMerkleProofIndexError* =
     "The given Merkle proof index is invalid"
+  FailedToObtainForkVersionError* =
+    "Failed to obtain fork version"
+  FailedToObtainConsensusForkError* =
+    "Failed to obtain ConsensusFork information"
```
cheatfate: Is `ConsensusFork` is widely used term and will be well understood?

tersec: Well, "consensus fork" or "beacon chain fork" probably are. If in doubt, can look at EF documentation to see what they use. One issue is that their public communication  refers to the joint forks (Shapella, Dencun) rather than Capella/Deneb. The EL forks aren't really relevant for this PR's purposes. But the idea that there are beacon chain forks, whatever their name, is widely known.

etan-status: I think it's clear enough with 'consensus fork'. Have used @tersec 's spelling.

etan-status:
```suggestion
    "Failed to obtain consensus fork information"
```

beacon_chain/nimbus_validator_client.nim
etan-status: you mean, because it lacks the `Opt` wrapper?

inside `forkConfig`, `CAPELLA_FORK_VERSION` itself is another `Opt`. 

- `vc.runtimeConfig.forkConfig.isSome` --> We have downloaded a valid config from a BN
- `vc.runtimeConfig.forkConfig.get().CAPELLA_FORK_VERSION.get()` --> We have a BN that has Capella information in its config.

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -367,8 +369,6 @@ type
 
   ProduceBlockResponseV3* = ForkedMaybeBlindedBeaconBlock
 
-  VCRuntimeConfig* = Table[string, string]
```
cheatfate: `VCRuntimeConfig` is exactly on place.

etan-status: moved back

arnetheduck (approved):

PR #5988, opened 2024-02-27
add trivial `{.push raises: [].}` to `research` and `test` folders

etan-status:
Annotate the `research` and `test` files for which no further changes are needed to successfully compile them, to not interfere with periodic tasks such as spec reference bumps.

### DIFF
--- a/tests/test_keymanager_api.nim
+++ b/tests/test_keymanager_api.nim
@@ -5,7 +5,6 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
-{.push raises: [].}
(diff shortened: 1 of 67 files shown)


etan-status:
- https://github.com/status-im/nimbus-eth2/pull/5991

tests/test_keymanager_api.nim
```diff
@@ -5,6 +5,7 @@
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
+{.push raises: [].}
```
etan-status:
```suggestion
```

This file is cursed. It invents random fake raises, and is also the one triggering big bug source in proposeBlockAux. Investigating it separately from this PR.

PR #5989, opened 2024-02-28
assert exceptions that statically cannot occur in test code

etan-status:
There are some spurious exceptions that statically cannot occur. Add `raiseAssert` for these.

### DIFF
--- a/research/stack_sizes.nim
+++ b/research/stack_sizes.nim
@@ -15,6 +15,7 @@ proc print(t: auto, n: string, indent: int) =
   try:
     echo fmt"{sizeof(t):>8}  {spaces(indent)}{n}: {typeof(t).name}"
   except ValueError:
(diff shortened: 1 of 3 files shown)


research/stack_sizes.nim
```diff
@@ -5,12 +5,17 @@
...
+
 import ../beacon_chain/spec/datatypes/phase0
 
 import std/[typetraits, strformat, strutils]
 
 proc print(t: auto, n: string, indent: int) =
-  echo fmt"{sizeof(t):>8}  {spaces(indent)}{n}: {typeof(t).name}"
+  try:
+    echo fmt"{sizeof(t):>8}  {spaces(indent)}{n}: {typeof(t).name}"
+  except ValueError:
```
arnetheduck, line 17: https://github.com/nim-lang/Nim/pull/23356 - this could be backported via stew

PR #5992, opened 2024-02-28
use `closureScope` with `{.raises.}` in `test_peer_pool`

etan-status:
`test_peer_pool` is a bit different from the other tests as it uses `closureScope` which doesn't play well with `{.push raises: [].}`. Define an overload instead that allows passing `{.raises.}` to the `template`. This allows using `unittest2`'s exception handler without having to refactor the test.

### DIFF
--- a/tests/test_peer_pool.nim
+++ b/tests/test_peer_pool.nim
@@ -1,16 +1,20 @@
 # beacon_chain
-# Copyright (c) 2019-2021 Status Research & Development GmbH
+# Copyright (c) 2019-2024 Status Research & Development GmbH
 # Licensed and distributed under either of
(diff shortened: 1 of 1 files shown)


tests/test_peer_pool.nim
```diff
@@ -237,7 +241,7 @@ suite "PeerPool testing suite":
       itemFut23.finished == false
       itemFut24.finished == false
 
-  test "Acquire/Sorting and consistency test": closureScope:
+  test "Acquire/Sorting and consistency test": closureScope([CatchableError]):
```
tersec, line 244: It  might be worth figuring out why `closureScope` is used at all here

PR #5993, opened 2024-02-28
handle `scanf` exceptions in `test_fixture_ssz_generic_types`

etan-status:
`scanf` apparently has both a `bool` return as well as raising random exceptions depending on what functions get called by the `macro`. To make this explicit, catch the `ValueError` from the generated `parseInt` call, to separate `scanf` behaviour from the actual SSZ test logic. In the end, it mostly doesn't matter as there are some `doAssert wasMatched` on the next line (not everywhere though). But it still makes the `scanf` internals explicit, so is clearer.

### DIFF
--- a/tests/consensus_spec/test_fixture_ssz_generic_types.nim
+++ b/tests/consensus_spec/test_fixture_ssz_generic_types.nim
@@ -192,11 +192,12 @@ proc checkBitList(
     expectedHash: SSZHashTreeRoot
 ) {.raises: [IOError, SerializationError, UnconsumedInput, ValueError].} =
   var maxLen: int
-  discard (
+  let wasMatched =
     try:
       scanf(sszSubType, "bitlist_$i", maxLen)
     except ValueError:
-      false)  # Raised if the parsed integer is out of the valid range
+      false  # Raised if the parsed integer is out of the valid range
+  doAssert wasMatched
   case maxLen
   of 0: checkBasic(BitList[0], dir, expectedHash)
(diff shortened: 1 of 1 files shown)


etan-status:
Also ensures that someone copy pasting our `scanf` for some other purpose gets the error handling right. Not entirely pointless, given AIs may also use it ;-)

etan-status:
From `scanf` internal logic, it seems that this is intentional, although not very well documented in the outer macro.

- `return 0` means there was no match
- `ValueError` means there was a match but it was out of range

tests/consensus_spec/test_fixture_ssz_generic_types.nim
```diff
@@ -169,9 +187,17 @@ proc checkBitVector(sszSubType, dir: string, expectedHash: SSZHashTreeRoot) =
   else:
     raise newException(TestSizeError, "Unsupported BitVector of size " & $size)
 
-proc checkBitList(sszSubType, dir: string, expectedHash: SSZHashTreeRoot) =
+proc checkBitList(
+    sszSubType, dir: string,
+    expectedHash: SSZHashTreeRoot
+) {.raises: [IOError, SerializationError, UnconsumedInput, ValueError].} =
   var maxLen: int
-  discard scanf(sszSubType, "bitlist_$i", maxLen)
```
etan-status, line 174: There was an unused `wasMatched` variable in the past. #2318 then proceeded to `discard` the result, silencing the warning.
However, the "unused" variable is necessary, but the `doAssert wasMatched` was missing!
Fixing now.

PR #5994, opened 2024-02-28
rm REST blinded forked Capella block support

tersec:


beacon_chain/spec/forks.nim
```diff
@@ -452,29 +452,23 @@ template ExecutionPayloadForSigning*(kind: static ConsensusFork): auto =
 template BlindedBeaconBlock*(kind: static ConsensusFork): auto =
   when kind == ConsensusFork.Deneb:
     typedesc[deneb_mev.BlindedBeaconBlock]
-  elif kind == ConsensusFork.Capella:
-    typedesc[capella_mev.BlindedBeaconBlock]
-  elif kind == ConsensusFork.Bellatrix:
+  elif kind == ConsensusFork.Capella or kind == ConsensusFork.Bellatrix:
```
etan-status, line 455:
```suggestion
  elif kind >= ConsensusFork.Bellatrix:
```

etan-status: ah, no, then it breaks for electra

etan-status (approved):

PR #5998, opened 2024-02-28
allow `--external-beacon-api-url` to fallback to genesis if viable

etan-status:
When using `--external-beacon-api-url`, one has to accompany it with either `--trusted-block-root` or `--trusted-state-root`. If neither is specified, we can fallback to a deeply finalized noncontroversial block root. For networks that started post Altair, e.g., Holesky, the genesis block root fulfills that requirement, as in, it is implicitly trusted. Therefore, if only `--external-beacon-api-url` is provided without any `--trusted-block-root` or `--trusted-state-root`, use genesis block root if it is a viable starting point (post-Altair).

```
build/nimbus_beacon_node \
    --network=holesky \
    --data-dir="$HOME/Downloads/nimbus/data/holesky" \
    "--external-beacon-api-url=http://unstable.holesky.beacon-api.nimbus.team" \
    --tcp-port=9010 --udp-port=9010 \
    --rest --log-level=DEBUG \
    --no-el
```

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -54,6 +54,43 @@ declareGauge next_action_wait,
 declareCounter db_checkpoint_seconds,
   "Time spent checkpointing the database to clear the WAL file"
 
+proc fetchGenesisState(
+    config: BeaconNodeConf,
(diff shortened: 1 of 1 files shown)


etan-status:
image

PR #6001, opened 2024-02-29
block pull requests to `stable`

etan-status:
Pull requests should be opened against the `unstable` branch. See https://nimbus.guide/contribute.html#build-and-deploy


etan-status:
Most recent instance: #6000

arnetheduck:
meh, or we can just retarget them as was done with 6000 instead of being obnoxious about it

etan-status:
The CI is not re-run on base change / it may show as green on stable, but actually fail after merge.
Failing allows quickly re-submitting directly. If it's a meaningful PR, it also may need resolving merge conflicts.

We can keep it pending until the next one that doesn't cleanly retarget :-) They happen from time to time.

tersec:
Yeah, it's always a gamble with retargeting, because it's either:

a) ask them to do it on their own, which is the human version of "obnoxious" here;
b) just try to target, and hope it goes okay without further intevention; because
c) failed version of (b) is bad, where tried to "just retarget them".

That one was a documentation fix so it was low-risk but typically for anything vaguely nontrivial I ask the person filing the PR to do it because of the risk of (c).

In favor of this.

etan-status:
- #6068

tersec:
- https://github.com/status-im/nimbus-eth2/pull/6103

tersec:
One edge case here is Dependabot, which uses `stable` and might not be able to use any other branch except if it were the default (i.e. navigating to `https://github.com/status-im/nimbus-eth2/` navigated to that branch instead): https://github.com/status-im/nimbus-eth2/pull/6111

These are 100% of the time in the Nimbus codebase trivial changes, because thye're in Python dependencies, and just version bumps in `requirements.txt` and typically are rebased by GitHub automatically and easily.

etan-status:
There is an option to specify the target branch for Dependabot, but it is documented to be ignored for security patches: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch

Therefore, have updated the action to not trigger on Dependabot triggered PRs.

tersec:
https://github.com/status-im/nimbus-eth2/pull/6219

tersec:
https://github.com/status-im/nimbus-eth2/pull/6253

arnetheduck (approved):

tersec (approved):

PR #6010, opened 2024-03-01
use debug- prefix for Yamux experimental support

tersec:
Discussed in https://github.com/status-im/nimbus-eth2/pull/5080#issuecomment-1592988315

per https://nimbus.guide/options.html#available-options
> Any `debug`-prefixed flags are considered ephemeral and subject to removal without notice.

It's already not documented because it's hidden; this just makes the existing policy explicit.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -335,7 +335,7 @@ type
         hidden
         desc: "Enable the Yamux multiplexer"
         defaultValue: false
-        name: "enable-yamux" .}: bool
+        name: "debug-enable-yamux" .}: bool
 
       weakSubjectivityCheckpoint* {.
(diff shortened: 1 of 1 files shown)


zah:
Looks like the old name is in use here:
https://github.com/status-im/infra-nimbus/blob/aaa3c437a9bc23e5be854940990ce10d91f2bef9/ansible/group_vars/nimbus.mainnet.yml#L171

@jakubgs, would you coordinate this change?

jakubgs:
Should I make the change before this is merged?

tersec:
> Should I make the change before this is merged?

If it's only the `libp2p` nodes without any validators (because mainnet test fleet) anyway, yeah. Make the change there, I'll merge, and within 24 hours, things will be synced again.

PR #6022, opened 2024-03-04
add proposer reward accounting to block transitions

tersec:
For
- https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.5.0#/Validator/produceBlockV3
- https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.5.0#/Rewards/getBlockRewards

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -968,40 +968,41 @@ proc advanceSlots*(
 proc applyBlock(
     dag: ChainDAGRef, state: var ForkedHashedBeaconState, bid: BlockId,
     cache: var StateCache, info: var ForkedEpochInfo): Result[void, cstring] =
-
   loadStateCache(dag, cache, bid, getStateField(state, slot).epoch)
 
-  case dag.cfg.consensusForkAtEpoch(bid.slot.epoch)
+  discard case dag.cfg.consensusForkAtEpoch(bid.slot.epoch)
   of ConsensusFork.Phase0:
     let data = getBlock(dag, bid, phase0.TrustedSignedBeaconBlock).valueOr:
       return err("Block load failed")
-    state_transition(
+    ? state_transition(
       dag.cfg, state, data, cache, info,
       dag.updateFlags + {slotProcessed}, noRollback)
   of ConsensusFork.Altair:
     let data = getBlock(dag, bid, altair.TrustedSignedBeaconBlock).valueOr:
       return err("Block load failed")
-    state_transition(
+    ? state_transition(
       dag.cfg, state, data, cache, info,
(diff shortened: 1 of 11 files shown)


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -968,41 +968,42 @@ proc advanceSlots*(
 proc applyBlock(
     dag: ChainDAGRef, state: var ForkedHashedBeaconState, bid: BlockId,
     cache: var StateCache, info: var ForkedEpochInfo): Result[void, cstring] =
-
   loadStateCache(dag, cache, bid, getStateField(state, slot).epoch)
 
-  case dag.cfg.consensusForkAtEpoch(bid.slot.epoch)
+  discard case dag.cfg.consensusForkAtEpoch(bid.slot.epoch)
```
etan-status, line 973: can make the callee `{.discardable.}` to allow implicit discards. as the result here is just some metadata and not an actual result that needs checking in regular operation, maybe cleaner.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -186,11 +186,9 @@ proc check_proposer_slashing*(
 proc process_proposer_slashing*(
     cfg: RuntimeConfig, state: var ForkyBeaconState,
     proposer_slashing: SomeProposerSlashing, flags: UpdateFlags,
-    cache: var StateCache):
-    Result[void, cstring] =
+    cache: var StateCache): Result[Gwei, cstring] =
   let proposer_index = ? check_proposer_slashing(state, proposer_slashing, flags)
-  ? slash_validator(cfg, state, proposer_index, cache)
-  ok()
+  ok(? slash_validator(cfg, state, proposer_index, cache))
```
etan-status:
```suggestion
  slash_validator(cfg, state, proposer_index, cache)
```

beacon_chain/consensus_object_pools/blockchain_dag.nim
etan-status: (just on the outermost layer, obviously. the inner layers need to forward between them)

tersec: I think that'd apply to the `Result[..., cstring]` part too, which is important, even if the `Gwei` part is not. `applyBlock` hadn't been discarding it, it had been forwarding the results:
```nim
proc applyBlock(
    dag: ChainDAGRef, state: var ForkedHashedBeaconState, bid: BlockId,
    cache: var StateCache, info: var ForkedEpochInfo): Result[void, cstring] =
```

just as the various fork operations tests are effectively converting from `Result[Gwei, cstring]` to `Result[void, cstring]` for consistency.

So making it discardable wouldn't help here, I think.

PR #6028, opened 2024-03-04
metrics for proposer and sync duties

### LINKED ISSUE #6025: Provide a way to detect next proposal time
**Is your feature request related to a problem? Please describe.**
Currently there is a metric called `next_action_wait` which allows to avoid missing an attestation on a node with low number of validators. This is neat and all, but for notes with large numbers of validators that metric isn't really that useful, because you can hardly find more than 30 second between attestations on a node with ~30 validators, which means you might as well just restart without consulting the metric.

What would be great is a `next_proposal_wait` metric that would allow us to avoid missing proposals, which are far more important.
There's very little harm in missing a few attestations compared to missing a proposal.

**Describe alternatives you've considered**
Alternatively an API endpoint would also be useful. A log message is nice too, but a metric is easier to use in scripts.

**Additional context**
We currently use `next_action_wait` metric a restart script in our Ansible roles:
https://github.com/status-im/infra-role-beacon-node-linux/blob/master/templates/scripts/restart.sh.j2

But it would be great to add support to it for also checking for future proposals.

etan-status:
To avoid restarting with scheduled proposals or sync duties, which are rare but have high rewards, add metrics to simpify tracking them.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -52,7 +52,7 @@ declareGauge next_action_wait,
   "Seconds until the next attestation will be sent"
 
 declareGauge next_proposal_wait,
-  "Seconds until the next proposal will be sent, or -1 if no proposal scheduled"
+  "Seconds until the next proposal will be sent, or Inf if not known"
 
 declareGauge sync_committee_active,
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Thank you.

PR #6069, opened 2024-03-12
chore(vendor): update nim-libp2p path

kaiserd:
The nim-libp2p repo has been moved from the status github org to the vacp2p org.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,7 +22,7 @@
 	branch = master
 [submodule "vendor/nim-libp2p"]
 	path = vendor/nim-libp2p
-	url = https://github.com/status-im/nim-libp2p.git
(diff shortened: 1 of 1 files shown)


etan-status (approved):
the old link still works fine (as a redirect), so this should not make it annoying to re-build old (and even the current) nimbus releases.

etan-status:
Why is there a `Signed-off-by` in the commit message? There is no contributor license agreement, Nimbus is permissively licensed.

kaiserd:
@etan-status Sorry for the signed-off. That was a mistake. I used `-s` instead of `-S`.

etan-status:
Ah, was just curious. No worries :)

PR #6087, opened 2024-03-15
bump `nim-web3` to `428c46c94f58cf19d052b25cd3ae6f4c8c722e56`

etan-status:
- Add writeValue for BlockNumber

PR #6090, opened 2024-03-16
make `Gwei` `distinct`

etan-status:
#6087 introduced a subtle change to `nim-web3` resulting in `Gwei` to be serialized differently than before. Using a `distinct` type for `Gwei` improves type safety and avoids such problems in the future.


etan-status:
- https://github.com/status-im/nimbus-eth2/pull/6089

etan-status:
- https://github.com/status-im/nimbus-eth2/pull/6094

etan-status:
- https://github.com/status-im/nim-ssz-serialization/pull/79

PR #6098, opened 2024-03-19
extend seen ttl to cover 2 epochs

arnetheduck:
this allows us to drop these useless messages earlier in the pipeline

https://github.com/ethereum/consensus-specs/pull/3627


arnetheduck:
This PR increases seen cache memory usage - https://github.com/vacp2p/nim-libp2p/pull/1073 claws it back

PR #6108, opened 2024-03-20
make sure `clearanceState` builds on top of `headState` in chain stall

etan-status:
The `clearanceState` points to the latest resolved block, regardless of whether that block is canonical according to fork choice. If chain is stalled and we want to prepare for resuming validator duties, we need a recent state according to fork choice to avoid lag spikes and missing slot timings.


etan-status:
It only makes sense to do this in degraded mode. Otherwise, we may interfere with an active import process of a side branch that only becomes canonical later. In degraded mode, we know that no active import is ongoing, so the `clearanceState` is unused and can be borrowed for catching up on validator duties.

PR #6112, opened 2024-03-20
when adding duplicates to quarantine, schedule deepest missing parent

etan-status:
During sync, sometimes the same block gets encountered and added to quarantine multiple times. If its parent is already known, quarantine incorrectly registers it as missing, leading to re-download. This can be fixed by registering the parent's deepest missing parent recursively.

Also increase the stickiness of `missing`. We only perform 4 attempts within ~16 seconds before giving up. Very frequently, this is not enough and there is no progress until sync manager kicks in even on holesky.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -103,9 +103,10 @@ OK: 5/5 Fail: 0/5 Skip: 0/5
 OK: 1/1 Fail: 0/1 Skip: 0/1
 ## Block quarantine
 ```diff
++ Recursive missing parent                                                                   OK
 + Unviable smoke test                                                                        OK
 ```
-OK: 1/1 Fail: 0/1 Skip: 0/1
+OK: 2/2 Fail: 0/2 Skip: 0/2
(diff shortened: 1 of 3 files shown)


arnetheduck:
> This can be fixed by registering the parent's deepest missing parent recursively.

why is this needed though? that ancestor should already be registered in this case?

etan-status:
Entries from the `missing` list expire if not found in a handful of attempts.

arnetheduck:
> Entries from the missing list expire if not found in a handful of attempts.

this sounds like a broader issue then - ie in the quarantine, we have limited space (in part because it lives in-memory) - it seems to me then that when running low on space, we should strive to keep "early" entries and discard entries that build on top of them first so that we don't inadvertedly remove an ancestor over a descendant from the list of "interesting" blocks -> the "retry" count of an ancestor should be the sum of all its known descendants and used to prioritise requests (since resolving such a quarantine item would have the effect of unblocking a large number of blocks)

PR #6114, opened 2024-03-20
turn on `request_manager` while chain is stalled

etan-status:
While chain is stalled, both sync manager and request manager could be useful to discover additional branches. `request_manager` should only be inhibited while syncing on a regularly operating chain.


etan-status:
It would be better to optimize rewind logic instead in sync manager, to use a single `byRoot` request on the parent when rewinding spans larger epoch ranges. This way, the request manager is not needed while sync manager is running.

PR #6120, opened 2024-03-21
make blobs use less quota when other nodes sync from us

etan-status:
Each individual blob currently uses as much quota from the network limit as an entire block does, 128 items per second shared across all peers. Blobs are 128 KB each instead of up to several MB and are simpler to encode. There can be multiple per block (6 currently), so allow 2000 blobs per second across all peers. That decreases the cost per block from `3125 + 3125 * blobs.len` quota (= `[3125, 21875]`) to a lower `3125 + 200 * blobs.len` quota (= `[3125, 4325]`), accounting for the slight increase in data transfer and encoding time.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -479,18 +479,19 @@ template awaitQuota*(peerParam: Peer, costParam: float, protocolIdParam: string)
 
   if not peer.quota.tryConsume(cost.int):
     let protocolId = protocolIdParam
-    debug "Awaiting peer quota", peer, cost, protocolId
+    debug "Awaiting peer quota", peer, cost = cost, protocolId = protocolId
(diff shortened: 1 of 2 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -479,18 +479,19 @@ template awaitQuota*(peerParam: Peer, costParam: float, protocolIdParam: string)
...
+template awaitQuota*(
+    networkParam: Eth2Node, costParam: float, protocolIdParam: string) =
   let
     network = networkParam
     cost = int(costParam)
 
   if not network.quota.tryConsume(cost.int):
     let protocolId = protocolIdParam
-    debug "Awaiting network quota", peer, cost, protocolId
+    debug "Awaiting network quota", peer, cost = cost, protocolId = protocolId
```
etan-status, line 494: these otherwise have nasty `gensym4177` visible, maybe because of template shenanigans.

PR #6121, opened 2024-03-22
only request blobs if a sync response actually provided blocks

etan-status:
During sync, we can skip the `blobSidecarsByRange` request when there are no blocks with `kzg_commitments` in the blocks data. Avoids running into throttling from peers during long periods of non-finality.


etan-status:
Ideally, would download the blobs from a separate peer in parallel that also claims to have that history available (based on status head root). But as long as it's sequential, we can know whether it's pointless to ask for blobs.

PR #6122, opened 2024-03-22
enqueue missing parent block if stored in local DB

etan-status:
When checking for `MissingParent`, it may be that the parent block was already discovered as part of a prior run. In that case, it can be loaded from storage and processed without having to rediscover the entire branch from the network. This is similar to #6112 but for blocks that are discovered via gossip / sync mgr instead of via request mgr.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -474,6 +474,16 @@ proc storeBlock(
 
   if parent.isErr():
     if parent.error() == VerifierError.MissingParent:
+      # This indicates that no `BlockRef` is available for the `parent_root`.
+      # However, the block may still be available in local storage. On startup,
+      # only the canonical branch is imported into `blockchain_dag`, while
+      # non-canonical branches are re-discovered with sync/request managers.
+      # Data from non-canonical branches that has already been verified during
+      # a previous run of the beacon node is already stored in the database but
+      # only lacks a `BlockRef`. Loading the branch from the database saves a
+      # lot of time, especially when a non-canonical branch has non-trivial
+      # depth. Note that if it turns out that a non-canonical branch eventually
+      # becomes canonical, it is vital to import it as quickly as possible.
       let
         parent_root = signedBlock.message.parent_root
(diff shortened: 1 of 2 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -473,6 +473,29 @@ proc storeBlock(
     parent = dag.checkHeadBlock(signedBlock)
 
   if parent.isErr():
+    if parent.error() == VerifierError.MissingParent:
+      let
+        parent_root = signedBlock.message.parent_root
+        parentBlck = dag.getForkedBlock(parent_root)
```
arnetheduck, line 493: why would we have the block here if checkHeadBlock fails with missing parent?

arnetheduck: ie that would indicate that a race happened between index and block write which should be .. rare?

etan-status: It happens when there are non-canonical branches from a previous run of the beacon node. We only assign `BlockRef` to the canonical chain on startup while rediscovering all other branches via sync / request manager. If non-canonical branches have significant length, that takes a very long time.

`checkParentRoot` actually already logs this case differently from the regular case:

```nim
  let parent = dag.getBlockRef(blck.parent_root).valueOr:
    # There are two cases where the parent won't be found: we don't have it or
    # it has been finalized already, and as a result the branch the new block
    # is on is no longer a viable fork candidate - we can't tell which is which
    # at this stage, but we can check if we've seen the parent block previously
    # and thus prevent requests for it to be downloaded again.
    let parentId = dag.getBlockId(blck.parent_root)
    if parentId.isSome() and parentId.get.slot < dag.finalizedHead.slot:
      debug "Block unviable due to pre-finalized-checkpoint parent",
        parentId = parentId.get()
      return err(VerifierError.UnviableFork)

    debug "Block parent unknown or finalized already", parentId
    return err(VerifierError.MissingParent)
```

`parentId` is set to `some` in case that the block is already available in the database. So, let's just load it from there to speed up discovery of other branches.

PR #6125, opened 2024-03-23
add branch discovery module for supporting chain stall situation

etan-status:
In split view situation, the canonical chain may only be served by a tiny amount of peers, and branches may span long durations. Minority branches may still have a large weight from attestations and should be discovered. To assist with that, add a branch discovery module that assists in such a situation by specifically targeting peers with unknown histories and downloading from them, in addition to sync manager work which handles popular branches.

### DIFF
--- a/beacon_chain/beacon_node.nim
+++ b/beacon_chain/beacon_node.nim
@@ -25,7 +25,7 @@ import
     attestation_pool, sync_committee_msg_pool, validator_change_pool],
   ./spec/datatypes/[base, altair],
   ./spec/eth2_apis/dynamic_fee_recipients,
-  ./sync/[sync_manager, request_manager],
+  ./sync/[branch_discovery, sync_manager, request_manager],
   ./validators/[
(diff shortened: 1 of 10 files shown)


tersec:
General question, is this testable without depending on rare events such as the Goerli deprecation?

etan-status:
It has the same problems as other network tests, we don't have infrastructure to mock timers / network in a way that would allow running through an accelerated but deterministic simulation. so beyond basic tests for the happy case, it's challenging.

The goal of structuring these chain-stall related PRs as separate modules etc is to avoid adding risk to the regular case. The additional modules apply only in a situation that is known to not be handled well to begin with. But, for reliable testing, additional tooling is necessary. Goerli right now is amazing for testing split view scenarios as there are 10+ heads available.

etan-status:
This module is incredibly effective. Discovered the canonical chain within ~8 hrs with only outbound peers, and managed to so before catching up on validator duties, ensuring that the situation is not made worse through adding extra useless heads.

PR #6135, opened 2024-03-25
add option to incrementally compute proposal state if behind

etan-status:
There can be situations where proposals need to be made on top of stale parent state. Because the distance to the wall slot is big, the proposal state (`getProposalState`) has to be computed incrementally to avoid lag spikes. Proposals have to be timely to have a chance to propagate on the network. Peak memory requirements don't change, but the proposal state needs to be allocated for a longer duration than usual.


etan-status:
This was working well on Goerli, but only in conjunction with #6136 . Without it, the BN would not discover the canonical chain in split view scenarios and would build on a different head.

Furthermore, while the memory requirement of these incrementally computed proposals is the same as during a regular proposal, it was noted that prolonging the memory consumption should rather not be done. Without keeping proposal state memory allocated across multiple slots, it is not possible to compute a proposal incrementally, because `epochRefState` keeps changing, and `clearanceState` tracks the latest imported state from _any_ branch (not just the canonical one) so also keeps changing as new branches are discovered by #6136 .

Putting the feature behind an opt-in flag was also discussed, but the preference is on a mechanism that can be turned on by default, whatever that would be.

PR #6136, opened 2024-03-25
add branch discovery module for use in split view scenarios

etan-status:
When the network is partitioned for a long time, e.g., Goerli, branches start forming where different peers have distinct views about the chain state. The current syncing solution with sync manager doesn't handle the case well, as it is optimized for a healthy network where syncing can be parallelized across different peers. To support sync manager discovering additional branches, a new module is added that pulls in histories from peers on unknown branches in a backwards manner.


etan-status:
This worked really well in the final month of Goerli, with Nimbus having kept up canonical block production until the end (https://goerli.beaconcha.in/blocks).

However, as documented in the file comments, better solutions could be engineered that potentially work even better. Based on discussion with team, we should only improve this once there is a realiable way for testing this.

PR #6140, opened 2024-03-26
ci: move help docs check script to its own file

jakubgs:
@etan-status please don't squeeze unreadable blobs like that into Jenkinsfiles.

And please don't squish a check/test into a build stage. Stages have names for a reason. Make use of them.

### DIFF
--- a/scripts/check_docs_help_msg.sh
+++ b/scripts/check_docs_help_msg.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2023-2024 Status Research & Development GmbH.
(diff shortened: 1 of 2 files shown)


scripts/check_docs_help_msg.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2023 Status Research & Development GmbH.
```
etan-status:
```suggestion
# Copyright (c) 2023-2024 Status Research & Development GmbH.
```

etan-status (approved):

PR #6149, opened 2024-03-27
rank peers by their score instead of their memory address

etan-status:
The `<` function to compare peers was not exported, leading to the same peer be acquired over and over again until kick. `mixin` doesn't pull it into `peerCmp` without `*` export, and with the export no mixin needed.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -456,15 +456,15 @@ func netKbps*(peer: Peer): float {.inline.} =
   ## Returns current network throughput average value in Kbps for peer ``peer``.
   round(((peer.netThroughput.average / 1024) * 10_000) / 10_000)
 
-func `<`(a, b: Peer): bool =
-  ## Comparison function, which first checks peer's scores, and if the peers'
-  ## score is equal it compares peers' network throughput.
-  if a.score < b.score:
-    true
-  elif a.score == b.score:
-    (a.netThroughput.average < b.netThroughput.average)
+# /!\ Must be exported to be seen by `peerCmp`
+func `<`*(a, b: Peer): bool =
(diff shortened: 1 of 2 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -456,15 +456,15 @@ func netKbps*(peer: Peer): float {.inline.} =
...
+# /!\ Must be exported to be seen by `peerCmp`
+func `<`*(a, b: Peer): bool =
+  ## Comparison function indicating `true` if peer `a` ranks worse than peer `b`
+  if a.score != b.score:
+    a.score < b.score
+  elif a.netThroughput.average != b.netThroughput.average:
+    a.netThroughput.average < b.netThroughput.average
   else:
-    false
+    system.`<`(a, b)
```
etan-status, line 467: memory address comparison added as fallback, just in case it is put into usage that requires `<` to be strictly monotonic. e.g., libc has some noobtraps there, not sure about nim :)

etan-status:
Will give this some testing, because the peer pool normally is full of null pointers that may now surface :)

cheatfate:
PR name is confusing, you adding compare by memory address actually. Also its almost impossible to get same network throughput.

etan-status:
Current `unstable` uses `system.<` because `peer_pool.<` is not exported.

In this PR, `system.<` is added as a fallback to `peer_pool.<` so that the total ordering that is used by the current implementation based on memory addresses is kept in-tact (so that if `a < b`, then `b > a`, which may or may not be required to use `HeapQueue` or other Nim data structures (it doesn't hurt to retain the total ordering).

PR name is correct, it adds the score (and bandwidth) comparisons to what peers are primarily sorted by.

tersec (approved):

PR #6151, opened 2024-03-28
filter out `nil` values when iterating peers

etan-status:
Iterating peers should only yield peers present in registry, otherwise `nil` pointers are returned and depending on comparison function it will break, see #6149.


cheatfate:
What the purpose of such iteration? You should not iterate peers of PeerPool.

etan-status:
The iterators are useful for tests and for diagnostic debug endpoints.

Note the iterators already exist, this just fixes them so that switching from `system.<` to `peer_pool.<` (#6149) becomes possible without leading to SIGSEGV when accessing the iterator. `peer_pool.<` expects non-nil arguments.

tersec (approved):

PR #6155, opened 2024-03-28
ssz: bump

arnetheduck:
should fix `dereferencing pointer to incomplete type` errors seen on some gcc versions


tersec:
https://github.com/nim-lang/Nim/issues/23453 is upstream Nim issue

PR #6159, opened 2024-03-30
nix: add initial basic Nix shell for development

jakubgs:
I understand the use of `USE_SYSTEM_NIM=1` is contentious but it does save a lot of compilation time, we just need to update `nixpkgs` to match the Nim compiler version used in `vendor` folder.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,32 @@
+{
+  pkgs ? import (builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/ae32f552b46fea56c8873504e8a199f68faa2d7f.tar.gz";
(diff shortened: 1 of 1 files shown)


arnetheduck:
https://github.com/status-im/nimbus-eth2/tree/stable/installer/nix / https://github.com/status-im/nimbus-eth2/blob/stable/.envrc?

jakubgs:
Oh, I had no idea this was done, nice! It was just hidden in a subfolder. That makes my life easier.

jakubgs:
Which does raise a question, should we use this Nix shell for CI builds?

jakubgs:
Also, any good reason why we don't keep the `flake.nix` in repo root for easier access?

arnetheduck:
> Which does raise a question, should we use this Nix shell for CI builds?

no, we should use the same docker images that we produce compiled binaries with

PR #6180, opened 2024-04-06
add packages to Nix flake config, move to repo root

jakubgs:
This way we can actually build and run a beacon node using just:
```sh
nix run 'github:status-im/nimbus-eth2?submodules=1'
```
The `?submodules=1` part should eventually not be necessary. For more details see:
* https://github.com/NixOS/nix/issues/4423


jakubgs:
Looks like the `submodule=1` flag doesn't work remotely with branch name for some reason:
```
 > nix run 'github:status-im/nimbus-eth2/nix-flake-packages?submodules=1#beacon_node'
error: builder for '/nix/store/kdf1l45rx93mz70nf7hzjgfb7m0szkzz-nimbus-eth2-unknown.drv' failed with exit code 2;
...
       > find: 'vendor/nimbus-build-system/scripts': No such file or directory
```
Seems like a bug. It works fie locally using `nix run '.?submodule=1#beacon_node'`.

nix/default.nix
```diff
@@ -0,0 +1,70 @@
...
+# See: https://github.com/status-im/nimbus-build-system/commits/master/vendor
+assert (
+  lib.assertMsg
+  (builtins.elem nim1.version ["1.6.16" "1.6.18"])
+  "Unsupported Nim version: ${nim1.version}"
+);
+
+stdenv.mkDerivation rec {
+  pname = "nimbus-eth2";
+  version = "unknown";
```
jakubgs: This is also a problem. In `go-waku` I can just do:
```nix
version = builtins.readFile ./VERSION
```
But this repo has no such file.

jakubgs: I've decided to parse the `beacon_chain/version.nim` rather than try to load a `VERSION` file at compile time.

nix/default.nix
```diff
@@ -0,0 +1,76 @@
...
+  targets ? ["nimbus_beacon_node"],
+  # These are the only platforms tested in CI and considered stable.
+  stableSystems ? [
+    "x86_64-linux" "aarch64-linux" "armv7a-linux"
+    "x86_64-darwin" "aarch64-darwin"
+    "x86_64-windows"
+  ],
+}:
+
+# Nim version(s) that are known to be stable
```
arnetheduck: if this is to be an "official" nix package in this repo, it must ascertain that the exact version of nim we lock in the submodule is used for nix building as well, ie no version leniency and no maintenance when we bump it (or it will go stale) - it should check what version is in the submodule and use that.

we do not condone any other versions to be used as they are frequently buggy and we remove workarounds once we've upgraded.

jakubgs: Right, but the problem with that is that [Nim compilation appears to require `git` usage to get `csources`](https://github.com/status-im/nimbus-build-system/blob/d7e7f1044def3c62d0b4bcec60c0b53876e44d71/scripts/build_nim.sh#L139-L174):
```
Building: Nim compiler
/build/source/vendor/nimbus-build-system/vendor/Nim /build/source
cmd: git clone -q --depth 1 -b master https://github.com/nim-lang/csources_v1.git csources_v1
cmd: cd csources_v1
ci/funs.sh: line 10: cd: csources_v1: No such file or directory
make[1]: *** [vendor/nimbus-build-system/makefiles/targets.mk:81: build-nim] Error 1
make: *** [vendor/nimbus-build-system/makefiles/targets.mk:124: vendor/nimbus-build-system/vendor/Nim/bin/nim] Error 2
```
Which is impossible to have available in a Nix build.

I thought I could avoid it using `QUICK_AND_DIRTY_COMPILER=1` and `QUICK_AND_DIRTY_NIMBLE=1`, but it doesn't appear to work. So we'd need some kind of custom setup to actually get the `csources`, which would also be separate from what `scripts/build_nim.sh` in `nimbus-build-system` fetches.

By not using Nim compiler form `nixpkgs` we are essentially forced to do a LOT of complex custom stuff to allow Nim compilation in the context of Nix.

jakubgs: I honestly don't understand why `nimbus-build-system` uses Git submodules for Nim compiler source code but does not use Git submodules for these `csources_v1`, `csources_v2`, and `nimble` sources. Do you know if there's any reason?

jakubgs: I see this was introduced by @etan-status in https://github.com/status-im/nimbus-build-system/commit/3866a8ab98fc6e0e6d406b88800aed72163d5fd4.

Why are we cloning these sources with a script instead of using submodules, like everything else does?

jakubgs: This used to work, but this commit I listed above has broken Nix builds of Nim compiler because the submodules were removed:

image

Which essentially means now we need to do a lot of extra custom work to support Nim compiler building.

etan-status: From the linked PR, you can see that `if grep -q skipIntegrityCheck koch.nim; then`, that none of the submodules were used, which is the case for any modern Nim version. The extra downloads via script are solely used when compiling legacy Nim versions. This change was introduced in https://github.com/status-im/nimbus-build-system/pull/61

Furthermore, these submodules could not have worked. The `vendor/Nim-csources-v1` and `vendor/nimble` submodules were outdated / references not maintained, and removed in the PR to avoid them being used accidentally (e.g., by someone passing the path to the outdated versions to the build command). This way, the correct version is always downloaded by the script when using it to build legacy Nim versions. Note that newer Nim versions also use `nim-lang/csources_v2` which was never checked out as a submodule.

The PR you linked also contains a comment that specifies the locations that specify the commit being used when compiling modern versions of Nim:

```nim
		# Run Nim buildchain, with matching dependency versions
		# - CSOURCES_REPO from Nim/config/build_config.txt (nim_csourcesUrl)
		# - CSOURCES_COMMIT from Nim/config/build_config.txt (nim_csourcesHash)
		# - NIMBLE_REPO from Nim/koch.nim (bundleNimbleExe)
		# - NIMBLE_COMMIT from Nim/koch.nim (NimbleStableCommit)
```

jakubgs: Thanks for the explanation.

So in theory we could just parse `Nim/koch.nim` and `Nim/config/build_config.txt` using Nix and get the commits of `csources_v1` and `nimble` to fetch via Nix. One snag is we'd need to update the SHA manually, but that is in a way good mechanism to prevent lagging behind with those sources when building via Nix, so it is actually a benefit.

It might be the simpler way than just trying to change how current build system works.

etan-status: `csources_v2` (the URL is also defined by the variables). Also have to teach Nim build system to actually pick up the custom checkouts.

jakubgs: Would we though? Maybe I misunderstand but it seems like `csources` and `nimble` overrides are more intended for developers, and actual package builds done by Nix would simply follow the pinned `Nim` version(and its dependencies) inside of `nimbus-build-system`. Is that wrong?

Araq: @jakubgs brought this issue to my attention and here is how Atlas builds a Nim version:

```nim

  var major, minor, patch: int
  if nimVersion != "devel":
    if not scanf(nimVersion, "$i.$i.$i", major, minor, patch):
      error c, "nim", "cannot parse version requirement"
      return
  let csourcesVersion =
    if nimVersion.isDevel or (major == 1 and minor >= 9) or major >= 2:
      # already uses csources_v2
      "csources_v2"
    elif major == 0:
      "csources" # has some chance of working
    else:
      "csources_v1"
```

from https://github.com/nim-lang/atlas/blob/master/src/nimenv.nim#L50

No patches are ever applied and the commit hash is not used either. However, more future proof would be: `"csources_v" & $major` if major >= 2 as version 3 should use csources_v3 etc.

jakubgs: Thanks for the explanation. If we don't have to worry about custom patches and the commits defined in the files listed earlier can be trusted we can fetch those things with Nix, main thing is how will build systems react if they are already in place.

We can't use a moving target like Atlas, both because Nix needs a concrete target with a hash, and also because it's just bad practice to use a moving target for a dependency when building complex software.

Appreciate the input.

Araq: I'm not suggesting to use Atlas, I'm saying that this is an officially supported way of building Nim because Atlas does it this way and Atlas is semi-official (its code is being integrated into Nimble).

The csources are not a moving target, they are frozen. But of course, use the hashes, it's fine.

jakubgs: The sad part is that Nim build setup actually tried to check if `nimble` folder is a Git repo and calls `git` via `/bin/sh`:
```sh
  if isGitRepo(destDir):
    let oldDir = getCurrentDir()
    setCurrentDir(destDir)
    try:
      execRetry "git fetch -q"
      exec fmt"git checkout -q {commit}"
    finally:
      setCurrentDir(oldDir)
```
https://github.com/status-im/Nim/blob/a749a8b742bd0a4272c26a65517275db4720e58a/tools/deps.nim#L30-L37

Which of course fails because there's no `/bin/sh`.

jakubgs: I have resolved this in a very ugly but effective way:
https://github.com/status-im/nimbus-eth2/blob/cb8d3628a31b52c9522d57ac0056b5972391bcc3/nix/default.nix#L50-L53
Not pretty. The clean way of doing it would involve flipping around that `if`/`elif` in upstream, which would take ages.

jakubgs: Made an attempt to fix this anyway:

* https://github.com/nim-lang/Nim/pull/23515

apentori (approved):
LGTM

tersec (approved):

tersec:
`jenkins/nimbus-eth2/linux/x86_64/nimv2` CI failure is because https://github.com/nim-lang/Nim/issues/23568

PR #6220, opened 2024-04-18
specifically use upstream Nim v2.0.4 for 2.0.x

tersec:
try working around https://github.com/nim-lang/Nim/issues/23519 until that's resolved either in `nimbus-eth2` or Nim


tersec:
right idea, slightly wrong execution, folding into another PR

PR #6228, opened 2024-04-22
el_manager initial refactor.

cheatfate:
Goals
1. Eliminate all the usages of "helpers" with proper primitives.
2. Add more error handlers and more error reporting (mostly on DEBUG level).
3. Adopt asyncraises usage.

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -1208,20 +1208,20 @@ proc processResponse(
             status2 = status
 
 proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
-  doAssert len(futures) > 0
-  try:
+  block:
     let pending = futures.filterIt(not(it.finished()))
-    await allFutures(pending).wait(30.seconds)
-  except CancelledError:
-    discard
-  except AsyncTimeoutError:
-    discard
+    if len(pending) > 0:
+      try:
+        await allFutures(pending).wait(30.seconds)
+      except CancelledError:
+        discard
+      except AsyncTimeoutError:
+        discard
 
-  try:
+  block:
     let pending = futures.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
-    await allFutures(pending)
-  except CancelledError:
-    discard
+    if len(pending) > 0:
+      await noCancel allFutures(pending)
 
 proc sendNewPayload*(
     m: ELManager,
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -303,7 +303,7 @@ proc engineApiRequest[T](
     failureAllowed = false
 ): Future[T] {.async: (raises: [CatchableError]).} =
   ## This procedure raises `CancelledError` and `DataProviderTimeout`
-  ## exceptions, and everything what `request` could raise.
+  ## exceptions, and everything which `request` could raise.
   try:
     let res = await request.wait(deadline)
     engine_api_request_duration_seconds.observe(
@@ -1580,10 +1580,10 @@ when hasDepositRootChecks:
         depositContract.get_deposit_root.call(blockNumber = blk.number)
       rawCountFut =
         depositContract.get_deposit_count.call(blockNumber = blk.number)
-      engineFut1 = connection.awaitEngineApiRequest(
+      engineFut1 = connection.engineApiRequest(
         depositRootFut, "get_deposit_root", startTime, deadline,
         failureAllowed = true)
-      engineFut2 = connection.awaitEngineApiRequest(
+      engineFut2 = connection.engineApiRequest(
         rawCountFut, "get_deposit_count", startTime, deadline,
         failureAllowed = true)
 
@@ -1778,7 +1778,7 @@ proc syncBlockRange(m: ELManager,
                 block_number = blk.number
           raise exc
         except CatchableError as exc:
-          debug "Request for block timestamp was failed",
+          debug "Request for block timestamp failed",
                 block_number = blk.number, reason = exc.msg
 
         let lastBlock = m.eth1Chain.blocks.peekLast
@@ -1794,7 +1794,7 @@ proc syncBlockRange(m: ELManager,
                     block_number = n
               raise exc
             except CatchableError as exc:
-              debug "Request for block was failed", block_number = n,
+              debug "Request for block failed", block_number = n,
                     reason = exc.msg
               raise exc
 
@@ -1912,7 +1912,7 @@ proc syncEth1Chain(m: ELManager, connection: ELConnection) {.
                   last_known_block_number = lastKnownBlock.number
             raise exc
           except CatchableError as exc:
-            debug "getBlockByNumber request has been failed",
+            debug "getBlockByNumber request failed",
                   last_known_block_number = lastKnownBlock.number,
                   reason = exc.msg
             raise exc
@@ -1938,7 +1938,7 @@ proc syncEth1Chain(m: ELManager, connection: ELConnection) {.
                 finalized_block_hash = finalizedBlockHash
           raise exc
         except CatchableError as exc:
-          debug "getBlockByHash() request has been failed",
+          debug "getBlockByHash() request has failed",
                 finalized_block_hash = finalizedBlockHash,
                 reason = exc.msg
           raise exc
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -19,7 +19,7 @@ import
   # Local modules:
   ../spec/[eth2_merkleization, forks],
   ../networking/network_metadata,
-  ".."/[beacon_node_status, future_combinators],
+  ".."/beacon_node_status,
   "."/[eth1_chain, el_conf]
 
 from std/sequtils import anyIt, mapIt
@@ -39,6 +39,7 @@ type
   WithdrawalCredentialsBytes = DynamicBytes[32, 32]
   SignatureBytes = DynamicBytes[96, 96]
   Int64LeBytes = DynamicBytes[8, 8]
+  WithoutTimeout* = distinct int
 
 contract(DepositContract):
   proc deposit(pubkey: PubKeyBytes,
@@ -56,6 +57,7 @@ contract(DepositContract):
                     index: Int64LeBytes) {.event.}
 
 const
+  noTimeout = WithoutTimeout(0)
   hasDepositRootChecks = defined(has_deposit_root_checks)
 
   targetBlocksPerLogsRequest = 1000'u64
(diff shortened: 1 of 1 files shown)


beacon_chain/el/el_manager.nim
```diff
@@ -1812,10 +1902,20 @@ proc syncEth1Chain(m: ELManager, connection: ELConnection) {.async.} =
...
+          try:
+            raiseIfNil await connection.engineApiRequest(
+              rpcClient.getBlockByNumber(lastKnownBlock.number),
+              "getBlockByNumber", Moment.now(), web3RequestsTimeout)
+          except CancelledError as exc:
+            debug "getBlockByNumber request has been interrupted",
+                  last_known_block_number = lastKnownBlock.number
+            raise exc
+          except CatchableError as exc:
+            debug "getBlockByNumber request has been failed",
```
tersec:
```suggestion
            debug "getBlockByNumber request has failed",
```

beacon_chain/el/el_manager.nim
```diff
@@ -1707,14 +1771,32 @@ proc syncBlockRange(m: ELManager,
...
+            try:
+              raiseIfNil await connection.engineApiRequest(
+                rpcClient.getBlockByNumber(n),
+                "getBlockByNumber", Moment.now(), web3RequestsTimeout)
+            except CancelledError as exc:
+              debug "The process of obtaining the block was interrupted",
+                    block_number = n
+              raise exc
+            except CatchableError as exc:
+              debug "Request for block was failed", block_number = n,
```
tersec:
```suggestion
              debug "Request for block failed", block_number = n,
```

beacon_chain/el/el_manager.nim
```diff
@@ -1707,14 +1771,32 @@ proc syncBlockRange(m: ELManager,
...
       if blk.number > fullSyncFromBlock:
-        await fetchTimestamp(connection, rpcClient, blk)
+        try:
+          await fetchTimestamp(connection, rpcClient, blk)
+        except CancelledError as exc:
+          debug "Request for block timestamp was interrupted",
+                block_number = blk.number
+          raise exc
+        except CatchableError as exc:
+          debug "Request for block timestamp was failed",
```
tersec:
```suggestion
          debug "Request for block timestamp failed",
```

beacon_chain/el/el_manager.nim
```diff
@@ -264,79 +276,64 @@ proc setDegradedState(connection: ELConnection,
...
+proc engineApiRequest[T](
+    connection: ELConnection,
+    request: Future[T],
+    requestName: string,
+    startTime: Moment,
+    deadline: Future[void] | Duration,
+    failureAllowed = false
+): Future[T] {.async: (raises: [CatchableError]).} =
+  ## This procedure raises `CancelledError` and `DataProviderTimeout`
+  ## exceptions, and everything what `request` could raise.
```
tersec:
```suggestion
  ## exceptions, and everything which `request` could raise.
```

beacon_chain/el/el_manager.nim
```diff
@@ -8,18 +8,18 @@
...
   web3, web3/[engine_api, primitives, conversions],
   eth/common/eth_types,
-  eth/async_utils, results,
+  results,
   stew/[assign2, byteutils, objects],
   # Local modules:
   ../spec/[eth2_merkleization, forks],
   ../networking/network_metadata,
-  ".."/[beacon_node_status, future_combinators],
+  ".."/beacon_node_status,
```
etan-status, line 22: Can delete the entire `future_combinators` file, it has no other usage.

beacon_chain/el/el_manager.nim
```diff
@@ -689,8 +702,7 @@ func getJsonRpcRequestHeaders(jwtSecret: Opt[seq[byte]]):
 
 proc newWeb3*(engineUrl: EngineApiUrl): Future[Web3] =
   newWeb3(engineUrl.url,
-          getJsonRpcRequestHeaders(engineUrl.jwtSecret),
-          httpFlags = {HttpClientFlag.NewConnectionAlways})
```
etan-status, line 693: Do we know why `NewConnectionAlways` was originally used here?

beacon_chain/el/el_manager.nim
```diff
@@ -264,79 +281,75 @@ proc setDegradedState(connection: ELConnection,
 
       connection.state = Degraded
 
-      asyncSpawn connection.close()
-      connection.web3 = none[Web3]()
-  of Degraded:
+      await connection.close()
```
etan-status, line 284: Could this increase the delay until a fallback client is used? `close` may run for 30s, and with `await` no progress is done during that time.

The new version is at least cleaner in tracking what is going on, the old version possibly run into situation where multiple close processes were running at same time I guess.

beacon_chain/el/el_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/d784672c107f846163082262334d4d7a4e625bd5

There is a possible failure mode in general without `NewConnectionAlways` where if the HTTP endpoint drops the connection first due to an internal timeout, then the client retries but too late because it had assumed the connection would be open from before because the connection pool hadn't properly noted the timeout -- this is what why the builder API RPCs do this, because in general they're not as frequent as engine API calls.

But engine API calls have to be happening every 12 seconds for the client to stay synced, and probably all the ELs hold open connections that long -- but avoiding said failure mode would be one reason to use this flag. It might be worth checking/testing how long, empirically, the ELs hold open the connections before introducing a potential regression.

tersec: To address the obvious point, yes, of course, HTTP clients should be prepared to retry things. However, sometimes the retries would happen too late, so it's important to try to maximize the chances of the first attempt working, which means unless the connection pool can virtually guarantee that the connection it will use hasn't died in the meantime, it's safer to bypass the connection pool. So "just retry, it's what network code has to handle anyway" is not an answer.

cheatfate: `NewConnectionAlways` was used here originally to avoid HTTP/1.1 pipeline issues we had. Because chronos HttpClient do not support HTTP/1.1 pipelining by default there is no reason to specify this flag anymore. Every request is being sent with new connection, connection is being closed as response is being received.

beacon_chain/el/el_manager.nim
```diff
@@ -1310,82 +1426,107 @@ proc forkchoiceUpdated*(m: ELManager,
...
+        if len(pendingRequests) == 0:
+          # All requests failed, we will continue our attempts until deadline
+          # is not finished.
+          break mainLoop
 
 # TODO can't be defined within exchangeConfigWithSingleEL
 func `==`(x, y: Quantity): bool {.borrow.}
 
-proc exchangeConfigWithSingleEL(m: ELManager, connection: ELConnection) {.async.} =
+proc exchangeConfigWithSingleEL(
```
tersec, line 1517: This is a bit of a misnomer now -- it hasn't been really `exchangeConfigWithSingleEL` since https://github.com/status-im/nimbus-eth2/pull/5585 and https://github.com/status-im/nimbus-eth2/pull/5889

beacon_chain/el/el_manager.nim
```diff
@@ -1763,18 +1949,27 @@ func hasProperlyConfiguredConnection*(m: ELManager): bool =
 
   false
 
-proc startExchangeTransitionConfigurationLoop(m: ELManager) {.async.} =
+proc startExchangeTransitionConfigurationLoop(
```
tersec, line 1952: Also a misnomer since https://github.com/status-im/nimbus-eth2/pull/5585 and https://github.com/status-im/nimbus-eth2/pull/5889 along with the `debug` log message, etc.

It checks chain ID now, does not exchange transition configuration

beacon_chain/el/el_manager.nim
```diff
@@ -1603,28 +1777,23 @@ proc new*(T: type ELManager,
...
+proc stop(m: ELManager) {.async: (raises: []).} =
+  if m.managerState notin {ELManagerState.Closing, ELManagerState.Closed}:
+    m.managerState = ELManagerState.Closing
+    var pending: seq[Future[void].Raising([])]
+    if not(m.chainSyncingLoopFut.isNil()) and
+       not(m.chainSyncingLoopFut.finished()):
+      pending.add(m.chainSyncingLoopFut.cancelAndWait())
+    if not(m.exchangeTransitionConfigurationLoopFut.isNil()) and
+       not(m.exchangeTransitionConfigurationLoopFut.finished()):
+      pending.add(m.exchangeTransitionConfigurationLoopFut.cancelAndWait())
```
tersec, line 1792: Do the waits here delay clean database closing in case of stuck connections, or at least allow the database to close cleanly?

cheatfate: This waits are proper cancellation, you should wait until loops will not be cancelled.

beacon_chain/el/el_manager.nim
cheatfate: What is your name proposal for it? It still performs network_id check.

beacon_chain/el/el_manager.nim
tersec: Ideally, yes. My question is whether in the case where it takes arbitrarily long to finish the loops, whether the database cleanup still happens first. If it doesn't, then it places the user in a less than great position. Ultimately the Chronos state is ephemeral and the state which has to remain intact for the next run is in the database.

beacon_chain/el/el_manager.nim
cheatfate: Every possible async task could run for 30s, that's not how we should protect the code from running for 30s.

beacon_chain/el/el_manager.nim
cheatfate: If you want to close database first - do it first, if you want to work properly in async world, you should signal your workers and wait them to complete their own cleanup processes. Overall construction of exit procedure in BN is incorrect, because it does not allow tasks to perform any cleanup procedures. You working with database in non-async way, if you think that you first task is close database - do it before signaling async tasks to finish.

beacon_chain/el/el_manager.nim
cheatfate: Procedures should not spawn procedure with `close` and do not wait for it actually being closed. As we seen this many times before its not a good practice which leads to, leaks and UB (when just closed transport being reused by OS and you will have 2 transports with same FD in process, one is closing and another was just opened).

beacon_chain/el/el_manager.nim
tersec: Fair enough, and out of scope of this PR.

beacon_chain/el/el_manager.nim
tersec: Alright, in which case I don't see any issue in removing it.

beacon_chain/el/el_manager.nim
tersec: sure, so `checkNetworkIdWithSingleEL`, say, or `checkChainIdWithSingleEL`

arnetheduck (approved):

beacon_chain/el/el_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6289

PR #6236, opened 2024-04-25
bump `geth` to `v1.14.0`

etan-status:
- https://github.com/ethereum/go-ethereum/releases/tag/v1.14.0

### DIFF
--- a/scripts/geth_binaries.sh
+++ b/scripts/geth_binaries.sh
@@ -21,7 +21,7 @@ source "${SCRIPTS_DIR}/bash_utils.sh"
 
 download_geth_stable() {
   if [[ ! -e "${STABLE_GETH_BINARY}" ]]; then
-    GETH_VERSION="1.13.15-c5ba367e"  # https://geth.ethereum.org/downloads
+    GETH_VERSION="1.14.0-87246f3c"  # https://geth.ethereum.org/downloads
     GETH_URL="https://gethstore.blob.core.windows.net/builds/"
 
     case "${OS}-${ARCH}" in


etan-status:
@jakubgs 

> Geth v1.14.0 switches over the default state trie representation from hash mode to path mode (i.e. --state.scheme flipped form hash to path) (https://github.com/ethereum/go-ethereum/pull/29108). This change does not affect Geth instances with pre-existing databases, in the case of which Geth continues to use whatever the existing database's format is. If no previous database exists however, for full nodes, Geth will now default to pathdb. The main advantage is built-in, online historical state pruning; no more runaway state growth.
>
> - If you want to force the old behaviour, you can run Geth with --state.scheme=hash for now. That said, we will be dropping hash mode sooner rather than later, so we advise everyone running full nodes to gradually switch, if they haven't yet.
> - Archive mode support is not yet finalised for path mode, so archive nodes will still run in hash mode. Naturally, hash mode will not be dropped until a full path archive lands and people have enough time to switch to it.

if we still have hash scheme based EL instances, may need to resync them eventually. for now, no action needed though, just a headsup.

etan-status:
@tersec

> Geth v1.14.0 switched to using Go v1.22 by default (https://github.com/ethereum/go-ethereum/pull/28946), which means we've dropped support for Go v1.20. Geth also started using of new features from Go v1.21 in our codebase, so building with Go v1.20 will probably error from now on. Just a heads up.

Does this increase the minimum OS that is realistically needed to run a full node?

tersec:
> @tersec
> 
> > Geth v1.14.0 switched to using Go v1.22 by default ([ethereum/go-ethereum#28946](https://github.com/ethereum/go-ethereum/pull/28946)), which means we've dropped support for Go v1.20. Geth also started using of new features from Go v1.21 in our codebase, so building with Go v1.20 will probably error from now on. Just a heads up.
> 
> Does this increase the minimum OS that is realistically needed to run a full node?

Run, no, but build Geth, yes.

[Debian](https://packages.debian.org/search?keywords=golang-go&searchon=names&exact=1&suite=all&section=all) and [Ubuntu](https://packages.ubuntu.com/search?keywords=golang-go&searchon=names&exact=1&suite=all&section=all), for example, have limited support for Go v1.21 with current stable versions. For Debian stable, the [backports repo](https://packages.debian.org/bookworm-backports/golang-go) contains v1.21, but that's not always installed or enabled by default, and some people using Debian stable versions might have specific reason not to want to introduce backports -- they chose a stable snapshot of package versions for a reason, potentially.

Ubuntu doesn't even offer that. The next Ubuntu LTS (24.04) will be out by end of month, and will support v1.22 by default, the previous LTS (22.04, supported for years yet) appears to be stuck on v1.18. There are ways of getting newer Go versions onto it, but if someone wants to customize/recompile Geth on their Nimbus node running older distros (which also [don't support v1.21 or v1.22 unless they're a rolling distro or very new](https://repology.org/project/go/versions), it'd be easier for them to cross-compile Geth from a Docker image of a current distro and use the resulting binary. Fortunately for this purpose, Go on Linux does not use libc unless the application specifically requests/requires it, so probably for Geth, which likely does not, one can use a compiled artifact of Geth, whether from the Geth binary release distribution or from any other source, regardless of distro version.

So, run, no, build, yes.

jakubgs:
Yes, some nodes probably still use the old schema. Is there a simple way to figure out which ones are those?

Also, commits titles - or bodies - are not markdown, so please stop putting markdown syntax in commit titles. Just because GitHub renders them doesn't mean it's a good practice.

PR #6240, opened 2024-04-25
bump `nim-web3` to `290acb6e6200230fc625209e6654f6bb60aceb13`

etan-status:
- Change some FeeHistoryResult field types from Quantity to UInt256
- bump `execution-apis` to `v1.0.0-beta.4`


tersec:
It's convenient anyway this apparently hasn't happened, or this bump doesn't include such commits, because fixing this would interfere with the EL refactor, but there probably should be some of the other renames of data types in `nim-web3` at some point -- the EL withdrawal requests were/are labeled as `ExitV1`s initially, but that should be fixed before interop.

etan-status:
The EL manager doesn't touch fee history, that's why this bump should be safe at this time.

PR #6253, opened 2024-04-29
Chore/update libp2p branch master

kaiserd:
changes the default branch of libp2p from unstable to master.
See https://github.com/vacp2p/nim-libp2p/issues/1088

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,7 +24,7 @@
 	path = vendor/nim-libp2p
 	url = https://github.com/vacp2p/nim-libp2p.git
 	ignore = untracked
-	branch = unstable
+	branch = master
 [submodule "vendor/nimbus-build-system"]
 	path = vendor/nimbus-build-system
 	url = https://github.com/status-im/nimbus-build-system.git


tersec:
- should target `unstable` branch of `nimbus-eth2`
- it looks like all the non-`gitmodule` changes are artifacts of this

kaiserd:
Oh sorry, I missed that. Fixed.

tersec:
Lint failure is
```
Run while read -r file; do
Submodule 'vendor/nim-libp2p': '2b5319622c997ce1c80bc62c863e30f3349ee0d7' (2024-04-25 13:01:29 +0000) is more recent than latest 'master' (2023-12-18 18:04:02 +0000) (branch config: '.gitmodules')
```

This is triggered by
https://github.com/status-im/nimbus-eth2/blob/1ab6f1654fa5221e1f79c54bf5150f4f4410e57c/.github/workflows/ci.yml#L293-L297

Need to update `nim-libp2p` submodule itself along with this, in `vendor/`, because right now it's pointing to/commithash-pinned to https://github.com/vacp2p/nim-libp2p/tree/2b5319622c997ce1c80bc62c863e30f3349ee0d7 from 5 days ago where in the version of libp2p vendored it's still part of the `unstable` branch from which it was renamed:
```
commit 2b5319622c997ce1c80bc62c863e30f3349ee0d7 (grafted, HEAD, origin/unstable, origin/HEAD, unstable)
Author: Diego <diego@status.im>
Date:   Thu Apr 25 15:01:29 2024 +0200

    Revert "always allow new data be received if the recvWindow is > 0"
    
    This reverts commit 5cbb473d1b765307ee15c73f47e180ae3a81fdb3.
```
whereas as it notes, the most recent `master` commit before the recent libp2p branch reorganization was from last year. So this change and the module update/bump need to happen mutually/internally consistently.

.gitmodules
```diff
@@ -24,7 +24,7 @@
 	path = vendor/nim-libp2p
 	url = https://github.com/vacp2p/nim-libp2p.git
 	ignore = untracked
-	branch = unstable
+	branch = master
```
etan-status, line 27: This is not pointing to the default branch, but instead to the one that should be used for building. If we should no longer use unstable version of libp2p, have to checkout a different ref

kaiserd:
I added an empty commit to libp2p and did a dummy-bump in the latest commit of this PR.
This allowed pleasing the linter.

tersec (approved):

.gitmodules
tersec: That was the intent; the vacuous bump there now enforces/ensures this gets picked up by `git`. libp2p is switching from `unstable` to `master` for this purpose. Before `master` hadn't been updating enough to make it worthwhile; now it will.

PR #6255, opened 2024-05-01
initial electra attestation pool changes; electra block_sim

tersec:
WIP, not mergeable yet, but make sure there's an internally consistent branch ready.

Captures some of the logic changes, but too much duplication; most can become generic.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -30,17 +30,24 @@ const
     ## that potentially could be added to a newly created block
 
 type
-  OnAttestationCallback = proc(data: phase0.Attestation) {.gcsafe, raises: [].}
+  OnPhase0AttestationCallback =
+    proc(data: phase0.Attestation) {.gcsafe, raises: [].}
+  OnElectraAttestationCallback =
+    proc(data: electra.Attestation) {.gcsafe, raises: [].}
 
-  Validation = object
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -32,25 +32,38 @@ const
 type
   OnAttestationCallback = proc(data: phase0.Attestation) {.gcsafe, raises: [].}
 
-  Validation = object
+  Validation[CVBType] = object
     ## Validations collect a set of signatures for a distict attestation - in
     ## eth2, a single bit is used to keep track of which signatures have been
     ## added to the aggregate meaning that only non-overlapping aggregates may
     ## be further combined.
-    aggregation_bits: CommitteeValidatorsBits
```
arnetheduck, line 40: why not unify at this abstraction level? ie the information is the same, just different encoding..

tersec: There's still the `committee_bits` change between these attestations, though that can be handled in other ways. Still needs some distinct codepaths, regardless.

More immediately, as long as those somewhat-distinct codepaths exist, the main gains right now possible to get this mergeable are from the various functions which I duplicated for convenience but which shouldn't be.

tersec:
Not in great condition, but doesn't hurt existing code either. Better given time constraints to merge this and improve it from `unstable`.

Issue #6075, opened 2024-03-14
POST to `eth/v1/builder/blinded_blocks` missing header


mcdee:
**Describe the bug**

When receiving a POST to `/eth/v1/builder/blinded_blocks` the spec says that we should obtain the version of the blinded block in the `Eth-Consensus-Version` header, however this is not set by Nimbus.

```
Nimbus beacon node v24.2.2-fc9c72-stateofus
Copyright (c) 2019-2024 Status Research & Development GmbH

eth2 specification v1.4.0-beta.7-hotfix

Nim Compiler Version 1.6.18 [Linux: amd64]
```

tersec:
https://github.com/status-im/nimbus-eth2/pull/6256

PR #6256, opened 2024-05-02
Fix POST to eth/v1/builder/blinded_blocks missing header Eth-Consensus-Version.

cheatfate:
This PR addresses #6075

### DIFF
--- a/beacon_chain/validators/message_router_mev.nim
+++ b/beacon_chain/validators/message_router_mev.nim
@@ -59,16 +59,21 @@ proc unblindAndRouteBlockMEV*(
   # protection check
   let response =
     try:
-      awaitWithTimeout(
-          payloadBuilderRestClient.submitBlindedBlock(blindedBlock),
-          BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE):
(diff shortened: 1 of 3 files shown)


beacon_chain/validators/message_router_mev.nim
```diff
@@ -59,16 +59,21 @@ proc unblindAndRouteBlockMEV*(
   # protection check
   let response =
     try:
-      awaitWithTimeout(
-          payloadBuilderRestClient.submitBlindedBlock(blindedBlock),
-          BUILDER_BLOCK_SUBMISSION_DELAY_TOLERANCE):
-        return err("Submitting blinded block timed out")
+      await payloadBuilderRestClient.submitBlindedBlock(blindedBlock).
```
tersec, line 61: Does `message_router_mev` still need to
```nim
from eth/async_utils import awaitWithTimeout
```
with this change?

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6256/commits/8ba8f406db045f3d1069d101baade76340c62b41

tersec (approved):

PR #6263, opened 2024-05-04
use correct minimum size when reading block / state headers

etan-status:
`sizeof` also includes padding between fields, while SSZ defines `fixedPortionSize` (on type) or `sszSize` (on value) to denote required bytes to encode. Switch forked block/state readers to SSZ size. As blocks/states are much larger than the padding, this doesn't affect practical use cases but is slightly more correct this way.


tersec:
https://github.com/nim-lang/Nim/issues/23568

tersec:
Updating to latest `unstable` should get this building again

beacon_chain/spec/forks.nim
```diff
@@ -1360,31 +1360,31 @@ func readSszForkedHashedBeaconState*(cfg: RuntimeConfig, data: openArray[byte]):
...
+  if data.len() < numHeaderBytes:
+    raise (ref MalformedSszError)(msg: "Incomplete BeaconState header")
   let header = SSZ.decode(
-    data.toOpenArray(0, sizeof(BeaconStateHeader) - 1),
-    BeaconStateHeader)
+    data.toOpenArray(0, numHeaderBytes - 1), BeaconStateHeader)
 
   # TODO https://github.com/nim-lang/Nim/issues/19357
-  result = readSszForkedHashedBeaconState(cfg, header.slot, data)
+  readSszForkedHashedBeaconState(cfg, header.slot, data)
```
tersec: In theory, this `result =` is to work around https://github.com/nim-lang/Nim/issues/19357 despite its being less-than-ideal code style. https://github.com/status-im/nimbus-eth2/pull/3050 appears to have introduced this, and notes:
> * fix several cases of states being stored on stack in tests, causing
random failures on some platforms

Maybe it's not necessary, but the referenced Nim issue has not been resolved/closed.

If it's not necessary anymore in this situation, that would be better, but that would require some evidence of the codegen improvement.

beacon_chain/spec/forks.nim
```diff
@@ -1360,31 +1360,31 @@ func readSszForkedHashedBeaconState*(cfg: RuntimeConfig, data: openArray[byte]):
     ForkedHashedBeaconState {.raises: [SerializationError].} =
   ## Read a state picking the right fork by first reading the slot from the byte
   ## source
-  if data.len() < sizeof(BeaconStateHeader):
-    raise (ref MalformedSszError)(msg: "Not enough data for BeaconState header")
+  const numHeaderBytes = fixedPortionSize(BeaconStateHeader)
+  if data.len() < numHeaderBytes:
+    raise (ref MalformedSszError)(msg: "Incomplete BeaconState header")
```
arnetheduck, line 1365: do we even need this raise? feels like `SSZ.decode` should already raise when there's not enough data

etan-status: Yes, we only pass `toOpenArray(0, numHeaderBytes - 1)`, which would possibly `Defect` without the check.

beacon_chain/spec/forks.nim
arnetheduck: finding out requires going to the c code and looking for copies - it's a messy process - at some point, this should probably be pursued as a task, ie codegen quality review, with upstream .. including tests.

PR #6273, opened 2024-05-08
Uses `nimbus-eth2` 24.3 stable release, joining Kaustinen Devnet (Verkle Readiness)

agnxsh:
Adds support for Verkle-enabled Execution Clients following a hard fork. 2 such examples are:
- [Geth](https://github.com/gballet/go-ethereum)
- [Nethermind](https://github.com/NethermindEth/nethermind/tree/verkle-devnet-7)

References for spec and other implementations:
- https://github.com/sigp/lighthouse/issues/4042
- https://github.com/ethereum/consensus-specs/pull/3230
- https://github.com/ethereum/EIPs/pull/6800

NOTE: For the devnet, Kaustinen is rebased on top of Capella, skipping Deneb, for some reason this fork name has been renamed from `Verge` to `Electra`.

The plan to make this work is to set DENEB_FORK_EPOCH to FAR_FUTURE_EPOCH and make Capella, Altair, Bellatrix to 0, as the devnet starts from genesis.

### DIFF
--- a/beacon_chain/spec/datatypes/electra.nim
+++ b/beacon_chain/spec/datatypes/electra.nim
@@ -549,16 +549,16 @@ type
   #   bls_to_execution_changes*:
   #     List[SignedBLSToExecutionChange, Limit MAX_BLS_TO_EXECUTION_CHANGES]
 
-  BeaconStateDiffPreSnapshot* = object
-    eth1_data_votes_recent*: seq[Eth1Data]
-    eth1_data_votes_len*: int
-    slot*: Slot
-    historical_summaries_len*: int
-    eth1_withdrawal_credential*: seq[bool]
+  # BeaconStateDiffPreSnapshot* = object
+  #   eth1_data_votes_recent*: seq[Eth1Data]
+  #   eth1_data_votes_len*: int
+  #   slot*: Slot
+  #   historical_summaries_len*: int
+  #   eth1_withdrawal_credential*: seq[bool]
 
-  IndexedWithdrawalCredentials* = object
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/datatypes/electra.nim
```diff
@@ -549,16 +549,16 @@ type
...
+  #   eth1_data_votes_recent*: seq[Eth1Data]
+  #   eth1_data_votes_len*: int
+  #   slot*: Slot
+  #   historical_summaries_len*: int
+  #   eth1_withdrawal_credential*: seq[bool]
 
-  IndexedWithdrawalCredentials* = object
-    validator_index*: uint64
-    withdrawal_credentials*: Eth2Digest
+  # IndexedWithdrawalCredentials* = object
```
tersec, line 559: Why add this back? There was never a reason to have added this, and there still isn't, commented or otherwise.

beacon_chain/spec/forks.nim
```diff
@@ -61,8 +61,8 @@ type
     altair.HashedBeaconState |
     bellatrix.HashedBeaconState |
     capella.HashedBeaconState |
-    deneb.HashedBeaconState #|
-    # electra.HashedBeaconState
```
tersec, line 65: Presumably this changes anyway retargeting Capella fork instead of Electra fork, but having `electra.Foo` as part of these shouldn't have been causing harm to begin with, regardless. Not sure why they were changed/removed?

tersec:
Obsolete.

PR #6292, opened 2024-05-15
hashtree: bump

arnetheduck:
Fixes compilation on osx and arm64


tersec:
This PR now shows no changes left -- does it/should it still being doing something?

arnetheduck:
oops, bad merge - here's the bump

PR #6293, opened 2024-05-16
Fix el-manager warnings and issues after refactoring.

cheatfate:


beacon_chain/el/el_manager.nim
```diff
@@ -747,21 +746,21 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
 proc getBlockByHash(
     rpcClient: RpcClient,
     hash: BlockHash
-): Future[BlockObject] {.async: (raising: [CatchableError], raw: true).} =
-  rpcClient.eth_getBlockByHash(hash, false)
+): Future[BlockObject] {.async: (raises: [CatchableError]).} =
+  await rpcClient.eth_getBlockByHash(hash, false)
```
arnetheduck, line 750: this gives a warning? it shouldn't - ie raw here saves a few allocations when forwarding to another async..

arnetheduck, line 750:
```suggestion
): Future[BlockObject] {.async: (raises: [CatchableError], raw: true).} =
  rpcClient.eth_getBlockByHash(hash, false)
```

cheatfate: of course but we can't propagate `asyncraises` here, because `rpcClient.eth_getBlockByHash(hash, false)` returns `Future[T]` and not `Future[T, E]`.
Approach you want is of course desirable but only after we adopt `nim-web3`, `nim-json-rpc` and `nim-websock`.

cheatfate: There is no way to unwrap Future[T] returned by `rpcClient.eth_getBlockByHash(hash, false)` to Future[T, E] required for `{.asyncraises.}` without `await`.

arnetheduck: ah, good point :+1:

arnetheduck (approved):

PR #6294, opened 2024-05-16
[VC] Add builderBoostFactor support.

cheatfate:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -1101,26 +1101,37 @@ proc collectBids(
...
-  let
-    scaledBuilderValue = (builderValue shr scalingBits) * 100
-    scaledEngineValue = engineValue shr scalingBits
-  scaledBuilderValue >
-    scaledEngineValue * (localBlockValueBoost.uint16 + 100).u256
+func builderBetterBid*(builderBoostFactor: uint64,
+                       builderValue: UInt256, engineValue: Wei): bool =
+  if builderBoostFactor == 0'u64:
+    false
+  elif builderBoostFactor == 100'u64:
```
tersec, line 1138: Seems like a special case which shouldn't be needed. Possibly slightly faster, but also means that the interesting codepath isn't executed as much.

beacon_chain/conf.nim
```diff
@@ -640,12 +640,12 @@ type
         defaultValue: ""
         name: "payload-builder-url" .}: string
 
-      # Flag name and semantics borrowed from Prysm
```
tersec, line 643: - this undoes the 10% local default boost; intentional?
- replaces the same `--builder-boost-factor` with a same-name option with different semantics, which will surprise people who have set it.

probably as a start, might make more sense to leave existing conf.nim option unchanged and translate it into the `produceBlockV3` semantics. Can change the command-line option and the (now more general) backend separately.

tersec, line 643: - this undoes the 10% local default boost; intentional?
- replaces the same `--builder-boost-factor` with a same-name option with different semantics, which will surprise people who have set it.

probably as a start, might make more sense to leave existing conf.nim option unchanged and translate it into the `produceBlockV3` semantics. Can change the command-line option and the (now more general) backend separately.

beacon_chain/validators/beacon_validators.nim
cheatfate: According to

> builder_boost_factor

> Percentage multiplier to apply to the builder's payload value when choosing between a builder payload header and payload from the paired execution node. This parameter is only relevant if the beacon node is connected to a builder, deems it safe to produce a builder payload, and receives valid responses from both the builder endpoint and the paired execution node. When these preconditions are met, the server MUST act as follows:

    if exec_node_payload_value >= builder_boost_factor * (builder_payload_value // 100), then return a full (unblinded) block containing the execution node payload.
    otherwise, return a blinded block containing the builder payload header.

> Servers must support the following values of the boost factor which encode common preferences:

    builder_boost_factor=0: prefer the execution node payload unless an error makes it unviable.
    builder_boost_factor=100: default profit maximization mode; choose whichever payload pays more.
    builder_boost_factor=2**64 - 1: prefer the builder payload unless an error or beacon node health check makes it unviable.

> Servers should use saturating arithmetic or another technique to ensure that large values of the builder_boost_factor do not trigger overflows or errors. If this parameter is provided and the beacon node is not configured with a builder then the beacon node MUST respond with a full block, which the caller can choose to reject if it wishes. If this parameter is not provided then it should be treated as having the default value of 100. If the value is provided but out of range for a 64-bit unsigned integer, then an error response with status code 400 MUST be returned.

In case of `100` we just multiplying with `1`, so i do not see any reasons to make multiplication and overflow check which involves division.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1101,26 +1101,37 @@ proc collectBids(
...
+    builderValue >= engineValue
+  elif builderBoostFactor == high(uint64):
+    true
+  else:
+    let
+      zero256 = 0.u256
```
tersec: can/should be `const`, if useful

etan-status: there's actually UInt256.zero and UInt256.one ;-)

beacon_chain/conf.nim
```diff
@@ -1031,6 +1031,13 @@ type
       defaultValue: false
       name: "distributed".}: bool
 
+    builderBoostFactor* {.
+      desc: "Percentage multiplier to apply to the builder's payload value " &
+            "when choosing between a builder payload header and payload " &
+            "from the paired execution node."
+      defaultValue: 100,
+      name: "builder-boost-factor".}: uint64
```
tersec, line 1040: Other issue noted above aside, this switches from `uint8` to `uint64`. The `uint8` was deliberate, and whether or not a good idea, it's pointless flexibility now to add to a basically deprecated option. Any new changes should involve the spec version, and would create additional scenarios to support

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6294/commits/6a8f10fb48e9222a27d7320d781dcaa5e6052c83

beacon_chain/validators/beacon_validators.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6294/commits/6a8f10fb48e9222a27d7320d781dcaa5e6052c83

beacon_chain/conf.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6294/commits/6a8f10fb48e9222a27d7320d781dcaa5e6052c83

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6294/commits/6a8f10fb48e9222a27d7320d781dcaa5e6052c83

tersec (approved):

PR #6300, opened 2024-05-20
VC: Deprecate fallback path to produceBlockV2, use produceBlockV3 always.

cheatfate:


tersec:
In general, seems like a good idea if the BNs support it -- which is likely, as `produceBlockV3` is standard in spec, but not yet certainly known.

Not yet confirmed CLs are mainly Teku and Prysm, with Prysm being a bit ambiguous because of their gRPC focus and generally not-widespread interoperability in the most recent EF VC/BN tests anyway.

tersec:
https://github.com/status-im/nimbus-eth2/pull/5644#issuecomment-1843025498

Teku's fine.

Going off https://github.com/kurtosis-tech/ethereum-package?tab=readme-ov-file#beacon-node--validator-client-compatibility Prysm's BN does support other VCs in general (the default incompatibility is from the VC side), so that should be verified too.

tersec:
Prysm confirmed

tersec (approved):

PR #6322, opened 2024-05-29
Adding Kzg Specs, integration to `gossipValidation` and it's adjoining tests

agnxsh:
This branch is for parallely adding kzg/ other EIP7594 (peerDAS) specs/ tests as and when things are completed. Once things look decent enough, they are going to be merged into the `wip-peerdas` branch.

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -13,7 +13,7 @@ import
   results,
   # Internals
   ../spec/[
-    beaconstate, state_transition_block, forks, helpers, network, signatures],
+    beaconstate, state_transition_block, forks, helpers, network, signatures, eip7594_helpers],
   ../consensus_object_pools/[
     attestation_pool, blockchain_dag, blob_quarantine, block_quarantine,
     spec_cache, light_client_pool, sync_committee_msg_pool,
@@ -206,6 +206,18 @@ func check_blob_sidecar_inclusion_proof(
     return errReject(res.error)
 
   ok()
+
+func check_data_column_sidecar_inclusion_proof(
+  data_column_sidecar: DataColumnSidecar): Result[void, ValidationError] =
+  let res = data_column_sidecar.verify_data_column_sidecar_inclusion_proof()
+  if res.isErr:
+    return errReject(res.error)
+
+proc check_data_column_sidecar_kzg_proofs(
+  data_column_sidecar: DataColumnSidecar): Result[void, ValidationError] =
+  let res = data_column_sidecar.verify_data_column_sidecar_kzg_proofs()
+  if res.isErr:
+    return errReject(res.error)
 
 # Gossip Validation
 # ----------------------------------------------------------------
@@ -502,11 +514,19 @@ proc validateDataColumnSidecar*(
   if not (block_header.slot > dag.finalizedHead.slot):
     return errIgnore("DataColumnSidecar: slot already finalized")
 
-  # TODO: [REJECT] The sidecar's `kzg_commitments` inclusion proof is valid as verified by
+  # [REJECT] The sidecar's `kzg_commitments` inclusion proof is valid as verified by
   # `verify_data_column_sidecar_inclusion_proof(sidecar)`.
+  block:
+    let v = check_data_column_sidecar_inclusion_proof(data_column_sidecar)
+    if v.isErr:
+      return dag.checkedReject(v.error)
 
-  # TODO: [REJECT] The sidecar's column data is valid as 
+  # [REJECT] The sidecar's column data is valid as 
   # verified by `verify_data_column_kzg_proofs(sidecar)`
+  block:
+    let r = check_data_column_sidecar_kzg_proofs(data_column_sidecar)
+    if r.isErr:
+      return dag.checkedReject(r.error)
 
   # [IGNORE] The sidecar is the first sidecar for the tuple
   # (block_header.slot, block_header.proposer_index, blob_sidecar.index)
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -521,22 +521,22 @@ proc validateDataColumnSidecar*(
     if v.isErr:
       return dag.checkedReject(v.error)
 
-  # [REJECT] The sidecar's column data is valid as 
-  # verified by `verify_data_column_kzg_proofs(sidecar)`
-  block:
-    let r = check_data_column_sidecar_kzg_proofs(data_column_sidecar)
-    if r.isErr:
-      return dag.checkedReject(r.error)
+  # # [REJECT] The sidecar's column data is valid as 
+  # # verified by `verify_data_column_kzg_proofs(sidecar)`
+  # block:
+  #   let r = check_data_column_sidecar_kzg_proofs(data_column_sidecar)
+  #   if r.isErr:
+  #     return dag.checkedReject(r.error)
 
   # [IGNORE] The sidecar is the first sidecar for the tuple
   # (block_header.slot, block_header.proposer_index, blob_sidecar.index)
   # with valid header signature, sidecar inclusion proof, and kzg proof.
   let block_root = hash_tree_root(block_header)
   if dag.getBlockRef(block_root).isSome():
-    return errIgnore("BlobSidecar: already have block")
+    return errIgnore("DataColumnSidecar: already have block")
   if blobQuarantine[].hasBlob(
       block_header.slot, block_header.proposer_index, data_column_sidecar.index):
-    return errIgnore("BlobSidecar: already have valid blob from same proposer")
+    return errIgnore("DataColumnSidecar: already have valid blob from same proposer")
 
   # [IGNORE] The sidecar's block's parent (defined by
   # `block_header.parent_root`) has been seen (via both gossip and
@@ -545,13 +545,13 @@ proc validateDataColumnSidecar*(
   #
   # [REJECT] The sidecar's block's parent (defined by
   # `block_header.parent_root`) passes validation.
-  let parent = dag.getBlockRef(block_header.parent_root).valueOr:
-    if block_header.parent_root in quarantine[].unviable:
-      quarantine[].addUnviable(block_root)
-      return dag.checkedReject("DataColumnSidecar: parent not validated")
-    else:
-      quarantine[].addMissing(block_header.parent_root)
-      return errIgnore("DataColumnSidecar: parent not found")
+  # let parent = dag.getBlockRef(block_header.parent_root).valueOr:
+  #   if block_header.parent_root in quarantine[].unviable:
+  #     quarantine[].addUnviable(block_root)
+  #     return dag.checkedReject("DataColumnSidecar: parent not validated")
+  #   else:
+  #     quarantine[].addMissing(block_header.parent_root)
+  #     return errIgnore("DataColumnSidecar: parent not found")
 
   # [REJECT] The sidecar is proposed by the expected `proposer_index`
   # for the block's slot in the context of the current shuffling
@@ -562,10 +562,10 @@ proc validateDataColumnSidecar*(
   # REJECT, instead IGNORE this message.
   let proposer = getProposer(dag, parent, block_header.slot).valueOr:
     warn "cannot compute proposer for blob"
-    return errIgnore("BlobSidecar: Cannot compute proposer") # internal issue
+    return errIgnore("DataColumnSidecar: Cannot compute proposer") # internal issue
 
   if uint64(proposer) != block_header.proposer_index:
-    return dag.checkedReject("BlobSidecar: Unexpected proposer")
+    return dag.checkedReject("DataColumnSidecar: Unexpected proposer")
 
   # [REJECT] The proposer signature of `blob_sidecar.signed_block_header`,
   # is valid with respect to the `block_header.proposer_index` pubkey.
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -549,8 +549,8 @@ proc validateDataColumnSidecar*(
     if block_header.parent_root in quarantine[].unviable:
       quarantine[].addUnviable(block_root)
       return dag.checkedReject("DataColumnSidecar: parent not validated")
-    else:
-      quarantine[].addMissing(block_header.parent_root)
+    # else:
+    #   quarantine[].addMissing(block_header.parent_root)
       # return errIgnore("DataColumnSidecar: parent not found")
 
   # [REJECT] The sidecar is proposed by the expected `proposer_index`
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -213,11 +213,15 @@ func check_data_column_sidecar_inclusion_proof(
   if res.isErr:
     return errReject(res.error)
 
+  ok()
+
 proc check_data_column_sidecar_kzg_proofs(
   data_column_sidecar: DataColumnSidecar): Result[void, ValidationError] =
   let res = data_column_sidecar.verify_data_column_sidecar_kzg_proofs()
   if res.isErr:
     return errReject(res.error)
+
+  ok()
 
 # Gossip Validation
 # ----------------------------------------------------------------
@@ -523,10 +527,10 @@ proc validateDataColumnSidecar*(
 
   # # [REJECT] The sidecar's column data is valid as 
   # # verified by `verify_data_column_kzg_proofs(sidecar)`
-  # block:
-  #   let r = check_data_column_sidecar_kzg_proofs(data_column_sidecar)
-  #   if r.isErr:
-  #     return dag.checkedReject(r.error)
+  block:
+    let r = check_data_column_sidecar_kzg_proofs(data_column_sidecar)
+    if r.isErr:
+      return dag.checkedReject(r.error)
 
   # [IGNORE] The sidecar is the first sidecar for the tuple
   # (block_header.slot, block_header.proposer_index, blob_sidecar.index)
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -536,7 +536,7 @@ proc validateDataColumnSidecar*(
     return errIgnore("DataColumnSidecar: already have block")
   if blobQuarantine[].hasBlob(
       block_header.slot, block_header.proposer_index, data_column_sidecar.index):
-    return errIgnore("DataColumnSidecar: already have valid blob from same proposer")
+    return errIgnore("DataColumnSidecar: already have valid data column from same proposer")
 
   # [IGNORE] The sidecar's block's parent (defined by
   # `block_header.parent_root`) has been seen (via both gossip and
@@ -545,13 +545,13 @@ proc validateDataColumnSidecar*(
   #
   # [REJECT] The sidecar's block's parent (defined by
   # `block_header.parent_root`) passes validation.
-  let parent = dag.getBlockRef(block_header.parent_root).valueOr:
-    if block_header.parent_root in quarantine[].unviable:
-      quarantine[].addUnviable(block_root)
-      return dag.checkedReject("DataColumnSidecar: parent not validated")
-    # else:
-    #   quarantine[].addMissing(block_header.parent_root)
-      # return errIgnore("DataColumnSidecar: parent not found")
+  # let parent = dag.getBlockRef(block_header.parent_root).valueOr:
+  #   if block_header.parent_root in quarantine[].unviable:
+  #     quarantine[].addUnviable(block_root)
+  #     return dag.checkedReject("DataColumnSidecar: parent not validated")
+  #   else:
+  #     quarantine[].addMissing(block_header.parent_root)
+  #     return errIgnore("DataColumnSidecar: parent not found")
 
   # [REJECT] The sidecar is proposed by the expected `proposer_index`
   # for the block's slot in the context of the current shuffling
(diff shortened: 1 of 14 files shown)


beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -8,649 +8,20 @@
 {.push raises: [].}
 
 # Uncategorized helper functions from the spec
-
 import
+  tables,
```
tersec:
```suggestion
  std/[algorithm, macros, tables],
```
and then the separate `algorithm` and `std/macros` lines can be removed. it's preferred to use `std/foo` for stdlib modules to ensure they're sourced from the stdlib specifically, not a similarly named module from elsewhere.

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
 proc get_custody_columns*(node_id: NodeId, custody_subnet_count: uint64): Result[seq[ColumnIndex], cstring] =
     
-    # assert custody_subnet_count <= DATA_COLUMN_SIDECAR_SUBNET_COUNT
-    if not (custody_subnet_count <= DATA_COLUMN_SIDECAR_SUBNET_COUNT):
-        return err("Eip7594: Custody subnet count exceeds the DATA_COLUMN_SIDECAR_SUBNET_COUNT")
+  # assert custody_subnet_count <= DATA_COLUMN_SIDECAR_SUBNET_COUNT
+  if not (custody_subnet_count <= DATA_COLUMN_SIDECAR_SUBNET_COUNT):
+    return err("Eip7594: Custody subnet count exceeds the DATA_COLUMN_SIDECAR_SUBNET_COUNT")
+
+  var subnet_ids: HashSet[uint64]
```
tersec: Typically try to combine these, e.g.
```nim
var
  subnet_ids: HashSet[uint64]
  current_id = node_id
```

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+  var current_id = node_id
+
+  while subnet_ids.len < int(custody_subnet_count):
+
```
tersec: not sure why empty line

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+    return err("Eip7594: Custody subnet count exceeds the DATA_COLUMN_SIDECAR_SUBNET_COUNT")
+
+  var subnet_ids: HashSet[uint64]
+  var current_id = node_id
+
+  while subnet_ids.len < int(custody_subnet_count):
+  
+    # var subnet_id_bytes: seq[byte]
+    let subnet_id_bytes = eth2digest(current_id.toBytesLE().toOpenArray(0,8))
+    var subnet_id = bytes_to_uint64(subnet_id_bytes.data) mod DATA_COLUMN_SIDECAR_SUBNET_COUNT
```
tersec: This is a 94-character line, see https://status-im.github.io/nim-style-guide/formatting.style.html#style-formattingstyle for using 80 character lines when feasible.

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+  var rowIndices: seq[RowIndex] = @[]
+  for i in 0..<sidecar.column.len:
+    rowIndices.add(RowIndex(i))
+
+  # Iterate through the column indices
+  var colIndices: seq[ColumnIndex] = @[]
+  for i in 0..<sidecar.column.len:
+    colIndices.add(sidecar.index * uint64(sidecar.column.len))
+
+  let kzgCommits = sidecar.kzg_commitments.asSeq
```
tersec:
```nim
  let
    kzgCommits = ...
    sidecarCol = ...
    kzgProofs = ...
    res = ...
```
per https://status-im.github.io/nim-style-guide/language.vardecl.html#variable-declaration-languagevardecl ("Group related variables")

tersec: if `res` should be separate, it can be, but the first three are clearly related

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
-#     #         kzg_commitments=block.body.blob_kzg_commitments,
-#     #         kzg_proofs=kzg_proof_of_column,
-#     #         signed_block_header=signed_block_header,
-#     #         kzg_commitments_inclusion_proof=kzg_commitments_inclusion_proof,
-#     #     ))
-#     # return sidecars
-#     discard
+    return err("DataColumnSidecar: inclusion proof not valid")
+
+  ok()
```
tersec: ends with no blank line -- this seems to be something editors vary on. We generally try to avoid this, represented in GitHub's view that red symbol at the end.

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
-
-    # assert len(subnet_ids) == len(set(subnet_ids))
-    if not (subnet_ids.len == subnet_ids.len):
-        return err("Eip7594: Subnet ids are not unique")
-
-    # columns_per_subnet = NUMBER_OF_COLUMNS // DATA_COLUMN_SIDECAR_SUBNET_COUNT
-    let columns_per_subnet = NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
+# https://github.com/ethereum/consensus-specs/blob/5f48840f4d768bf0e0a8156a3ed06ec333589007/specs/_features/eip7594/das-core.md#recover_matrix    
+proc recover_matrix*(cells_dict: Table[(BlobIndex, CellID), KzgCell], blobCount: uint64): Result[ExtendedMatrix, cstring] =
+
```
tersec: avoid empty line here

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+  # Check is the sidecar column length == sidecar.kzg_commitments length == sidecar.kzg_proofs mixInLength
+  if not (sidecar.column.len == sidecar.kzg_commitments.len):
+    return err("EIP7594: Data column sidecar length is not equal to the kzg_commitments length")
+
+  if not (sidecar.kzg_commitments.len == sidecar.kzg_proofs.len):
+    return err("EIP7594: Data column sidecar kzg_commitments length is not equal to the kzg_proofs length")
+
+  # Iterate through the row indices
+  var rowIndices: seq[RowIndex] = @[]
```
tersec:
```suggestion
  var rowIndices: seq[RowIndex]
```
either is ok, depends how explicit one wants to be, but it's guaranteed that the default value of a `seq` is the empty sequence

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+  for i in 0..<sidecar.column.len:
+    rowIndices.add(RowIndex(i))
+
+  # Iterate through the column indices
+  var colIndices: seq[ColumnIndex] = @[]
```
tersec: see above, regarding default `seq` value

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+  for i in 0..<sidecar.column.len:
```
tersec:
```suggestion
  for _ in 0..<sidecar.column.len:
```
the loop index isn't used, can use `_` to make this clear and improve efficiency somewhat

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+    colIndices.add(sidecar.index * uint64(sidecar.column.len))
```
tersec:
```suggestion
    colIndices.add(sidecar.index * sidecar.column.lenu64)
```

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+    var cellIds: seq[CellID] = @[]
+    var blIdx: BlobIndex
+    var cellId: CellID
+    let key = (blIdx, cellId)
+
+    for key, cell in pairs(cells_dict):
+      if blIdx == blobIndex:
+        cellIds.add(cellId)
+
+    var cells: seq[KzgCell] = @[]
```
tersec: usual, the ` = @[]` is not strictly necessary, to taste though

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+      
+      if cells_dict.hasKey(interim_key):
+        try:
+          let cell = cells_dict[interim_key]
+          cells.add(cell)
+        except:
+          debug "DataColumn: Key not found in Cell Dictionary", interim_key
+
+    let allCellsForRow = recoverAllCells(ctx, cellIds, cells)
+    discard extended_matrix.add(allCellsForRow.get())
```
tersec: why is the return value being discarded? This generally (not always) indicates either missing error handling or an API which doesn't actually need to return that value to begin with

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +32,205 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+          debug "DataColumn: Key not found in Cell Dictionary", interim_key
+
+    let allCellsForRow = recoverAllCells(ctx, cellIds, cells)
+    discard extended_matrix.add(allCellsForRow.get()) 
+
+  ok(extended_matrix)
+
+# https://github.com/ethereum/consensus-specs/blob/5f48840f4d768bf0e0a8156a3ed06ec333589007/specs/_features/eip7594/das-core.md#get_data_column_sidecars
+proc get_data_column_sidecars*(signed_block: deneb.SignedBeaconBlock, blobs: seq[KzgBlob]): Result[seq[DataColumnSidecar], cstring] =
+
```
tersec: rm empty line at start of function

beacon_chain/spec/eip7594_helpers.nim
tersec: https://nim-lang.org/docs/manual.html#statements-and-expressions-special-identifier-nim-underscore

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,83 @@
...
+
+func toUInt64(s: SomeInteger): Opt[uint64] =
+  if s < 0:
+    return Opt.none uint64
+  try:
+    Opt.some uint64(s)
+  except ValueError:
+    Opt.none uint64
+
+func fromHex[N: static int](s: string): Opt[array[N, byte]] =
```
tersec, line 34: There's not some other `fromHex` already available? this function looks very familiar

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -540,47 +564,47 @@ proc validateDataColumnSidecar*(
...
-    # to the finalized checkpoint (else it wouldn't be in the DAG)
-    return errIgnore("DataColumnSidecar: Can't find ancestor")
-
-  if not (
-      finalized_checkpoint.root == ancestor.root or
-      finalized_checkpoint.root.isZero):
-    quarantine[].addUnviable(block_root)
-    return dag.checkedReject(
-      "DataColumnSidecar: Finalized checkpoint not an ancestor")
+  # let
```
tersec, line 593: in general, for anything mergeable, it's usually preferable not to have large blocks of commented-out code. but so long as this part won't be merged, it's not particularly important, whatever works best for this branch.

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -20,6 +29,8 @@ type
   Coset* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
   CosetEvals* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
   Cell* = KzgCell
```
tersec, line 31: Is `Cell` used anywhere, or should it be, in place of `KzgCell`?

beacon_chain/gossip_processing/gossip_validation.nim
agnxsh: Whenever this will be merged, it can be uncommented out. This is just for Kurtosis, so that I can localise if any errors around the newer gossip validation rules.

beacon_chain/spec/datatypes/eip7594.nim
agnxsh: Well, using `Cell` should be used ideally and consistently everywhere, I probably used `KzgCell` as as experimental debug strategy to locate the buggy non-upstreamed funcs in c-kzg

agnxsh: Can definitely fix this

tests/consensus_spec/test_fixture_networking.nim
agnxsh: Well this fromHex returns any Option for me to check if the YAML parser parsed things properly using the `isNone` function, now this function does exist in the kzg fixture test file too, but I wrote this `fromHex` separately to keep the imports clean enough, so that every same dir test file has an independent set of import paths

tersec: ok

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +27,209 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+# https://github.com/ethereum/consensus-specs/blob/5f48840f4d768bf0e0a8156a3ed06ec333589007/specs/_features/eip7594/das-core.md#recover_matrix    
+proc recover_matrix*(cells_dict: Table[(BlobIndex, CellID), Cell], blobCount: uint64): Result[ExtendedMatrix, cstring] =
+  # This helper demonstrates how to apply recover_all_cells
+  # The data structure for storing cells is implementation-dependent
+
+  var extended_matrix: ExtendedMatrix
+
+  for blobIndex in 0'u64..<blobCount:
+    var cellIds: seq[CellID] = @[]
```
tersec: Would typically group these kinds of things as:
```nim
    var
      cellIds: seq[CellID] = @[]
      blIdx: BlobIndex
      cellId: CellID
```

it's three related `var`s next to each other

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -661,103 +27,209 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
...
+  var blck = signed_block.message
+
+  var cellsAndProofs: seq[KzgCellsAndKzgProofs] = @[]
+
+  for blob in blobs:
+    let
+      computed_cell = computeCellsAndKzgProofs(blob)
+
+    if computed_cell.isErr():
+      fatal "EIP7549: Could not compute cells"
```
tersec: Apparently not a very fatal `fatal`? It might be an `error`, but evidently it's not `fatal`

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,83 @@
...
+          check column == uint64(reslt[count])
+          count = count + 1
+
+suite "EF - EIP7594 - Networking" & preset():
+  const presetPath = SszTestsDir/const_preset
+  let basePath =
+    presetPath/"eip7594"/"networking"/"get_custody_columns"/"pyspec_tests"
+  for kind, path in walkDir(basePath, relative = true, checkDir = true):
+    runGetCustodyColumns(suiteName, basePath/path)
+
```
tersec: Bunch of odd empty-looking lines at end

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,83 @@
+# beacon_chain
```
tersec, line 1: Is this specifically a 7594-related test? If so, it could be more clearly labeled/named as such and apparently not included in `all_eip7594_fixtures`. I'm not sure if there are other test fixtures you'd intended to at some point add to that or not, or if `consensus-specs` will in general include networking tests so that there will be a more general purpose to something like `test_fixture_networking`, but currently it's not clear to me the purpose of its organization

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -183,9 +193,135 @@ proc runComputeBlobKzgProofTest(suiteName, suitePath, path: string) =
...
+    # blocks does not represent a BLS field element) it should error, i.e. the
+    # the output should be `null`.
+    if blob.isNone:
+      check output.kind == JNull
+    else:
+      let p = computeCells(blob.get)
+      if p.isErr:
+        check output.kind == JNull
+      else:
+        for i in 0..<128:
```
tersec: This 128 is presumably
```nim
  KzgCells* = array[CELLS_PER_EXT_BLOB, KzgCell]
  KzgCellsAndKzgProofs* = object
    cells*: array[CELLS_PER_EXT_BLOB, KzgCell]
    proofs*: array[CELLS_PER_EXT_BLOB, KzgProof]
```
from https://github.com/agnxsh/c-kzg-4844/blob/peerdas-nim/bindings/nim/kzg.nim where `CELLS_PER_EXT_BLOB` is from https://github.com/agnxsh/c-kzg-4844/blob/peerdas-nim/bindings/nim/kzg_abi.nim

But these are `array`s where the `length` is known statically at compile time, so it'd be better to either:
- use `CELLS_PER_EXT_BLOB`, or maybe even better;
- `for i in 0 ..< len(p.get)`

or similar. the `get` is not the most aesthetic, but that's a more generic issue in this code flow

The point is to:
- explain where the 128 comes from (i.e. it's the entire object);
- ensure if that constant ever changes, the loop body changes too; and
- clarify why it won't ever index out of the array

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -183,9 +193,135 @@ proc runComputeBlobKzgProofTest(suiteName, suitePath, path: string) =
...
+    # blocks does not represent a BLS field element) it should error, i.e. the
+    # the output should be `null`.
+    if blob.isNone:
+      check output.kind == JNull
+    else:
+      let p = computeCellsAndProofs(blob.get)
+      if p.isErr:
+        check output.kind == JNull
+      else:
+        for i in 0..<128:
```
tersec: https://github.com/status-im/nimbus-eth2/pull/6322/files#r1637485498

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -183,9 +193,135 @@ proc runComputeBlobKzgProofTest(suiteName, suitePath, path: string) =
...
+    # blocks does not represent a BLS field element) it should error, i.e. the
+    # the output should be `null`.
+    if cell_ids.anyIt(it.isNone) or cells.anyIt(it.isNone):
+      check output.kind == JNull
+    else:
+      let v = recoverAllCells(cell_ids.mapIt(it.get), cells.mapIt(it.get))
+      if v.isErr:
+        check output.kind == JNull
+      else:
+        for i in 0..<128:
```
tersec: https://github.com/status-im/nimbus-eth2/pull/6322/files#r1637485498

beacon_chain/spec/eip7594_helpers.nim
agnxsh: my bad, forgot to cleanup, this was just sanity check, perhaps debugEcho is probably the best in this case

tests/consensus_spec/test_fixture_networking.nim
agnxsh: okay so this is `consensus-specs` specific and i think i named it exactly based on that, the dir was indeed named as `networking`, similar to `merkle_proof`, etc. and yes if there are further networking fixture tests, it could easily go in there

agnxsh: I do understand that `eip7594` is sort of hard-coded in the presetPath but that can be altered later, once more eips/specs are tested under this networking `dir`

agnxsh: I used the same rationale as to naming `test_fixture_kzg`, and previously just hardcoding `deneb` in one of the dir paths

tests/consensus_spec/test_fixture_kzg.nim
agnxsh: > This 128 is presumably
> 
> ```nim
>   KzgCells* = array[CELLS_PER_EXT_BLOB, KzgCell]
>   KzgCellsAndKzgProofs* = object
>     cells*: array[CELLS_PER_EXT_BLOB, KzgCell]
>     proofs*: array[CELLS_PER_EXT_BLOB, KzgProof]
> ```
> 
> from https://github.com/agnxsh/c-kzg-4844/blob/peerdas-nim/bindings/nim/kzg.nim where `CELLS_PER_EXT_BLOB` is from https://github.com/agnxsh/c-kzg-4844/blob/peerdas-nim/bindings/nim/kzg_abi.nim
> 
> But these are `array`s where the `length` is known statically at compile time, so it'd be better to either:
> 
> * use `CELLS_PER_EXT_BLOB`, or maybe even better;
> * `for i in 0 ..< len(p.get)`
> 
> or similar. the `get` is not the most aesthetic, but that's a more generic issue in this code flow
> 
> The point is to:
> 
> * explain where the 128 comes from (i.e. it's the entire object);
> * ensure if that constant ever changes, the loop body changes too; and
> * clarify why it won't ever index out of the array

sure getting it fixed, sometimes, i prefer hardcoding on a unit test based on the fact that it wouldn't break any other place, or maybe if there's some issue importing a constant, because it's defined in the C layer, which in this case isn't so okay can fix this

beacon_chain/spec/eip7594_helpers.nim
agnxsh: fixed

tersec:
Been overtaken by events.

PR #6329, opened 2024-06-07
VC/NIMBUS-BN validators protection.

cheatfate:


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -393,19 +394,26 @@ proc processAttestation*(
 
     let (attester_index, sig) = v.get()
 
-    self[].checkForPotentialDoppelganger(attestation, [attester_index])
+    if checkValidator and (attester_index in self.validatorPool[]):
+      warn "A validator client attempts to send an attestation from " &
```
```suggestion
      warn "A validator client has attempted to send an attestation from " &
```

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -393,19 +394,26 @@ proc processAttestation*(
...
+           "validator that is also manager by beacon node",
```
```suggestion
           "validator that is also managed by the beacon node",
```

beacon_chain/gossip_processing/eth2_processor.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6329/commits/b53f5bdbc43d0fc4308ce48f748e235c2e530670

beacon_chain/gossip_processing/eth2_processor.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6329/commits/b53f5bdbc43d0fc4308ce48f748e235c2e530670

beacon_chain/validator_client/api.nim
```diff
@@ -1597,10 +1678,10 @@ proc submitPoolSyncCommitteeSignature*(
         of 200:
           ApiResponse[bool].ok(true)
         of 400:
-          handle400()
+          handle400Indexed()
```
arnetheduck, line 1621: we should not have this checking on the sync committee

cheatfate: This template and call is supposed to fix error decoding process, because both calls:
https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolAttestations and https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolSyncCommitteeSignatures

are using [IndexedErrorMessage](https://github.com/ethereum/beacon-APIs/blob/3132da1c570a72792c0a5e4d0b412a856fbd4854/types/http.yaml#L152-L178) instead of [ErrorMessage](https://github.com/ethereum/beacon-APIs/blob/3132da1c570a72792c0a5e4d0b412a856fbd4854/types/http.yaml#L136-L151)

beacon_chain/validator_client/api.nim
```diff
@@ -1522,7 +1549,23 @@ proc submitPoolAttestations*(
         of 200:
           ApiResponse[bool].ok(true)
         of 400:
-          handle400()
+          # We check for Nimbus specific error which beacon node could produce
+          # if it manages same validator as we do.
+          let idxerr = response.unpackErrorMessage()
+          for item in idxerr.failures:
+            if "NIMBUS#E0001" in item.message:
+              inc(doppelgangerCount)
```
tersec: In general, should the VC even keep running in this situation? I guess one point here is that this `NIMBUS#E0001` error isn't part of the actual Beacon API, but a side channel via the textual response, which means that in theory it would be legal for a BN to respond this way regardless, and that would shut down the VC. But that would be essentially malicious and we can have enough `fatal`/`error`/`notice`/`info` (i.e. default) visible logging to be able to attribute such malice and work around if it someone does this.

But when it happens even once that means there's a misconfiguration and probably it would be better for the VC to just stop in a similar way as the BN does when it detects a doppelganger. Historically, it's not usually just been a small overlap between validators, but fallout from a more systemic mistake in allocating validator keys across Nimbus BN/VC instances.

beacon_chain/validator_client/block_service.nim
```diff
@@ -279,6 +279,9 @@ proc publishBlockV3(vc: ValidatorClientRef, currentSlot, slot: Slot,
           except CancelledError as exc:
             debug "Block publication has been interrupted"
             raise exc
+          except ValidatorApiCriticalError:
+            vc.doppelExit.fire()
```
tersec: Is `ValidatorApiCriticalError` essentially defined to be coterminous by design with doppelganger detection? Because if so, then it's a bit generically named. If the latter, then one can't use this kind of logic unconditionally with it.

beacon_chain/validator_client/api.nim
```diff
@@ -2284,10 +2368,23 @@ proc publishBlock*(
           return true
         of 202:
           debug BlockBroadcasted, node = node,
-           blck = shortLog(ForkedSignedBeaconBlock.init(data))
+                blck = shortLog(ForkedSignedBeaconBlock.init(data))
           return true
         of 400:
-          handle400()
+          let emsg = response.getErrorMessage()
+          if "NIMBUS#E0001" in emsg:
```
tersec: This exact `NIMBUS#E0001` string is copied a few times. Looks typo-prone (exact number of zeros, etc), so maybe better to define it as constant somewhere

beacon_chain/validator_client/block_service.nim
cheatfate: `ValidatorApiCriticalError` is an error which should bring us to VC shutdown. `doppelExit` event was renamed by @arnetheduck (initially it was named as `gracefulExit`) and it was designed to allow some parts of VC to request shutdown.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -366,7 +366,8 @@ proc checkForPotentialDoppelganger(
 proc processAttestation*(
     self: ref Eth2Processor, src: MsgSource,
     attestation: phase0.Attestation | electra.Attestation, subnet_id: SubnetId,
-    checkSignature: bool = true): Future[ValidationRes] {.async: (raises: [CancelledError]).} =
+    checkSignature: bool = true, checkValidator: bool = false
```
tersec: Perhaps this would work better as a non-optional parameter. `checkSignature` is "safe" to default to on; it just wastes a bit of time if it's truly not needed.

But `checkValidator`
- appears to be essentially the bypass mechanism very specifically for doppel-sent-by-BN, i.e. so the BN doesn't develop an autoimmune sort of doppelganger detection, so is less general-purpose than its name suggests; and
- more important, isn't something there's a good sensible default for. Every caller of `processAttestation` should consider specifically what to do here

So perhaps better for it not to be an optional parameter.

The problem there is that it has the same type as `checkSignature` though, so it could get confused with it (i.e. if something calls `processAttestation(foo, bar, true)` or `processAttestation(foo, bar, false)` they currently mean to control signature checking, not doppel checking)

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -393,19 +394,26 @@ proc processAttestation*(
 
     let (attester_index, sig) = v.get()
 
-    self[].checkForPotentialDoppelganger(attestation, [attester_index])
+    if checkValidator and (attester_index in self.validatorPool[]):
+      warn "A validator client has attempted to send an attestation from " &
+           "validator that is also managed by the beacon node",
+           validator_index = attester_index
+      errReject("An attestation could not be sent from a validator that is " &
+                "also managed by the beacon node. (NIMBUS#E0001)")
```
tersec: third instance of magic string literal

beacon_chain/validator_client/block_service.nim
tersec: Are there other possible `ValidatorApiCriticalError` examples which are not doppel?

beacon_chain/validator_client/api.nim
cheatfate: Removed.

beacon_chain/validator_client/block_service.nim
cheatfate: Removed.

beacon_chain/gossip_processing/eth2_processor.nim
cheatfate: magic string removed.

beacon_chain/validator_client/api.nim
cheatfate: Removed.

beacon_chain/gossip_processing/eth2_processor.nim
cheatfate: optional parameters removed.

beacon_chain/validators/validator_pool.nim
```diff
@@ -93,6 +93,7 @@ type
 
   ValidatorPool* = object
     validators*: Table[ValidatorPubKey, AttachedValidator]
+    indexSet*: HashSet[ValidatorIndex]
```
tersec, line 96: In theory, it should be an O(1) lookup to use `ValidatorIndex` to get a `ValidatorPubKey` and check `ValidatorPool.validators`, or are there advantages to this `indexSet`? It seems to be maintained in a synchronized way to `validators`.

cheatfate: If you talking about ValidatorIndex <-> ValidatorKey bloom filter table. Than this approach will require at least 2*O(1) lookups instead of O(1) lookup in current approach, also i have not find any references/documentation to that table, so i decided to use hashset which brings us to single check for O(1) time and does not require bringing bloom filter to the light.

tersec: No bloom filter. Looking up a `ValidatorPubKey` from a `ValidatorIndex` is one normal indexed `seq` lookup on a state. The VC might not have a state, but this is BN functionality, and it does have a head state.

Maybe there's a reason to prefer `ValidatorPool` to contain this information, but I'm not sure what this is.

cheatfate: So to get `ValidatorIndex` first i need to get corresponding state, after that i should get `ValidatorPubKey` after that i should query `ValidatorPool` to get `AttachedValidator` object and after that i should check for `index` being set.
So this is actually `O(1) + O(1) + branch` time.
Or i can do `O(1)` and obtain information i need.

beacon_chain/validators/validator_pool.nim
```diff
@@ -223,10 +224,24 @@ func contains*(pool: ValidatorPool, pubkey: ValidatorPubKey): bool =
...
+proc contains*(pool: ValidatorPool, index: ValidatorIndex): bool =
+  ## Returns ``true`` if validator with index ``index`` present in ``pool``.
+  pool.indexSet.contains(index)
+
+proc setValidatorIndex*(pool: var ValidatorPool, validator: AttachedValidator,
+                        index: ValidatorIndex) =
+  pool.indexSet.incl(index)
+  validator.index = Opt.some(index)
+
+proc removeValidatorIndex*(pool: var ValidatorPool, index: ValidatorIndex) =
```
tersec:
```suggestion
proc removeValidatorIndex(pool: var ValidatorPool, index: ValidatorIndex) =
```

beacon_chain/validators/validator_pool.nim
tersec: Sure, why not

beacon_chain/validators/validator_pool.nim
```diff
@@ -223,10 +224,24 @@ func contains*(pool: ValidatorPool, pubkey: ValidatorPubKey): bool =
   ## Returns ``true`` if validator with key ``pubkey`` present in ``pool``.
   pool.validators.contains(pubkey)
 
+proc contains*(pool: ValidatorPool, index: ValidatorIndex): bool =
```
tersec, line 227: Is this used anywhere?

beacon_chain/validators/validator_pool.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6329/commits/e628336bb0a0496aa0e5b9cb5b85c22e246f6211

tersec (approved):

PR #6354, opened 2024-06-14
bump nim-libp2p to `v1.2.1-rc`

etan-status:
- chore(README): small PRs
- Send IDONTWANT before validating message
- chore(peer-scoring): enhance score trace logs (x2)
- chore: fix typos
- fix(multicodec): remove unnecessary "!=" operator


tersec:
Will probably hold this one until after the release commit has been chosen. Probably safe, looks small, but also libp2p is typically one of the riskier bumps

PR #6358, opened 2024-06-14
EIP7594: p2p-interface

agnxsh:


beacon_chain/sync/sync_protocol.nim
```diff
@@ -368,6 +391,110 @@ p2pProtocol BeaconSync(version = 1,
...
+
+        await response.writeBytesSZ(
+          uncompressedLen, bytes,
+          peer.network.forkDigestAtEpoch(blockRef.slot.epoch).data)
+        inc found
+
+    debug "Data column root request done",
+      peer, roots = columnIds.len, count, found
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.2/specs/_features/eip7594/p2p-interface.md#datacolumnsidecarsbyrange-v1
```
tersec, line 438: In general, checking for that all of these are consistent with `v1.5.0-alpha.3` and updating the links accordingly would be worthwhile, because the next PeerDAS devnet will be based on this version:
2024-06-20T23:47:08,583161972+00:00

beacon_chain/sync/sync_protocol.nim
```diff
@@ -24,6 +24,8 @@ const
     ## Allow syncing ~64 blocks/sec (minus request costs)
   blobResponseCost = allowedOpsPerSecondCost(1000)
     ## Multiple can exist per block, they are much smaller than blocks
+  dataColumnResponseCost = allowedOpsPerSecondCost(1000)
+    ## 1 blob has an equivalent memory of 8 data columns
```
tersec, line 28: Wonder if it makes sense to scale the cost somewhat appropriately -- there's overhead which means 8x is too much, because packets/requests/etc per second is also a cost, and to the extent there's just local lookup cost/latency which is a fixed lower bound, that also doesn't scale. But more than `blobResponseCost` probably would be reasonable

beacon_chain/sync/sync_protocol.nim
```diff
@@ -37,6 +39,7 @@ type
 
   BlockRootsList* = List[Eth2Digest, Limit MAX_REQUEST_BLOCKS]
   BlobIdentifierList* = List[BlobIdentifier, Limit (MAX_REQUEST_BLOB_SIDECARS)]
```
tersec:
```suggestion
  BlobIdentifierList* = List[BlobIdentifier, Limit MAX_REQUEST_BLOB_SIDECARS]
```

beacon_chain/sync/sync_protocol.nim
```diff
@@ -37,6 +39,7 @@ type
...
+  DataColumnIdentifierList* = List[DataColumnIdentifier, Limit (MAX_REQUEST_DATA_COLUMNS)]
```
tersec:
```suggestion
  DataColumnIdentifierList* = List[DataColumnIdentifier, Limit MAX_REQUEST_DATA_COLUMNS]
```

beacon_chain/sync/sync_protocol.nim
```diff
@@ -368,6 +391,110 @@ p2pProtocol BeaconSync(version = 1,
...
+      count = columnIds.len
+
+    var
+      found = 0
+      bytes: seq[byte]
+
+    for i in 0..<count:
+      let blockRef = dag.getBlockRef(columnIds[i].block_root).valueOr:
+        continue
+      let index = columnIds[i].index
```
tersec, line 417: Might be worth enforcing:
> If any root in the request content references a block earlier than `minimum_request_epoch`, peers MAY respond with error code `3: ResourceUnavailable` or not include the data column sidecar in the response.

from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.3/specs/_features/eip7594/p2p-interface.md#datacolumnsidecarsbyroot-v1 here

beacon_chain/sync/sync_protocol.nim
```diff
@@ -368,6 +391,110 @@ p2pProtocol BeaconSync(version = 1,
...
+          GENESIS_EPOCH
+        else:
+          dag.head.slot.epoch - dag.cfg.MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS
+    
+    if startSlot.epoch < epochBoundary:
+      raise newException(ResourceUnavailableError, DataColumnsOutOfRange)
+
+    var blockIds: array[int(MAX_REQUEST_DATA_COLUMNS), BlockId]
+    let
+      count = int min(reqCount, blockIds.lenu64)
```
tersec, line 467: If feasible, it's better to avoid these conversions without some kind of protection against crashing on out-of-range `uint64` to `int` conversions with a `RangeDefect`:
```
Error: unhandled exception: value out of range [RangeDefect]
```

If it's provably not hittable (e.g., blockIds.lenu64 is bounded in a sensible way), that should be noted

In this case, `blockIds.lenu64 == MAX_REQUESTS_DATA_COLUMNS`, so that part is compile-time known. One approach there might be to wrap that in `static(blockIds.lenu64)` to add a kind of compile-time assertion of this much, that it's not really a variable, but is known ahead of time.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -368,6 +391,110 @@ p2pProtocol BeaconSync(version = 1,
...
+        dag.getBlockRange(startSlot, 1, blockIds.toOpenArray(0, endIndex))
+
+    var
+      found = 0
+      bytes: seq[byte]
+    
+    for i in startIndex..endIndex:
+      for j in 0..<MAX_REQUEST_DATA_COLUMNS:
+        if dag.db.getDataColumnSidecarSZ(blockIds[i].root, ColumnIndex(j), bytes):
+          if blockIds[i].slot.epoch >= dag.cfg.BELLATRIX_FORK_EPOCH and
```
tersec, line 479: Why is `blockIds[i].slot.epoch >= dag.cfg.BELLATRIX_FORK_EPOCH` being checked? It's neither sufficient nor (I think?) necessary.

It's not sufficient because the Bellatrix fork isn't coterminous with the merge -- the merge happened within the fork, not at the beginning. And it shouldn't be necessary, because `getDataColumnSidecarSZ` has already been called. It should not be possible for that to have content in a pre-Electra fork.

If it is worth checking this as a sanity check, maybe check against when PeerDAS actually starts, or at least bounding it at `ELECTRA_FORK_EPOCH` rather than `BELLATRIX_FORK_EPOCH`.

There's another oddity here, which is that it makes this the `ByRange` inconsistent in its checks with the `ByRoot` rep/resp call. Both have this information about the slots/epochs, and could perform the same check, and should allow/disallow based on this similarly.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -368,6 +391,110 @@ p2pProtocol BeaconSync(version = 1,
...
+              not dag.head.executionValid:
```
tersec, line 480: Why is this checked in `ByRange` but not `ByRoot`?

beacon_chain/sync/sync_protocol.nim
```diff
@@ -368,6 +391,110 @@ p2pProtocol BeaconSync(version = 1,
...
+      peer: Peer,
+      columnIds: DataColumnIdentifierList,
+      response: MultipleChunksResponse[
+        ref DataColumnSidecar, Limit(MAX_REQUEST_DATA_COLUMNS)])
+      {.async, libp2pProtocol("data_column_sidecars_by_root", 1).} =
+
+    trace "got data columns range request", peer, len = columnIds.len
+    if columnIds.len == 0:
+      raise newException(InvalidInputsError, "No data columns requested")
+
```
tersec, line 405: Should check here for
> No more than `MAX_REQUEST_DATA_COLUMN_SIDECARS` may be requested at a time.

from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.3/specs/_features/eip7594/p2p-interface.md#datacolumnsidecarsbyroot-v1 in which case the request is also invalid

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +467,120 @@ proc requestManagerBlobLoop(
...
+    if columnIds.len > 0:
+      debug "Requesting detected missing data columns", columns = shortLog(columnIds)
+      let start = SyncMoment.now(0)
+      var workers:
+        array[PARALLEL_REQUESTS, Future[void].Raising([CancelledError])]
+      for i in 0..<PARALLEL_REQUESTS:
+        workers[i] = rman.fetchDataColumnsFromNetwork(columnIds)
+      
+      await allFutures(workers)
+      let finish = SyncMoment.now(uint64(len(columnIds)))
```
tersec, line 628: Might be worth moving this into the `debug` statement so it's not run if it's not used; it's possible (albeit slightly pathological) for this to be a slow operation

beacon_chain/sync/sync_manager.nim
```diff
@@ -80,6 +80,7 @@ type
   BeaconBlocksRes =
     NetRes[List[ref ForkedSignedBeaconBlock, Limit MAX_REQUEST_BLOCKS]]
   BlobSidecarsRes = NetRes[List[ref BlobSidecar, Limit(MAX_REQUEST_BLOB_SIDECARS)]]
+  DataColumnSidecarsRes = NetRes[List[ref DataColumnSidecar, Limit(MAX_REQUEST_DATA_COLUMNS)]]
```
tersec: 93-character line

beacon_chain/sync/sync_protocol.nim
tersec: but also still a fairly long line

beacon_chain/consensus_object_pools/block_quarantine.nim
```diff
@@ -336,3 +338,16 @@ func popBlobless*(
...
+func popColumnless*(
+    quarantine: var Quarantine,
+    root: Eth2Digest): Opt[ForkedSignedBeaconBlock] =
+  var blck: ForkedSignedBeaconBlock
+  if quarantine.columnless.pop(root, blck):
+    Opt.some(blck)
+  else:
+    Opt.none(ForkedSignedBeaconBlock)
+
+iterator peekColumnless*(quarantine: var Quarantine): ForkedSignedBeaconBlock =
```
tersec, line 387: Why does this take a `var Quarantine` rather than `Quarantine`? `peekColumnless` shouldn't be changing it

beacon_chain/beacon_node.nim
```diff
@@ -22,7 +22,7 @@ import
   ./el/el_manager,
   ./consensus_object_pools/[
     blockchain_dag, blob_quarantine, block_quarantine, consensus_manager,
-    attestation_pool, sync_committee_msg_pool, validator_change_pool],
+    data_column_quarantine, attestation_pool, sync_committee_msg_pool, validator_change_pool],
```
tersec: 92 char line

beacon_chain/beacon_chain_db.nim
```diff
@@ -254,6 +254,13 @@ func blobkey(root: Eth2Digest, index: BlobIndex) : array[40, byte] =
 
   ret
 
+func columnkey(root: Eth2Digest, index: ColumnIndex) : array[40, byte] =
```
```suggestion
func columnkey(root: Eth2Digest, index: ColumnIndex): array[40, byte] =
```

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1410,6 +1411,24 @@ proc pruneBlobs(node: BeaconNode, slot: Slot) =
               count = count + 1
     debug "pruned blobs", count, blobPruneEpoch
 
+proc pruneDataColumns(node: BeaconNode, slot: Slot) =
+  let dataColumnPruneEpoch = (slot.epoch -
```
tersec, line 1458: This seems to underflow if called too early to genesis? Even if currently, no networks have launched with PeerDAS at genesis, presumably at some point there will be an Electra genesis, PeerDAS network just as there are now Deneb genesis, KZG-launched networks.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1444,6 +1463,7 @@ proc onSlotEnd(node: BeaconNode, slot: Slot) {.async.} =
       # the pruning for later
       node.dag.pruneHistory()
       node.pruneBlobs(slot)
+      node.pruneDataColumns(slot)
```
tersec: And here, it will trigger the underflow in a new devnet. This code won't, I think, be hit during syncing, so existing testnets won't have a problem, but new devnets will

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +467,120 @@ proc requestManagerBlobLoop(
...
+    wallSlot = wallTime.slotOrZero()
+    delay = wallTime - wallSlot.start_beacon_time()
+    waitDur = TimeDiff(nanoseconds: DATA_COLUMN_GOSSIP_WAIT_TIME_NS)
+
+  var
+    fetches: seq[DataColumnIdentifier]
+    ready: seq[Eth2Digest]
+  for columnless in rman.quarantine[].peekColumnless():
+    withBlck(columnless):
+      when consensusFork >= ConsensusFork.Deneb:
```
tersec, line 544: Is this being Deneb a temporary quirk of the PeerDAS activation epoch/fake fork mess or intended to be more permanent?

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +467,120 @@ proc requestManagerBlobLoop(
...
+          if len(missing.indices) == 0:
+            warn "quarantine is missing data columns, but missing indices are empty",
+             blk = columnless.root,
+             commitments = len(forkyBlck.message.body.blob_kzg_commitments)
+          for idx in missing.indices:
+            let id = DataColumnIdentifier(block_root: columnless.root, index: idx)
+            if id notin fetches:
+              fetches.add(id)
+        else:
+          # this is a programming error and it should occur
```
```suggestion
          # this is a programming error and it should not occur
```

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +259,46 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+        return
+
+      for col in ucolumns:
+        rman.dataColumnQuarantine[].put(col)
+      var curRoot: Eth2Digest
+      for col in ucolumns:
+        let block_root = hash_tree_root(col.signed_block_header.message)
+        if block_root != curRoot:
+          curRoot = block_root
+          if (let o = rman.quarantine[].popColumnless(curRoot); o.isSome):
```
tersec: This can probably use `valueOr` or a similar construction to avoid the `isSome`/`unsafeGet` pair and be more statically verifiably safe

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +259,46 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+                                 colIdList: seq[DataColumnIdentifier])
+                                 {.async: (raises: [CancelledError]).} =
+  var peer: Peer
+
+  try:
+    peer = await rman.network.peerPool.acquire()
+    debug "Requesting data columns by root", peer = peer, columns = shortLog(colIdList),
+                                                     peer_score = peer.getScore()
+    
+    let columns = await dataColumnSidecarsByRoot(peer, DataColumnIdentifierList colIdList)
```
tersec: In theory can an use `valueOr` here (let columns = await foo().valueOr:) and then avoid the `if columns.isOk`/`let ucolumns = columns.get` construction

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +259,46 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+      for col in ucolumns:
+        let block_root = hash_tree_root(col.signed_block_header.message)
+        if block_root != curRoot:
+          curRoot = block_root
+          if (let o = rman.quarantine[].popColumnless(curRoot); o.isSome):
+            let col = o.unsafeGet()
+            discard await rman.blockVerifier(col, false)
+
+    else:
+      debug "Data Columns by root request failed",
```
```suggestion
      debug "Data columns by root request failed",
```

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +467,120 @@ proc requestManagerBlobLoop(
             blobs_count = len(blobIds),
             sync_speed = speed(start, finish)
 
+proc getMissingDataColumns(rman: RequestManager): seq[DataColumnIdentifier] =
+  let
+    wallTime = rman.getBeaconTime()
+    wallSlot = wallTime.slotOrZero()
+    delay = wallTime - wallSlot.start_beacon_time()
+    waitDur = TimeDiff(nanoseconds: DATA_COLUMN_GOSSIP_WAIT_TIME_NS)
```
tersec: This can be `const`

beacon_chain/sync/sync_protocol.nim
```diff
@@ -109,6 +112,26 @@ proc readChunkPayload*(
...
+    Future[NetRes[MsgType]] {.async: (raises: [CancelledError]).} =
+  var contextBytes: ForkDigest
+  try:
+    await conn.readExactly(addr contextBytes, sizeof contextBytes)
+  except CancelledError as exc:
+    raise exc
+  except CatchableError:
+    return neterr UnexpectedEOF
+
+  if contextBytes == peer.network.forkDigests.deneb:
```
tersec, line 126: `peer.network.forkDigests.deneb`, is this a temporary hack/spec workaround or how it's supposed to work?

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,103 @@
...
+func shortLog*(x: seq[ColumnIndex]): string =
+  "<" & x.mapIt($it).join(", ") & ">"
+
+func shortLog*(x: seq[DataColumnFetchRecord]): string =
+  "[" & x.mapIt(shortLog(it.block_root) & shortLog(it.indices)).join(", ") & "]"
+
+func put*(quarantine: var DataColumnQuarantine, dataColumnSidecar: ref DataColumnSidecar) =
+  if quarantine.data_columns.lenu64 >= MaxDataColumns:
+    # FIFO if full. For example, sync manager and request manager can race to
+    # put blobs in at the same time, so one gets blob insert -> block resolve
```
tersec: The comment block here still discusses blobs

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +467,120 @@ proc requestManagerBlobLoop(
...
+          # this is a programming error and it should occur
+          warn "missing data column handler found columnless block with all data columns",
+             blk = columnless.root,
+             commitments=len(forkyBlck.message.body.blob_kzg_commitments)
+          ready.add(columnless.root)
+  
+  for root in ready:
+    let columnless = rman.quarantine[].popColumnless(root).valueOr:
+      continue
+    discard rman.blockVerifier(columnless, false)
```
tersec, line 570: This is `discard` because none of the error conditions are relevant? i.e. `MissingParent` is implied to some extent by it being in a kind of quarantine, `UnviableFork` and `Invalid` should have been already caught, etc?

beacon_chain/sync/request_manager.nim
agnxsh: temporary fork

beacon_chain/sync/sync_protocol.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/e2afc583cb1bf31a995c6c6bfa246561e507291c

beacon_chain/sync/sync_protocol.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/e2afc583cb1bf31a995c6c6bfa246561e507291c

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -404,6 +445,63 @@ proc checkForPotentialDoppelganger(
...
+    data_column_sidecars.add data_column[]
+    storedColumns.add data_column.index
+
+    if columnsOk:
+      debug "Loaded data column for reconstruction"
+
+  # storedColumn number is less than the NUMBER_OF_COLUMNS
+  # then reconstruction is not possible, and if all the data columns
+  # are already stored then we do not need to reconstruct at all
+  if storedColumns.len < NUMBER_OF_COLUMNS or storedColumns.len == NUMBER_OF_COLUMNS:
```
tersec: Does this know if they're valid?

beacon_chain/sync/sync_protocol.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6358/commits/93c3525955b841699fbd2f9d77b72595c0100e92

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1104,6 +1130,11 @@ proc addCapellaMessageHandlers(
   node.addAltairMessageHandlers(forkDigest, slot)
   node.network.subscribe(getBlsToExecutionChangeTopic(forkDigest), basicParams)
 
+proc fetchCustodySubnetCount* (res: var uint64, node: BeaconNode) =
```
tersec: Generally, the `node: BeaconNode` argument/parameter should be first here

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -830,7 +900,7 @@ proc processBlock(
 
   let res = withBlck(entry.blck):
     await self.storeBlock(
-      entry.src, wallTime, forkyBlck, entry.blobs, entry.maybeFinalized,
+      entry.src, wallTime, forkyBlck, entry.blobs, entry.data_columns, entry.maybeFinalized,
```
tersec: 93 character line

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -809,7 +879,7 @@ proc addBlock*(
   # - RequestManager (missing ancestor blocks)
   # - API
   let resfut = newFuture[Result[void, VerifierError]]("BlockProcessor.addBlock")
-  enqueueBlock(self, src, blck, blobs, resfut, maybeFinalized, validationDur)
+  enqueueBlock(self, src, blck, blobs, data_columns, resfut, maybeFinalized, validationDur)
```
tersec: 92 char line

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -842,7 +912,7 @@ proc processBlock(
     # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.2/sync/optimistic.md#execution-engine-errors
     await sleepAsync(chronos.seconds(1))
     self[].enqueueBlock(
-      entry.src, entry.blck, entry.blobs, entry.resfut, entry.maybeFinalized,
+      entry.src, entry.blck, entry.blobs, entry.data_columns, entry.resfut, entry.maybeFinalized,
```
tersec: 97 char line

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,102 @@
...
+    blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock): DataColumnFetchRecord =
+  var indices: seq[ColumnIndex]
+  for i in 0..<len(blck.message.body.blob_kzg_commitments):
+    let idx = ColumnIndex(i)
+    if not quarantine.data_columns.hasKey(
+        (blck.root, idx)):
+      indices.add(idx)
+  DataColumnFetchRecord(block_root: blck.root, indices: indices)
+
+func init*(
```
tersec, line 100: What is this `init` function doing -- is it likely in near future that it will be inadmissible to create a `DataColumnQuarantine` without arguments? It's fine as a placeholder until those are finalized, but when merging into `unstable`, it's cleaner to just use Nim's default mechanisms for this than add an empty shim init function. Just
```nim
  var x: DataColumnQuarantine
```
will also default-initialize this way (any substantive numeric fields are zero, strings empty, ref's not allocated, etc), as will other default-creation approaches.

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,102 @@
...
+    seq[ref DataColumnSidecar] =
+  var r: seq[ref DataColumnSidecar]
+  for idx in 0..<len(blck.message.body.blob_kzg_commitments):
+    var c: ref DataColumnSidecar
+    if quarantine.data_columns.pop((digest, ColumnIndex idx), c):
+      r.add(c)
+  r
+
+func hasDataColumns*(quarantine: DataColumnQuarantine,
+    blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock): bool =
```
tersec, line 84: There's just so much of this temporary workaround for broken devnet-1 fork schedule that's going to become kludgy jank the moment the first few devnets are over

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1410,6 +1441,24 @@ proc pruneBlobs(node: BeaconNode, slot: Slot) =
...
     debug "pruned blobs", count, blobPruneEpoch
 
+proc pruneDataColumns(node: BeaconNode, slot: Slot) =
+  let dataColumnPruneEpoch = (slot.epoch -
+                              node.dag.cfg.MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS - 1)
+  if slot.is_epoch() and dataColumnPruneEpoch >= node.dag.cfg.DENEB_FORK_EPOCH:
+    var blocks: array[SLOTS_PER_EPOCH.int, BlockId]
+    var count = 0
+    let startIndex = node.dag.getBlockRange(
+      dataColumnPruneEpoch.start_slot, 1, blocks.toopenArray(0, SLOTS_PER_EPOCH - 1))
```
tersec, line 1464:
```suggestion
      dataColumnPruneEpoch.start_slot, 1, blocks.toOpenArray(0, SLOTS_PER_EPOCH - 1))
```

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -28,9 +36,8 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex, subnetIds: HashSet[uint
   res.sort()
   res
 
-# https://github.com/ethereum/consensus-specs/blob/5f48840f4d768bf0e0a8156a3ed06ec333589007/specs/_features/eip7594/das-core.md#get_custody_columns
-proc get_custody_columns*(node_id: NodeId, custody_subnet_count: uint64): Result[seq[ColumnIndex], cstring] =
-    
+proc get_custody_column_subnet*(node_id: NodeId, custody_subnet_count: uint64): Result[HashSet[uint64], cstring] =
```
tersec: long line

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -224,21 +322,21 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar): Result[
...
     return err("DataColumnSidecar: validation failed")
 
   ok()
 
 # https://github.com/ethereum/consensus-specs/blob/5f48840f4d768bf0e0a8156a3ed06ec333589007/specs/_features/eip7594/p2p-interface.md#verify_data_column_sidecar_inclusion_proof
-proc verify_data_column_sidecar_inclusion_proof*(sidecar: DataColumnSidecar): Result[void, string] =
+func verify_data_column_sidecar_inclusion_proof*(sidecar: DataColumnSidecar): Result[void, string] =
   # Verify if the given KZG commitments are included in the beacon block
-  let gindex = kzg_commitment_inclusion_proof_gindex(sidecar.index)
+  let gindex = 27.GeneralizedIndex
```
tersec, line 362: this should be a named constant somewhere (and it also should be `const` here as it is, since it is a constant)

but also see `KZG_COMMITMENT_INCLUSION_PROOF_DEPTH`, being replaced by `4.int` -- going to need a better approach at some point, and/or the devnets will become more reasonable

beacon_chain/sync/sync_manager.nim
```diff
@@ -548,7 +548,7 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A)
         withBlck(blck[]):
           when consensusFork >= ConsensusFork.Deneb:
             if forkyBlck.message.body.blob_kzg_commitments.len > 0:
-              hasColumns = true
+              hasColumns = true
```
tersec: adds trailing space

beacon_chain/sync/sync_manager.nim
```diff
@@ -323,6 +327,83 @@ func checkDataColumns(data_columns: seq[DataColumnSidecars]): Result[void, strin
...
+    else:
+      return(enrFieldOpt.get)
+
+proc checkValidPeerCustody*[A, B](man: SyncManager[A, B], peer: A): bool =
+  # `constructValidaCustodyPeers` returns a list of peers that are an overall
+  # superset of the given local node's custody columns and the custody columns
+  # of the remote peers
+
+  # Get the local custody subnet count
+  var localCustodySubnetCount: uint64
```
tersec: in general this construction is better written
```nim
let localCustodySubnetCount =
  if man.supernode:
    DATA_COLUMN_SIDECAR_SUBNET_COUNT.uint64
  else:
    CUSTODY_REQUIREMENT
```

that way can still get `let` immutability and there's no point in the code/data flow where it's uninitialized

But more broadly, this seems a bit specialized/maybe misplaced -- is there really nowhere else which needs to know the local custody subnet count, to know `DATA_COLUMN_SIDECAR_SUBNET_COUNT` or `CUSTODY_REQUIREMENT`?

beacon_chain/sync/sync_manager.nim
```diff
@@ -323,6 +327,83 @@ func checkDataColumns(data_columns: seq[DataColumnSidecars]): Result[void, strin
...
+  # Filtering out the invalid peers
+  for column in localCustodyColumns:
+    if column notin remoteCustodyColumns:
+      return true
+  
+  # Otherwise add the peer to the set
+  # of valid peerIds
+  # validPeerIds.add(peer.peerId)
+  # validPeerIds
+  return true
```
tersec: Can use just `true` here

beacon_chain/sync/sync_manager.nim
```diff
@@ -548,51 +629,54 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A)
...
+            warn "Received data columns sequence is not in requested range",
+              data_columns_count = len(dataColumnData), data_columns_map = getShortMap(req, dataColumnData),
+                                  request = req
+            return
+        let groupedDataColumns = groupDataColumns(req, blockData, dataColumnData)
+        if groupedDataColumns.isErr():
+          # peer.updateScore(PeerScoreNoValues)
+          man.queue.push(req)
+          # warn "Received data columns is inconsistent",
+          #   data_columns_map = getShortMap(req, dataColumnData), request = req, msg=groupedDataColumns.error()
```
tersec, line 663: come Nim 2.0.10 should be able to uncomment this, because the `foo.error()` won't trigger https://github.com/nim-lang/Nim/issues/23790

beacon_chain/sync/sync_manager.nim
```diff
@@ -548,51 +629,54 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A)
...
+            data_columns_map = getShortMap(req, dataColumnData),
+            request = req,
+            msg = checkRes.error
           return
-      let groupedDataColumns = groupDataColumns(req, blockData, dataColumnData)
-      if groupedDataColumns.isErr():
-        # peer.updateScore(PeerScoreNoValues)
-        man.queue.push(req)
-        # warn "Received data columns is inconsistent",
-        #   data_columns_map = getShortMap(req, dataColumnData), request = req, msg=groupedDataColumns.error()
```
tersec: See also Nim 2.0.10 / https://github.com/nim-lang/Nim/issues/23790

beacon_chain/sync/sync_manager.nim
```diff
@@ -548,51 +629,54 @@ proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A)
...
+          if not(checkResponse(req, uniqueSlots)):
+            # peer.updateScore(PeerScoreBadResponse)
+            man.queue.push(req)
+            warn "Received data columns sequence is not in requested range",
+              data_columns_count = len(dataColumnData), data_columns_map = getShortMap(req, dataColumnData),
+                                  request = req
+            return
+        let groupedDataColumns = groupDataColumns(req, blockData, dataColumnData)
+        if groupedDataColumns.isErr():
+          # peer.updateScore(PeerScoreNoValues)
```
tersec, line 660: I guess these were are remain commented, so no change, but is this still the Deneb blob back or something else?

Issue #6306, opened 2024-05-22
Require flag when resuming from past-weak-subjectivity database / genesis


arnetheduck:
This proposal introduces a new flag, `--long-range-sync`, that takes effect whenever the most recent state in the database (or genesis, if no database) is older than the weak subjectivity period and no trusted pivot point is given:

* `--long-range-sync=light`, default - light client protocol is required to verify the long-range sync (https://github.com/metacraft-labs/DendrETH/tree/main/docs/long-range-syncing)
  - client shuts down otherwise with a message suggesting TNS or lenient
* `--long-range-sync=lenient` - allows forward-syncing beyond WSP, including syncing from genesis

tersec:
https://github.com/status-im/nimbus-eth2/pull/6361

PR #6361, opened 2024-06-15
BN: Disable genesis sync via long-range-sync argument.

cheatfate:


beacon_chain/sync/sync_queue.nim
```diff
@@ -8,7 +8,7 @@
 {.push raises: [].}
 
 import std/[heapqueue, tables, strutils, sequtils, math]
-import stew/base10, chronos, chronicles
+import stew/base10, chronos, chronicles, results
```
tersec, line 11: None of the changes here seem to require this additional import?

beacon_chain/sync/sync_manager.nim
```diff
@@ -8,7 +8,7 @@
 {.push raises: [].}
 
 import std/[strutils, sequtils, algorithm]
-import stew/base10, chronos, chronicles
+import stew/base10, chronos, chronicles, results
```
tersec, line 11: Another case of no changes in `results` usage?

beacon_chain/sync/sync_manager.nim
```diff
@@ -711,6 +722,18 @@ proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
                     man.avgSyncSpeed.formatBiggestFloat(ffDecimal, 4) &
                     "slots/s (" & map & ":" & currentSlot & ")"
 
+    info "Debug: Sync manager loop", queue_kind = man.queue.kind,
+         flags = man.flags, is_within = man.isWithinWeakSubjectivityPeriod()
+
+    if (man.queue.kind == SyncQueueKind.Forward) and
+       (SyncManagerFlag.NoGenesisSync in man.flags):
+      if not(man.isWithinWeakSubjectivityPeriod()):
+        warn "Unable to sync with the network, because of configuration " &
```
tersec: This message should somehow include a reference specifically to either
- the phrase "weak subjectivity" (people might not know initially, but it's something they can look up and it's a [term of art in the spec](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.3/specs/phase0/weak-subjectivity.md#weak-subjectivity-period)); or
- a sentence or two summary of basically what this weak subjectivity period is (i.e. "The Weak Subjectivity Period is the number of recent epochs within which there must be a Weak Subjectivity Checkpoint to ensure that an attacker who takes control of the validator set at the beginning of the period is slashed at least a minimum threshold in the event that a conflicting `Checkpoint` is finalized." but probably without all the jargon/terminology it uses, simplified somewhat)

Otherwise, it looks arbitrary, like, well, okay, this setting isn't letting me sync, that must be a bad setting, change it, which would (a) defeat the point, and (b) miss the point.

beacon_chain/sync/sync_manager.nim
```diff
@@ -711,6 +722,18 @@ proc syncLoop[A, B](man: SyncManager[A, B]) {.async.} =
                     man.avgSyncSpeed.formatBiggestFloat(ffDecimal, 4) &
                     "slots/s (" & map & ":" & currentSlot & ")"
 
+    info "Debug: Sync manager loop", queue_kind = man.queue.kind,
```
tersec: ...

beacon_chain/sync/sync_manager.nim
tersec: The other thing this should probably do is point people to https://nimbus.guide/trusted-node-sync.html or some other resource in terms of next steps, which, while it doesn't capture exactly the flow we need (because it TNS won't work with already-extant databases, so first one has to carefully drop `db/` but not `validators/` or `secrets/`, then resync with TNS, currently, pending light client sync actually existing, so just using the existing documentation will fail).

But something to guide people to approximate next steps, or at least make it clear it's a solvable problem with specific, reliable, anticipated solutions.

tersec:
One quality-of-implementation thing here is that while it's true that to catch the general case, it has to live in the sync manager, because of the limitation of TNS not to work with existing databases, this means that this currently will, in a new `--data-dir`:
- create the database, with some genesis state
- start trying to sync
- discover that it can't sync
- alert the user and stop syncing

From a UX flow perspective this is fine in isolation for now, but the trouble is that the next step is made harder by requiring the user to manually delete the database it pointlessly created which now does nothing but hinder the usage of TNS. Once light client sync exists, and if/when we can get an Altair or newer finalized state always available (already true for Sepolia and Holesky, not true for mainnet yet), this won't be a concern, but it's currently sort of UX jank.

It already kind of exists now, and this PR doesn't really make it worse, but it's also probably the right place to fix it, because it's already the PR adding the weak subjectivity period detection. The additional check would be at the beginning, checking as a special case if slot 0 (genesis) is within the WSP of the wall slot, the same way as it currently checks whether the head state is. The difference is that the head state conceptually can't exist until the database does, while the check can be performed specially for `GENESIS_SLOT` even before the database is created.

tersec:
References https://github.com/status-im/nimbus-eth2/issues/6306

beacon_chain/sync/sync_queue.nim
cheatfate: Have you seen export? Because if you have export you definitely require import.

beacon_chain/sync/sync_manager.nim
cheatfate: My phrase is just a placeholder for the proper one, so I'm open for your proposal.

beacon_chain/sync/sync_manager.nim
cheatfate: Somebody removed `results` import somehow, but still present `results` export.

beacon_chain/sync/sync_queue.nim
tersec: One can have, e.g.,
`b.nim`:
```nim
import std/sequtils
export sequtils
```

`c.nim`:
```nim
import ./b
export sequtils
```

and `d.nim`:
```nim
import ./c
discard toSeq([0])
```

and despite `c.nim` not explicitly importing `sequtils`, it can still export it, because the import/export network has provided it and Nim tracks this.

Also, even in a more standalone way:
`c.nim`:
```nim
proc toSeq(_: auto): seq[int] = discard
export toSeq
```
and
`d.nim`:
```nim
import ./c
discard toSeq([0])
```
where there's exactly an `export` with no `import`.

I'm not a fan of the import/export network, but if it's going to exist, might as well use it

beacon_chain/sync/sync_manager.nim
tersec: It gets the `result` from one of its imports.

tersec: It gets the `results` from one of its imports.

beacon_chain/sync/sync_queue.nim
cheatfate: If module provides API functions with `Result[T, E]` signature as return value it is definitely should import and export `results`.

beacon_chain/sync/sync_manager.nim
cheatfate: Removed in https://github.com/status-im/nimbus-eth2/pull/6361/commits/b5c5cc3ca10843563e80c946a1f9048d44088cc6

beacon_chain/sync/sync_manager.nim
cheatfate: Probably, but what if one of this imports decides to deprecated results import, so we will have compilation issue of missing import in here. Also there is no warnings about `results` was imported but not used.

tersec: That (not so hypothetical, in this case) set of imports should be fixed.

> there is no warnings about `results` was imported but not used.

`a.nim`:
```nim
proc f*() = discard
```

`b.nim`
```nim
{.deprecated.}
import "."/a
export a
```

`c.nim`:
```nim
import "."/[a, b]
f()
```

also doesn't. Nim in general doesn't warn on redundant imports, only (if lucky; generics and overloading can confuse it too, so it regularly fails to flag that `std/tables` for example isn't used, at all, and it's not because it's coming in somewhere indirectly via the import/export network) on completely unused imports.

arnetheduck: > It gets the results from one of its imports.

broadly, if a module uses `Result` in one of its own public API, it should import and export `results` itself no matter what the other modules it imports do - otherwise, if the module it inherited results from stops using results, this module will break as well .

tersec:
> It already kind of exists now, and this PR doesn't really make it worse, but it's also probably the right place to fix it, because it's already the PR adding the weak subjectivity period detection. The additional check would be at the beginning, checking as a special case if slot 0 (genesis) is within the WSP of the wall slot, the same way as it currently checks whether the head state is. The difference is that the head state conceptually can't exist until the database does, while the check can be performed specially for `GENESIS_SLOT` even before the database is created.

Done in https://github.com/status-im/nimbus-eth2/pull/6361/commits/228a3b2ea46fa0c3c2b5e80bf8b0bbbe74b3db8a

beacon_chain/sync/sync_manager.nim
```diff
@@ -34,13 +34,16 @@ const
   StatusExpirationTime* = chronos.minutes(2)
     ## Time time it takes for the peer's status information to expire.
 
+  WeakSubjectivityLogMessage* =
+    "Unable to sync with the network, <placeholder>"
```
arnetheduck:
```suggestion
    "Database state missing or too old, cannot sync - resync the client using a trusted node or allow lenient long-range syncing with the `--long-range-sync:lenient` option. See https://nimbus.guide/faq.html#what-is-long-range-sync for more information."
```

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6361/commits/ad68ff6b0cd34701fde605fb4d147dac31fc1f40

beacon_chain/sync/sync_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6361/commits/ad68ff6b0cd34701fde605fb4d147dac31fc1f40

tersec (approved):

arnetheduck:
So with long-range-sync=light, the default mode, we should be performing a light client sync automatically if we have sync committees available - does that happen now?

tersec:
> So with long-range-sync=light, the default mode, we should be performing a light client sync automatically if we have sync committees available - does that happen now?

No. Light client sync does not actually exist in Nimbus right now in a way integrated with the rest of sync.

There are some options of how to approach, including but not limited to
- treat the whole thing as a beta, `--debug-foo`, hidden, not documented, and set the default back to allowing long-range sync, but keep the mechanism (which is fine and needed regardless of light client sync);
- remove the command-line option and hardwire it to allowing long-range (don't even expose as beta, just keep the mechanism);
- keep as is, current defaults, etc

I don't think it makes sense to remove it.

Part of the point from my perspective is, right now, regardless of light client sync, we repeatedly have people come to help channels having inadvertently triggered genesis sync because it's default. Putting WSP concerns aside even, just a end-user UX it's bad, because people mostly don't want this, but they show up a day or two after genesis sync has started. It's ok to allow genesis sync, and I don't actually think the trusted node sync idea of separate commands is that bad, but the combination of:
- default genesis sync; and
- trusted node sync which depends on `--data-dir` lining up; and
- the `run-foo.sh` scripts and bare binaries using different default `--data-dir`s

predictably, repeatedly, reliably creates scenarios where  someone tries to do a TNS, falls into a genesis sync, and just thinks things are a bit slower than they should be.

So, it's better regardless to make sure that genesis sync is what people want. The people who want archive modes, and therefore currently need genesis sync (though in theory one could adjust backfill to get archive mode and 100% deprecate this genesis sync, this discussion has been had, so, sure) to create archive nodes already have to specify custom settings for this.

PR #6384, opened 2024-06-23
during proposal duties, perform sync aggregation duties if not received

etan-status:
Including sync contributions into a block affects validator rewards. When we have not received aggregate sync contributions, but have seen individual messages, we can produce the contributions locally, improving validator rewards when subscribing to all subnets or when having a non-aggregating attached validator in the sync committee.


etan-status:
This also makes certain tests a bit less flaky where due to small validator count it can happen that noone is selected as an aggregator.

PR #6396, opened 2024-07-01
chore: Integrate PeerDASKZG library

kevaundray:
Currently in DRAFT as theres quite a bit of experimentation happening.

This is a PR to integrate the rust peerdas-kzg library into nimbus. The branch being used to make quick changes is here: https://github.com/crate-crypto/peerdas-kzg/pull/44

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       
-      - name: Setup Rust toolchain
+      - name: Set up Rust toolchain
         uses: dtolnay/rust-toolchain@master
         with:
           toolchain: 1.78.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.78.0
 
       - name: Restore llvm-mingw (Windows) from cache
         if: runner.os == 'Windows'
--- a/beacon_chain/kzg.nim
+++ b/beacon_chain/kzg.nim
@@ -11,7 +11,7 @@ import
   std/os,
   results,
   kzg4844/kzg_ex,
-  ../../../nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg
+  ../nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg
 
 
 from std/strutils import rsplit
--- a/beacon_chain/kzg.nim
+++ b/beacon_chain/kzg.nim
@@ -1,0 +1,40 @@
+# beacon_chain
+# Copyright (c) 2024 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+{.used.}
+
+import
+  std/os,
+  stew/results,
+  kzg4844/kzg_ex,
+  ../../../vendor/nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg
+
+
+from std/strutils import rsplit
+
+var ctx: nim_peerdas_kzg.KZGCtx
+
+proc init_kzg*(): bool =
+  # TODO: no compilation flag here because c-kzg does more than peerdas functionality.   
+  ctx = newKZGCtx()
+  template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
+  return Kzg.loadTrustedSetup(
+    sourceDir &
+      "/../vendor/nim-kzg4844/kzg4844/csources/src/trusted_setup.txt").isOk
+
+proc free_kzg*(): Result[void, string] =
+  # TODO: add code to free nim_peerdas_kzg. Removed due to nim not allowing
+  # TODO; the particular destory function being called.
+  Kzg.freeTrustedSetup()
+
+proc blobToKZGCommitment*(blob : array[131_072, byte]): Result[array[48, byte], string] =
+  when defined(USE_NIMPEERDAS_KZG):
+    let commitment = ?ctx.blobToKzgCommitment(Blob(bytes: blob))
+    ok(commitment.bytes)
+  else:
+    kzg_ex.blobToKZGCommitment(blob)
--- a/beacon_chain/kzg.nim
+++ b/beacon_chain/kzg.nim
@@ -6,11 +6,10 @@
 # at your option. This file may not be copied, modified, or distributed except according to those terms.
 
 {.push raises: [].}
-{.used.}
 
 import
   std/os,
-  stew/results,
+  results,
   kzg4844/kzg_ex,
   ../../../nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg
 
@@ -23,7 +22,7 @@ proc initKZG*(): bool =
   # TODO: no compilation flag here because c-kzg does more than peerdas functionality.   
   ctx = newKZGCtx()
   template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
-  return Kzg.loadTrustedSetup(
+  Kzg.loadTrustedSetup(
     sourceDir &
       "/../vendor/nim-kzg4844/kzg4844/csources/src/trusted_setup.txt").isOk
 
--- a/beacon_chain/kzg.nim
+++ b/beacon_chain/kzg.nim
@@ -12,7 +12,7 @@ import
   std/os,
   stew/results,
   kzg4844/kzg_ex,
-  ../../../vendor/nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg
+  ../../../nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg
 
 
 from std/strutils import rsplit
--- a/beacon_chain/kzg.nim
+++ b/beacon_chain/kzg.nim
@@ -19,7 +19,7 @@ from std/strutils import rsplit
 
 var ctx: nim_peerdas_kzg.KZGCtx
 
-proc init_kzg*(): bool =
+proc initKZG*(): bool =
   # TODO: no compilation flag here because c-kzg does more than peerdas functionality.   
   ctx = newKZGCtx()
   template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
@@ -27,7 +27,7 @@ proc init_kzg*(): bool =
     sourceDir &
       "/../vendor/nim-kzg4844/kzg4844/csources/src/trusted_setup.txt").isOk
 
-proc free_kzg*(): Result[void, string] =
+proc freeKZG*(): Result[void, string] =
   # TODO: add code to free nim_peerdas_kzg. Removed due to nim not allowing
   # TODO; the particular destory function being called.
   Kzg.freeTrustedSetup()
--- a/build_peerdas_lib.sh
+++ b/build_peerdas_lib.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+# Current usage: ./build_peerdas_lib.sh vendor/nimpeerdaskzg
+
+set -e  # Exit immediately if a command exits with a non-zero status.
+
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 <destination_folder>"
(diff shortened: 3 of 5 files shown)


build_peerdas_lib.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
kevaundray: In the rust library, we call ./scripts/compile.sh and it will compile the rust code for the host platform and copy the static libs into the nim directory. From thereon, you can call the nim code as if it were regular nim code. This script:

- clones the rust code
- calls ./scripts/compile.sh
- Copies the nim directory to the destination folder
- deletes the rust code since we no longer need it

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -12,6 +12,7 @@ import
   std/json,
   yaml,
   kzg4844/kzg_ex,
+  ../../../vendor/nimpeerdaskzg/nim_peerdas_kzg/nim_peerdas_kzg,
```
kevaundray: A hack right now, to reference the peerdas-kzg lib

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -60,12 +64,13 @@ proc runBlobToKzgCommitmentTest(suiteName, suitePath, path: string) =
     if blob.isNone:
       check output.kind == JNull
     else:
-      let commitment = blobToKzgCommitment(blob.get)
+      var blobObject = Blob(bytes: blob.get)
```
kevaundray: Not happy with needing this; will modify the API so that this can take types like array

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -60,12 +64,13 @@ proc runBlobToKzgCommitmentTest(suiteName, suitePath, path: string) =
...
+      let commitment = ctx.blobToKzgCommitment(blobObject)
       check:
         if commitment.isErr:
           output.kind == JNull
         else:
-          commitment.get == fromHex[48](output.getStr).get
+          commitment.get.bytes == fromHex[48](output.getStr).get
```
kevaundray: same comment here, re https://github.com/status-im/nimbus-eth2/pull/6396/files#r1661590547

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -60,12 +64,13 @@ proc runBlobToKzgCommitmentTest(suiteName, suitePath, path: string) =
     if blob.isNone:
       check output.kind == JNull
     else:
-      let commitment = blobToKzgCommitment(blob.get)
+      var blobObject = Blob(bytes: blob.get)
+      let commitment = ctx.blobToKzgCommitment(blobObject)
```
kevaundray: Just as a reference, if this branch is to be merged into stable, we probably do not want to modify this method since its used in 4844

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -39,7 +40,10 @@ func fromHex[N: static int](s: string): Opt[array[N, byte]] =
   except ValueError:
     Opt.none array[N, byte]
 
+var ctx: nim_peerdas_kzg.KZGCtx
+
 block:
+  ctx = newKZGCtx()
```
kevaundray: This will be modified in the future to allow one to say how many threads the library should use

kevaundray:
After running the build script, kzg tests are currently being ran with `./env.sh nim c -r tests/consensus_spec/test_fixture_kzg.nim`

kevaundray:
Most of the skeleton is now here, setting to ready to get feedback

tests/consensus_spec/test_fixture_kzg.nim
kevaundray: Seems that the c-kzg library also needed this: https://github.com/status-im/nimbus-eth2/pull/6403

Not sure why it errors on the rust, pre gcc-14 but not on the c-kzg lib pre gcc-14

beacon_chain/kzg.nim
```diff
@@ -0,0 +1,40 @@
+# beacon_chain
+# Copyright (c) 2024 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+{.used.}
```
tersec: Does this really need `{.used.}`? The `tests/` often do this because their entire purpose is the side-effects of importing them, rather than any symbols they export, but here, it seems like if one doesn't use the exported symbols it's genuinely not providing value to the importing modules.

beacon_chain/kzg.nim
```diff
@@ -0,0 +1,40 @@
...
+
+import
+  std/os,
+  stew/results,
```
tersec: `stew/results` has been deprecated; can simply use `results`

build_peerdas_lib.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Copyright (c) 2020-2023 Status Research & Development GmbH. Licensed under
```
tersec: Unexpected copyright years

kevaundray: I guess this needs to be updated in the run-*-beacon-node.sh files?

.github/workflows/ci.yml
```diff
@@ -71,6 +71,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      
+      - name: Setup Rust toolchain
```
tersec: `Setup` is the noun; `Set up` is the verb phrase

build_peerdas_lib.sh
```diff
@@ -0,0 +1,64 @@
...
+# either of:
+# - Apache License, version 2.0
+# - MIT license
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+# Current usage: ./build_peerdas_lib.sh nimpeerdaskzg
+
+set -e  # Exit immediately if a command exits with a non-zero status.
```
tersec: If `set -eu` works, that would still be POSIX-compliant and avoid some issues we've had where accidentally someone introduced a behavior dependent on bash/sh/etc interpolating a missing variable into the empty string

build_peerdas_lib.sh
kevaundray: Ah the CI only looks at files which have been modified

beacon_chain/kzg.nim
```diff
@@ -0,0 +1,40 @@
...
+
+from std/strutils import rsplit
+
+var ctx: nim_peerdas_kzg.KZGCtx
+
+proc initKZG*(): bool =
+  # TODO: no compilation flag here because c-kzg does more than peerdas functionality.   
+  ctx = newKZGCtx()
+  template sourceDir: string = currentSourcePath.rsplit(DirSep, 1)[0]
+  return Kzg.loadTrustedSetup(
```
tersec: Don't need `return` here: [expression return](https://status-im.github.io/nim-style-guide/language.result.html#practical-notes) might work better here, by omitting it

build_peerdas_lib.sh
kevaundray: have made a draft PR here: https://github.com/status-im/nimbus-eth2/pull/6404

beacon_chain/kzg.nim
kevaundray: This is being used in quite a few places, I'll modify it in this file and then push a PR to modify it in the rest of the code that is unrelated to this PR

tersec: That's true, and you're welcome to fix/update any you find

build_peerdas_lib.sh
tersec: https://github.com/status-im/nimbus-eth2/pull/6396/commits/086d100f69154b3aeee67792df47970d0aff3f74

tests/consensus_spec/test_fixture_kzg.nim
tersec: In general, Nimbus except for a couple specific cryptography libraries doesn't use threads -- would these threads be wholly contained within the KZG library?

build_peerdas_lib.sh
tersec: That PR looks good, can merge whenever

tersec: Also this is why the CI linter is failing this PR:
```
Run excluded_files="config.yaml"
- build_peerdas_lib.sh
```

tests/consensus_spec/test_fixture_kzg.nim
kevaundray: Yep they would be wholly contained in the cryptography library, you can set the number to 1 to use a single thread -- I can also look into making it a compilation flag in the future

beacon_chain/kzg.nim
kevaundray: After modifying the codebase, seems a lot are in vendored git submoduled packages (might stash the changes for now)

build_peerdas_lib.sh
kevaundray: Yep sorry was quickly preparing the stew/results -> results PR

build_peerdas_lib.sh
```diff
@@ -0,0 +1,69 @@
...
+cd "$TEMP_DIR"
+
+echo "Checking out commit: $COMMIT_HASH"
+git checkout "$COMMIT_HASH"
+
+# Check if the user is on Windows and install gnu target instead of msvc
+if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "win32" ]]; then
+    echo "Detected Windows environment. Adding x86_64-pc-windows-gnu target..."
+    rustup target add x86_64-pc-windows-gnu
+fi
```
kevaundray: I have an issue to switch to the msvc toolchain for every language -- using gnu was to make it easier for cross compilation though `cargo-xwin` works pretty well for this

beacon_chain/kzg.nim
tersec: Yes, https://github.com/status-im/nimbus-eth2/pull/6343 should have gotten anything then remaining in this repository, so the rest are going to be vendored, and would have to be done sub-repo by sub-repo

build_peerdas_lib.sh
tersec: Well, there's maybe some concern about the open-source/free-software accessibility of MSVC? But I'm not sure what the current situation there is and I don't use Windows myself. @cheatfate might be someone who would have stronger opinions on this

tests/consensus_spec/test_fixture_kzg.nim
tersec: That's fine, then -- we also have
https://github.com/status-im/nimbus-eth2/blob/cac63a3a8266c0deea46a99891bf577e9f524be4/beacon_chain/conf.nim#L264-L267
already to configure the number of threads to be modified by the end-user, so could potentially hook into that.

It's already used for BLS signature verification threads, though, so one would have to consider how to keep it from becoming a misleading or double-counted setting

beacon_chain/kzg.nim
tersec: To be clear, you're still welcome to fix any you find, but yeah it's not going to be quite as convenient as the ones in this repo were

kevaundray:
Updated the script to not build the java bindings package -- this requires javac to generate a java header file

For windows, it seems to be having trouble picking up the linker -- will try to reproduce this environment

kevaundray:
Currently going to work on a different strategy to allow an easier way to integrate and release the library

kevaundray:
@tersec can you re-run CI please?

tersec:
> @tersec can you re-run CI please?

done

kevaundray:
The error on linux seems to stem from the fact that linux has a different stack limit than the others -- The place where it is happening is not somewhere this PR has modified, so unclear why it did not pop up in the parent branch.

The reason why it is happening, is possibly due to ComputeCellsAndProofs method computing 8192*32 + 128*48 = 268KB onto the stack for each call. If this is the case, I don't think this should be fixed in this PR as its already present on the parent branch and not related to integration

agnxsh:
> The error on linux seems to stem from the fact that linux has a different stack limit than the others -- The place where it is happening is not somewhere this PR has modified, so unclear why it did not pop up in the parent branch.
> 
> The reason why it is happening, is possibly due to ComputeCellsAndProofs method computing 8192_32 + 128_48 = 268KB onto the stack for each call. If this is the case, I don't think this should be fixed in this PR as its already present on the parent branch and not related to integration

Currently the c-kzg backend we're using is a fork that i've created, i think this issue was fixed in `ethereum/c-kzg` on the `das` branch a few weeks earlier, so a subsequent upstream and changing of the git submodule should fix the issue, the reason why this wasn't already resolved  from my side was because, there was also a function param change in `loadTrustedSetup` in general, which would cause several breaking changes in the `deneb` portion of nimbus-eth2, surprisingly this function change only exists in the `das` branch and not in `master`, so I found it not super productive to fix this.

PR #6414, opened 2024-07-05
BN: User agent recognition and error codes decoding.

cheatfate:
This PR adds, detection of remote peers via libp2p user agent string.
Using this we could properly decode application-defined error codes.
Also one more metric was added `nbc_disconnects_count` which indicates number of disconnects from specific CL nodes.

### DIFF
--- a/beacon_chain/networking/eth2_agents.nim
+++ b/beacon_chain/networking/eth2_agents.nim
@@ -24,7 +24,7 @@ type
 func `$`*(a: Eth2Agent): string =
   case a
   of Eth2Agent.Unknown:
-    "unrecognized"
+    "pending/unknown"
   of Eth2Agent.Nimbus:
     "nimbus"
   of Eth2Agent.Lighthouse:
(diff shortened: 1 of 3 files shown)


beacon_chain/networking/eth2_agents.nim
```diff
@@ -0,0 +1,142 @@
...
+  of Eth2Agent.Teku:
+    "teku"
+  of Eth2Agent.Lodestar:
+    "lodestar"
+  of Eth2Agent.Grandine:
+    "grandine"
+
+const
+  # Lighthouse errors could be found here
+  # https://github.com/sigp/lighthouse/blob/stable/beacon_node/lighthouse_network/src/rpc/methods.rs
```
tersec: These `stable` branch links drift over time (see also `develop`, `unstable`, et cetera). They work now, but they don't work well as permalinks, as these projects refactor/rename/etc. Better to use either a tag (some version) or commit hash.

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6414/commits/d5a0afe3edaa767614bf1bc57ec039cf4410997f

PR #6417, opened 2024-07-08
nullability annotations for `libnimbus_lc.h`

etan-status:
Annotate functions that may return `NULL` with `_Nullable` to properly bridge into Swift.


etan-status:
[6493-ios.txt](https://github.com/user-attachments/files/16131546/6493-ios.txt)

PR #6419, opened 2024-07-10
Replace deprecated ENR functions

kdeme:


beacon_chain/conf.nim
```diff
@@ -1269,8 +1269,11 @@ func completeCmdArg*(T: type WalletName, input: string): seq[string] =
   return @[]
 
 proc parseCmdArg*(T: type enr.Record, p: string): T {.raises: [ValueError].} =
-  if not fromURI(result, p):
-    raise newException(ValueError, "Invalid ENR")
+  let res = enr.Record.fromURI(p)
+  if res.isErr:
+    raise newException(ValueError, "Invalid ENR:" & $res.error)
+
+  res.value
```
arnetheduck, line 1276:
```suggestion
  enr.Record.fromURI(p).valueOr:
    raise newException(ValueError, "Invalid ENR:" & $error)
```

kdeme: I tried that but using the error template will not work:

```
/home/deme/repos/nimbus-eth2/beacon_chain/conf.nim(1273, 53) Error: type mismatch
Expression: $error
  [1] error: template (eventName: static[string], props: varargs[untyped]) | template (stream: typedesc, eventName: static[string], props: varargs[untyped]) | proc (self: Result[error.T, error.E]): E{.noSideEffect.}
```

I think this is still the same issue as https://github.com/arnetheduck/nim-results/issues/34, but with a Nim 2.0.x error msg.

PR #6430, opened 2024-07-17
bump nimcrypto to `71bca15508e2c0548f32b42a69bcfb1ccd9ab9ff`

etan-status:
- double keccak speed


tersec:
Slightly on the risky side this late (a few days at most) before the next release branches.

cheatfate:
keccak is not used anywhere in `nimbus-eth2` why it is needed?

etan-status:
It's used when syncing optimistically without an attached EL, or when EL is temporarily unreachable due to maintenance.

At the very least, some component must validate that the execution_payload.block_hash is correct. for that, keccak is needed. If EL is connected, it can be queried. Otherwise, CL has to do it themselves or wait until EL is back online.

etan-status:
https://github.com/status-im/nimbus-eth2/blob/377698d65abe0802a4520440c93d431a09ac60ab/beacon_chain/spec/helpers.nim#L607

PR #6438, opened 2024-07-22
Fix blob syncing for Electra

etan-status:
`BlobSidecar` requests on libp2p have a context prefix based on:

> The `<context-bytes>` field is calculated as context =
>     `compute_fork_digest(fork_version, genesis_validators_root)`

We currently only process blobs if that indicates Deneb, meaning that on Electra we incorrectly report `InvalidContextBytes` and refuse to process the blob response data.

Fix this, and also ensure that the code no longer needs maintenance with every fork unrelated to blobs.

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -78,8 +78,7 @@ proc readChunkPayload*(
       else:
         return err(res.error)
     else:
-      else:
-        return neterr InvalidContextBytes
+      return neterr InvalidContextBytes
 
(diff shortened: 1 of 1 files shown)


etan-status:
Spec is a bit vague about the fork digest to use, btw.

When producing response, we use `peer.network.forkDigestAtEpoch(blockRef.slot.epoch).data`.

It doesn't hurt to just accept either deneb/electra fork digests when reading.

etan-status:
- https://github.com/ethereum/consensus-specs/pull/3850

PR #6451, opened 2024-07-25
introduce `ForkedBlobSidecar` for EIP-7688 Electra period before PeerDAS

etan-status:
On `ELECTRA_FORK_EPOCH`, PeerDAS is not yet activated, hence the current mechanism based on `BlobSidecar` is still in use. With EIP-7688, the generalized indices of `BeaconBlockBody` get reindexed, changing the length of the inclusion proof within the `BlobSidecar`. Because network Req/Resp operations allow responses across fork boundaries, this creates the need for a `ForkedBlobSidecar` in that layer, same as already done for `ForkedSignedBeaconBock` for similar reasons.

Note: This PR is only needed if PeerDAS is adopted _after_ EIP-7688. If PeerDAS is adopted _before_ EIP-7688, a similar PR may be needed for forked columns. Coincidental `Forked` jank can only be fully avoided if both features activate at the same epoch, actual changes to blobs aside. Delaying EIP-7688 for sole purpose of epoch alignemnt is not worth it.


tersec:
> Note: This PR is only needed if PeerDAS is adopted _after_ EIP-7688.

Is this particularly possible now? PeerDAS appears to [100% be in Fulu](https://github.com/ethereum/consensus-specs/pull/3994), so I don't see the scenario where PeerDAS is adopted after EIP-7688, even if EIP-7688 also lands in Fulu.

etan-status:
I need this branch for 7688 testing, it is marked as draft.

etan-status:
nimv2_2 failure is due to CI using incorrect Nim 2.0.6 version

etan-status:
Integrated into `feat_eip-7688` branch for https://stabilitynow.box testing.

This will eventually be rebased on top of Fulu, making this particular component irrelevant (as Fulu likely transitions away from `BlobSidecar` to `DataColumnSidecar`).

PR #6484, opened 2024-08-09
Use Pippenger multiplication for combining multiple sigs of same msg

etan-status:
Newer `blst` releases expose multiscalar Pippenger multiplication that allows accelerated verification of signatures pertaining to same msg.

- https://gist.github.com/wemeetagain/d52fc4b077f80db6e423935244c2afb2


etan-status:
- Needs https://github.com/status-im/nim-blscurve/pull/178

beacon_chain/gossip_processing/batch_validation.nim
```diff
@@ -290,13 +307,18 @@ proc processBatch(
     # may not be beneficial to use batch verification:
     # https://github.com/status-im/nim-blscurve/blob/3956f63dd0ed5d7939f6195ee09e4c5c1ace9001/blscurve/bls_batch_verifier.nim#L390
     if numSets == 1:
-      blsVerify(batch[].sigsets[0])
+      var r = true
+      for multiSet in batch[].multiSets.values():
+        r = r and blsVerify(multiSet.combine(verifier))
```
tersec: why not early-exit on `not blsVerify()`? keep constant time, or is there some other motivation

etan-status: there's only a single entry in there anyway. but can add a break as well

etan-status: 574a99dd10426adbd58ffac537cb60d50749a991

PR #6486, opened 2024-08-11
add: metadata-v3 for custody subnet count

agnxsh:
This PR adds metadata V3 for custody_subnet_count, keeping metadata V2 compatible.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -2077,12 +2077,32 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
 import ./peer_protocol
 export peer_protocol
 
+proc metadataV2ToV3(metadata: altair.MetaData): eip7594.MetaData =
+  eip7594.MetaData(
+    seq_number: metadata.seq_number,
+    attnets: metadata.attnets,
+    syncnets: metadata.syncnets)
+
+proc getMetadata_vx(node: Eth2Node, peer: Peer): 
+                    Future[NetRes[eip7594.MetaData]]
+                   {.async: (raises: [CancelledError]).} =
+  let
+    res = 
+      if node.cfg.EIP7594_FORK_EPOCH != FAR_FUTURE_EPOCH:
+        # Directly fetch eip7594 metadata if available
+        await getMetadata_v3(peer)
+      else:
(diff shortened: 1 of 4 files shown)


beacon_chain/networking/peer_protocol.nim
```diff
@@ -175,10 +175,14 @@ p2pProtocol PeerSync(version = 1,
     {.libp2pProtocol("metadata", 1).} =
     raise newException(InvalidInputsError, "GetMetaData v1 unsupported")
 
-  proc getMetadata_v2(peer: Peer): altair.MetaData
+  proc getMetadata_v2(peer: Peer): eip7594.MetaData
```
tersec: The v2 endpoint is still defined to return an `altair.MetaData`. One way to do this is construct an `altair.MetaData` from the internal `eip7594.MetaData`

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -9,7 +9,7 @@
 
 import
   std/[sequtils],
-  "."/[base, deneb], 
+  "."/[base, deneb, altair],
```
tersec: Usually the codebase uses either the fork order (`base, altair, deneb`) or alphabetical (`altair, base, deneb`), but here fork order seems to be usual

beacon_chain/networking/eth2_network.nim
```diff
@@ -2082,7 +2082,7 @@ proc updatePeerMetadata(node: Eth2Node, peerId: PeerId) {.async: (raises: [Cance
 
   let
     peer = node.getPeer(peerId)
-    newMetadataRes = await peer.getMetadata_v2()
+    newMetadataRes = await peer.getMetadata_v3()
```
tersec: This unconditional replacement of the v2 endpoint with the v3 endpoint only works if all the other peers also support it. This might be true for the PeerDAS devnets, in which case, ok. But it's not something ultimately mergeable into Nimbus `unstable`

beacon_chain/networking/peer_protocol.nim
agnxsh: I think that should be ideal. ok

beacon_chain/networking/eth2_network.nim
agnxsh: Agree on this, I can probably write it in a way to support v2 as well.

PR #6489, opened 2024-08-13
kurtosis network testing script

advaita-saha:
The kurtosis package that will spin up a private Ethereum testnet over Docker or Kubernetes with multi-client support, Flashbot's mev-boost infrastructure for PBS-related testing/validation, and other useful network tools (transaction spammer, monitoring tools, etc). Kurtosis packages are entirely reproducible and composable, so this will work the same way over Docker or Kubernetes, in the cloud or locally on your machine. For more details https://github.com/ethpandaops/ethereum-package

This PR adds easier support to spin up test networks and run testnets locally. For running the testnet and then the checks run the `run-kurtosis-check.sh` script. This will spin up a testnet with your local nimbus-eth2 build and pull the latest releases from the other clients then run the following tests

- Check if atleast one client is ready
- Check if all EL and CL clients are synced
  - Check if CL clients are synced
  - Check if EL clients are synced
- Check consensus finality
- Check consensus attestation stats
- Check consensus reorgs
- Check consensus forks

- Check if all clients propose blocks with legacy EOA transactions
- 10 EOA transactions per block
- check block proposals
- Blob transaction tests
- Execute all different types of transactions 
- Check chain stability ( finality checks )

## The current client combination matrix

All combinations from the bellow table is made as client pairs and a network is setup for the tests

EL | CL | 
--- | --- |
geth | nimbus-eth2 (local) |
nethermind | lighthouse |
erigon | prysm |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,11 +35,9 @@ RUN apt update && apt -y upgrade
 
 RUN ldd --version ldd
 
-RUN rm -rf /home/user/nimbus-eth2/build/beacon_node
 RUN rm -rf /home/user/nimbus-eth2/build/nimbus_beacon_node
 
 # "COPY" creates new image layers, so we cram all we can into one command
-COPY --from=build /root/nimbus-eth2/build/nimbus_beacon_node /home/user/nimbus-eth2/build/beacon_node
 COPY --from=build /root/nimbus-eth2/build/nimbus_beacon_node /home/user/nimbus-eth2/build/nimbus_beacon_node
 
 ENV PATH="/home/user/nimbus-eth2/build:${PATH}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,49 @@
+# Copyright (c) 2024 Status Research & Development GmbH
+# Licensed under either of
+#  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
+#    http://www.apache.org/licenses/LICENSE-2.0)
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+FROM debian:testing-slim AS build
+
+SHELL ["/bin/bash", "-c"]
+
+RUN apt-get clean && apt update \
+ && apt -y install build-essential git-lfs
+
+RUN ldd --version ldd
+
+ADD . /root/nimbus-eth2
+
+RUN cd /root/nimbus-eth2 \
+ && make -j$(nproc) update \
+ && make -j$(nproc) V=1 NIMFLAGS="-d:const_preset=mainnet -d:disableMarchNative" LOG_LEVEL=TRACE nimbus_beacon_node
+
+
+# --------------------------------- #
+# Starting new image to reduce size #
+# --------------------------------- #
+FROM debian:testing-slim as deploy
+
+SHELL ["/bin/bash", "-c"]
+RUN apt-get clean && apt update \
+ && apt -y install build-essential
+RUN apt update && apt -y upgrade
+
+RUN ldd --version ldd
+
+RUN rm -rf /home/user/nimbus-eth2/build/beacon_node
+RUN rm -rf /home/user/nimbus-eth2/build/nimbus_beacon_node
+
+# "COPY" creates new image layers, so we cram all we can into one command
+COPY --from=build /root/nimbus-eth2/build/nimbus_beacon_node /home/user/nimbus-eth2/build/beacon_node
+COPY --from=build /root/nimbus-eth2/build/nimbus_beacon_node /home/user/nimbus-eth2/build/nimbus_beacon_node
+
+ENV PATH="/home/user/nimbus-eth2/build:${PATH}"
+ENTRYPOINT ["nimbus_beacon_node"]
+WORKDIR /home/user/nimbus-eth2/build
+
+STOPSIGNAL SIGINT
--- a/run-kurtosis-check.sh
+++ b/run-kurtosis-check.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+set -euo pipefail
 # Copyright (c) 2024 Status Research & Development GmbH
 # Licensed under either of
 #  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or
@@ -175,7 +175,7 @@ else
     status_lines+=("$(date +'%Y-%m-%d %H:%M:%S')  Test Status:")
 
     tests=$(curl -s "${assertoor_url}"/api/v1/test_runs | jq -c ".data[] | {run_id, test_id, name, status}")
-    while read test; do
+    while read -r test; do
       if [ -z "$test" ]; then
         continue
       fi
@@ -210,7 +210,7 @@ else
       echo -e "$status_line"
     done
 
-    if ! [ -z "$running_test" ]; then
+    if [ -n "$running_test" ]; then
       task_lines=$(get_tasks_status "$running_test")
       echo "Active Test Task Status:"
       echo "$task_lines"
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+FROM debian:testing-slim as deploy
+
+SHELL ["/bin/bash", "-c"]
+RUN apt-get clean && apt update \
+ && apt -y install build-essential
+RUN apt update && apt -y upgrade
+
+RUN ldd --version ldd
+
+RUN rm -rf /home/user/nimbus-eth2/build/beacon_node
```
tersec: Shouldn't have to remove this, because it shouldn't be created

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN rm -rf /home/user/nimbus-eth2/build/nimbus_beacon_node
+
+# "COPY" creates new image layers, so we cram all we can into one command
+COPY --from=build /root/nimbus-eth2/build/nimbus_beacon_node /home/user/nimbus-eth2/build/beacon_node
```
tersec: Why does this create a `beacon_node` copy?

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
+#!/bin/bash
+
```
tersec: `set -euo pipefail` would probably help. If this script fails, due to a missing variable (and by default, `bash` interpolates a reference to them to be the empty string, so it will fail/behave strangely silently), that should be an explicit error. Without `o pipefail`, the `-eu` part is limited in its effect because this does use pipes.

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+#  * MIT license ([LICENSE-MIT](LICENSE-MIT) or
+#    http://opensource.org/licenses/MIT)
+# at your option. This file may not be copied, modified, or distributed except
+# according to those terms.
+
+# ------------------------------------------------
+#          Inputs on how to run checks
+# ------------------------------------------------
+echo
+printf "Do you want to run the checks in terminal or visit the assertoor URL? (terminal/url) "
```
tersec, line 16: This `printf`/`read` pattern seem to only work interactively, or with some kludge such as feeding a prepared stdin to the the script?

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+new_cl_image="localtestnet"
+
+# ------------------------------------------------
+#            Build the Docker Image
+# ------------------------------------------------
+if [[ "$use_previous_image" == "no" ]]; then
+  echo "Using the previously built docker image"
+  echo
+  echo -n "Please enter the docker image name (default: localtestnet) "
+  read -r cl_image
```
tersec, line 64: [`read` without `-r` will mangle backslashes.](https://www.shellcheck.net/wiki/SC2162)

It seems better to use `read -r` by default, even, for each of these `read`s.

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+  # save test results & status to github output
+  echo "test_result=$(echo "$final_test_result")"
+  echo "test_status"
+  for status_line in "${status_lines[@]}"
+  do
+    echo -e "$status_line"
+  done
+  echo
+
+  if ! [ -z "$failed_test_id" ]; then
```
tersec:
```suggestion
  if [ -n "$failed_test_id" ]; then
```

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+      status_lines+=("  $(printf '%-3s' "$test_id") $status_name \t$test_name")
+      total_tests=$(expr $total_tests + 1)
+    done <<< $(echo "$tests")
+
+    for status_line in "${status_lines[@]}"
+    do
+      echo -e "$status_line"
+    done
+
+    if ! [ -z "$running_test" ]; then
```
tersec:
```suggestion
    if [ -n "$running_test" ]; then
```

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+    failed_tests=0
+    total_tests=0
+    running_test=""
+
+    status_lines=()
+    task_lines=""
+    status_lines+=("$(date +'%Y-%m-%d %H:%M:%S')  Test Status:")
+
+    tests=$(curl -s "${assertoor_url}"/api/v1/test_runs | jq -c ".data[] | {run_id, test_id, name, status}")
+    while read test; do
```
tersec: Just as `read -r` is used elsewhere with freeform input, probably makes sense here too

Dockerfile
advaita-saha: Was put as a precautionary measure, but yes can be removed

Dockerfile
advaita-saha: This is part of the docker image optimisation and reduces the size 
Well this is not really helpful for this context, but will be when we might be using this in the CI

tersec: howso? I understand the optimization of copying a particular binary out, sure. But why copy it out twice under slightly different names?
https://github.com/status-im/nimbus-eth2/blob/427cbec9a77f0c58678e36e37fd46a53e086a10e/Dockerfile#L41-L43

advaita-saha: Ohhhh right right, this can definitely be removed 
It was made initially for ethpandaops devnets

run-kurtosis-check.sh
advaita-saha: I mean yeah it was made to work just interactively, for deciding on few factors before running the tests. I find this way better than passing arguments with the script like `--terminal=no`

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+  new_cl_image="localtestnet"
+fi
+
+
+# ------------------------------------------------
+#             Run the Kurtosis Tests
+# ------------------------------------------------
+
+# Use sed to replace the el_image value in the file
+cat kurtosis-network-params.yml | envsubst > assertoor.yaml
```
tersec, line 84: Since this file is created and removed fairly quickly, is it intended for human examination/consumption? If not, might be more robust to `mktemp` a filename,

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+sed -i "s/cl_image: .*/cl_image: $new_cl_image/" assertoor.yaml
+
+sudo kurtosis run \
+  --enclave nimbus-localtestnet \
+  github.com/ethpandaops/ethereum-package \
+  --args-file assertoor.yaml
+
+enclave_dump=$(kurtosis enclave inspect nimbus-localtestnet)
+assertoor_url=$(echo "$enclave_dump" | grep assertoor | grep http | sed 's/.*\(http:\/\/[0-9.:]\+\).*/\1/')
```
tersec, line 93: This is a pretty fragile parsing mechanism -- is there a more supported way of getting this information?

run-kurtosis-check.sh
```diff
@@ -0,0 +1,252 @@
...
+  sudo apt-get install -y jq
+fi
+
+new_cl_image="localtestnet"
+
+# ------------------------------------------------
+#            Build the Docker Image
+# ------------------------------------------------
+if [[ "$use_previous_image" == "no" ]]; then
+  echo "Using the previously built docker image"
```
tersec: `Docker` is a proper noun

run-kurtosis-check.sh
advaita-saha: Nope not yet 
There's no better way to fetch it for now
but afaik, the the team us working towards a better approach, but not really a priority

run-kurtosis-check.sh
advaita-saha: In case of crashes get's useful for having the config
Also in future I plan to add devnet specific config loading from ethpandaops, in that case a lot of params will be changes, will need to be investigated in case of crashes ( which is very common in devnets  )

PR #6491, opened 2024-08-14
Synchronously check all `transactions` to have non-zero length

etan-status:
Reject blocks with zero length transactions early when no EL connected.

- https://github.com/ethereum/consensus-specs/pull/3885

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -543,7 +543,7 @@ proc storeBlock(
     # When the execution layer is not available to verify the payload, we do the
     # required checks on the CL instead and proceed as if the EL was syncing
     # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.4/specs/bellatrix/beacon-chain.md#verify_and_notify_new_payload
-    # https://github.com/ethereum/consensus-specs/blob/1.5.0-alpha.4/specs/deneb/beacon-chain.md#modified-verify_and_notify_new_payload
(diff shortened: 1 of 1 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -541,31 +541,32 @@ proc storeBlock(
...
-    # Client software MUST validate `blockHash` value as being equivalent to
-    # `Keccak256(RLP(ExecutionBlockHeader))`
-    # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.3/src/engine/paris.md#specification
-    #
-    # This should simulate an unsynced EL, which still must perform these
-    # checks. This means it must be able to do so without context, beyond
-    # whatever data the block itself contains.
+    # required checks on the CL instead and proceed as if the EL was syncing
+    # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.4/specs/bellatrix/beacon-chain.md#verify_and_notify_new_payload
+    # https://github.com/ethereum/consensus-specs/blob/1.5.0-alpha.4/specs/deneb/beacon-chain.md#modified-verify_and_notify_new_payload
```
tersec: these two lines refer to the spec version differently, with/without v prefix?

```suggestion
    # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.4/specs/deneb/beacon-chain.md#modified-verify_and_notify_new_payload
```

etan-status: thanks, have fixed it. the `v` prefix is correct.

PR #6496, opened 2024-08-17
config.nims: fix build/makefile_generation for riscv64

gballet:
I'm currently trying to build nimbus for my riscv64 laptop. Just like macosx, the `-march=native` command doesn't work for this gcc (abi version: `riscv64-linux-gnu`) and so the target needs to be set manually.


gballet:
nvm, that seems to be fixed in unstable, I'll have to try that.

PR #6498, opened 2024-08-20
 Improved electra attestations packing 

pedromiguelmiranda:
Improved attestations packing for Electra addressing creation of attestations with a single committee allong side with some general cleanups.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -18,6 +18,7 @@ import
   ../fork_choice/fork_choice,
   ../beacon_clock
 
+from std/algorithm import sorted, SortOrder
 from std/sequtils import keepItIf, maxIndex
 
 export blockchain_dag, fork_choice
@@ -888,6 +889,16 @@ proc getElectraAttestationsForBlock*(
         # remain valid
         candidates.add((score, slot, addr entry, j))
 
+  # Sort candidates by score (descending) and by slot as a tie-breaker
+  var
+    scoreSlotCmp = proc (a, b: typeof(candidates[0])): int =
+      if a.score == b.score:
+        cmp(b.slot, a.slot)
+      else:
+        cmp(b.score, a.score)
+
+  candidates = candidates.sorted(scoreSlotCmp, SortOrder.Ascending)
+
   # Using a greedy algorithm, select as many attestations as possible that will
   # fit in the block.
   #
@@ -904,32 +915,34 @@ proc getElectraAttestationsForBlock*(
   var
     prevEpoch = state.data.get_previous_epoch()
     prevEpochSpace = MAX_ATTESTATIONS_ELECTRA
+    candidatesPerBlock: Table[(Eth2Digest, Slot), seq[electra.Attestation]]
 
-  var res: seq[electra.Attestation]
   let totalCandidates = candidates.len()
-  while candidates.len > 0 and res.lenu64() <
+  while candidates.len > 0 and candidatesPerBlock.lenu64() <
       MAX_ATTESTATIONS_ELECTRA * MAX_COMMITTEES_PER_SLOT:
     let entryCacheKey = block:
-      # Find the candidate with the highest score - slot is used as a
-      # tie-breaker so that more recent attestations are added first
-      let
-        candidate =
-          # Fast path for when all remaining candidates fit
-          if candidates.lenu64 < MAX_ATTESTATIONS_ELECTRA * MAX_COMMITTEES_PER_SLOT:
-            candidates.len - 1
-          else:
-            maxIndex(candidates)
-        (_, _, entry, j) = candidates[candidate]
-
-      candidates.del(candidate) # careful, `del` reorders candidates
+      let (_, _, entry, j) =
+        # Fast path for when all remaining candidates fit
+        if candidates.lenu64 < MAX_ATTESTATIONS_ELECTRA * MAX_COMMITTEES_PER_SLOT:
+          candidates[candidates.len - 1]
+        else:
+          # Get the candidate with the highest score
+          candidates.pop()
 
       if entry[].data.target.epoch == prevEpoch:
         if prevEpochSpace < 1:
           continue # No need to rescore since we didn't add the attestation
 
-        prevEpochSpace -= 1
+      prevEpochSpace -= 1
 
-      res.add(entry[].toElectraAttestation(entry[].aggregates[j]))
+      #TODO: Merge candidates per block structure with the candidates one
+      # and score possible on-chain attestations while collecting candidates
+      # (previous loop) and reavaluate cache key definition
+      var
+        key = (entry.data.beacon_block_root, entry.data.slot)
+        newAtt = entry[].toElectraAttestation(entry[].aggregates[j])
+
+      candidatesPerBlock.mGetOrPut(key, newSeq[electra.Attestation](0)).add(newAtt)
 
       # Update cache so that the new votes are taken into account when updating
       # the score below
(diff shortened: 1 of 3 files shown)


beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -904,32 +915,34 @@ proc getElectraAttestationsForBlock*(
...
 
       if entry[].data.target.epoch == prevEpoch:
         if prevEpochSpace < 1:
           continue # No need to rescore since we didn't add the attestation
 
-        prevEpochSpace -= 1
+      prevEpochSpace -= 1
 
-      res.add(entry[].toElectraAttestation(entry[].aggregates[j]))
+      #TODO: Merge candidates per block structure with the candidates one
```
pedromiguelmiranda, line 923: This is something that needs to be addressed given that more valuable on chain aggregates might not be included. However we need to see these changes behaviour on dev nets

tests/test_attestation_pool.nim
```diff
@@ -855,3 +856,94 @@ suite "Attestation pool electra processing" & preset():
...
+
+    let attestations = pool[].getElectraAttestationsForBlock(state[], cache)
+
+    check:
+      # A single final chain aggregated attestation should be created
+      # with same data, 2 committee bits and 3 aggregation bits
+      attestations.len == 1
+      attestations[0].aggregation_bits.countOnes() == 3
+      attestations[0].committee_bits.countOnes() == 2
+
```
tersec: trailing newline should be removed

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -950,38 +963,38 @@ proc getElectraAttestationsForBlock*(
           it.entry[].aggregates[it.validation].aggregation_bits)
 
       candidates.keepItIf:
-        # Only keep candidates that might add coverage
+          # Only keep candidates that might add coverage
```
tersec: Why the indentation change?  In general Nimbus uses a two-spaces-per-level indentation and that was correct here

tersec: https://status-im.github.io/nim-style-guide/formatting.style.html notes

Where NEP-1 notes:
> Two spaces should be used for indentation of blocks

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -950,38 +963,38 @@ proc getElectraAttestationsForBlock*(
...
         it.score > 0
 
-  # TODO sort candidates by score - or really, rewrite the whole loop above ;)
-  var res2: seq[electra.Attestation]
-  var perBlock: Table[(Eth2Digest, Slot), seq[electra.Attestation]]
+      # reorder candidates
```
tersec: Two spaces for indentation

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -888,6 +889,16 @@ proc getElectraAttestationsForBlock*(
...
 
+  # Sort candidates by score (descending) and by slot as a tie-breaker
+  var
+    scoreSlotCmp = proc (a, b: typeof(candidates[0])): int =
+      if a.score == b.score:
+        cmp(b.slot, a.slot)
+      else:
+        cmp(b.score, a.score)
+
+  candidates = candidates.sorted(scoreSlotCmp, SortOrder.Ascending)
```
tersec: foo = foo.sorted can be done with [sort](https://nim-lang.org/docs/algorithm.html#sort%2CopenArray%5BT%5D), in-place, no additional dynamic memory allocations, no whole-`seq` copying

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -888,6 +889,16 @@ proc getElectraAttestationsForBlock*(
         # remain valid
         candidates.add((score, slot, addr entry, j))
 
+  # Sort candidates by score (descending) and by slot as a tie-breaker
```
tersec: Can do this without a custom sort function (which is a bit expensive since it creates a closure, an additional call every pairwise comparison, et cetera). Since `score` and `slot` are already the first elements of
```nim
    candidates: seq[tuple[
      score: int, slot: Slot, entry: ptr ElectraAttestationEntry,
      validation: int]]
```
can just `candidates.sort(SortOrder.Descending)` to achieve the same result.

tests/test_attestation_pool.nim
```diff
@@ -757,7 +758,7 @@ suite "Attestation pool electra processing" & preset():
     # Slot 0 is a finalized slot - won't be making attestations for it..
     check:
       process_slots(
-        dag.cfg, state[], getStateField(state[], slot) + 1, cache, info,
+        dag.cfg, state[], getStateField(state[], slot) + MIN_ATTESTATION_INCLUSION_DELAY, cache, info,
```
tersec: 100+ character line. Per https://status-im.github.io/nim-style-guide/formatting.style.html
> We strive to follow NEP-1 for style matters - naming, capitalization, 80-character limit etc.

beacon_chain/consensus_object_pools/attestation_pool.nim
pedromiguelmiranda: the custom sort function is to use slot as a tie breaker for equal scores, does default serves this purpose? or can we ignore slot as tie-breaker for a better performance?

pedromiguelmiranda: ok, just tested it. nice one!

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -8,6 +8,7 @@
 {.push raises: [].}
 
 import
+  algorithm,
```
tersec: Typically, we'd use `std/algorithm` (i.e. explicitly use the `std/` prefix to indicate it's from the stdlib and not some other `algorithm`)

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -904,32 +908,34 @@ proc getElectraAttestationsForBlock*(
...
           continue # No need to rescore since we didn't add the attestation
 
-        prevEpochSpace -= 1
+      prevEpochSpace -= 1
+
+      #TODO: Merge candidates per block structure with the candidates one
+      # and score possible on-chain attestations while collecting candidates
+      # (previous loop) and reavaluate cache key definition
+      var
+        key = (entry.data.beacon_block_root, entry.data.slot)
```
tersec, line 927: For both `key` and `newAtt`, https://status-im.github.io/nim-style-guide/language.vardecl.html notes:
> Use the most restrictive of `const`, `let` and `var` that the situation allows.

Here, `key` and `newAtt` are used immutably, and not themselves changed once declared and initialized, so they can be `let` (immutable) rather than `var` (mutable).

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -904,32 +908,34 @@ proc getElectraAttestationsForBlock*(
...
+        newAtt = entry[].toElectraAttestation(entry[].aggregates[j])
 
-      res.add(entry[].toElectraAttestation(entry[].aggregates[j]))
+      candidatesPerBlock.mGetOrPut(key, newSeq[electra.Attestation](0)).add(newAtt)
```
tersec:
```suggestion
      candidatesPerBlock.mGetOrPut(key, @[newAtt])
```
there's a simpler `seq` constructor

pedromiguelmiranda: The suggestion idea is good, it is a simpler seq constructor, however does not mimic the intended code behaviour (require additional uncessary code)

We want to add a new attestation to an existing sequence associated with a given root. if this root does not exist, we want to create its entry on the table and insert it.

etan-status: There is also `.withValue` for some alternate syntax

```nim
  candidatesPerBlock.withValue(key, candidate):
    candidate[].add newAtt
  do:
    candidatesPerBlock[key] = @[newAtt]
```

has the advantage of not creating the `newSeq[electra.Attestation](0)` in every case (for regular calls, all arguments are non-lazily evaluated before function is called -- that includes the default value passed here).

but either's fine here, probably. `mgetOrPut` is more concise but sometimes is a bit clunky if there's not already a function defined that can work on the returned `var`

tersec: Oops, yeah, misread where the parens went.

Agree with both of the last two points -- it wasn't `newSeq[electra.Attestation](0).add(newAtt)` or similar being passed to `mgetOrPut`, it was rather the empty seq.

So, would go with @etan-status 's approach here.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -904,32 +908,37 @@ proc getElectraAttestationsForBlock*(
...
-          if candidates.lenu64 < MAX_ATTESTATIONS_ELECTRA * MAX_COMMITTEES_PER_SLOT:
-            candidates.len - 1
-          else:
-            maxIndex(candidates)
-        (_, _, entry, j) = candidates[candidate]
-
-      candidates.del(candidate) # careful, `del` reorders candidates
+      let (_, _, entry, j) =
+        # Fast path for when all remaining candidates fit
+        if candidates.lenu64 < MAX_ATTESTATIONS_ELECTRA * MAX_COMMITTEES_PER_SLOT:
```
tersec: I do see that this inherits from the previous code, but I'm not sure this is the correct bound for the fast path. The point here is when the candidates all fit, which is only truly the case when under `MAX_ATTESTATIONS_ELECTRA` per block.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -904,32 +908,37 @@ proc getElectraAttestationsForBlock*(
...
+        else:
+          # Get the candidate with the highest score
+          candidates.pop()
 
       if entry[].data.target.epoch == prevEpoch:
         if prevEpochSpace < 1:
           continue # No need to rescore since we didn't add the attestation
 
-        prevEpochSpace -= 1
+      prevEpochSpace -= 1
```
tersec: See https://github.com/status-im/nimbus-eth2/pull/6508 -- the entire `prevEpoch`/`prevEpochSpace`/`if entry[].data.target.epoch == prevEpoch:` machinery can go and was only ever a phase0 spec workaround, irrelevant to Electra attestation packing.

tersec: To avoid interfering with this PR, I did it there only for non-Electra attestation packing, but basically the same steps work here.

pedromiguelmiranda: Perfect. Thank you. will add

PR #6505, opened 2024-08-21
inhibit LC sync while DAG is synced

etan-status:
Normally, running LC and DAG sync at same time is fine, but on tiny devnet where some peer may not support the LC data, we can end up in situation where peer gets disconnected when DAG is in sync, because DAG sync never uses any req/resp on local devnet (perfect nw conditions) so the LC sync over minutes removes the peer as sync is stuck.

We don't need to actively sync LC from network if DAG is already synced, preventing this specific low peer devnet issue (there are others still). LC is still locally updated when DAG finalized checkpoint advances.


etan-status:
Tested on Holesky with additional trace logging to see that the inhibit fn result is properly computed and honored.

PR #6506, opened 2024-08-22
Revert "stop testing broken upstream version-2-0"

arnetheduck:
Reverts status-im/nimbus-eth2#6499

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,18 @@ jobs:
             cpu: amd64
           - os: windows
             cpu: amd64
-        branch: [~]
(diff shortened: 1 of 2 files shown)


arnetheduck:
https://github.com/nim-lang/Nim/issues/24002

tersec:
No objections to merging this once the issue is resolved

arnetheduck:
so per the comments in the PR, this is actually a latent bug in our code exposed by additional strictness - ie the template never worked as intended basically, and prints `None` in the logs.

tersec:
In which case, fixing that code in `nimbus-eth2` works also

arnetheduck:
seems to work

PR #6511, opened 2024-08-27
Add GetAggregateAttestation V2 endpoint version

pedromiguelmiranda:
Added new version (v2) of GetAggregateAttestation endpoint to support EIP-7549.
spec: https://github.com/ethereum/beacon-APIs/pull/447

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -199,11 +199,16 @@ proc addForkChoiceVotes(
       # hopefully the fork choice will heal itself over time.
       error "Couldn't add attestation to fork choice, bug?", err = v.error()
 
-func candidateIdx(pool: AttestationPool, slot: Slot): Opt[int] =
+func candidateIdx(pool: AttestationPool, slot: Slot,
+  isElectra: bool = false): Opt[int] =
   static: doAssert pool.phase0Candidates.len == pool.electraCandidates.len
+
+  let poolLength = if isElectra:
+    pool.electraCandidates.lenu64 else: pool.phase0Candidates.lenu64
+
   if slot >= pool.startingSlot and
-      slot < (pool.startingSlot + pool.phase0Candidates.lenu64):
-    Opt.some(int(slot mod pool.phase0Candidates.lenu64))
+      slot < (pool.startingSlot + poolLength):
+    Opt.some(int(slot mod poolLength))
   else:
     Opt.none(int)
 
@@ -978,7 +983,8 @@ proc getElectraAttestationsForBlock*(
     else:
       default(seq[electra.Attestation])
 
-func bestValidation(aggregates: openArray[Phase0Validation]): (int, int) =
+func bestValidation(
+  aggregates: openArray[Phase0Validation | ElectraValidation]): (int, int) =
   # Look for best validation based on number of votes in the aggregate
   doAssert aggregates.len() > 0,
     "updateAggregates should have created at least one aggregate"
@@ -992,6 +998,29 @@ func bestValidation(aggregates: openArray[Phase0Validation]): (int, int) =
       best = count
       bestIndex = i
   (bestIndex, best)
+
+func getElectraAggregatedAttestation*(
+    pool: var AttestationPool, slot: Slot,
+    attestationDataRoot: Eth2Digest, committeeIndex: CommitteeIndex):
+    Opt[electra.Attestation] =
+
+  let candidateIdx = pool.candidateIdx(slot)
+  if candidateIdx.isNone:
+    return Opt.none(electra.Attestation)
+
+  var res: Opt[electra.Attestation]
+  for _, entry in pool.electraCandidates[candidateIdx.get].mpairs():
+    if entry.data.index != committeeIndex.uint64:
+      continue
+
+    entry.updateAggregates()
+
+    let (bestIndex, best) = bestValidation(entry.aggregates)
+
+    if res.isNone() or best > res.get().aggregation_bits.countOnes():
+      res = Opt.some(entry.toElectraAttestation(entry.aggregates[bestIndex]))
+
+  res
 
 func getAggregatedAttestation*(
     pool: var AttestationPool, slot: Slot, attestation_data_root: Eth2Digest):
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -984,7 +984,7 @@ proc getElectraAttestationsForBlock*(
       default(seq[electra.Attestation])
 
 func bestValidation(
-  aggregates: openArray[Phase0Validation | ElectraValidation]): (int, int) =
+    aggregates: openArray[Phase0Validation | ElectraValidation]): (int, int) =
   # Look for best validation based on number of votes in the aggregate
   doAssert aggregates.len() > 0,
     "updateAggregates should have created at least one aggregate"
@@ -1010,7 +1010,7 @@ func getElectraAggregatedAttestation*(
 
   var res: Opt[electra.Attestation]
   for _, entry in pool.electraCandidates[candidateIdx.get].mpairs():
-    if entry.data.index != committeeIndex.uint64:
+    if entry.data.index != committeeIndex.distinctBase:
       continue
 
     entry.updateAggregates()
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -776,11 +776,6 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
     attestation_data_root: Option[Eth2Digest],
     slot: Option[Slot]) -> RestApiResponse:
 
-    let contextFork = node.dag.cfg.consensusForkAtEpoch(node.currentSlot.epoch)
-    if contextFork >= ConsensusFork.Electra:
-      return RestApiResponse.jsonError(Http410,
-                                        DeprecatedGetAggregatedAttestation)
-
     let attestation =
       block:
         let qslot =
@@ -803,7 +798,13 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
                              InvalidAttestationDataRootValueError, $res.error())
             res.get()
         let res =
-          node.attestationPool[].getAggregatedAttestation(qslot, qroot)
+          block:
+            let contextFork = node.dag.cfg.consensusForkAtEpoch(epoch(qslot))
+            if contextFork >= ConsensusFork.Electra:
+              return RestApiResponse.jsonError(Http410,
+                             DeprecatedGetAggregatedAttestation)
+
+            node.attestationPool[].getAggregatedAttestation(qslot, qroot)
         if res.isNone():
           return RestApiResponse.jsonError(Http400,
                                           UnableToGetAggregatedAttestationError)
@@ -816,14 +817,9 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
     committee_index: Option[CommitteeIndex],
     slot: Option[Slot]) -> RestApiResponse:
 
-    let contextFork = node.dag.cfg.consensusForkAtEpoch(node.currentSlot.epoch)
-    if contextFork < ConsensusFork.Electra:
-      return RestApiResponse.jsonError(Http400,
-                                        UnableToGetAggregatedAttestationError)
-
     let attestation =
       block:
-        let slot =
+        let qslot =
           block:
             if slot.isNone():
               return RestApiResponse.jsonError(Http400, MissingSlotValueError)
@@ -854,7 +850,13 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
                              InvalidAttestationDataRootValueError, $res.error())
             res.get()
         let res =
-          node.attestationPool[].getElectraAggregatedAttestation(slot, root, committee_index)
+          block:
+            let contextFork = node.dag.cfg.consensusForkAtEpoch(epoch(qslot))
+            if contextFork < ConsensusFork.Electra:
+              return RestApiResponse.jsonError(Http400,
+                             UnableToGetAggregatedAttestationError)
+
+            node.attestationPool[].getElectraAggregatedAttestation(qslot, root, committee_index)
         if res.isNone():
           return RestApiResponse.jsonError(Http400,
                                           UnableToGetAggregatedAttestationError)
--- a/beacon_chain/rpc/rest_validator_api.nim
+++ b/beacon_chain/rpc/rest_validator_api.nim
@@ -771,7 +771,40 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
         makeAttestationData(epochRef, qhead.atSlot(qslot), qindex)
     RestApiResponse.jsonResponse(adata)
 
-  # https://ethereum.github.io/beacon-APIs/#/Validator/getAggregatedAttestationV2
+  # https://ethereum.github.io/beacon-APIs/#/Validator/getAggregatedAttestation
+  router.api2(MethodGet, "/eth/v1/validator/aggregate_attestation") do (
+    attestation_data_root: Option[Eth2Digest],
+    slot: Option[Slot]) -> RestApiResponse:
+    let attestation =
+      block:
+        let qslot =
+          block:
+            if slot.isNone():
+              return RestApiResponse.jsonError(Http400, MissingSlotValueError)
+            let res = slot.get()
+            if res.isErr():
+              return RestApiResponse.jsonError(Http400, InvalidSlotValueError,
+                                               $res.error())
+            res.get()
+        let qroot =
+          block:
+            if attestation_data_root.isNone():
+              return RestApiResponse.jsonError(Http400,
+                                           MissingAttestationDataRootValueError)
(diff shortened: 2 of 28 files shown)


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -775,6 +775,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
   router.api2(MethodGet, "/eth/v1/validator/aggregate_attestation") do (
     attestation_data_root: Option[Eth2Digest],
     slot: Option[Slot]) -> RestApiResponse:
+
+    let contextFork = node.dag.cfg.consensusForkAtEpoch(node.currentSlot.epoch)
+    if contextFork >= ConsensusFork.Electra:
+      return RestApiResponse.jsonError(Http410,
+                                        DeprecatedGetAggregatedAttestation)
+
```
pedromiguelmiranda: Not sure if this is the best strategy. we will have nimbus-eth2 releases before the actual eletra fork, so this is my suggestion to pick the right endpoint version.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -775,6 +775,12 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
   router.api2(MethodGet, "/eth/v1/validator/aggregate_attestation") do (
     attestation_data_root: Option[Eth2Digest],
     slot: Option[Slot]) -> RestApiResponse:
+
+    let contextFork = node.dag.cfg.consensusForkAtEpoch(node.currentSlot.epoch)
```
arnetheduck: it's probably better to base this check off the slot that is being requested

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -993,6 +999,29 @@ func bestValidation(aggregates: openArray[Phase0Validation]): (int, int) =
...
+    attestationDataRoot: Eth2Digest, committeeIndex: CommitteeIndex):
+    Opt[electra.Attestation] =
+
+  let candidateIdx = pool.candidateIdx(slot)
+  if candidateIdx.isNone:
+    return Opt.none(electra.Attestation)
+
+  var res: Opt[electra.Attestation]
+  for _, entry in pool.electraCandidates[candidateIdx.get].mpairs():
+    if entry.data.index != committeeIndex.uint64:
```
tersec:
```suggestion
    if entry.data.index != committeeIndex.distinctBase:
```
one goal here to retain more type-safety, and avoid potentially unsafe integer conversions in general, while continuing to use the semi-random mix of `distinct` integer types sometimes used.

beacon_chain/rpc/rest_validator_api.nim
tersec: It's reasonable, yes.

Eventually this can be deprecated and the implementation aside from some placeholder removed entirely, because there won't be any non-Electra attestations in supported networks to return, but indeed, there will be at least several months of overlap before then.

As a minor formatting point, this section of the code seems to line up the `Http410` and `DeprecatedGet...` horizontally/vertically in the same column.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -978,7 +983,8 @@ proc getElectraAttestationsForBlock*(
     else:
       default(seq[electra.Attestation])
 
-func bestValidation(aggregates: openArray[Phase0Validation]): (int, int) =
+func bestValidation(
+  aggregates: openArray[Phase0Validation | ElectraValidation]): (int, int) =
```
tersec: Indent this `aggregates: ...` line 4 spaces to distinguish it from the function body

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -797,7 +798,65 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+            let res = attestation_data_root.get()
+            if res.isErr():
+              return RestApiResponse.jsonError(Http400,
+                             InvalidAttestationDataRootValueError, $res.error())
+            res.get()
+        let res =
+          block:
+            let contextFork = node.dag.cfg.consensusForkAtEpoch(epoch(qslot))
+            if contextFork < ConsensusFork.Electra:
+              return RestApiResponse.jsonError(Http400,
```
tersec: From Eth R&D Discord #apis channel earlier this month:
2024-08-30T03:36:13,675188267+00:00

And, indeed, https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/getAggregatedAttestationV2 doesn't seem to exclude pre-Electra usage.

This would make sense given that https://github.com/ethereum/beacon-APIs/pull/447 also deprecates the V1 API. If this endpoint did not support pre-Electra attestations, then that would prohibit using non-deprecated beacon API endpoints to run pre-Electra Ethereum at all, which seems unlikely to be the intention.

Regarding Nimbus's approach, it actually makes sense to get here eventually, to eventually return nothing, but arguably as a simplification of implementation of, we eventually won't need a pre-Electra attestation pool at all, so, well, the answer will be always nothing, with maybe even `Http400` as used here.

Just, not until Nimbus is ready to simply not support attestations pre-Electra, period, which as a baseline needs
- EF mainnet to fork to Electra;
- Gnosis mainnet to fork to Electra; and
- EF devnets and Kurtosis setups start with Electra rather than Deneb genesis.

These probably happen by mid-2025, at which point this can/should (maintaining effectively two attestation pools, one of which will never be used, is expensive) be re-evaluated, around the same time we might be able to deprecate/remove implementation of the V1 endpoints entirely.

Until then, the V2 endpoint is supposed to be a complete replacement pre-Electra of the V1 endpoints' functionality.

beacon_chain/spec/eth2_apis/rest_validator_calls.nim
```diff
@@ -71,6 +71,15 @@ proc getAggregatedAttestationPlain*(
...
   ## https://ethereum.github.io/beacon-APIs/#/Validator/getAggregatedAttestation
 
+proc getAggregatedAttestationPlainV2*(
+       attestation_data_root: Eth2Digest,
+       slot: Slot,
+       committee_index: CommitteeIndex
+     ): RestPlainResponse {.
+     rest, endpoint: "/eth/v2/validator/aggregate_attestation"
+     meth: MethodGet.}
+  ## https://ethereum.github.io/beacon-APIs/#/Validator/getAggregatedAttestation
```
tersec: This seems like a link to the deprecated V1 endpoint documentation, not the V2 endpoint?

beacon_chain/rpc/rest_validator_api.nim
pedromiguelmiranda: Excellent finding. Thank you.
For some reason, interpreted V1 for pre-electra and V2 after the fork, however as you noticed, the spec states that the new version should be backwards compatible and completely replace/deprecate V1.

Until mid-2025 pretty sure several refinements will/should happend, so we still got some time to improve/refine.

will proceed with the complete V1 replacement

beacon_chain/spec/eth2_apis/rest_validator_calls.nim
pedromiguelmiranda: The official consensus spec hasn't been released, so we got two options here: 

- Predictable release link ( https://ethereum.github.io/beacon-APIs/#/Validator/getAggregatedAttestationV2)
- Current dev link (https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getPoolAttestationsV2)

Should we proceed with expected release link ?

tersec: There are a couple other styles one could look at too:
- https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.5.0#/Node/getNetworkIdentity (different endpoint, but here, use the version number as part of the link, and acts as a stable link. can't be used yet though, for this endpoint)
- https://github.com/ethereum/beacon-APIs/blob/4fc68cb7e30973bb948c66084a2b94f1fe57c1c4/apis/validator/aggregate_attestation.v2.yaml or similar where one can use a usual Git commit hash to have a stable link which might not be a friendly as the  https://ethereum.github.io/beacon-APIs/ links

Maybe a combination of your dev link, which works for now but isn't predictable, with the GitHub link with commit hash to have a predictable part.

tersec: It can be adjusted later, but at any given point the link should at least work

tersec:
https://github.com/status-im/nimbus-eth2/actions/runs/10667055348/job/29568268145
```
2024-09-02T14:16:25.3542317Z /github-runner/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/validator_client/attestation_service.nim(248, 63) template/generic instantiation of `async` from here
2024-09-02T14:16:25.3544530Z /github-runner/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/validator_client/attestation_service.nim(283, 17) Error: type mismatch
2024-09-02T14:16:25.3546172Z Expression: getAggregatedAttestation(vc, slot, attestationRoot, committeeIndex, Best)
2024-09-02T14:16:25.3547078Z   [1] vc: ValidatorClientRef
2024-09-02T14:16:25.3547523Z   [2] slot: Slot
2024-09-02T14:16:25.3547931Z   [3] attestationRoot: Digest
2024-09-02T14:16:25.3548412Z   [4] committeeIndex: uint64
2024-09-02T14:16:25.3656072Z   [5] Best: ApiStrategyKind
2024-09-02T14:16:25.3656467Z 
2024-09-02T14:16:25.3656729Z Expected one of (first mismatch at [position]):
2024-09-02T14:16:25.3657569Z [3] proc getAggregatedAttestation(vc: ValidatorClientRef; slot: Slot;
2024-09-02T14:16:25.3658498Z                               committee_index: CommitteeIndex; root: Eth2Digest;
2024-09-02T14:16:25.3659324Z                               strategy: ApiStrategyKind): Future[Attestation]
2024-09-02T14:16:25.3659828Z 
2024-09-02T14:16:25.5278872Z make: *** [Makefile:447: nimbus_validator_client] Error 1
2024-09-02T14:16:25.5279590Z make: *** Waiting for unfinished jobs....
```

running `make nimbus_validator_client` locally can find this too

beacon_chain/spec/eth2_apis/rest_validator_calls.nim
```diff
@@ -63,13 +63,14 @@ proc produceAttestationDataPlain*(
      meth: MethodGet.}
   ## https://ethereum.github.io/beacon-APIs/#/Validator/produceAttestationData
 
-proc getAggregatedAttestationPlain*(
```
tersec: why remove the old one? Until can use the new APIs exclusively (coming up to/post-Electra, after the old APIs aren't there), still need this as something the VC can call.

But yes, this is another endpoint from a client perspective that can be removed at some point because it won't be useful. Just not yet.

pedromiguelmiranda: A little bit confused here...

from [here](https://github.com/status-im/nimbus-eth2/pull/6511#discussion_r1737865179):
> Until then, the V2 endpoint is supposed to be a complete replacement pre-Electra of the V1 endpoints' functionality.

if it is suppose to be a complete replacement, in my opinion, this change makes sense since it matches the spec and matches the complete endpoint replacement.

tersec: Well, at least until that point where EF and Gnosis mainnets fork and EF devnets/Kurtosis simulations start from Electra genesis, it can't be assumed that all other VCs and BNs in the network support the V2 endpoints.

Currently https://github.com/ethpandaops/ethereum-package?tab=readme-ov-file#beacon-node--validator-client-compatibility suggests that the Nimbus VC (consumer/client of these endpoints) is compatible with all the other BNs, and the Nimbus BN (server of these endpoints) is likewise compatible with all the VCs. Prematurely switching the Nimbus VC over to the V2 endpoints for not yet any gain (because they only help in Electra, otherwise they're equivalent to the V1 endpoints) unconditionally means that it would likely lose some of this compatibility.

To remain compatible with Pectra networks, all the VCs and BNs will by then have to implement the V2 endpoints, and this will not be true anymore, but as long as Dencun is the deployed fork, can't rely on other VCs or BNs supporting the V2 endpoint.

So yes, V2 is supposed to provide a complete functional replacement for the V1 endpoint, per spec, but it will be practically necessary to retain both, in both the client (`rest_validator_calls`) and server (`/rpc/rest_*`) forms for several months, concurrently. The V1 endpoints being deprecated doesn't mean they're not the ones almost 100% used today and in the near future.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -804,6 +804,53 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+                                           MissingAttestationDataRootValueError)
+            let res = attestation_data_root.get()
+            if res.isErr():
+              return RestApiResponse.jsonError(Http400,
+                             InvalidAttestationDataRootValueError, $res.error())
+            res.get()
+        let res =
+          block:
+            node.attestationPool[].getElectraAggregatedAttestation(qslot, root, committee_index)
```
tersec: Well, so it won't return an error, but it also won't ever access the pre-Electra part of the attestation pool.

And regarding how to handle the inevitable issue of their being different types -- the aggregated attestations might be, but the REST responses are the same type, so can return different REST responses and unify types that way (among others)

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -993,6 +999,29 @@ func bestValidation(aggregates: openArray[Phase0Validation]): (int, int) =
...
+    Opt[electra.Attestation] =
+
+  let candidateIdx = pool.candidateIdx(slot)
+  if candidateIdx.isNone:
+    return Opt.none(electra.Attestation)
+
+  var res: Opt[electra.Attestation]
+  for _, entry in pool.electraCandidates[candidateIdx.get].mpairs():
+    if entry.data.index != committeeIndex.distinctBase:
+      continue
```
tersec, line 1014: The point of the Electra changes is that attestations across committees can be aggregated.

https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/phase0/validator.md#construct-aggregate (phase 0, i.e. pre-Electra) states that:
> #### Construct aggregate
>
> If the validator is selected to aggregate (`is_aggregator()`), they construct an aggregate attestation via the following.
>
> Collect `attestations` seen via gossip during the `slot` that have an equivalent `attestation_data` to that constructed by the validator. If `len(attestations) > 0`, create an `aggregate_attestation: Attestation` with the following fields.
>
> ##### Data
>
> Set `aggregate_attestation.data = attestation_data` where `attestation_data` is the `AttestationData` object that is the same for each individual attestation being aggregated.
>
> ##### Aggregation bits
>
> Let `aggregate_attestation.aggregation_bits` be a `Bitlist[MAX_VALIDATORS_PER_COMMITTEE]` of length `len(committee)`, where each bit set from each individual attestation is set to `0b1`.

But the `index` field of https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/phase0/beacon-chain.md#attestationdata is the committee index, until Electra. So this condition means that aggregation cannot happen across committees pre-Electra.

https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/electra/validator.md#construct-attestation changes this:
> ### Construct attestation
>
> - Set `attestation_data.index = 0`.
> - Let `attestation.aggregation_bits` be a `Bitlist[MAX_VALIDATORS_PER_COMMITTEE * MAX_COMMITTEES_PER_SLOT]` of length `len(committee)`, where the bit of the index of the validator in the `committee` is set to `0b1`.
> - Let `attestation.committee_bits` be a `Bitvector[MAX_COMMITTEES_PER_SLOT]`, where the bit at the index associated with the validator's committee is set to `0b1`.
>
> *Note*: Calling `get_attesting_indices(state, attestation)` should return a list of length equal to 1, containing `validator_index`.

And the procedure for https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/electra/validator.md#attestation-aggregation similarly allows cross-committee index-aggregation.

pedromiguelmiranda: Yes, Electra changes will allow cross committee index-aggregationss, however, the new [getAgregatedAttestationV2](https://github.com/ethereum/beacon-APIs/pull/447) endpoint version is only interested in attestations in its corresponding committee(*) (previous version, could aggregate attestations from multiple committees since committee_index is not in its params, and that 's not the objective wanted for the API according to the spec).

Also, and more important, we are still using the _data.index_ field from phase0 (maybe as "transition/temp" structure while we still have 2 attestations pools):

-  _template addAttToPool_ on __proc addAttestation_ uses _data.index_ field to save the index obtained from _get_committee_index_one_
- _toElectraAttestation_ recovers this _index_ and saves it on _committee_bits_ while building the final electra aggregated attestation

This particular conditional statement in the code is to achieve the objective defined in (*) given the code particularites referred before.

beacon_chain/rpc/rest_validator_api.nim
pedromiguelmiranda: nice catch.

beacon_chain/consensus_object_pools/attestation_pool.nim
tersec: Oh, indeed.

PR #6515, opened 2024-08-28
Light forward sync mechanism

cheatfate:
Light forward sync uses the light client protocol to accelerate forward syncing done after checkpoint sync, when restarting the client or indeed when syncing up after a longer period of being offline.

It works by establishing a "safe" head to sync to using the light client protocol, tracing this history back to the latest known state in the database and then using a less computationally intensive method of verifying blocks based on the knowledge that the blocks lead to a valid head.

It can currently be enabled by specifying `--debug-long-range-sync=light`, though it should be understood that `--debug-` options are not compatible across client versions and should not be used in production setups.


beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+        of ConsensusFork.Capella:
+          ForkedSignedBeaconBlock.init(
+            SSZ.decode(decompressed, capella.SignedBeaconBlock))
+        of ConsensusFork.Deneb:
+          ForkedSignedBeaconBlock.init(
+            SSZ.decode(decompressed, deneb.SignedBeaconBlock))
+        of ConsensusFork.Electra:
+          ForkedSignedBeaconBlock.init(
+            SSZ.decode(decompressed, electra.SignedBeaconBlock))
+      except SerializationError as exc:
```
tersec: `exc` here isn't used

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,529 @@
...
+              bchunk.resfut.complete(Result[void, string].err(msg))
+              break innerLoop
+
+          withBlck(bdata.blck):
+            let res =
+              try:
+                await updateHead(consensusManager, validatorMonitor,
+                  overseer.getBeaconTimeFn, forkyBlck,
+                  NewPayloadStatus.noResponse)
+              except CancelledError as exc:
```
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(270, 40) Hint: 'exc' is declared but not used [XDeclaredButNotUsed]`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,529 @@
...
+  is_within_weak_subjectivity_period(
+    dag.cfg, currentSlot, dag.headState, checkpoint)
+
+proc isBackfillEmpty(backfill: BeaconBlockSummary): bool =
+  (backfill.slot == GENESIS_SLOT) and isFullZero(backfill.parent_root.data)
+
+proc isUntrustedBackfillEmpty(clist: ChainListRef): bool =
+  clist.tail.isNone()
+
+proc needsUntrustedBackfill(clist: ChainListRef, dag: ChainDagRef): bool =
```
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(173, 55) Hint: 'ChainDagRef' should be: 'ChainDAGRef' [type declared in nimbus-eth2/beacon_chain/consensus_object_pools/block_pools_types.nim(72, 3)] [Name]`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,529 @@
...
+                break mainLoop
+            if res.isErr():
+              let msg = "Unable to update head [" & res.error & "]"
+              bchunk.resfut.complete(Result[void, string].err(msg))
+              break innerLoop
+
+        bchunk.resfut.complete(Result[void, string].ok())
+
+proc verifyBlockProposer(
+    dag: ChainDagRef,
```
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(282, 10) Hint: 'ChainDagRef' should be: 'ChainDAGRef' [type declared in nimbus-eth2/beacon_chain/consensus_object_pools/block_pools_types.nim(72, 3)] [Name]`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,529 @@
...
+
+    ok()
+
+proc rebuildState(overseer: SyncOverseerRef): Future[void] {.
+     async: (raises: [CancelledError]).} =
+  overseer.statusMsg = Opt.some("rebuilding state")
+  let
+    consensusManager = overseer.consensusManager
+    dag = consensusManager.dag
+    attestationPool = consensusManager.attestationPool
```
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(308, 5) Hint: 'attestationPool' is declared but not used [XDeclaredButNotUsed]`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,529 @@
...
+    validatorMonitor = overseer.validatorMonitor
```
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(309, 5) Hint: 'validatorMonitor' is declared but not used [XDeclaredButNotUsed]`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,529 @@
...
+    checkpoint = Checkpoint(
+      epoch:
+        getStateField(dag.headState, slot).epoch(),
+      root:
+        getStateField(dag.headState, latest_block_header).state_root)
+
+  is_within_weak_subjectivity_period(
+    dag.cfg, currentSlot, dag.headState, checkpoint)
+
+proc isBackfillEmpty(backfill: BeaconBlockSummary): bool =
```
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(167, 6) Hint: 'isBackfillEmpty' is declared but not used [XDeclaredButNotUsed]`

beacon_chain/sync/sync_overseer.nim
tersec: `nimbus-eth2/beacon_chain/sync/sync_overseer.nim(173, 6) Hint: 'needsUntrustedBackfill' is declared but not used [XDeclaredButNotUsed]`

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f22779491aa960d65755654523d54443014f31cd

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bc640a064a116b909b7575561d54f7ce94280b28

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+  defer:
+    overseer.eventQueue.unregister(eventKey)
+
+  let events =
+    try:
+      await overseer.eventQueue.waitEvents(eventKey)
+    except CancelledError as exc:
+      raise exc
+    except AsyncEventQueueFullError:
+      raiseAssert "AsyncEventQueueFullError should not be happened!"
```
tersec: > should not happen!

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -589,11 +631,20 @@ proc init*(T: type BeaconNode,
...
+    notice "Genesis state information",
+           genesis_fork = genesisState.kind,
+           is_post_altair = (cfg.ALTAIR_FORK_EPOCH == GENESIS_EPOCH)
+
     if config.longRangeSync == LongRangeSyncMode.Light:
       if not is_within_weak_subjectivity_period(metadata.cfg, currentSlot,
                                                 genesisState[], checkpoint):
-        fatal WeakSubjectivityLogMessage, current_slot = currentSlot
-        quit 1
+        # We do support any network which starts from Altair fork.
```
tersec: well, either Altair or later fork.

right now, yeah, this is only Holesky really. But eventually Sepolia will be phased out and presumably its replacement will start in Bellatrix or later, like Holesky.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+proc getBlockConsensusFork(header: ChainFileHeader): ConsensusFork =
+  let hkind = unmaskKind(header.kind)
+  case hkind
+  of 0'u64: ConsensusFork.Phase0
+  of 1'u64: ConsensusFork.Altair
+  of 2'u64: ConsensusFork.Bellatrix
+  of 3'u64: ConsensusFork.Capella
+  of 4'u64: ConsensusFork.Deneb
+  of 5'u64: ConsensusFork.Electra
+  else: raiseAssert("Should not be happened")
```
tersec: > Should not happen

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+    copyMem(addr dst[offset], unsafeAddr data[0], len(data))
+    offset += len(data)
+
+  footer.store(dst.toOpenArray(offset, offset + ChainFileFooterSize - 1))
+  dst
+
+template getBlockChunkKind(kind: ConsensusFork, last: bool): uint64 =
+  let res =
+    case kind
+    of ConsensusFork.Phase0: 0'u64
```
tersec: In both `getBlockChunkKind` and `getBlockConsensusFork`, might make sense to ensure that
https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/spec/forks.nim#L44-L50
maps (as it should anyway) from 0 to 5 and just reuse the properties of that to convert to/from, rather than these `case` statements.

Yeah, going from int to the enum value isn't fully typesafe but it asserts in the existing code here too, so it'd be equivalent without adding one more place to maintain the fork list.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+    of ConsensusFork.Electra: 5'u64
+  if last:
+    maskKind(res)
+  else:
+    res
+
+template getBlobChunkKind(kind: ConsensusFork, last: bool): uint64 =
+  let res =
+    case kind
+    of ConsensusFork.Phase0, ConsensusFork.Altair, ConsensusFork.Bellatrix,
```
tersec:
```nim
    of ConsensusFork.Phase0 .. ConsensusFork.Capella:
```
more concise

beacon_chain/beacon_chain_file.nim
tersec: though see https://github.com/status-im/nimbus-eth2/pull/6515/files#r1751101376 -- the entire thing can be an (unsafe, sure, but `raiseAssert` anyway) type conversion to `ConsensusFork`.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+            blobs: Opt[BlobSidecars], compressed = true): Result[void, string] =
+  let origOffset =
+    updateFilePos(chandle.handle, 0'i64, SeekPosition.SeekEnd).valueOr:
+      return err(ioErrorMsg(error))
+
+  block:
+    let
+      kind = getBlockChunkKind(signedBlock.kind, blobs.isNone())
+      (data, plainSize) =
+        if compressed:
```
tersec: Is `store()` ever called with `compressed = false`? I couldn't find anything, and if so, the code is handling various cases it doesn't have to. Snappy is supposed to be fast enough that one can use it by default, without worrying much about its speed, and Nimbus already uses it unconditionally for various objects in the block/state database.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+            SSZ.decode(decompressed, electra.SignedBeaconBlock))
+      except SerializationError:
+        return err("Incorrect block format")
+  ok(blck)
+
+proc decodeBlob(
+    header: ChainFileHeader,
+    data: openArray[byte]
+): Result[BlobSidecar, string] =
+  if header.plainSize > uint64(high(uint32)):
```
tersec: Can be more precise than this. Uncompressed blobs are 32 * 4096 = 131,072 bytes:
https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/spec/datatypes/deneb.nim#L53-L54

Snappy might increase their size somewhat in pathological cases, but it has bounds on this.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+  if not(dataRead):
+    ok(Opt.some(Chunk(header: header, footer: footer)))
+  else:
+    ok(Opt.some(Chunk(header: header, footer: footer, data: data)))
+
+proc decodeBlock(
+    header: ChainFileHeader,
+    data: openArray[byte]
+): Result[ForkedSignedBeaconBlock, string] =
+  if header.plainSize > uint64(high(uint32)):
```
tersec: Can bound more tightly: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/phase0/p2p-interface.md#configuration and https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/configs/mainnet.yaml and https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/configs/minimal.yaml note that `GOSSIP_MAX_SIZE` is 10MiB (for "uncompressed gossip messages").

That limit applies to all gossiped SSZ objects, including blocks (and blobs, but blobs are simply fixed-size).

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+            (snappy.encode(res), len(res))
+        else:
+          withBlck(signedBlock):
+            let res = SSZ.encode(forkyBlck)
+            (res, len(res))
+      slot = signedBlock.slot
+      buffer = Chunk.init(kind, uint64(slot), uint32(plainSize), data)
+      wrote = writeFile(chandle.handle, buffer).valueOr:
+        discard truncate(chandle.handle, origOffset)
+        discard fsync(chandle.handle)
```
tersec, line 308: Why are some `fsync` calls fine to discard/ignore the return value of but others not?

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+        return err(
+          ChainFileError.init(ChainFileErrorType.HeaderError, error))
+
+  if not(dataRead):
+    setFilePos(handle, int64(header.comprSize),
+               SeekPosition.SeekCurrent).isOkOr:
+      return err(
+        ChainFileError.init(ChainFileErrorType.IoError, ioErrorMsg(error)))
+  else:
+    data.setLen(int64(header.comprSize))
```
tersec: https://nim-lang.org/docs/system.html#setLen%2Cseq%5BT%5D%2CNatural states that `newLen` is a `Natural`, which https://nim-lang.org/docs/system.html#Natural defines as ranging from `0 .. high(int)`. `int64` happens to work here because the only targets tested in CI (and, realistically, supported by `nimbus_beacon_node`, though the other components such as `ncli` and the validator client probably can still work fine as 32-bit builds) are 64-bit. But this is an accidental type-aliasing.

So I'm not sure what the best type to use here, but it's probably not `int64`.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+    return err(
+      ChainFileError.init(ChainFileErrorType.MismatchError, error))
+
+  if not(dataRead):
+    let position = int64(-ChainFileHeaderSize)
+    setFilePos(handle, position, SeekPosition.SeekCurrent).isOkOr:
+      return err(
+        ChainFileError.init(ChainFileErrorType.IoError, ioErrorMsg(error)))
+  else:
+    data.setLen(int64(header.comprSize))
```
tersec: See https://github.com/status-im/nimbus-eth2/pull/6515/files#r1751127475

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+
+template maskKind(k: uint64): uint64 =
+  k or 0x8000_0000_0000_0000'u64
+
+template isLast(k: uint64): bool =
+  (k and 0x8000_0000_0000_0000'u64) != 0'u64
+
+proc checkKind(kind: uint64): Result[void, string] =
+  let hkind = unmaskKind(kind)
+  if hkind notin [0'u64, 1, 2, 3, 4, 5, 64, 65]:
```
tersec: Presumably the 0 through 5 here are the fork numbers? This could use one of the
```nim
  static: doAssert high(ConsensusFork) == ConsensusFork.Electra
```
around the codebase for exactly this kind of not-easily-grepped/typesystem-found location which needs to be adjusted with new forks.

Better yet is to avoid adding such things, but sometimes that's infeasible.

tersec: I guess this schema also breaks after 64 consensus spec hardforks?

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,962 @@
...
+    return err("Unsuppoted chunk kind value")
+  ok()
+
+proc check(a: ChainFileHeader): Result[void, string] =
```
tersec, line 118: As far as I can tell, none of these `Result[void, string]` returns anything but a constant (amenable to `cstring`) literal string.

There's an argument more generally that we've biased a bit too much towards this in some subsystems, where it turns out they weren't so performance-sensitive as initially anticipated, but the `cstring` returns made it hard to report useful errors, but here it doesn't look like there's any reason not to avoid the memory allocations/GC pressure/churn.

tersec: (Applies to whole PR, picked a semi-arbitrary example)

beacon_chain/beacon_chain_file.nim
cheatfate: So this is because of error in `writeFile` (initial error we going to report to user), so this is attempt to cleanup. In case where number of bytes written was less than we expect, it means that file become inconsistent, so we trying to truncate file to known good size and we trying to fsync it, but we still want to report original error message to the user.

beacon_chain/beacon_chain_file.nim
cheatfate: > I guess this schema also breaks after 64 consensus spec hardforks?

Yes you are right but we can increase it to 128 or 256 if you think that 64 is not enough.

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/nimbus_beacon_node.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Such cases are not pathological because we here mostly working with high entropy sources (crypto hashes and or encrypted data).

Decreased to `MaxChunkSize` in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
cheatfate: Increased number of supported forks to 16384.
And this part was also reworked to satisfy your comments in https://github.com/status-im/nimbus-eth2/pull/6515/commits/361bc0b354fa1df13f3a0adabc8302d51739a981

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,957 @@
...
+            SSZ.decode(decompressed, bellatrix.SignedBeaconBlock))
+        of ConsensusFork.Capella:
+          ForkedSignedBeaconBlock.init(
+            SSZ.decode(decompressed, capella.SignedBeaconBlock))
+        of ConsensusFork.Deneb:
+          ForkedSignedBeaconBlock.init(
+            SSZ.decode(decompressed, deneb.SignedBeaconBlock))
+        of ConsensusFork.Electra:
+          ForkedSignedBeaconBlock.init(
+            SSZ.decode(decompressed, electra.SignedBeaconBlock))
```
etan-status:
```suggestion
        withConsensusFork(fork):
          ForkedSignedBeaconBlock.init(
            SSZ.decode(decompressed, consensusFork.SignedBeaconBlock))
```

beacon_chain/beacon_node_light_client.nim
```diff
@@ -53,10 +53,7 @@ proc initLightClient*(
       getBeaconTime, optimisticHandler)
 
     shouldInhibitSync = func(): bool =
-      if node.syncManager != nil:
-        not node.syncManager.inProgress  # No LC sync needed if DAG is in sync
-      else:
-        false
+      not node.syncOverseer.syncInProgress # No LC sync needed if DAG is in sync
```
etan-status: `node.syncOverseer` is not yet initialized here, but it should still work out because the `lightClient` implementation probably does not callback into `shouldInhibitSync` until `node.syncOverseer` exists.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -406,6 +406,112 @@ proc enqueueBlock*(
...
+              newHead.get.blck.bid.slot.epoch)
+          of ConsensusFork.Deneb, ConsensusFork.Electra:
+            # https://github.com/ethereum/execution-apis/blob/90a46e9137c89d58e818e62fa33a0347bba50085/src/engine/prague.md
+            # does not define any new forkchoiceUpdated, so reuse V3 from Dencun
+            callExpectValidFCU(payloadAttributeType = PayloadAttributesV3)
+          of ConsensusFork.Capella:
+            callExpectValidFCU(payloadAttributeType = PayloadAttributesV2)
+          of  ConsensusFork.Phase0, ConsensusFork.Altair,
+              ConsensusFork.Bellatrix:
+            callExpectValidFCU(payloadAttributeType = PayloadAttributesV1)
```
```suggestion
        template callForkChoiceUpdated: untyped = 
          withConsensusFork(consensusManager.dag.cfg.consensusForkAtEpoch(
              newHead.get.blck.bid.slot.epoch)):
            when consensusFork >= ConsensusFork.Bellatrix:
              await elManager.expectValidForkchoiceUpdated(
                headBlockPayloadAttributesType = consensusFork.PayloadAttributes,
                headBlockHash = headExecutionBlockHash,
                safeBlockHash = newHead.get.safeExecutionBlockHash,
                finalizedBlockHash = newHead.get.finalizedExecutionBlockHash,
                receivedBlock = signedBlock)
```

beacon_chain/spec/signatures_batch.nim
```diff
@@ -229,6 +229,39 @@ func bls_to_execution_change_signature_set*(
 
   SignatureSet.init(pubkey, signing_root, signature)
 
+proc collectSignatureSets*(
+  sigs: var seq[SignatureSet],
+  blocks: openArray[ForkedSignedBeaconBlock],
+  validatorKeys: openArray[ImmutableValidatorData2],
+  state: ForkedHashedBeaconState
+): Result[void, string] =
```
etan-status: Possibly safer to name this one `collectLightSignatureSets` to make it explicit that only the proposer signature is being checked. From the caller code perspective it is not obvious to me.

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+    except CancelledError as exc:
+      raise exc
+    except AsyncEventQueueFullError:
+      raiseAssert "AsyncEventQueueFullError should not happen!"
+
+  withForkyHeader(events[^1]):
+    when lcDataFork > LightClientDataFork.None:
+      forkyHeader.beacon
+    else:
+      raiseAssert "Should not be happened"
```
etan-status:
```suggestion
      raiseAssert "Should not happen"
```

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+          Opt.none(BlobSidecars)
+    ok(BlockData(blck: res.blocks[0][], blob: blob))
+  finally:
+    overseer.pool.release(peer)
+
+proc `==`(a, b: BeaconBlockHeader): bool =
+  (a.slot == b.slot) and (a.proposer_index == b.proposer_index) and
+  (a.parent_root.data == b.parent_root.data) and
+  (a.state_root.data == b.state_root.data) and
+  (a.body_root.data == b.body_root.data)
```
etan-status: isn't memberwise comparison the default `==` in nim?

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+
+proc getBlock*(
+    overseer: SyncOverseerRef,
+    slot: Slot,
+    blockHeader: Opt[BeaconBlockHeader]
+): Future[BlockData] {.async: (raises: [CancelledError]).} =
+  var workers:
+    array[PARALLEL_REQUESTS, Future[BlockDataRes].Raising([CancelledError])]
+
+  while true:
```
etan-status, line 111: would this benefit from rate limiting in case of retries?

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+                        blck = shortLog(signedBlock),
+                        reason = res.error
+              return err(VerifierError.Invalid)
+            ok()
+
+          for bchunk in blocks.chunks(onStateUpdate, BLOCKS_PROCESS_CHUNK_SIZE):
+            try:
+              overseer.blocksQueue.addLastNoWait(bchunk)
+            except AsyncQueueFullError:
+              raiseAssert "Should not be happened with unbound AsyncQueue"
```
etan-status:
```suggestion
              raiseAssert "Should not happen with unbounded AsyncQueue"
```

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+      # Checkpoint/Trusted state we have is too old.
+      error "Trusted node sync started too long time ago"
+      quit 1
+
+    if not(isUntrustedBackfillEmpty(clist)):
+      let headSlot = clist.head.get().slot
+      if not(overseer.isWithinWeakSubjectivityPeriod(headSlot)):
+        # Light forward sync file is too old.
+        warn "Light client sync was started too long time ago",
+             current_slot = currentSlot, backfill_data_slot = headSlot
```
etan-status, line 439: Do these checks apply in `Lenient` sync mode? especially the one that exits may block certain sync use cases.

Maybe we need a third mode, `LongRangeSyncMode.Strict`, that does the check + exit without a fallback to the LC sync.

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+
+    if overseer.config.longRangeSync == LongRangeSyncMode.Lenient:
+      # Starting forward sync manager/monitor only.
+      overseer.forwardSync.start()
+      return
+
+    if overseer.config.longRangeSync == LongRangeSyncMode.Light:
+      let dagHead = dag.finalizedHead
+      if dagHead.slot < dag.cfg.ALTAIR_FORK_EPOCH.start_slot:
+        fatal "Light syncing could not work"
```
etan-status:
```suggestion
        fatal "Light syncing requires a post-Altair state",
          headSlot = dagHead.slot, 
          altairStartSlot = dag.cfg.ALTAIR_FORK_EPOCH.start_slot
```

beacon_chain/winservice.nim
etan-status, line 1: Extractable to separate PR

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,521 @@
...
+    overseer.forwardSync.start()
+    # Starting backfill/backward sync manager.
+    if dag.needsBackfill():
+      asyncSpawn overseer.startBackfillTask()
+    return
+  else:
+    if dag.needsBackfill():
+      # Checkpoint/Trusted state we have is too old.
+      error "Trusted node sync started too long time ago"
+      quit 1
```
etan-status, line 432: It may be useful to have a separate exit code when user interaction is needed for recovery, i.e., this requires a manual re-sync to a newer trusted checkpoint and is not a configuration error or software bug.

beacon_chain/beacon_node_light_client.nim
```diff
@@ -92,6 +89,8 @@ proc initLightClient*(
                 .shouldSyncOptimistically(node.currentSlot):
               return
 
+            node.eventBus.optHeaderUpdateQueue.emit(optimisticHeader)
+
```
etan-status: When syncing via libp2p by range, responses are produced based on the current head branch as selected by fork choice. However, the optimistic headers may potentially pertain to a different branch, if there was a reorg.

Either, we have to handle the possibility of getting stuck, because we may have selected a block as our sync target that was orphaned and is no longer available via libp2p.

Or, alternatively, we could use the finalized header as our sync target. As long as there is no slashing, the protocol guarantees that there is only a single finalized branch within the weak subjectivity period. `LightClient` can be configured with an `onFinalizedHeader` callback as well.

beacon_chain/beacon_node_light_client.nim
```diff
@@ -92,6 +89,8 @@ proc initLightClient*(
                 .shouldSyncOptimistically(node.currentSlot):
               return
```
etan-status: The new sync mechanism works for all `lcDataFork > LightClientDataFork.None`, but here we have an additional `>= Capella` restriction, because that's required for the EL `blockHash` and `setOptimisticHead`.

All networks that we actively support have advanced past Capella, so letting this remain in the `>= Capella` section is also fine.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -449,3 +453,110 @@ proc addBackfillBlock*(
...
+    ) {.gcsafe, raises: [].} =
+      onBlockAdded(
+        blckRef,
+        ForkedTrustedSignedBeaconBlock.init(trustedBlock),
+        epochRef,
+        unrealized)
+
+    let blockHandler: BlockAdded(consensusFork) = onBlockAddedHandler
+
+    discard addResolvedHeadBlock(
```
etan-status, line 554: `addResolvedHeadBlock` calls `dag.processNewBlockForLightClient(state, trustedBlock, parent.bid)`, and that may trigger a beacon state root computation.

If `conf.lightClientDataImportMode` is set to `Full`, this is not avoidable, as the user explicitly indicates interest in full historical data availability. However, if it is set to `OnlyNew` or `OnDemand`, the new sync overseer could gain some performance by setting `dag.lcDataStore.cache.tailSlot` to the LC sync target so that the state root computation can be avoided until the sync has caught up to the LC target.

beacon_chain/beacon_chain_file.nim
etan-status, line 1: The general idea of downloading trust-minimized data into a separate location before processing it is great.

I wonder though, is there an overlap between this format and `.era`/`.erb` files?
And if no, is there an advantage of using a custom format over era?

etan-status (approved):
This is a great improvement for the security of default sync compared to the existing genesis sync.

As this is timely to properly test, would be great to have at least a happy case test with a few epochs worth of minimal blocks to avoid accidental regressions.

If data size is a concern, one may consider syncing 1 (or N) sync committee period (~27 hrs / 8192 slots) at a time by treating each intermediate light client sync step as its own separate sync target, applying each of them separately before syncing to the next period. The hook to obtain intermediate results is `LightClientUpdateObserver`. It would add quite a bit of complexity, though, so maybe waiting for demand is better for now.

beacon_chain/beacon_node_light_client.nim
cheatfate: If `syncManager` is initialized at this place, `syncOverseer` is initialized too.

etan-status: `syncManager` isn't, that's why there was the `node.syncManager != nil` check.

beacon_chain/sync/sync_overseer.nim
cheatfate: I'm not sure i understand the question, where you see rate limiting?

beacon_chain/sync/sync_overseer.nim
cheatfate: We did not use any other return codes except 1 for all the exit failures we have.

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/beacon_node_light_client.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/gossip_processing/block_processor.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/spec/signatures_batch.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/bbce4d33f8d1e2a4f6cd16e409d63677dc31be1d

beacon_chain/sync/sync_overseer.nim
etan-status: When requests fail, wondering whether there should be a delay between retries (as in, directly below the `while true` a token bucket. Not sure if it's necessary, just wanting to ensure it's really not needed.

beacon_chain/sync/sync_overseer.nim
etan-status: We also have a special exit code when the BN detects a slashing. But sure, maybe 1 is fine as well.

beacon_chain/beacon_chain_file.nim
cheatfate: ERA files is more about static storage of limited number of blocks + state, without any need to read it from the tail.
Current file and format is made with sense that we downloading files from the head to tail and applying from tail to head.

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f3451ad474fcaf6a7cfdb70cf3c827789ce87ab0

beacon_chain/beacon_node_light_client.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/f3451ad474fcaf6a7cfdb70cf3c827789ce87ab0

beacon_chain/sync/sync_overseer.nim
cheatfate: It just a warning, which will be printed if it finds blockchain-list file in database directory, and its head will be out of WSP. If you supply `Lenient` mode it will start genesis sync.

beacon_chain/consensus_object_pools/block_clearance.nim
cheatfate: I just want to understand this comment, by default `conf.lightClientDataImportMode` is `OnlyNew`, and we already using `dag.lcDataStore.cache.tailSlot` to be set  https://github.com/status-im/nimbus-eth2/pull/6515/files#diff-f22697df9a71937c613670a8034b67d712074868b5da6bb0973b58b665d16993R323-R324 here.

etan-status: When rebasing, deep links break on github, not sure where it's pointing.

`dag.lcDataStore.cache.tailSlot` defines the minimum slot from which light client data is collected to serve to others. With `OnlyNew` (the default) and `OnDemand`, data collection is opportunistic and runs only when it is efficient to do so. Because data collection involves computing the `BeaconState` root, it may be necessary to adjust `dag.lcDataStore.cache.tailSlot` higher so that expensive state root computations are not run when running this new sync mode. With `Full`, the user wants full data (even when it slows down regular operation) and we have to do the state root computations while syncing.

`tailSlot` in `beacon_chain_file.nim` appears to have different semantics from `dag.lcDataStore.cache.tailSlot`.

cheatfate: I'm talking about this line in `sync_overseer.nim`.
```nim
  # Set minimum slot number from which LC data is collected.
  dag.lcDataStore.cache.tailSlot = clist.head.get().slot
```

etan-status: ooooh! I missed that one. yes, with this it should already be ok :) thanks for pointing it out!

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -406,6 +406,102 @@ proc enqueueBlock*(
...
+      # However, in practice, an EL client may not have completed importing all
+      # block headers, so may be unaware of a block's ancestor status.
+      # Therefore, hopping back and forth between the optimistic head and the
+      # chain DAG head does not work well in practice, e.g., Geth:
+      # - "Beacon chain gapped" from DAG head to optimistic head,
+      # - followed by "Beacon chain reorged" from optimistic head back to DAG.
+      consensusManager[].updateHead(newHead.get.blck)
+
+      template callForkchoiceUpdated(attributes: untyped) =
+        if NewPayloadStatus.noResponse != payloadStatus and
```
tersec: https://github.com/status-im/nimbus-eth2/blob/fac3f12b1611a9b472aaab2095b3d47819c1a284/beacon_chain/sync/sync_overseer.nim#L264-L266
always calls this function with `NewPayloadStatus.noResponse` so this `callForkchoiceUpdated` resolves to a no-op as used in this PR. There aren't any other users of this new function, so just dead code at the moment.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -406,6 +406,102 @@ proc enqueueBlock*(
...
+      withConsensusFork(consensusFork):
+        when consensusFork >= ConsensusFork.Bellatrix:
+          callForkchoiceUpdated(consensusFork.PayloadAttributes)
+    else:
+      let headExecutionBlockHash =
+        consensusManager.dag.loadExecutionBlockHash(
+          newHead.get.blck).get(ZERO_HASH)
+
+      if headExecutionBlockHash.isZero or
+         NewPayloadStatus.noResponse == payloadStatus:
```
tersec: Likewise here,
https://github.com/status-im/nimbus-eth2/blob/fac3f12b1611a9b472aaab2095b3d47819c1a284/beacon_chain/sync/sync_overseer.nim#L264-L266
means that as `sync_overseer` calls this function, it amounts to an indirect way of calling `consensusManager[].updateHead(...)`.

beacon_chain/spec/signatures_batch.nim
```diff
@@ -229,6 +229,39 @@ func bls_to_execution_change_signature_set*(
...
+    fork = getStateField(state, fork)
+    genesis_validators_root = getStateField(state, genesis_validators_root)
+
+  for forkedBlock in blocks:
+    let item =
+      withBlck(forkedBlock):
+        let
+          proposerKey =
+            validatorKeys.load(forkyBlck.message.proposer_index).valueOr:
+              let msg = "collectSignatureSets: invalid proposer index (" &
```
tersec: function name in `msg` doesn't reflect `collectProposerSignatureSet`

beacon_chain/spec/signatures_batch.nim
```diff
@@ -229,6 +229,39 @@ func bls_to_execution_change_signature_set*(
...
+                        $forkyBlck.message.proposer_index & ")"
+              return err(msg)
+          signature =
+            forkyBlck.signature.load().valueOr:
+              let msg = "collectSignatureSets: cannot load signature (" &
```
tersec: function name in `msg` doesn't reflect `collectProposerSignatureSet`

beacon_chain/sync/sync_manager.nim
```diff
@@ -269,8 +282,96 @@ func checkBlobs(blobs: seq[BlobSidecars]): Result[void, string] =
...
+          res.get().asSeq()
+
+      debug "Received blobs on request",
+            blobs_count = len(blobData),
+            blobs_map = getShortMap(req, blobData), request = req
+
+      if len(blobData) > 0:
+        let
+          slots = mapIt(blobData, it[].signed_block_header.message.slot)
+          uniqueSlots = foldl(slots, combine(a, b), @[slots[0]])
```
tersec: This `foldl(..., combine, ...)` is essentially O(n^2) in number of slots, where per https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/deneb/p2p-interface.md#configuration `slots` might be up to `MAX_REQUEST_BLOB_SIDECARS == MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK == 128 * 6 == 768` items long, which is a bit much when it's an intrinsically O(n) task it's performing.
https://github.com/status-im/nimbus-eth2/blob/fac3f12b1611a9b472aaab2095b3d47819c1a284/beacon_chain/sync/sync_manager.nim#L181-L185

Some observations:
- it always copies the list (this is part of the overall O(n^2)), even though per https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/deneb/p2p-interface.md#blobsidecarsbyrange-v1 "blob sidecars, where they exist, MUST be sent in consecutive `(slot, index)` order" i.e. most of the time it's going to do nothing but copy a list and the not modify it;
- the `slots` `seq` is only ever used essentially as an iterable, but still gets created in full, also with up to 768 8-byte `uint64`s.

All of these points apply to the other user of this function at
https://github.com/status-im/nimbus-eth2/blob/fac3f12b1611a9b472aaab2095b3d47819c1a284/beacon_chain/sync/sync_manager.nim#L553-L554
also, incidentally, so it's least not new.

beacon_chain/sync/sync_manager.nim
```diff
@@ -269,8 +282,96 @@ func checkBlobs(blobs: seq[BlobSidecars]): Result[void, string] =
...
+
+  logScope:
+    peer_score = peer.getScore()
+    peer_speed = peer.netKbps()
+    topics = "syncman"
+
+  doAssert(not(req.isEmpty()), "Request must not be empty!")
+
+  debug "Requesting blocks from peer", request = req
+  let blocksRange =
```
tersec, line 298: This is all remarkably similar to https://github.com/status-im/nimbus-eth2/blob/fac3f12b1611a9b472aaab2095b3d47819c1a284/beacon_chain/sync/sync_manager.nim#L300-L351

beacon_chain/sync/sync_manager.nim
```diff
@@ -269,8 +282,96 @@ func checkBlobs(blobs: seq[BlobSidecars]): Result[void, string] =
       ? blob_sidecar[].verify_blob_sidecar_inclusion_proof()
   ok()
 
-proc syncStep[A, B](man: SyncManager[A, B], index: int, peer: A)
-    {.async: (raises: [CancelledError]).} =
+proc getSyncBlockData*[T](
+    peer: T,
+    req: SyncRequest,
+    blobsPresent: bool
```
tersec: This is only ever called with `blobsPresent == true`:
https://github.com/status-im/nimbus-eth2/blob/fac3f12b1611a9b472aaab2095b3d47819c1a284/beacon_chain/sync/sync_overseer.nim#L93-L96

beacon_chain/sync/sync_manager.nim
```diff
@@ -269,8 +282,96 @@ func checkBlobs(blobs: seq[BlobSidecars]): Result[void, string] =
...
+        blocks_map = getShortMap(req, blocksRange), request = req
+
+  let slots = mapIt(blocksRange, it[].slot)
+
+  if not(checkResponse(req, slots)):
+    peer.updateScore(PeerScoreBadResponse)
+    return err("Received blocks sequence is not in requested range")
+
+  let shouldGetBlobs =
+    if not(blobsPresent):
```
tersec: This branch can't ever be taken

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+import std/[strutils, sequtils]
+import stew/base10, chronos, chronicles, results
+import
+  ../consensus_object_pools/blockchain_list,
+  ../spec/datatypes/[phase0, altair],
```
tersec: If importing `spec/forks`, it's unnecessary to also import `spec/datatypes/foo`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+      blob =
+        if res.blobs.isSome():
+          Opt.some(res.blobs.get()[0])
+        else:
+          Opt.none(BlobSidecars)
+    ok(BlockData(blck: res.blocks[0][], blob: blob))
+  finally:
+    overseer.pool.release(peer)
+
+# proc `==`(a, b: BeaconBlockHeader): bool =
```
tersec: Is this commented-out function useful/necessary?

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+      block:
+        overseer.clist.seekForSlot(dag.head.slot).isOkOr:
+          fatal "Unable to find slot in backfill data", reason = error,
+                path = overseer.clist.path
+          quit 1
+        overseer.clist
+
+  var
+    blocks: seq[BlockData]
+    processEpoch: Epoch = FAR_FUTURE_EPOCH
```
tersec: `state_transition_epoch` has a function named `process_epoch`, so for the sake of not confusing Nim's stylecheck hints, it would be nicer to pick some other name than `processEpoch` here.

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+from ../consensus_object_pools/spec_cache import get_attesting_indices
+
+export sync_types
+
+logScope:
+  topics = "overseer"
+
+const
+  PARALLEL_REQUESTS* = 3
+    ## Number of peers we using to resolve our request.
```
tersec: either
- "Number of peers we use to ..."; or
- "Number of peers we're using to ..."

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+  BLOCKS_PROCESS_CHUNK_SIZE* = 2
+    ## Number of blocks sent to processing (CPU heavy task).
+
+type
+  BlockDataRes* = Result[BlockData, string]
```
tersec: This `BlockDataRes` type does not appear to be referenced by name outside this module, not sure why it's exported

https://status-im.github.io/nim-style-guide/language.proc.html

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+
+export sync_types
+
+logScope:
+  topics = "overseer"
+
+const
+  PARALLEL_REQUESTS* = 3
+    ## Number of peers we using to resolve our request.
+  BLOCKS_PROCESS_CHUNK_SIZE* = 2
```
tersec: only reference is in `for bchunk in blocks.chunks(onStateUpdate, BLOCKS_PROCESS_CHUNK_SIZE)`, no need to export

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -809,6 +861,16 @@ proc init*(T: type BeaconNode,
 
     getBeaconTime = beaconClock.getBeaconTimeFn()
 
+  let clist =
+    block:
+      # TODO (cheatfate): We should reset (delete) blockchain file if tail is
+      # not in weak subjectivity period.
+      let
+        res = ChainListRef.init(config.databaseDir())
+      info "Backfill database has been loaded", path = config.databaseDir(),
```
tersec: Not really sure why the `info` is placed inside the `let`?

beacon_chain/consensus_object_pools/blockchain_list.nim
```diff
@@ -0,0 +1,243 @@
...
+    store(clist.handle.get(), signedBlock, blobs)
+
+proc checkBlobs(signedBlock: ForkedSignedBeaconBlock,
+                blobsOpt: Opt[BlobSidecars]): Result[void, VerifierError] =
+  withBlck(signedBlock):
+    when consensusFork >= ConsensusFork.Deneb:
+      if blobsOpt.isSome():
+        let
+          blobs = blobsOpt.get()
+          commits = forkyBlck.message.body.blob_kzg_commitments.asSeq
```
tersec: This is a kind of pointless copy of `forkyBlck.message.body.blob_kzg_commitments`. Yes, `validate_blobs` wants a `seq`, but `asSeq` is just an l-value-preserving `distinct`-removal so if one just calls (either via a `template` shortener or directly) `validate_blobs(forkyBlck.message.body.blob_kzg_commitments.asSeq, ...)` it accomplishes the same thing without the copy.

beacon_chain/consensus_object_pools/blockchain_list.nim
```diff
@@ -0,0 +1,243 @@
...
+
+        if len(blobs) > 0 or len(commits) > 0:
```
tersec: `len(blobs) == len(commits)` or it's definitely not valid per https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/cancun.md#specification-2
> 4. The call **MUST** return `blobs` and `proofs` that match the `commitments` list, i.e. `assert len(blobsBundle.commitments) == len(blobsBundle.blobs) == len(blobsBundle.proofs)` and `assert verify_blob_kzg_proof_batch(blobsBundle.blobs, blobsBundle.commitments, blobsBundle.proofs)`.

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,953 @@
...
+    maskKind(getBlobForkCode(kind))
+  else:
+    getBlobForkCode(kind)
+
+proc getBlockConsensusFork(header: ChainFileHeader): ConsensusFork =
+  let hkind = unmaskKind(header.kind)
+  if int(hkind) in BlockForkCodeRange:
+    cast[ConsensusFork](hkind)
+  else:
+    raiseAssert("Should not be happened")
```
tersec: `Should not happen`

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+            if res.isErr():
+              let msg = "Unable to add block data to database [" &
+                        $res.error & "]"
+              bchunk.resfut.complete(Result[void, string].err(msg))
+              break innerLoop
+
+          withBlck(bdata.blck):
+            let res =
+              try:
+                await updateHead(consensusManager, validatorMonitor,
```
tersec: Because (see elsewhere) this can't trigger FCU, it's also not actually asynchronous, no `await`, no `CancelledError`, etc

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+        if len(blocks) != 0:
+          let
+            startTick = Moment.now()
+            blocksOnly = blocks.mapIt(it.blck)
+
+          proc onStateUpdate(slot: Slot): Result[void, VerifierError] {.
+               gcsafe, raises: [].} =
+
+            if slot != blocksOnly[0].slot:
+              # We going to verify signatures only at the beginning of
```
tersec: "We're going to ..."

tersec: also why?

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+  "[" & map & "]:" & futureState
+
+iterator chunks*(data: openArray[BlockData],
+                 stateCallback: OnStateUpdated,
+                 maxCount: Positive): BlockDataChunk =
+  for i in countup(0, len(data) - 1, maxCount):
+    yield BlockDataChunk.init(stateCallback,
+      data.toOpenArray(i, min(i + maxCount, len(data)) - 1))
+
+proc getLatestBeaconHeader*(
```
tersec: Only used by `let blockHeader = await overseer.getLatestBeaconHeader()` in module; no need to export

https://status-im.github.io/nim-style-guide/language.proc.html

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+        results.add(workers[i].value.get())
+
+    if blockHeader.isSome:
+      if len(results) > 0:
+        for item in results:
+          withBlck(item.blck):
+            if forkyBlck.message.toBeaconBlockHeader() == blockHeader.get():
+              return item
+    else:
+      # TODO (cheatfate): Compare received blocks
```
tersec: this seems important?

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+    except AsyncEventQueueFullError:
+      raiseAssert "AsyncEventQueueFullError should not happen!"
+
+  withForkyHeader(events[^1]):
+    when lcDataFork > LightClientDataFork.None:
+      forkyHeader.beacon
+    else:
+      raiseAssert "Should not happen"
+
+proc getPeerBlock*(
```
tersec: Only used by `let blockHeader = await overseer.getLatestBeaconHeader()` in same module; no need to export

https://status-im.github.io/nim-style-guide/language.proc.html

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+              return item
+    else:
+      # TODO (cheatfate): Compare received blocks
+      if len(results) > 0:
+        return results[0]
+
+    # Wait for 2 seconds before trying one more time.
+    await sleepAsync(2.seconds)
+
+proc isWithinWeakSubjectivityPeriod(
```
tersec, line 137: `isWithinWeakSubjectivityPeriod` and `is_within_weak_subjectivity_period` next to each other confuse Nim's style checker

beacon_chain/sync/sync_manager.nim
tersec: Also, it's only ever called with `count == 1`, i.e. a single block. But the whole thing still has the copy/paste structure of a full-on general-purpose multiple block sync function.

beacon_chain/sync/sync_manager.nim
```diff
@@ -269,8 +282,96 @@ func checkBlobs(blobs: seq[BlobSidecars]): Result[void, string] =
...
+      res.get().asSeq
+
+  debug "Received blocks on request", blocks_count = len(blocksRange),
+        blocks_map = getShortMap(req, blocksRange), request = req
+
+  let slots = mapIt(blocksRange, it[].slot)
+
+  if not(checkResponse(req, slots)):
+    peer.updateScore(PeerScoreBadResponse)
+    return err("Received blocks sequence is not in requested range")
```
tersec: there shouldn't really be a "blocks sequence" I think? just one block

beacon_chain/sync/sync_overseer.nim
```diff
@@ -0,0 +1,526 @@
...
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+import std/[strutils, sequtils]
+import stew/base10, chronos, chronicles, results
+import
+  ../consensus_object_pools/blockchain_list,
+  ../spec/datatypes/[phase0, altair],
+  ../spec/eth2_apis/rest_types,
```
tersec, line 15: how are REST JSON types relevant here? this function should be dealing in 100% SSZ

tersec, line 15: how are REST JSON types relevant here? this module should be dealing in 100% SSZ

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,953 @@
...
+      int(res)
+  if (hkind in BlockForkCodeRange) or (hkind in BlobForkCodeRange):
+    ok()
+  else:
+    err("Unsuppoted chunk kind value")
+
+proc check(a: ChainFileHeader): Result[void, string] =
+  if a.header != ChainFileHeaderValue:
+    return err("Incorrect chunk header [NIMB]")
+  if a.version != 1'u32:
```
tersec: Is this intentionally not comparing with `ChainFileVersion`?

beacon_chain/beacon_chain_file.nim
```diff
@@ -0,0 +1,953 @@
...
+  int(hkind) in BlockForkCodeRange
+
+template isBlob(h: ChainFileHeader | ChainFileFooter): bool =
+  let hkind = unmaskKind(h.kind)
+  int(hkind) in BlobForkCodeRange
+
+template isLast(h: ChainFileHeader | ChainFileFooter): bool =
+  h.kind.isLast()
+
+proc isFilePresent*(filename: string): bool =
```
tersec: What does this wrapper function add?

etan-status (approved):

etan-status:
I think it's good to have, as an improvement over genesis sync when `--external-beacon-api-url` is not provided (pure decentralized mode). By default this is only used on Holesky and when syncing from outdated database, so the risk of accidental interference with validator duties is minimal.

The latest changes make it track the finalized checkpoint, meaning no reorgs are expected, so the branch being backfilled is expected to remain available during the sync.

One remaining item could be file system performance when lots of blocks are downloaded. Maybe it needs subdirectories if millions of blocks are queued up, or an incremental sync that only does 1 sync committee period at a time. But such changes could be introduced later if needed.

beacon_chain/sync/sync_manager.nim
cheatfate: Why do you think so? We always requested and i think we will continue to request range of blocks we just processing it one by one.

beacon_chain/spec/signatures_batch.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/spec/signatures_batch.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/nimbus_beacon_node.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/consensus_object_pools/blockchain_list.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/consensus_object_pools/blockchain_list.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/commit/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/commit/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/beacon_chain_file.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/commit/646bef4e79ceed8a1d41bb46fb90729e60302bbf

beacon_chain/sync/sync_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/d71139db1343167811d0f2feefe3dd453f6d1d99

beacon_chain/winservice.nim
cheatfate: This code was removed and merged already into `unstable`.

beacon_chain/beacon_chain_file.nim
cheatfate: Wrapper was removed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/157416a14631f5e95cc9d189e50fa8e975346e62

beacon_chain/sync/sync_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/87682a8ebda24ff3f3b84a5ca3cd604f82303e0f but you should check.

tersec: Yeah, seems fine now.

It's maybe useful to have this `uniqueSlots` thing but probably not that useful, so better to use the simpler version, as this does.

cheatfate: I think `checkResponse` is enough to check that list has unique slots.
I have added tests to keep this functionality in place. https://github.com/status-im/nimbus-eth2/pull/6515/commits/f45978b5a11d4118663c4536e89f6780560df605

beacon_chain/spec/signatures_batch.nim
```diff
@@ -229,6 +229,41 @@ func bls_to_execution_change_signature_set*(
...
+proc collectProposerSignatureSet*(
+  sigs: var seq[SignatureSet],
+  blocks: openArray[ForkedSignedBeaconBlock],
+  validatorKeys: openArray[ImmutableValidatorData2],
+  state: ForkedHashedBeaconState
+): Result[void, string] =
+  mixin load
+
+  let
+    fork = getStateField(state, fork)
```
tersec: The only thing this function does with `state` is get the `fork` and `genesis_validator_root`. It would clarify the dataflow and actual dependencies to pass these explicitly, because `state` is basically the heaviest object in the application, for fields totaling about 40 bytes.

It should pass it efficiently enough, it's just easier to reason about knowing that the rest of the `state` can't possibly affect it.

beacon_chain/gossip_processing/block_processor.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/d3d65143222bba9fb417a8218948972cfbdef548

beacon_chain/gossip_processing/block_processor.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/d3d65143222bba9fb417a8218948972cfbdef548

beacon_chain/sync/sync_overseer.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/d3d65143222bba9fb417a8218948972cfbdef548

beacon_chain/spec/signatures_batch.nim
cheatfate: Have you seen this file? whole file is accepting `state` as parameter, do you want me to change whole file?

tersec: The only use of `BeaconState` I can find in the file is
https://github.com/status-im/nimbus-eth2/blob/f54a0366abfba26c7694fad1e14350d19aa0a228/beacon_chain/spec/signatures_batch.nim#L232-L239

and that's necessary because
https://github.com/status-im/nimbus-eth2/blob/f54a0366abfba26c7694fad1e14350d19aa0a228/beacon_chain/spec/signatures_batch.nim#L371-L386
and
https://github.com/status-im/nimbus-eth2/blob/f54a0366abfba26c7694fad1e14350d19aa0a228/beacon_chain/spec/signatures_batch.nim#L422-L432
for example.

It would be possible to split all those things out, but it's probably better to just provide the state.

There's not some overarching convention in `signatures_batch` of the state though, only that one other `collectSignatureSets` call.

Here, it's just those two fields, one of which is a constant for any given network (`genesis_validators_root`).

So, no, I'm not suggesting changing the whole file, or anything outside this PR.

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6515/commits/fe6452e7b2c0ebd985ffcde3aca35c6303723b2e

PR #6524, opened 2024-08-28
bump nim-libp2p to `v1.5.0`

etan-status:
- chore: setting dialing canceled log to trace
- fix(ci): windows-amd64 (Nim version-1-6)
- fix: add gcc 14 support
- feat: adding onValidated observer
- fix(test): interop transport
- chore: Cleanup CI
- fix: infinite loop when connection is aborted before being accepted
- fix(yamux): future leak
- chore(version): update libp2p.nimble to 1.5.0


tersec:
Will wait until closer to 24.8 release

PR #6532, opened 2024-09-02
add: `get_custody_columns` for das_core

agnxsh:
This Pr adds `get_custody_columns` and it's adjoining helper functions, along with _networking_ fixture tests provided by EF

Ref: 
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#get_custody_columns
- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/tests/formats/networking/get_custody_columns.md

### DIFF
--- a/beacon_chain/spec/datatypes/eip7594.nim
+++ b/beacon_chain/spec/datatypes/eip7594.nim
@@ -24,9 +24,12 @@ const
     # The number of bytes in a cell |
   CELLS_PER_EXT_BLOB* = FIELD_ELEMENTS_PER_EXT_BLOB div FIELD_ELEMENTS_PER_CELL
     # The number of cells in an extended blob |
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/p2p-interface.md#preset
   KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH* = 4
 
 type
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/polynomial-commitments-sampling.md#custom-types 
   BLSFieldElement* = KzgBytes32
   G2Point* = array[96, byte]
   PolynomialCoeff* = List[BLSFieldElement, FIELD_ELEMENTS_PER_EXT_BLOB]
@@ -35,24 +38,27 @@ type
   Cell* = KzgCell
   Cells* = KzgCells
   CellsAndProofs* = KzgCellsAndKzgProofs
-  CellID* = uint64
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#custom-types
   RowIndex* = uint64
   ColumnIndex* = uint64
   CellIndex* = uint64
 
 const
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#data-size
   NUMBER_OF_COLUMNS* = 128
-  MAX_CELLS_IN_EXTENDED_MATRIX* = MAX_BLOBS_PER_BLOCK * NUMBER_OF_COLUMNS
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#networking
   DATA_COLUMN_SIDECAR_SUBNET_COUNT* = 128
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#custody-setting
   SAMPLES_PER_SLOT* = 8
   CUSTODY_REQUIREMENT* = 4
-  TARGET_NUMBER_OF_PEERS* = 70
 
 type
   DataColumn* = List[KzgCell, Limit(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
-  ExtendedMatrix* = List[KzgCell, Limit(MAX_CELLS_IN_EXTENDED_MATRIX)]
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#datacolumnsidecar
   DataColumnSidecar* = object
     index*: ColumnIndex # Index of column in extended matrix
     column*: DataColumn
@@ -64,10 +70,12 @@ type
 
   DataColumnSidecars* = seq[ref DataColumnSidecar]
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/p2p-interface.md#datacolumnidentifier
   DataColumnIdentifier* = object
     block_root*: Eth2Digest
     index*: ColumnIndex
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#matrixentry
   MatrixEntry* = object
     cell*: Cell
     kzg_proof*: KzgProof
@@ -79,7 +87,7 @@ type
 
   CscCount* = uint8
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.4/specs/_features/eip7594/p2p-interface.md#metadata
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/p2p-interface.md#metadata
   MetaData* = object
     seq_number*: uint64
     attnets*: AttnetBits
--- a/beacon_chain/spec/datatypes/eip7594.nim
+++ b/beacon_chain/spec/datatypes/eip7594.nim
@@ -1,0 +1,101 @@
+# beacon_chain
+# Copyright (c) 2022-2024 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+import
+  std/[sequtils],
+  "."/[altair, base, deneb], 
+  kzg4844/[kzg, kzg_abi]
+
+from std/strutils import join
+
+export base
+
+const
+  FIELD_ELEMENTS_PER_EXT_BLOB* = 2 * kzg_abi.FIELD_ELEMENTS_PER_BLOB
+    # Number of field elements in a Reed-Solomon extended blob |
+  FIELD_ELEMENTS_PER_CELL* = 64 # Number of field elements in a cell |
+  BYTES_PER_CELL* = FIELD_ELEMENTS_PER_CELL * kzg_abi.BYTES_PER_FIELD_ELEMENT
+    # The number of bytes in a cell |
+  CELLS_PER_EXT_BLOB* = FIELD_ELEMENTS_PER_EXT_BLOB div FIELD_ELEMENTS_PER_CELL
+    # The number of cells in an extended blob |
+  KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH* = 4
+
+type
+  BLSFieldElement* = KzgBytes32
+  G2Point* = array[96, byte]
+  PolynomialCoeff* = List[BLSFieldElement, FIELD_ELEMENTS_PER_EXT_BLOB]
+  Coset* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
+  CosetEvals* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
+  Cell* = KzgCell
+  Cells* = KzgCells
+  CellsAndProofs* = KzgCellsAndKzgProofs
+  CellID* = uint64
+  RowIndex* = uint64
+  ColumnIndex* = uint64
+  CellIndex* = uint64
+
+const
+  NUMBER_OF_COLUMNS* = 128
+  MAX_CELLS_IN_EXTENDED_MATRIX* = MAX_BLOBS_PER_BLOCK * NUMBER_OF_COLUMNS
+
+  DATA_COLUMN_SIDECAR_SUBNET_COUNT* = 128
+  SAMPLES_PER_SLOT* = 8
+  CUSTODY_REQUIREMENT* = 4
+  TARGET_NUMBER_OF_PEERS* = 70
+
+type
+  DataColumn* = List[KzgCell, Limit(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
+  ExtendedMatrix* = List[KzgCell, Limit(MAX_CELLS_IN_EXTENDED_MATRIX)]
+
+  DataColumnSidecar* = object
+    index*: ColumnIndex # Index of column in extended matrix
+    column*: DataColumn
+    kzg_commitments*: KzgCommitments
+    kzg_proofs*: KzgProofs
+    signed_block_header*: SignedBeaconBlockHeader
+    kzg_commitments_inclusion_proof*:
+      array[KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH, Eth2Digest]
+
+  DataColumnSidecars* = seq[ref DataColumnSidecar]
+
+  DataColumnIdentifier* = object
+    block_root*: Eth2Digest
+    index*: ColumnIndex
+
+  MatrixEntry* = object
+    cell*: Cell
+    kzg_proof*: KzgProof
+    column_index*: ColumnIndex
+    row_index*: RowIndex
+  
+  # Not in spec, defined in order to compute custody subnets
+  CscBits* = BitArray[DATA_COLUMN_SIDECAR_SUBNET_COUNT]
+
+  CscCount* = uint8
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.4/specs/_features/eip7594/p2p-interface.md#metadata
+  MetaData* = object
+    seq_number*: uint64
+    attnets*: AttnetBits
+    syncnets*: SyncnetBits
+    custody_subnet_count*: CscCount 
+
+func shortLog*(v: DataColumnSidecar): auto =
+  (
+    index: v.index,
+    kzg_commitments: v.kzg_commitments.len,
+    kzg_proofs: v.kzg_proofs.len,
+    block_header: shortLog(v.signed_block_header.message),
+  )
+
+func shortLog*(v: seq[DataColumnSidecar]): auto =
+  "[" & v.mapIt(shortLog(it)).join(", ") & "]"
+
+func shortLog*(x: seq[DataColumnIdentifier]): string =
+  "[" & x.mapIt(shortLog(it.block_root) & "/" & $it.index).join(", ") & "]"
--- a/beacon_chain/spec/datatypes/eip7594.nim
+++ b/beacon_chain/spec/datatypes/eip7594.nim
@@ -17,13 +17,14 @@ from std/strutils import join
 export base
 
 const
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/polynomial-commitments-sampling.md#cells
   FIELD_ELEMENTS_PER_EXT_BLOB* = 2 * kzg_abi.FIELD_ELEMENTS_PER_BLOB
-    # Number of field elements in a Reed-Solomon extended blob |
+  # Number of field elements in a Reed-Solomon extended blob |
   FIELD_ELEMENTS_PER_CELL* = 64 # Number of field elements in a cell |
   BYTES_PER_CELL* = FIELD_ELEMENTS_PER_CELL * kzg_abi.BYTES_PER_FIELD_ELEMENT
-    # The number of bytes in a cell |
+  # The number of bytes in a cell |
   CELLS_PER_EXT_BLOB* = FIELD_ELEMENTS_PER_EXT_BLOB div FIELD_ELEMENTS_PER_CELL
-    # The number of cells in an extended blob |
+  # The number of cells in an extended blob |
 
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/p2p-interface.md#preset
   KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH* = 4
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -11,14 +11,9 @@
 import
   std/[algorithm, hashes],
   results,
-  kzg4844/[kzg, kzg_abi],
   eth/p2p/discoveryv5/[node],
   ./[helpers, digest],
-  ./datatypes/[eip7594, deneb],
-  ssz_serialization/[
-    proofs,
-    types],
-  ./[beacon_time, crypto]
+  ./datatypes/[eip7594]
 
 proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
                           subnetIds: HashSet[uint64]):
@@ -32,8 +27,8 @@ proc sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
   res
 
 proc sortedColumnIndexList*(columnsPerSubnet: ColumnIndex, 
-                          subnetIds: HashSet[uint64]): 
-                          List[ColumnIndex, NUMBER_OF_COLUMNS] =
+                            subnetIds: HashSet[uint64]): 
+                            List[ColumnIndex, NUMBER_OF_COLUMNS] =
   var
     res: seq[ColumnIndex]
     list: List[ColumnIndex, NUMBER_OF_COLUMNS]
@@ -46,7 +41,7 @@ proc sortedColumnIndexList*(columnsPerSubnet: ColumnIndex,
     discard list.add(ColumnIndex(elem))
   list
 
-proc get_custody_column_subnet*(node_id: NodeId, 
+proc get_custody_column_subnets*(node_id: NodeId, 
                                 custody_subnet_count: uint64): 
                                 Result[HashSet[uint64], cstring] =
 
@@ -61,7 +56,7 @@ proc get_custody_column_subnet*(node_id: NodeId,
     subnet_ids: HashSet[uint64]
     current_id = node_id
 
-  while subnet_ids.len < int(custody_subnet_count):
+  while subnet_ids.lenu64 < custody_subnet_count:
     var
       current_id_bytes: array[32, byte]
       hashed_bytes: array[8, byte]
@@ -73,11 +68,10 @@ proc get_custody_column_subnet*(node_id: NodeId,
       hashed_current_id = eth2digest(current_id_bytes)
       
     hashed_bytes[0..7] = hashed_current_id.data.toOpenArray(0,7)
-    var subnet_id = bytes_to_uint64(hashed_bytes) mod 
+    let subnet_id = bytes_to_uint64(hashed_bytes) mod 
       DATA_COLUMN_SIDECAR_SUBNET_COUNT
     
-    if subnet_id notin subnet_ids:
-      subnet_ids.incl(subnet_id)
+    discard subnet_ids.containsOrIncl(subnet_id)
 
     if current_id == UInt256.high.NodeId:
       # Overflow prevention
@@ -92,7 +86,8 @@ proc get_custody_columns*(node_id: NodeId,
                           seq[ColumnIndex] =
   let
     subnet_ids = 
-      get_custody_column_subnet(node_id, custody_subnet_count).get
+      get_custody_column_subnets(node_id, custody_subnet_count).get
+  const
     columns_per_subnet = 
       NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
 
@@ -107,7 +102,8 @@ proc get_custody_column_list*(node_id: NodeId,
   # before sending, data_column_sidecars_by_range requests
   let
     subnet_ids = 
-      get_custody_column_subnet(node_id, custody_subnet_count).get
+      get_custody_column_subnets(node_id, custody_subnet_count).get
+  const
     columns_per_subnet = 
       NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
   
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -9,21 +9,20 @@
 
 # Uncategorized helper functions from the spec
 import
-  std/[algorithm],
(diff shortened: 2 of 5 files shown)


beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,116 @@
...
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+# Uncategorized helper functions from the spec
+import
+  std/[algorithm],
+  results,
+  sequtils,
```
tersec: `sequtils` and `hashes` are also part of Nim's stdlib and should be prefixed by `std/`, just like `algorithms` (can be in same list, if no other preference alphabetically is nice as a convention)

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,116 @@
...
+    
+  let subnet_ids = get_custody_column_subnet(node_id, custody_subnet_count).get
+
+  # columns_per_subnet = NUMBER_OF_COLUMNS // DATA_COLUMN_SIDECAR_SUBNET_COUNT
+  let columns_per_subnet = NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
+
+  ok(sortedColumnIndices(ColumnIndex(columns_per_subnet), subnet_ids))
+
+
+proc get_custody_column_list*(node_id: NodeId,
```
tersec, line 92: Can this function fail? I'm not sure why it's a Result[...] really, since it only ever returns `ok(...)`

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,116 @@
...
+
+    if current_id == UInt256.high.NodeId:
+        # Overflow prevention
+        current_id = NodeId(StUint[256].zero)
+    current_id += NodeId(StUint[256].one)
+
+  ok(subnet_ids)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#get_custody_columns
+proc get_custody_columns*(node_id: NodeId,
```
tersec, line 79: can be `func` (like `get_custody_column_list`) but more importantly, why does it return `Result[...]` when it's only ever an `ok(...)`? Seems like it could return just `seq[ColumnIndex]`

beacon_chain/spec/eip7594_helpers.nim
agnxsh: Ok

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,116 @@
...
+  ok(subnet_ids)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#get_custody_columns
+proc get_custody_columns*(node_id: NodeId, 
+                          custody_subnet_count: uint64): 
+                          Result[seq[ColumnIndex], cstring] =
+    
+  let subnet_ids = get_custody_column_subnet(node_id, custody_subnet_count).get
+
+  # columns_per_subnet = NUMBER_OF_COLUMNS // DATA_COLUMN_SIDECAR_SUBNET_COUNT
```
tersec: This comment from the Python code is close enough to the Nim it's probably unnecessary

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  DATA_COLUMN_SIDECAR_SUBNET_COUNT* = 128
+  SAMPLES_PER_SLOT* = 8
+  CUSTODY_REQUIREMENT* = 4
+  TARGET_NUMBER_OF_PEERS* = 70
+
+type
+  DataColumn* = List[KzgCell, Limit(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
+  ExtendedMatrix* = List[KzgCell, Limit(MAX_CELLS_IN_EXTENDED_MATRIX)]
+
+  DataColumnSidecar* = object
```
tersec, line 63: Include link to https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#datacolumnsidecar

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+    kzg_commitments_inclusion_proof*:
+      array[KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH, Eth2Digest]
+
+  DataColumnSidecars* = seq[ref DataColumnSidecar]
+
+  DataColumnIdentifier* = object
+    block_root*: Eth2Digest
+    index*: ColumnIndex
+
+  MatrixEntry* = object
```
tersec, line 80: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#matrixentry

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  BLSFieldElement* = KzgBytes32
+  G2Point* = array[96, byte]
+  PolynomialCoeff* = List[BLSFieldElement, FIELD_ELEMENTS_PER_EXT_BLOB]
+  Coset* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
+  CosetEvals* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
+  Cell* = KzgCell
+  Cells* = KzgCells
+  CellsAndProofs* = KzgCellsAndKzgProofs
+  CellID* = uint64
+  RowIndex* = uint64
```
tersec, line 44: Should be a comment linking `RowIndex` and `ColumnIndex` being `uint64` to https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#custom-types

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  ColumnIndex* = uint64
+  CellIndex* = uint64
+
+const
+  NUMBER_OF_COLUMNS* = 128
+  MAX_CELLS_IN_EXTENDED_MATRIX* = MAX_BLOBS_PER_BLOCK * NUMBER_OF_COLUMNS
+
+  DATA_COLUMN_SIDECAR_SUBNET_COUNT* = 128
```
tersec, line 53: from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#networking

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  SAMPLES_PER_SLOT* = 8
```
tersec, line 56: `SAMPLES_PER_SLOT` and `CUSTODY_REQUIREMENT` from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#custody-setting

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+    kzg_proof*: KzgProof
+    column_index*: ColumnIndex
+    row_index*: RowIndex
+  
+  # Not in spec, defined in order to compute custody subnets
+  CscBits* = BitArray[DATA_COLUMN_SIDECAR_SUBNET_COUNT]
+
+  CscCount* = uint8
+
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.4/specs/_features/eip7594/p2p-interface.md#metadata
```
tersec: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/p2p-interface.md#metadata (unchanged contents)

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  BYTES_PER_CELL* = FIELD_ELEMENTS_PER_CELL * kzg_abi.BYTES_PER_FIELD_ELEMENT
+    # The number of bytes in a cell |
+  CELLS_PER_EXT_BLOB* = FIELD_ELEMENTS_PER_EXT_BLOB div FIELD_ELEMENTS_PER_CELL
+    # The number of cells in an extended blob |
+  KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH* = 4
+
+type
+  BLSFieldElement* = KzgBytes32
+  G2Point* = array[96, byte]
+  PolynomialCoeff* = List[BLSFieldElement, FIELD_ELEMENTS_PER_EXT_BLOB]
```
tersec, line 36: `PolynomialCoeff`, `Coset`, `CosetEvals`, `Cell`, and `CellIndex` are from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/polynomial-commitments-sampling.md#custom-types

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  Coset* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
+  CosetEvals* = array[FIELD_ELEMENTS_PER_CELL, BLSFieldElement]
+  Cell* = KzgCell
+  Cells* = KzgCells
+  CellsAndProofs* = KzgCellsAndKzgProofs
+  CellID* = uint64
```
tersec: What is a `CellID`? Looks like a remnant from before https://github.com/ethereum/consensus-specs/pull/3797

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  RowIndex* = uint64
+  ColumnIndex* = uint64
+  CellIndex* = uint64
+
+const
+  NUMBER_OF_COLUMNS* = 128
+  MAX_CELLS_IN_EXTENDED_MATRIX* = MAX_BLOBS_PER_BLOCK * NUMBER_OF_COLUMNS
```
tersec: https://github.com/ethereum/consensus-specs/pull/3887 seems to remove `MAX_CELLS_IN_EXTENDED_MATRIX` as a spec entity?

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  Cell* = KzgCell
+  Cells* = KzgCells
+  CellsAndProofs* = KzgCellsAndKzgProofs
+  CellID* = uint64
+  RowIndex* = uint64
+  ColumnIndex* = uint64
+  CellIndex* = uint64
+
+const
+  NUMBER_OF_COLUMNS* = 128
```
tersec, line 50: `NUMBER_OF_COLUMNS` is from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#data-size

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,101 @@
...
+  MAX_CELLS_IN_EXTENDED_MATRIX* = MAX_BLOBS_PER_BLOCK * NUMBER_OF_COLUMNS
+
+  DATA_COLUMN_SIDECAR_SUBNET_COUNT* = 128
+  SAMPLES_PER_SLOT* = 8
+  CUSTODY_REQUIREMENT* = 4
+  TARGET_NUMBER_OF_PEERS* = 70
```
tersec: `TARGET_NUMBER_OF_PEERS` has been removed by https://github.com/ethereum/consensus-specs/pull/3852 ?

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,116 @@
...
+
+  if not (custody_subnet_count <= DATA_COLUMN_SIDECAR_SUBNET_COUNT):
+    return err("Eip7594: Custody subnet count exceeds the DATA_COLUMN_SIDECAR_SUBNET_COUNT")
+
+  var 
+    subnet_ids: HashSet[uint64]
+    current_id = node_id
+
+  while subnet_ids.len < int(custody_subnet_count):
+
```
tersec: rm empty line after `while` loop condition

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,116 @@
...
+    let
+      hashed_current_id = eth2digest(current_id_bytes)
+      
+    hashed_bytes[0..7] = hashed_current_id.data.toOpenArray(0,7)
+    var subnet_id = bytes_to_uint64(hashed_bytes) mod 
+      DATA_COLUMN_SIDECAR_SUBNET_COUNT
+    
+    if subnet_id notin subnet_ids:
+        subnet_ids.incl(subnet_id)
```
tersec: This (and the body of `if current_id == UInt256.high.NodeId:` a couple lines later) should use two-space indentation

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+    let
+      meta = block:
+        var s = openFileStream(path/"meta.yaml")
+        defer: close(s)
+        var res: TestMetaYaml
+        yaml.load(s, res)
+        res
+      node_id = UInt256.fromDecimal(meta.node_id)
+      custody_subnet_count = toUInt64(meta.custody_subnet_count)
+      reslt = (meta.result.get).mapIt(uint64(it))
```
tersec: `meta.result.get` is already a `seq[uint64]`, so what's gained by doing the `uint64(it)` in the `mapIt`?

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+
+    if custody_subnet_count.isNone:
+      check meta.result.isNone
+    else:
+      let columns = get_custody_columns(node_id, custody_subnet_count.get)
+      if columns.isErr:
+        check meta.result.isNone
+      else:
+        for i in 0..<columns.get.lenu64:
+          check columns.get[i] == uint64(reslt[i])
```
tersec: Another conversion to `uint64` where as far as I can tell, it's being converted already from a `uint64` (i.e. it's an element of the `seq` returned by `(meta.result.get).mapIt(uint64(it))`, which will have the type `seq[uint64]`)

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+        defer: close(s)
+        var res: TestMetaYaml
+        yaml.load(s, res)
+        res
+      node_id = UInt256.fromDecimal(meta.node_id)
+      custody_subnet_count = toUInt64(meta.custody_subnet_count)
+      reslt = (meta.result.get).mapIt(uint64(it))
+
+    if custody_subnet_count.isNone:
+      check meta.result.isNone
```
tersec: can this happen? I don't see how when passed a `uint64`, toUInt64 can ever return an `Opt.none`

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+      result: Option[seq[uint64]]
+    let
+      meta = block:
+        var s = openFileStream(path/"meta.yaml")
+        defer: close(s)
+        var res: TestMetaYaml
+        yaml.load(s, res)
+        res
+      node_id = UInt256.fromDecimal(meta.node_id)
+      custody_subnet_count = toUInt64(meta.custody_subnet_count)
```
tersec: `meta.custody_subnet_count` is declared as a `uint64`, so what's the `toUInt64` call doing here? The whole thing looks like a roundabout no-op designed for more general situations. I'm not really even sure what the point of the type games is with the `custody_subnet_count` handling -- it:
- starts as a `uint64`
- gets wrapped in `Opt` as part of the `toUInt64`
- stored as a local variable, saving 5 characters
- later (I think?) more or less unambiguously creates an `Opt.some` rather than `Opt.none`, because `toUInt64` should always return `Opt.some x` for a `uint64` `x`
- which gets unwrapped with `custody_subnet_count.get`

it's defined as a `uint64` (not an `Opt[uint64]` in `TestMetaYaml`, so it seems simpler to drop the `toUInt64`/call to `toUInt64`, then drop the impossible `if custody_subnet_count.isNone:` branch and refer to `let columns = get_custody_columns(node_id, meta.custody_subnet_count)` directly

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+  results,
+  stew/[byteutils],
+  ../../beacon_chain/spec/eip7594_helpers,
+  ../testutil,
+  ./fixtures_utils, ./os_ops
+
+from std/sequtils import anyIt, mapIt, toSeq
+from std/strutils import rsplit
+
+func toUInt64(s: SomeInteger): Opt[uint64] =
```
tersec: See elsewhere, pretty sure this function isn't doing anything useful here

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+  stint,
+  chronicles,
+  eth/p2p/discoveryv5/[node],
+  results,
+  stew/[byteutils],
+  ../../beacon_chain/spec/eip7594_helpers,
+  ../testutil,
+  ./fixtures_utils, ./os_ops
+
+from std/sequtils import anyIt, mapIt, toSeq
```
tersec: Are `anyIt` or `toSeq` used here?

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,78 @@
...
+from std/strutils import rsplit
```
tersec: Doesn't seem like a used import?

beacon_chain/spec/eip7594_helpers.nim
agnxsh: It can fail if the user tries to boot up (tho an experimental scenario totally) nimbus-eth2, using custom subnet count > DATA COLUMN SIDECAR SUBNET COUNT, from the CLI flags.

agnxsh: ok no that call happens in the subnet function, yes no need to ok() then

agnxsh: same for custody column list

tests/consensus_spec/test_fixture_networking.nim
agnxsh: in the test format all the fields are `int`s

agnxsh: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/tests/formats/networking/get_custody_columns.md#test-case-format

agnxsh: i thought the parser tries to resolve the test format mentioned in the spec, removed all the unwanted conversions/Opt

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,64 @@
...
+  eth/p2p/discoveryv5/[node],
+  results,
+  stew/[byteutils],
+  ../../beacon_chain/spec/eip7594_helpers,
+  ../testutil,
+  ./fixtures_utils, ./os_ops
+
+from std/sequtils import mapIt
+
+func fromHex[N: static int](s: string): Opt[array[N, byte]] =
```
tersec: `fromHex` doesn't seem to be used?

Maybe it will be at some point, but it can be introduced then

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -0,0 +1,64 @@
...
+{.push raises: [].}
+{.used.}
+
+import
+  std/[json, streams],
+  yaml,
+  kzg4844/[kzg, kzg_abi],
+  stint,
+  chronicles,
+  eth/p2p/discoveryv5/[node],
```
tersec, line 16: Surprising import?

beacon_chain/spec/datatypes/eip7594.nim
```diff
@@ -0,0 +1,109 @@
...
+  std/[sequtils],
+  "."/[altair, base, deneb], 
+  kzg4844/[kzg, kzg_abi]
+
+from std/strutils import join
+
+export base
+
+const
+  FIELD_ELEMENTS_PER_EXT_BLOB* = 2 * kzg_abi.FIELD_ELEMENTS_PER_BLOB
```
tersec, line 21: These appear to be from https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/polynomial-commitments-sampling.md#cells

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+      hashed_bytes: array[8, byte]
+      
+    current_id_bytes = current_id.toBytesBE()
+    current_id_bytes.reverse()
+
+    let
+      hashed_current_id = eth2digest(current_id_bytes)
+      
+    hashed_bytes[0..7] = hashed_current_id.data.toOpenArray(0,7)
+    var subnet_id = bytes_to_uint64(hashed_bytes) mod
```
tersec: Only used immutably, can be a `let`

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+  # in order to maintain subscription to specific column subnets.
+
+  if not (custody_subnet_count <= DATA_COLUMN_SIDECAR_SUBNET_COUNT):
+    return err("Eip7594: Custody subnet count exceeds the DATA_COLUMN_SIDECAR_SUBNET_COUNT")
+
+  var 
+    subnet_ids: HashSet[uint64]
+    current_id = node_id
+
+  while subnet_ids.len < int(custody_subnet_count):
```
tersec: More typesafe to compare `subnet_ids.lenu64` with `custody_subnet_count`, which can't in theory truncate

Not a major issue, practically, but easier to remove the dubious-looking `int(foo)` conversions, since `int(foo)` can in theory `Defect` with certain input at runtime

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+    current_id_bytes.reverse()
+
+    let
+      hashed_current_id = eth2digest(current_id_bytes)
+      
+    hashed_bytes[0..7] = hashed_current_id.data.toOpenArray(0,7)
+    var subnet_id = bytes_to_uint64(hashed_bytes) mod 
+      DATA_COLUMN_SIDECAR_SUBNET_COUNT
+    
+    if subnet_id notin subnet_ids:
```
tersec: It has to probe the hash table twice this way if `notin`. Can just do the `incl` and do one hash table probe.

https://nim-lang.org/docs/sets.html#incl%2CHashSet%5BA%5D%2CA already states that
> This doesn't do anything if `key` is already in `s`.

And implementation-wise, https://github.com/nim-lang/Nim/blob/4bf323d6c44b329fe1955f210580a73952c081d0/lib/pure/collections/setimpl.nim#L45-L55 already basically does this:
```nim
template inclImpl() {.dirty.} =
  if s.data.len == 0:
    initImpl(s, defaultInitialSize)
  var hc: Hash
  var index = rawGet(s, key, hc)
  if index < 0:
    if mustRehash(s):
      enlarge(s)
      index = rawGetKnownHC(s, key, hc)
    rawInsert(s, s.data, key, hc, -1 - index)
    inc(s.counter)
```

i.e. that `if index < 0: ...`

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+# Uncategorized helper functions from the spec
+import
+  std/[algorithm, hashes],
+  results,
+  kzg4844/[kzg, kzg_abi],
+  eth/p2p/discoveryv5/[node],
```
tersec, line 14: Is this import used here?

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+    subnet_ids: HashSet[uint64]
+    current_id = node_id
+
+  while subnet_ids.len < int(custody_subnet_count):
+    var
+      current_id_bytes: array[32, byte]
+      hashed_bytes: array[8, byte]
+      
+    current_id_bytes = current_id.toBytesBE()
+    current_id_bytes.reverse()
```
tersec: `toBytesBE`/`reverse` should also be spelled as `toBytesLE`

tersec: (at which point it can be done immutably, so `let`)

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+  ok(subnet_ids)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#get_custody_columns
+proc get_custody_columns*(node_id: NodeId, 
+                          custody_subnet_count: uint64): 
+                          seq[ColumnIndex] =
+  let
+    subnet_ids = 
+      get_custody_column_subnet(node_id, custody_subnet_count).get
+    columns_per_subnet =
```
tersec, line 86: Looks `const`

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+proc get_custody_column_list*(node_id: NodeId, 
+                          custody_subnet_count: uint64): 
+                          List[ColumnIndex, NUMBER_OF_COLUMNS] =
+
+  # Not in spec in the exact format, but it is useful in sorting custody columns 
+  # before sending, data_column_sidecars_by_range requests
+  let
+    subnet_ids = 
+      get_custody_column_subnet(node_id, custody_subnet_count).get
+    columns_per_subnet =
```
tersec, line 102: `const` should work here

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+  var res: seq[ColumnIndex] = @[]
+  for i in 0'u64 ..< columnsPerSubnet:
+    for subnetId in subnetIds:
+      let index = DATA_COLUMN_SIDECAR_SUBNET_COUNT * i + subnetId
+      res.add(ColumnIndex(index))
+  res.sort
+  res
+
+proc sortedColumnIndexList*(columnsPerSubnet: ColumnIndex, 
+                          subnetIds: HashSet[uint64]):
```
tersec: Odd indentation (compare with `sortedColumnIndices` just before it)

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+  for i in 0'u64 ..< columnsPerSubnet:
+    for subnetId in subnetIds:
+      let index = DATA_COLUMN_SIDECAR_SUBNET_COUNT * i + subnetId
+      res.add(ColumnIndex(index))
+  res.sort()
+  for elem in res:
+    discard list.add(ColumnIndex(elem))
+  list
+
+proc get_custody_column_subnet*(node_id: NodeId,
```
tersec: Name suggests it returns one subnet (singular) but it in fact returns an entire collection of subnets, plural

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+  std/[algorithm, hashes],
+  results,
+  kzg4844/[kzg, kzg_abi],
+  eth/p2p/discoveryv5/[node],
+  ./[helpers, digest],
+  ./datatypes/[eip7594, deneb],
+  ssz_serialization/[
+    proofs,
+    types],
+  ./[beacon_time, crypto]
```
tersec: It doesn't seem to use anything from `beacon_time`?

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -0,0 +1,114 @@
...
+  var
+    res: seq[ColumnIndex]
+    list: List[ColumnIndex, NUMBER_OF_COLUMNS]
+  for i in 0'u64 ..< columnsPerSubnet:
+    for subnetId in subnetIds:
+      let index = DATA_COLUMN_SIDECAR_SUBNET_COUNT * i + subnetId
+      res.add(ColumnIndex(index))
+  res.sort()
+  for elem in res:
+    discard list.add(ColumnIndex(elem))
```
tersec: `res` is already a `seq[ColumnIndex]`, so `elem` is already a `ColumnIndex`, so this `ColumnIndex(elem)` conversion is a no-op.

One can more directly construct `list` with `init` from https://github.com/status-im/nim-ssz-serialization/blob/b5aa78cae197cc97277be050ba69ef9c5d33aad1/ssz_serialization/types.nim#L167-L168:
```nim
template init*[T, N](L: type List[T, N], x: seq[T]): auto =
  List[T, N](x)
```
which takes advantage of that:
```nim
type
  List*[T; maxLen: static Limit] = distinct seq[T]
```
i.e. it doesn't need a copy, `List` literally is a distinct `seq`.

for example,
`List[ColumnIndex, NUMBER_OF_COLUMNS].init(res)` instead of this whole `for elem in res:` loop

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -46,7 +41,7 @@ proc sortedColumnIndexList*(columnsPerSubnet: ColumnIndex,
     discard list.add(ColumnIndex(elem))
   list
 
-proc get_custody_column_subnet*(node_id: NodeId, 
+proc get_custody_column_subnets*(node_id: NodeId, 
                                 custody_subnet_count: uint64):
```
tersec: indentations here need adjustment after rename

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -73,11 +68,10 @@ proc get_custody_column_subnet*(node_id: NodeId,
       hashed_current_id = eth2digest(current_id_bytes)
       
     hashed_bytes[0..7] = hashed_current_id.data.toOpenArray(0,7)
-    var subnet_id = bytes_to_uint64(hashed_bytes) mod 
+    let subnet_id = bytes_to_uint64(hashed_bytes) mod 
       DATA_COLUMN_SIDECAR_SUBNET_COUNT
     
-    if subnet_id notin subnet_ids:
-      subnet_ids.incl(subnet_id)
+    discard subnet_ids.containsOrIncl(subnet_id)
```
tersec: just `subnet_ids.incl(subnet_id)` without the `notin` check works, is simpler, and doesn't require this `discard` which is a bit of a suboptimal thing to have because it indicates ignoring return values. The point is kind of that adding an element to a set which is already present in that set is not an error, it's a safe, well-defined idempotent operation, which doesn't need to be guarded

PR #6537, opened 2024-09-04
bump nim-eth to `f2568a64c0933c3a0bf959568df400cb1d16c2bf`

etan-status:
- Differentiate log for invalid packet seqnr without reorder buffer and not
- Using unsigned types for message type and requst IDs
- Reassign logging weights in RLPx handler
- Fix potential uTP clock drift overflow


etan-status:
image

this one is a bit sketchy, but should only affect logging, if at all

PR #6538, opened 2024-09-10
bump nim-eth and nim-nat-traversal

tersec:
From `nim-eth`:
- https://github.com/status-im/nim-eth/pull/728

From `nim-nat-traversal`:
- https://github.com/status-im/nim-nat-traversal/pull/29

Avoids a potential crash while setting up UPnP.


etan-status:
Thanks. Good catch, and sustainable fix, that we did not handle unexpected output codes gracefully.
Have smoke-tested locally but IGD worked fine so didn't hit it.

PR #6447, opened 2024-07-24
hypergeometric distribution CDF for PeerDAS

tersec:
For [get_extended_sample_count](https://github.com/ethereum/consensus-specs/blob/bb8f3caafc92590cdcf2d14974adb602db9b5ca3/specs/_features/eip7594/das-core.md#get_extended_sample_count)

The reference implementation intrinsically relies on Python's arbitrary-sized integers. This implementation computes the same function with machine-sized data types and in O(1) memory, without bringing in any new dependencies Nim or C.

PR #6544, opened 2024-09-11
add: `get_extended_sample_count` with test

agnxsh:
Adds https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/peer-sampling.md#get_extended_sample_count with test to facilitate Lossy Sampling

Prequel: https://github.com/status-im/nimbus-eth2/pull/6447

### DIFF
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -111,7 +111,7 @@ func get_extended_sample_count*(samples_per_slot: int,
   const columnsCount = NUMBER_OF_COLUMNS.int
 
   # If 50% of the columns are missing, we are able to reconstruct the data
-  # If 50% + 1 columns are missing, we are NO MORE able to reconstruct the data
+  # If 50% + 1 columns are missing, we are become unable to reconstruct the data
   let worstCaseConditionCount = (columnsCount div 2) + 1
 
   # Compute the false positive threshold
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -100,12 +100,12 @@ func get_custody_column_list*(node_id: NodeId,
   sortedColumnIndexList(ColumnIndex(columns_per_subnet), subnet_ids)
 
 # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/peer-sampling.md#get_extended_sample_count
-proc get_extended_sample_count*(samples_per_slot: int,
+func get_extended_sample_count*(samples_per_slot: int,
                                 allowed_failures: int):
                                 int =
-  # `get_extended_sample_count` computes the number of samples we
-  # should query from peers, given the SAMPLES_PER_SLOT and 
-  # the number of allowed failures
+  ## `get_extended_sample_count` computes the number of samples we
+  ## should query from peers, given the SAMPLES_PER_SLOT and 
+  ## the number of allowed failures
 
   # Retrieving the column count
   const columnsCount = NUMBER_OF_COLUMNS.int
@@ -122,9 +122,10 @@ proc get_extended_sample_count*(samples_per_slot: int,
 
   # Finally, compute the extended sample count
   for i in samples_per_slot .. columnsCount + 1:
-    if hypergeom_cdf(allowed_failures,
-                     columnsCount, 
-                     worstCaseConditionCount, i) <= falsePositiveThreshold:
+    if hypergeom_cdf(
+        allowed_failures,
+        columnsCount, 
+        worstCaseConditionCount, i) <= falsePositiveThreshold:
       sampleCount = i
       break
     sampleCount = i
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -126,4 +126,4 @@ proc get_extended_sample_count*(samples_per_slot: int,
       break
     sampleCount = i
   
-  return sampleCount
+  sampleCount
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -108,7 +108,7 @@ func get_extended_sample_count*(samples_per_slot: int,
   ## the number of allowed failures
 
   # Retrieving the column count
-  const columnsCount = NUMBER_OF_COLUMNS.int
+  const columnsCount = NUMBER_OF_COLUMNS
 
   # If 50% of the columns are missing, we are able to reconstruct the data
   # If 50% + 1 columns are missing, we cannot reconstruct the data
@@ -119,11 +119,11 @@ func get_extended_sample_count*(samples_per_slot: int,
     hypergeom_cdf(0, columnsCount, worstCaseConditionCount, samples_per_slot)
 
   # Finally, compute the extended sample count
-  for i in samples_per_slot .. columnsCount + 1:
+  for i in samples_per_slot .. columnsCount:
     if hypergeom_cdf(
         allowed_failures,
         columnsCount, 
         worstCaseConditionCount, i) <= falsePositiveThreshold:
       return i
 
-  return -999
+  columnsCount
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -108,20 +108,23 @@ proc get_extended_sample_count*(samples_per_slot: int,
   # the number of allowed failures
 
   # Retrieving the column count
-  let columnsCount = NUMBER_OF_COLUMNS.int
+  const columnsCount = NUMBER_OF_COLUMNS.int
 
   # If 50% of the columns are missing, we are able to reconstruct the data
   # If 50% + 1 columns are missing, we are NO MORE able to reconstruct the data
   let worstCaseConditionCount = (columnsCount div 2) + 1
 
   # Compute the false positive threshold
-  let falsePositiveThreshold = hypergeom_cdf(0, columnsCount, worstCaseConditionCount, samples_per_slot)
+  let falsePositiveThreshold = 
+    hypergeom_cdf(0, columnsCount, worstCaseConditionCount, samples_per_slot)
 
   var sampleCount: int
 
   # Finally, compute the extended sample count
   for i in samples_per_slot .. columnsCount + 1:
-    if hypergeom_cdf(allowed_failures, columnsCount, worstCaseConditionCount, i) <= falsePositiveThreshold:
+    if hypergeom_cdf(allowed_failures,
+                     columnsCount, 
+                     worstCaseConditionCount, i) <= falsePositiveThreshold:
       sampleCount = i
       break
     sampleCount = i
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -107,23 +107,20 @@ func get_extended_sample_count*(samples_per_slot: int,
   ## should query from peers, given the SAMPLES_PER_SLOT and 
   ## the number of allowed failures
 
-  # Retrieving the column count
-  const columnsCount = NUMBER_OF_COLUMNS
-
   # If 50% of the columns are missing, we are able to reconstruct the data
   # If 50% + 1 columns are missing, we cannot reconstruct the data
-  const worstCaseConditionCount = (columnsCount div 2) + 1
+  const worstCaseConditionCount = (NUMBER_OF_COLUMNS div 2) + 1
 
   # Compute the false positive threshold
   let falsePositiveThreshold = 
-    hypergeom_cdf(0, columnsCount, worstCaseConditionCount, samples_per_slot)
+    hypergeom_cdf(0, NUMBER_OF_COLUMNS, worstCaseConditionCount, samples_per_slot)
 
   # Finally, compute the extended sample count
-  for i in samples_per_slot .. columnsCount:
+  for i in samples_per_slot .. NUMBER_OF_COLUMNS:
     if hypergeom_cdf(
         allowed_failures,
-        columnsCount, 
+        NUMBER_OF_COLUMNS, 
         worstCaseConditionCount, i) <= falsePositiveThreshold:
       return i
 
-  columnsCount
+  NUMBER_OF_COLUMNS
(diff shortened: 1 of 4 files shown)


tests/test_eip7594_helpers.nim
```diff
@@ -0,0 +1,45 @@
...
+        (15, 55),
+        (16, 57),
+        (17, 59),
+        (18, 61),
+        (19, 63),
+        (20, 65)
+      ]
+
+      for (allowed_failures, extendedSampleCount) in tests:
+        check: get_extended_sample_count(samplesPerSlot, allowed_failures) == extendedSampleCount
```
tersec: 98 character line

agnxsh: checking

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,32 @@ func get_custody_column_list*(node_id: NodeId,
...
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/peer-sampling.md#get_extended_sample_count
+proc get_extended_sample_count*(samples_per_slot: int,
+                                allowed_failures: int):
+                                int =
+  # `get_extended_sample_count` computes the number of samples we
+  # should query from peers, given the SAMPLES_PER_SLOT and 
+  # the number of allowed failures
+
+  # Retrieving the column count
+  let columnsCount = NUMBER_OF_COLUMNS.int
```
tersec: `const`?

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,32 @@ func get_custody_column_list*(node_id: NodeId,
...
+
+  # If 50% of the columns are missing, we are able to reconstruct the data
+  # If 50% + 1 columns are missing, we are NO MORE able to reconstruct the data
+  let worstCaseConditionCount = (columnsCount div 2) + 1
+
+  # Compute the false positive threshold
+  let falsePositiveThreshold = hypergeom_cdf(0, columnsCount, worstCaseConditionCount, samples_per_slot)
```
tersec: long line

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,32 @@ func get_custody_column_list*(node_id: NodeId,
...
+
+  var sampleCount: int
+
+  # Finally, compute the extended sample count
+  for i in samples_per_slot .. columnsCount + 1:
+    if hypergeom_cdf(allowed_failures, columnsCount, worstCaseConditionCount, i) <= falsePositiveThreshold:
```
tersec: long line

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,35 @@ func get_custody_column_list*(node_id: NodeId,
       NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
 
   sortedColumnIndexList(ColumnIndex(columns_per_subnet), subnet_ids)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/peer-sampling.md#get_extended_sample_count
+proc get_extended_sample_count*(samples_per_slot: int,
```
tersec: `func`

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,35 @@ func get_custody_column_list*(node_id: NodeId,
...
+                                allowed_failures: int):
+                                int =
+  # `get_extended_sample_count` computes the number of samples we
```
tersec: These kind of function-purpose-documentation headers can/should use `##` rather than `#`: https://nim-lang.org/docs/manual.html#lexical-analysis-comments
> Documentation comments are comments that start with two `##`. Documentation comments are tokens; they are only allowed at certain places in the input file as they belong to the syntax tree.

They'll get picked up by API/documentation generation tools, et cetera

tersec: This does not mean all comments should use `##` -- normal comments still use `#`

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,35 @@ func get_custody_column_list*(node_id: NodeId,
...
+  # should query from peers, given the SAMPLES_PER_SLOT and 
+  # the number of allowed failures
+
+  # Retrieving the column count
+  const columnsCount = NUMBER_OF_COLUMNS.int
+
+  # If 50% of the columns are missing, we are able to reconstruct the data
+  # If 50% + 1 columns are missing, we are NO MORE able to reconstruct the data
```
tersec:
```suggestion
  # If 50% + 1 columns are missing, we become unable to reconstruct the data
```

tests/test_eip7594_helpers.nim
```diff
@@ -0,0 +1,46 @@
+# beacon_chain
```
tersec, line 1: Is this hooked up anywhere into `all_tests` or similar? Otherwise it won't run

agnxsh: oops

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,36 @@ func get_custody_column_list*(node_id: NodeId,
...
+
+  var sampleCount: int
+
+  # Finally, compute the extended sample count
+  for i in samples_per_slot .. columnsCount + 1:
+    if hypergeom_cdf(
+        allowed_failures,
+        columnsCount, 
+        worstCaseConditionCount, i) <= falsePositiveThreshold:
+      sampleCount = i
```
tersec: Maybe cleaner to early-return here rather than set this `sampleCount` variable?

Yeah, some people really view structured programming as one-exit-per-function but Nimbus emphatically does not adopt that style

agnxsh: okay

tests/test_eip7594_helpers.nim
agnxsh: running make test once to update all test.md

beacon_chain/spec/eip7594_helpers.nim
tersec: That would mean having some placeholder return though, if the loop falls through -- or can one (see comments above) 100% guarantee that's not the case?

Keep in mind `hypergeom_cdf` is approximate to some extent, so even if platonically ideally mathematically this held, floating point and numerical analysis issues could still strike

tests/all_tests.nim
```diff
@@ -27,6 +27,7 @@ import # Unit test
   ./test_discovery,
   ./test_engine_api_conversions,
   ./test_engine_authentication,
+  ./test_eip7594_helpers,
```
tersec, line 30: Ok, great, now need to regenerate the test summary results by running `make test` locally

tests/test_eip7594_helpers
tersec, line 1: Shouldn't be included in the PR

tests/all_tests.nim
agnxsh: yesyes

tests/test_eip7594_helpers
agnxsh: what shouldn't be included?

tersec: it's a binary artifact, not a source file

tersec: i.e. the `.nim` file `tests/test_eip7594_helpers.nim` is obviously fine/intended. this one (`tests/test_eip7594_helpers`) is not.

agnxsh: ok ok yes

agnxsh: my bad, done

tests/all_tests.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6544/commits/6df42c894d548c64c746bda0c7c5b5736a887071

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,36 @@ func get_custody_column_list*(node_id: NodeId,
...
+                                int =
+  ## `get_extended_sample_count` computes the number of samples we
+  ## should query from peers, given the SAMPLES_PER_SLOT and 
+  ## the number of allowed failures
+
+  # Retrieving the column count
+  const columnsCount = NUMBER_OF_COLUMNS.int
+
+  # If 50% of the columns are missing, we are able to reconstruct the data
+  # If 50% + 1 columns are missing, we are become unable to reconstruct the data
```
tersec:
```
# If 50% + 1 columns are missing, we become unable to reconstruct the data
```
or
```
# If 50% + 1 columns are missing, we cannot reconstruct the data
```

depends on whether one wants to emphasize change-in-condition or condition

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,36 @@ func get_custody_column_list*(node_id: NodeId,
...
+  let worstCaseConditionCount = (columnsCount div 2) + 1
```
tersec: this can be `const` too

tests/test_eip7594_helpers
tersec: https://github.com/status-im/nimbus-eth2/pull/6544/commits/e1567c2928f12fcf12387d5e72335db1ef03db1f

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,36 @@ func get_custody_column_list*(node_id: NodeId,
...
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/peer-sampling.md#get_extended_sample_count
+func get_extended_sample_count*(samples_per_slot: int,
+                                allowed_failures: int):
+                                int =
+  ## `get_extended_sample_count` computes the number of samples we
+  ## should query from peers, given the SAMPLES_PER_SLOT and 
+  ## the number of allowed failures
+
+  # Retrieving the column count
+  const columnsCount = NUMBER_OF_COLUMNS.int
```
tersec: I'm not sure what integer data type this is going to be called with (`uint64`), but whatever it is, it's probably better to start by adjusting
https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/spec/helpers.nim#L612-L630
from using `int` (which was kind of a placeholder) to whatever that type is, in `ln_binomial` and `hypergeom_cdf`, and be consistent about its usage here and use the same times throughout without conveting. Those two functions should work (and the tests should be able to run easily) with any reasonable integer type.

beacon_chain/spec/eip7594_helpers.nim
agnxsh: why don't we just wrap it with results?

agnxsh: we early return with ok, or we return err saying there's some issue with the input, because `samples_per_slot` can never be 0.

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,32 @@ func get_custody_column_list*(node_id: NodeId,
...
+
+  # Finally, compute the extended sample count
+  for i in samples_per_slot .. columnsCount + 1:
+    if hypergeom_cdf(
+        allowed_failures,
+        columnsCount, 
+        worstCaseConditionCount, i) <= falsePositiveThreshold:
+      return i
+
+  return -999
```
tersec:
```python
    for sample_count in range(SAMPLES_PER_SLOT, NUMBER_OF_COLUMNS + 1):
        if hypergeom_cdf(allowed_failures, NUMBER_OF_COLUMNS,
                         worst_case_missing, sample_count) <= false_positive_threshold:
            break
    return sample_count
```

This is the algorithm from the spec.

While in general, mirroring the Python code is nice, the issue here is that the variable scoping of the loop variable is different enough that it's clumsy to do so in Nim.

But one can extract the different outcomes in each case:
- if the `if hypergeom(...) <= falsePositiveThreashold` condition holds for some `i`, return that `i`.
- if the `if hypergeom(...) <= falsePositiveThreashold` condition never holds, then in the Python version of `for` loops what happen is that the loop variable will take the last value from the loop:
```
>>> for i in [1, 2, 3, 4, 5]: pass
... 
>>> print(i)
5
```
i.e. here, `NUMBER_OF_COLUMNS`, because
```
>>> print(list(range(5, 10)))
[5, 6, 7, 8, 9]
```

So to mirror the Python algorithm, fallthrough case should return `NUMBER_OF_COLUMNS` in Python spec terms and here that or `columnsCount` depending on type desired.

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -101,3 +101,32 @@ func get_custody_column_list*(node_id: NodeId,
...
+  # If 50% of the columns are missing, we are able to reconstruct the data
+  # If 50% + 1 columns are missing, we cannot reconstruct the data
+  const worstCaseConditionCount = (columnsCount div 2) + 1
+
+  # Compute the false positive threshold
+  let falsePositiveThreshold = 
+    hypergeom_cdf(0, columnsCount, worstCaseConditionCount, samples_per_slot)
+
+  # Finally, compute the extended sample count
+  for i in samples_per_slot .. columnsCount + 1:
```
tersec:
```python
    for sample_count in range(SAMPLES_PER_SLOT, NUMBER_OF_COLUMNS + 1):
```
does not iterate all the way to `NUMBER_OF_COLUMNS + 1` but only to `NUMBER_OF_COLUMNS`.

To emulate Python's `range` bounds, can use `..<`:
```Nim
for i in samples_per_slot ..< columnsCount + 1:
```
or can use `..` and adjust the upper bound:
```Nim
for i in samples_per_slot .. columnsCount:
```

In pure clarity terms the latter is probably better, but the former has the virtue of mapping more precisely to the Python spec.

beacon_chain/spec/eip7594_helpers.nim
tersec: This was determined to be fine

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -111,7 +111,7 @@ func get_extended_sample_count*(samples_per_slot: int,
   ## the number of allowed failures
 
   # Retrieving the column count
-  const columnsCount = NUMBER_OF_COLUMNS.int
+  const columnsCount = NUMBER_OF_COLUMNS
```
tersec: at this point why even have a separate `columsCount` symbol? Not sure what it's providing on top of `NUMBER_OF_COLUMNS`.

It's harmless but also unnecessary

agnxsh: true

agnxsh: done

PR #6546, opened 2024-09-12
Add publishAggregateAndProofsV2

pedromiguelmiranda:
Added new endpoint version(v2) for validator aggregate and proof.
spec: [ethereum/beacon-APIs/pull/456](https://github.com/ethereum/beacon-APIs/pull/456)


beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -890,6 +890,66 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+    let
+      consensusVersion = request.headers.getString("Eth-Consensus-Version")
+      isElectra = consensusVersion == "electra"
+
+    if contentBody.isNone():
+      return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
+
+    let proofs =
+      block:
+        if isElectra:
```
tersec: If this section did one additional step, and called `node.router.routeSignedAggregateAndProof(...)` (which returns a `Future` which does not yet run) here, it would all return `SendFuture` and one could avoid the whole `MultiForkProof` aspect. So instead of constructing
```nim
  MultiForkProof(isElectra: true, electraProofs: dres.get())
```
one can construct
```nim
  node.router.routeSignedAggregateAndProof(dres.get())
```
and skip most of the type-management here by unifying the types a section earlier.

pedromiguelmiranda: Perfect. Thank you. More elegant your suggestion.
Was a little bit skeptical on using a _variant_ to solve the static check.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -890,6 +890,49 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
                "Unexpected server failure, while sending aggregate and proof")
     RestApiResponse.jsonMsgResponse(AggregateAndProofValidationSuccess)
 
+  # https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/publishAggregateAndProofsV2
+  router.api2(MethodPost, "/eth/v2/validator/aggregate_and_proofs") do (
+    contentBody: Option[ContentBody]) -> RestApiResponse:
+
+    let
+      consensusVersion = request.headers.getString("Eth-Consensus-Version")
+      isElectra = consensusVersion == "electra"
```
tersec: Well, the next forks, starting with f, g, h, etc, will also not use phase0 attestations at least but, as far as is currently known, Electra attestations, unless it changes again. But this check finds only Electra specifically.

One approach is to use
https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/spec/forks.nim#L797-L816
i.e. use `ConsensusFork.init(consensusVersion)` to detect if it's a valid fork string and if so, which. Then one can compare, e.g.
```nim
  let isElectra =
    case foo
    of ConsensusFork.Phase0 .. ConsensusFork.Deneb: false
    of ConsensusFork.Electra: true
```
which is a bit more verbose than just `>`/`<`/`>=`/`<=` but makes sure that when new forks are added, Nim's enum case exhaustiveness checks will notice and not automatically assume how that new fork maps to a potential new or existing attestation type.

`ConsensusFork.init("some string")` returns an `Opt[ConsensusFork]` so the case of someone sending gibberish which isn't a fork would be detected as, of the three errors allowed (200, 400, and 500), most appropriately probably HTTP error 400.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -890,6 +890,49 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+    if contentBody.isNone():
+      return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
+
+    var proofs: seq[Future[SendResult]]
+    if isElectra:
+      let dres = decodeBody(seq[electra.SignedAggregateAndProof], contentBody.get())
```
tersec: This is an improvement, but still a bit redundant. The two branches of the `if isElectra` are identical except for the first argument to `decodeBody`.

This can be abstracted to an internal template declared within the handler for example (or close to this; this exact code untested):
```nim
    template addDecodedProofs(ProofType: untyped) =
      let dres = decodeBody(seq[ProofType], contentBody.get())
      if dres.isErr():
        return RestApiResponse.jsonError(Http400,
                                         InvalidAggregateAndProofObjectError,
                                         $dres.error())
      for proof in dres.get():
        proofs.add(node.router.routeSignedAggregateAndProof(proof))
```
which can be called as either `addDecodedProofs(phase0.SignedAggregateAndProof)` or `addDecodedProofs(electra.SignedAggregateAndProof)`

tersec: Because of how `template`s work, that `return` actually `return`s out of the handler as a whole, not only the `template`.

beacon_chain/rpc/rest_validator_api.nim
pedromiguelmiranda: Perfect. It's more verbose but a whole lote correct in all aspects.

beacon_chain/rpc/rest_validator_api.nim
pedromiguelmiranda: Perfect.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -890,6 +890,48 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+    let
+      headerVersion = request.headers.getString("Eth-Consensus-Version")
+      consensusVersion = ConsensusFork.init(headerVersion)
+    if consensusVersion.isNone():
+      return RestApiResponse.jsonError(Http400, FailedToObtainConsensusForkError)
+
+    var proofs: seq[Future[SendResult]]
+    template addDecodedProofs(ProofType: untyped) =
+        let dres = decodeBody(seq[ProofType], contentBody.get())
```
tersec, line 908: Should use 2 space indentation instead of 4 for `template` body

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -890,6 +890,48 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        if dres.isErr():
+          return RestApiResponse.jsonError(Http400,
+                                          InvalidAggregateAndProofObjectError,
```
tersec, line 911: Probably more consistent to indent `InvalidAggregateAndProofObjectError` and `$dres.error()` with `Http400,` rather than `(Http400,`, i.e. one space more. it's what you do later in the same function (`AggregateAndProofValidationError`) at least, and what most of this module uses.

tersec:
https://github.com/status-im/nimbus-eth2/actions/runs/10853866443/job/30123076903?pr=6546 from `CI / linux-amd64 (Nim version-2-0) (pull_request)` is failing on:
`2024-09-13T18:05:58.0116534Z /github-runner/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/libnimbus_lc/libnimbus_lc.nim(1465, 11) Error: type mismatch: got 'uint64' for '0'u' but expected 'Quantity = distinct uint64'`

probably unrelated to this PR

tersec:
https://github.com/status-im/nimbus-eth2/actions/runs/10853866443/job/30123076903?pr=6546 from `CI / linux-amd64 (Nim version-2-0) (pull_request)` is failing on:
```
2024-09-13T18:05:58.0116534Z /github-runner/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/libnimbus_lc/libnimbus_lc.nim(1465, 11) Error: type mismatch: got 'uint64' for '0'u' but expected 'Quantity = distinct uint64'
```
probably unrelated to this PR

tersec:
https://github.com/nim-lang/Nim/issues/24104

beacon_chain/rpc/rest_validator_api.nim
tersec: Addressing these in https://github.com/status-im/nimbus-eth2/pull/6548

PR #6549, opened 2024-09-15
add some electra support to REST JSON decoding; rm unused code

tersec:
Stage-setting for fixing forked block body reading for Electra.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -15,14 +15,10 @@ import results, stew/[assign2, base10, byteutils, endians2], presto/common,
        stint, chronicles
 import ".."/[eth2_ssz_serialization, forks, keystore],
        ".."/../consensus_object_pools/block_pools_types,
-       ".."/datatypes/[phase0, altair, bellatrix],
        ".."/mev/[bellatrix_mev, capella_mev],
        ".."/../validators/slashing_protection_common,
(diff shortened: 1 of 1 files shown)


tersec:
CI / linux-amd64 (Nim version-2-0) failure is unrelated: https://github.com/nim-lang/Nim/issues/24104

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -4336,14 +4295,16 @@ proc decodeString*(t: typedesc[ValidatorFilter],
   else:
     err("Incorrect validator state identifier value")
 
-proc decodeString*(t: typedesc[ConsensusFork],
+func decodeString*(t: typedesc[ConsensusFork],
                    value: string): Result[ConsensusFork, cstring] =
+  static: doAssert ConsensusFork.high == ConsensusFork.Electra
```
etan-status, line 4300: There's `ConsensusFork.init` in `forks.nim` which happens to do the same

tersec: Yes, need to verify differences in case handling -- these lowercase first and that still needs to be done separately

etan-status: ConsensusFork.init(value.toLower)?

tersec: https://github.com/status-im/nimbus-eth2/pull/6587

PR #6550, opened 2024-09-15
add: compute matrix

agnxsh:
Ref: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#compute_matrix

This PR adds the `compute_matrix` function for peerDAS, it enables the BN to take a decision on when can rest of the data columns be reconstructed by obtaining a `MatrixEntry` like view.

### DIFF
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -119,7 +119,6 @@ proc compute_matrix* (blobs: seq[KzgBlob]): Result[seq[MatrixEntry], cstring] =
 
   ok(extended_matrix)
 
-
 # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/peer-sampling.md#get_extended_sample_count
 func get_extended_sample_count*(samples_per_slot: int,
                                 allowed_failures: int):
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -99,6 +99,27 @@ func get_custody_column_list*(node_id: NodeId,
 
   sortedColumnIndexList(ColumnIndex(columns_per_subnet), subnet_ids)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#compute_matrix
+proc compute_matrix* (blobs: seq[KzgBlob]): Result[seq[MatrixEntry], cstring] =
+  # This helper demonstrates the relationship between blobs and the `MatrixEntries`
+  var extended_matrix: seq[MatrixEntry]
+
+  for blbIdx, blob in blobs.pairs:
+    let cellsAndProofs = computeCellsAndKzgProofs(blob)
+    if not cellsAndProofs.isOk:
+      return err("Computing Extended Matrix: Issue computing cells and proofs")
+
+    for i in 0..<eip7594.CELLS_PER_EXT_BLOB:
+      extended_matrix.add(MatrixEntry(
+        cell: cellsAndProofs.get.cells[i],
+        kzg_proof: cellsAndProofs.get.proofs[i],
+        row_index: blbIdx.uint64,
+        column_index: i.uint64
+      ))
+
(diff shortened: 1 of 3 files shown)


agnxsh:
Linux Amd failure for Nim 2-0 is unrelated, ref https://github.com/status-im/nimbus-eth2/pull/6549#issuecomment-2351440393

tests/test_eip7594_helpers.nim
```diff
@@ -9,8 +9,57 @@
...
+  test "EIP-7594: Compute Matrix":
+    proc testComputeExtendedMatrix() =
+      const 
+        blob_count = 2
+      let
+        input_blobs = createSampleKzgBlobs(blob_count)
+        extended_matrix = compute_matrix(input_blobs)
+      doAssert extended_matrix.get.len == kzg_abi.CELLS_PER_EXT_BLOB * blob_count
+      let rows = chunks(extended_matrix.get, kzg_abi.CELLS_PER_EXT_BLOB)
+      for row in rows:
```
tersec: This is the only use of `rows`, so it's more efficient in general (doesn't matter as much in the test code, but it's still nicer) to use an `iterator`. This means writing
```nim
      for row in chunks(extended_matrix.get, kzg_abi.CELLS_PER_EXT_BLOB):
        ...
```
where `chunks` becomes not a `func` or `proc` but:
```nim
iterator chunks[T](lst: seq[T], n: int): seq[T] =
  ## Helper that splits a list into N sized chunks.
  for i in countup(0, len(lst) - 1, n):
    yield chunk_of_interest
```
i.e. it's never accumulated, but streamed through

tests/test_eip7594_helpers.nim
```diff
@@ -9,8 +9,57 @@
...
+      "/../vendor/nim-kzg4844/kzg4844/csources/src/trusted_setup.txt", 0).isOk
+
+const MAX_TOP_BYTE = 114
+
+proc createSampleKzgBlobs(n: int): seq[KzgBlob] =
+  var
+    blob: array[BYTES_PER_BLOB, byte]
+    blobs: seq[KzgBlob]
+  for i in 0..<n:
+    discard urandom(blob)
```
tersec: This becomes effectively a kind of (constrained, due to the `MAX_TOP_BYTE` checks) random fuzzing of the input space, which is a useful technique. However, it means that if this check does fail due to unforeseen input-dependency in matrix computation, reproducing that without knowing the exact input is problematic.

There are at least a couple approaches here:
- pre-generate some number of random blobs. Probably too large to be feasible when there's not actually interesting structure in them from the test perspective; and
- either use specifically seeded-pseudorandom PRNGs, which don't necessarily for this purpose have to be cryptographically secure, to fill the blobs, and if the test fails, print the seeds; and
- something like this, but dump the blobs on failure. Possible, but I'm not sure it's worth it vs using a more conventional seeded PRNG to allow reproducing it from an easily printed integer seed.

Regardless, the overall goal is, if the test fails, whether in CI or locally (`make test` for example), it should be possible to investigate without first spending significant time just reproducing the issue.

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -99,6 +99,26 @@ func get_custody_column_list*(node_id: NodeId,
 
   sortedColumnIndexList(ColumnIndex(columns_per_subnet), subnet_ids)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#compute_matrix
+proc compute_matrix*(blobs: seq[KzgBlob]): Result[seq[MatrixEntry], cstring] =
+  ## `compute_matrix` helper demonstrates the relationship between blobs and the `MatrixEntries`
```
tersec: Can wrap 97-character comment to 80 characters

tests/test_eip7594_helpers.nim
```diff
@@ -9,8 +9,62 @@
 {.used.}
 
 import
+  random,
```
tersec: prefer `std/random` for stdlib

agnxsh: ok

PR #6551, opened 2024-09-16
Electra REST forked BeaconBlockBody reading support

tersec:
Won't be merged in September's release cycle, because it affects Deneb REST block reading and can't be readily isolated into future-forks-only codepaths aside from not very usefully disabling it by default.

It's not yet complete, either, because of the requests in `ExecutionPayload`, but those might be changing soon, so won't address them yet.


tersec:
`version-2-0` CI failure is unrelated: https://github.com/nim-lang/Nim/issues/24104

tersec:
https://github.com/status-im/nimbus-eth2/pull/6552 is better

PR #6552, opened 2024-09-16
remove block fork guessing/inference from REST JSON decoding

tersec:
Better, more sustainable alternative to https://github.com/status-im/nimbus-eth2/pull/6551

Not for September release cycle; would be for October cycle.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -129,7 +129,6 @@ RestJson.useDefaultSerializationFor(
   RestDepositContract,
   RestEpochRandao,
   RestEpochSyncCommittee,
-  RestExecutionPayload,
   RestExtraData,
   RestGenesis,
   RestIndexedErrorMessage,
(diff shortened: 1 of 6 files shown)


tersec:
`version-2-0` CI failure is unrelated: https://github.com/nim-lang/Nim/issues/24104

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2854,98 +2854,31 @@ proc decodeBody*(
...
-                                           [version, exc.formatMsg("<data>")]))
-        except CatchableError as exc:
-          return err(RestErrorMessage.init(Http400, UnexpectedDecodeError,
-                                           [version, $exc.msg]))
+      of ConsensusFork.Phase0:    getBlck(phase0.SignedBeaconBlock)
+      of ConsensusFork.Altair:    getBlck(altair.SignedBeaconBlock)
+      of ConsensusFork.Bellatrix: getBlck(bellatrix.SignedBeaconBlock)
+      of ConsensusFork.Capella:   getBlck(capella.SignedBeaconBlock)
+      of ConsensusFork.Deneb:     getBlck(deneb.SignedBeaconBlock)
+      of ConsensusFork.Electra:   getBlck(electra.SignedBeaconBlock)
```
etan-status: `getBlck(consensusFork.SignedBeaconBlock)` avoids the switch as well and makes this maintenance free

etan-status: (`withConsensusFork(consensusFork)` to convert `consensusFork` from `let` to `const`)

tersec: Well, using the usual-fake-Nim-feature which has to be defined in `forks` and which Nim seems not to care much about supporting.

But sure, as long as that's maintained, then it can be used here.

tersec: https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/spec/forks.nim#L402-L522

There's 120 lines of this

tersec: It also increases the coupling which makes for a higher risk/reward setup -- when `withConsensusFork` works automatically, it's nice, but when it doesn't, it's yet another thing which needs to be temporarily disabled/enabled during fork transitions.

The net effect of this kind of transformation is to centralize the case statements into fewer places, with more effect/impact each. That typically decreases maintenance, until it doesn't. Can't put some `debugComment`/`debugAssert`/etc on only the eth2 REST serialization, it has to be in the centralized forks module which affects that and a dozen other things.

But, sure. Maybe risk worth taking, since the point of this overall PR's refactoring is to have REST serialization not use entirely separate block reading for a few specific cases than anything else does, so it should follow other fixes more automatically regardless: https://github.com/status-im/nimbus-eth2/pull/6552/commits/8644d89ad76af4e9c2496d84285fd2b11c34fda9

PR #6553, opened 2024-09-16
add: `recover_matrix`

agnxsh:
This Pr adds the second part to data column reconstruction, where the data columns that would not be unavailable via gossip or req/resp would effectively be reconstructed within the slot period.

ref: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#recover_matrix

### DIFF
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -120,6 +120,41 @@ proc compute_matrix*(blobs: seq[KzgBlob]): Result[seq[MatrixEntry], cstring] =
 
   ok(extended_matrix)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/das-core.md#recover_matrix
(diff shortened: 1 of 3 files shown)


tersec:
https://github.com/status-im/nimbus-eth2/pull/6554 fixes the build issues, for a certain definition of fix

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -120,6 +120,41 @@ proc compute_matrix*(blobs: seq[KzgBlob]): Result[seq[MatrixEntry], cstring] =
...
+      cells: seq[Cell]
+  
+    for e in partial_matrix:
+      if e.row_index == uint64(blob_index):
+        cell_indices.add(e.column_index)
+        cells.add(e.cell)
+
+    let recoveredCellsAndKzgProofs = 
+      recoverCellsAndKzgProofs(cell_indices, cells)
+    if not recoveredCellsAndKzgProofs.isOk:
```
tersec, line 142: `    if recoveredCellsAndKzgProofs.isErr:`

PR #6556, opened 2024-09-17
REST: getBlockRewards() and getSyncCommitteeRewards() implementation

cheatfate:


beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+  ../beacon_node,
+  ../consensus_object_pools/[blockchain_dag, spec_cache, validator_change_pool],
+  ../spec/[forks, state_transition]
+
+export rest_utils
+
+logScope: topics = "rest_rewardsapi"
+
+proc installRewardsApiHandlers*(router: var RestRouter, node: BeaconNode) =
+  let
```
tersec, line 42: Does this work as `const`?

beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+
+    var
+      cache = StateCache()
+      tmpState = assignClone(node.dag.headState)
+
+    if not updateState(
+      node.dag, tmpState[], targetBlock, false, cache):
+        return RestApiResponse.jsonError(Http404, ParentBlockMissingStateError)
+
+    func restore(v: var ForkedHashedBeaconState) =
```
tersec: This `restore` function usage doesn't really make sense if using `tmpState[]`, because now `node.dag.clearanceState`, and `node.dag.headState`  aren't involved at all.

This is what Nimbus has
https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/spec/state_transition.nim#L153-L169
for

beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+
+          for i in 0 ..< min(
+            len(forkyState.data.current_sync_committee.pubkeys),
+            len(sync_aggregate.sync_committee_bits)):
+            let
+              pubkey = forkyState.data.current_sync_committee.pubkeys.data[i]
+              vindex = pubkeyIndices.getOrDefault(pubkey)
+              reward =
+                block:
+                  let res = uint64(get_participant_reward(total_active_balance))
```
tersec: The per-sync-committee-membership-item reward is identical for all sync committee members. This `get_participant_reward` calculation can be lifted out of the loop and not repeated.

If some validator is in the sync committee multiple times, this amount just gets added (or removed, if their contribution was present in the sync committee multiple times) multiple times, but the increment is the same regardess

tersec:
https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest-rules.json?

beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+
+  # https://ethereum.github.io/beacon-APIs/#/Rewards/getBlockRewards
+  router.api2(MethodGet, "/eth/v1/beacon/rewards/blocks/{block_id}") do (
+    block_id: BlockIdent) -> RestApiResponse:
+    let
+      bident = block_id.valueOr:
+        return RestApiResponse.jsonError(Http400, InvalidBlockIdValueError,
+                                         $error)
+
+    if (bident.kind == BlockQueryKind.Named) and
```
tersec: If someone specifies the genesis block in some other way than literally naming it `genesis`, the same logic should apply.

beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+      sync_aggregate =
+        withBlck(bdata):
+          when consensusFork > ConsensusFork.Phase0:
+            forkyBlck.message.body.sync_aggregate
+          else:
+            default(TrustedSyncAggregate)
+
+      targetBlock =
+        withBlck(bdata):
+          if (bident.kind == BlockQueryKind.Named) and
```
tersec: If someone specifies the genesis block in some other way than literally naming it genesis, the same logic should apply.

beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+                  if pubkey in committeeKeys:
+                    res[pubkey] = vindex
+                res
+
+          for i in 0 ..< min(
+            len(forkyState.data.current_sync_committee.pubkeys),
+            len(sync_aggregate.sync_committee_bits)):
+            let
+              pubkey = forkyState.data.current_sync_committee.pubkeys.data[i]
+              vindex = pubkeyIndices.getOrDefault(pubkey)
```
tersec: In theory, this should never hit the "or default" case, because of the `committeeKeys`/`pubkeyIndices` construction above, but if it does, that will incorrectly return validator index 0 as rewarded or penalized.

It's true that the usual `Table` `[]` function generates `KeyError`, but it's still probably better either use the usual `except KeyError: raiseAssert(...)` approach or otherwise ensure that the value it returns in the `OrDefault` case isn't confusable with an actual validator index.

beacon_chain/rpc/rest_rewards_api.nim
```diff
@@ -0,0 +1,226 @@
...
+    let response =
+      withState(tmpState[]):
+        let total_active_balance =
+          get_total_active_balance(forkyState.data, cache)
+        var resp: seq[RestSyncCommitteeReward]
+        when consensusFork > ConsensusFork.Phase0:
+          let
+            keys =
+              block:
+                var res: seq[ValidatorPubKey]
```
tersec: What is the advantage of constructing this as a `seq[ValidatorPubKey]` and then later using `toHashSet` on it rather than directly constructing it as a `HashSet[ValidatorPubKey]` and `incl`ing in the loop.

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: My attempt to convert it to `const` failed with an error
`nimbus-eth2\vendor\nim-stew\stew\ptrops.nim(38, 21) Error: VM does not support 'cast' from tyUInt to tyPointer`

tersec: Oh, not surprising, ok

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6556/commits/b216cbab9be08c17587bcf0f0ecaba4a89f5523e

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6556/commits/b216cbab9be08c17587bcf0f0ecaba4a89f5523e

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6556/commits/b216cbab9be08c17587bcf0f0ecaba4a89f5523e and https://github.com/status-im/nimbus-eth2/pull/6556/commits/23e6d3bbffc2d37dce94a8046dfc1e472b43c132

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6556/commits/b216cbab9be08c17587bcf0f0ecaba4a89f5523e and https://github.com/status-im/nimbus-eth2/pull/6556/commits/23e6d3bbffc2d37dce94a8046dfc1e472b43c132

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6556/commits/b216cbab9be08c17587bcf0f0ecaba4a89f5523e

beacon_chain/rpc/rest_rewards_api.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6556/commits/b216cbab9be08c17587bcf0f0ecaba4a89f5523e

cheatfate:
> https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest-rules.json?

Added in https://github.com/status-im/nimbus-eth2/pull/6556/commits/23e6d3bbffc2d37dce94a8046dfc1e472b43c132

PR #6557, opened 2024-09-17
Attestation API updates for Electra

pedromiguelmiranda:
New GET and POST v2 versions for block and pool attestations according to spec[ ethereum/beacon-APIs/pull/448 ](https://github.com/ethereum/beacon-APIs/pull/448).

~~New Attester slashing endpoints still need to be implemented, given that they still lack support on code base.~~ pool attester slashings api changes #6585


tersec:
Could use updates to https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest-rules.json

Not sure why https://github.com/status-im/nimbus-eth2/pull/6557/commits/1115e4d8fb0a82e39e080179b410f56a2ec23aa8 was premature?

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1302,6 +1322,43 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        let rslot = slot.get()
+        if rslot.isErr():
+          return RestApiResponse.jsonError(Http400, InvalidSlotValueError,
+                                           $rslot.error)
+        Opt.some(rslot.get())
+      else:
+        Opt.none(Slot)
+
+    withConsensusFork(
+      node.dag.cfg.consensusForkAtEpoch(vslot.get().epoch)):
```
tersec: Flip indentation here -- the body of `withConsensusFork` (`if consensusFork < ConsensusFork.Electra:`, et cetera) shold be 2-spaces indented.

So one can do e.g.
```nim
   withConsensusFork(
       node.dag.cfg.consensusForkAtEpoch(vslot.get().epoch)):  # 4-spaces indented to keep it unambiguous/visually clear [and this comment just for exposition in PR discussion of course]
     if consensusFork < ...    # all of this then is 2-space-indented
```

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1349,6 +1406,53 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        if dres.isErr():
+          return RestApiResponse.jsonError(Http400,
+                                           InvalidAttestationObjectError,
+                                           $dres.error)
+        dres.get()
+
+    # Since our validation logic supports batch processing, we will submit all
+    # attestations for validation.
+    let pending =
+      block:
```
tersec: This `block`/`for` loop can be replaced by `mapIt(attestations, node.router.routeAttestation(it))`

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1349,6 +1406,53 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
       RestApiResponse.jsonMsgResponse(AttestationValidationSuccess)
 
+  # https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/submitPoolAttestationsV2
+  router.api2(MethodPost, "/eth/v2/beacon/pool/attestations") do (
+    contentBody: Option[ContentBody]) -> RestApiResponse:
+    let attestations =
+      block:
+        if contentBody.isNone():
+          return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
+        let dres = decodeBody(seq[phase0.Attestation], contentBody.get())
```
cheatfate: This should use different decoding type according to HTTP header `eth-consensus-version' value.

pedromiguelmiranda: excellent catch. thank you.

pedromiguelmiranda:
> Could use updates to https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest-rules.json
> 
> Not sure why [1115e4d](https://github.com/status-im/nimbus-eth2/commit/1115e4d8fb0a82e39e080179b410f56a2ec23aa8) was premature?

The reason why it is premature is because these new endpoint versions are not yet implemented in ncli and without those missing developments, the tests will fail if we add the resttest-rules.json.

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1347,17 +1347,17 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        node.dag.cfg.consensusForkAtEpoch(vslot.get().epoch)):
+      if consensusFork < ConsensusFork.Electra:
+        var res: seq[phase0.Attestation]
+        for item in node.attestationPool[].attestations(vslot, vindex):
+          res.add(item)
+        return RestApiResponse.jsonResponseWVersion(res, consensusFork)
+      else:
+        var res: seq[electra.Attestation]
+        for item in node.attestationPool[].electraAttestations(vslot, vindex):
+          res.add(item)
```
tersec: Should be able to use [toSeq](https://nim-lang.org/docs/sequtils.html#toSeq.t%2Cuntyped) to transform the iterable from `attestations`/`electraAttestations` into the `seq`/`openArray` `jsonResponseWVersion` needs:
```nim
        return RestApiResponse.jsonResponseWVersion(
          toSeq(node.attestationPool[].attestations(vslot, vindex)),
          consensusFork)
```
and
```nim
        return RestApiResponse.jsonResponseWVersion(
          toSeq(node.attestationPool[].electraAttestations(vslot, vindex)),
          consensusFork)
```
the intent is clearer this way

tersec:
> > Could use updates to https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest-rules.json
> > Not sure why [1115e4d](https://github.com/status-im/nimbus-eth2/commit/1115e4d8fb0a82e39e080179b410f56a2ec23aa8) was premature?
> 
> The reason why it is premature is because these new endpoint versions are not yet implemented in ncli and without those missing developments, the tests will fail if we add the resttest-rules.json.

I'm not sure I understand, probably I'm missing something. https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest.nim doesn't really implement any endpoint versions to begin with. It just uses the input from `resttest-rules.json` to prepare basically arbitrary REST requests:
https://github.com/status-im/nimbus-eth2/blob/e3fcd8b031e767ba6cd81bad8e3296556eadb7e4/ncli/resttest.nim#L395-L486 and then compares the results, just using the provided string patterns.

It does not use, for example, `beacon_chain/spec/eth2_apis/rest_beacon_calls.nim` or any similar API definitions

pedromiguelmiranda:
> > > Could use updates to https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest-rules.json
> > > Not sure why [1115e4d](https://github.com/status-im/nimbus-eth2/commit/1115e4d8fb0a82e39e080179b410f56a2ec23aa8) was premature?
> > 
> > 
> > The reason why it is premature is because these new endpoint versions are not yet implemented in ncli and without those missing developments, the tests will fail if we add the resttest-rules.json.
> 
> I'm not sure I understand, probably I'm missing something. https://github.com/status-im/nimbus-eth2/blob/unstable/ncli/resttest.nim doesn't really implement any endpoint versions to begin with. It just uses the input from `resttest-rules.json` to prepare basically arbitrary REST requests:
> 
> https://github.com/status-im/nimbus-eth2/blob/e3fcd8b031e767ba6cd81bad8e3296556eadb7e4/ncli/resttest.nim#L395-L486
> and then compares the results, just using the provided string patterns.
> 
> It does not use, for example, `beacon_chain/spec/eth2_apis/rest_beacon_calls.nim` or any similar API definitions

Apologies, confused with different calls on validator interfaces. 
Which leads to the source of the error, the tests added in [1115e4](https://github.com/status-im/nimbus-eth2/commit/1115e4d8fb0a82e39e080179b410f56a2ec23aa8) were working and show that there is either a use case missing in beacon api spec or that I'm missing some default interpretation of the spec:
- both input parameters (slot and committee_index) are not required. Without a slot, how can we get the correct consensus fork?
- Or, should we just assume the default value of uint64 ( 0 ) and use slot 0 ?

tersec:
The existing [getPoolAttestations](https://ethereum.github.io/beacon-APIs/?urls.primaryName=v2.5.0#/Beacon/getPoolAttestations) semantics, and the apparently intended semantics of [getPoolAttestationsV2](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getPoolAttestationsV2), treat each parameter as an optional filter:
- if only slot is specified, filter on slot;
- if only committee index is specified, filter on committee index;
- if both slot and committee index are specified, filter on both; and
- if neither is specified, don't filter, and enumerate the entire pool.

The existing `getPoolAttestations` implementation does this for the non-V2 call already:
https://github.com/status-im/nimbus-eth2/blob/bf4abf8b9e07c35442b00966f51cc9af5857af33/beacon_chain/rpc/rest_beacon_api.nim#L1277-L1303

However, this does not address
> both input parameters (slot and committee_index) are not required. Without a slot, how can we get the correct consensus fork?

which, indeed. The documentation is clear that it's supposed to be exactly one consensus fork, and this is where V2 is trickier:
> The active consensus version to which the data belongs. Required in response so client can deserialize returned json or ssz data more effectively.

is not in the non-V2 getPoolAttestations, so the question doesn't come up.

> Or, should we just assume the default value of uint64 ( 0 ) and use slot 0 ?
is not useful or specified, so wouldn't do that.

The most spec-correct way I can see is, when `slot` isn't specified:
1. get the current fork
2. get the first epoch of the current fork (`RuntimeConfig` object, various other ways, the REST API knows this information)
3. return only attestations within that epoch. If that means only one slot, so be it. It's a boundary condition, and it's more important to keep the
> The active consensus version to which the data belongs. Required in response so client can deserialize returned json or ssz data more effectively.

invariant than to return a few more attestations for a basically debugging endpoint every few months.

tersec:
This came up in https://github.com/ethereum/beacon-APIs/issues/458 and that's basically the conclusion they came to also

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1349,6 +1405,62 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
     else:
       RestApiResponse.jsonMsgResponse(AttestationValidationSuccess)
 
+  # https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/submitPoolAttestationsV2
+  router.api2(MethodPost, "/eth/v2/beacon/pool/attestations") do (
+    contentBody: Option[ContentBody]) -> RestApiResponse:
+
+    let
+      headerVersion = request.headers.getString("Eth-Consensus-Version")
```
tersec, line 1413: https://github.com/status-im/nim-chronos/blob/c04576d829b8a0a1b12baaa8bc92037501b3a4a0/chronos/apps/http/httptable.nim#L52-L63
```nim
proc getString*(ht: HttpTables, key: string,
                default: string = ""): string =
  ## Returns concatenated value of headers with key ``key``.
  ##
  ## If there multiple headers with the same name ``key`` the result value will
  ## be concatenation using `,`.
  var defseq: seq[string]
  let res = ht.table.getOrDefault(key.toLowerAscii(), defseq)
  if len(res) == 0:
    return default
  else:
    res.join(",")
```
`Eth-Consensus-Version` will never match, because `getString` lowercases the header name first.

tersec: ah, misread the code, not an issue

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1302,6 +1322,46 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+        let rslot = slot.get()
+        if rslot.isErr():
+          return RestApiResponse.jsonError(Http400, InvalidSlotValueError,
+                                           $rslot.error)
+        Opt.some(rslot.get())
+      else:
+        Opt.none(Slot)
+
+    let consensusFork =
+      block:
```
tersec: Don't need this `block:` when entire body of it is one expression (an `if` expression)

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1302,6 +1322,46 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+          node.dag.cfg.consensusForkAtEpoch(node.currentSlot().epoch())
+        else:
+          node.dag.cfg.consensusForkAtEpoch(vslot.get().epoch)
+
+    if consensusFork < ConsensusFork.Electra:
+      return RestApiResponse.jsonResponseWVersion(
+        toSeq(node.attestationPool[].attestations(vslot, vindex)),
+        consensusFork)
+    else:
+      return RestApiResponse.jsonResponseWVersion(
```
tersec, line 1360: This kind of naturally implements the filtering in https://github.com/ethereum/beacon-APIs/issues/458 and that's probably enough.

It seems like the consensus there isn't to filter between forks in post-electra attestation, e.g., if one's in the first slots of next fork Fusaka and requests all slots, this will also get some Electra attestations. That's defensible, and apparently what other CLs do, even if it's not 100% accurate. It's simpler to implement, so can start with these semantics and if the beacon API wants to tighten that later, we can do so.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -695,6 +695,31 @@ proc jsonResponseFinalized*(t: typedesc[RestApiResponse], data: auto,
...
+          writer.beginRecord()
+          writer.writeField("version", version.toString())
+          if exec.isSome():
+            writer.writeField("execution_optimistic", exec.get())
+          writer.writeField("finalized", finalized)
+          writer.writeField("data", data)
+          writer.endRecord()
+          stream.getOutput(seq[byte])
+        except IOError:
+          default
```
tersec, line 720: - [`default`](https://nim-lang.org/docs/system.html#default%2Ctypedesc%5BT%5D) is already a `template` defined by the Nim standard library, so just as a point of naming, unless there's a quite compelling reason to name a symbol literally `default` (or one is writing effectively an type-specialized implementation of that `template`) it's less confusing to use a different name; and
- more significantly, it's unclear why at all to define `var default: seq[byte]` in general and then only use it in the `except` part of the `try`/`except`. One can just directly use `default(seq[byte])` here and only incur the runtime cost on the `except` case.

I do see that various other things in `eth2_rest_serialization` adopt this pattern, but I don't see any reason to.

tersec:
> New Attester slashing endpoints still need to be implemented, given that they still lack support on code base.

https://github.com/status-im/nimbus-eth2/pull/6579

still doesn't have to be done in this PR though

beacon_chain/rpc/rest_beacon_api.nim
pedromiguelmiranda: Glad you find a discussion envolving this topic. Was about to create one to check other CLs.
Regarding next fork Fusaka, it depend of it will required new attestation data structures or not. However, this endpoint (and others with similar strategy) will need to be revisited

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
pedromiguelmiranda: Nice one. 
If I'm not mistaken "default" is a reserved keyword in several languages.  Given that the whole file is using the same nomenclature, I didn't pay much attention to it.

The second point, I would say that the runtime cost is not a problem given that the whole exception mechanism is already a performance hit, but exception handling should be simple and mem allocation/deallocation should never fail. 
Regarding your suggestion, template code will be injected/replace in place, however, what are the impacts of allocating a sequence on except scope? any leak possbility or anything that we might be missing nim-wise?

Neverthless, I'm adressing this on #6580

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -570,6 +570,48 @@ iterator attestations*(
...
+        candidateIdx.get() .. candidateIdx.get()
+      else:
+        1 .. 0
+    else:
+      0 ..< pool.electraCandidates.len()
+
+  for candidateIndex in candidateIndices:
+    for _, entry in pool.electraCandidates[candidateIndex]:
+      ## data.index field from phase0 is still being used while we have
+      ## 2 attestations pools (pre and pos electra). Refer to template addAttToPool
```
tersec: pos -> post

tersec (approved):

tersec:
https://github.com/status-im/nimbus-eth2/pull/6582 should fill in the rest of attester slashing infrastructure

PR #6570, opened 2024-09-20
Do not cache resolved hostnames in REST clients.

cheatfate:


tersec (approved):
For merging after the upcoming release is branched

PR #6581, opened 2024-09-24
add gossip validation for dc, and data column quarantine strategy

agnxsh:


agnxsh:
[https://github.com/status-im/nimbus-eth2/blob/80d71d084cdbb60994dbcfb1f10e36b8cc782443/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L18-L21](url)

Doubt regarding `MaxDataColumns`, I think we should bump this number up (?)

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+import
+  std/tables,
+  ../spec/datatypes/eip7594,
+  ../spec/helpers
+
+from std/sequtils import mapIt
+from std/strutils import join
+
+const
+  MaxDataColumns = 3 * SLOTS_PER_EPOCH * NUMBER_OF_COLUMNS
```
agnxsh, line 19: Doubt regarding `MaxDataColumns`, I think we should bump this number up (?)

beacon_chain/spec/forks.nim
```diff
@@ -16,12 +16,12 @@ import
   "."/[
     block_id, eth2_merkleization, eth2_ssz_serialization,
     forks_light_client, presets],
-  ./datatypes/[phase0, altair, bellatrix, capella, deneb, electra],
+  ./datatypes/[phase0, altair, bellatrix, capella, deneb, electra, eip7594],
```
tersec, line 19: As a general note, at some point this is probably going to be worth refactoring because eip7594 isn't really a fork. But yeah, it's in `datatypes` and that's what matters here. The spec also treats it as a quasi-fork currently.

But once it's merged, it will become just part of some other fork, and for the most part `nimbus-eth2` doens't have a have a `kzg` datatypes, it's part of `deneb`.

But for now, sure.

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+      OrderedTable[(Eth2Digest, ColumnIndex), ref DataColumnSidecar]
+    onDataColumnSidecarCallback*: OnDataColumnSidecarCallback
+  
+  DataColumnFetchRecord* = object
+    block_root*: Eth2Digest
+    indices*: seq[ColumnIndex]
+
+  OnDataColumnSidecarCallback = proc(data: DataColumnSidecar) {.gcsafe, raises: [].}
+
+func shortLog*(x: seq[ColumnIndex]): string =
```
tersec: This should probably live in `beacon_chain/spec/datatypes/eip7594.nim`, since it's for a type defined there. Otherwise, whether `shortLog(...)` works on `ColumnIndex` depends seemingly randomly on whether `beacon_chain/consensus_object_pools/data_column_quarantine.nim` is also imported.

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+
+func hasDataColumns*(quarantine: DataColumnQuarantine,
+    blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock): bool =
+  for idx in 0..<len(blck.message.body.blob_kzg_commitments):
+    if (blck.root, ColumnIndex idx) notin quarantine.data_columns:
+      return false
+  true
+
+func dataColumnFetchRecord*(quarantine: DataColumnQuarantine,
+    blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock): DataColumnFetchRecord =
```
tersec: 88 character line

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -475,6 +492,131 @@ proc validateBlobSidecar*(
...
+  if not verify_block_signature(
+      dag.forkAtEpoch(block_header.slot.epoch),
+      getStateField(dag.headState, genesis_validators_root),
+      block_header.slot,
+      block_root,
+      dag.validatorKey(proposer).get(),
+      data_column_sidecar.signed_block_header.signature):
+    return dag.checkedReject("DataColumnSidecar: Invalid proposer signature")
+
+  # [REJECT] The sidecar's column data is valid as
```
tersec, line 610: Which is more expensive, verifying `check_data_column_sidecar_kzg_proofs` or `verify_block_signature`? Whichever takes longer to run should generally go later in these gossip rejection functions where it's reasonable to reorder them -- means more efficient rejections.

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+      data_column_sidecar.signed_block_header.message
+    if block_header.slot == slot and
+        block_header.proposer_index == proposer_index and
+        data_column_sidecar.index == index:
+      return true
+  false
+
+func popDataColumns*(
+    quarantine: var DataColumnQuarantine, digest: Eth2Digest,
+    blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock):
```
tersec: In `popDataColumns`, `hasDataColumns`, and `dataColumnFetchRecord` are all defined here for both `deneb.SignedBeaconBlock` and `electra.SignedBeaconBlock`. Past devnet 2, running anything `DataColumn`-related on `deneb.SignedBeaconBlock` should be a bug, and ideally be caught by Nim's typechecker. Since `devnet-2` is not running `unstable` anyway, I'm not sure why it's useful to include `deneb.SignedBeaconBlock` in these function signatures.

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+  var indices: seq[ColumnIndex]
+  for i in 0..<len(blck.message.body.blob_kzg_commitments):
+    let idx = ColumnIndex(i)
+    if not quarantine.data_columns.hasKey(
+        (blck.root, idx)):
+      indices.add(idx)
+  DataColumnFetchRecord(block_root: blck.root, indices: indices)
+
+func init*(T: type DataColumnQuarantine,
+           onDataColumnSidecarCallback: OnDataColumnSidecarCallback):
```
tersec: What is this callback for?

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -475,6 +492,131 @@ proc validateBlobSidecar*(
 
   ok()
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#data_column_sidecar_subnet_id
+proc validateDataColumnSidecar*(
+    dag: ChainDAGRef, quarantine: ref Quarantine,
+    dataColumnQuarantine: ref DataColumnQuarantine, data_column_sidecar: DataColumnSidecar,
```
tersec: 91 character line

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+
+func shortLog*(x: seq[ColumnIndex]): string =
+  "<" & x.mapIt($it).join(", ") & ">"
+
+func shortLog*(x: seq[DataColumnFetchRecord]): string =
+  "[" & x.mapIt(shortLog(it.block_root) & shortLog(it.indices)).join(", ") & "]"
+
+func put*(quarantine: var DataColumnQuarantine,
+          dataColumnSidecar: ref DataColumnSidecar) =
+  if quarantine.data_columns.lenu64 >= MaxDataColumns:
```
tersec: Slightly safer and more efficient to do the type conversion at compile-time than runtime here, something like:
```nim
  if quarantine.data_columns.len >= static(MaxDataColumns.int):
```
since `lenu64` is just `.len.uint64`.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -475,6 +492,131 @@ proc validateBlobSidecar*(
...
+  # -- i.e. `blob_sidecar.index < NUMBER_OF_COLUMNS`
+  if not (data_column_sidecar.index < NUMBER_OF_COLUMNS):
+    return dag.checkedReject("DataColumnSidecar: The sidecar's index should be consistent with NUMBER_OF_COLUMNS")
+
+  # [REJECT] The sidecar is for the correct subnet 
+  # -- i.e. `compute_subnet_for_data_column_sidecar(blob_sidecar.index) == subnet_id`.
+  if not (compute_subnet_for_data_column_sidecar(data_column_sidecar.index) == subnet_id):
+    return dag.checkedReject("DataColumnSidecar: The sidecar is not for the correct subnet")
+
+  # [IGNORE] The sidecar is not from a future slot (with a `MAXIMUM_GOSSIP_CLOCK_DISPARITY` allowance)
```
tersec: can wrap this comment block to 80 lines

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+    # data columns that are correctly signed, point to either correct block
+    # root which isn't ever seen, and then for any reason simply never used.
+    var oldest_column_key: (Eth2Digest, ColumnIndex)
+    for k in quarantine.data_columns.keys:
+      oldest_column_key = k
+      break
+    quarantine.data_columns.del(oldest_column_key)
+  let block_root = 
+    hash_tree_root(dataColumnSidecar.signed_block_header.message)
+  discard quarantine.data_columns.hasKeyOrPut(
```
tersec, line 61: `hasKeyOrPut` means that a `(block_root, dataColumnSidecar.index)` pair being inserted into the quarantine blocks new `DataColumnSidecar`s for that same `(block_root, dataColumnSidecar.index)` pair blocks future matching keys.

Does this enable a denial of service attack? It seems like based on https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#data_column_sidecar_subnet_id there are three conditions relevant:
> - _[REJECT]_ The sidecar's `kzg_commitments` field inclusion proof is valid as verified by `verify_data_column_sidecar_inclusion_proof(sidecar)`.
> - _[REJECT]_ The sidecar's column data is valid as verified by `verify_data_column_sidecar_kzg_proofs(sidecar)`.
> - _[IGNORE]_ The sidecar is the first sidecar for the tuple `(block_header.slot, block_header.proposer_index, sidecar.index)` with valid header signature, sidecar inclusion proof, and kzg proof.

Taken together, the first two authenticate that the KZG blobs apply to the block (but, foreshadowing, not vice versa anywhere I can find) and that the columns map to the blobs, respectively.

But an attacker can create arbitrary fake blobs, tie them to the `block_root` they see via gossip, and then construct (valid) proofs, commitments, columns, and the rest from them and race to send the resulting columns so they're received sooner than the legitimate columns.

The gossip validation conditions have no cryptographic way to distinguish between the legitimacy of the two, and so allow it through to this quarantine codepath, where it then sits there and blocks legitimate columns' arrivals, because it's only keyed on `(block_root, dataColumnSidecar.index)`, rather than the full-3-tuple-equivalent:
> - _[IGNORE]_ The sidecar is the first sidecar for the tuple `(block_header.slot, block_header.proposer_index, sidecar.index)` with valid header signature, sidecar inclusion proof, and kzg proof.

which includes the `proposer_index`. Eventually, the attack-columns will get pushed out of the quarantine, and it might recover, but a persistent attacker can keep spamming them.

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+        block_header.proposer_index == proposer_index and
+        data_column_sidecar.index == index:
+      return true
+  false
+
+func popDataColumns*(
+    quarantine: var DataColumnQuarantine, digest: Eth2Digest,
+    blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock):
+    seq[ref DataColumnSidecar] =
+  var r: seq[ref DataColumnSidecar]
```
tersec: Both `seq[ref DataColumnSidecar]` are exactly `DataColumnSidecars`:
https://github.com/status-im/nimbus-eth2/blob/f2d616609982f61790ad19b131fce76f603d02c4/beacon_chain/spec/datatypes/eip7594.nim#L73

beacon_chain/consensus_object_pools/data_column_quarantine.nim
```diff
@@ -0,0 +1,107 @@
...
+
+const
+  MaxDataColumns = 3 * SLOTS_PER_EPOCH * NUMBER_OF_COLUMNS
+  ## Same limit as `MaxOrphans` in `block_quarantine`
+  ## data columns may arrive before an orphan is tagged `columnless`
+
+type
+  DataColumnQuarantine* = object
+    data_columns*:
+      OrderedTable[(Eth2Digest, ColumnIndex), ref DataColumnSidecar]
```
tersec: The `Table` key type here (problematic key per https://github.com/status-im/nimbus-eth2/pull/6581/files#r1774595981 but regardless) seems like a `tuple` version of
https://github.com/status-im/nimbus-eth2/blob/f2d616609982f61790ad19b131fce76f603d02c4/beacon_chain/spec/datatypes/eip7594.nim#L75-L78

Seems cleaner to reuse that type here, since it's [already required per specs](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#datacolumnidentifier), if it's still relevant after addressing the blob-confusion attack.

beacon_chain/spec/forks.nim
tersec, line 19: Eventually will want to fold this `eip7594` pseudo-fork into the `datatypes` module of whatever actual fork it lands in once it's not a slightly extra-spec `_feature`, presumably Fusaka, but for now sure, better to leave it separately.

beacon_chain/spec/network.nim
```diff
@@ -109,6 +109,10 @@ func getBlobSidecarTopic*(forkDigest: ForkDigest,
 func compute_subnet_for_blob_sidecar*(blob_index: BlobIndex): BlobId =
   BlobId(blob_index mod BLOB_SIDECAR_SUBNET_COUNT)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#data_column_sidecar_subnet_id
```
tersec: Correct link seems to be https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#compute_subnet_for_data_column_sidecar

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: agree, moving...

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: fixed

beacon_chain/gossip_processing/gossip_validation.nim
agnxsh: `check_blob_signature` is around 40% faster than `check_data_column_sidecar_kzg_proofs`, so no need really to rearrange

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: fixed

beacon_chain/gossip_processing/gossip_validation.nim
agnxsh: fixed

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: fixed

beacon_chain/spec/network.nim
agnxsh: fixed

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: Done

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: Not relevant in this PR, this would go later in `nimbus_beacon_node`, when we initialise the full node

agnxsh: Can remove it for now

beacon_chain/gossip_processing/gossip_validation.nim
agnxsh: done

beacon_chain/consensus_object_pools/data_column_quarantine.nim
agnxsh: fixed

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -475,6 +492,134 @@ proc validateBlobSidecar*(
 
   ok()
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#data_column_sidecar_subnet_id
```
tersec: Newly added code should be using alpha.8

beacon_chain/spec/network.nim
```diff
@@ -109,6 +109,10 @@ func getBlobSidecarTopic*(forkDigest: ForkDigest,
 func compute_subnet_for_blob_sidecar*(blob_index: BlobIndex): BlobId =
   BlobId(blob_index mod BLOB_SIDECAR_SUBNET_COUNT)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.6/specs/_features/eip7594/p2p-interface.md#compute_subnet_for_data_column_sidecar
```
tersec: Newly added code should be using alpha.8

beacon_chain/consensus_object_pools/data_column_quarantine.nim
tersec: As pointed out, the proposer is checked too.

beacon_chain/gossip_processing/gossip_validation.nim
agnxsh: sure

beacon_chain/spec/network.nim
agnxsh: sure

beacon_chain/consensus_object_pools/data_column_quarantine.nim
tersec: Can be tweaked later.

PR #6583, opened 2024-09-26
Update nim-eth types

arnetheduck:
Minimal changes neede for compatiblity with
https://github.com/status-im/nim-eth/pull/733 which aligns core types with execution spec.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -1133,6 +1133,8 @@ type
 
   AnyConf* = BeaconNodeConf | ValidatorClientConf | SigningNodeConf
 
+  Address = primitives.Address
+
 proc defaultDataDir*[Conf](config: Conf): string =
   let dataDir = when defined(windows):
     "AppData" / "Roaming" / "Nimbus"
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -35,11 +35,13 @@ logScope:
   topics = "elman"
 
 type
+  FixedBytes[N: static int] =  web3.FixedBytes[N]
   PubKeyBytes = DynamicBytes[48, 48]
   WithdrawalCredentialsBytes = DynamicBytes[32, 32]
   SignatureBytes = DynamicBytes[96, 96]
   Int64LeBytes = DynamicBytes[8, 8]
   WithoutTimeout* = distinct int
+  Address = web3.Address
 
   SomeEnginePayloadWithValue =
     BellatrixExecutionPayloadWithValue |
--- a/beacon_chain/libnimbus_lc/libnimbus_lc.nim
+++ b/beacon_chain/libnimbus_lc/libnimbus_lc.nim
@@ -1282,25 +1282,25 @@ proc ETHExecutionBlockHeaderCreateFromJson(
   if data.nonce.isNone:
     return nil
   let blockHeader = ExecutionBlockHeader(
-    parentHash: data.parentHash.asEth2Digest,
-    ommersHash: data.sha3Uncles.asEth2Digest,
-    coinbase: distinctBase(data.miner),
-    stateRoot: data.stateRoot.asEth2Digest,
-    txRoot: data.transactionsRoot.asEth2Digest,
-    receiptsRoot: data.receiptsRoot.asEth2Digest,
-    logsBloom: distinctBase(data.logsBloom),
+    parentHash: data.parentHash.asEth2Digest.to(Hash32),
+    ommersHash: data.sha3Uncles.asEth2Digest.to(Hash32),
+    coinbase: distinctBase(data.miner).to(EthAddress),
+    stateRoot: data.stateRoot.asEth2Digest.to(Hash32),
+    transactionsRoot: data.transactionsRoot.asEth2Digest.to(Hash32),
+    receiptsRoot: data.receiptsRoot.asEth2Digest.to(Hash32),
+    logsBloom: distinctBase(data.logsBloom).to(Bloom),
     difficulty: data.difficulty,
     number: distinctBase(data.number),
     gasLimit: distinctBase(data.gasLimit),
     gasUsed: distinctBase(data.gasUsed),
     timestamp: EthTime(distinctBase(data.timestamp)),
     extraData: distinctBase(data.extraData),
-    mixHash: data.mixHash.asEth2Digest,
-    nonce: distinctBase(data.nonce.get),
+    mixHash: data.mixHash.asEth2Digest.to(Hash32),
+    nonce: distinctBase(data.nonce.get).to(Bytes8),
     baseFeePerGas: data.baseFeePerGas,
     withdrawalsRoot:
       if data.withdrawalsRoot.isSome:
-        Opt.some(data.withdrawalsRoot.get.asEth2Digest)
+        Opt.some(data.withdrawalsRoot.get.asEth2Digest.to(Hash32))
       else:
         Opt.none(ExecutionHash256),
     blobGasUsed:
@@ -1315,12 +1315,12 @@ proc ETHExecutionBlockHeaderCreateFromJson(
         Opt.none(uint64),
     parentBeaconBlockRoot:
       if data.parentBeaconBlockRoot.isSome:
-        Opt.some distinctBase(data.parentBeaconBlockRoot.get.asEth2Digest)
+        Opt.some data.parentBeaconBlockRoot.get.asEth2Digest.to(Hash32)
       else:
         Opt.none(ExecutionHash256),
     requestsRoot:
       if data.requestsRoot.isSome:
-        Opt.some(data.requestsRoot.get.asEth2Digest)
+        Opt.some(data.requestsRoot.get.asEth2Digest.to(Hash32))
       else:
         Opt.none(ExecutionHash256))
   if rlpHash(blockHeader) != executionHash[]:
@@ -1342,7 +1342,7 @@ proc ETHExecutionBlockHeaderCreateFromJson(
         wd = ExecutionWithdrawal(
           index: distinctBase(data.index),
           validatorIndex: distinctBase(data.validatorIndex),
-          address: distinctBase(data.address),
+          address: distinctBase(data.address).to(EthAddress),
           amount: distinctBase(data.amount))
         rlpBytes =
           try:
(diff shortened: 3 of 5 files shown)


beacon_chain/libnimbus_lc/libnimbus_lc.nim
```diff
@@ -1861,10 +1861,8 @@ proc ETHTransactionsCreateFromJson(
         of DestinationType.Regular:
           tx.to.get
         of DestinationType.Create:
-          var res {.noinit.}: array[20, byte]
-          res[0 ..< 20] = keccakHash(rlp.encodeList(fromAddress, tx.nonce))
-            .data.toOpenArray(12, 31)
-          res
+          let hash = keccakHash(rlp.encodeList(fromAddress, tx.nonce))
+          hash.to(EthAddress)
```
etan-status, line 1865: This one's a bit tricky, keccak returns 32 bytes and `to` suggests simple data type conversion. The idea here is to drop the prefix and only keep the last 20 bytes.

beacon_chain/libnimbus_lc/libnimbus_lc.nim
```diff
@@ -1890,14 +1888,14 @@ proc ETHTransactionsCreateFromJson(
...
       value: tx.value,
       input: tx.payload,
       accessList: tx.accessList.mapIt(ETHAccessTuple(
-        address: ExecutionAddress(data: it.address),
-        storageKeys: it.storageKeys.mapIt(Eth2Digest(data: it)))),
+        address: ExecutionAddress(data: it.address.data),
+        storageKeys: it.storageKeys.mapIt(Eth2Digest(data: it.data)))),
       maxFeePerBlobGas: tx.maxFeePerBlobGas,
-      blobVersionedHashes: tx.versionedHashes,
+      blobVersionedHashes: tx.versionedHashes.mapIt(Eth2Digest(data: it.data)),
```
etan-status, line 1898: isn't this `asEth2Digest`? (if we want to align type names maybe `Eth2Digest` -> `Root` could be another one)

beacon_chain/spec/helpers.nim
```diff
@@ -462,6 +462,12 @@ func toExecutionWithdrawal*(
     address: EthAddress withdrawal.address.data,
     amount: distinctBase(withdrawal.amount))
 
+proc append(w: var RlpWriter, val: EthAddress) =
+  # TODO for some reason, the same implementation from addresses_rlp isn't
+  #      picked up, so we need to repeat it here - odd.
+  mixin append
+  w.append(val.data())
+
```
etan-status: `mixin` don't always work from generics / templates when the mixed-in stuff isn't re-exported all the way to the innermost non-generic/non-template caller. we have the same problem in `helpers_el.nim` and there the fix was just to wrap the call that requires mixin internally in a non-generic func to avoid polluting the entire codebase with rlp exports.

beacon_chain/spec/helpers.nim
```diff
@@ -604,7 +610,7 @@ proc blockToBlockHeader*(blck: ForkyBeaconBlock): ExecutionBlockHeader =
     requestsRoot          : requestsRoot)          # EIP-7685
 
 proc compute_execution_block_hash*(blck: ForkyBeaconBlock): Eth2Digest =
-  rlpHash blockToBlockHeader(blck)
+  rlpHash(blockToBlockHeader(blck)).to(Eth2Digest)
```
etan-status, line 612: what's the difference between `Eth2Digest` and `Root`?

beacon_chain/conf.nim
```diff
@@ -1133,6 +1133,8 @@ type
 
   AnyConf* = BeaconNodeConf | ValidatorClientConf | SigningNodeConf
 
+  Address = primitives.Address
+
```
etan-status, line 1137: This one is not exported, and it worked before without it, is it necessary?

beacon_chain/el/el_manager.nim
```diff
@@ -35,11 +35,13 @@ logScope:
   topics = "elman"
 
 type
+  FixedBytes[N: static int] =  web3.FixedBytes[N]
```
etan-status, line 38:
```suggestion
  FixedBytes[N: static int] = web3.FixedBytes[N]
```

this one is also not exported and it worked without this definition before.

beacon_chain/libnimbus_lc/libnimbus_lc.nim
```diff
@@ -1282,25 +1282,25 @@ proc ETHExecutionBlockHeaderCreateFromJson(
...
     difficulty: data.difficulty,
     number: distinctBase(data.number),
     gasLimit: distinctBase(data.gasLimit),
     gasUsed: distinctBase(data.gasUsed),
     timestamp: EthTime(distinctBase(data.timestamp)),
     extraData: distinctBase(data.extraData),
-    mixHash: data.mixHash.asEth2Digest,
-    nonce: distinctBase(data.nonce.get),
+    mixHash: data.mixHash.asEth2Digest.to(Hash32),
+    nonce: distinctBase(data.nonce.get).to(Bytes8),
```
etan-status, line 1299: shouldn't nonce be a number type? is it still getting serialized properly with this change? as in, as Quantity and not as a byte array, on engine?

etan-status (commented):
`ethereum/execution-specs` does not pass full `ethereum/execution-spec-tests` at this stage (constantinople doesn't work, and 7702 support only got added recently). `ethereum/execution-spec-tests` is generated with ethereumjs. Still agreeing with the overall change though, sharing identifiers that are closer to "specs" (where specs exist -- devp2p specs are not even reflecting dencun changes yet).

beacon_chain/libnimbus_lc/libnimbus_lc.nim
arnetheduck: https://github.com/ethereum/execution-specs/blob/51fac24740e662844446439ceeb96a460aae0ba0/src/ethereum/paris/blocks.py#L41

beacon_chain/conf.nim
arnetheduck: `nim-eth` adds an `Address` symbol - this local type has precedence over everything imported so it sets a preference between primitives and nim-eth for the rest of the file

beacon_chain/libnimbus_lc/libnimbus_lc.nim
etan-status: is that compatible with engine API?

etan-status: (or, json-rpc, probably the better one, as nonce is not used in engine itself)

etan-status: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash says it's bytes8, but the reality seems to encode it as quantity:

```sh
 curl https://docs-demo.quiknode.pro/ \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{"method":"eth_getBlockByNumber","params":["0xc5043f",true],"id":1,"jsonrpc":"2.0"}' | jq '.'
```

e.g. here nonce is "0x1d" not a bytes8

```sh
      {
        "blockHash": "0xa917fcc721a5465a484e9be17cda0cc5493933dd3bc70c9adbee192cb419c9d7",
        "blockNumber": "0xc5043f",
        "from": "0xb0cba44805f096094e8759ee9824a05590c10f27",
        "gas": "0x5208",
        "gasPrice": "0x7f084b500",
        "hash": "0x3dc2776aa483c0eee09c2ccc654bf81dccebead40e9bb664289637bfb5e7e954",
        "input": "0x",
        "nonce": "0x1d",
        "to": "0xb0cba44805f096094e8759ee9824a05590c10f27",
        "transactionIndex": "0xcb",
        "value": "0x3f3b45b9331",
        "type": "0x0",
        "chainId": "0x1",
        "v": "0x26",
        "r": "0x19610db3a46538e10c4db8597510902374ff66ef339a2251c82faebb3c00f8b9",
        "s": "0x7af4c008f4bc1edbe081c8e32f49893f035a9eeacb615dd67808f6a9f5e13b52"
      }
```

the specs don't seem authoritative. https://github.com/ethereum/devp2p/blob/master/caps/eth.md here the blocks don't even have excess blob fees or parent beacn blob roots, as per spec, in reality they have.

or also, there's https://github.com/ethereum/execution-apis/pull/573 which is implemented in practice but is not specced out like that.

arnetheduck: `BytesX` _generally_ also encodes as `0x...` in the new nim-eth - but that doesn't matter, we're dealing with an eth type here, not web3 - this is internal to the execution client, not a web3 impl

beacon_chain/spec/helpers.nim
arnetheduck: https://github.com/ethereum/execution-specs/blob/master/src/ethereum/paris/fork_types.py#L29

`Root == Hash32 == distinct array` - `Eth2Digest == MDigest[256] == object`

it's a rather pointless alias in the spec

beacon_chain/libnimbus_lc/libnimbus_lc.nim
arnetheduck: https://github.com/status-im/nim-eth/pull/733/files#diff-cfc06cb7a2a013ce859388f992b07f54d334dae23b5e6ae22d32faba8cc38c64R29

It "kind of" fits with the idea of a canonical conversion from `Hash` to `Address` - in the spec, it's even defined as dropping the first 12 bytes supporting the `to` idea and being well-defined.

arnetheduck: I guess it could be named something else, if you happen to have a good name handy :)

beacon_chain/libnimbus_lc/libnimbus_lc.nim
arnetheduck: most of this will go away in the next set of pr:s which will reuse the new nim-eth types in nim-web3, getting rid of most of the conversion friction - this is just a "make it work" step.

beacon_chain/libnimbus_lc/libnimbus_lc.nim
jangko: ExecutionPayload of web3 is not using nonce from eth block header. And the example posted by @etan-status  is not a block header, but a transaction. And a transaction's nonce is a uint64. There will be no conversion friction or serialization issue in the case of this block header nonce. And there is no problem also with the json encoding nonce of block header so far in our json-rpc.

beacon_chain/spec/helpers.nim
arnetheduck: yeah, this didn't work at all ideed ; it's a pretty annoying bug though in lookup

PR #6595, opened 2024-10-04
Fix last-seen-p2p-address values in /eth/v1/node/peers API endpoint.

cheatfate:
… field.

### DIFF
--- a/beacon_chain/rpc/rest_node_api.nim
+++ b/beacon_chain/rpc/rest_node_api.nim
@@ -7,6 +7,7 @@
 {.push raises: [].}
 
 import
+  std/algorithm,
   stew/byteutils,
   chronicles,
   eth/p2p/discoveryv5/enr,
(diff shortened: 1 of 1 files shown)


beacon_chain/rpc/rest_node_api.nim
```diff
@@ -97,15 +98,93 @@ proc toString(direction: PeerType): string =
...
+              (address, boost + 2)
+            else:
+              (address, boost + 11)
+          elif IP6.matchPartial(address):
+            let address6 =
+              address.getProtocolArgument(multiCodec("ip6")).valueOr:
+                continue
+            var ta6 = TransportAddress(family: AddressFamily.IPv6)
+            ta6.address_v6[0 .. 15] = address6[0 .. 15]
+            if ta6.isLoopback():
```
Redidacove, line 166: I think using switch case statement would be better here for code readability as there are too many if else statements already in parent statement. wdyt

beacon_chain/rpc/rest_node_api.nim
```diff
@@ -97,15 +98,93 @@ proc toString(direction: PeerType): string =
...
+          # TODO (cheatfate): We match TCP here because `nim-libp2p` do not have
+          # QUIC support yet. So we give TCP addresses priority.
+          let boost = if isUDP: 100 else: 0
+          if IP4.matchPartial(address):
+            let address4 =
+              address.getProtocolArgument(multiCodec("ip4")).valueOr:
+                continue
+            var ta4 = TransportAddress(family: AddressFamily.IPv4)
+            ta4.address_v4[0 .. 3] = address4[0 .. 3]
+            if ta4.isLoopback():
```
Redidacove, line 150: here as well

beacon_chain/rpc/rest_node_api.nim
```diff
@@ -97,15 +98,93 @@ proc toString(direction: PeerType): string =
...
+
 proc getLastSeenAddress(node: BeaconNode, id: PeerId): string =
-  # TODO (cheatfate): We need to provide filter here, which will be able to
-  # filter such multiaddresses like `/ip4/0.0.0.0` or local addresses or
-  # addresses with peer ids.
-  let addrs = node.network.switch.peerStore[AddressBook][id]
-  if len(addrs) > 0:
-    $addrs[len(addrs) - 1]
+  # Because `last_seen_p2p_address` is single address but libp2p reports lot of
+  # addresses, we going to sort list of addresses before selecting one. Here
```
Redidacove, line 115:
```suggestion
  # addresses, we are going to sort list of addresses before selecting one. Here
```

Redidacove (changes requested):

Issue #6297, opened 2024-05-17
Single command for beacon node and checkpoint sync. Remove separate command for `trustedNodeSync`


jgresham:
**Is your feature request related to a problem? Please describe.**
No just a UX improvement

**Describe the solution you'd like**
No need to run `trustedNodeSync` as a separate process/command at the beginning. Just run one command like so:
`<start beacon node as normal> --trusted-node-url=<checkpoint-sync-url>`

**Describe alternatives you've considered**
Most other clients support this

**Additional context**
This will help simplify logic for apps that run nodes for users

cmd0s:
That will be a good improvement. 
In addition, it will be nice to have API to monitor trustedNodeSync progress.

marmarek:
I'd prefer to retain separate command (even if standard run would support trusted node sync too). I have a connection to the trusted node via a tunnel that needs to be specifically enabled, which means:
1. It won't be available on further startup, so using the option for normal startup means I would need to add to add it to standard startup script and then remove after the sync.
2. I need to know when sync completes to cleanup the tunnel, and it's more convenient to script by simply seeing when the command completes, instead of watching an API (and then triggering service restart anyway - see earlier point).

tersec:
A general point is that within the weak subjectivity period, a node can trust its own state not to be mislead by arbitrary hostile/malicious input, but if there's an automated mechanism that runs ever startup or similarly which allows effectively rewriting a node's perceived history and puts it outside the safety of this weak subjectivity period, that's a potential attack vector.

We might look at this at some point, but right now, it would fundamentally alter the trust model of a node from only having to trust its own data to every time having to trust some external entity, precisely because the advantage here would be that one does't have to remove the TNS parameters each time.

The T stands for trusted, yes, but part of that is the verification by another node, per what the Nimbus guide suggests, and that's not feasible in an automated way.

We're exploring alternate approaches to this, but as it stands, we're not likely to change this without other circumstances changing.

PR #6598, opened 2024-10-05
Remove `trustedNodeSync` logic from default CLI command flow

chirag-parmar:
### Changelog

- Removes the `external-beacon-api-url` from the default CLI command (`noCommand`)
- marks related CLI options `trusted-state-root` and `trusted-block-root` as deprecated in the default CLI flow
- tidies deprecated options to be processed at one place
- updates descriptions for deprecated options

### Rationale

- View the discussion [here](https://github.com/status-im/nimbus-eth2/issues/6297#issuecomment-2228735449)
- The logic is stale and doesn't adapt well with many centralized checkpoint providers

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -225,7 +225,7 @@ type
 
     requireEngineAPI* {.
       hidden  # Deprecated > 22.9
-      desc: "Require Nimbus to be configured with an Engine API end-point after the Bellatrix fork epoch"
+      desc: "Require Nimbus to be configured with an Engine API end-point after the Bellatrix fork epoch (deprecated)"
       name: "require-engine-api-in-bellatrix" .}: Option[bool]
 
     nonInteractive* {.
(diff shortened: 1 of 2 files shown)


etan-status:
> The logic is stale and doesn't adapt well with many centralized checkpoint providers

https://github.com/ethpandaops/checkpointz/issues/143

checkpointz is GPL-3.0 licensed so I personally prefer not touching it (risk of tainting our own permissive code with GPL), but in principle someone could simply adjust it to provide the LC data, making this mechanism work fine with `--trusted-block-root` parameter (it already works with `--trusted-state-root`).

chirag-parmar:
Closing this PR for the rationale below

There are two logical flows for a checkpoint sync in Nimbus. 
1. Using the `trustedNodeSync` subcommand, which by default grabs the `finalized` state from either a centralized checkpoint server or a manually configured trusted node (`--trusted-node-url`). The subcommand also accepts two other parameters `--trusted-state-root`, for specifying an explicit state that the user trusts in, OR, `--trusted-block-root` for specifying an explicit block root against which light client data is queried for the sync ([explanation](https://github.com/ethpandaops/checkpointz/issues/143)). 
3. Using `--external-beacon-api-url` which provides the same parameters(`--trusted-state-root` and `--trusted-block-root`) as above but does not provide a default behaviour. This mandates either a state root or a block root to be provided.

The rationale for removing `--external-beacon-api-url`(this PR) was to remove redundancy in code and to require an explicit interaction from the user for the "trusted" nature of the sync. However, since the second flow of sync requires explicit specification of a state root (or block root) it still maintains the explicit interactions while providing a 'one-shot' command to sync AND start the node.

[This PR](https://github.com/status-im/nimbus-eth2/pull/6515) solves all our problems for the long term scenario

PR #6606, opened 2024-10-07
fix deprecated warnings

narimiran:
_I don’t like deprecation warnings. They're coarse and rough and irritating… and they get everywhere._

### DIFF
--- a/beacon_chain/spec/presets.nim
+++ b/beacon_chain/spec/presets.nim
@@ -42,7 +42,7 @@ type
 
     # Transition
     TERMINAL_TOTAL_DIFFICULTY*: UInt256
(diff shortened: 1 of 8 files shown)


beacon_chain/spec/presets.nim
```diff
@@ -164,7 +164,7 @@ when const_preset == "mainnet":
     TERMINAL_TOTAL_DIFFICULTY:
       u256"115792089237316195423570985008687907853269984665640564039457584007913129638912",
     # By default, don't use these params
-    TERMINAL_BLOCK_HASH: BlockHash.fromHex(
+    TERMINAL_BLOCK_HASH: Hash32.fromHex(
```
arnetheduck, line 167: this is `hash32"0x..."` in the new code

PR #6610, opened 2024-10-08
Ordered trie for trie root computations

arnetheduck:
Gives a 100x speed boost for this operation which happens during optimistic sync when computing the block hash.

### DIFF
--- a/beacon_chain/libnimbus_lc/libnimbus_lc.nim
+++ b/beacon_chain/libnimbus_lc/libnimbus_lc.nim
@@ -1363,7 +1363,7 @@ proc ETHExecutionBlockHeaderCreateFromJson(
         amount: wd.amount,
         bytes: rlpBytes)
 
-    var tr = orderedTrieRoot(wds)
+    let tr = orderedTrieRoot(wds)
     if tr != data.withdrawalsRoot.get.asEth2Digest:
       return nil
 
--- a/beacon_chain/libnimbus_lc/libnimbus_lc.nim
+++ b/beacon_chain/libnimbus_lc/libnimbus_lc.nim
@@ -9,11 +9,11 @@
 
 import
   std/[json, sequtils, times],
-  eth/common/[eth_types_rlp, transaction],
+  eth/common/eth_types_rlp,
   eth/keys,
   eth/p2p/discoveryv5/random2,
   eth/rlp,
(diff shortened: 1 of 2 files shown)


beacon_chain/libnimbus_lc/libnimbus_lc.nim
```diff
@@ -1357,13 +1363,8 @@ proc ETHExecutionBlockHeaderCreateFromJson(
...
 
-    var tr = initHexaryTrie(newMemoryDB())
-    for i, wd in wds:
-      try:
-        tr.put(rlp.encode(i.uint), wd.bytes)
-      except RlpError:
-        raiseAssert "Unreachable"
-    if tr.rootHash() != data.withdrawalsRoot.get.asEth2Digest:
+    var tr = orderedTrieRoot(wds)
+    if tr != data.withdrawalsRoot.get.asEth2Digest:
```
etan-status:
```suggestion
    let tr = orderedTrieRoot(wds)
    if tr != data.withdrawalsRoot.get.asEth2Digest:
```

beacon_chain/spec/helpers.nim
```diff
@@ -440,115 +440,67 @@ func compute_timestamp_at_slot*(state: ForkyBeaconState, slot: Slot): uint64 =
...
 proc computeRequestsTrieRoot(
     requests: electra.ExecutionRequests): ExecutionHash256 =
-  if requests.deposits.len == 0 and
-      requests.withdrawals.len == 0 and
-      requests.consolidations.len == 0:
+  let n = requests.deposits.len +
+      requests.withdrawals.len +
+      requests.consolidations.len
+
+  if n == 0:
```
etan-status: Can the `EMPTY_ROOT_HASH` optimization be integrated into the `OrderedTrieRootBuilder`? Then we don't need `EMPTY_ROOT_HASH` in `helpers` anymore.

arnetheduck: > Can the EMPTY_ROOT_HASH optimization be integrated into the OrderedTrieRootBuilder? Then we don't need EMPTY_ROOT_HASH in helpers anymore.

it's already there actually, [429bb2e](https://github.com/status-im/nimbus-eth2/pull/6610/commits/429bb2ec9651541eaed5665b8c123ac778a4adbb)

PR #6613, opened 2024-10-08
bump nimbus-build-system to always bundle checksums

tersec:
- https://github.com/status-im/nimbus-build-system/pull/89

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -5,7 +5,7 @@
   # Options: nimbus_light_client, nimbus_validator_client, nimbus_signing_node, all
   targets ? ["nimbus_beacon_node"],
   # Options: 0,1,2
-  verbosity ? 0,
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -865,9 +865,6 @@ proc init*(T: type BeaconNode,
   func getDenebForkEpoch(): Opt[Epoch] =
     Opt.some(cfg.DENEB_FORK_EPOCH)
 
-  func getElectraForkEpoch(): Opt[Epoch] =
```
arnetheduck: intended?

tersec: yeah, not used, and this seemed as harmless as any a PR to include it in.

PR #6614, opened 2024-10-08
use types from `eth/common/eth_types`

narimiran:
Depends on https://github.com/status-im/nim-eth/pull/745 — `EthWithdrawal` is introduced there.


etan-status (approved):
lc stuff looks good

PR #6657, opened 2024-10-17
update to devnet-4-compatible engine API

tersec:


beacon_chain/libnimbus_lc/libnimbus_lc.nim
etan-status, line 1: LGTM. the EL no longer provides requests, I think, and the libnimbus_lc function is specifically to parse, verify and expose execution data there from execution JSON-RPC. so, the depositrequests / withdrawalrequests / consolidationrequests could be removed and replaced with an api that exposes the requests hash (similar to state root / receipts root except it's not an MPT root but a random geth hash)

etan-status (approved):
the libnimbus_lc changes can be applied lateron, beyond what's already in this PR.

beacon_chain/libnimbus_lc/libnimbus_lc.nim
tersec: ok, I want this to land asap, so I'm going to let it do that.

But follow-up PRs can handle this.

PR #6659, opened 2024-10-20
json_rpc: bump

arnetheduck:
* refactored error handling, logging and other fixes
* crash fix on overlong array in json


arnetheduck:
this one is a bit risky, can wait til the next release

arnetheduck:
this will need a look to see that deposits syncing is still fine - should be, because the client updates to json-rpc are mostly harmless, but still

tersec:
Sure, `testing` branch updated but like to keep `unstable` clear of anything risky for 3 or 4 days after, ideally until actual release, in case small fixes need to go into `testing` with minimal disruption/risk.

Hasn't happened yet, but in case.

PR #6660, opened 2024-10-20
add column support to beacon chain db

agnxsh:


beacon_chain/beacon_chain_db.nim
```diff
@@ -561,6 +571,8 @@ proc new*(T: type BeaconChainDB,
 
   var blobs = kvStore db.openKvStore("deneb_blobs").expectDb()
 
+  var columns = kvStore db.openKvStore("electra_columns").expectDb()
```
etan-status: this should be gated on `ELECTRA_FORK_EPOCH != FAR_FUTURE_EPOCH`, as in, if electra is not scheduled don't touch the DB with the experimental data.

agnxsh: sure ok

beacon_chain/beacon_chain_db.nim
```diff
@@ -20,7 +20,7 @@ import
           forks,
           presets,
           state_transition],
-  ./spec/datatypes/[phase0, altair, bellatrix],
+  ./spec/datatypes/[phase0, altair, bellatrix, eip7594],
```
tersec: https://github.com/status-im/nimbus-eth2/blob/250a80eb0ce714d000e88dbb67e0ea61fb2c2073/beacon_chain/spec/forks.nim#L22-L25
from `spec/forks` already exports all of these, so there's no need to import of these manually (`phase`, `altair`, `bellatrix`, or `eip7594` from `datatypes`) since it already imports `forks`.

The whole line can be removed.

agnxsh: I thought eip7594 was being treated as a feature and not a fork, which is why I think I had removed it from the `forks.nim` import re-export thing. But ok can revert.

tersec: Well minimally, removing `phase0`, `altair`, and `bellatrix`.

tersec: Also, descriptively, https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/spec/forks.nim does export `eip7594`

agnxsh: ah ok resolving this

tersec:
My concern is that PeerDAS won't, in fact, be in PectrA/Pectra, and as such this will trigger for end-users when the first of Sepolia or Holesky forks. Want to keep this out of `unstable` until that's resolved. One way of doing this is to create new Fulu fork in the codebase and merge fork-specific PeerDAS code into that, which presumably is where it will ship, and if not, can rename/move later.

tests/testdbutil.nim
```diff
@@ -26,6 +26,7 @@ proc makeTestDB*(
     flags: UpdateFlags = {},
     cfg = defaultRuntimeConfig): BeaconChainDB =
   # Blob support requires DENEB_FORK_EPOCH != FAR_FUTURE_EPOCH
+  # Data column support requires ELECTRA_FORK_EPOCH != FAR_FUTURE_EPOCH
```
tersec: The check in this PR is `if cfg.FULU_FORK_EPOCH != FAR_FUTURE_EPOCH`

agnxsh: yes missed it

PR #6662, opened 2024-10-20
don't use {.noinit.} with case objects

tersec:
The semantics aren't compatible, because to then assign to the case objects, it needs to check the existing discriminators.

### DIFF
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -170,7 +170,7 @@ proc updateLightClientFromDag*(node: BeaconNode) =
 
   let bdata = node.dag.getForkedBlock(dagHead.blck.bid).valueOr:
     return
-  var header {.noinit.}: ForkedLightClientHeader
+  var header: ForkedLightClientHeader
   withBlck(bdata):
     const lcDataFork = lcDataForkAtConsensusFork(consensusFork)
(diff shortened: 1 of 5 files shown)


etan-status:
Why does assigning to a case object read its discriminator?

etan-status:
As in, these assignments overwrite the full object without caring about what case it was before.

arnetheduck:
> As in, these assignments overwrite the full object without caring about what case it was before.

technically, on assignment the compiler needs to generate code for releasing ref counts and running destructors and to know what is actually stored in the type, it has to read the kind field - technically, for variants that only have trivial types, it should not have to do this - otoh, it would also be harmless to do an uninit read because no matter the type, it would be a no-op, though this is ub and ub is bad.

etan-status:
doesn't that mean that `{.noinit.}` can never be used? except for primitive types

arnetheduck:
> doesn't that mean that {.noinit.} can never be used? except for primitive types

pretty much, yes - the closest concept in C++ is the trivial type, ie primitive types and trivial combinations thereof (structs and arrays with simple members, no constructor and no destructor) - same goes for nim.

etan-status:
Probably more restrictive, as the case object shows...

etan-status:
There's no `ref` involved here, and also no constructor

PR #6665, opened 2024-10-21
BN: Fix el_manager timeouts issue in block processing.

cheatfate:
Use predefined array of exponential timeouts when all the requests to EL has been failed.

So now both operations `newPayload` and `updateHead` sharing single timeout of `4.seconds` on `mainnet` preset.

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -251,8 +251,8 @@ proc variedSleep*(
 ): Future[void] {.async: (raises: [CancelledError], raw: true).} =
   doAssert(len(durations) > 0, "Empty durations array!")
   let index =
-    if (counter < 0) or (counter > len(durations)):
-      len(durations) - 1
+    if (counter < 0) or (counter > high(durations)):
+      high(durations)
     else:
       counter
   inc(counter)
@@ -1073,7 +1073,7 @@ proc sendNewPayload*(
     blck: SomeForkyBeaconBlock
 ): Future[PayloadExecutionStatus] {.
     async: (raises: [CancelledError], raw: true).} =
-  sendNewPayload(m, blck, DeadlineObject.init(FORKCHOICEUPDATED_TIMEOUT))
+  sendNewPayload(m, blck, DeadlineObject.init(NEWPAYLOAD_TIMEOUT))
 
 proc forkchoiceUpdatedForSingleEL(
     connection: ELConnection,
@@ -1262,7 +1262,7 @@ proc forkchoiceUpdated*(
                        Opt[PayloadAttributesV2] |
                        Opt[PayloadAttributesV3]
 ): Future[(PayloadExecutionStatus, Opt[BlockHash])] {.
-   async: (raises: [CancelledError], raw: true).} =
+    async: (raises: [CancelledError], raw: true).} =
   forkchoiceUpdated(
     m, headBlockHash, safeBlockHash, finalizedBlockHash,
     payloadAttributes, DeadlineObject.init(FORKCHOICEUPDATED_TIMEOUT))
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -34,6 +34,10 @@ export
 logScope:
   topics = "elman"
 
+const
+  SleepDurations =
+    [100.milliseconds, 200.milliseconds, 500.milliseconds, 1.seconds]
+
 type
   FixedBytes[N: static int] =  web3.FixedBytes[N]
   PubKeyBytes = DynamicBytes[48, 48]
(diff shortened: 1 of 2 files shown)


beacon_chain/el/el_manager.nim
```diff
@@ -233,6 +242,22 @@ declareCounter engine_api_last_minute_forkchoice_updates_sent,
...
+proc init*(t: typedesc[DeadlineObject], d: Duration): DeadlineObject =
+  DeadlineObject(future: sleepAsync(d))
+
+proc variedSleep*(
+    counter: var int,
+    durations: openArray[Duration]
+): Future[void] {.async: (raises: [CancelledError], raw: true).} =
+  doAssert(len(durations) > 0, "Empty durations array!")
+  let index =
+    if (counter < 0) or (counter > len(durations)):
```
etan-status:
```suggestion
    if (counter < 0) or (counter > high(durations)):
```
or >=

beacon_chain/el/el_manager.nim
```diff
@@ -1037,14 +1064,17 @@ proc sendNewPayload*(
...
+          # going to sleep until next attempt.
+          await variedSleep(sleepCounter, SleepDurations)
           break mainLoop
 
+proc sendNewPayload*(
+    m: ELManager,
+    blck: SomeForkyBeaconBlock
+): Future[PayloadExecutionStatus] {.
+    async: (raises: [CancelledError], raw: true).} =
+  sendNewPayload(m, blck, DeadlineObject.init(FORKCHOICEUPDATED_TIMEOUT))
```
etan-status:
```suggestion
  sendNewPayload(m, blck, DeadlineObject.init(NEWPAYLOAD_TIMEOUT))
```

beacon_chain/el/el_manager.nim
```diff
@@ -1218,14 +1251,22 @@ proc forkchoiceUpdated*(
...
           break mainLoop
 
+proc forkchoiceUpdated*(
+    m: ELManager,
+    headBlockHash, safeBlockHash, finalizedBlockHash: Eth2Digest,
+    payloadAttributes: Opt[PayloadAttributesV1] |
+                       Opt[PayloadAttributesV2] |
+                       Opt[PayloadAttributesV3]
+): Future[(PayloadExecutionStatus, Opt[BlockHash])] {.
+   async: (raises: [CancelledError], raw: true).} =
```
etan-status:
```suggestion
    async: (raises: [CancelledError], raw: true).} =
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -424,6 +439,14 @@ proc storeBlock(
...
     dag = self.consensusManager.dag
     wallSlot = wallTime.slotOrZero
+    deadlineTime =
+      block:
+        let slotTime = (wallSlot + 1).start_beacon_time() - 1.seconds
+        if slotTime <= wallTime:
+          0.seconds
+        else:
+          chronos.nanoseconds((slotTime - wallTime).nanoseconds)
+    deadlineObj = DeadlineObject.init(deadlineTime)
```
etan-status, line 455: hmm. I think this proc is called also when syncing. If a block gets synced shortly before the end of a slot, this may give it a very short `newPayload` deadline (or even 0).

etan-status (commented):
Overall, the exponential timeouts make more sense than the hardcoded ones from before. They are also short enough to work on both `minimal` and `mainnet` presets, and they should also improve latency to get back online.

https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md

For reference, timeouts defined by spec are 8s on these, but have been removed in Prague for v4 newPayload. @tersec may know more about what that means or if it is intentional.

beacon_chain/gossip_processing/block_processor.nim
cheatfate: There is present an overloaded procedure which still uses `8.seconds` timeout and it will be used in all the places where `DeadlineObject`  is not used as last argument.

beacon_chain/el/el_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6665/commits/b1f2ac91b6af5fa036b01b8727d8c6003991e010

beacon_chain/el/el_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6665/commits/b1f2ac91b6af5fa036b01b8727d8c6003991e010

beacon_chain/el/el_manager.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/6665/commits/b1f2ac91b6af5fa036b01b8727d8c6003991e010

PR #6667, opened 2024-10-21
Implement latest EIP-7685 `requestsHash` method

etan-status:
It's now this weird one-off hashing method so that EL can partially forget in some modules what the data schema is.

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -12,6 +12,7 @@
 import
   # Status libraries
   stew/[byteutils, endians2, objects],
+  nimcrypto/sha2,
   chronicles,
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/helpers.nim
```diff
@@ -466,24 +467,26 @@ func append*(w: var RlpWriter, request: electra.ConsolidationRequest) =
     targetPubkey: Bytes48 request.target_pubkey.blob)
 
 # https://eips.ethereum.org/EIPS/eip-7685
-proc computeRequestsTrieRoot(
+func computeRequestsTrieRoot(
```
arnetheduck: it's not a trieroot any more then

PR #6675, opened 2024-10-23
Add REST endpoint to retrieve historical_summaries

kdeme:
In draft currently as it is not in spec. Needs to be accompanied still with a proposal for beacon REST API specifications.

The idea is that `historical_summaries` is very useful for verifying if beacon blocks (with their accompanied proofs) are part of the canonical chain, however there is no way to actually retrieve the `historical_summaries` except by using the `/eth/v2/debug/beacon/states/{state_id}` endpoint, which is quite the download in terms of size (We are currently using this endpoint in Portal/fluffy  for our `portal_bridge`, but it takes long).

This PR adds an endpoint that provides the `historical_summaries` with its proof against the state root. This is done so that the `historical_summaries` can then be verified for example with the finalized state root that a (light) node might hold from its finality update.
Again, this is specifically useful in Portal (as a `portal_bridge` can inject it in the p2p network as is, and receiving nodes can verify it if they are light client synced) . But you could make an argument that it could also be useful on a consensus light client (assuming blocks with proofs would be somehow available).

edit: moved to `/nimbus/v1/debug/` namespace and removed draft


beacon_chain/rpc/rest_debug_api.nim
```diff
@@ -138,3 +138,52 @@ proc installDebugApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+            let response = RestHistoricalSummaries(
+              historical_summaries: forkyState.data.historical_summaries.asSeq(),
+              proof: proof,
+              slot: bslot.slot,
+            )
+
+            RestApiResponse.jsonResponseFinalized(
+              response, node.getStateOptimistic(state), node.dag.isFinalized(bslot.bid)
+            )
+          elif contentType == sszMediaType:
```
kdeme: I opted for JSON and SSZ, as that is what the debug state endpoint does and also because I was using the SSZ version already in our `portal_bridge`. 

I'm not sure what the convention is, or if there is a convention, of when to opt also for SSZ. I've seen quite some big responses that do not seem to offer SSZ as an option in the beacon API.

beacon_chain/rpc/rest_debug_api.nim
```diff
@@ -138,3 +138,52 @@ proc installDebugApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+            let
+              headers = [("eth-consensus-version", consensusFork.toString())]
```
kdeme: I added this as it seems to be added always (?), even though I can't think of any use for it in the specific case of `historical_summaries`. (Perhaps if the `BeaconState` amount of fields grows > 32 it could be useful for the proof)

beacon_chain/rpc/rest_debug_api.nim
```diff
@@ -138,3 +138,52 @@ proc installDebugApiHandlers*(router: var RestRouter, node: BeaconNode) =
           bestDescendant: item.bestDescendant))
 
     RestApiResponse.jsonResponsePlain(response)
+
+  router.metricsApi2(
+    MethodGet,
+    "/eth/v1/debug/beacon/states/{state_id}/historical_summaries",
```
kdeme: Added it under the `debug` namespace for now.

Although I realize that `debug` is not actually purely a list of debug endpoints as the state one is used by users for trusted node sync.

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
```diff
@@ -73,3 +73,56 @@ proc getStateV2*(client: RestClientRef, state_id: StateIdent,
...
+): RestPlainResponse {.
+  rest,
+  endpoint: "/eth/v1/debug/beacon/states/{state_id}/historical_summaries",
+  accept: preferSSZ,
+  meth: MethodGet
+.}
+
+proc getHistoricalSummariesV1*(
+    client: RestClientRef, state_id: StateIdent, cfg: RuntimeConfig, restAccept = ""
+): Future[Option[GetHistoricalSummariesV1Response]] {.async.} =
```
cheatfate: We are not using simple `{.async.}` anymore, it should use new syntax with `asyncraises`.
By default PRESTO annotates client functions which returns `RestPlainResponse` with
`[CancelledError, RestEncodingError, RestDnsResolveError, RestCommunicationError]` as you can see here
https://github.com/status-im/nim-presto/blob/master/presto/client.nim#L489-L509

Of course you can use `[RestError, CancelledError]` because `RestError` is common ancestor for all exceptions generated by PRESTO, but its not recommended.

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
```diff
@@ -73,3 +73,56 @@ proc getStateV2*(client: RestClientRef, state_id: StateIdent,
...
+    if len(restAccept) > 0:
+      await client.getHistoricalSummariesV1Plain(state_id, restAcceptType = restAccept)
+    else:
+      await client.getHistoricalSummariesV1Plain(state_id)
+
+  return
+    case resp.status
+    of 200:
+      if resp.contentType.isNone() or isWildCard(resp.contentType.get().mediaType):
+        raise newException(RestError, "Missing or incorrect Content-Type")
```
cheatfate: It should be `RestDecodingError` instead

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
```diff
@@ -73,3 +73,56 @@ proc getStateV2*(client: RestClientRef, state_id: StateIdent,
...
+      else:
+        let mediaType = resp.contentType.get().mediaType
+        if mediaType == ApplicationJsonMediaType:
+          let summaries = decodeBytes(
+            GetHistoricalSummariesV1Response, resp.data, resp.contentType
+          ).valueOr:
+            raise newException(RestError, $error)
```
cheatfate: It should be `RestDecodingError`.

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
```diff
@@ -73,3 +73,56 @@ proc getStateV2*(client: RestClientRef, state_id: StateIdent,
...
+          some(summaries)
+        elif mediaType == OctetStreamMediaType:
+          let summaries =
+            try:
+              SSZ.decode(resp.data, GetHistoricalSummariesV1Response)
+            except SerializationError as exc:
+              raise newException(RestError, exc.msg)
```
cheatfate: `RestDecodingError`

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
```diff
@@ -73,3 +73,56 @@ proc getStateV2*(client: RestClientRef, state_id: StateIdent,
...
+          some(summaries)
+        else:
+          raise newException(RestError, "Unsupported Content-Type")
```
cheatfate: `RestDecodingError`

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
cheatfate: In your case it should also include `RestResponseError` which you generating manually.

beacon_chain/rpc/rest_debug_api.nim
cheatfate: `debug` section is part of official specification, you should not add endpoints to this section if its not yet part of official specification, we have `nimbus` section for such calls.

kdeme: Yes, this is why the PR is in draft for now. I want to hopefully get it added in official specification.

I was not aware about our specific `nimbus` prefixed API. I'd be happy to move it there already if that would mean earlier merging in nimbus :)

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
kdeme: Right, I noticed the `{.async.}` still being used in the debug calls and wasn't sure if everything was ready to use `asyncraises`. But looking at the other APIs now I see it has been using `asyncraises` everywhere.
I willl adjust to `asyncraises`.

tersec:
One consideration for me is that if this does get standardized, we don't end up with both a pre-standardized and standardized version which needs to be supported.

kdeme:
> One consideration for me is that if this does get standardized, we don't end up with both a pre-standardized and standardized version which needs to be supported.

One approach could be that I move it under the `/nimbus/v1/debug/` prefix for now, assuming that `/nimbus/v1/debug/` would mean custom and subject to removal without notice (like for the debug cli flags).

beacon_chain/spec/eth2_apis/rest_debug_calls.nim
cheatfate: All the calls which are using `rest` pragma annotation already include `asyncraises` (but its hidden). All other wrappers will be converted soon.

cheatfate:
> > One consideration for me is that if this does get standardized, we don't end up with both a pre-standardized and standardized version which needs to be supported.
> 
> One approach could be that I move it under the `/nimbus/v1/debug/` prefix for now, assuming that `/nimbus/v1/debug/` would mean custom and subject to removal without notice (like for the debug cli flags).

Why you need to remove it? You can establish `redirect()` call from one endpoint to another when it will be part of specification, we was doing this many times.

tersec:
> > > One consideration for me is that if this does get standardized, we don't end up with both a pre-standardized and standardized version which needs to be supported.
> > 
> > 
> > One approach could be that I move it under the `/nimbus/v1/debug/` prefix for now, assuming that `/nimbus/v1/debug/` would mean custom and subject to removal without notice (like for the debug cli flags).
> 
> Why you need to remove it? You can establish `redirect()` call from one endpoint to another when it will be part of specification, we was doing this many times.

Well, part of the point is not to end up with extra, redundant endpoints. The point of this subject to removal without notice would be to exactly allow experimentation without obligation such directions later.

beacon_chain/rpc/rest_debug_api.nim
cheatfate: `preferredContentType` function provides interface to choose what client want to get. Client could use different versions of this construction:

* Accept: application/json
* Accept: application/octet-stream (SSZ)
* Accept: application/json;q=0.5, application/octet-stream;q=0.9
* Accept: application/json;q=0.5, application/octet-stream

and so on...

when using `preferredContentStream` server side portion could which order is preferred for the server. And `preferredContentStream()` establish consensus and return you agreed content type.

cheatfate:
> > > > One consideration for me is that if this does get standardized, we don't end up with both a pre-standardized and standardized version which needs to be supported.
> > > 
> > > 
> > > One approach could be that I move it under the `/nimbus/v1/debug/` prefix for now, assuming that `/nimbus/v1/debug/` would mean custom and subject to removal without notice (like for the debug cli flags).
> > 
> > 
> > Why you need to remove it? You can establish `redirect()` call from one endpoint to another when it will be part of specification, we was doing this many times.
> 
> Well, part of the point is not to end up with extra, redundant endpoints. The point of this subject to removal without notice would be to exactly allow experimentation without obligation such directions later.

There could be exactly 2 situations:
1. Change will be proposed, but declined.
2. Change will be proposed and accepted.

In both this cases it could be better to start with `/nimbus/` endpoint. Because `/nimbus/` endpoint could not be regulated by EF, so we could add/remove any functions. Otherwise in case `1` you will be forced to remove it from `/debug/`.

beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -531,3 +531,52 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+      bslot = node.getBlockSlotId(sid).valueOr:
+        return RestApiResponse.jsonError(Http404, StateNotFoundError, $error)
+      contentType = preferredContentType(jsonMediaType, sszMediaType).valueOr:
+        return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
+
+    node.withStateForBlockSlotId(bslot):
+      return withState(state):
+        when consensusFork >= ConsensusFork.Capella:
+          # Build the proof for historical_summaries field (28th field in BeaconState)
+          let gIndex = GeneralizedIndex(59) # 31 + 28 = 59
```
cheatfate: I think this index is heavily depends on current version of `BeaconState` object, which means that if `BeaconState` object will change in next fork this `index` should be modified or at least visible for new fork maintainers.
So we should add something like
```nim
static: doAssert high(ConsensusFork) == ConsensusFork.Electra
```
here.

etan-status: Yep, and also, `>= Capella` is wrong because Electra has different gindex than Deneb.

kdeme: Right, I see in Electra the amount of fields grows > 32. Basically what I write here (https://github.com/status-im/nimbus-eth2/pull/6675#discussion_r1813609277) is already happening.

I will adjust index for different forks and assert for unsupported/unknown forks.

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -337,6 +337,11 @@ type
   RestEpochRandao* = object
     randao*: Eth2Digest
 
+  RestHistoricalSummaries* = object
+    historical_summaries*: seq[HistoricalSummary]
+    proof*: array[5, Eth2Digest]
```
kdeme: Comment https://github.com/status-im/nimbus-eth2/pull/6675#discussion_r1822537862 means also that this will become `array[6, Eth2Digest]` for Electra. For the JSON part that is not really an issue. In Nim this means I will need to either create different versions of `RestHistoricalSummaries` depending on fork or perhap use `seq` instead.

beacon_chain/rpc/rest_debug_api.nim
kdeme: So this became mandatory due to Electra fork changing the types.

beacon_chain/rpc/rest_nimbus_api.nim
kdeme: added support for forks in https://github.com/status-im/nimbus-eth2/pull/6675/commits/631a155d94a6d60b2f08fde10911bd4a56034fac

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -1076,3 +1076,88 @@ func toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
...
+  # REST API types
+  GetHistoricalSummariesV1Response* = object
+    historical_summaries*: HashList[HistoricalSummary, Limit HISTORICAL_ROOTS_LIMIT]
+    proof*: HistoricalSummariesProof
+    slot*: Slot
+
+  GetHistoricalSummariesV1ResponseElectra* = object
+    historical_summaries*: HashList[HistoricalSummary, Limit HISTORICAL_ROOTS_LIMIT]
+    proof*: HistoricalSummariesProofElectra
+    slot*: Slot
```
kdeme, line 1101: Went with different `array`s for the different proofs as this is for example also what is done for `FinalityBranch`, `CurrentSyncCommitteeBranch` and `NextSyncCommitteeBranch`.

But could in theory also just use a `seq[Eth2Digest]` which I think would make the whole `ForkedHistoricalSummariesWithProof` helpers not required.

kdeme:
@etan-status and/or @cheatfate Would you mind having a look at this again? It is something crucial for our development on Portal network.

As mentioned in the comments, fork management is added, which I think was the only issue (at least for getting it under nimbus/debug namespace)?

Some remarks:
- Those `HistoricalSummariesProof` and `ForkedHistoricalSummariesProof` are added in the rest types code as I did not want to pollute the specs and forks code with types that are not actually part of the consensus specifications. But if you believe they now pollute too much the rest types, I can also move them into another new file.
- The `Forked` type is required as we use arrays for the proofs (which differ in size between forks). We could also opt for going with a sequence instead, but that is not consistent with what is done for example for `FinalityBranch`, `CurrentSyncCommitteeBranch` and `NextSyncCommitteeBranch`.

beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -531,3 +531,69 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+      bslot = node.getBlockSlotId(sid).valueOr:
+        return RestApiResponse.jsonError(Http404, StateNotFoundError, $error)
+      contentType = preferredContentType(jsonMediaType, sszMediaType).valueOr:
+        return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
+
+    node.withStateForBlockSlotId(bslot):
+      return withState(state):
+        when consensusFork >= ConsensusFork.Electra:
+          var proof: HistoricalSummariesProofElectra
+          if forkyState.data.build_proof(HISTORICAL_SUMMARIES_GINDEX_ELECTRA, proof).isErr:
```
etan-status: You could reduce code duplication by extending `forks.nim` with 

```nim
template historical_summaries_gindex*(
    kind: static ConsensusFork): GeneralizedIndex =
  when kind >= ConsensusFork.Electra:
    HISTORICAL_SUMMARIES_GINDEX_ELECTRA
  elif kind >= ConsensusFork.Capella:
    HISTORICAL_SUMMARIES_GINDEX
  else:
    {.error: "historical_summaries_gindex does not support " & $kind.}
```

and `eth2_rest_serialization.nim` with

```nim
template GetHistoricalSummariesResponse(
    kind: static ConsensusFork): auto =
  when kind >= ConsensusFork.Electra:
    GetHistoricalSummariesV1ResponseElectra
  elif kind >= ConsensusFork.Capella:
    GetHistoricalSummariesV1Response
  else:
    {.error: "GetHistoricalSummariesResponse does not support " & $kind.}
```

then can collapse the duplicated logic here:

```nim
        when consensusFork >= ConsensusFork.Capella:
          let response = consensusFork.GetHistoricalSummariesResponse(
            historical_summaries: forkyState.data.historical_summaries,
            proof: forkyState.data.build_proof(
              consensusFork.HISTORICAL_SUMMARIES_GINDEX).expect("Valid gindex"),
            slot: bslot.slot)

          if contentType == jsonMediaType:
            RestApiResponse.jsonResponseFinalizedWVersion(
              response,
              node.getStateOptimistic(state),
              node.dag.isFinalized(bslot.bid),
              consensusFork)
          elif contentType == sszMediaType:
            let headers = [("eth-consensus-version", consensusFork.toString())]
            RestApiResponse.sszResponse(response, headers)
          else:
            RestApiResponse.jsonError(Http500, InvalidAcceptError)

File: beacon_chain/spec/eth2_apis/rest_types.nim [hunk 13]
```diff
@@ -1076,3 +1076,120 @@ func toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
...
+    of ConsensusFork.Bellatrix: discard
+    of ConsensusFork.Capella: capellaData*: GetHistoricalSummariesV1Response
+    of ConsensusFork.Deneb: denebData*: GetHistoricalSummariesV1Response
+    of ConsensusFork.Electra: electraData*: GetHistoricalSummariesV1ResponseElectra
+    of ConsensusFork.Fulu: fuluData*: GetHistoricalSummariesV1ResponseElectra
+
+template init*(
+    T: type ForkedHistoricalSummariesWithProof,
+    historical_summaries: GetHistoricalSummariesV1Response,
+    fork: ConsensusFork,
```
etan-status: Hmm. Do we really need to retain the exact consensusFork?

Or would a `HistoricalSummariesFork` make sense, similar to `LightClientDataFork`, `EpochInfoFork` or (experimental `BlobFork` in #6451)? That way, this code no longer has to be maintained with every fork, by simply having a `capellaData` and `electraData` member, and no `denebData` / `fuluData` etc

```nim
case x.kind
of HistoricalSummariesFork.Electra:
  const historicalFork {.inject, used.} = HistoricalSummariesFork.Electra
  template forkySummaries: untyped {.inject, used.} = x.electraData
  body
of ConsensusFork.Capella:
  const historicalFork {.inject, used.} = HistoricalSummariesFork.Capella
  template forkySummaries: untyped {.inject, used.} = x.capellaData
  body
```

One can still recover exact `consensusFork` if needed from `Eth-Consensus-Version` HTTP header, if needed.

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -1076,3 +1076,120 @@ func toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
...
+    const consensusFork {.inject, used.} = ConsensusFork.Deneb
+    template forkySummaries: untyped {.inject, used.} = x.denebData
+    body
+  of ConsensusFork.Capella:
+    const consensusFork {.inject, used.} = ConsensusFork.Capella
+    template forkySummaries: untyped {.inject, used.} = x.capellaData
+    body
+  of ConsensusFork.Bellatrix:
+    const consensusFork {.inject, used.} = ConsensusFork.Deneb
+    template forkySummaries: untyped {.inject, used.} = GetHistoricalSummariesV1Response()
```
etan-status: Or just don't define `forkySummaries` instead of dummy value. accessing it `when consensusFork < ConsensusFork.Capella` then no longer compiles and makes misuse easier to detect at compiletime.

See `template withForkyStore*(` in `forks_light_client.nim` for comparison.

beacon_chain/spec/eth2_apis/rest_types.nim
etan-status: I think having it consistent with other endpoints is better than `seq`, even though it adds some boilerplate here. There is for every fork only a single correct proof length, and using `array` guarantees that incorrect proof lengths are rejected early at deserialization time.

The real solution would be to adopt EIP-7688... then there are no more random proof length changes for different consensusFork. It's a design flaw that every client has to keep maintaining gindices whenever ethereum decides to release some random functionality unrelated to the client's interests.

beacon_chain/spec/eth2_apis/rest_nimbus_calls.nim
```diff
@@ -76,3 +76,117 @@ proc getTimeOffset*(client: RestClientRef,
...
+    of 400, 500:
+      let error = decodeBytes(RestErrorMessage, resp.data, resp.contentType).valueOr:
+        let msg =
+          "Incorrect response error format (" & $resp.status & ") [" & $error & "]"
+        raise (ref RestResponseError)(msg: msg, status: resp.status)
+      let msg = "Error response (" & $resp.status & ") [" & error.message & "]"
+      raise
+        (ref RestResponseError)(msg: msg, status: error.code, message: error.message)
+    else:
+      raiseRestResponseError(resp)
```
etan-status: editor should be configured to add trailing newline

etan-status (approved):
There's also an effort from @Inspector-Butters and @wemeetagain to look into a general proof API. But I think as part of Nimbus namespace it's alright to have specific endpoints to unblock portal progress.

Code looks correct to me, I have put some remarks to make it less maintenance-heavy and reduce duplication. Once comments are addressed, good to go.

beacon_chain/spec/eth2_apis/rest_types.nim
kdeme: > The real solution would be to adopt EIP-7688

Yes, I agree. But that is not in my sole hands here :)

beacon_chain/spec/eth2_apis/rest_types.nim
kdeme: No, I don't think we need it.

In the past, I mostly found it confusing to work with these different fork types and always had to look back at which one contained which forks, hence why I did not go for that.

But I think I can understand how it makes maintenance probably quite a bit easier when you are dealing with that all the time. Though for an external/new reader of the code, having all these different fork types without a reference in the spec, might be confusing.

beacon_chain/spec/eth2_apis/rest_nimbus_calls.nim
kdeme: editor is actually configured to use nph, but I have to disable this for some repos :)

kdeme: https://github.com/status-im/nimbus-eth2/pull/6675/commits/6cebf610d8ac9047bc8cc44427fb3194581b845f

beacon_chain/spec/eth2_apis/rest_types.nim
kdeme: https://github.com/status-im/nimbus-eth2/pull/6675/commits/6cebf610d8ac9047bc8cc44427fb3194581b845f

beacon_chain/rpc/rest_nimbus_api.nim
kdeme: https://github.com/status-im/nimbus-eth2/pull/6675/commits/ebb2ec464bcfed218cba0be57c1629cae82c401e

beacon_chain/spec/eth2_apis/rest_types.nim
kdeme: So in the case we would add a `HistoricalSummariesFork`, I assume the code would work in such way that in the metrics API where there is now the statement:
```
    node.withStateForBlockSlotId(bslot):
      return withState(state):
        when consensusFork >= ConsensusFork.Capella:
```
we would set the `HistoricalSummariesFork` based on the ConsensusFork?

And then from then onward work with the `HistoricalSummariesFork`?

I can see the benefit of not having to update the fork templates for that, but it also allows for not failing compilation when a fork gets added which might require a change again?

I am currently undecided what I like the most, but if there is a strong pull towards adding `HistoricalSummariesFork` then I don't mind doing so (especially since it is already the habit in the code base).

kdeme:
@etan-status I addressed latest feedback, aside from the specific fork addition, which I commented on (but am open to add still).

beacon_chain/spec/eth2_apis/rest_types.nim
kdeme: @tersec any opinion on this? Else I'll go ahead and merge as is now.

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -1076,3 +1076,135 @@ func toValidatorIndex*(value: RestValidatorIndex): Result[ValidatorIndex,
...
+  of ConsensusFork.Electra:
+    ForkedHistoricalSummariesWithProof(
+      kind: ConsensusFork.Electra, electraData: historical_summaries
+    )
+  of ConsensusFork.Fulu:
+    ForkedHistoricalSummariesWithProof(
+      kind: ConsensusFork.Fulu, fuluData: historical_summaries
+    )
+
+template withForkyHistoricalSummariesWithProof*(
```
tersec, line 1148: Is this used anywhere?

kdeme: It is used in this PR: https://github.com/status-im/nimbus-eth1/pull/2775/files#diff-9f656396622d7567025b612acdc5da46a6a2b795b1346fd3ecf773de786417e4R255

The consumer of this API basically. But it isn't required functionality for the actual server side of the endpoint. 
But as we seem to provide here in general also client side API functionality, I considered it fitting for this repository to be part of.

beacon_chain/spec/eth2_apis/rest_types.nim
etan-status: > but it also allows for not failing compilation when a fork gets added which might require a change again

Typically, adding a new fork is quite a mechanical process, so I guess _if_ this doesn't compile, in practice the first step I expect is that the person adding a new fork would do a simple copy paste of the prior fork's logic, effectively ending up with the same bugs as if the code would implicitly inherit the prior fork's logic rather than explicitly. The mental effort is comparable, one has to be actually aware of generalized indices getting re-indexed during a fork to prevent the mistake.

If you want to protect against the bugs, you can add a `static: doAssert` that checks that `log2trunc(consensusFork.BeaconState.totalSerializedFields)` matches the compatible value, to prevent someone from messing it up with a mechanical change when adding a future fork. Such an assertion cannot be lazily adjusted to make the code work and requires more thought to address :-) That way, you have the maintenance-free implicit inheritance for all the forks that don't need a new `HistoricalSummariesFork`, and you also have the compile-time error in case a new `HistoricalSummariesFork` is required.

Most `ConsensusFork` won't need a new `HistoricalSummariesFork`, I would personally slightly tend towards the `HistoricalSummariesFork`. But I'm also fine with the current solution. Up to @tersec

etan-status: And, `consensusFork.BeaconState` actually works. There's a template in `forks.nim` to support accessing fork specific types based on `static ConsensusFork` constant.

```nim
withConsensusFork(consensusFork):  # or withState / withBlock, they also inject `consensusFork` constant
  static: doAssert log2trunc(consensusFork.BeaconState.totalSerializedFields) == ....
```

tersec (approved):

KolbyML:
Is this still blocked? Because I want to use it!

KolbyML:
Ideally we standardized this endpoint across the board and with loadstar https://github.com/ChainSafe/lodestar/pull/7245

tersec:
The failing CI, https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Flinux%2Fx86_64%2Fnimv2_2/detail/PR-6675/2/pipeline is a bit confused; it's supposed to be Nim 2.2, but it's apparently using Nim 2.0.6:
```

[2025-04-26T13:53:52.899Z] + make LOG_LEVEL=TRACE
[2025-04-26T13:53:53.682Z] Building: build/deposit_contract
[2025-04-26T13:53:53.682Z] Building: build/resttest
[2025-04-26T13:53:53.682Z] Building: build/mev_mock
[2025-04-26T13:53:53.682Z] Building: build/ncli
[2025-04-26T13:53:53.682Z] [using Nim version v2.0.6]
[2025-04-26T13:53:53.682Z] [using Nim version v2.0.6]
[2025-04-26T13:53:53.682Z] [using Nim version v2.0.6]
[2025-04-26T13:53:53.682Z] [using Nim version v2.0.6]
[2025-04-26T13:54:07.291Z] @["Unknown constants in file: ", "MAX_REQUEST_DATA_COLUMN_SIDECARS", "MAX_BLOBS_PER_BLOCK_FULU", "NUMBER_OF_CUSTODY_GROUPS", "CUSTODY_REQUIREMENT", "MAX_REQUEST_PAYLOADS", "WHISK_PROPOSER_SELECTION_GAP", "WHISK_EPOCHS_PER_SHUFFLING_PHASE", "DATA_COLUMN_SIDECAR_SUBNET_COUNT", "SAMPLES_PER_SLOT", "MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS", "TARGET_BLOBS_PER_BLOCK_ELECTRA", "NUMBER_OF_COLUMNS"]
[2025-04-26T13:54:07.291Z] @["Unknown constants in file: ", "MAX_REQUEST_DATA_COLUMN_SIDECARS", "MAX_BLOBS_PER_BLOCK_FULU", "NUMBER_OF_CUSTODY_GROUPS", "CUSTODY_REQUIREMENT", "MAX_REQUEST_PAYLOADS", "WHISK_PROPOSER_SELECTION_GAP", "WHISK_EPOCHS_PER_SHUFFLING_PHASE", "DATA_COLUMN_SIDECAR_SUBNET_COUNT", "SAMPLES_PER_SLOT", "MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS", "TARGET_BLOBS_PER_BLOCK_ELECTRA", "NUMBER_OF_COLUMNS"]
[2025-04-26T13:54:07.291Z] /home/jenkins/workspace/rms_linux_x86_64_nimv2_2_PR-6675/beacon_chain/validator_bucket_sort.nim(50, 19) Error: undeclared identifier: 'newSeqUninit'
[2025-04-26T13:54:07.291Z] candidates (edit distance, scope distance); see '--spellSuggest': 
[2025-04-26T13:54:07.291Z]  (6, 4): 'newSeq'
[2025-04-26T13:54:07.291Z]  (6, 4): 'newSeqOfCap'
[2025-04-26T13:54:07.291Z]  (6, 4): 'newString'
[2025-04-26T13:54:07.291Z] make: *** [Makefile:441: ncli] Error 1
[2025-04-26T13:54:07.291Z] make: *** Waiting for unfinished jobs....
[2025-04-26T13:54:08.074Z] @["Unknown constants in file: ", "MAX_REQUEST_DATA_COLUMN_SIDECARS", "MAX_BLOBS_PER_BLOCK_FULU", "NUMBER_OF_CUSTODY_GROUPS", "CUSTODY_REQUIREMENT", "MAX_REQUEST_PAYLOADS", "WHISK_PROPOSER_SELECTION_GAP", "WHISK_EPOCHS_PER_SHUFFLING_PHASE", "DATA_COLUMN_SIDECAR_SUBNET_COUNT", "SAMPLES_PER_SLOT", "MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS", "TARGET_BLOBS_PER_BLOCK_ELECTRA", "NUMBER_OF_COLUMNS"]
[2025-04-26T13:54:08.074Z] /home/jenkins/workspace/rms_linux_x86_64_nimv2_2_PR-6675/beacon_chain/validator_bucket_sort.nim(50, 19) Error: undeclared identifier: 'newSeqUninit'
[2025-04-26T13:54:08.074Z] candidates (edit distance, scope distance); see '--spellSuggest': 
[2025-04-26T13:54:08.074Z]  (6, 4): 'newSeq'
[2025-04-26T13:54:08.074Z]  (6, 4): 'newSeqOfCap'
[2025-04-26T13:54:08.074Z]  (6, 4): 'newString'
[2025-04-26T13:54:08.074Z] make: *** [Makefile:441: mev_mock] Error 1
[2025-04-26T13:54:11.249Z] /home/jenkins/workspace/rms_linux_x86_64_nimv2_2_PR-6675/beacon_chain/validator_bucket_sort.nim(50, 19) Error: undeclared identifier: 'newSeqUninit'
[2025-04-26T13:54:11.249Z] candidates (edit distance, scope distance); see '--spellSuggest': 
[2025-04-26T13:54:11.249Z]  (6, 4): 'newSeq'
[2025-04-26T13:54:11.249Z]  (6, 4): 'newSeqOfCap'
[2025-04-26T13:54:11.249Z]  (6, 4): 'newString'
[2025-04-26T13:54:11.249Z] make: *** [Makefile:441: deposit_contract] Error 1
[2025-04-26T13:54:20.545Z] Build completed successfully: build/resttest
script returned exit code 2
```
It can be safely ignored.

beacon_chain/spec/eth2_apis/rest_types.nim
kdeme: Introduced `HistoricalSummariesFork` in https://github.com/status-im/nimbus-eth2/pull/6675/commits/c3166c5f041bec72c7145532ed6b5f414505eae8

Assuming this still is the preferred way, if not I can revert the last commit and we go with that.

kdeme:
> @etan-status I addressed latest feedback, aside from the specific fork addition, which I commented on (but am open to add still).

So I added this historical summaries fork in https://github.com/status-im/nimbus-eth2/commit/c3166c5f041bec72c7145532ed6b5f414505eae8 (see msg https://github.com/status-im/nimbus-eth2/pull/6675#discussion_r2063261774). 

@etan-status @tersec Any preference on with or without historical summaries fork implementation?
If no preference, I'll merge as is. If yes, and preferred not, I don't mind reverting last commit and merging that.

beacon_chain/rpc/rest_nimbus_api.nim
```diff
@@ -531,3 +531,46 @@ proc installNimbusApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+      bslot = node.getBlockSlotId(sid).valueOr:
+        return RestApiResponse.jsonError(Http404, StateNotFoundError, $error)
+      contentType = preferredContentType(jsonMediaType, sszMediaType).valueOr:
+        return RestApiResponse.jsonError(Http406, ContentNotAcceptableError)
+
+    node.withStateForBlockSlotId(bslot):
+      return withState(state):
+        when consensusFork >= ConsensusFork.Capella:
+          let summariesFork = historicalSummariesForkAtConsensusFork(consensusFork).valueOr:
+            return RestApiResponse.jsonError(Http404, HistoricalSummariesUnavailable)
```
etan-status: `withState` gives you a `const consensusFork`, so you can make this a `const summariesFork` and force-unwrap with `.expect`, then you don't need the nested `withHistoricalSummariesFork`.

etan-status (approved):
Personally preferring the version that includes the `HistoricalSummariesFork`, seems to require less maintenance than the earlier version.

beacon_chain/rpc/rest_nimbus_api.nim
kdeme: https://github.com/status-im/nimbus-eth2/pull/6675/commits/fbbbdda1a80873b39cc0c58536d7b97a70337454

etan-status: I guess that may also remove the need for the `withHistoricalSummariesFork` template itself. But yeah, looks cleaner. Thanks!

kdeme: Yeah, I forgot about removing that. Done now.

PR #6718, opened 2024-11-21
Adopt latest changes to request hash computation

etan-status:
The `requestType` of empty lists is no longer part of the requests hash.

- https://github.com/ethereum/EIPs/pull/8989

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -462,7 +462,7 @@ func computeRequestsHash(
 
   let requestsHash = computeDigest:
     template mixInRequests(requestType, requestList): untyped =
-      block:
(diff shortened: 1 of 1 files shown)


tersec:
Probably best to merge as part of when https://github.com/ethereum/consensus-specs/releases gets a 1.5.0-alpha.9.

etan-status:
Tested using https://github.com/ethereum/consensus-specs/pull/3998 -- hashes are the same across all consensus-spec-tests except the one that test 0-length transactions (https://github.com/status-im/nim-eth/pull/744#issuecomment-2492124906). the 0-length tx is invalid anyway so it doesn't matter too much whether it gets rejected for invalid block hash or for being 0-length.

PR #6732, opened 2024-11-27
fix testing on arm64

narimiran:


narimiran:
Let's continue in #6918.

PR #6737, opened 2024-11-28
Run LC data collection tests

etan-status:
Add test runner for future tests / ensure it gets compiled.


etan-status:
Passes https://github.com/ethereum/consensus-specs/pull/3553

tersec:
No tests show up on any test summary -- should https://github.com/status-im/nimbus-eth2/blob/unstable/tests/consensus_spec/test_fixture_light_client_data_collection.nim be doing so?

etan-status:
Only once linked PR from consensus-spec gets merged. But yes it's that test runner, and this PR here only adds it to the compilation. It will start producing output after specs are updated to cover the new tests.

PR #6739, opened 2024-11-29
`getMetaData_v3`

agnxsh:
implements https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip7594/p2p-interface.md#metadata

conditionally calls getMetaData v2 or v3 by checking if `FULU_FORK_EPOCH` is scheduled or not.

### DIFF
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -2086,11 +2086,11 @@ proc updateMetadataV2ToV3(metadataRes: NetRes[altair.MetaData]):
                           NetRes[fulu.MetaData] =
   if metadataRes.isOk:
     let metadata = metadataRes.get
-    return ok(fulu.MetaData(seq_number: metadata.seq_number,
+    ok(fulu.MetaData(seq_number: metadata.seq_number,
                             attnets: metadata.attnets,
                             syncnets: metadata.syncnets))
   else:
-    return err(metadataRes.error)
+    err(metadataRes.error)
 
 proc getMetadata_vx(node: Eth2Node, peer: Peer): 
                     Future[NetRes[fulu.MetaData]]
--- a/beacon_chain/networking/eth2_network.nim
+++ b/beacon_chain/networking/eth2_network.nim
@@ -69,7 +69,7 @@ type
     protocols: seq[ProtocolInfo]
       ## Protocols managed by the DSL and mounted on the switch
     protocolStates*: seq[RootRef]
-    metadata*: altair.MetaData
+    metadata*: fulu.MetaData
     connectTimeout*: chronos.Duration
     seenThreshold*: chronos.Duration
     connQueue: AsyncQueue[PeerAddr]
@@ -108,7 +108,7 @@ type
     lastReqTime*: Moment
     connections*: int
     enr*: Opt[enr.Record]
-    metadata*: Opt[altair.MetaData]
+    metadata*: Opt[fulu.MetaData]
     failedMetadataRequests: int
     lastMetadataTime*: Moment
     direction*: PeerType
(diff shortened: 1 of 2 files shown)


beacon_chain/networking/eth2_network.nim
```diff
@@ -2082,12 +2082,32 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
 import ./peer_protocol
 export peer_protocol
 
+proc updateMetadataV2ToV3(metadata: altair.MetaData): fulu.MetaData =
```
tersec: `func`

beacon_chain/networking/eth2_network.nim
```diff
@@ -2082,12 +2082,32 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
...
+    seq_number: metadata.seq_number,
+    attnets: metadata.attnets,
+    syncnets: metadata.syncnets)
+
+proc getMetadata_vx(node: Eth2Node, peer: Peer): 
+                    Future[NetRes[fulu.MetaData]]
+                   {.async: (raises: [CancelledError]).} =
+  let
+    res = 
+      if node.cfg.FULU_FORK_EPOCH != FAR_FUTURE_EPOCH:
```
tersec, line 2100: I guess it's likely/certain enough that PeerDAS or something similar will show up in Fulu that this isn't much of a gamble, so, sure.

beacon_chain/networking/eth2_network.nim
```diff
@@ -2082,12 +2082,32 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
...
+        # Directly fetch fulu metadata if available
+        await getMetadata_v3(peer)
+      else:
+        let metadata_v2_result = await getMetadata_v2(peer)
+        metadata_v2_result.map(proc (altairData: altair.MetaData):
```
tersec: [mapIt](https://nim-lang.org/docs/sequtils.html#mapIt.t%2Ctyped%2Cuntyped) works here more simply and efficiently:
```nim
       metadata_v2_result.mapIt(it.updateMetadataV2ToV3)
```
or
```nim
       metadata_v2_result.mapIt(updateMetadataV2ToV3(it))
```

beacon_chain/networking/eth2_network.nim
```diff
@@ -2082,12 +2082,32 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
 import ./peer_protocol
 export peer_protocol
 
+proc updateMetadataV2ToV3(metadata: altair.MetaData): fulu.MetaData =
+  fulu.MetaData(
+    seq_number: metadata.seq_number,
+    attnets: metadata.attnets,
+    syncnets: metadata.syncnets)
+
+proc getMetadata_vx(node: Eth2Node, peer: Peer):
```
tersec, line 2095: The name emphasizes that it works with any of the `getMetadata` versions, but at this point does that convey useful information, i.e. are there similar/comparable functions which don't have that property?

beacon_chain/networking/eth2_network.nim
agnxsh: ok

beacon_chain/networking/eth2_network.nim
agnxsh: i think vx does convey the fact that it internally handles the metadata versions (whatever version there is, and is NOT deprecated).

beacon_chain/networking/eth2_network.nim
agnxsh: changed to a simpler/cleaner alternative

beacon_chain/networking/eth2_network.nim
```diff
@@ -2082,12 +2082,33 @@ proc p2pProtocolBackendImpl*(p: P2PProtocol): Backend =
 import ./peer_protocol
 export peer_protocol
 
+proc updateMetadataV2ToV3(metadataRes: NetRes[altair.MetaData]): 
+                          NetRes[fulu.MetaData] =
+  if metadataRes.isOk:
+    let metadata = metadataRes.get
+    return ok(fulu.MetaData(seq_number: metadata.seq_number,
```
tersec: Neither branch needs the `return`, can use implicit return style here

beacon_chain/networking/eth2_network.nim
tersec: not relevant anymore, other changes

PR #6741, opened 2024-12-02
add: column support to ENR, Metadata and Request Manager

agnxsh:
This PR adds `custody_subent_count` to both ENR and Metadata as corresponding field, in peerDAS, it helps peers to filer other peers from the peer pool as `full nodes` and `supernodes`.

### DIFF
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -35,16 +35,6 @@ func sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
   res.sort
   res
 
-func sortedColumnIndicesToHashSet*(columnsPerSubnet: ColumnIndex,
-                                   subnetIds: HashSet[uint64]):
-                                   HashSet[ColumnIndex] =
-  var res: HashSet[ColumnIndex] = initHashSet[ColumnIndex]()
-  for i in 0'u64 ..< columnsPerSubnet:
-    for subnetId in subnetIds:
-      let index = DATA_COLUMN_SIDECAR_SUBNET_COUNT * i + subnetId
-      res.incl(ColumnIndex(index))  # Add to HashSet
-  res
-
 func sortedColumnIndexList*(columnsPerSubnet: ColumnIndex,
                             subnetIds: HashSet[uint64]):
                             List[ColumnIndex, NUMBER_OF_COLUMNS] =
@@ -115,7 +105,7 @@ func get_custody_columns_set*(node_id: NodeId,
     columns_per_subnet =
       NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
 
-  sortedColumnIndicesToHashSet(ColumnIndex(columns_per_subnet), subnet_ids)
+  sortedColumnIndices(ColumnIndex(columns_per_subnet), subnet_ids).toHashSet()
 
 func get_custody_column_list*(node_id: NodeId,
                               custody_subnet_count: uint64):
--- a/beacon_chain/spec/eip7594_helpers.nim
+++ b/beacon_chain/spec/eip7594_helpers.nim
@@ -35,6 +35,16 @@ func sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
   res.sort
   res
 
+func sortedColumnIndicesToHashSet*(columnsPerSubnet: ColumnIndex,
+                                   subnetIds: HashSet[uint64]):
+                                   HashSet[ColumnIndex] =
+  var res: HashSet[ColumnIndex] = initHashSet[ColumnIndex]()
+  for i in 0'u64 ..< columnsPerSubnet:
+    for subnetId in subnetIds:
+      let index = DATA_COLUMN_SIDECAR_SUBNET_COUNT * i + subnetId
+      res.incl(ColumnIndex(index))  # Add to HashSet
+  res
+
 func sortedColumnIndexList*(columnsPerSubnet: ColumnIndex,
                             subnetIds: HashSet[uint64]):
                             List[ColumnIndex, NUMBER_OF_COLUMNS] =
@@ -92,6 +102,20 @@ func get_custody_columns*(node_id: NodeId,
       NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
 
   sortedColumnIndices(ColumnIndex(columns_per_subnet), subnet_ids)
+
+func get_custody_columns_set*(node_id: NodeId,
+                              custody_subnet_count: uint64):
+                              HashSet[ColumnIndex] =
+  # This method returns a HashSet of column indices, 
+  # the method is specifically relevant while peer filtering
+  let
+    subnet_ids =
+      get_custody_column_subnets(node_id, custody_subnet_count)
+  const
+    columns_per_subnet =
+      NUMBER_OF_COLUMNS div DATA_COLUMN_SIDECAR_SUBNET_COUNT
+
+  sortedColumnIndicesToHashSet(ColumnIndex(columns_per_subnet), subnet_ids)
 
 func get_custody_column_list*(node_id: NodeId,
                               custody_subnet_count: uint64):
--- a/beacon_chain/sync/request_manager.nim
+++ b/beacon_chain/sync/request_manager.nim
@@ -155,22 +155,27 @@ proc checkResponse(idList: seq[BlobIdentifier],
   true
 
 proc checkResponse(idList: seq[DataColumnIdentifier],
-                   columns: openArray[ref DataColumnSidecar]):
-                   bool =
+                   columns: openArray[ref DataColumnSidecar]): bool =
   if columns.len > idList.len:
     return false
-  for column in columns:
-    let block_root =
-      hash_tree_root(column.signed_block_header.message)
-    var found = false
-    for id in idList:
-      if id.block_root == block_root and id.index == column.index:
-        found = true
-        break
-      if not found:
-        return false
-    column[].verify_data_column_sidecar_inclusion_proof().isOkOr:
+  var i = 0
+  while i < columns.len:
+    let
+      block_root = hash_tree_root(columns[i].signed_block_header.message)
+      id = idList[i]
+
+    # Check if the column reponse is a subset
+    if binarySearch(idList, columns[i], cmpSidecarIdentifier) == -1:
       return false
+
+    # Verify block root and index match
+    if id.block_root != block_root or id.index != columns[i].index:
+      return false
+
+    # Verify inclusion proof
+    columns[i][].verify_data_column_sidecar_inclusion_proof().isOkOr:
+      return false
+    inc i
   true
 
 proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async: (raises: [CancelledError]).} =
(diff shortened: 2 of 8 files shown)


beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +271,95 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+      let remoteCustodySubnetCount =
+        peer.lookupCscFromPeer()
+
+      # Extract remote peer's nodeID from peerID
+      # Fetch custody columns from remote peer
+      let
+        remoteNodeId = fetchNodeIdFromPeerId(peer)
+        remoteCustodyColumns =
+          remoteNodeId.get_custody_columns(max(SAMPLES_PER_SLOT.uint64,
+                                              remoteCustodySubnetCount))
```
tersec: probably should line up with `SAMPLES_PER_SLOT` rather than the `(` if indenting this way; it's another parameter

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +271,95 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+
+      for local_column in localCustodyColumns:
+        if local_column notin remoteCustodyColumns:
```
tersec, line 334: This is an O(n^2) look -- this check is possible easily in O(n log n) way and slightly less easily in an O(n) way. The difference between them is small enough that probably if simple + fast enough in constant overhead, O(n log n) is fine.

This doesn't make much difference unless both peers are supernodes, but  then it gets inefficient.

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +271,95 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+        for col in ucolumns:
+          rman.dataColumnQuarantine[].put(col)
+        var curRoot: Eth2Digest
+        for col in ucolumns:
+          let block_root = hash_tree_root(col.signed_block_header.message)
+          if block_root != curRoot:
+            curRoot = block_root
+            if (let o = rman.quarantine[].popColumnless(curRoot); o.isSome):
+              let col = o.unsafeGet()
+              discard await rman.blockVerifier(col, false)
```
tersec, line 371: It's not important the result?

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +528,130 @@ proc requestManagerBlobLoop(
...
+            let id = DataColumnIdentifier(block_root: columnless.root, index: idx)
+            let local_csc = 
+              if rman.supernode:
+                DATA_COLUMN_SIDECAR_SUBNET_COUNT.uint64
+              else:
+                CUSTODY_REQUIREMENT.uint64
+            let local_custody =
+              rman.network.nodeId.get_custody_columns(max(SAMPLES_PER_SLOT.uint64,
+                                                          local_csc))
+            if id.index in local_custody and id notin fetches and
```
tersec: This is another O(n^2) loop effectively because it's O(n) each time it adds to `fetches`. This could be up to 64 or 128 columns, I guess, which is slightly on the high side to do that with -- one simple approach might be to use a `HashSet` for `fetches` unless the order specifically matters.

It's probably not a practical problem, but if the number of columns grows, code of this sort rapidly becomes a bottleneck which has to be later refactored.

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +271,95 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+            if (let o = rman.quarantine[].popColumnless(curRoot); o.isSome):
+              let col = o.unsafeGet()
+              discard await rman.blockVerifier(col, false)
+      else:
+        debug "Data columns by root request not done, peer doesn't have custody column",
+          peer = peer, columns = shortLog(colIdList), err = columns.error()
+        peer.updateScore(PeerScoreNoValues)
+
+  finally:
+    if not(isNil(peer)):
```
tersec, line 378: It's good to have this check, but is there a situation in which this would be `nil`?

beacon_chain/sync/request_manager.nim
agnxsh: resolved

beacon_chain/sync/request_manager.nim
agnxsh: not ideally, not in any case during bn running

beacon_chain/sync/request_manager.nim
agnxsh: missing parent, or unviable fork, etc, should be handled by request manager block loop, if i can successfully pop a blob/columnless block, it means the block verifier has already handled the cases when it gets an erroneous value earlier/almost at the same time during runtime, atleast that's what i understand from request manager blob loop

beacon_chain/sync/request_manager.nim
agnxsh: resolved, `fetches` as well as `local custody` are now HashSets.

beacon_chain/sync/request_manager.nim
agnxsh: resolved

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6741/commits/27a4d8a0b17cd7a6021e68c919c563924a6a259a

beacon_chain/sync/request_manager.nim
```diff
@@ -31,8 +32,13 @@ const
...
     ## Number of peers we using to resolve our request.
 
+  PARALLEL_REQUESTS_DATA_COLUMNS* = 32
+
   BLOB_GOSSIP_WAIT_TIME_NS* = 2 * 1_000_000_000
-    ## How long to wait for blobs to arrive over gossip before fetching.
+    ## How long to wait for blobs to arri ve over gossip before fetching.
+
+  DATA_COLUMN_GOSSIP_WAIT_TIME_NS* = 2 * 1_000_000_000
+    ## How long to wait for blobs to arri ve over gossip before fetching.
```
tersec, line 42:
```suggestion
    ## How long to wait for blobs to arrive over gossip before fetching.
```

beacon_chain/sync/request_manager.nim
```diff
@@ -31,8 +32,13 @@ const
   PARALLEL_REQUESTS* = 2
     ## Number of peers we using to resolve our request.
 
+  PARALLEL_REQUESTS_DATA_COLUMNS* = 32
+
   BLOB_GOSSIP_WAIT_TIME_NS* = 2 * 1_000_000_000
-    ## How long to wait for blobs to arrive over gossip before fetching.
+    ## How long to wait for blobs to arri ve over gossip before fetching.
```
tersec, line 39:
```suggestion
    ## How long to wait for blobs to arrive over gossip before fetching.
```

beacon_chain/sync/request_manager.nim
```diff
@@ -119,6 +142,25 @@ proc checkResponse(idList: seq[BlobIdentifier],
...
+                   bool =
+  if columns.len > idList.len:
+    return false
+  for column in columns:
+    let block_root =
+      hash_tree_root(column.signed_block_header.message)
+    var found = false
+    for id in idList:
+      if id.block_root == block_root and id.index == column.index:
+        found = true
```
tersec: What is the logic here supposed to do?

As is, `found = true` followed by `break` means that the `if not found` won't run.

But nothing else actually checks/reads `found`, so it's not clear what the purpose of the variable is at all.

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +274,89 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
       self.network.peerPool.release(peer)
 
+proc checkPeerCustody*(rman: RequestManager,
+                       peer: Peer):
+                       bool =
+  # Returns true if the peer custodies atleast
+  # ONE of the common custody columns, straight 
+  # away returns true if the peer is a supernode.
+  if rman.supernode:
+    # For a supernode, it is always best/optimistic
```
tersec, line 305: The logic makes sense, locally/self-interestedly for a given node, but seems like it'd create a cluster of supernodes specifically preferentially peering with each other, which more generally seems potentially troublesome

beacon_chain/sync/request_manager.nim
```diff
@@ -232,6 +274,89 @@ proc fetchBlobsFromNetwork(self: RequestManager,
...
+    
+    else:
+      return false
+
+proc fetchDataColumnsFromNetwork(rman: RequestManager,
+                                 colIdList: seq[DataColumnIdentifier])
+                                 {.async: (raises: [CancelledError]).} =
+  var peer: Peer
+  try:
+    peer = await rman.network.peerPool.acquire()
```
tersec: It seems like `peer` is never re-assigned, and right now, it's notionally briefly declared but default-initialized to some meaningless zero-value. It seems like one could do something like `var peer = await rman ...` outside the `try`, equivalently but avoiding this?

beacon_chain/sync/request_manager.nim
```diff
@@ -400,14 +525,123 @@ proc requestManagerBlobLoop(
...
+  const waitDur = TimeDiff(nanoseconds: DATA_COLUMN_GOSSIP_WAIT_TIME_NS)
+
+  var
+    fetches: HashSet[DataColumnIdentifier]
+    ready: seq[Eth2Digest]
+
+  for columnless in rman.quarantine[].peekColumnless():
+    withBlck(columnless):
+      when consensusFork >= ConsensusFork.Fulu:
+        # granting data columns a chance to arrive over gossip
```
tersec, line 564: As a more general point, protocols/specs practically depending on this kind of logic, especially for interop purposes, probably should be a bit explicit about it. Not an issue with the code though

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -35,6 +35,16 @@ func sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
   res.sort
   res
 
+func sortedColumnIndicesToHashSet*(columnsPerSubnet: ColumnIndex,
```
tersec: What's `sorted` about this? Iterating over the `HashSet[ColumnIndex]` won't happen in any particular order, sorted or otherwise

tersec: Not that it should affect functionality, just that I'm not sure why the name.

beacon_chain/sync/request_manager.nim
agnxsh: resolved

beacon_chain/sync/request_manager.nim
agnxsh: resolved

beacon_chain/sync/request_manager.nim
agnxsh: so in general, it's a pessimistic approach, by default the logic suggests that i am convinced that i have NOT found that right response, if the root and index both are correct then ON found, else it's OFF

beacon_chain/sync/request_manager.nim
agnxsh: as it does not release/kick any peer from the peer pool, thus supernode/fullnode, will _never_ be expelled, hence, everyone who were originally peering will keep peering, this is just preferential peer acquisition for column req/resp

beacon_chain/sync/request_manager.nim
agnxsh: fixed

beacon_chain/spec/eip7594_helpers.nim
agnxsh: got rid of that method, with something simpler

PR #6751, opened 2024-12-05
Support reading `bootstrap_nodes.yaml`

etan-status:
`bootstrap_nodes.txt` is retired in lieu of `bootstrap_nodes.yaml`, start reading `.yaml` format (similar to `.enr`).

### DIFF
--- a/beacon_chain/networking/network_metadata.nim
+++ b/beacon_chain/networking/network_metadata.nim
@@ -18,7 +18,7 @@ import
 
 from std/sequtils import deduplicate, filterIt, mapIt
 from std/strutils import
-  escape, parseBiggestUInt, replace, splitLines, startsWith, strip,
+  endsWith, escape, parseBiggestUInt, replace, splitLines, startsWith, strip,
   toLowerAscii
 
 # TODO(zah):
@@ -116,6 +116,8 @@ proc readBootEnr(path: string): seq[string] {.raises: [IOError].} =
       let line = line.strip()
       if line.startsWith("- enr:"):
         res.add line[2 .. ^1]
+      elif line.startsWith("- \"enr:") and line.endsWith("\""):
+        res.add line[3 .. ^2]  # Gnosis Chiado `boot_enr.yaml`
       elif line.len == 0 or line.startsWith("#"):
         discard
       else:
--- a/beacon_chain/networking/network_metadata.nim
+++ b/beacon_chain/networking/network_metadata.nim
@@ -91,23 +91,39 @@ type
 func hasGenesis*(metadata: Eth2NetworkMetadata): bool =
   metadata.genesis.kind != NoGenesis
 
-proc readBootstrapNodes*(path: string): seq[string] {.raises: [IOError].} =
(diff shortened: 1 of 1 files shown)


beacon_chain/networking/network_metadata.nim
```diff
@@ -126,7 +144,8 @@ proc loadEth2NetworkMetadata*(
       deployBlockPath = path & "/deploy_block.txt"
       depositContractBlockPath = path & "/deposit_contract_block.txt"
       depositContractBlockHashPath = path & "/deposit_contract_block_hash.txt"
-      bootstrapNodesPath = path & "/bootstrap_nodes.txt"
+      bootstrapNodesLegacyPath = path & "/bootstrap_nodes.txt"  # <= Dec 2024
```
tersec, line 147: This is a good way to more safely transition especially given (a) a potentially early December release and (b) possible legacy-format-metadata custom networks people might have. Worth considering at some point removing the older versions. But or this PR, sure.

beacon_chain/networking/network_metadata.nim
```diff
@@ -91,23 +91,41 @@ type
...
+proc readBootstrapNodes(path: string): seq[string] {.raises: [IOError].} =
   # Read a list of ENR values from a YAML file containing a flat list of entries
+  var res: seq[string]
   if fileExists(path):
-    splitLines(readFile(path)).
-      filterIt(it.startsWith("enr:")).
-      mapIt(it.strip())
-  else:
-    @[]
+    for line in splitLines(readFile(path)):
```
tersec, line 98: This isn't really a YAML parser, this is a parser for that specific still-basically-line-oriented format. Is that enshrined as what's necessary, or is it legal now to have true general-purpose YAML files?

We do have `NimYAML` if necessary.

beacon_chain/networking/network_metadata.nim
```diff
@@ -91,23 +91,41 @@ type
...
+proc readBootEnr(path: string): seq[string] {.raises: [IOError].} =
   # Read a list of ENR values from a YAML file containing a flat list of entries
+  var res: seq[string]
   if fileExists(path):
-    splitLines(readFile(path)).
-      filterIt(it.startsWith("- enr:")).
-      mapIt(it[2..^1].strip())
-  else:
-    @[]
+    for line in splitLines(readFile(path)):
```
tersec, line 115: This isn't really a YAML parser, this is a parser for that specific still-basically-line-oriented format. Is that enshrined as what's necessary, or is it legal now to have true general-purpose YAML files?

We do have `NimYAML` if necessary.

etan-status: This is in line with how we parse `config.yaml` for network configurations. We kinda parse what exists in the wild; the `.yaml` extension for both the `boot_enr.yaml` (previously existed, and did not work on Chiado file) and the new `bootstrap_nodes.yaml` (same format seems to work) suggests that this is general purpose YAML in theory.

`NimYAML` so far is solely used in testing, and if we add it to production it should be audited accordingly. For now, I think going with the existing parser (plus the fix for Chiado which quotes strings, which is valid in YAML) is probably alright.

PR #6755, opened 2024-12-10
fix: response checking from RequestManager for blobs

agnxsh:
Previously `checkResponse` for blobs could potentially return some erroneous results for certain values of `idList`. The algo went like:

```nim
proc checkResponse(idList: seq[BlobIdentifier],
                   blobs: openArray[ref BlobSidecar]): bool =
  if len(blobs) > len(idList):
    return false
  for blob in blobs:
    let block_root = hash_tree_root(blob.signed_block_header.message)
    var found = false
    for id in idList:
      if id.block_root == block_root and id.index == blob.index:
        found = true
        break
    if not found:
      return false
    blob[].verify_blob_sidecar_inclusion_proof().isOkOr:
      return false
  true
```

here, initially `found` is `false`, say we check the first elem of `idList`, we see it passes correctly through the `if` statement, we may see `found` as `true` and exit the inner loop, thereby NOT checking the other elements of the `idList`, `else` we straightway return `false`. hence, in any case we DON'T correctly iterate through `idList` and check it.

one simple attack, could be a situation where the seq of `BlobSidecar` has the correct inclusion proof, but any of the non 1st response element(s) is/are missing/incorrect, `checkResponse` regardless should still respond as `true`.

looking at all of this, here's an updated algo, this iterates through the `idList` and the responded seq using a common loop control variable, and efficiently jumps if only a subset of the request has been responded correctly, it performs this check using a binary search.

thereby, the final time complexity of this response checker drops from `O(n^2)` (if `idList` was correctly iterated), to `O(n log n)`. this optimization should be more relevant for data columns later as well, as we would often ask for columns in the order of 50-60 of them, if we were a supernode.

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -10,6 +10,7 @@
 # Uncategorized helper functions from the spec
 
 import
+  std/sequtils,
   # Status libraries
   stew/[byteutils, endians2, objects],
   nimcrypto/sha2,
@@ -250,6 +251,24 @@ func create_blob_sidecars*(
       sidecar.kzg_commitment_inclusion_proof).expect("Valid gindex")
     res.add(sidecar)
   res
+
+func search_sidecar_identifier*(
+    data: seq[BlobIdentifier] | seq[DataColumnIdentifier], 
+    target: ref BlobSidecar | ref DataColumnSidecar): int =
+  var
+    low_pt = 0
+    high_pt = data.high
+
+  while low_pt <= high_pt:
+    let mid = (low_pt + high_pt) div 2
+    if data[mid].index == target.index:
+      return mid  # Target found at index `mid`
+    elif data[mid].index < target.index:
+      low_pt = mid + 1
+    else:
+      high_pt = mid - 1
+
+  -1  # Target not found
 
 # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/altair/light-client/sync-protocol.md#is_sync_committee_update
 template is_sync_committee_update*(update: SomeForkyLightClientUpdate): bool =
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -543,7 +543,6 @@ proc compute_execution_block_hash*(blck: ForkyBeaconBlock): Eth2Digest =
   rlpHash(blockToBlockHeader(blck)).to(Eth2Digest)
 
 from std/math import exp, ln
-from std/sequtils import foldl
 
 func ln_binomial(n, k: int): float64 =
   if k > n:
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/helpers.nim
```diff
@@ -251,6 +252,24 @@ func create_blob_sidecars*(
     res.add(sidecar)
   res
 
+func search_sidecar_identifier*(
```
tersec: Is this custom binary search required?

```nim
import std/algorithm

func compareStringToInt(x: string; y: int): int =
  cmp(x, $y)

var storageKeys = ["1", "2", "3", "4", "5"]
storageKeys.sort()
var searchKey = 3
var index = binarySearch(storageKeys, searchKey, compareStringToInt)

if index != -1:
  echo "Found key at index: ", index
else:
  echo "Key not found"
```
for example -- the built-in `std/algorithm` one can arbitrarily map search keys to underlying storage types for comparison purposes

agnxsh: yeah because the key and the search space are 2 different data types.

tersec: sure, but that's what's happening here, no?

tersec: https://github.com/status-im/nimbus-eth2/pull/6755/commits/98bc8b9f653f0ca484d22b792a85d36e8313b8ff

beacon_chain/spec/helpers.nim
```diff
@@ -10,6 +10,7 @@
 # Uncategorized helper functions from the spec
 
 import
+  std/sequtils,
```
tersec, line 13: is this import still necessary now that the rest was removed?

beacon_chain/sync/request_manager.nim
```diff
@@ -102,21 +103,33 @@ proc checkResponse(roots: openArray[Eth2Digest],
...
-    if not found:
+  var i = 0
+  while i < blobs.len:
+    let
+      block_root = hash_tree_root(blobs[i].signed_block_header.message)
+      id = idList[i]
+
+    # Check if the blob response is a subset
+    if binarySearch(idList, blobs[i], cmpSidecarIdentifier) == -1:
+      i = if i != blobs.len - 1: i + 2 else: i + 1
```
tersec: There are basically three cases here, and all of them trigger next a `while i < blobs.len` check:
- `i < blobs.len - 2`: `i = i + 2`, loop continues
- `i == blobs.len - 2`: this does `i = i + 2`, which triggers a loop halt
- `i == blobs.len - 1`: this is the case where `i = i + 1` happens, but either `i = i + 1` or `i = i + 2` would trigger the exact same loop halt because either it'd make `i == blobs.len` or `i == blobs.len + 1`

I don't understand the `if` here.

PR #6758, opened 2024-12-13
small poc library

pedromiguelmiranda:
tests, not meant to be commited, ignore.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -63,7 +63,7 @@ declareGauge sync_committee_active,
 declareCounter db_checkpoint_seconds,
   "Time spent checkpointing the database to clear the WAL file"
 
-proc fetchGenesisState(
+proc fetchGenesisState*(
     metadata: Eth2NetworkMetadata,
     genesisState = none(InputFile),
     genesisStateUrl = none(Uri)
@@ -101,7 +101,7 @@ proc fetchGenesisState(
   else:
     nil
 
-proc doRunTrustedNodeSync(
+proc doRunTrustedNodeSync*(
     db: BeaconChainDB,
     metadata: Eth2NetworkMetadata,
     databaseDir: string,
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1869,7 +1869,7 @@ proc onSlotStart(node: BeaconNode, wallTime: BeaconTime,
 
 proc onSecond(node: BeaconNode, time: Moment) =
   # Nim GC metrics (for the main thread)
-  updateThreadMetrics()
+  # updateThreadMetrics()
```
pedromiguelmiranda, line 1894: Metrics collection not generated, and this code is commented. possible solutions may required a significant rewrite (using some strategy simliar to an aggregator pattern)

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2364,7 +2364,7 @@ proc doRunBeaconNode(config: var BeaconNodeConf, rng: ref HmacDrbgContext) {.rai
 
   # Nim GC metrics (for the main thread) will be collected in onSecond(), but
   # we disable piggy-backing on other metrics here.
-  setSystemMetricsAutomaticUpdate(false)
+  # setSystemMetricsAutomaticUpdate(false)
```
pedromiguelmiranda, line 2389: same

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2548,7 +2548,7 @@ proc handleStartUpCmd(config: var BeaconNodeConf) {.raises: [CatchableError].} =
 
 {.pop.} # TODO moduletests exceptions
 
-programMain:
+when isMainModule:
```
pedromiguelmiranda, line 2573: This might be an outdated strategy.however it's present in other "main" modules as well.
What is the impact of using the "when is" approach vs this one?

pedromiguelmiranda:
not needed anymore

PR #6764, opened 2024-12-16
adds peerdas-supernode flag

agnxsh:
In order to measure resources efficiently it's beneficial to have a separate peerdas supernode flag, this is because if we launch a peerdas supernode on `subscribe-all-subnets`, then we also end up subscribing to all attnets and syncnets, which would thereby use of more resources to run the peerdas supernode.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -254,9 +254,10 @@ type
       name: "subscribe-all-subnets" .}: bool
 
     peerdasSupernode* {.
+      hidden
       defaultValue: false,
       desc: "Subscribe to all column subnets, thereby becoming a peerdas supernode"
-      name: "peerdas-supernode" .}: bool
+      name: "debug-peerdas-supernode" .}: bool
 
     slashingDbKind* {.
       hidden
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -253,6 +253,11 @@ type
       desc: "Subscribe to all subnet topics when gossiping"
       name: "subscribe-all-subnets" .}: bool
 
+    peerdasSupernode* {.
+      defaultValue: false,
+      desc: "Subscribe to all column subnets, thereby becoming a peerdas supernode"
+      name: "peerdas-supernode" .}: bool
+
     slashingDbKind* {.
       hidden
       defaultValue: SlashingDbKind.v2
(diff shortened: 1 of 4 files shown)


beacon_chain/conf.nim
```diff
@@ -253,6 +253,11 @@ type
       desc: "Subscribe to all subnet topics when gossiping"
       name: "subscribe-all-subnets" .}: bool
 
+    peerdasSupernode* {.
+      defaultValue: false,
+      desc: "Subscribe to all column subnets, thereby becoming a peerdas supernode"
+      name: "peerdas-supernode" .}: bool
```
tersec: `debug-peerdas-supernode` would be nicer.

https://nimbus.guide/options.html
> Any `debug`-prefixed flags are considered ephemeral and subject to removal without notice.

Until `Fulu` becomes production, this is a debugging flag for devnet/testnet/Kurtosis purposes, so shouldn't be documented or supported-per-se.

Also, probably, `hidden`, because again, it's not really part of the supported UI/UX yet, certainly before Fulu is in public testnets (months from now).

It does seem possible it will at some point, but no need to do that prematurely.

agnxsh: ok, resolving

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -24,7 +24,7 @@ type
   CellBytes = array[fulu.CELLS_PER_EXT_BLOB, Cell]
   ProofBytes = array[fulu.CELLS_PER_EXT_BLOB, KzgProof]
 
-func sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
+func sortedColumnIndices(columnsPerSubnet: ColumnIndex,
                           subnetIds: HashSet[uint64]):
```
tersec: the indentation of these lines is off now

beacon_chain/spec/eip7594_helpers.nim
```diff
@@ -35,7 +35,7 @@ func sortedColumnIndices*(columnsPerSubnet: ColumnIndex,
   res.sort
   res
 
-func sortedColumnIndexList*(columnsPerSubnet: ColumnIndex,
+func sortedColumnIndexList(columnsPerSubnet: ColumnIndex,
                             subnetIds: HashSet[uint64]):
```
tersec: indentation

beacon_chain/spec/eip7594_helpers.nim
agnxsh: resolved

beacon_chain/spec/eip7594_helpers.nim
agnxsh: resolved

PR #6774, opened 2024-12-18
multiple typos

crStiv:


tersec:
https://en.wikipedia.org/wiki/Artefact

PR #6776, opened 2024-12-19
fix: multiple typos

crStiv:


tersec:
Thank you, these have been fixed in `unstable`.

PR #6777, opened 2024-12-19
custody subnet count decoding during discovery

agnxsh:


beacon_chain/spec/network.nim
```diff
@@ -235,3 +235,12 @@ func getSyncSubnets*(
 iterator blobSidecarTopics*(forkDigest: ForkDigest): string =
   for subnet_id in BlobId:
     yield getBlobSidecarTopic(forkDigest, subnet_id)
+
+func getDataColumnSidecarTopic*(forkDigest: ForkDigest,
```
tersec, line 240: Useful to link to https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/p2p-interface.md#data_column_sidecar_subnet_id here which seems to be where the `data_column_sidecar_` prefix is defined in the specs.

agnxsh: linked

PR #6783, opened 2024-12-22
implement SingleAttestation

tersec:
Required for https://notes.ethereum.org/@ethpandaops/pectra-devnet-5

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -1061,6 +1061,16 @@ proc validateAttestation*(
   # defined by attestation.data.beacon_block_root -- i.e.
   # get_checkpoint_block(store, attestation.data.beacon_block_root,
   # store.finalized_checkpoint.epoch) == store.finalized_checkpoint.root
+
+  var sig: CookedSig
+  let shufflingRefMissing =
+    pool.dag.findShufflingRef(target.blck.bid, target.slot.epoch).isNone
+  if shufflingRefMissing:
+    # getShufflingRef might be slow here, so first try to eliminate by
+    # signature check
+    sig = attestation.signature.load().valueOr:
+      return pool.checkedReject("SingleAttestation: unable to load signature")
+
   let shufflingRef =
     pool.dag.getShufflingRef(target.blck, target.slot.epoch, false).valueOr:
       # Target is verified - shouldn't happen
@@ -1080,7 +1090,6 @@ proc validateAttestation*(
       shufflingRef, attestation.data.slot,
       attestation.committee_index.CommitteeIndex)
     index_in_committee = find(beacon_committee, validator_index)
-  doAssert index_in_committee < beacon_committee.len
   if index_in_committee < 0:
     return pool.checkedReject("SingleAttestation: attester index not in beacon committee")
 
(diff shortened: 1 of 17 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1023,31 +1031,57 @@ proc validateAttestation*(
...
   # In the spec, is_valid_indexed_attestation is used to verify the signature -
   # here, we do a batch verification instead
-  let sig =
-    attestation.signature.load().valueOr:
-      return pool.checkedReject("Attestation: unable to load signature")
+  let sig = attestation.signature.load().valueOr:
+    return pool.checkedReject("SingleAttestation: unable to load signature")
 
-  return ok((validator_index, sig))
+  echo "FOOBAR4: ", beacon_committee.len, "; ", index_in_committee
```
arnetheduck: ..

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1023,31 +1031,57 @@ proc validateAttestation*(
...
+  # attestation.attester_index in
+  # get_beacon_committee(state, attestation.data.slot, index).
   let
-    shufflingRef =
-      pool.dag.getShufflingRef(target.blck, target.slot.epoch, false).valueOr:
-        # Target is verified - shouldn't happen
-        warn "No shuffling for attestation - report bug",
-          attestation = shortLog(attestation), target = shortLog(target)
-        return errIgnore("Attestation: no shuffling")
+    beacon_committee = get_beacon_committee(
```
arnetheduck, line 1106: here, ideally we'd have a fast and a slow path - the fast path is for when the shuffling is cached - then we check beacon committee and other "fast" checks and finally sig.. 

if instead the committee is unknown, we first check the signature and only then do we compute the shuffling to check the other fields

tersec: https://github.com/status-im/nimbus-eth2/pull/6783/commits/659d03d7f291513e6049ca7839dd9fb420d116ab

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6783/commits/659d03d7f291513e6049ca7839dd9fb420d116ab

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1023,31 +1031,68 @@ proc validateAttestation*(
   # defined by attestation.data.beacon_block_root -- i.e.
   # get_checkpoint_block(store, attestation.data.beacon_block_root,
   # store.finalized_checkpoint.epoch) == store.finalized_checkpoint.root
+
+  var sig: CookedSig
+  let shufflingRefMissing =
```
arnetheduck: why not:

```nim
var sigchecked = false
let shufflingRef = findShufflingRef(...).valueOr:
  # check sig
  sigchecked = true
  getShufflingRefRef(...)
```

avoids the double lookup in the majority (happy) case

arnetheduck (approved):

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6783/commits/836b28f6429226019723b5b4e7775124c1924cca

PR #6807, opened 2025-01-02
fix loading Nix Flake shell via direnv, update readme

jakubgs:
And some other Nix related fixes.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,16 +1,6 @@
 NBS_ONLY_LOAD_ENV_VARS=1 source env.sh
 
-if command -v nix > /dev/null
-then
-  export NIMBUS_NIX_ENV=1
-
-  cd installer/nix
-  # watch_file tells direnv that changes to any of the watched files
-  # should trigger a re-evalution of the environment
-  watch_file flake.nix
-  watch_file flake.lock
-  watch_file shell.nix
-
-  mkdir -p .flake-profiles
-  eval "$(nix print-dev-env --profile ".flake-profiles/profile")"
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.6; then
+  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.6/direnvrc" "sha256-RYcUJaRMf8oF5LznDrlCXbkOQrywm0HDv1VjYGaJGdM="
 fi
+use flake
--- a/nix/README.md
+++ b/nix/README.md
@@ -19,11 +19,12 @@ https://github.com/NixOS/nix/issues/4423
 
 It can be also done without even cloning the repo:
 ```sh
-nix build 'github:status-im/nimbus-eth2?submodules=1'
+nix build 'github:status-im/nimbus-eth2?submodules=1#'
 ```
+The trailing `#` is required due to [URI parsing bug in Nix](https://github.com/NixOS/nix/issues/6633).
 
 ## Running
 
 ```sh
-nix run 'github:status-im/nimbus-eth2?submodules=1'
+nix run 'github:status-im/nimbus-eth2?submodules=1#'
 ```
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,7 +254,7 @@ jobs:
         if: ${{ !cancelled() }} && github.event_name == 'pull_request'
         run: |
           excluded_files="config.yaml"
-          excluded_extensions="ans|bin|cfg|json|json\\.template|md|png|service|ssz|txt|lock|nix"
+          excluded_extensions="ans|bin|cfg|yml|json|json\\.template|md|png|service|ssz|txt|lock|nix|gitignore|envrc"
 
           current_year=$(date +"%Y")
           problematic_files=()


jakubgs:
There is an issue with Nimble build and missing `sat` sources:
```
bin/nim c -o:bin/nimble -d:release --mm:refc --noNimblePath -d:release --skipUserCfg --skipParentCfg --warnings:off --hints:off dist/nimble/src/nimble.nim
/build/source/vendor/nimbus-build-system/vendor/Nim/dist/nimble/src/nimble.nim(14, 13) Error: cannot open file: sat/sat
```
This is caused by this import:
```nim
when defined(nimNimbleBootstrap):
  import ../dist/sat/src/sat/sat
else:
  import sat/sat
```
https://github.com/nim-lang/nimble/blob/46e2ae13eeb95619a371a74d16efc5aff30ea371/src/nimble.nim#L11-L14

Which hits the `else` because when I try to set `-d:nimNimbleBootstrap=true` in `NIMFLAGS` it doesn't seem to propagate to the build of Nimbus when using the `nimbus-build-system`. I'm not sure where `sat` sources are supposed to be for `import sat/sat` to work.

jakubgs:
I have checked and indeed the hash issues appear to be caused by some Nix caching mechanism which messes with results.

Indeed use of `fetchSubmodules = true` here should be sufficient to provide `sat` sources:
https://github.com/status-im/nimbus-eth2/blob/bfec79eac3253deb71543dd55c35eb61bdf7dd49/nix/nimble.nix#L9

I need to research how to avoid these Nix caching confusion.

tersec:
> There is an issue with Nimble build and missing `sat` sources:
> 
> ```
> bin/nim c -o:bin/nimble -d:release --mm:refc --noNimblePath -d:release --skipUserCfg --skipParentCfg --warnings:off --hints:off dist/nimble/src/nimble.nim
> /build/source/vendor/nimbus-build-system/vendor/Nim/dist/nimble/src/nimble.nim(14, 13) Error: cannot open file: sat/sat
> ```
> 
> This is caused by this import:
> 
> ```nim
> when defined(nimNimbleBootstrap):
>   import ../dist/sat/src/sat/sat
> else:
>   import sat/sat
> ```
> 
> https://github.com/nim-lang/nimble/blob/46e2ae13eeb95619a371a74d16efc5aff30ea371/src/nimble.nim#L11-L14
> 
> Which hits the `else` because when I try to set `-d:nimNimbleBootstrap=true` in `NIMFLAGS` it doesn't seem to propagate to the build of Nimbus when using the `nimbus-build-system`. I'm not sure where `sat` sources are supposed to be for `import sat/sat` to work.

If you see the line 14 error, it's the wrong/old Nimble version.  [Nimble 0.16.2](https://github.com/nim-lang/nimble/blob/v0.16.2/src/nimble.nim) attempts to import `sat` on either line 12 or 14; This code targets [Nimble 0.16.4](https://github.com/nim-lang/nimble/blob/v0.16.4/src/nimble.nim), which attempts to import `sat` on line 11.

jakubgs:
Indeed, this is the hash caching behavior that we talked about. Did you review the PR?

nix/README.md
```diff
@@ -19,11 +19,12 @@ https://github.com/NixOS/nix/issues/4423
 
 It can be also done without even cloning the repo:
 ```sh
-nix build 'github:status-im/nimbus-eth2?submodules=1'
+nix build 'github:status-im/nimbus-eth2?submodules=1#'
 ```
+The trailing `#` is required due to URL parsing bug in Nix.
```
tersec: Is this bug documented upstream somewhere to link to so one can check whether it's fixed yet, in the future?

.envrc
```diff
@@ -1,16 +1,6 @@
...
-  cd installer/nix
-  # watch_file tells direnv that changes to any of the watched files
-  # should trigger a re-evalution of the environment
-  watch_file flake.nix
-  watch_file flake.lock
-  watch_file shell.nix
-
-  mkdir -p .flake-profiles
-  eval "$(nix print-dev-env --profile ".flake-profiles/profile")"
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.6; then
```
tersec, line 3: Is there something specific/special about version 3.0.6 or is this a way of pinning a semi-arbitrary version?

tersec:
> Indeed, this is the hash caching behavior that we talked about. Did you review the PR?

Mostly fine/reasonable, a couple questions

nix/README.md
jakubgs: Not that I know of.

.envrc
jakubgs: It's the same version we've been using in all of our `infra-*` repositories for our Nix Flake shell for a few months now:
https://github.com/status-im/infra-template/blob/3c85b88671a8b48c5d267801bf0ad678f0e739d4/.envrc#L1-L3

So it's known to be stable.

nix/README.md
jakubgs: This appears to be the issue:

- https://github.com/NixOS/nix/issues/6633#issuecomment-1926198701

Updated the readme.

tersec (approved):

PR #6812, opened 2025-01-03
Fix EL manager to wait for non-syncing/accepted response.

cheatfate:


tersec:
```
Run excluded_files="config.yaml"
- beacon_chain/el/el_manager.nim
```

PR #6820, opened 2025-01-06
clean up eip7594 naming, useful links, test suite naming

agnxsh:


tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2024 Status Research & Development GmbH
+# Copyright (c) 2025 Status Research & Development GmbH
```
tersec: Not 2024-2025?

tests/consensus_spec/test_fixture_networking.nim
```diff
@@ -43,7 +43,7 @@ proc runGetCustodyColumns(suiteName, path: string) =
     for i in 0..<columns.lenu64:
       check columns[i] == reslt[i]
 
-suite "EF - EIP7594 - Networking" & preset():
+suite "EF - PEERDAS - Networking" & preset():
```
tersec: PEER is not in this context an acronym, nor is it capitalized as such

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -329,7 +329,7 @@ suite suiteName:
     for kind, path in walkDir(testsDir, relative = true, checkDir = true):
       runComputeBlobKzgProofTest(suiteName, testsDir, testsDir / path)
 
-suiteName = "EF - KZG - EIP7594"
+suiteName = "EF - KZG - PEERDAS"
```
tersec: PeerDAS, not PEERDAS

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2024 Status Research & Development GmbH
+# Copyright (c) 2025 Status Research & Development GmbH
```
tersec: 2024-2025

agnxsh: ok

tests/consensus_spec/test_fixture_kzg.nim
agnxsh: ok

tests/consensus_spec/test_fixture_networking.nim
agnxsh: ok

tests/consensus_spec/test_fixture_networking.nim
agnxsh: fair enough

beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -117,14 +117,14 @@ type
...
-  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.5/specs/_features/eip7594/p2p-interface.md#metadata
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/p2p-interface.md#metadata
   MetaData* = object
     seq_number*: uint64
     attnets*: AttnetBits
     syncnets*: SyncnetBits
     custody_group_count*: uint64
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/deneb/beacon-chain.md#executionpayload
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/beacon-chain.md#execution-payload
```
tersec: This isn't the best URL here -- https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/beacon-chain.md#execution-payload documents the Fulu form of the function `process_execution_payload()`, whereas the URL here should document the actual type `ExecutionPayload`.

As far a I can tell, https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/beacon-chain.md doesn't change the `ExecutionPayload` at all, nor does https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/electra/beacon-chain.md

So while updating the URL just for spec version purposes (i.e. https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/deneb/beacon-chain.md#executionpayload) makes sense, it's still the same Deneb container.

The reason Nimbus redefines this for Electra and Fulu is mainly to have a 1:1 type mapping between the EL/engine API `ExecutionPayload` types and the CL-side ones, to allow type-based disambiguation/dispatching for converting back and forth between them. But they're the same types in each case.

agnxsh: fixed

PR #6835, opened 2025-01-10
assert submodules are available to Nix derivation

jakubgs:
This should result in a human-readable error when `?submodules=1#` is not appended to the project URL as the README explains:
```
 > nix build '.'
       error: Unable to build without submodules. Append '?submodules=1#' to the URL.

 > nix build '.?submodules=1#'
```

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -16,6 +16,10 @@
   ],
 }:
 
+# The 'or' is to handle src fallback to ../. which lack submodules attribue.
+assert pkgs.lib.assertMsg ((src.submodules or true) == true)
+  "Unable to build without submodules. Append '?submodules=1#' to the URI.";
+
 let
   inherit (pkgs) stdenv lib writeScriptBin callPackage;
 


tersec:
Why would this error condition happen to begin with?

jakubgs:
Because if we do not specify `?submodules=1#` in the URL when building the node via Nix then submodules will not be present and the build will fail. In a perfect would this would be configurable in the `flake.nix` itself, but currently it is not. For more context on problems with submodules you can read this issue:

- https://github.com/NixOS/nix/issues/6633

I've added this assert because without it the error is incomprehensible to a mere mortal.

tersec:
> Because if we do not specify `?submodules=1#` in the URL when building the node via Nix then submodules will not be present and the build will fail. In a perfect would this would be configurable in the `flake.nix` itself, but currently it is not. For more context on problems with submodules you can read this issue:
> 
>     * [Submodules of flakes are not working NixOS/nix#6633](https://github.com/NixOS/nix/issues/6633)
> 
> 
> I've added this assert because without it the error is incomprehensible to a mere mortal.

Ok, but why would it be intermittent, in that case?

jakubgs:
@tersec it's not intermittent, it's just that Nix behavior with submodules is a bit inconsistent which can cause confusion:

- https://github.com/NixOS/nix/issues/6633#issuecomment-2580528070

Having this clear message avoids the confusion.

PR #6849, opened 2025-01-15
Adjust default port numbers for tests.

cheatfate:


Makefile
```diff
@@ -210,10 +210,10 @@ libbacktrace:
...
 # - --base-el-ws-port + --el-port-offset * [0, --nodes + --light-clients)
 # - --base-el-auth-rpc-port + --el-port-offset * [0, --nodes + --light-clients)
-UNIT_TEST_BASE_PORT := 9960
-REST_TEST_BASE_PORT := 9990
-MINIMAL_TESTNET_BASE_PORT := 5001
-MAINNET_TESTNET_BASE_PORT := 6501
+UNIT_TEST_BASE_PORT := 39960
+REST_TEST_BASE_PORT := 40990
+MINIMAL_TESTNET_BASE_PORT := 35001
+MAINNET_TESTNET_BASE_PORT := 36501
```
etan-status, line 216: ephemeral ports could be in use randomly, https://en.wikipedia.org/wiki/Ephemeral_port#Range - unfortunately it's quite platform specific which OS uses what ports.

richard-ramos: perhaps something like this would work? (after replacing `nim` by  `$(ENV_SCRIPT) $(NIMC)`. That way the port numbers wouldn't have to be hardcoded, and the port number would be obtained by nim automatically regardless of the OS being used.

```make
UNIT_TEST_BASE_PORT := $(shell nim r -d:release --eval:"import net; let s=newSocket(); s.bindAddr(Port(0)); echo s.getLocalAddr()[1]; s.close()")
REST_TEST_BASE_PORT := $(shell nim r -d:release --eval:"import net; let s=newSocket(); s.bindAddr(Port(0)); echo s.getLocalAddr()[1]; s.close()")
MINIMAL_TESTNET_BASE_PORT := $(shell nim r -d:release --eval:"import net; let s=newSocket(); s.bindAddr(Port(0)); echo s.getLocalAddr()[1]; s.close()")
MAINNET_TESTNET_BASE_PORT := $(shell nim r -d:release --eval:"import net; let s=newSocket(); s.bindAddr(Port(0)); echo s.getLocalAddr()[1]; s.close()")

main:
	@echo "Port1: $(UNIT_TEST_BASE_PORT)"
	@echo "Port2: $(REST_TEST_BASE_PORT)"
	@echo "Port3: $(MINIMAL_TESTNET_BASE_PORT)"
	@echo "Port4: $(MAINNET_TESTNET_BASE_PORT)"
```

cheatfate: lol, but it will not work because there is a big gap in time between this variables being used and your incredible script being executed.
One more thing, your script does not work when executed many times, it will always provide you with same port number.

etan-status: And also, there are entire ranges of port being used that add on top of the base port xD With the script you could even get the same port multiple times.

richard-ramos: Oh well. RIP my script idea
image

PR #6872, opened 2025-01-25
prepare builder API, modify electra BuilderBid

agnxsh:


tersec:
```
- beacon_chain/spec/mev/electra_mev.nim
- beacon_chain/spec/mev/fulu_mev.nim
```

tersec:
The comment at
https://github.com/status-im/nimbus-eth2/blob/4d89d9531dda1562463a41aa5414375b419648d7/beacon_chain/spec/state_transition.nim#L538-L538
is addressed by this PR and should be removed.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -707,11 +708,27 @@ proc getBlindedExecutionPayload[
...
-      executionPayloadValue: builderBid.value))
+    when EPH is deneb_mev.BlindedExecutionPayloadAndBlobsBundle:
+      return ok(BuilderBid[EPH](
+        blindedBlckPart: EPH(
+          execution_payload_header: builderBid.header,
+          blob_kzg_commitments: builderBid.blob_kzg_commitments),
+        executionRequests: Opt.none(ExecutionRequests),
+        executionPayloadValue: builderBid.value))
+    elif EPH is electra_mev.BlindedExecutionPayloadAndBlobsBundle:
+      return ok(BuilderBid[EPH](
```
tersec, line 720: The `electra_mev.BlindedExecutionPayloadAndBlobsBundle` and `fulu_mev.BlindedExecutionPayloadAndBlobsBundle` cases are identical and can be combined.

tersec: Also, in general whenever `nimbus-eth2` has these
```nim
  when cond1:
    ...
  elif cond2:
    ...
  elif cond3:
    ...
```
conditions, it adds on an
```nim
  else:
    static: doAssert false
```

or
```nim
  else:
    static: raiseAssert "..."
```

e.g.
https://github.com/status-im/nimbus-eth2/blob/4e440277cf8a3fed72f32eb2f01fc5e910ad6768/beacon_chain/spec/state_transition_epoch.nim#L1016-L1037

That it's `static: ...` is important; the exact way it causes the compilation to fail is less important. This way, if/when the conditions fail to hold -- e.g., here, a new fork is added after Fulu -- it won't silently stop working, but rather loudly fail to compile.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1072,11 +1094,26 @@ proc getBuilderBid[
...
-  ))
+  template execution_requests: untyped =
+      unsignedBlindedBlock.get.message.body.execution_requests
+  when SBBB is deneb_mev.SignedBlindedBeaconBlock:
+   return ok(BuilderBid[SBBB](
+      blindedBlckPart: unsignedBlindedBlock.get,
+      executionRequests: Opt.none(ExecutionRequests),
+      executionPayloadValue: bidValue,
+      consensusBlockValue: consensusValue))
+  elif SBBB is electra_mev.SignedBlindedBeaconBlock:
```
tersec: Same issues as https://github.com/status-im/nimbus-eth2/pull/6872/files#r1929864463

beacon_chain/validators/beacon_validators.nim
```diff
@@ -983,11 +1001,13 @@ proc getBlindedBlockParts[
     debugComment "verify (again, after change) this is what builder API needs"
```
tersec: this `debugComment` is addressed by this PR, can be removed

agnxsh: resolved in https://github.com/status-im/nimbus-eth2/pull/6872/commits/90ece63496ba173c68690c2c3c50caa563b98fcc

beacon_chain/validators/beacon_validators.nim
agnxsh: resolved in https://github.com/status-im/nimbus-eth2/pull/6872/commits/90ece63496ba173c68690c2c3c50caa563b98fcc

beacon_chain/validators/beacon_validators.nim
agnxsh: resolved in https://github.com/status-im/nimbus-eth2/pull/6872/commits/90ece63496ba173c68690c2c3c50caa563b98fcc

beacon_chain/validators/beacon_validators.nim
```diff
@@ -968,26 +981,27 @@ proc getBlindedBlockParts[
...
     type PayloadType = deneb.ExecutionPayloadForSigning
     template actualEPH: untyped =
-      executionPayloadHeader.get.blindedBlckPart.execution_payload_header
+      blindedBlockRes.get.blindedBlckPart.execution_payload_header
     let
       withdrawals_root = Opt.some actualEPH.withdrawals_root
       kzg_commitments = Opt.some(
-        executionPayloadHeader.get.blindedBlckPart.blob_kzg_commitments)
+        blindedBlockRes.get.blindedBlckPart.blob_kzg_commitments)
+      execution_requests = blindedBlockRes.get.executionRequests.get
```
tersec: This isn't valid when `EPH is deneb_mev.BlindedExecutionPayloadAndBlobsBundle`, and also, why do it at all?

Can just assign `let` it to a `default` value. It won't be used anyway, and it's even kind of semantically valid -- there are 0 execution requests for Deneb, always.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -968,26 +981,27 @@ proc getBlindedBlockParts[
...
     type PayloadType = electra.ExecutionPayloadForSigning
     template actualEPH: untyped =
-      executionPayloadHeader.get.blindedBlckPart.execution_payload_header
+      blindedBlockRes.get.blindedBlckPart.execution_payload_header
     let
       withdrawals_root = Opt.some actualEPH.withdrawals_root
       kzg_commitments = Opt.some(
-        executionPayloadHeader.get.blindedBlckPart.blob_kzg_commitments)
+        blindedBlockRes.get.blindedBlckPart.blob_kzg_commitments)
+      execution_requests = blindedBlockRes.get.executionRequests.get
```
tersec: Meanwhile, these unwrap the `executionRequests` from `blindedBlockRes` unconditionally, which is fine really, because they know the type. But it points to the unnecessary use of `Opt` here -- the code flow is based on the `SBBB`/`EPH` types statically, not on detecting whether these are `Opt.some` or `Opt.none` (which is never checked anyway, only assumed).

beacon_chain/validators/beacon_validators.nim
```diff
@@ -84,6 +84,7 @@ type
 
   BuilderBid[SBBB] = object
     blindedBlckPart*: SBBB
+    executionRequests*: Opt[ExecutionRequests]
```
tersec: This doesn't need to be `Opt`, either.

This is a generic object which knows (at compile-time) what `SBBB` it uses, and depending on that `SBBB`, it uses the value of `executionRequests` or not. This already happens, so all the `Opt` adds is some wrapping/unwrapping into/from the `Result` type:
```nim
    when EPH is deneb_mev.BlindedExecutionPayloadAndBlobsBundle:
      return ok(BuilderBid[EPH](
        blindedBlckPart: EPH(
          execution_payload_header: builderBid.header,
          blob_kzg_commitments: builderBid.blob_kzg_commitments),
        executionRequests: Opt.none(ExecutionRequests),
        executionPayloadValue: builderBid.value))
    elif EPH is electra_mev.BlindedExecutionPayloadAndBlobsBundle or
        EPH is fulu_mev.BlindedExecutionPayloadAndBlobsBundle:
      return ok(BuilderBid[EPH](
        blindedBlckPart: EPH(
          execution_payload_header: builderBid.header,
          blob_kzg_commitments: builderBid.blob_kzg_commitments),
        executionRequests: Opt.some(builderBid.execution_requests),
        executionPayloadValue: builderBid.value))
```
and then all of the readers of this object (even the Deneb one, for unclear reasons) use:
```nim
  execution_requests = blindedBlockRes.get.executionRequests.get
```
just without checking `Opt.some`/`Opt.none` (or `isSome`/`isNone`, etc). Can just drop the `Opt` aspect entirely, since this whole code flow already is based around knowing the block types statically at compile-time and therefore whether the `executionRequests` are valid/in use.

beacon_chain/validators/beacon_validators.nim
agnxsh: right

Issue #3973, opened 2022-08-16
1031 is not a valid program exit code for posix/systemd


arnetheduck:
> The value of status may be 0, EXIT_SUCCESS, EXIT_FAILURE, [[CX](javascript:open_code('CX'))] [Option Start]  or any other value, though only the least significant 8 bits (that is, status & 0377) shall be available to a waiting parent process. [Option End]

https://pubs.opengroup.org/onlinepubs/009695399/functions/exit.html

systemd in particular does not support parsing values >255:
```
Aug 16 16:37:42 raspi3 systemd[1]: /lib/systemd/system/nimbus_validator_client.service:49: Failed to parse value, ignoring: 1031
```

22.7.0

tersec:
And it's consistent with https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Exit%20Codes too which is why per https://github.com/status-im/nimbus-eth2/issues/3687 a >1000 code was chosen to begin with, so the usable gap is a possibly non-strict subset of 8 through 199 inclusive.

zah:
Fixed by https://github.com/status-im/nimbus-eth2/pull/3977 and shipped in Nimbus v22.8.0

PR #6881, opened 2025-01-28
Eliminate sporadic quit() calls.

cheatfate:
This PR converts all the established `quit` calls into `raiseDefect` statements and also establishes place where all this specific `Defects` being handled and appropriate exit code returned.

The only `quit()` left is Doppelganger one, i make it available for 2 reasons.
1. Is it ok to convert it to Defect?
2. `129` is invalid exit code number because it falls into the signals range. For Linux `SIGHUP` signal sent to process could also return exit code `129`.

### DIFF
--- a/beacon_chain/spec/defects.nim
+++ b/beacon_chain/spec/defects.nim
@@ -7,7 +7,7 @@
 
 type
   ExitProcessDefect* = object of Defect
-    exitCode*: int16
+    exitCode*: int8
 
   SuccessDefect* = object of ExitProcessDefect
   DatabaseDefect* = object of ExitProcessDefect
@@ -30,6 +30,9 @@ type
   TrustedSyncDefect* = object of ExitProcessDefect
   WalletsDefect* = object of ExitProcessDefect
   BeaconNodeDefect* = object of ExitProcessDefect
+  RecordDefect* = object of ExitProcessDefect
+  Web3CmdDefect* = object of ExitProcessDefect
+  SlashInterDefect* = object of ExitProcessDefect
 
 template declareExitHelpers(defect, code: untyped): untyped =
   template `raise defect`*() =
@@ -59,3 +62,15 @@ declareExitHelpers(DepositsDefect, 30)
 declareExitHelpers(TrustedSyncDefect, 31)
 declareExitHelpers(WalletsDefect, 32)
 declareExitHelpers(BeaconNodeDefect, 33)
+declareExitHelpers(RecordDefect, 34)
+declareExitHelpers(Web3CmdDefect, 35)
+declareExitHelpers(SlashInterDefect, 36)
+
+template withDefectsHandlers*(pbody, fbody: untyped) =
+  try:
+    pbody
+  except ExitProcessDefect as exc:
+    # This defects are already logged, so we just quit.
+    quit(exc.exitCode)
+  finally:
+    fbody
--- a/beacon_chain/spec/defects.nim
+++ b/beacon_chain/spec/defects.nim
@@ -38,6 +38,7 @@ type
 
   ValidatorClientDefect* = object of ExitProcessDefect
   LightClientDefect* = object of ExitProcessDefect
+  DepositContractDefect* = object of ExitProcessDefect
 
 template declareExitHelpers(defect, code: untyped): untyped =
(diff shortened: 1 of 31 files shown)


kdeme:
> 2\. `129` is invalid exit code number because it falls into the signals range. For Linux `SIGHUP` signal sent to process could also return exit code `129`

I don't remember exactly but wasn't that number chosen (arbitrary?) so that for example a service could be set up not to restart the process when the program terminated with that specific exit code?

beacon_chain/spec/defects.nim
```diff
@@ -0,0 +1,92 @@
...
+  except ExitProcessDefect as exc:
+    # This defects are already logged, so we just quit.
+    quit(exc.exitCode)
+  finally:
+    fbody
+
+template withDefectsHandlers*(pbody: untyped) =
+  try:
+    pbody
+  except ExitProcessDefect as exc:
```
arnetheduck, line 90: Catching `Defect` is [undefined behavior](https://nim-lang.org/docs/manual.html#definitions) - this is not something we want to use/rely on broadly - it can change meaning in future nim versions and/or not behave properly . 

Rather than defect, in order to avoid `quit`, the fix would be to return errors from the functions and deal with them _or_ let the defect be raised and not caught.

arnetheduck:
> 129 is invalid exit code number because it falls into the signals range. For Linux SIGHUP signal sent to process could also return exit code 129.

I don't remember the reasoning either, but it would break a lot of installations to change it now, because it's also part of the systemd service  file that we recommend. If SIGHUP can really cause this, we should probably install a handler to ignore it.

tersec:
> > 129 is invalid exit code number because it falls into the signals range. For Linux SIGHUP signal sent to process could also return exit code 129.
> 
> I don't remember the reasoning either, but it would break a lot of installations to change it now, because it's also part of the systemd service file that we recommend. If SIGHUP can really cause this, we should probably install a handler to ignore it.

It addresses https://github.com/status-im/nimbus-eth2/issues/3973 where previously it had been 1031, which was not a valid choice, and it doesn't conflict with https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Process%20Exit%20Codes.

https://nimbus.guide/doppelganger-detection.html documents this:
> If any activity is detected, the node shuts down with exit code 129.

beacon_chain/spec/defects.nim
cheatfate: Its first time i see that catching `Defect` is `undefined behavior`, because in such case generating `Defect` is also `undefined behavior` and code which was written with `Result.expect()` and `quit X` in mind should be totally rewritten before attempts to combine CL and EL into single binary. Otherwise `undefined behavior` created by CL could cause significant losses in EL (e.g. database could be left in `undefined state`).

cheatfate:
It doesn't matter where we used `129` - we should change this number to any number which is < 128.

beacon_chain/spec/defects.nim
tersec: > Otherwise `undefined behavior` created by CL could cause significant losses in EL (e.g. database could be left in `undefined state`).

ELs should not be able to put into undefined states by anything the engine API does. That's an EL bug.

arnetheduck: > generating Defect is also undefined behavior

a raised `Defect` is guaranteed to terminate the application - how it does that is undefined - ie `except Defect` may or may not work, ie you can't expect that `quit(exc.exitCode)` or anything inside this exception handler will actually run. 

This undefined behavior is why we don't use `except:` and `except Exception:` - they are both undefined as a consequence of catching `Defect` being undefined (I consider it a design flaw in the language).

The third correct option is to turn `ExitProcessDefect` into a `CatchableError`. 

From a code perspective, it's also bad to have untracked exceptions / hidden control flows like this - this goes back to our original decision to use `raises` consistently - `raises` tracking does not work with `Defect` either.

`expect` is the same as `doAssert`, but with a better error message - it works as intended, ie it terminates the application.

arnetheduck: > (e.g. database could be left in undefined state)

generally, all our database interactions happen in a transaction - the worst thing that can happen is a rollback.

PR #6889, opened 2025-02-04
Eliminate build warnings in applications.

cheatfate:
* elimination of 2 warnings in `nimbus_beacon_node` and 1 warning in `nimbus_signing_node`.
* make metrics server management for VC and BN equal.

P.S. It still has one build warning when building BN, but it requires Nim compiler version bump.

### DIFF
--- a/beacon_chain/nimbus_binary_common.nim
+++ b/beacon_chain/nimbus_binary_common.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2018-2024 Status Research & Development GmbH
+# Copyright (c) 2018-2025 Status Research & Development GmbH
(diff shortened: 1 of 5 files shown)


beacon_chain/nimbus_binary_common.nim
```diff
@@ -448,6 +448,40 @@ proc initKeymanagerServer*(
 
   KeymanagerInitResult(server: keymanagerServer, token: token)
 
+proc initMetricsServer*(
```
arnetheduck, line 451: it still runs in a separate thread, right?

cheatfate: Yup.

Issue #6901, opened 2025-02-06
POST `/eth/v1/validator/{pubkey}/feerecipient` of an existing remotekey requires beacon-chain restart to register in the relay


pablomendezroyo:
**Describe the bug**

After successfully [POST feerecipient](https://ethereum.github.io/keymanager-APIs/#/Fee%20Recipient/setFeeRecipient) through **validator keymanager API**, it seems like the beacon-chain service (BN) does not register this new FeeRecipient to MEVBoost/relays. When querying the relay the [validator registration data](https://flashbots.github.io/relay-specs/#/Data/getValidatorRegistration), the old FR is still there

**Expected behaviour**

Nimbus beacon-chain registers the new fee recipient in the relay.

**To Reproduce**
1. Platform details (OS, architecture):  running nimbus as **multiservice**  with docker.
2. Branch/commit used: nimbus version 25.1.0
3. Commands being executed: POST fee recipient endpoint. See https://ethereum.github.io/keymanager-APIs/#/Fee%20Recipient
4. Relevant log lines: 

**Additional context**
Running nimbus in dappnode. See https://github.com/dappnode/DAppNodePackage-nimbus-generic

pablomendezroyo:
It might be related to https://github.com/status-im/nimbus-eth2/issues/5983 . But still facing the same issue

tersec:
Related to https://github.com/status-im/nimbus-eth2/pull/6412 -- i.e. the issue here is likely that it's the VC, not the BN, which knows about this, since the in-BN version of this should be fixed already.

tersec:
More detail in https://github.com/status-im/nimbus-eth2/issues/5599#issuecomment-1813913879
> Specifying per-VC validator builder URLs is probably useful, but the rest of the infrastructure for that isn't there, so it'd be unused and untested code. `ValidatorRegistration` and that REST route, yeah, do not specify the builder API anywhere, so it's kind of a strangely fragmented set of configuration but it's what's currently specified.
> 
> There's also not really a mechanism currently to specify that the BN doesn't know/own the key pair of some VC/external/REST API validator, but does want to specify the route, which is another approach (still fragmented, but, would at least allow this to work).
> 
> So:
> 
> > If we have both VC and attached validators, we do not want the VC to be registered to all builders, but only to the one specified by `payload-builder-url`, so we should consider VC validators only when handling that builder's URL.
> 
[...]

PR #6907, opened 2025-02-07
VC: Fix validator registration cache issue.

cheatfate:
Should address #6901

### DIFF
--- a/beacon_chain/validator_client/common.nim
+++ b/beacon_chain/validator_client/common.nim
@@ -236,7 +236,6 @@ type
     beaconGenesis*: RestGenesis
     proposerTasks*: Table[Slot, seq[ProposerTask]]
     dynamicFeeRecipientsStore*: ref DynamicFeeRecipientsStore
-    validatorsRegCache*: Table[ValidatorPubKey, SignedValidatorRegistrationV1]
     blocksSeen*: Table[Slot, BlockDataItem]
     rootsSeen*: Table[Eth2Digest, Slot]
     processingDelay*: Opt[Duration]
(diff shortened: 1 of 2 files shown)


beacon_chain/validator_client/common.nim
```diff
@@ -1059,68 +1058,61 @@ proc isExpired(vc: ValidatorClientRef,
...
           return err(RegistrationKind.IncorrectTime)
         res.slot
 
-  if cached.isDefault() or vc.isExpired(cached, currentSlot):
-    if not cached.isDefault():
-      # Want to send it to relay, but not recompute perfectly fine cache
-      return ok(PendingValidatorRegistration(registration: cached, future: nil))
+  if validator.externalBuilderRegistration.isSome():
+    let cached = validator.externalBuilderRegistration.get()
+    if not(vc.isExpired(cached, currentSlot)):
```
tersec: It seems like this implementation basically recomputes all the signatures every epoch, even if nothing has changed (including either fee recipient or gas limit)?

If the
https://github.com/status-im/nimbus-eth2/blob/e1c616887f987a799f765eca2d26d9acd8f023b2/beacon_chain/validator_client/common.nim#L1079-L1082
case doesn't catch it, i.e. if it's more than an epoch old (regardless of changes), it unconditionally signs it again. This isn't efficient for either the VC or the relay -- I've gotten specific requests from a relay to ensure Nimbus caches the signatures, because that way, they can check (much more cheaply) whether the cached signature was shown to be valid for the object as a whole, rather than re-verify the signature. More importantly, this doesn't really scale for the VC, either, with hundreds or thousands of signatures at a few milliseconds each.

```nim
     # Want to send it to relay, but not recompute perfectly fine cache
```
is still a perfectly good goal overall, it's just that it should detect when the cache should be invalidated due to changes in one of the inputs (here, fee recipient or gas limit).

tersec: https://github.com/status-im/nimbus-eth2/pull/6907/commits/4c1b5bbe92512304073450db1c977e797851362e

PR #6913, opened 2025-02-10
implement `getBlobs` and bypass blob gossip validation on successful blob retrievals from EL

agnxsh:
https://github.com/ethereum/execution-apis/pull/559

if EL supports `engine_getBlobsV1`
- makes the block available as soon as the CL gets all the required blobs from the blob pool
- populates blob quarantine with that info (should probably verify proofs and inclusion once, like gossip?)
- pops from blob quarantine and add blobs to the block processor

else
- does everything the conventional way

Note: `getBlobs` in `nimbus-eth2` here is made to be backward compatible from Electra and NOT Deneb (Deneb is what the spec suggests)

Pros
- doesn't call kzg proof verification and inclusion proof verification for 9 blobs per slot (assuming the EL can be trusted)
- faster block import (in happy cases)

Cons
- relying heavily on the EL mempool can cause threats of various sorts, however, our block processor runs another kzg proof check before persisting the block and blobs, in case of a failure, the `RequestManager` should be able to fetch back the otherwise malicious blob/proof data sent by the bad EL, EL can be quickly swapped, Nimbus would rely on blob gossip that entire duration and switch back to optimistically pulling blobs from EL.

~~CI failing: because of some nim-web3 upstream incompatibilities~~

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -1001,7 +1001,7 @@ proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
 
 proc sendGetBlobs*(
     m: ELManager,
-    blck: electra.SignedBeaconBlock | fulu.SignedBeaconBlock
+    blck: electra.SignedBeaconBlock | fulu.SignedBeaconBlock,
 ): Future[Opt[seq[BlobAndProofV1]]] {.async: (raises: [CancelledError]).} =
   if m.elConnections.len == 0:
     return err()
@@ -1042,7 +1042,7 @@ proc sendGetBlobs*(
     await noCancel allFutures(pending)
 
     if bestResponse.isSome():
-      return ok(requests[bestResponse.get()].value().blobsAndProofs)
+      return ok(requests[bestResponse.get()].value())
 
     if timeoutExceeded:
       break
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -57,6 +57,7 @@ type
     GetPayloadV3Response |
     GetPayloadV4Response
 
+
 contract(DepositContract):
   proc deposit(pubkey: PubKeyBytes,
                withdrawalCredentials: WithdrawalCredentialsBytes,
@@ -108,6 +109,8 @@ const
   # https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.3/src/engine/shanghai.md#request-2
   GETPAYLOAD_TIMEOUT = 1.seconds
 
+  # https://github.com/ethereum/execution-apis/blob/ad9102b11212d51b736a0413c8655a8da93e55fc/src/engine/cancun.md#request-3
+  GETBLOBS_TIMEOUT = 1.seconds
   connectionStateChangeHysteresisThreshold = 15
     ## How many unsuccesful/successful requests we must see
     ## before declaring the connection as degraded/restored
@@ -862,6 +865,13 @@ proc sendNewPayloadToSingleEL(
     payload, versioned_hashes, Hash32 parent_beacon_block_root,
     executionRequests)
 
+proc sendGetBlobsToSingleEL(
+    connection: ELConnection,
+    versioned_hashes: seq[engine_api.VersionedHash]
+): Future[GetBlobsV1Response] {.async: (raises: [CatchableError]).} =
+  let rpcClient = await connection.connectedRpcClient()
+  await rpcClient.engine_getBlobsV1(versioned_hashes)
+
 type
   StatusRelation = enum
     newStatusIsPreferable
@@ -989,6 +999,61 @@ proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
     let pending = futures.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
     if len(pending) > 0:
       await noCancel allFutures(pending)
+
+proc sendGetBlobs*(
+    m: ELManager,
+    blck: electra.SignedBeaconBlock | fulu.SignedBeaconBlock
+): Future[Opt[seq[BlobAndProofV1]]] {.async: (raises: [CancelledError]).} =
+  if m.elConnections.len == 0:
+    return err()
+  let
+    timeout = GETBLOBS_TIMEOUT
+    deadline = sleepAsync(timeout)
+
+  var bestResponse = Opt.none(int)
+
+  while true:
+    let
+      requests = m.elConnections.mapIt(
+        sendGetBlobsToSingleEL(it, mapIt(
+            blck.message.body.blob_kzg_commitments,
+            engine_api.VersionedHash(kzg_commitment_to_versioned_hash(it)))))
+      timeoutExceeded =
+        try:
+          await allFutures(requests).wait(deadline)
+          false
+        except AsyncTimeoutError:
+          true
+        except CancelledError as exc:
+          let pending =
+            requests.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
+          await noCancel allFutures(pending)
+          raise exc
+
+    for idx, req in requests:
+      if not(req.finished()):
+        warn "Timeout while getting blob and proof",
+             url = m.elConnections[idx].engineUrl.url,
+             reason = req.error.msg
+      else:
+        if bestResponse.isNone:
+          bestResponse = Opt.some(idx)
+
+    let pending =
+      requests.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
+    await noCancel allFutures(pending)
+
+    if bestResponse.isSome():
+      return ok(requests[bestResponse.get()].value().blobsAndProofs)
+
+    else:
+      # should not reach this case
+      discard
+
+    if timeoutExceeded:
+      break
+
+  err()
 
 proc sendNewPayload*(
     m: ELManager,
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -57,7 +57,6 @@ type
     GetPayloadV3Response |
     GetPayloadV4Response
 
-
 contract(DepositContract):
   proc deposit(pubkey: PubKeyBytes,
                withdrawalCredentials: WithdrawalCredentialsBytes,
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -1007,8 +1007,7 @@ proc sendGetBlobs*(
   if m.elConnections.len == 0:
     return err()
   let
-    timeout = GETBLOBS_TIMEOUT
-    deadline = sleepAsync(timeout)
+    deadline = sleepAsync(GETBLOBS_TIMEOUT)
 
   var bestResponse = Opt.none(int)
 
@@ -1045,10 +1044,6 @@ proc sendGetBlobs*(
 
     if bestResponse.isSome():
       return ok(requests[bestResponse.get()].value().blobsAndProofs)
-
-    else:
-      # should not reach this case
-      discard
 
     if timeoutExceeded:
       break
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2018-2024 Status Research & Development GmbH
+# Copyright (c) 2018-2025 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -12,6 +12,7 @@ import
   chronicles, chronos, metrics,
   taskpools,
   ../spec/[helpers, forks],
+  ../el/el_manager,
   ../consensus_object_pools/[
     blob_quarantine, block_clearance, block_quarantine, blockchain_dag,
     attestation_pool, light_client_pool, sync_committee_msg_pool,
@@ -138,6 +139,10 @@ type
     # ----------------------------------------------------------------
     batchCrypto*: ref BatchCrypto
 
+    # EL integration
+    # ----------------------------------------------------------------
+    elManager*: ELManager
+
     # Missing information
     # ----------------------------------------------------------------
     quarantine*: ref Quarantine
@@ -169,6 +174,7 @@ proc new*(T: type Eth2Processor,
           blobQuarantine: ref BlobQuarantine,
           rng: ref HmacDrbgContext,
           getBeaconTime: GetBeaconTimeFn,
+          elManager: ELManager,
           taskpool: Taskpool
          ): ref Eth2Processor =
   (ref Eth2Processor)(
@@ -186,6 +192,7 @@ proc new*(T: type Eth2Processor,
     quarantine: quarantine,
     blobQuarantine: blobQuarantine,
     getCurrentBeaconTime: getBeaconTime,
+    elManager: elManager,
     batchCrypto: BatchCrypto.new(
       rng = rng,
       # Only run eager attestation signature verification if we're not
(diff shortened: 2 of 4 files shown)


agnxsh:
for devnet testing locally, use `ethpandaops/nimbus-eth2:getBlobs`

beacon_chain/el/el_manager.nim
```diff
@@ -57,6 +57,7 @@ type
     GetPayloadV3Response |
     GetPayloadV4Response
 
+
```
tersec: ?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -459,21 +460,69 @@ proc initFullNode(
                              maybeFinalized: bool):
         Future[Result[void, VerifierError]] {.async: (raises: [CancelledError]).} =
       withBlck(signedBlock):
-        when consensusFork >= ConsensusFork.Deneb:
+        when consensusFork >= ConsensusFork.Electra:
+          # Pull blobs and proofs from the EL blob pool
+          let blobsFromElOpt = await node.elManager.sendGetBlobs(forkyBlck)
+          debugEcho "pulled blobs from el"
```
tersec: Obligatory comment about `debugEcho` (but yeah, it's a draft PR and useful for now)

beacon_chain/el/el_manager.nim
```diff
@@ -990,6 +1000,61 @@ proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
...
+    let pending =
+      requests.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
+    await noCancel allFutures(pending)
+
+    if bestResponse.isSome():
+      return ok(requests[bestResponse.get()].value().blobsAndProofs)
+
+    else:
+      # should not reach this case
```
tersec: How "should" is this (i.e. if it's a logic bug, or a potentially buggy EL, or etc)?

beacon_chain/el/el_manager.nim
```diff
@@ -990,6 +1000,61 @@ proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
...
       await noCancel allFutures(pending)
 
+proc sendGetBlobs*(
+    m: ELManager,
+    blck: electra.SignedBeaconBlock | fulu.SignedBeaconBlock
+): Future[Opt[seq[BlobAndProofV1]]] {.async: (raises: [CancelledError]).} =
+  if m.elConnections.len == 0:
+    return err()
+  let
+    timeout = GETBLOBS_TIMEOUT
```
tersec, line 1010: It looks like `timeout` isn't used anywhere here, but also, why not directly use `GETBLOBS_TIMEOUT`?

beacon_chain/el/el_manager.nim
```diff
@@ -990,6 +1000,61 @@ proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
...
+    deadline = sleepAsync(timeout)
```
tersec: Why not directly use `GETBLOBS_TIMEOUT`?

agnxsh: i was considering to add some extra processing latency

beacon_chain/nimbus_beacon_node.nim
agnxsh: just for now, until i see things are ok in the local devnet

beacon_chain/el/el_manager.nim
agnxsh: oops, will cleanup

beacon_chain/el/el_manager.nim
tersec: Well, at least it's nice to keep the `timeout` itself `const`, even then.

tersec: https://github.com/status-im/nimbus-eth2/pull/6913/commits/82857482ff9c8520eda5267ed413ed5b78e74e91

beacon_chain/el/el_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6913/commits/82857482ff9c8520eda5267ed413ed5b78e74e91

beacon_chain/el/el_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6913/commits/3fa2603243ed78d3affd2fe400701fee1d55f11d

agnxsh:
image

```
2025-02-14 13:24:27 INF 2025-02-14 07:54:27.006+00:00 Slot end                                   topics="beacnde" slot=234 nextActionWait=1s993ms123us121ns nextAttestationSlot=235 nextProposalSlot=236 syncCommitteeDuties=current head=4932c763:234
2025-02-14 13:24:28 WRN 2025-02-14 07:54:28.713+00:00 Peer count low, no new peers discovered    topics="networking" discovered_nodes=0 new_peers=@[] current_peers=1 wanted_peers=160
2025-02-14 13:24:29 INF 2025-02-14 07:54:28.999+00:00 Slot start                                 topics="beacnde" nextFork=Fulu:100000001 head=4932c763:234 delay=137us922ns finalized=5:5177fdfd peers=1 slot=235 sync=synced epoch=7
2025-02-14 13:24:29 pulled blobs from el
2025-02-14 13:24:29 9
2025-02-14 13:24:31 INF 2025-02-14 07:54:31.295+00:00 Attestation sent                           topics="beacval" attestation="(committee_index: 0, attester_index: 28, data: (slot: 235, index: 0, beacon_block_root: \"61632bd0\", source: \"6:f956b1f3\", target: \"7:a67fac0a\"), signature: \"929f5919\")" delay=-1s703ms988us327ns subnet_id=11
2025-02-14 13:24:31 INF 2025-02-14 07:54:31.295+00:00 Attestation sent                           topics="beacval" attestation="(committee_index: 0, attester_index: 52, data: (slot: 235, index: 0, beacon_block_root: \"61632bd0\", source: \"6:f956b1f3\", target: \"7:a67fac0a\"), signature: \"b5d6aa5c\")" delay=-1s703ms537us202ns subnet_id=11
2025-02-14 13:24:31 INF 2025-02-14 07:54:31.296+00:00 Sync committee message sent 
```

we're finalising with nimbus running `getBlobs` in a devnet.

agnxsh:
we can get some better decision making on whether to validate blob gossip or pull blobs from EL every slot, by caching `engine_exchangeCapabilities`, however a good interval should be decided as ELs can be swapped on the run.

agnxsh:
https://discordapp.com/channels/595666850260713488/1252403418941624532/1341996653149945927

an interesting mainnet research from Jimmy (Lighthouse) suggests, that there are 0-8 blob misses per minute by the EL, hence we shall only bypass gossip validation for blobs completely when the EL can serve a complete response (i.e, blobs from EL == kzg comms len).

an alternative can be done, where for every block we can fetch _most_ blobs from the EL, and the rest from gossip val, but in that case, we'd likely spend some time checking what we missed from EL, which isn't very ideal.

image
grafana for ref

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -266,10 +274,63 @@ proc processSignedBeaconBlock*(
...
-    self: var Eth2Processor, src: MsgSource,
-    blobSidecar: deneb.BlobSidecar, subnet_id: BlobId): ValidationRes =
+    self: ref Eth2Processor, src: MsgSource,
+    blobSidecar: deneb.BlobSidecar, subnet_id: BlobId):
+    Future[ValidationRes] {.async: (raises: [CancelledError]).} =
   template block_header: untyped = blobSidecar.signed_block_header.message
+  let block_root = hash_tree_root(block_header)
+
+  let vEl =
+    await self.validateBlobSidecarFromEL(block_root)
```
tersec: it seems like in the case of a pathological EL, this could delay things by 1 second?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2095,6 +2098,33 @@ proc installMessageValidators(node: BeaconNode) =
...
+              ): Future[ValidationResult] {.async: (raises: [CancelledError]).} =
+                let
+                  fut1 =
+                    (node.processor.processBlobSidecarFromEL(blobSidecar))
+                  fut2 =
+                    (node.processor.processBlobSidecar(MsgSource.gossip,
+                                                       blobSidecar,
+                                                       subnet_id))
+
+                return waitFor toValidationRace(fut1, fut2))
```
tersec: `waitFor` isn't ideal

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -266,10 +308,77 @@ proc processSignedBeaconBlock*(
...
+            return ok()
+
+          elif blobsEl.len < forkyBlck.message.body.blob_kzg_commitments.len and
+              blobsEl.len != 0:
+            let end_time = Moment.now()
+
+            el_blob_loss = forkyBlck.message.body.blob_kzg_commitments.len - blobsEl.len
+
+            debug "Time taken to receive partially response from EL",
+                  received_percent = float((blobsEl.len div forkyBlck.message.body.blob_kzg_commitments.len) * 100),
```
tersec, line 366: `div` is [integer division](https://nim-lang.org/docs/system.html#div%2Cint%2Cint); this doesn't do what it seems to want, because the `float` conversion is pointless. It will return 0 all the time (or 100)

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/6913/commits/36c848b217a2ce770858633a26ad8cf68fba6849

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -266,10 +308,77 @@ proc processSignedBeaconBlock*(
...
+              blobs_from_el = blobsEl.len
+            return ok()
+
+          elif blobsEl.len < forkyBlck.message.body.blob_kzg_commitments.len and
+              blobsEl.len != 0:
+            let end_time = Moment.now()
+
+            el_blob_loss = forkyBlck.message.body.blob_kzg_commitments.len - blobsEl.len
+
+            debug "Time taken to receive partially response from EL",
```
tersec, line 365: `partial response` (or `partially receive response`, whichever seems more appropriate, but the "receive" here seems done and not in progress, so I'd lean toward former)

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -266,10 +308,77 @@ proc processSignedBeaconBlock*(
...
+          await self.elManager.sendGetBlobs(forkyBlck)
+        if blobsFromElOpt.get.len > 0 and blobsFromElOpt.isSome():
+          let blobsEl = blobsFromElOpt.get()
+
+          # check lengths of array[BlobAndProofV1] with blobs
+          # kzg commitments of the signed block
+          if blobsEl.len == forkyBlck.message.body.blob_kzg_commitments.len:
+
+            # we have got all blobs from EL, now we can
+            # conveniently the blobless block from quarantine
```
tersec, line 335: verb?

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -149,9 +157,41 @@ type
 
   ValidationRes* = Result[void, ValidationError]
 
+
```
tersec, line 160: ?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2095,6 +2098,33 @@ proc installMessageValidators(node: BeaconNode) =
               await node.processor.processBlsToExecutionChange(
                 MsgSource.gossip, msg)))
 
+      when consensusFork >= ConsensusFork.Electra:
+        # blob_sidecar_{subnet_id}
+        # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.5/specs/deneb/p2p-interface.md#blob_sidecar_subnet_id
```
tersec, line 2103: Newly introduced URLs should be current version (currently `v1.5.0-beta.3`)

beacon_chain/el/el_manager.nim
```diff
@@ -990,6 +999,56 @@ proc lazyWait(futures: seq[FutureBase]) {.async: (raises: []).} =
...
+          true
+        except CancelledError as exc:
+          let pending =
+            requests.filterIt(not(it.finished())).mapIt(it.cancelAndWait())
+          await noCancel allFutures(pending)
+          raise exc
+
+    for idx, req in requests:
+      if not(req.finished()):
+        warn "Timeout while getting blob and proof",
```
tersec, line 1033: Is this really `warn`? It's not necessarily a Nimbus error, and it shouldn't disrupt Nimbus functioning because of the `race`. So one could end up with a scary-looking `WRN` but no malfunction.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -149,9 +157,41 @@ type
...
+  except ValueError:
+    # This shouldn't normally happen with 2 futures
+    return ValidationResult.Ignore
+
+  let
+    first = if raced == f1b: f1 else: f2
+    second = if raced == f1b: f2 else: f1
+
+  try:
+    let res1 = await first
```
tersec, line 183:
```nim
  ## Returns a future which will complete and return finished FutureBase,
  ## when one of the given futures will be completed, failed or canceled.
  ##
  ## On success returned Future will hold finished FutureBase.
  ##
  ## On cancel futures in ``futs`` WILL NOT BE cancelled.
```
under what circumstance should `first` have to be `await`ed?

Explicitly:
```
Returns a future which will complete and return finished FutureBase
```
That is, one should just be able to call `complete()` on it? cc @cheatfate 

This might not apply to `await second`, because that could still be in progress, but by definition the winner of the race should be `complete`able.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2095,6 +2098,33 @@ proc installMessageValidators(node: BeaconNode) =
...
+        let subnets =
+          when consensusFork >= ConsensusFork.Electra:
+            node.dag.cfg.BLOB_SIDECAR_SUBNET_COUNT_ELECTRA
+          else:
+            node.dag.cfg.BLOB_SIDECAR_SUBNET_COUNT
+        # It's safe to try as many times to fetch EL blobs as
+        # there are blob subnets.
+        for it in 0.BlobId ..< subnets.BlobId:
```
tersec, line 2111: There might be more blobs than subnets, see:
- https://github.com/status-im/nimbus-eth2/pull/6967

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -266,10 +308,77 @@ proc processSignedBeaconBlock*(
...
+    self: ref Eth2Processor,
+    blobSidecar: deneb.BlobSidecar):
+    Future[ValidationRes]
+    {.async: (raises: [CancelledError]).} =
+  template block_header: untyped = blobSidecar.signed_block_header.message
+  let block_root = hash_tree_root(block_header)
+  if (let o = self.quarantine[].getBlobless(block_root); o.isSome):
+    let blobless = o.get()
+    withBlck(blobless):
+      when consensusFork >= ConsensusFork.Electra:
```
tersec, line 321: When this isn't true, it's a bug, but also kind of falls off the edge of the control flow of `withBlck(blobless)` and implicitly returns some default thing. Better to handle this case explicitly somehow and flag it as an error (`doAssert`, even, if it's really a logic error to enter this function with a pre-Electra block, but only if that can be actually guaranteed, otherwise some usual runtime error).

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2095,6 +2098,33 @@ proc installMessageValidators(node: BeaconNode) =
...
+      when consensusFork >= ConsensusFork.Electra:
+        # blob_sidecar_{subnet_id}
+        # https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.5/specs/deneb/p2p-interface.md#blob_sidecar_subnet_id
+        let subnets =
+          when consensusFork >= ConsensusFork.Electra:
+            node.dag.cfg.BLOB_SIDECAR_SUBNET_COUNT_ELECTRA
+          else:
+            node.dag.cfg.BLOB_SIDECAR_SUBNET_COUNT
+        # It's safe to try as many times to fetch EL blobs as
+        # there are blob subnets.
```
tersec, line 2110: There might be more blobs than blob subnets, so for any given block, each subnet might be triggered multiple times. This doesn't happen in Dencun and won't in Pectra, but in theory it could, so the comment at least is misleading.

The code structure works fine for this -- it will just trigger the `toValidationRace(...)` per blob, not per blob subnet.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2106,12 +2136,12 @@ proc installMessageValidators(node: BeaconNode) =
...
             let subnet_id = it
-            node.network.addValidator(
+            node.network.addAsyncValidator(
               getBlobSidecarTopic(digest, subnet_id), proc (
                 blobSidecar: deneb.BlobSidecar
-              ): ValidationResult =
-                toValidationResult(
-                  node.processor[].processBlobSidecar(
+              ): Future[ValidationResult] {.async: (raises: [CancelledError]).} =
+                return toValidationResult(
```
tersec, line 2143: Chronos doesn't' need the explicit `return` anymore for async functions.

tersec:
Decided not to proceed with this approach.

PR #6914, opened 2025-02-11
Bump `is_better_update` to match latest spec

etan-status:
Add `signature_slot` comparison to `is_better_update` LC data comparator for robustness in edge cases, and sync style with latest specs.

- https://github.com/ethereum/consensus-specs/pull/4124

### DIFF
--- a/beacon_chain/spec/helpers.nim
+++ b/beacon_chain/spec/helpers.nim
@@ -279,7 +279,7 @@ func get_safety_threshold*(store: ForkyLightClientStore): uint64 =
     store.current_max_active_participants
   ) div 2
 
-# https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.5/specs/altair/light-client/sync-protocol.md#is_better_update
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/helpers.nim
```diff
@@ -279,7 +279,7 @@ func get_safety_threshold*(store: ForkyLightClientStore): uint64 =
     store.current_max_active_participants
   ) div 2
 
-# https://github.com/ethereum/consensus-specs/blob/v1.4.0-beta.5/specs/altair/light-client/sync-protocol.md#is_better_update
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.3/specs/altair/light-client/sync-protocol.md#is_better_update
```
tersec, line 282: doesn't seem to exist?

tersec: or is this coming in next spec release?

etan-status: the linked PR is not part of b.2, so anticipating b.3

etan-status: Like, it was merged into `dev` upstream from which they branch releases.

PR #6917, opened 2025-02-12
update Electra attestation and aggregation gossip validation

tersec:


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1085,31 +1085,74 @@ proc validateAttestation*(
...
-    return errReject("SingleAttestation: attester index too high")
-  let validator_index = attestation.attester_index.ValidatorIndex
+  # [REJECT] The committee index is within the expected range -- i.e.
+  # data.index < get_committee_count_per_slot(state, data.target.epoch).
+  let committee_index = block:
+    let idx = shufflingRef.get_committee_index(attestation.committee_index)
+    if idx.isErr():
+      return pool.checkedReject(
+        "Attestation: committee index not within expected range")
+    idx.get()
```
etan-status, line 1155: In the `get_beacon_committee` below, we should probably use `committee_index` variable instead of `attestation.committee_index.CommitteeIndex`

tersec: https://github.com/status-im/nimbus-eth2/commit/0ec1cc9bce12ae76645ce36263cb4d2352d6618a

PR #6918, opened 2025-02-12
correctly build arm64 macos Nim compiler

narimiran:
Trying to see if the correct version of Nim compiler will be built on arm64 macos.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,12 @@ jobs:
             builder: ['self-hosted','ubuntu-22.04']
           - target:
               os: macos
+              cpu: amd64
             builder: macos-13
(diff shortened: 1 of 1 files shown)


etan-status:
As the arm64 test fails with a stacktrace and libunwind is used on macOS, maybe this helps:

- https://github.com/status-im/nimbus-eth2/pull/6919

tersec:
> As the arm64 test fails with a stacktrace and libunwind is used on macOS, maybe this helps:
> 
>     * [Bump nim-libbacktrace to `616e5f4a6e2c751b85487e2e855c6b7f90848503` #6919](https://github.com/status-im/nimbus-eth2/pull/6919)

Merged into `unstable`.

PR #6929, opened 2025-02-16
add: basic validator custody

agnxsh:
https://github.com/ethereum/consensus-specs/pull/3871, a prequel to https://github.com/ethereum/consensus-specs/pull/3779

### DIFF
--- a/beacon_chain/spec/datatypes/fulu.nim
+++ b/beacon_chain/spec/datatypes/fulu.nim
@@ -64,6 +64,15 @@ const
   CUSTODY_REQUIREMENT* = 4
   NUMBER_OF_CUSTODY_GROUPS* = 128
 
+  # Minimum number of custody groups an honest node with
+  # validators attached custodies and serves samples from
+  VALIDATOR_CUSTODY_REQUIREMENT* = 8
+
+
+  # Balance increment corresponding to one additional group to custody
+  # 2**5 * 10**9 (= 32,000,000,000) Gwei
+  BALANCE_PER_ADDITIONAL_CUSTODY_GROUP*: uint64 = 32000000000'u64
+
   # Number of columns in the network per custody group
   COLUMNS_PER_GROUP* = NUMBER_OF_COLUMNS div NUMBER_OF_CUSTODY_GROUPS
 
--- a/beacon_chain/spec/datatypes/fulu.nim
+++ b/beacon_chain/spec/datatypes/fulu.nim
@@ -59,7 +59,7 @@ const
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/p2p-interface.md#configuration
   DATA_COLUMN_SIDECAR_SUBNET_COUNT* = 128
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/das-core.md#custody-setting
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/fulu/das-core.md#custody-setting
   SAMPLES_PER_SLOT* = 8
   CUSTODY_REQUIREMENT* = 4
   NUMBER_OF_CUSTODY_GROUPS* = 128
@@ -67,7 +67,6 @@ const
   # Minimum number of custody groups an honest node with
   # validators attached custodies and serves samples from
   VALIDATOR_CUSTODY_REQUIREMENT* = 8
-
 
   # Balance increment corresponding to one additional group to custody
   # 2**5 * 10**9 (= 32,000,000,000) Gwei
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -342,3 +342,14 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
     return err("DataColumnSidecar: validation failed")
 
   ok()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/fulu/das-core.md#validator-custody
+proc get_validators_custody_requirement*(state: fulu.BeaconState,
+                                         validator_indices: seq[ValidatorIndex]):
+                                         uint64 =
+  let
+    totalNodeBalance =
+      validator_indices.mapIt(state.balances[it]).foldl(a + b, 0.Gwei)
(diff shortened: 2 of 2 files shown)


beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -64,6 +64,15 @@ const
   CUSTODY_REQUIREMENT* = 4
   NUMBER_OF_CUSTODY_GROUPS* = 128
 
+  # Minimum number of custody groups an honest node with
+  # validators attached custodies and serves samples from
+  VALIDATOR_CUSTODY_REQUIREMENT* = 8
```
tersec, line 69: https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/fulu/das-core.md#custody-setting (updated from
```nim
  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/das-core.md#custody-setting
``` 
would document where these two come from

beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -64,6 +64,15 @@ const
...
+
+
```
tersec: extra newline?

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -342,3 +342,14 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
     return err("DataColumnSidecar: validation failed")
 
   ok()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/fulu/das-core.md#validator-custody
+proc get_validators_custody_requirement*(state: fulu.BeaconState,
```
tersec: `func`

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -342,3 +342,14 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
...
+                                         validator_indices: seq[ValidatorIndex]):
+                                         uint64 =
+  let
+    totalNodeBalance =
+      validator_indices.mapIt(state.balances[it]).foldl(a + b, 0.Gwei)
```
tersec: This starts by making a copy of all the relevant validator balances, first, when it can/should "stream" them in with O(1) memory usage. 

For comparison, the spec formulation is
```python
def get_validators_custody_requirement(state: BeaconState, validator_indices: Sequence[ValidatorIndex]) -> uint64:
    total_node_balance = sum(state.balances[index] for index in validator_indices)
    count = total_node_balance // BALANCE_PER_ADDITIONAL_CUSTODY_GROUP
    return min(max(count, VALIDATOR_CUSTODY_REQUIREMENT), NUMBER_OF_CUSTODY_GROUPS)
```
It's true that what you wrote is a very literal copy of the spec (though to be even more literal it could use [`sum()`](https://nim-lang.org/docs/math.html#sum%2CopenArray%5BT%5D)), but it's fine/preferred to avoid O(n^2) costs when O(n) is achievable, or O(n) where O(1) is achievable. And accumulating a sum does not require an initial `mapIt`/`seq`-copy.

Typically we'd copy the `snake_case` from the spec when transliterating the spec functions this directly, i.e. `total_node_balance`.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -342,3 +342,14 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
     return err("DataColumnSidecar: validation failed")
 
   ok()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/fulu/das-core.md#validator-custody
+proc get_validators_custody_requirement*(state: fulu.BeaconState,
+                                         validator_indices: seq[ValidatorIndex]):
```
tersec: It's nice to use `openArray` for these parameters when feasible, because then can call them with an `array` too. Here it looks feasible.

beacon_chain/spec/datatypes/fulu.nim
agnxsh: i thought it would have gotten auto-updated

tersec: evidently it didn't -- the auto-update script is very sensitive. any change at all means it doesn't auto-update. that could be whitespace, etc.

tersec: also it needs an entire .md or HTML file not to change, not only the one section it points to. It's safe/conservative in its replacements but simpleminded. It does no markdown parsing at all.

beacon_chain/spec/peerdas_helpers.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/6929/commits/a2955681b5b7a4b88f5aae0a6d6f3fa70d746494

beacon_chain/spec/peerdas_helpers.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/6929/commits/a2955681b5b7a4b88f5aae0a6d6f3fa70d746494

beacon_chain/spec/peerdas_helpers.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/6929/commits/a2955681b5b7a4b88f5aae0a6d6f3fa70d746494

beacon_chain/spec/datatypes/fulu.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/6929/commits/a2955681b5b7a4b88f5aae0a6d6f3fa70d746494

beacon_chain/spec/datatypes/fulu.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/6929/commits/a2955681b5b7a4b88f5aae0a6d6f3fa70d746494

beacon_chain/spec/peerdas_helpers.nim
tersec: `total_node_balance`?

agnxsh: https://github.com/status-im/nimbus-eth2/pull/6929/commits/cb4d787b4f748626bb424811b90875d05bd709f6

PR #6936, opened 2025-02-19
removes the boolean "payload withheld" from the payload envelope and adds the slot to the payload envelope.

Tomi-3-0:


beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -79,7 +79,7 @@ const
   # https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip7732/beacon-chain.md#payload-status
   PAYLOAD_ABSENT* = PTCStatus(0)
   PAYLOAD_PRESENT* = PTCStatus(1)
-  PAYLOAD_WITHHELD* = PTCStatus(2)
+  PAYLOAD_WITHHELD* = PTCStatus(2)
```
agnxsh: why this space?

Tomi-3-0: An oversight...thanks

PR #6939, opened 2025-02-20
Move defaultDataDir procedure to nimbus_binary_common.

pedromiguelmiranda:
This intends to fix all Nimbus clients  declaring the same function, and once its merged, unify this use on all of them.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -55,6 +55,7 @@ const
   defaultGasLimit* = 36_000_000
   defaultAdminListenAddressDesc* = $defaultAdminListenAddress
   defaultBeaconNodeDesc = $defaultBeaconNode
+  defaultDataDirPath* = "BeaconNode"
 
 when defined(windows):
   {.pragma: windowsOnly.}
@@ -161,7 +162,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store all blockchain data"
-      defaultValue: config.defaultDataDir("BeaconNode")
+      defaultValue: config.defaultDataDir(defaultDataDirPath)
       defaultValueDesc: ""
       abbr: "d"
       name: "data-dir" .}: OutDir
@@ -897,7 +898,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store all blockchain data"
-      defaultValue: config.defaultDataDir("BeaconNode")
+      defaultValue: config.defaultDataDir(defaultDataDirPath)
       defaultValueDesc: ""
       abbr: "d"
       name: "data-dir" .}: OutDir
@@ -1078,7 +1079,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store validator's keys"
-      defaultValue: config.defaultDataDir("BeaconNode")
+      defaultValue: config.defaultDataDir(defaultDataDirPath)
       defaultValueDesc: ""
       abbr: "d"
       name: "data-dir" .}: OutDir
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -30,7 +30,7 @@ import
   ./el/el_conf,
   ./filepath
 
-from std/os import getHomeDir, parentDir, `/`
+from std/os import `/`
 from std/strutils import parseBiggestUInt, replace
 from consensus_object_pools/block_pools_types_light_client
   import LightClientDataImportMode
@@ -161,7 +161,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store all blockchain data"
-      defaultValue: config.defaultDataDir()
+      defaultValue: config.defaultDataDir("BeaconNode")
       defaultValueDesc: ""
       abbr: "d"
       name: "data-dir" .}: OutDir
@@ -897,7 +897,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store all blockchain data"
-      defaultValue: config.defaultDataDir()
+      defaultValue: config.defaultDataDir("BeaconNode")
       defaultValueDesc: ""
       abbr: "d"
       name: "data-dir" .}: OutDir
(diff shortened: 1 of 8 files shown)


beacon_chain/conf.nim
```diff
@@ -161,7 +161,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store all blockchain data"
-      defaultValue: config.defaultDataDir()
+      defaultValue: config.defaultDataDir("BeaconNode")
```
cheatfate: This is definitely should not be string literal, because its easy to make mistake which you have made.

beacon_chain/conf.nim
```diff
@@ -897,7 +897,7 @@ type
...
```
cheatfate: This is not BeaconNode configuration it is Validator Client configuration.

beacon_chain/conf.nim
```diff
@@ -1078,7 +1078,7 @@ type
 
     dataDir* {.
       desc: "The directory where nimbus will store validator's keys"
-      defaultValue: config.defaultDataDir()
+      defaultValue: config.defaultDataDir("BeaconNode")
```
cheatfate: This is not BeaconNode configuration it is Signing Node configuration

beacon_chain/conf.nim
pedromiguelmiranda: ~~thanks. Which mistake was made?~~
Ok, I see your point, force type checking with an enum or a table. That would be interesting, however check the comments below.
Deserves a `const` though

beacon_chain/conf.nim
pedromiguelmiranda: thanks.
all of these declarations were using the same `defaultDataDir` procedure with "BeaconNode" hardcoded on the procedure it self.
I just made it generic, and move it from `beacon_chain/conf` to `nimbus_binary_common` module.

Is it safe to start using a diferent data dir for each client and/or component?

beacon_chain/conf.nim
pedromiguelmiranda: same comment as before.

arnetheduck:
The general direction we want to be going in is that we have a directory structure that allows the applications to co-exist in it - ie imagine beacon node, vc and execution client as separate binaries - if I pass the same `--data-dir` to each of them separately, the outcome should be the same if I run them in a single binary - if I run the single binary, stop it and then run just the beacon node with the same `--data-dir`, that should "just work".

arnetheduck:
the consequence of the above is that we want to tie directory structure to "functionality" and not to any particular end-user application (ie it doesn't matter if the user starts a "validator client" as a separate binary or use the integrated beacon validator - it's the "validator functionality" is what guides what the layout looks like - in this sense, it doesn't make sense to have names like "beacon node" or "validator client" as part of the directory structure.

pedromiguelmiranda:
> the consequence of the above is that we want to tie directory structure to "functionality" and not to any particular end-user application (ie it doesn't matter if the user starts a "validator client" as a separate binary or use the integrated beacon validator - it's the "validator functionality" is what guides what the layout looks like - in this sense, it doesn't make sense to have names like "beacon node" or "validator client" as part of the directory structure.

 I agree and personally like the direction, and this is the first step towards it: 
   - Make the procedure generic for all clients (they will be addressed later on a different PR) and removing the hardcoded path "BeaconNode" keeping the same behaviour as before.

Other steps, besides the obvious one of applying the same strategy to other clients still in development phase, is that Nimbus beacon node client is public and users are not aware of the changes proposed (removing the "BeaconNode" path)

   A strategy and specially a delivery method need to be defined and tested among all clients. The delivery part is more concerning given that we we need to either warn users using default data directory to move the data manually, or doing it during installation or doing it via code.
   
Strategies, in my opinion, to be addressed and reasoned about on upcoming PR's

beacon_chain/conf.nim
```diff
@@ -55,6 +55,7 @@ const
   defaultGasLimit* = 36_000_000
   defaultAdminListenAddressDesc* = $defaultAdminListenAddress
   defaultBeaconNodeDesc = $defaultBeaconNode
+  defaultDataDirPath* = "BeaconNode"
```
cheatfate, line 58: I dont understand why we need this subfolder which would be named `BeaconNode` for `Validator Client` and for `Signing Node`???

cheatfate: and why we need this argument?

pedromiguelmiranda: - The "BeaconNode" subfolder was already there hardcoded in `defaultDataDir*`

- The argument allows to keep eth2 users upgrades without issues if any of them is using default data directory. 
Until we define a directory structure for all nimbus clients.

- Fullfills the PR objective: move `defaultDataDir*` procedure to nimbus_common module to be used in other nimbus clients, without "_breaking_" current distributed eth2 functionalities.

beacon_chain/conf.nim
arnetheduck: So - this actually brings to light a problem with the defaults: 

if `BeaconNode` is hardcoded, both the beacon node *and* the validator client default to `~/.cache/nimbus/BeaconNode`, it means you will end up with *both* the validator client and the beacon node reading validator keys from the same folder and therefore both attesting and therefore risk slashing. 

This is something to fix, so that the validator client keys are kept in a separately named folder in the defaults.

beacon_chain/conf.nim
arnetheduck: per the validator client comments above about validator keys, this is actually something to address, perhaps even in this PR. 

What my comments about the layout imply is that _under_ the data directory, applications should not introduce additional "levels" of named folders, but the default _for_ the data directory needs to take into consideration a user launching the application without arguments and here, we need to separate the data directories of vc and bn or they will conflict / use the same validator keys.

cc @cheatfate have you tested importing keys into the vc without supplying a data directory, and then running both bn and vc without data directory? what happens?

cheatfate: what does it mean importing? copy files from one folder to another? If you will run BN or VC without data directory they both will going to access same folder...

pedromiguelmiranda:
thank you @arnetheduck  and @cheatfate for the inputs. Seems like we need to test all scenarios of nimbus clients using the same base directory or if a set of directories is needed.
More important, a delivery strategy needs to be defined which wouldn't affect users.

Given that the risk is higher for such low gains in the codebase, let's  close this PR and circle back in the future
when we address/test these topics.

PR #6951, opened 2025-02-24
Use consistent merkleization limit for `/eth/v1/beacon/blob_sidecars`

etan-status:
Use the `MAX_BLOB_COMMITMENTS_PER_BLOCK` constant as merkleization limit for the `/eth/v1/beacon/blob_sidecars` response so that `hash_tree_root` results in same value regardless of configured serialization limits. At compile-time, we don't have access to the actual serialization limit, so using a mainnet derived value in the type definition is not flexible.

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -1702,10 +1702,12 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
         res.get()
 
     # https://github.com/ethereum/beacon-APIs/blob/v2.4.2/types/deneb/blob_sidecar.yaml#L2-L28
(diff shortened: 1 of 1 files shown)


tersec:
Sure, why not, though these should never be merkleized before being serialized for network sending anyway. The REST server does not calculate htr, otherwise this hack wouldn't work.

etan-status:
The theoretical limit is 4096. The runtime limit is configurable (but we cannot put the runtime limit into the type definition).

tersec:
It's not even clear that it needs a `List[...]` because no htr or similar -- it's just that the serialization machinery is better-tested/tuned for `List[...]`s. Any limit in this `List` is pointless/arbitrary. `MAX_BLOB_COMMITMENTS_PER_BLOCK` is as good as any.

PR #6957, opened 2025-02-24
Extend `RuntimeConfig` with constants for blob sidecar limits

etan-status:
Since v1.5.0-alpha.4 the blob sidecar limits are no longer configured in the presets but are now part of the network configuration. This sets up `RuntimeConfig` to parse those values so that can phase out the compile time constants later.


etan-status:
Note Gnosis also changed Deneb from 6 from preset to 2 via config.

tersec:
> Note Gnosis also changed Deneb from 6 from preset to 2 via config.

How can this possibly be safe and not trigger forking, for a hardfork already in progress?

etan-status:
#minority client :)

they have a weird config, subnet count is 6 but blob count is 2.

- https://github.com/gnosischain/specs/blob/master/consensus/config/gnosis.yaml#L137-L140
- https://github.com/gnosischain/specs/pull/54/files

Here they removed `MAX_BLOBS_PER_BLOCK: 6` from the preset and put it into `MAX_BLOBS_PER_BLOCK: 2` in gnosis.yaml with documentation that suggests it is intentional.

whatever

etan-status:
It seems safe because in the EL they might already be using 2: "Gnosis currently has a non-equal max variable between CL and EL"

PR #6964, opened 2025-02-25
Allow runtime config of `MAX_REQUEST_BLOB_SIDECARS`

etan-status:
With `MAX_REQUEST_BLOB_SIDECARS` becoming dynamic, we can no longer use the value from network config in the `List` datatype and have to rely on an upper bound constant for the maximum supported value instead and then check the actual length at runtime.

As a side effect, this also extends `BlobIdentifierList` to use the correct electra specific request limit instead of the Deneb constant.


beacon_chain/spec/datatypes/constants.nim
```diff
@@ -93,7 +93,3 @@ const
 
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.0/specs/electra/beacon-chain.md#execution-1
   MAX_BLOBS_PER_BLOCK_ELECTRA* = 9'u64
-
-  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/electra/p2p-interface.md#configuration
-  MAX_REQUEST_BLOB_SIDECARS_ELECTRA* =
-    MAX_REQUEST_BLOCKS_DENEB * MAX_BLOBS_PER_BLOCK_ELECTRA
```
etan-status, line 99: In `presets.nim` we have the same equivalence check dynamically:

```nim
  checkCompatibility MAX_REQUEST_BLOCKS_DENEB * cfg.MAX_BLOBS_PER_BLOCK,
                     "MAX_REQUEST_BLOB_SIDECARS"
  checkCompatibility MAX_REQUEST_BLOCKS_DENEB * cfg.MAX_BLOBS_PER_BLOCK_ELECTRA,
                     "MAX_REQUEST_BLOB_SIDECARS_ELECTRA"
```

PR #6977, opened 2025-03-03
Respond to un-synced blocks request with `ResourceUnavailable`

etan-status:
When serving beacon blocks by range requests, there are three outcomes:

1. We have blocks available: Response is non-empty
2. The entire epoch actually is empty: Response is empty
3. We have not backfilled that far or have pruned the history

To distinguish (2) from (3), send `ResourceUnavailable` if the requested range is for blocks earlier than our backfill progress.

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -238,6 +238,7 @@ p2pProtocol BeaconSync(version = 1,
       # Peers that are unable to reply to block requests within the
       # `MIN_EPOCHS_FOR_BLOCK_REQUESTS` epoch range SHOULD respond with
       # error code `3: ResourceUnavailable`.
+      # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/phase0/p2p-interface.md#responding-side
       raise newException(ResourceUnavailableError, BlocksUnavailable)
 
     let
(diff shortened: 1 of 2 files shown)


beacon_chain/sync/sync_protocol.nim
```diff
@@ -233,8 +233,14 @@ p2pProtocol BeaconSync(version = 1,
       raise newException(InvalidInputsError, "Empty range requested")
 
     var blocks: array[MAX_REQUEST_BLOCKS.int, BlockId]
+    let dag = peer.networkState.dag
+    if startSlot < dag.backfill.slot:
+      # Peers that are unable to reply to block requests within the
```
tersec, line 238: Appears to come from https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/phase0/p2p-interface.md#responding-side which would be useful to link

beacon_chain/sync/sync_protocol.nim
```diff
@@ -233,8 +233,14 @@ p2pProtocol BeaconSync(version = 1,
...
+      # `MIN_EPOCHS_FOR_BLOCK_REQUESTS` epoch range SHOULD respond with
+      # error code `3: ResourceUnavailable`.
```
etan-status, line 240:
```suggestion
      # error code `3: ResourceUnavailable`.
      # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/phase0/p2p-interface.md#responding-side
```

PR #6985, opened 2025-03-05
security: fixed sidecar response checking allowing subsets

agnxsh:
Sidecar responses that have partial responses to what was requested should be considered _happy_ responses even if the response is _atleast_ ONE sidecar.

This PR splits response sanity checking and subset checking in order to give better visibility in `DEBUG` logs.

Thanks to @fredriksvantes (EF Security) for the find!

### DIFF
--- a/beacon_chain/sync/request_manager.nim
+++ b/beacon_chain/sync/request_manager.nim
@@ -128,24 +128,22 @@ func checkResponse(roots: openArray[Eth2Digest],
 
 func cmpSidecarIdentifier(x: BlobIdentifier | DataColumnIdentifier,
                           y: ref BlobSidecar | ref DataColumnSidecar): int =
-  cmp(x.index, y.index)
+  cmp(x.index, y[].index)
 
-func checkResponse(idList: seq[BlobIdentifier],
-                   blobs: openArray[ref BlobSidecar]): bool =
+func checkResponseSanity(idList: seq[BlobIdentifier],
+                         blobs: openArray[ref BlobSidecar]): bool =
+  # Cannot respond more than what I have asked
   if blobs.len > idList.len:
     return false
   var i = 0
   while i < blobs.len:
     let
-      block_root = hash_tree_root(blobs[i].signed_block_header.message)
-      id = idList[i]
+      block_root =
+        hash_tree_root(blobs[i][].signed_block_header.message)
+      idListKey = binarySearch(idList, blobs[i], cmpSidecarIdentifier)
 
-    # Check if the blob response is a subset
-    if binarySearch(idList, blobs[i], cmpSidecarIdentifier) == -1:
-      return false
-
-    # Verify block_root and index match
-    if id.block_root != block_root or id.index != blobs[i].index:
+    # Verify the block root
+    if idList[idListKey].block_root != block_root:
       return false
 
     # Verify inclusion proof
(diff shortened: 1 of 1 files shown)


beacon_chain/sync/request_manager.nim
```diff
@@ -178,6 +181,13 @@ func checkResponse(idList: seq[DataColumnIdentifier],
     inc i
   true
 
+func checkResponseSubset(idList: seq[DataColumnIdentifier],
+                         columns: openArray[ref DataColumnSidecar]): bool =
+  ## Clients MUST respond with at least one sidecar, if they have it.
+  ## Clients MAY limit the number of blocks and sidecars in the response.
+  ## https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/fulu/p2p-interface.md#datacolumnsidecarsbyroot-v1
+  idList.mapIt(it.index).toHashSet() <= columns.mapIt(it[].index).toHashSet()
```
tersec: This doesn't have to be two `toHashSet` constructions -- can construct one of them only and then iterate linearly through the other to do the O(1) checks

beacon_chain/sync/request_manager.nim
```diff
@@ -154,22 +152,27 @@ func checkResponse(idList: seq[BlobIdentifier],
...
   true
 
-func checkResponse(idList: seq[DataColumnIdentifier],
-                   columns: openArray[ref DataColumnSidecar]): bool =
+func checkResponseSubset(idList: seq[BlobIdentifier],
+                         blobs: openArray[ref BlobSidecar]): bool =
+  ## Clients MUST respond with at least one sidecar, if they have it.
+  ## Clients MAY limit the number of blocks and sidecars in the response.
+  ## https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/deneb/p2p-interface.md#blobsidecarsbyroot-v1
+  idList.mapIt(it.index).toHashSet() <= blobs.mapIt(it[].index).toHashSet()
```
tersec: also this doesn't have to be two toHashSet constructions -- can construct one of them only and then iterate linearly through the other to do the O(1) checks

agnxsh: https://github.com/status-im/nimbus-eth2/pull/6985/commits/5410dc53d254c9374cd7c7e61db9b009b43ffba9

beacon_chain/sync/request_manager.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/6985/commits/5410dc53d254c9374cd7c7e61db9b009b43ffba9

beacon_chain/sync/request_manager.nim
```diff
@@ -154,22 +152,31 @@ func checkResponse(idList: seq[BlobIdentifier],
...
   true
 
-func checkResponse(idList: seq[DataColumnIdentifier],
-                   columns: openArray[ref DataColumnSidecar]): bool =
+func checkResponseSubset(idList: seq[BlobIdentifier],
+                         blobs: openArray[ref BlobSidecar]): bool =
+  ## Clients MUST respond with at least one sidecar, if they have it.
+  ## Clients MAY limit the number of blocks and sidecars in the response.
+  ## https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/deneb/p2p-interface.md#blobsidecarsbyroot-v1
+  let idHSet = idList.mapIt(it.index).toHashSet()
```
tersec: This also creates a useless intermediate result, the fully-reified list of validator indices.

It'd be nice if Nim used one of the many approaches available to avoid this code style (which I like, basically), from generating useless objects/GC load, but it does, the result of that `idList.mapIt(it.index)` is only ever iterated over, but it's created in full.

Ditto the other example of this code pattern here.

agnxsh: https://github.com/status-im/nimbus-eth2/pull/6985/commits/1275a8b21df3bfe93905a38e3b3ac1ab5aa0489c

PR #6987, opened 2025-03-05
fix: removed unnecessary `fi` after function

fragchain:
Fixed an issue with an unnecessary `fi` that was left after the function.
It's now removed, making the code cleaner and working properly.


tersec:
This isn't a correct fix

PR #7021, opened 2025-03-14
Add `hoodi` testnet

etan-status:
Hoodi is the new public testnet. Add support to use it like Holesky.

- https://github.com/eth-clients/hoodi


Makefile
```diff
@@ -600,35 +599,6 @@ define CLEAN_NETWORK
 	rm -rf build/data/shared_$(1)*/*.log
 endef
 
-###
```
tersec: Why remove Sepolia yet?

etan-status: Unused makefile bitrot

docs/the_nimbus_book/src/eth1.md
```diff
@@ -36,9 +36,9 @@ Select an execution client and install it, configuring it such that the authenti
         geth --authrpc.addr localhost --authrpc.port 8551 --authrpc.vhosts localhost --authrpc.jwtsecret /tmp/jwtsecret
         ```
 
-    === "Holesky"
+    === "Hoodi"
         ```
-        geth --holesky --authrpc.addr localhost --authrpc.port 8551 --authrpc.vhosts localhost --authrpc.jwtsecret /tmp/jwtsecret
+        geth --hoodi --authrpc.addr localhost --authrpc.port 8551 --authrpc.vhosts localhost --authrpc.jwtsecret /tmp/jwtsecret
```
tersec, line 41: I'm not sure `nimbus-eth1` at least supports it yet, creating an issue there presently.

docs/the_nimbus_book/src/faq.md
```diff
@@ -181,7 +181,7 @@ In other words, if you stood to earn ≈0.01 ETH, you would instead be penalized
 
 ### How can I keep track of my validator?
 
-One way of keeping track is using an online service such as beaconcha.in: [Mainnet](https://beaconcha.in/) or [Holesky](https://holesky.beaconcha.in).
+One way of keeping track is using an online service such as beaconcha.in: [Mainnet](https://beaconcha.in/) or [Hoodi](https://hoodi.beaconcha.in).
```
tersec, line 184: does this URL exist yet

docs/the_nimbus_book/src/metrics-pretty-pictures.md
```diff
@@ -14,7 +14,7 @@ You can learn more about Grafana [here](https://github.com/grafana/grafana).
 To enable the metrics server, run the beacon node  with the `--metrics` flag:
 
 ```
-./run-holesky-beacon-node.sh --metrics
+./run-hoodi-beacon-node.sh --metrics
```
tersec: ideally we'd stop recommending using this `run-foo-beacon-node.sh` (which also might not exist for Hoodi as of this PR? and shouldn't, ideally?) antipattern.

It's just inertia and it creates problems more than anything else.

docs/the_nimbus_book/src/faq.md
etan-status: no, it's a draft. however will eventually exist.

docs/the_nimbus_book/src/suggested-fee-recipient.md
```diff
@@ -37,9 +37,9 @@ Fee recipients are recorded publicly on-chain as part of proposed blocks, so sug
     ./run-mainnet-beacon-node.sh --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842
     ```
 
-=== "Holesky"
+=== "Hoodi"
     ```sh
-    ./run-holesky-beacon-node.sh --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842
+    ./run-hoodi-beacon-node.sh --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842
```
tersec: The trouble with this is partly that it ends up putting things in different data directories, then the sensible thing for virtually any troubleshooting is to stop using these wrapper scripts, and then the trouble compounds

docs/the_nimbus_book/src/eth1.md
etan-status: yeah, all of the links and EL refs will eventually resolve, they are not yet reachable.

run-hoodi-beacon-node.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
tersec: I said this last time for last testnet, but it's _really_ time to stop making these pointless wrapper scripts, they're worse than not having them.

Yes, the documentation needs to point to something. We have weeks, it's fine.

tersec:
More generally, all of this appears to depend on
- https://github.com/eth-clients/hoodi/pull/3

merging. Which presumably will happen but hasn't yet.

etan-status:
Yes, it's a draft. It also depends on the .ssz.sz download becoming available.

Makefile
tersec: Okay, it's reasonable actually, not sure it's best as part of this PR

docs/the_nimbus_book/src/eth1.md
tersec: - https://github.com/status-im/nimbus-eth1/issues/3143

Makefile
tersec: https://github.com/status-im/nimbus-eth2/pull/7021/commits/7c20f3e3d0574d495b9b1b30142d061ebd43ba9d

docs/the_nimbus_book/src/eth1.md
tersec: Also probably geth doesn't support `--hoodi` either yet, at least not released versions.

etan-status: They will, eventually. Hoodi is the replacement for Holesky, essentially. Genesis is tomorrow, if the predicted names / links don't end up being correct we can fix them again. It's a similar case of when future spec tag name has to be anticipated. chicken and egg.

run-hoodi-beacon-node.sh
etan-status: Maybe we can inline them into the documentation, I guess that's their point anyway? For Holesky, the same approach was used as for Sepolia. I think updating the docs shouldn't be that big of a task.

tersec: They key part of what the script does is impose the defaults:
```
: ${NODE_ID:=0}
: ${DATA_DIR_NAME:="shared_${NETWORK}_${NODE_ID}"}
: ${DATA_DIR:="build/data/${DATA_DIR_NAME}"}
: ${BASE_P2P_PORT:=9000}
: ${BASE_REST_PORT:=5052}
```

The `BASE_P2P_PORT` and `BASE_REST_PORT` are such already, but they got offset by `NODE_ID`. The documentation doesn't describe any scenarios where `NODE_ID` is set to something non-default/0 and by definition they can't possibly exist yet for a testnet which was created in the last couple of days and for which there hasn't been a release. Even now, I've never seen someone deploying Nimbus at scale reference using a non-zero`NODE_ID` for the existing scripts (but sure, keep them unchanged). This leaves `DATA_DIR` and `DATA_DIR_NAME`, and that's where the `build/data/shared_hooli_0` names come from.

So, yes -- basically instead of
```
./run-hoodi-beacon-node.sh --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842
```
for example, can use either
- `build/nimbus_beacon_node --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` to allow Nimbus to use its default of, e.g., on Linux, `~/.cache/nimbus/BeaconNode`  (see `getDefaultDataDir()` for details). UX on this is debatable;
- `build/nimbus_beacon_node --data-dir=build/data/shared_hoodi_0 --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` to replicate the existing wrapper semantics in the way the documentation uses it;
- `build/nimbus_beacon_node --data-dir=build/data/shared_hoodi --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` to drop the suffix; or
- same idea, `build/nimbus_beacon_node --data-dir="${FOO}" --suggested-fee-recipient=0x70E47C843E0F6ab0991A3189c28F2957eb6d3842` to taste for anything else. As long as it's internally consistent.

And not have the wrapper script created at all. It's already not how `log-rotate.md`, `pi-guide.md`, `trusted-node-sync.md`, and `validator-client.md`, refer to it, and they use the technique depicted for compatibility.

tersec: In general, this doesn't have the initial `cd "$(dirname $0)"` either, but none of the are a good way to run a node longer-term except through some kind of process supervisor such as `systemd`, to handle automatic restarts, to hook into logging (otherwise, this is how we get people trying to report issues with screenshots of consoles of logs), et cetera. So these are all already kind of demo-commands except for a few pages which get into actual deployment discipline, and copying and pasting them doesn't create viable long-term running node. Better to make that clear early.

etan-status: It's weird to combine these changes for Hoodi, but have added a commit with those script cleanups as well.

etan-status:
```
# 2025-Mar-17 12:00:00 PM UTC
MIN_GENESIS_TIME: 1742212800
```

run-hoodi-beacon-node.sh
tersec: Once they exist, they'll have to "always" exist. The best/only time to avoid their creation is before anyone anywhere can possibly have relied on them.

The other option is to split off the documentation changes from the `--network=holesky` bump.

PR #7029, opened 2025-03-21
`ColumnSyncer`

agnxsh:


beacon_chain/sync/column_syncer_assist.nim
```diff
@@ -0,0 +1,927 @@
...
+        await cas.resetWait(some(rewindSlot))
+        break
+
+    if processingCb != nil:
+      processingCb()
+
+    # Validating received blocks one by one
+    var
+      hasInvalidBlock = false
+      unviableBlock: Option[(Eth2Digest, Slot)]
```
tersec, line 572: In general, `Opt` is typically preferable to `Option`

beacon_chain/sync/column_syncer_assist.nim
```diff
@@ -0,0 +1,927 @@
...
+proc checkDataColumnsResponse*[T](req: ColumnSyncRequest[T],
+                                  data: openArray[Slot],
+                                  maxBlobsPerBlockElectra: uint64):
+                                  Result[void, string] =
+  if data.len == 0:
+    return ok()
+
+  if lenu64(data) > (req.count * NUMBER_OF_COLUMNS):
+    # Number of data columns in the response should be less than
+    # or equal to (MAX_BLOBS_PER_BLOCK_FULU * NUMBER_OF_COLUMNS).
```
tersec, line 146: `MAX_BLOBS_PER_BLOCK_FULU` isn't a thing anymore.

beacon_chain/sync/column_syncer_assist.nim
```diff
@@ -0,0 +1,927 @@
...
+  cas.outSlot = minSlot
+  # Waking up all waiters and wait for last one.
+  await cas.wakeupAndWaitWaiters()
+
+proc isEmpty*[T](sr: ColumnSyncResult[T]): bool {.inline.} =
+  ## Returns ``true`` if response chain of blocks is empty (has only
+  ## empty slots).
+  len(sr.data) == 0
+
+proc hasEndGap*[T](sr: ColumnSyncResult[T]): bool {.inline.} =
```
tersec, line 287: Not called outside `column_syncer_assist`, so doesn't need to be exported.

beacon_chain/sync/column_syncer_assist.nim
```diff
@@ -0,0 +1,927 @@
...
+  ## Returns ``true`` if response chain of blocks has a gap at the end
+  let lastSlot =
+    sr.request.slot + sr.request.count - 1'u64
+  if len(sr.data) == 0:
+    return true
+  if sr.data[^1][].slot != lastslot:
+    return true
+  return false
+
+proc getLastNonEmptySlot*[T](sr: ColumnSyncResult[T]): Slot {.inline.} =
```
tersec, line 297: Not called outside `column_syncer_assist`, so doesn't need to be exported.

beacon_chain/sync/column_syncer_assist.nim
```diff
@@ -0,0 +1,927 @@
...
+  case cas.direction
+  of ColumnSyncerDirection.Forward:
+    # Calculate the latest finalized epoch
+    let finalizedEpoch = epoch(safeSlot)
+
+    # Calculate failure epoch
+    let failEpoch = epoch(failSlot)
+
+    # Calculate exponential rewind point in number of epochs.
+    let epochCount =
```
tersec, line 362: Isn't this substantially replicating the kind of logic already in the existing block syncer?

tersec: In general, is it important that the column syncer have consistent behavior with the block syncer in this rewind behavior, for DA purposes?

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+
+proc getColumnSyncerStats[A, B](man: ColumnManager[A, B]):
+                                tuple[map: string,
+                                      sleeping: int,
+                                      waiting: int,
+                                      pending: int] =
+  var map = newString(len(man.workers))
+  var sleeping, waiting, pending: int
+  for i in 0..<len(man.workers):
+    var ch: char
```
tersec, line 990: This whole design seems mostly copy/paste from the existing block syncer.

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+    map[i] = ch
+  (map, sleeping, waiting, pending)
+
+proc startColumnSyncWorkers[A, B](man: ColumnManager[A, B]) =
+  # Starting all the column sync workers
+  if ColumnSyncerFlag.Greedy in man.flags:
+    for i in 0..<len(man.workers):
+      man.workers[i].future =
+        columnSyncWorkerGreedy[A, B](man, i)
+  if ColumnSyncerFlag.Impartial in man.flags:
```
tersec, line 1022: Is it valid for `ColumnSyncerFlag.Greedy in man.flags` and `ColumnSyncerFlag.Impartial in man.flags` to be true? If not, seems maybe better to use an `enum`.

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+
+proc stopColumnSyncWorkers[A, B](man: ColumnManager[A, B])
+                                {.async: (raises: []).} =
+  # Cancelling all the column sync workers
+  let pending = man.workers.mapIt(it.future.cancelAndWait())
+  await noCancel allFutures(pending)
+
+proc timeLeftForColumnSyncer*(d: Duration): string =
+  if d == InfiniteDuration:
+    "--h--m"
```
tersec, line 1035: Pretty sure this formatting code already lives somewhere (probably the syncer) in Nimbus

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+            sleeping_workers_count = sleeping,
+            waiting_workers_count = waiting,
+            pending_workers_count = pending,
+            wall_head_slot = wallSlot,
+            local_head_slot = headSlot,
+            pause_time = $chronos.seconds(pauseTime),
+            avg_sync_speed = man.avgSyncSpeed.formatBiggestFloat(ffDecimal, 4),
+            ins_sync_speed = man.insSyncSpeed.formatBiggestFloat(ffDecimal, 4),
+            topics = "columnsync"
+    of ColumnSyncerDirection.Backward:
```
tersec, line 1134: Does the column syncer need to work backwards at all? Just don't say the columns are in custody yet, and if the network built on them, well...

This isn't how DA checking works for blobs, but maybe it does for columns.

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
+
+import std/[strutils, sequtils, algorithm]
+import stew/base10, chronos, chronicles, results
+import
+  ../spec/datatypes/[phase0, altair],
```
tersec, line 13: Nothing which imports `forks` needs to import `phase0, altair` also, it's covered already.

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+  ../spec/eth2_apis/rest_types,
+  ../spec/[helpers, forks, network, peerdas_helpers],
+  ../networking/[peer_pool, peer_scores, eth2_network],
+  ../gossip_processing/block_processor,
+  ../beacon_clock,
+  "."/[sync_protocol, sync_queue, column_syncer_assist]
+
+export phase0, altair, merge, chronos, chronicles, results,
```
tersec, line 21: Or exports it, likewise.

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+
+  StatusExpirationTime* = chronos.minutes(2)
+
+type
+  ColumnSyncerStatus* {.pure.} = enum
+    Sleeping, WaitingPeer, UpdatingStatus, Requesting, Downloading,
+    Queueing, Processing
+
+  ColumnAndBlockResponse* = object
+    blk*: fulu.SignedBeaconBlock
```
tersec, line 40: Does this store the block redundantly with non-column-syncer things?

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+
+  debugEcho "fetching block for column navigation"
+
+  doAssert(not(req.isEmpty()), "Request must not be empty!")
+  debug "Requesting blocks from peer",
+         peer_score = req.item.getScore(),
+         peer_speed = req.item.netKbps(),
+         topics = "columnsync"
+
+  beaconBlocksByRange_v2(peer, req.slot, req.count, 1'u64)
```
tersec, line 207: Can this redundantly fetch blocks that the block syncer also is fetching?

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+proc filterRelevantPeers[A, B](man: ColumnManager[A, B],
+                               peers: seq[A],
+                               w_index: int):
+                               Future[seq[A]] {.async: (raises: [CancelledError]).} =
+  ## This iterates over the available peers
+  ## and returns a refreshed peer list based on
+  ## whichever's peer status is recent and relevant
+  ##
+
+
```
tersec, line 326: kind of trivial formatting issue, but this is just a lot of whitespace

beacon_chain/sync/column_syncer.nim
tersec: Also, it's fine for now, but assuming Glamsterdam also uses PeerDAS, this won't always be a `fulu.SignedBeaconBlock`.

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+  ColumnManager*[A, B] = ref object
+    pool: PeerPool[A, B]
+    cfg*: RuntimeConfig
+    amIsupernode*: bool
+    custody_columns_set*: HashSet[ColumnIndex]
+    custody_columns_list*: List[ColumnIndex, NUMBER_OF_COLUMNS]
+    column_syncer_table*: OrderedTable[Slot, ColumnAndBlockResponse]
+    FULU_FORK_EPOCH: Epoch
+    MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS: uint64
+    MAX_BLOBS_PER_BLOCK_ELECTRA: uint64
```
tersec, line 62: Would this need to know about BPOs?

beacon_chain/sync/column_syncer.nim
```diff
@@ -0,0 +1,1289 @@
...
+      dur = toFloatSeconds(finish.timestamp - start.timestamp)
+    slots / dur
+
+proc initColumnSyncerAssist[A, B](man: ColumnManager[A, B]) =
+  case man.direction
+  of ColumnSyncerDirection.Forward:
+    man.assist = ColumnSyncerAssist.init(A, man.direction, man.getFirstSlot(),
+                                         man.getLastSlot(), man.chunkSize,
+                                         man.getSafeSlot, man.peerdasBlockVerifier, 1)
+  of ColumnSyncerDirection.Backward:
```
tersec, line 120: A general comment on backfilling, and a concern here a bit too: the existing backfill mechanism has a tendency to crowd out forward syncing if there's less than a couple dozen peers. This seems to add yet more demand for syncing peers.

Is the idea that the "main" block sync gets shut off, or is the pool of sync peers now split between multiple mini-pools of 10 workers each?

beacon_chain/sync/column_syncer.nim
agnxsh: > Does this store the block redundantly with non-column-syncer things?

downloading blocks for this pass, is useful to detect rewinding of slots

beacon_chain/sync/column_syncer.nim
agnxsh: this fetches the blocks just in order to cross-check the blocks by the dag and decide on potential rewinds and other slot movements

beacon_chain/sync/column_syncer.nim
agnxsh: yeah i think it will

beacon_chain/sync/column_syncer_assist.nim
agnxsh: > In general, is it important that the column syncer have consistent behavior with the block syncer in this rewind behavior, for DA purposes?

yes, it's important in a few ways because here we are range downloading columns, and it's significant to understand and detect partial/malicious responses, responses from orphaned blocks that can be quickly rejected, and slot movements can be performed accordingly.

beacon_chain/sync/column_syncer.nim
agnxsh: mostly because i thought these concepts shouldn't change, as they didn't change the same for blob sidecars, either, but i do agree that there can be leaner ways of integrating columns within the existing block syncer, i will take a look at that

beacon_chain/sync/column_syncer.nim
agnxsh: > seems maybe better to use an enum.

ok

beacon_chain/sync/column_syncer.nim
agnxsh: yeah overall, because the sync speed math shouldn't change significantly for this either, but i'll still try to reduce this much code reuse,i'd say i did keep most of the code draft-ish in order to see if things worked properly

beacon_chain/sync/column_syncer.nim
agnxsh: > is the pool of sync peers now split between multiple mini-pools of 10 workers each?

something like this

beacon_chain/sync/column_syncer.nim
tersec: Right, but how likely is it that it and the block syncer are/have fetched the same blocks around the same time?

This kind of thing works ok, if wastefully, if there are surplus syncing peers and notably less well when there are fewer potential sync peers.

beacon_chain/sync/column_syncer.nim
tersec: What's the practical minimum number of total peers to be comfortable with the combinations of
- forward syncing blocks
- backfilling blocks
- syncing columns?

beacon_chain/sync/column_syncer.nim
tersec: That's reasonable here, sure.

beacon_chain/sync/column_syncer.nim
tersec: As you note, for a draft, for mainly devnets, and since there might be another syncer implementation entirely coming, it's reasonable enough for what it is.

tersec:
Superceded by other PeerDAS branches.

Issue #7019, opened 2025-03-14
Nimbus returns 404 post-Electra for non-zero committee index on `GET /eth/v2/validator/aggregate_attestation`


eth2353:
**Describe the bug**
Hey team, it looks like Nimbus does not return an aggregated attestation correctly post-Electra and only returns one correctly for a committee index value of 0. The same issue is _not_ present on Teku, Lodestar or Lighthouse.

**To Reproduce**
1. Platform details (OS, architecture): `N/A`
2. Branch/commit used: latest `ethpandaops/nimbus-eth2:unstable-minimal`
3. Commands being executed: `nimbus_beacon_node --non-interactive=true [...]` (standard Kurtosis config for Nimbus CL)
4. Relevant log lines: No apparently related log lines in the Nimbus beacon node logs

validator-client-side log lines show the 404 response:
`ERROR: Failed to get response from 172.16.0.15 for GET /eth/v2/validator/aggregate_attestation: ValueError('Received status code 404 for request to http://172.16.0.15:4000/eth/v2/validator/aggregate_attestation?attestation_data_root=0xd267290189e2c954e30e43c6d21ba4e35e88403e87b466b6a62ec61cd86c6e52&slot=219&committee_index=1 Full response text: {"code":404,"message":"Unable to retrieve an aggregated attestation"}')`

I included a Kurtosis config for reproducibility below.

**Screenshots**

Metric values for `rate(requests_total{service="vc-2-geth-nimbus-vero", path="/eth/v2/validator/aggregate_attestation"}[1m])`
<img width="1481" alt="Image" src="https://github.com/user-attachments/assets/c300a0fd-8d1b-4bb2-9077-f821fea56aac" />
--> you can see in Deneb, the first 2 minutes, all requests return a 200 status code (green area). Post-Electra, a significant part of the responses are 404 responses (blue area).

**Additional context**
I see this same issue on Holesky too.

This can be reproduced easily the following way using the current `master` version of `ethereum-package` (minimal config is used in order to get >1 committee):

```
participants:
  - el_type: reth
    el_image: ghcr.io/paradigmxyz/reth:v1.2.0
    cl_type: teku
    cl_image: consensys/teku:25.2.0
    use_separate_vc: true
    vc_type: vero
    vc_image: ghcr.io/serenita-org/vero:v1.0.0-rc.1
    use_remote_signer: true
    remote_signer_image: "consensys/web3signer:develop"
  - el_type: geth
[...]

PR #7037, opened 2025-03-28
fix incorrect indexing of attestationdata in attestationpool

tersec:
Probably responsible for https://github.com/status-im/nimbus-eth2/issues/7019

@pedromiguelmiranda supplied the repro test case.


pedromiguelmiranda:
Confirmed. solves the issue.

PR #7045, opened 2025-04-01
Make MEV decoding less strict for Deneb encoded responses.

cheatfate:
…eb fork.

### DIFF
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -3285,7 +3285,7 @@ proc decodeBytesJsonOrSsz*(
     version: string
 ): Result[T, RestErrorMessage] =
   var res {.noinit.}: T
-  let typeFork = kind(typeof(res.data))
+  const typeFork = kind(typeof(res.data))
 
   if contentType == ApplicationJsonMediaType:
     res =
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -3282,15 +3282,21 @@ proc decodeBytesJsonOrSsz*(
     T: typedesc[MevDecodeTypes],
     data: openArray[byte],
     contentType: Opt[ContentTypeData],
-    version: string
+    version: string,
+    strictDecode: bool = false
 ): Result[T, RestErrorMessage] =
   var res {.noinit.}: T
 
   let
     typeFork = kind(typeof(res.data))
-    consensusFork = ConsensusFork.decodeString(version).valueOr:
-      return err(RestErrorMessage.init(Http400, UnableDecodeVersionError,
-                                       [version, $error]))
+    consensusFork =
+      if strictDecode:
+        ConsensusFork.decodeString(version).valueOr:
+          return err(RestErrorMessage.init(Http400, UnableDecodeVersionError,
+                                           [version, $error]))
+      else:
+        typeFork
+
   if typeFork != consensusFork:
     return err(
       RestErrorMessage.init(Http400, UnexpectedForkVersionError,
--- a/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
+++ b/beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
@@ -3282,26 +3282,10 @@ proc decodeBytesJsonOrSsz*(
     T: typedesc[MevDecodeTypes],
     data: openArray[byte],
     contentType: Opt[ContentTypeData],
-    version: string,
-    strictDecode: bool = false
+    version: string
 ): Result[T, RestErrorMessage] =
   var res {.noinit.}: T
-
-  let
-    typeFork = kind(typeof(res.data))
-    consensusFork =
-      if strictDecode:
-        ConsensusFork.decodeString(version).valueOr:
-          return err(RestErrorMessage.init(Http400, UnableDecodeVersionError,
-                                           [version, $error]))
-      else:
-        typeFork
-
-  if typeFork != consensusFork:
-    return err(
-      RestErrorMessage.init(Http400, UnexpectedForkVersionError,
-                            ["eth-consensus-version", consensusFork.toString(),
-                             typeFork.toString()]))
+  let typeFork = kind(typeof(res.data))
 
   if contentType == ApplicationJsonMediaType:
     res =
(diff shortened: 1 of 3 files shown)


tersec:
As https://github.com/attestantio/vouch/issues/309#issuecomment-2770794230 points out, `eth-consensus-version` is not a required header for any fork; it shouldn't depend on it at all.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3282,26 +3282,10 @@ proc decodeBytesJsonOrSsz*(
...
-                                           [version, $error]))
-      else:
-        typeFork
-
-  if typeFork != consensusFork:
-    return err(
-      RestErrorMessage.init(Http400, UnexpectedForkVersionError,
-                            ["eth-consensus-version", consensusFork.toString(),
-                             typeFork.toString()]))
+  let typeFork = kind(typeof(res.data))
```
tersec: This should be able to be `const`

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -3321,6 +3311,17 @@ proc decodeBytesJsonOrSsz*(
                                typeFork.toString()]))
     ok(res)
   elif contentType == OctetStreamMediaType:
+    let consensusFork =
+      ConsensusFork.decodeString(version).valueOr:
```
tersec, line 3315: `version` here comes from:
https://github.com/status-im/nimbus-eth2/blob/c00f35a99e2be159d98353b143802f6a4a952452/beacon_chain/validators/beacon_validators.nim#L649-L687
i.e. `eth-consensus-version`

Neither https://github.com/ethereum/builder-specs/blob/v0.5.0/apis/builder/header.yaml nor https://github.com/ethereum/builder-specs/blob/v0.5.0/apis/builder/blinded_blocks.yaml appears to require it even for SSZ mode, so it might compliantly be empty in/missing from the relay's REST response.

But, again, this isn't actually a real issue, because at any given callsite, there's only one valid fork, and it does not ever depend on `eth-consensus-version`, in no edge case.

https://github.com/status-im/nimbus-eth2/blob/c00f35a99e2be159d98353b143802f6a4a952452/beacon_chain/validators/beacon_validators.nim#L641-L651
can only happen when it's Deneb, and therefore the consensus version it's expecting must, always, be Deneb.

https://github.com/status-im/nimbus-eth2/blob/c00f35a99e2be159d98353b143802f6a4a952452/beacon_chain/validators/beacon_validators.nim#L658-L668
similarly with Electra

https://github.com/status-im/nimbus-eth2/blob/c00f35a99e2be159d98353b143802f6a4a952452/beacon_chain/validators/beacon_validators.nim#L675-L687
and with Fulu

This is all known at compile-time, statically; there's no need to use `eth-consensus-version`. There isn't never was any runtime detection of fork here.

If it's supplied, and it's incorrect, then, sure, can flag on that error, but otherwise its absence shouldn't matter at all.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7045/commits/55e479bc6373c1a98dbe1492b3d339cedec10a0a

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
cheatfate: I would consider your opinion when specification will say exactly **optional** or **not required**, because right now it is:

>The active consensus version to which the data belongs. Required in response so client can deserialize returned json or ssz data more effectively.

As you can see here present `Required in response`.
https://ethereum.github.io/builder-specs/#/Builder/getHeader

cheatfate: One more proof to my opinion is present in original PR message https://github.com/ethereum/builder-specs/pull/104

> Adds `Eth-Consensus-Version` to responses, only required if response body is SSZ encoded but preferably should be added to all responses eventually.

tersec: That phrasing appears to come in via https://ethereum.github.io/builder-specs/releases/v0.5.0/builder-oapi.json via https://github.com/ethereum/builder-specs/blob/1569fc45d43e3ed2bed94ec6788c3f8fa7d2a4b2/builder-oapi.yaml#L108-L112
```
  headers:
    Eth-Consensus-Version:
      $ref: "./beacon-apis/beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version"
      schema:
        $ref: "#/components/schemas/ConsensusVersion"
```

which in turn points to https://github.com/ethereum/beacon-APIs/blob/2b1d7b5ac4756881bd29e7adacc9b7032343d981/beacon-node-oapi.yaml#L482-L489
```
  headers:
    Eth-Consensus-Version:
      description: |
        The active consensus version to which the data belongs. Required in response so client can deserialize returned json or ssz data 
        more effectively.
      required: true
      schema:
        $ref: '#/components/schemas/ConsensusVersion'
```

That's descriptive/explanatory phrasing from another non-builder API spec in a human-readable `description` field only, which sort of leaks its way into the website description (not the spec proper).

Meanwhile the builder-API spec itself is clear on this:
https://github.com/ethereum/builder-specs/blob/c7b0abad0d0b995915cde8d2e5ebcd650de5b163/apis/builder/header.yaml#L35-L41
```
  responses:
    "200":
      description: Success response.
      headers:
        Eth-Consensus-Version:
          $ref: "../../builder-oapi.yaml#/components/headers/Eth-Consensus-Version"
          required: false
```

and https://github.com/ethereum/builder-specs/blob/c7b0abad0d0b995915cde8d2e5ebcd650de5b163/apis/builder/blinded_blocks.yaml#L49-L55
```
  responses:
    "200":
      description: Success response.
      headers:
        Eth-Consensus-Version:
          $ref: "../../beacon-apis/beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version"
          required: false
```

The "required" is [flavor text](https://en.wikipedia.org/wiki/Flavor_text), and doesn't even .... necessarily mean it's required, per se. It's vague and handwavy enough even without the contradictory evidence from the builder API itself it's be questionable.

I can say, X is required for Y to happen, and it's a conditional requirement: if one wants Y to be true, X must be true, a kind of logical implication operator.

tersec: > One more proof to my opinion is present in original PR message [ethereum/builder-specs#104](https://github.com/ethereum/builder-specs/pull/104)
> 
> > Adds `Eth-Consensus-Version` to responses, only required if response body is SSZ encoded but preferably should be added to all responses eventually.

That speaks to intent of the PR creator, yes. Not what the spec actually says.

nflaig: sorry for that, I have opened a PR to clarify / formalize this on the spec https://github.com/ethereum/builder-specs/pull/118

PR #7053, opened 2025-04-04
Use ordered table for attestation packing

pedromiguelmiranda:
In order to pack electra chained attestations, nimbus scores and aggregates with an ordered fashion to ease the packing of these attestations.

However, after the scoring and aggregation, a table (not ordered) was being used, and such, not producing optimal chain aggregates after the ordered scoring.

Solution arise from the use of an ordered table.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -102,6 +102,12 @@ type
     onPhase0AttestationAdded: OnPhase0AttestationCallback
     onSingleAttestationAdded: OnSingleAttestationCallback
 
+  CandidateKey = tuple
+    ## Search key for selecting the final candidates for
+    ## electra chain aggregates.
+    hash: Eth2Digest
+    slot: Slot
+
 logScope: topics = "attpool"
 
 declareGauge attestation_pool_block_attestation_packing_time,
@@ -1005,7 +1011,7 @@ proc getElectraAttestationsForBlock*(
   # For each round, we'll look for the best attestation and add it to the result
   # then re-score the other candidates.
   var
-    candidatesPerBlock: Table[(Eth2Digest, Slot), seq[electra.Attestation]]
+    candidatesPerBlock: OrderedTable[CandidateKey, seq[electra.Attestation]]
 
   let totalCandidates = candidates.len()
   while candidates.len > 0 and candidatesPerBlock.lenu64() <
@@ -1027,13 +1033,10 @@ proc getElectraAttestationsForBlock*(
           var e2 = entry.data
           e2.index = 0
           e2
-        key = (hash_tree_root(entry2), entry.data.slot)
+        key: CandidateKey = (hash_tree_root(entry2), entry.data.slot.Slot)
         newAtt = entry[].toElectraAttestation(entry[].aggregates[j])
 
-      candidatesPerBlock.withValue(key, candidate):
-        candidate[].add newAtt
-      do:
-        candidatesPerBlock[key] = @[newAtt]
+      candidatesPerBlock.mGetOrPut(key, @[]).add(newAtt)
 
       # Update cache so that the new votes are taken into account when updating
       # the score below
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -1033,7 +1033,7 @@ proc getElectraAttestationsForBlock*(
           var e2 = entry.data
           e2.index = 0
           e2
-        key: CandidateKey = (hash_tree_root(entry2), entry.data.slot.Slot)
+        key = (hash_tree_root(entry2), entry.data.slot.Slot)
         newAtt = entry[].toElectraAttestation(entry[].aggregates[j])
 
       candidatesPerBlock.mGetOrPut(key, @[]).add(newAtt)
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -11,7 +11,8 @@ AllTests-mainnet
 + Attestations with disjoint comittee bits and equal data into single on-chain aggregate [Pr OK
 + Can add and retrieve simple electra attestations [Preset: mainnet]                         OK
 + Working with electra aggregates [Preset: mainnet]                                          OK
-+ simple add and get with electra nonzero committee [Preset: mainnet]                        OK
++ Simple add and get with electra nonzero committee [Preset: mainnet]                        OK
++ Cache coherence on chain aggregates [Preset: mainnet]                                      OK
 ```
 ## Attestation pool processing [Preset: mainnet]
 ```diff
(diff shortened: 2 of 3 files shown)


tersec:
As https://nim-lang.org/docs/tables.html#OrderedTable notes, they aren't sorted tables in any way depending on the key, they're ordered by insertion order. The insertion order is possible to game here -- it's legal to send old attestations (up to at least 32 slots old) on the network, so this is not a robust heuristic.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -1027,13 +1033,10 @@ proc getElectraAttestationsForBlock*(
...
           e2
-        key = (hash_tree_root(entry2), entry.data.slot)
+        key: CandidateKey = (hash_tree_root(entry2), entry.data.slot.Slot)
         newAtt = entry[].toElectraAttestation(entry[].aggregates[j])
 
-      candidatesPerBlock.withValue(key, candidate):
-        candidate[].add newAtt
-      do:
-        candidatesPerBlock[key] = @[newAtt]
+      candidatesPerBlock.mGetOrPut(key, @[]).add(newAtt)
```
tersec, line 1039: `mGetOrPut` is notably less efficient than `withValue`, e.g., it always generates that `@[]` default value

pedromiguelmiranda: i don t think `withValue` is implemented for `OrderedTable`

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -102,6 +102,12 @@ type
     onPhase0AttestationAdded: OnPhase0AttestationCallback
     onSingleAttestationAdded: OnSingleAttestationCallback
 
+  CandidateKey = tuple
```
tersec, line 105: `object`? Tuples are declared differently

beacon_chain/consensus_object_pools/attestation_pool.nim
tersec: Oh, you're right: https://nim-lang.org/docs/tables.html#withValue.t%2CTable%5BA%2CB%5D%2CA%2Cuntyped%2Cuntyped%2Cuntyped

pedromiguelmiranda:
> As https://nim-lang.org/docs/tables.html#OrderedTable notes, they aren't sorted tables in any way depending on the key, they're ordered by insertion order. The insertion order is possible to game here -- it's legal to send old attestations (up to at least 32 slots old) on the network, so this is not a robust heuristic.

Ordered inserts is exactly what we want. The list of candidates are ordered by score and slot as tie breaker. If a normal hash table is used when selecting attestations to fit in the block, this scoring effort will be lost and attestations with larger distance and low scoring will be used.

beacon_chain/consensus_object_pools/attestation_pool.nim
tersec: But probably tuples are still better for this purpose.  Per https://nim-lang.org/docs/manual.html#types-tuples-and-object-types one can have either `(3, 5)` or `(fieldname0: 3, fieldname1: 5)`, so if they benefit from having field labels (e.g., to access fields in an ergonomic way) one can get that. Objects are mainly useful for larger things which gets modified, but for a pair, I'm not sure why to use an `object` rather than a `tuple`.

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -1027,13 +1033,10 @@ proc getElectraAttestationsForBlock*(
           var e2 = entry.data
           e2.index = 0
           e2
-        key = (hash_tree_root(entry2), entry.data.slot)
+        key: CandidateKey = (hash_tree_root(entry2), entry.data.slot.Slot)
```
tersec: this is basically tuple syntax (which is fine), but it's not how it's declared

tersec:
> > As https://nim-lang.org/docs/tables.html#OrderedTable notes, they aren't sorted tables in any way depending on the key, they're ordered by insertion order. The insertion order is possible to game here -- it's legal to send old attestations (up to at least 32 slots old) on the network, so this is not a robust heuristic.
> 
> Ordered inserts is exactly what we want. The list of candidates are ordered by score and slot as tie breaker. If a normal hash table is used when selecting attestations to fit in the block, this scoring effort will be lost and attestations with larger distance and low scoring will be used.

Hm, yeah, it set it up for
https://github.com/status-im/nimbus-eth2/blob/eec3f170d4fea358ccd4e463a88f287b09cbbbab/beacon_chain/consensus_object_pools/attestation_pool.nim#L1063-L1076

With pre-electra, the ordering really didn't matter, because it only created exactly (128) as many attestations in the initial sorting as were relevant to begin with. But with Electra,
https://github.com/status-im/nimbus-eth2/blob/eec3f170d4fea358ccd4e463a88f287b09cbbbab/beacon_chain/consensus_object_pools/attestation_pool.nim#L1011-L1012
means it could end up with attestations which can't all fit because they're overlapping committee indices, so while the previous step collects `MAX_ATTESTATIONS_ELECTRA * MAX_COMMITTEES_PER_SLOT` aggregates (each with one committee index only), they could easily require more than `MAX_ATTESTATIONS_ELECTRA` spaces in the block to fill due to interaction with `compute_on_chain_aggregate`.

Alright.

tersec:
Do you have a benchmark comparison here of empirically, on Hoodi, attestation packing/inclusion distance 1 (especially) efficiency? With the pack-attestations-every-slot and if necessary `--subscribe-all-subnets` it should be possible to readily empirically compare the status quo and this PR.

pedromiguelmiranda:
> Do you have a benchmark comparison here of empirically, on Hoodi, attestation packing/inclusion distance 1 (especially) efficiency? With the pack-attestations-every-slot and if necessary `--subscribe-all-subnets` it should be possible to readily empirically compare the status quo and this PR.

Not yet, but started wondering about it after your findings about `mGetOrPut`. and it would be nice to check what/how this improves the packing

beacon_chain/consensus_object_pools/attestation_pool.nim
pedromiguelmiranda: fix: [d63997b](https://github.com/status-im/nimbus-eth2/pull/7053/commits/d63997b83817d896c0319d026515677004d88040)

pedromiguelmiranda:
some benchmarks (will update the comment with scripts used and instructions), and they were compared with benchmarks using the _--subscribe-all-subnets_ (s-a-s), which isn't active by default on nimbus beacon node.
Samples were taken for an hour for each combination.

minimal distance :
image

- minimal distance seen of 1, with/o s-a-s 🟢 

average validators :
image
- We can argue that _avg validators_ is not a good metric, but an improvement is seen again with/o s-a-s 🟢 

given that _avg validators_ might not be a good metric, we measured results obtained against the information on beacon chain using this metric:
(small 100 slots sample graph)
image
image

**seems like an improvement, however**
image

- The number attestations of submitted per slot with the new version is on average below from the same ones seen on the beacon chain


In conclusion:
- distance is improved, to a minimal seen of 1 and averaging 1.
- avg validators per slot improved when compared with previsous version and with the ones seen on beacon chain.
- (opinion) to improve: number of attestations per slot (if this is an improvement at all). to be further investigated.

PR #7091, opened 2025-04-22
Exit on first slashing

ahshum:


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -504,6 +504,15 @@ proc processBlsToExecutionChange*(
 
   return v
 
+proc checkAttachedValidatorSlashing(
+    self: var Eth2Processor,
+    msg: ProposerSlashing | phase0.AttesterSlashing | electra.AttesterSlashing) =
+  for idx in getValidatorIndices(msg):
+    let i = ValidatorIndex.init(idx).valueOr:
+      continue
+    if self.validatorPool[].contains(i):
```
tersec: https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/validators/validator_pool.nim only tracks validators attached to the beacon node, not those which run through the beacon node's REST API, i.e. those attached to a validator client. It's insufficient as a sole source of information.

As one alternative,
https://github.com/status-im/nimbus-eth2/blob/4879afbc93147d51d99fd4afef846f164f9d9481/beacon_chain/validators/action_tracker.nim#L63-L64
tracks both BN and VC/REST-attached validators.

As a result, `checkAttachedValidatorSlashing` isn't really an accurate name anymore, since it would also be checking non-attached (but known) validators on behalf of which this beacon node acts.

beacon_chain/nimbus_binary_common.nim
```diff
@@ -493,3 +493,9 @@ proc quitDoppelganger*() =
 
   const QuitDoppelganger = 129
   quit QuitDoppelganger
+
+proc quitSlashing*() =
+  fatal "An attached validator is slashed"
```
tersec: See discussion regarding validator pool vs action tracker; "attached" probably isn't the best word here, since it's (a) fairly Nimbus-specific terminology, but more importantly, (b) within Nimbus it specifically refers to which entity controls the private key, the BN or VC, and this feature shouldn't distinguish between them for this purpose.

beacon_chain/nimbus_binary_common.nim
```diff
@@ -493,3 +493,9 @@ proc quitDoppelganger*() =
...
+
+  const QuitSlashing = 130
```
tersec: 130 is potentially confusable with exiting the process via `SIGINT`.

Example, where I sent SIGINT to the process with ctrl-C in both `bash` and `dash`:
```
$ sleep 10
^C
$ echo $?
130
```
This is something [the shell does by calculating 128+signal number](https://www.gnu.org/software/bash/manual/html_node/Simple-Commands.html).

129 turned out not to have been a great choice, either, for basically the same reason (it's confusable with [SIGHUP](https://www.man7.org/linux/man-pages/man7/signal.7.html)), but at least we can avoid repeating that particular mistake.

The other main set of constraints to consider is [systemd's exit code handling](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Process%20Exit%20Codes). In particular, tables 7 through 10 document that exit code ranges 0-7, 64-78, and 200-245 each inclusive already denote specific semantics via one of LSB, BSD, or systemd. Combined with POSIX shells using 128+signal number, where standard Linux signal numbers [range from 0 to 31](https://www.man7.org/linux/man-pages/man7/signal.7.html), implies that 128 through 159 inclusive should be avoided too, if feasible.

This leaves  exit codes(ranges all inclusive):
- 9-63
- 79-127
- 160-199
- 246-255

As potentially available.

beacon_chain/gossip_processing/eth2_processor.nim
ahshum: A quick question, it seems that the doppelganger detection is using validator pool (https://github.com/status-im/nimbus-eth2/blob/ecabeee59291491c8312b8e499f600c970f37c70/beacon_chain/gossip_processing/eth2_processor.nim#L342C6-L342C35). Should it use `action_tracker` for remote attached validators?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -146,6 +146,12 @@ proc new*(T: type BlockProcessor,
 func hasBlocks*(self: BlockProcessor): bool =
   self.blockQueue.len() > 0
 
+# Validator status
+# ------------------------------------------------------------------------------
+
+proc hasKnownValidator*(self: var BlockProcessor, idx: ValidatorIndex): bool =
```
tersec: https://status-im.github.io/nim-style-guide/language.proc.html#functions-and-procedures-languageproc
> Prefer `func` - use `proc` when side effects cannot conveniently be avoided.

tersec: Also this doesn't appear to be used outside `block_processor`, so:
> Avoid public functions and variables (`*`) that don't make up an intended part of public API.

beacon_chain/gossip_processing/eth2_processor.nim
tersec: VC does its own doppelganger detection using the same mechanism, so it's ok for BN to only check its own (validator pool/"attached validators"):
https://github.com/status-im/nimbus-eth2/blob/77cfa783762085af0d6a9caf17dc5b2c1c15f046/beacon_chain/validator_client/doppelganger_service.nim#L26-L45

beacon_chain/nimbus_binary_common.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7091/commits/4fe8d19132402a1d1590b8bad069aca0b0f88e2b

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -120,6 +120,7 @@ type
     validatorPool*: ref ValidatorPool
     syncCommitteeMsgPool: ref SyncCommitteeMsgPool
     lightClientPool: ref LightClientPool
+    consensusManager: ref ConsensusManager
```
tersec: `BlockProcessor` already knows about `ConsensusManager`; probably better to export that symbol than add new initialization:
https://github.com/status-im/nimbus-eth2/blob/77cfa783762085af0d6a9caf17dc5b2c1c15f046/beacon_chain/gossip_processing/block_processor.nim#L97-L99

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7091/commits/b8a4f535b0fae32fa3ddaf473450efed6597ab06

beacon_chain/nimbus_binary_common.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7091/commits/4fe8d19132402a1d1590b8bad069aca0b0f88e2b

beacon_chain/nimbus_binary_common.nim
```diff
@@ -493,3 +493,9 @@ proc quitDoppelganger*() =
 
   const QuitDoppelganger = 129
   quit QuitDoppelganger
+
+proc quitSlashing*() =
+  fatal "A known validator is slashed"
+
+  const QuitSlashing = 160
```
cheatfate: According to https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html
only range of `150-199` is reserved for applications.

cheatfate: But `150-199` is not very safe because SHELL is using formula
```nim
  func exitStatusLikeShell(status: int): int =
    if WAITIFSIGNALED(cint(status)):
      # like the shell!
      128 + WAITTERMSIG(cint(status))
    else:
      WAITEXITSTATUS(cint(status))
```
it means that signal exit code reported by SHELL is in range `[128, 128 + 64]`, where 64 is current `NSIG` value in Linux.
So to not intersect with SHELL signal reports we could have range of `[192, 199]` only because exit codes `200+` are busy by systemd (https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Process%20Exit%20Codes).

So lets assume we could start from `199` (lets reserve it for doppelganger exit code which is currently 129), and move to `192`. So i would recommend to choose `198` instead of `160`.

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7091/commits/f79371e2edbeb29329bd078b00df6928448dfa9f

beacon_chain/nimbus_binary_common.nim
ahshum: Thanks for the comment. 

@tersec I started thinking that why are the occupied status codes matter to us. I think `[128, 128+64]` makes sense to me as it is for linux kernel. It seems that `1-125` is safe to use as we are different exe from e.g. systemd.

cheatfate: Oh, its not a comment from somebody, i'm asking you to change this value to `198`

ahshum: Yes I understand it is a change request but would be great to have more detail before changes. Anyway updated.

tersec (approved):

jakubgs:
Sorry, only noticed now, you missed a spot:
https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service#L61

PR #7100, opened 2025-04-23
Refactored version of BlobSidecar/DataColumnSidecar quarantine.

cheatfate:
With this new version of BlobQuarantine/DataColumnQuarantine number of issues being solved.
1. Number of API get converted from O(n) difficulty to O(1).
2. Ability to get access  to blobs by `slot` or by `root` only.
3. Elimination of hasBlobs/popBlobs (double hash table access) pattern via using `popBlobs: Opt[T]`.
4. Test suite added.

### DIFF
--- a/beacon_chain/consensus_object_pools/blob_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/blob_quarantine.nim
@@ -343,12 +343,13 @@ func hasSidecars*(
 ): bool =
   ## Function returns ``true`` if quarantine has all the blobs for block
   ## ``blck`` with block root ``blockRoot``.
+  if len(blck.message.body.blob_kzg_commitments) == 0:
+    return true
+
   let record = quarantine.roots.getOrDefault(blockRoot)
   if len(record.sidecars) == 0:
     # block root not found, record.sidecars sequence was not initialized.
     return false
-  if len(blck.message.body.blob_kzg_commitments) == 0:
-    return true
 
   if record.count < len(blck.message.body.blob_kzg_commitments):
     # Quarantine does not hold enough blob sidecars.
@@ -362,12 +363,13 @@ func hasSidecars*(
 ): bool =
   ## Function returns ``true`` if quarantine has all the columns for block
   ## ``blck`` with block root ``blockRoot``.
+  if len(blck.message.body.blob_kzg_commitments) == 0:
+    return true
+
   let record = quarantine.roots.getOrDefault(blockRoot)
   if len(record.sidecars) == 0:
     # block root not found, record.sidecars sequence was not initialized.
     return false
-  if len(blck.message.body.blob_kzg_commitments) == 0:
-    return true
 
   let
     supernode = (len(quarantine.custodyColumns) == NUMBER_OF_COLUMNS)
@@ -409,17 +411,16 @@ func popSidecars*(
   ## block ``blck``.
   ## If some of the blob sidecars are missing Opt.none() is returned.
   ## If block do not have any blob sidecars Opt.some([]) is returned.
+  let sidecarsCount = len(blck.message.body.blob_kzg_commitments)
+  if sidecarsCount == 0:
+    # Block does not have any blob sidecars.
+    quarantine.remove(blockRoot)
+    return Opt.some(default(seq[ref BlobSidecar]))
+
   let record = quarantine.roots.getOrDefault(blockRoot)
   if len(record.sidecars) == 0:
     # block root not found, record.sidecars sequence was not initialized.
     return Opt.none(seq[ref BlobSidecar])
-
-  let sidecarsCount = len(blck.message.body.blob_kzg_commitments)
-
-  if sidecarsCount == 0:
-    # Block does not have any blob sidecars.
-    quarantine.remove(blck.message.slot)
-    return Opt.some(default(seq[ref BlobSidecar]))
 
   if record.count < sidecarsCount:
     # Quarantine does not hold enough blob sidecars.
@@ -443,17 +444,16 @@ func popSidecars*(
   ## and block ``blck``.
   ## If some of the column sidecars are missing Opt.none() is returned.
   ## If block do not have any column sidecars bundledd Opt.some([]) is returned.
+  let sidecarsCount = len(blck.message.body.blob_kzg_commitments)
+  if sidecarsCount == 0:
+    # Block does not have any blob sidecars.
+    quarantine.remove(blockRoot)
+    return Opt.some(default(seq[ref DataColumnSidecar]))
+
   let record = quarantine.roots.getOrDefault(blockRoot)
   if len(record.sidecars) == 0:
     # block root not found, record.sidecars sequence was not allocated.
     return Opt.none(seq[ref DataColumnSidecar])
-
-  let sidecarsCount = len(blck.message.body.blob_kzg_commitments)
-
-  if sidecarsCount == 0:
-    # Block does not have any blob sidecars.
-    quarantine.remove(blck.message.slot)
-    return Opt.some(default(seq[ref DataColumnSidecar]))
 
   let
     supernode = (len(quarantine.custodyColumns) == NUMBER_OF_COLUMNS)
--- a/beacon_chain/consensus_object_pools/blob_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/blob_quarantine.nim
@@ -438,7 +438,8 @@ func fetchMissingSidecars*(
 
   let commitmentsCount = len(blck.message.body.blob_kzg_commitments)
   if (commitmentsCount == 0) or (record.count == commitmentsCount):
-    # Fast-path if block do not have any blobs or record holds all the blobs.
+    # Fast-path if ``blck`` does not have any blobs or if quarantine's record
+    # holds enough blobs.
     return res
 
   for bindex in 0 ..< commitmentsCount:
--- a/beacon_chain/consensus_object_pools/blob_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/blob_quarantine.nim
@@ -208,11 +208,7 @@ func put*[A, B](
     # used.
     quarantine.pruneRoot()
 
-  let
-    slot =
-      # All the blobs related to single block we could get `slot` once.
-      sidecars[0][].slot()
-    rootRecord = RootTableRecord.init(quarantine)
+  let rootRecord = RootTableRecord.init(quarantine)
 
   quarantine.roots.mgetOrPut(blockRoot, rootRecord).put(
     quarantine, sidecars)
@@ -510,6 +506,32 @@ func fetchMissingSidecars*(
         if (index == -1) or (record.sidecars[index].isNil()):
           res.add(DataColumnIdentifier(block_root: blockRoot, index: column))
   res
+
+func pruneAfterFinalization*[A, B](
+    quarantine: var SidecarQuarantine[A, B],
+    epoch: Epoch
+) =
+  let epochSlot = epoch.start_slot()
+  var
+    sidecarsCount = 0
+    rootsToRemove: seq[Eth2Digest]
+
+  for mkey, mrecord in quarantine.roots.mpairs():
+    var removeRoot = false
+    for index in 0 ..< len(mrecord.sidecars):
+      if not(isNil(mrecord.sidecars[index])) and
+         mrecord.sidecars[index][].slot < epochSlot:
+        removeRoot = true
+        # Preemptively freeing `ref` object reference.
+        mrecord.sidecars[index] = nil
+        inc(sidecarsCount)
+    if removeRoot:
+      rootsToRemove.add(mkey)
+
+  for root in rootsToRemove:
+    quarantine.roots.del(root)
+
+  dec(quarantine.sidecarsCount, sidecarsCount)
 
 template onBlobSidecarCallback*(
     quarantine: BlobQuarantine
--- a/beacon_chain/consensus_object_pools/blob_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/blob_quarantine.nim
@@ -14,7 +14,7 @@ import
   ../spec/datatypes/[deneb, electra, fulu],
   ../spec/[presets, helpers]
 
-from std/sequtils import mapIt
+from std/sequtils import mapIt, toSeq
 from std/strutils import join
 
 export results
@@ -63,19 +63,29 @@ func init*(t: typedesc[ColumnMap], columns: openArray[ColumnIndex]): ColumnMap =
 func `and`*(a, b: ColumnMap): ColumnMap =
   ColumnMap(data: [a.data[0] and b.data[0], a.data[1] and b.data[1]])
 
-func asSeq(a: ColumnMap): seq[ColumnIndex] =
+iterator items*(a: ColumnMap): ColumnIndex =
   var
-    res1: seq[ColumnIndex]
-    res2: seq[ColumnIndex]
-  for i in 0 ..< 64:
-    if a.data[0].getBit(i):
-      res1.add(ColumnIndex(i))
-    if a.data[1].getBit(i):
-      res2.add(ColumnIndex(64 + i))
-  res1 & res2
+    data0 = a.data[0]
+    data1 = a.data[1]
+
+  while data0 != 0'u64:
+    let
+      # t = data0 and -data0
+      t = data0 and ((0xFFFF_FFFF_FFFF_FFFF'u64 - data0) + 1'u64)
+      res = firstOne(data0)
+    yield ColumnIndex(res - 1)
+    data0 = data0 xor t
+
+  while data1 != 0'u64:
+    let
+      # t = data0 and -data0
+      t = data1 and ((0xFFFF_FFFF_FFFF_FFFF'u64 - data1) + 1'u64)
+      res = firstOne(data1)
+    yield ColumnIndex(64 + res - 1)
+    data1 = data1 xor t
 
 func `$`*(a: ColumnMap): string =
-  "[" & a.asSeq().mapIt($it).join(", ") & "]"
+  "[" & a.items().toSeq().mapIt($it).join(", ") & "]"
 
 func maxSidecars(maxSidecarsPerBlock: uint64): int =
   # Same limit as `MaxOrphans` in `block_quarantine`;
@@ -496,10 +506,10 @@ func fetchMissingSidecars*(
       else:
         ColumnMap.init(quarantine.custodyColumns)
     if len(record.sidecars) == 0:
-      for column in (peerMap and quarantine.custodyMap).asSeq():
+      for column in (peerMap and quarantine.custodyMap).items():
         res.add(DataColumnIdentifier(block_root: blockRoot, index: column))
     else:
-      for column in (peerMap and quarantine.custodyMap).asSeq():
+      for column in (peerMap and quarantine.custodyMap).items():
         let index = quarantine.getIndex(column)
         if (index == -1) or (record.sidecars[index].isNil()):
           res.add(DataColumnIdentifier(block_root: blockRoot, index: column))
--- a/beacon_chain/consensus_object_pools/blob_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/blob_quarantine.nim
@@ -8,39 +8,231 @@
 {.push raises: [].}
 
 import
-  std/tables,
-  ../spec/helpers
+  stew/bitops2,
+  std/[sets, tables],
+  results,
+  ../spec/datatypes/[deneb, electra, fulu],
+  ../spec/[presets, helpers]
 
 from std/sequtils import mapIt
 from std/strutils import join
 
-func maxBlobs(MAX_BLOBS_PER_BLOCK_ELECTRA: uint64): uint64 =
-  # Same limit as `MaxOrphans` in `block_quarantine`;
-  # blobs may arrive before an orphan is tagged `blobless`
-  3 * SLOTS_PER_EPOCH * MAX_BLOBS_PER_BLOCK_ELECTRA
+export results
 
 type
-  BlobQuarantine* = object
-    maxBlobs: uint64
-    blobs*:
-      OrderedTable[(Eth2Digest, BlobIndex, KzgCommitment), ref BlobSidecar]
-    onBlobSidecarCallback*: OnBlobSidecarCallback
+  ColumnMap = object
+    data: array[2, uint64]
 
-  BlobFetchRecord* = object
-    block_root*: Eth2Digest
-    indices*: seq[BlobIndex]
+  RootTableRecord[A] = object
+    sidecars: seq[ref A]
+    count: int
 
-  OnBlobSidecarCallback = proc(
-      data: BlobSidecarInfoObject) {.gcsafe, raises: [].}
+  SlotRecord[A] = object
+    blockRoot: Eth2Digest
+    sidecar: ref A
+
+  SlotTableRecord[A] = seq[seq[SlotRecord[A]]]
+
+  SidecarQuarantine[A, B] = object
+    maxSidecarsCount: int
+    maxSidecarsPerBlockCount: int
+    sidecarsCount: int
+    custodyColumns: seq[ColumnIndex]
+    custodyMap: ColumnMap
+    roots: Table[Eth2Digest, RootTableRecord[A]]
+    slots: Table[Slot, SlotTableRecord[A]]
+    usage: OrderedSet[Slot]
+    indexMap: seq[int]
+    onSidecarCallback*: B
+
+  OnBlobSidecarCallback* = proc(
+    data: BlobSidecarInfoObject) {.gcsafe, raises: [].}
+  OnDataColumnSidecarCallback* = proc(
+    data: DataColumnSidecar) {.gcsafe, raises: [].}
+
+  BlobQuarantine* =
+    SidecarQuarantine[BlobSidecar, OnBlobSidecarCallback]
+  ColumnQuarantine* =
+    SidecarQuarantine[DataColumnSidecar, OnDataColumnSidecarCallback]
+
+func init(t: typedesc[ColumnMap], columns: openArray[ColumnIndex]): ColumnMap =
+  var res: ColumnMap
+  for column in columns:
+    let
+      index = int(uint64(column) shr 6)
+      offset = int(uint64(column) and 0x3F'u64)
+    res.data[index].setBit(offset)
+  res
+
+# func `xor`(a, b: ColumnMap): ColumnMap =
+#   ColumnMap(data: [a.data[0] xor b.data[0], a.data[1] xor b.data[1]])
+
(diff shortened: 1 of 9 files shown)


beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -49,66 +241,387 @@ func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
...
+    T: typedesc[ColumnQuarantine],
+    cfg: RuntimeConfig,
+    custodyColumns: openArray[ColumnIndex],
+    onBlobSidecarCallback: OnDataColumnSidecarCallback
+): ColumnQuarantine =
+  let size = maxSidecars(cfg.MAX_BLOBS_PER_BLOCK_ELECTRA)
+
+  var indexMap = newSeq[int](NUMBER_OF_COLUMNS)
+  for i in 0 ..< len(indexMap):
+    indexMap[i] = -1
```
tersec: In theory [repeat](https://nim-lang.org/docs/sequtils.html#repeat%2CT%2CNatural) does precisely this, but it does it also in the same way, with just `newSeq`, so first zero-fills it then fills it with whatever value one wants. It would be more efficient to use `newSeqUninit` and then the loop you have here.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -8,39 +8,231 @@
...
+func init(t: typedesc[ColumnMap], columns: openArray[ColumnIndex]): ColumnMap =
+  var res: ColumnMap
+  for column in columns:
+    let
+      index = int(uint64(column) shr 6)
+      offset = int(uint64(column) and 0x3F'u64)
+    res.data[index].setBit(offset)
+  res
+
+# func `xor`(a, b: ColumnMap): ColumnMap =
```
tersec: There are several of these commented-out sections (also `or`, `$`, `genFuluSignedBeaconBlock` at least), preferable not to merge such disabled/commented-out code sections.

cheatfate: This is draft PR, i want to show some of the operations which could be needed so it is commented because not used. I will try to eliminate all the commented sections before merge.

beacon_chain/consensus_object_pools/blob_quarantine.nim
cheatfate: Source of `repeat` is
```nim
proc repeat*[T](x: T, n: Natural): seq[T] =
  ## Returns a new sequence with the item `x` repeated `n` times.
  ## `n` must be a non-negative number (zero or more).
  ##
  runnableExamples:
    let
      total = repeat(5, 3)
    assert total == @[5, 5, 5]

  result = newSeq[T](n)
  for i in 0 ..< n:
    result[i] = x
```
which means that `newSeqUninit()` is much more efficient.

tersec: yes, I agree, as indicated. though this is really a bug with the Nim stdlib.

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/7100/commits/70f0ecc911607feb32697f44d00a6a79fc637a89

beacon_chain/sync/request_manager.nim
```diff
@@ -471,28 +466,32 @@ proc getMissingBlobs(rman: RequestManager): seq[BlobIdentifier] =
...
-             commitments=len(forkyBlck.message.body.blob_kzg_commitments)
-          ready.add(blobless.root)
+          if commitmentsCount == 0:
+            # this is a programming error should it occur.
+            warn "missing blob handler found blobless block with all blobs",
+                 blk = blobless.root,
+                 commitments = len(forkyBlck.message.body.blob_kzg_commitments)
+            ready.add(blobless.root)
+          else:
+            # This should not be happen too...
```
tersec: "This shouldn't happen either" or "This should not happen either"

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -434,9 +434,8 @@ proc validateBlobSidecar*(
   #
   # It would be good to fix this more properly, but this has come up often on
   # Pectra devnet-6.
-  if blobQuarantine[].hasBlob(
-      block_header.slot, block_header.proposer_index, blob_sidecar.index,
-      blob_sidecar.kzg_commitment):
+  if blobQuarantine[].hasSidecar(block_header.slot, block_header.proposer_index,
```
tersec: Does this still handle what
- https://github.com/status-im/nimbus-eth2/pull/6912

and the comment block just above it about devnet-6 describes?

I'm wondering because it removes precisely the `kzg_commitment` which that PR added. It's true that it now matches (again, and as it was before, more closely) the gossip spec, but that didn't work particularly well in these conditions. I've asked about it on the Eth R&D Discord a couple of weeks ago, not responses, but working > not working, and I'm still not sure how a 100% truly spec-compliant implementation was supposed to handle these, because other clients did (and so did Nimbus after #6912).

To clarify that PR's description a bit, there were regular double-proposer, slashable events (some of which got slashed, but that takes at least a couple of slots typically to be noticed, it's not instant). What would happen is, Nimbus would be going fine, following the chain, until one of these double proposals came up. Each had, independently, some set of blobs:
- separately valid block 1, with a set of valid blobs; and
- separately valid block 2, with a set of valid blobs (different than the first set, created by a different node).

Both of these proposals shared a slot and proposer index, because they were the same proposer. Indeed, the signatures were all valid too, because, well, they were both legitimately running that private key.

But what would happen is,
- if block 1's blobs came in, and block 1 came in, and block 1 turned out to be the one the chain followed, then, great, the ignore condition here worked fine (WLOG extend to block 2); but 
- if the blobs came in interleaved, this wasn't always true, and, crucially, this gossip condition as spec-written prevented Nimbus's gossip from being able to collect all the blobs from block 1.

Maybe other clients did/do this by having a very efficient request manager-equivalent, I'm not sure. But without something, either receiving via gossip or req/resp, Nimbus just got stuck until a suitable reorg happened, typically dozens of slots later, because this gossip condition prevented it from seeing all the blobs corresponding to either block.

Also, it would be basically random chance which, if asked by req/resp, of the two different (or more, but the devnet-6 case was two slashable blocks at a time) sets of blobs would be returned, so it seemed to sometimes have to retry this. All of this took enough time Nimbus lost the chain basically deterministically every time this slashable double-proposal situation came up.

I don't see anything obviously corresponding to this in the tests, either, to show this is otherwise addressed.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -49,66 +278,361 @@ func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
...
+    quarantine: BlobQuarantine,
+    blockRoot: Eth2Digest,
     blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock |
-          fulu.SignedBeaconBlock): bool =
-    # Having a fulu SignedBeaconBlock is incorrect atm, but
-    # shall be fixed once data columns are rebased to fulu
-  for idx, kzg_commitment in blck.message.body.blob_kzg_commitments:
-    if (blck.root, BlobIndex idx, kzg_commitment) notin quarantine.blobs:
-      return false
+          fulu.SignedBeaconBlock
```
tersec, line 269: Is it valid to ask for/quarantine blob sidecars in Fulu?

cheatfate: Its not valid, but this was not addressed in code base, so when removing this i will get compilation errors inside validation and processing logic.

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/1c91d9d48fdf369cec4773a3077bc741db1442b9

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/c9b960a8d529107aaffd148bc66bc39874be5003

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/f30a8f5afe24d90a036c4e851ac893b1e1859350

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: Specifically,
```
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(888, 52) template/generic instantiation of `async` from here
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(900, 21) template/generic instantiation of `withBlck` from here
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(901, 15) template/generic instantiation of `storeBlock` from here
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(862, 11) template/generic instantiation of `setResult` from here
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(840, 13) template/generic instantiation of `withBlck` from here
nimbus-eth2/beacon_chain/gossip_processing/block_processor.nim(855, 34) Error: type mismatch
Expression: popSidecars(self.blobQuarantine[], quarantined.fuluData.root,
            quarantined.fuluData)
  [1] self.blobQuarantine[]: BlobQuarantine
  [2] quarantined.fuluData.root: Eth2Digest
  [3] quarantined.fuluData: SignedBeaconBlock

Expected one of (first mismatch at [position]):
[1] func popSidecars(quarantine: var ColumnQuarantine; blck: fulu.SignedBeaconBlock): Opt[
    seq[ref DataColumnSidecar]]
[1] func popSidecars(quarantine: var ColumnQuarantine; blockRoot: Eth2Digest;
                 blck: fulu.SignedBeaconBlock): Opt[seq[ref DataColumnSidecar]]
[2] func popSidecars(quarantine: var BlobQuarantine; blck: deneb.SignedBeaconBlock |
    electra.SignedBeaconBlock |
    fulu.SignedBeaconBlock): Opt[seq[ref BlobSidecar]]
[3] func popSidecars(quarantine: var BlobQuarantine; blockRoot: Eth2Digest;
                 blck: deneb.SignedBeaconBlock | electra.SignedBeaconBlock): Opt[
    seq[ref BlobSidecar]]
```
because
https://github.com/status-im/nimbus-eth2/blob/3fc3bc5083d700cf80976ec373495d851432c65b/beacon_chain/gossip_processing/block_processor.nim#L840-L860
doesn't check for an upper bound on the relevant fork.

It's fine for this PR to ignore that, because as you allude to, cleaning that up is a basically separate task, but at some point should happen.

tersec:
looks like maybe `make test` wasn't rerun:
```
[2025-05-12T17:51:56.808Z] + git diff --exit-code --ignore-submodules=all
[2025-05-12T17:51:56.808Z] diff --git a/AllTests-mainnet.md b/AllTests-mainnet.md
[2025-05-12T17:51:56.808Z] index 1805a5a5..9960f7b2 100644
[2025-05-12T17:51:56.808Z] --- a/AllTests-mainnet.md
[2025-05-12T17:51:56.808Z] +++ b/AllTests-mainnet.md
[2025-05-12T17:51:56.808Z] @@ -111,6 +111,7 @@ AllTests-mainnet
[2025-05-12T17:51:56.808Z]  ```diff
[2025-05-12T17:51:56.808Z]  + overfill protection test                                                                   OK
[2025-05-12T17:51:56.808Z]  + popSidecars()/hasSidecars() return []/true on block without blobs                          OK
[2025-05-12T17:51:56.808Z] ++ pruneAfterFinalization() test                                                              OK
[2025-05-12T17:51:56.808Z]  + put()/fetchMissingSidecars/remove test                                                     OK
[2025-05-12T17:51:56.808Z]  + put()/hasSidecar(index, slot, proposer_index)/remove() test                                OK
[2025-05-12T17:51:56.808Z]  + put(sidecar)/put([sidecars])/hasSidecars/popSidecars/remove() test                         OK
[2025-05-12T17:51:56.808Z] @@ -159,6 +160,7 @@ AllTests-mainnet
[2025-05-12T17:51:56.808Z]  + ColumnMap test                                                                             OK
[2025-05-12T17:51:56.808Z]  + overfill protection test                                                                   OK
[2025-05-12T17:51:56.808Z]  + popSidecars()/hasSidecars() return []/true on block without columns                        OK
[2025-05-12T17:51:56.808Z] ++ pruneAfterFinalization() test                                                              OK
[2025-05-12T17:51:56.808Z]  + put()/fetchMissingSidecars/remove test [node]                                              OK
[2025-05-12T17:51:56.808Z]  + put()/fetchMissingSidecars/remove test [supernode]                                         OK
[2025-05-12T17:51:56.808Z]  + put()/hasSidecar(index, slot, proposer_index)/remove() test                                OK
script returned exit code 1
```

beacon_chain/sync/request_manager.nim
```diff
@@ -471,28 +466,32 @@ proc getMissingBlobs(rman: RequestManager): seq[BlobIdentifier] =
...
-             blk=blobless.root,
-             commitments=len(forkyBlck.message.body.blob_kzg_commitments)
-          for idx in missing.indices:
-            let id = BlobIdentifier(block_root: blobless.root, index: idx)
-            if id notin fetches:
-              fetches.add(id)
+        let
+          commitmentsCount = len(forkyBlck.message.body.blob_kzg_commitments)
+          missing =
+            rman.blobQuarantine[].fetchMissingSidecars(blobless.root, forkyBlck)
```
tersec, line 472: This appears to be the only caller of the `BlobQuarantine`, and it appears to be the only thing which fills in `idents: HashSet[BlobIdentifier]`. It might be more direct for at least the `BlobQuarantine` `fetchMissingSidecars()` to directly build and return the `HashSet[BlobIdentifier]`, rather than return a `seq[BlobIdentifier]` which is converted to said `HashSet` before it's ever used.

The main question would be is if somehow constructing the empty `HashSet[BlobIdentifier]` is both common enough and more expensive enough to warrant the current two-stage approach as an optimization, but it seems unlikely.

beacon_chain/sync/request_manager.nim
```diff
@@ -471,28 +466,32 @@ proc getMissingBlobs(rman: RequestManager): seq[BlobIdentifier] =
...
+            warn "quarantine missing blobs, but missing indices is empty",
+                 blk = blobless.root,
+                 commitments = len(forkyBlck.message.body.blob_kzg_commitments)
 
   for root in ready:
     let blobless = rman.quarantine[].popBlobless(root).valueOr:
       continue
     discard rman.blockVerifier(blobless, false)
-  fetches
+  idents.items.toSeq()
```
tersec: The current flow has `fetchMissingSidecars()` return a `seq`, which is accumulated into `HashSet`, then iterated back out into a `seq` again. Is the purpose here to deduplicate, and if so, are duplicates possible?

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -8,39 +8,196 @@
...
+    let
+      index = int(uint64(column) shr 6)
+      offset = int(uint64(column) and 0x3F'u64)
+    res.data[index].setBit(offset)
+  res
+
+func `and`*(a, b: ColumnMap): ColumnMap =
+  ColumnMap(data: [a.data[0] and b.data[0], a.data[1] and b.data[1]])
+
+func asSeq(a: ColumnMap): seq[ColumnIndex] =
```
tersec: This has 3 callers, `$` and
```nim

    if len(record.sidecars) == 0:
      for column in (peerMap and quarantine.custodyMap).asSeq():
        res.add(DataColumnIdentifier(block_root: blockRoot, index: column))
    else:
      for column in (peerMap and quarantine.custodyMap).asSeq():
        let index = quarantine.getIndex(column)
        if (index == -1) or (record.sidecars[index].isNil()):
          res.add(DataColumnIdentifier(block_root: blockRoot, index: column))
```
of which the two main ones only need `iterator`-level access, and `$` can use `toSeq` or similar, but at least `asSeq` (or a different name if it's more appropriate; I'm indifferent on that point) can by default avoid all the `seq`s it builds up in ordinary circumstances.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -8,39 +8,196 @@
...
+  if len(quarantine.usage) == 0:
+    return
+  var oldestRoot: Eth2Digest
+  for blockRoot in quarantine.usage:
+    oldestRoot = blockRoot
+    break
+  quarantine.remove(oldestRoot)
+
+func getIndex(quarantine: BlobQuarantine, index: BlobIndex): int =
+  doAssert(index < lenu64(quarantine.indexMap))
```
tersec: isn't this `doAssert` redundant, i.e. `seq` access should raise an `IndexDefect` on any out of bounds access?

See also the `ColumnQuarantine` `getIndex` `doAssert`.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -8,39 +8,196 @@
...
+    # This also therefore automatically garbage-collects otherwise valid garbage
+    # blobs which are correctly signed, point to either correct block roots or a
+    # block root which isn't ever seen, and then are for any reason simply never
+    # used.
+    quarantine.pruneRoot()
+
+  let rootRecord = RootTableRecord.init(quarantine)
+  quarantine.roots.mgetOrPut(blockRoot, rootRecord).put(
+    quarantine, [sidecar])
+  discard quarantine.usage.containsOrIncl(blockRoot)
```
tersec: If discarding the return value anyway, why not just `incl` instead of `containsOrIncl`?

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -49,66 +206,384 @@ func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
...
-    (block_root, blobSidecar.index, blobSidecar.kzg_commitment), blobSidecar)
-
-func hasBlob*(
+    quarantine.pruneRoot()
+
+  let rootRecord = RootTableRecord.init(quarantine)
+
+  quarantine.roots.mgetOrPut(blockRoot, rootRecord).put(
+    quarantine, sidecars)
+  discard quarantine.usage.containsOrIncl(blockRoot)
```
tersec: Just `incl`, no `discard`?

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -49,66 +206,384 @@ func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
...
+    fulu.SignedBeaconBlock
+): seq[BlobIdentifier] =
+  ## Function returns sequence of BlobIdentifiers for blobs which are missing
+  ## for block root ``blockRoot`` and block ``blck``.
+  var res: seq[BlobIdentifier]
+  let record = quarantine.roots.getOrDefault(blockRoot)
+
+  let commitmentsCount = len(blck.message.body.blob_kzg_commitments)
+  if (commitmentsCount == 0) or (record.count == commitmentsCount):
+    # Fast-path if block do not have any blobs or record holds all the blobs.
```
tersec: "if the block does not have any ..."

(could use `blck`, i.e. "if blck does not have any ..." if one wants the similar article-less thing as "or record holds all"...)

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -8,39 +8,196 @@
...
+    # blobs which are correctly signed, point to either correct block roots or a
+    # block root which isn't ever seen, and then are for any reason simply never
+    # used.
+    quarantine.pruneRoot()
+
+  let rootRecord = RootTableRecord.init(quarantine)
+  quarantine.roots.mgetOrPut(blockRoot, rootRecord).put(
+    quarantine, [sidecar])
+  discard quarantine.usage.containsOrIncl(blockRoot)
+  inc(quarantine.sidecarsCount)
```
tersec: If this `put` function is called twice in a row with the same (duplicate) `sidecar`, does it increment `quarantine.sidecarsCount` twice? Should it?

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -49,66 +206,384 @@ func put*(quarantine: var BlobQuarantine, blobSidecar: ref BlobSidecar) =
...
-
-func hasBlob*(
+    quarantine.pruneRoot()
+
+  let rootRecord = RootTableRecord.init(quarantine)
+
+  quarantine.roots.mgetOrPut(blockRoot, rootRecord).put(
+    quarantine, sidecars)
+  discard quarantine.usage.containsOrIncl(blockRoot)
+  inc(quarantine.sidecarsCount, len(sidecars))
```
tersec: Same question about calling this with duplicate `sidecars` and how often `quarantine.sidecarsCount` increments

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/982e62a16019c6ef6ff1d24b240be076be4e4fe0

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/982e62a16019c6ef6ff1d24b240be076be4e4fe0

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/982e62a16019c6ef6ff1d24b240be076be4e4fe0

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/982e62a16019c6ef6ff1d24b240be076be4e4fe0

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7100/commits/982e62a16019c6ef6ff1d24b240be076be4e4fe0

tersec:
https://github.com/status-im/nimbus-eth2/pull/7100/commits/982e62a16019c6ef6ff1d24b240be076be4e4fe0 adds tests but doesn't update the test summaries.

https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Flinux%2Fx86_64%2Fmain/detail/PR-7100/14/pipeline
```
[2025-05-13T11:54:52.523Z] + git diff --exit-code --ignore-submodules=all
[2025-05-13T11:54:52.523Z] diff --git a/AllTests-mainnet.md b/AllTests-mainnet.md
[2025-05-13T11:54:52.523Z] index 1805a5a5..dd1b338d 100644
[2025-05-13T11:54:52.523Z] --- a/AllTests-mainnet.md
[2025-05-13T11:54:52.523Z] +++ b/AllTests-mainnet.md
[2025-05-13T11:54:52.523Z] @@ -111,6 +111,8 @@ AllTests-mainnet
[2025-05-13T11:54:52.523Z]  ```diff
[2025-05-13T11:54:52.523Z]  + overfill protection test                                                                   OK
[2025-05-13T11:54:52.523Z]  + popSidecars()/hasSidecars() return []/true on block without blobs                          OK
[2025-05-13T11:54:52.523Z] ++ pruneAfterFinalization() test                                                              OK
[2025-05-13T11:54:52.523Z] ++ put() duplicate items should not affect counters                                           OK
[2025-05-13T11:54:52.523Z]  + put()/fetchMissingSidecars/remove test                                                     OK
[2025-05-13T11:54:52.523Z]  + put()/hasSidecar(index, slot, proposer_index)/remove() test                                OK
[2025-05-13T11:54:52.523Z]  + put(sidecar)/put([sidecars])/hasSidecars/popSidecars/remove() test                         OK
[2025-05-13T11:54:52.523Z] @@ -159,6 +161,8 @@ AllTests-mainnet
[2025-05-13T11:54:52.523Z]  + ColumnMap test                                                                             OK
[2025-05-13T11:54:52.523Z]  + overfill protection test                                                                   OK
[2025-05-13T11:54:52.523Z]  + popSidecars()/hasSidecars() return []/true on block without columns                        OK
[2025-05-13T11:54:52.523Z] ++ pruneAfterFinalization() test                                                              OK
[2025-05-13T11:54:52.523Z] ++ put() duplicate items should not affect counters                                           OK
[2025-05-13T11:54:52.523Z]  + put()/fetchMissingSidecars/remove test [node]                                              OK
[2025-05-13T11:54:52.523Z]  + put()/fetchMissingSidecars/remove test [supernode]                                         OK
[2025-05-13T11:54:52.523Z]  + put()/hasSidecar(index, slot, proposer_index)/remove() test                                OK
script returned exit code 1
```

beacon_chain/sync/request_manager.nim
cheatfate: I think you right, duplicates should not be possible.

beacon_chain/sync/request_manager.nim
cheatfate: There is no reason to return `HashSet[BlobIdentifier]` because duplicates are impossible.

cheatfate: Addressed in https://github.com/status-im/nimbus-eth2/pull/7100/commits/995ebf417c6d52c4ec18477b360200bafda78296

beacon_chain/sync/request_manager.nim
cheatfate: Addressed in https://github.com/status-im/nimbus-eth2/pull/7100/commits/995ebf417c6d52c4ec18477b360200bafda78296

beacon_chain/consensus_object_pools/blob_quarantine.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/7100/commits/cda8613fd0b3fa13694737593afdf613a9620350

beacon_chain/consensus_object_pools/blob_quarantine.nim
cheatfate: Addressed in https://github.com/status-im/nimbus-eth2/pull/7100/commits/d7d6b5304b1f4bc1fa590a3cea880ee246877654

PR #7108, opened 2025-04-25
Create database table for blob data sidecar quarantine

etan-status:
Add low-level interface to persist blob data sidecars by block root + index. Also support deleting junk that's been finalized and deletion of all data associated with a block root.

Can pass `db.getQuarantineDB()` into the quarantine to expose the new DB tables while hiding APIs that deal with trusted data, and then use the new `beacon_chain_db_quarantine.nim` APIs to store / fetch / delete.


etan-status:
The quarantine DB works as is for blobs (and also for blocks in earlier commit), but it seems that a custom solution based on SQLite implicit rowid is preferred over explicit rowid through primary key.

beacon_chain/beacon_chain_db_quarantine.nim
```diff
@@ -0,0 +1,211 @@
...
+
+    electraDataSidecar: DataSidecarStore
+      ## Proposer signature verified data blob sidecars.
+    fuluDataSidecar: DataSidecarStore
+      ## Proposer signature verified data column sidecars.
+
+template tableName(sidecar: typedesc[ForkyDataSidecar]): string =
+  when sidecar is deneb.BlobSidecar:
+    "electra_sidecars_quarantine"
+  else:
```
tersec: In all of these, it would be more future-proof to adopt the structure used elsewhere in the CL of:
```nim
  when sidecar is deneb.BlobSidecar:
    ...
  elif sidecar is fulu.DataColumnSidecar:
    ...
  else:
    static: doAssert false/static: raiseAssert/{.error.}/...
```
because the next DAS approach after PeerDAS likely won't be based around `fulu.DataColumnSidecar` and it'd be good to be able to catch all such issues at compile-time

beacon_chain/beacon_chain_db_quarantine.nim
```diff
@@ -0,0 +1,211 @@
...
+    discard statement.exec do (res: int64):
+      recordCount = res
+  recordCount
+
+proc initQuarantineDB*(
+    backend: SqStoreRef,
+): KvResult[QuarantineDB] =
+  let
+    electraDataSidecar =
+      ? backend.initDataSidecarStore(tableName(deneb.BlobSidecar))
```
tersec, line 208: Maybe comment explicitly in some places that this means there's no compatibility guarantees, forward or backward, around this schema.

beacon_chain/beacon_chain_db_quarantine.nim
```diff
@@ -0,0 +1,211 @@
...
+    "electra_sidecars_quarantine"
+  else:
+    "fulu_sidecars_quarantine"
+
+proc initDataSidecarStore(
+    backend: SqStoreRef,
+    name: string
+): KvResult[DataSidecarStore] =
+  if name == "":
+    return ok(DataSidecarStore())
```
tersec: how does this early-exit path actually function, i.e. when something tries to use it?

beacon_chain/beacon_chain_db_quarantine.nim
```diff
@@ -0,0 +1,211 @@
...
+  # Status libraries
+  chronicles,
+  eth/db/kvstore_sqlite3,
+  # Beacon chain internals
+  spec/helpers,
+  ./db_utils
+
+# Without this export compilation fails with error
+# vendor\nim-chronicles\chronicles.nim(352, 21) Error: undeclared identifier: 'activeChroniclesStream'
+# It actually does not needed, because chronicles is not used in this file,
```
tersec: - It actually isn't needed; or
- It actually is not needed

beacon_chain/beacon_chain_db_quarantine.nim
```diff
@@ -0,0 +1,211 @@
...
+# but because decodeSZSSZ() generic and uses chronicles generic expansion
```
tersec: decodeSZSSZ() is generic

beacon_chain/beacon_chain_db_quarantine.nim
tersec: Or alternatively just be very explicit, indeed, about the lack of durability/persistence between instances, so any Fulu quarantine tables made now don't constrain the future.

beacon_chain/beacon_chain_db_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7108/commits/3d7c3afa0eb1bfdc8d565e42db3726e8d95050ac

beacon_chain/beacon_chain_db_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7108/commits/3d7c3afa0eb1bfdc8d565e42db3726e8d95050ac

beacon_chain/beacon_chain_db_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7108/commits/3d7c3afa0eb1bfdc8d565e42db3726e8d95050ac

beacon_chain/beacon_chain_db_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7108/commits/3d7c3afa0eb1bfdc8d565e42db3726e8d95050ac

beacon_chain/beacon_chain_db_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7108/commits/3d7c3afa0eb1bfdc8d565e42db3726e8d95050ac

PR #7117, opened 2025-04-29
replace:`DataColumnIdentifier` -> `DataColumnsByRootIdentifier` while root requesting

agnxsh:
https://github.com/ethereum/consensus-specs/pull/4284

### DIFF
--- a/beacon_chain/spec/datatypes/fulu.nim
+++ b/beacon_chain/spec/datatypes/fulu.nim
@@ -95,6 +95,7 @@ type
 
 type
   DataColumn* = List[KzgCell, Limit(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
+  ColumnIndices* = List[ColumnIndex, Limit(NUMBER_OF_COLUMNS)]
 
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.3/specs/fulu/das-core.md#datacolumnsidecar
   DataColumnSidecar* = object
@@ -115,7 +116,7 @@ type
 
   DataColumnsByRootIdentifier* = object
     block_root*: Eth2Digest
-    indices*: List[ColumnIndex, Limit NUMBER_OF_COLUMNS]
+    indices*: ColumnIndices
 
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/das-core.md#matrixentry
   MatrixEntry* = object
--- a/beacon_chain/spec/datatypes/fulu.nim
+++ b/beacon_chain/spec/datatypes/fulu.nim
@@ -113,6 +113,10 @@ type
     block_root*: Eth2Digest
     index*: ColumnIndex
 
+  DataColumnsByRootIdentifier* = object
+    block_root*: Eth2Digest
+    indices*: List[ColumnIndex, Limit NUMBER_OF_COLUMNS]
+
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/das-core.md#matrixentry
   MatrixEntry* = object
     cell*: Cell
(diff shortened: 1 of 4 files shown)


beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -95,6 +95,7 @@ type
 
 type
   DataColumn* = List[KzgCell, Limit(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
+  ColumnIndices* = List[ColumnIndex, Limit(NUMBER_OF_COLUMNS)]
```
cheatfate: Please rename this type to DataColumnIndices.

agnxsh: Ok

beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -113,6 +114,10 @@ type
     block_root*: Eth2Digest
     index*: ColumnIndex
 
+  DataColumnsByRootIdentifier* = object
```
tersec, line 118: This appears to come from https://github.com/ethereum/consensus-specs/pull/4284 which, sure, part of the spec now, so it's fine, but should be linked to, to indicate that it's a spec entity and not some randomly changeable internal Nimbus-only construction: https://github.com/ethereum/consensus-specs/blob/b8b5fbb8d16f52d42a716fa93289062fe2124c7c/specs/fulu/p2p-interface.md#datacolumnsbyrootidentifier

No spec release yet, but at least that's a stable link for now.

beacon_chain/sync/sync_protocol.nim
```diff
@@ -393,7 +395,7 @@ p2pProtocol BeaconSync(version = 1,
   # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/p2p-interface.md#datacolumnsidecarsbyroot-v1
```
tersec: this URL is presumably outdated and doesn't reflect the new implementation.

beacon_chain/spec/datatypes/fulu.nim
agnxsh: no stable release, which is why i waited but yeah, can add it

tersec: https://github.com/status-im/nimbus-eth2/pull/7117/commits/3848e271257a4fac9ed8d94446de0bf4d72a8a59

beacon_chain/spec/datatypes/fulu.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7117/commits/3848e271257a4fac9ed8d94446de0bf4d72a8a59

beacon_chain/sync/sync_protocol.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7117/commits/3848e271257a4fac9ed8d94446de0bf4d72a8a59

agnxsh: https://github.com/status-im/nimbus-eth2/pull/7117/commits/95c3d67f16b2c0d059bb43264d6f7cbd1386d21f

beacon_chain/sync/request_manager.nim
```diff
@@ -162,36 +166,24 @@ func checkResponseSubset(idList: seq[BlobIdentifier],
...
-  ## Clients MAY limit the number of blocks and sidecars in the response.
-  ## https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.3/specs/fulu/p2p-interface.md#datacolumnsidecarsbyroot-v1
-  for col in columns:
-    if binarySearch(idList, col, cmpSidecarIdentifier) == -1:
+func checkColumnResponse(idList: seq[DataColumnsByRootIdentifier],
+                    columns: openArray[ref DataColumnSidecar]): bool =
+  for colresp in columns:
+    let block_root =
+      hash_tree_root(colresp[].signed_block_header.message)
+    if block_root notin idList.mapIt(it.block_root):
```
tersec, line 174: It doesn't have to be in this PR, but in general, this is pointlessly allocating the `mapIt`-resulting `seq`.

It only consumes this as an iterator-equivalent and at no point does it need nor should it allocate all of `idList.mapIt(it.block_root)` (rather, it checks against `foo1.block_root`, `foo2.block_root`, in sequence).

tersec: Also, it means that it's calculated a bunch of these ahead of the item it might detect a match

PR #7126, opened 2025-05-05
Experimental splitview support that was used on Holesky / Goerli

etan-status:
This branch contains the experimental changes to better support splitview scenarios, and was used to successfully sync Goerli in its final month, and Holesky during the Pectra upgrade. The changes include:

- `--debug-invalidate-block-root` flag to treat a block root as INVALID; needed because certain EL forget that a payload is invalid and replies SYNCING/ACCEPTED when Nimbus is restarted, tricking Nimbus into optimistically importing the known-invalid block
- `--debug-split-views-merge` flag to enable a branch discovery module that pulls in all advertised histories from all peers if a chain stall is detected. This feature will no longer work with PeerDAS, because not every peer will be a supernode anymore.
- `--debug-propose-stale` flag to keep proposing when a chain stall is detected, even when the head is behind by a long timespan. This feature allocates a temporary extra state to incrementally advance to present epoch for doing a timely proposal, but if `epochRefState` state could be merged with `clearanceState` this would no longer be the needed.


etan-status:
The branch is available as `ethpandaops/nimbus-eth2:splitview` on Docker.

As the simple branch pulling strategy from this PR will no longer be viable with Fusaka, there's no point in maintaining this proof-of-concept past Electra - the Docker image can still be used as a safety net during the Pectra transition (or simply build from the PR branch which is based on Nimbus v25.4.1). 

Further, the individual elements are fairly "cherry pickable" if needed (from the diff, not from the commit history).

etan-status:
`--debug-invalidate-block-root`: https://github.com/status-im/nimbus-eth2/pull/7128

PR #7127, opened 2025-05-05
implement validator custody with column refill mechanism

agnxsh:


beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -487,9 +487,11 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
   ok()
 
 # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.3/specs/fulu/das-core.md#validator-custody
-func get_validators_custody_requirement*(cfg: RuntimeConfig, state: fulu.BeaconState,
+proc get_validators_custody_requirement*(cfg: RuntimeConfig,
```
tersec: Why change from `func` to `proc`?

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -487,9 +487,11 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
...
+                                         hstate: ForkyHashedBeaconState,
                                          validator_indices: openArray[ValidatorIndex]):
                                          uint64 =
+  let state = hstate.data
```
tersec: This is an expensive copy, since Nim copies by value. If you want just the more convenient name, use
```nim
template state: untyped = hstate.data
```

tersec: Other options:
- keep this function's params the same, and call it with `foo.data` rather than `foo` at the callsite; and
- in the
```nim
    total_node_balance += state.balances[index]
```
line, instead directly use
```nim
    total_node_balance += hstate.data.balances[index]
```

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+    supernode: supernode,
+    getLocalHeadSlot: getLocalHeadSlotCb,
+    older_column_set: older_column_set,
+    getBeaconTime: getBeaconTime,
+    inhibit: inhibit,
+    dataColumnQuarantine: dataColumnQuarantine)
+
+proc detectNewValidatorCustody(vcus: ValidatorCustodyRef): seq[ColumnIndex] =
+  let
+    headState =  vcus.dag.headState
```
tersec: https://github.com/status-im/nimbus-eth2/blob/77cfa783762085af0d6a9caf17dc5b2c1c15f046/beacon_chain/consensus_object_pools/block_pools_types.nim#L194-L196
this also is a quite expensive state copy, for apparently no particular need except to refer to it more easily.
Also see `template` tricks on this one.

tersec: Or even better it's only actually referred to once, in `withState(headState)` -- so can just use `withState(vcus.dag.headState)`

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+  let dataColumnRefillEpoch = (slot.epoch -
+                              vcus.dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS - 1)
+  if slot.is_epoch() and dataColumnRefillEpoch >= vcus.dag.cfg.FULU_FORK_EPOCH:
+    var blocks: array[SLOTS_PER_EPOCH.int, BlockId]
+    let startIndex = vcus.dag.getBlockRange(
+      dataColumnRefillEpoch.start_slot, blocks.toOpenArray(0, SLOTS_PER_EPOCH - 1))
+    for i in startIndex..<SLOTS_PER_EPOCH:
+      let blck = vcus.dag.getForkedBlock(blocks[int(i)]).valueOr: continue
+      withBlck(blck):
+        when typeof(forkyBlck).kind < ConsensusFork.Fulu: continue
```
tersec: Should this ever happen? it's much cheaper to detect this before the block is ever fetched from the database, based on the slot number + fork schedule.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -487,9 +487,11 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
   ok()
 
 # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.3/specs/fulu/das-core.md#validator-custody
```
tersec: Seems to have moved to https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.5/specs/fulu/validator.md#validator-custody

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+import
+  ../validators/action_tracker,
+  ../spec/[forks, network, helpers, peerdas_helpers],
+  ../networking/eth2_network,
+  ../consensus_object_pools/blockchain_dag,
+  ../consensus_object_pools/block_dag,
+  ../consensus_object_pools/data_column_quarantine,
+  "."/[request_manager,  sync_manager, sync_protocol]
+
+from std/algorithm import binarySearch, sort
```
tersec: It doesn't seem to call `binarySearch`?

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+import chronos, chronicles
+import ssz_serialization/types
+import
+  ../validators/action_tracker,
+  ../spec/[forks, network, helpers, peerdas_helpers],
+  ../networking/eth2_network,
+  ../consensus_object_pools/blockchain_dag,
+  ../consensus_object_pools/block_dag,
+  ../consensus_object_pools/data_column_quarantine,
+  "."/[request_manager,  sync_manager, sync_protocol]
```
tersec: Why the extra space before `sync_manager`?

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+proc detectNewValidatorCustody(vcus: ValidatorCustodyRef): seq[ColumnIndex] =
+  let
+    headState =  vcus.dag.headState
+  var
+    res: seq[ColumnIndex]
+    diff_set: HashSet[ColumnIndex]
+  withState(headState):
+    when consensusFork >= ConsensusFork.Fulu:
+      var validator_indices =
+        get_active_validator_indices(forkyState.data, vcus.network.getBeaconTime().slotOrZero.epoch)
```
tersec: This is an expensive scan across all validators in the set -- usually one would try to use some kind of cache. Also, it's already basically implemented via
https://github.com/status-im/nimbus-eth2/blob/77cfa783762085af0d6a9caf17dc5b2c1c15f046/beacon_chain/spec/beaconstate.nim#L911-L925 by way of https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.5/specs/phase0/beacon-chain.md#get_total_active_balance

So basically the operation is `get_total_active_balance` then do some processing to get the validator custody set, and Nimbus can do that significantly more efficiently than what you're doing here.

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+      vcus.dag.cfg.resolve_columns_from_custody_groups(
+        remoteNodeId,
+        max(vcus.dag.cfg.SAMPLES_PER_SLOT.uint64,
+            remoteCustodyGroupCount))
+  for request_item in vcus.requested_columns:
+    var colIds: seq[ColumnIndex]
+    for cindex in request_item.indices:
+      let lookup = DataColumnIdentifier(block_root: request_item.block_root,
+                                        index: cindex)
+      if lookup notin vcus.global_refill_list and cindex in remoteCustodyColumns:
```
tersec, line 152: How often is this `cindex in RemoteCustodyColumns` expected to occur? Each one is an `O(n)` lookup, in a `for cindex in request_items.indices` loop, due to
https://github.com/status-im/nimbus-eth2/blob/77cfa783762085af0d6a9caf17dc5b2c1c15f046/beacon_chain/spec/peerdas_helpers.nim#L73-L75

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,246 @@
...
+      vcus.dataColumnQuarantine[].custody_columns =
+        newer_columns
+
+      # check which custody set is larger
+      if newer_column_set.len > vcus.older_column_set.len:
+        diff_set = newer_column_set.difference(vcus.older_column_set)
+      vcus.newer_column_set = newer_column_set
+
+  for i in diff_set.items():
+    res.add(i)
```
tersec: This is the first code in this function which touches `res` except for the declaration, and `res` turns out to be basically `toSeq(diff_set)`. So no need for the whole explicit `for i in diff_set.items()` loop, explicit `res` variable, et cetera.

It's fundamentally the same set of operations, but cleaner/clearer.

tersec:
https://github.com/status-im/nimbus-eth2/actions/runs/14842398481/job/41668253800?pr=7127
```
2025-05-05T18:21:02.9784725Z Hint: mm: refc; threads: on; opt: speed; options: -d:release
2025-05-05T18:21:02.9787538Z 378176 lines; 215.223s; 10.377GiB peakmem; proj: /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/tests/all_tests.nim; out: /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/nimcache/release/all_tests/all_tests.json [SuccessX]
2025-05-05T18:24:28.9640792Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/iterators.nim: In function ‘_ZN9confutils8loadImplE8typeDescIN4conf14BeaconNodeConfEE3seqI6stringE6string6string4bool4bool3refIN9anonymous16SecondarySourcesEE4procIN4conf14BeaconNodeConfE3refIN9anonymous16SecondarySourcesEEE6string.constprop’:
2025-05-05T18:24:28.9649194Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-confutils/confutils.nim:910:15: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
2025-05-05T18:24:28.9786592Z   910 | proc loadImpl[C, SecondarySources](
2025-05-05T18:24:28.9787035Z       |               ^
2025-05-05T18:25:06.9981057Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/iterators.nim: In function ‘_ZN9confutils8loadImplE8typeDescIN4conf14BeaconNodeConfEE3seqI6stringE6string6string4bool4bool3refIN14initBeaconNode16SecondarySourcesEE4procIN4conf14BeaconNodeConfE3refIN14initBeaconNode16SecondarySourcesEEE6string.constprop’:
2025-05-05T18:25:06.9985700Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-confutils/confutils.nim:910:15: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
2025-05-05T18:25:06.9986981Z   910 | proc loadImpl[C, SecondarySources](
2025-05-05T18:25:06.9987472Z       |               ^
2025-05-05T18:25:26.2738160Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nimbus-build-system/vendor/Nim/lib/system/iterators_1.nim: In function ‘_ZN17validator_custody25detectNewValidatorCustodyE3refIN17validator_custody16ValidatorCustodyEE’:
2025-05-05T18:25:26.2740847Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/sync/validator_custody.nim:67:15: error: stack usage is 11312192 bytes [-Werror=stack-usage=]
2025-05-05T18:25:26.2742260Z    67 | proc detectNewValidatorCustody(vcus: ValidatorCustodyRef): seq[ColumnIndex] =
2025-05-05T18:25:26.2742997Z       |               ^
2025-05-05T18:26:16.7631562Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim: In function ‘_ZN22test_gossip_transition18runTestX60gensym3_E6string6string’:
2025-05-05T18:26:16.7736235Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim:1101:15: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
2025-05-05T18:26:16.7737632Z  1101 |   proc runTest(suiteName, testName: string): TestStatus {.raises: [], gensym.} =
2025-05-05T18:26:16.7738172Z       |               ^
2025-05-05T18:26:18.2169452Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim:1101:15: note: variable tracking size limit exceeded
2025-05-05T18:26:19.0940375Z lto1: some warnings being treated as errors
2025-05-05T18:26:19.1458084Z make[2]: *** [/tmp/ccScWUVj.mk:305: /tmp/ccDZizXd.ltrans101.ltrans.o] Error 1
2025-05-05T18:26:19.1459044Z make[2]: *** Waiting for unfinished jobs....
2025-05-05T18:26:45.1057573Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim: In function ‘_ZN19test_toblindedblock21runTestX60gensym1069_E6string6string’:
2025-05-05T18:26:45.1134244Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim:1101:15: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
2025-05-05T18:26:45.1135690Z  1101 |   proc runTest(suiteName, testName: string): TestStatus {.raises: [], gensym.} =
2025-05-05T18:26:45.1136273Z       |               ^
2025-05-05T18:27:20.9297450Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim: In function ‘_ZN19test_toblindedblock21runTestX60gensym2038_E6string6string’:
2025-05-05T18:27:20.9300358Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim:1101:15: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
2025-05-05T18:27:22.1837786Z lto-wrapper: fatal error: make returned 2 exit status
2025-05-05T18:27:22.1838783Z compilation terminated.
2025-05-05T18:27:22.3302505Z /usr/bin/ld: error: lto-wrapper failed
2025-05-05T18:27:22.3436028Z collect2: error: ld returned 1 exit status
2025-05-05T18:27:22.3468012Z make[1]: *** [nimcache/release/all_tests/all_tests.makefile:3760: build] Error 1
```

beacon_chain/spec/peerdas_helpers.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/spec/peerdas_helpers.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/spec/peerdas_helpers.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/3dc502b7940014e9ef1a44a5017483cf1e79f75c

beacon_chain/spec/peerdas_helpers.nim
tersec: Well, funny timing, since this exchange, https://github.com/ethereum/consensus-specs/blob/v1.5.0/specs/fulu/validator.md#validator-custody

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,242 @@
...
+    supernode: supernode,
+    getLocalHeadSlot: getLocalHeadSlotCb,
+    older_column_set: older_column_set,
+    getBeaconTime: getBeaconTime,
+    inhibit: inhibit,
+    dataColumnQuarantine: dataColumnQuarantine)
+
+proc detectNewValidatorCustody(vcus: ValidatorCustodyRef): seq[ColumnIndex] =
+  var
+    cache = StateCache()
```
tersec: How/in what contexts is `detectNewValidatorCustody` going to be called? While this is, in isolation, at least no worse than the previous approach, and helps integrate with the caching infrastructure, would the future callers of ` `detectNewValidatorCustody` have access to an already-filled-in-`StateCache`?

beacon_chain/sync/validator_custody.nim
tersec: > [3dc502b](https://github.com/status-im/nimbus-eth2/commit/3dc502b7940014e9ef1a44a5017483cf1e79f75c)

How does this commit affect `checkIntersectingCustody()`?

beacon_chain/spec/peerdas_helpers.nim
agnxsh: lol

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,242 @@
...
+      let vcustody =
+        vcus.dag.cfg.get_validators_custody_requirement(forkyState, total_node_balance)
+
+      let
+        newer_columns =
+          vcus.dag.cfg.resolve_columns_from_custody_groups(
+            vcus.network.nodeId,
+            max(vcus.dag.cfg.SAMPLES_PER_SLOT.uint64,
+            vcustody))
+        newer_column_set = newer_columns.toHashSet()
```
tersec: I'm not sure why
https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/beacon_chain/spec/peerdas_helpers.nim#L73-L85
doesn't just always construct in-place with `incl` or similar and then return a `HashSet[ColumnIndex]` rather than this `seq[ColumnIndex]` which as far as I can tell, every single user of `resolve_columns_from_custody_groups()` converts to a `HashSet` before using or storing:

https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/beacon_chain/nimbus_beacon_node.nim#L425-L428
calls only the wrapper function, to get its `HashSet[ColumnIndex]`.

https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/beacon_chain/sync/request_manager.nim#L65-L68
stores a `HashSet[ColumnIndex]`, not a `seq[ColumnIndex]`, which it gets from the `nimbus_beacon_node` code.

https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/beacon_chain/sync/request_manager.nim#L352-L359
uses the `resolve_column_sets_from_custody_groups()` wrapper again.

And this code here doesn't use `resolve_column_sets_from_custody_groups()` but does the same thing. So it seems like in general `resolve_columns_from_custody_groups()` should natively return a `HashSet` and avoid the redundant copying, allocating, conversions, et cetera.

At that point, one wouldn't need a separate
https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/beacon_chain/spec/peerdas_helpers.nim#L87-L91
at all, because all those `HashSet[ColumnIndex]` users would just be able to call `resolve_columns_from_custody_groups()` directly.

tersec: https://github.com/status-im/nimbus-eth2/blob/566257eadd130ab627c002ce1ee6360b80b43797/beacon_chain/validators/message_router.nim#L191-L200
doesn't use the `HashSet` but probably should, and it wouldn't particularly be more expensive if it were constructed natively rather than converted from the `seq`.

agnxsh: you are looking at the wrong branch, i think there are significant use cases of `resolve_columns_from_custody_groups`, and hashset conversion is only applicable to most syncers/refillers/backfillers where lookup is more frequent

tersec: > you are looking at the wrong branch

Yes, this is true -- those were mostly from the `stable` branch, so e.g., `resolve_column_sets_from_custody_groups` isn't present in either `column-syncer` or `vcus`, so that part isn't relevant/accurate.

That said:
> i think there are significant use cases of `resolve_columns_from_custody_groups`, and hashset conversion is only applicable to most syncers/refillers/backfillers where lookup is more frequent

Looking again specifically at the current state of the `column-syncer` and `vcus` branches:

In `column-syncer`:

Doesn't use `HashSet`, probably should, only cares interface-wise that `in` works:
https://github.com/status-im/nimbus-eth2/blob/566257eadd130ab627c002ce1ee6360b80b43797/beacon_chain/validators/message_router.nim#L189-L200

Immediately converts to `HashSet` via `toHashSet`:
https://github.com/status-im/nimbus-eth2/blob/566257eadd130ab627c002ce1ee6360b80b43797/beacon_chain/sync/request_manager.nim#L345-L355

Immediately converts to `HashSet` via `toHashSet` for `RequestManager` initialization purposes, but also uses `dataColumnQuarantine[].custody_columns` with the `seq`:
https://github.com/status-im/nimbus-eth2/blob/566257eadd130ab627c002ce1ee6360b80b43797/beacon_chain/nimbus_beacon_node.nim#L414-L422

https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim from `column-syncer` itself uses `custody_columns` field in a few ways:
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L87
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L99
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L116
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L155
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L160
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim#L169

The field is marked for export, and probably shouldn't be, because that's mainly there to let it be initialized from `nimbus_beacon_node`, but nothing reads from it later outside that module.

Anyway, it does two things, iterates, in a way it's unclear that order matters (i.e. `HashSet` arbitrary ordering is ok) and checks length, either of which is fine in a `HashSet`. And this `HashSet` is constructed anyway, a line later in `nimbus_beacon_node`.

That's all the usage in `column-syncer`. On to `vcus` (this PR):

`message_router`'s usage is identical:
https://github.com/status-im/nimbus-eth2/blob/3dc502b7940014e9ef1a44a5017483cf1e79f75c/beacon_chain/validators/message_router.nim#L189-L200

`request_manager`'s usage is identical:
https://github.com/status-im/nimbus-eth2/blob/3dc502b7940014e9ef1a44a5017483cf1e79f75c/beacon_chain/sync/request_manager.nim#L345-L355

`nimbus_beacon_node`'s usage is identical:
https://github.com/status-im/nimbus-eth2/blob/3dc502b7940014e9ef1a44a5017483cf1e79f75c/beacon_chain/nimbus_beacon_node.nim#L426-L434

The data column quarantine usage is the same: https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/consensus_object_pools/data_column_quarantine.nim so, iteration in a way where exact order shouldn't matter and length-checking, both easy and reasonably efficient with the `HashSet` that's constructed anyway.

But `vcus` adds `validator_custody` usage.

Immediately converts via `toHashSet`:
https://github.com/status-im/nimbus-eth2/blob/3dc502b7940014e9ef1a44a5017483cf1e79f75c/beacon_chain/sync/validator_custody.nim#L79-L85

Doesn't use a `HashSet`, but probably should:
https://github.com/status-im/nimbus-eth2/blob/3dc502b7940014e9ef1a44a5017483cf1e79f75c/beacon_chain/sync/validator_custody.nim#L133-L146

tersec: Deliberately putting this in a separate comment:
https://github.com/status-im/nimbus-eth2/blob/083ed100bab70e07331f0c3e9692b1b5d3eed412/beacon_chain/spec/peerdas_helpers.nim#L119-L131

could just as easily directly construct the `HashSet` that most of those use case either already use or should use, and never construct the `seq` at all. Then no repeatedly converting to it, no issues with some places using the `seq` (and incurring O(n) search times) and some places not, et cetera.

The data column quarantine itself doesn't really benefit, but it's not hurt either that I can tell.

tersec: https://nim-lang.org/docs/packedsets.html provides another alternative, I'm not that familiar with its performance tradeoffs.

beacon_chain/sync/validator_custody.nim
```diff
@@ -0,0 +1,243 @@
...
+    PARALLEL_REFILL_REQUESTS = 32
+
+type
+  InhibitFn = proc: bool {.gcsafe, raises: [].}
+
+  ValidatorCustody* = object
+    network*: Eth2Node
+    dag*: ChainDAGRef
+    supernode*: bool
+    cache*: StateCache
```
tersec: Does the linear chain segment assumption apply here?
https://github.com/status-im/nimbus-eth2/blob/d2f23389ab657608261be9f4dcef1009d67fbc2d/beacon_chain/spec/datatypes/base.nim#L395-L401

If `ValidatorCustody` can be before/after a reorg, it's not necessarily safe to use `StateCache` directly.

That said, if you're careful to only use this on a sequence of finalized checkpoints, that would qualify, but that assumption needs to be documented and checked if so.

But it gets back to, what is the calling context in which `detectNewValidatorCustody()` is called, and might that calling context have access to a filled-in `StateCache` already?

beacon_chain/sync/validator_custody.nim
agnxsh: > detect this before the block is ever fetched from the database

the only way to make this cheaper is to see if the earliest data column in the database is against a Fulu block only then it makes sense to skip this check, imo, otherwise everytime we pull a finalized block from the DB, it's _safe_ to check the fork, also say it can be a situation just a couple epochs after Fulu transition that on wanting block range i get `EEEEEEFFFFFFF`, i want to refill only against the last Fulu blocks, and wanna continue for the first few Es, note that Es and Fs should ideally way more in number, this is just for an example

beacon_chain/spec/peerdas_helpers.nim
agnxsh: this has been updated, resolving this comment

beacon_chain/sync/validator_custody.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7127/commits/ae8ed8896395948b63566efb4621dc0a86b65029

this should be resolved here, although there may be different ways to get this done as well, we can merge this for now.

beacon_chain/sync/validator_custody.nim
agnxsh: this can be tackled in the `column-syncer` branch with a subsequent rebase

tersec: ok

beacon_chain/sync/validator_custody.nim
tersec: The answer is was that it's, yes, linear.

beacon_chain/sync/validator_custody.nim
agnxsh: remote custody columns is now a hashset after https://github.com/status-im/nimbus-eth2/pull/7210

beacon_chain/sync/validator_custody.nim
agnxsh: > because all those HashSet[ColumnIndex] users would just be able to call resolve_columns_from_custody_groups() directly.

custody columns now directly returns a HashSet. https://github.com/status-im/nimbus-eth2/pull/7210

beacon_chain/sync/validator_custody.nim
```diff
@@ -98,14 +98,21 @@ proc detectNewValidatorCustody(vcus: ValidatorCustodyRef, cache: var StateCache)
...
   let dataColumnRefillEpoch = (slot.epoch -
                               vcus.dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS - 1)
   if slot.is_epoch() and dataColumnRefillEpoch >= vcus.dag.cfg.FULU_FORK_EPOCH:
-    var blocks: array[SLOTS_PER_EPOCH.int, BlockId]
+    var blocks = newSeq[BlockId](slot.epoch().int)
     let startIndex = vcus.dag.getBlockRange(
-      dataColumnRefillEpoch.start_slot, blocks.toOpenArray(0, SLOTS_PER_EPOCH - 1))
-    for i in startIndex..<SLOTS_PER_EPOCH:
+      dataColumnRefillEpoch.start_slot, blocks.toOpenArray(0, slot.epoch().int - 1))
+    for i in startIndex..<slot.epoch().int:
```
tersec, line 116: practically save but not that great to do this because (a) in theory 32/64-bit differences and (b) introduces potential `Defect` due to `int` conversion

tersec: I see `i` also has to be `int(i)` a couple lines later, worth looking at the types here holistically.

agnxsh: it's the same technique used in pruneSidecars

tersec: It's not really great there either, but sure, it's consistent

PR #7139, opened 2025-05-07
Fix VC does not send eth-consensys-version header with publishBlindedBlockV2 call.

cheatfate:


cheatfate:
Closing this PR in favor of https://github.com/status-im/nimbus-eth2/pull/7140

PR #7152, opened 2025-05-12
Add peerDAS reconstruction metrics

KatyaRyazantseva:
Add peerDAS reconstruction metrics from the [beacon metrics specs](https://github.com/ethereum/beacon-metrics/pull/14):
- beacon_data_availability_reconstructed_columns_total
- beacon_data_availability_reconstruction_time_seconds


KatyaRyazantseva:
There is also a recovery in [eth2_processor.nim](https://github.com/status-im/nimbus-eth2/blob/column-metrics/beacon_chain/gossip_processing/eth2_processor.nim#L401). Should I declare metrics globally? Where is the better place for it?

agnxsh:
we are not reconstructing as of yet, we intend to reconstruct while serving columns, however.

Issue #6754, opened 2024-12-05
Node should exit with a unique code on first slashing


jakubgs:
**Is your feature request related to a problem? Please describe.**
Currently if a slashing occurs for one of the attached validators the node continues to run. I consider this behavior as inviting further slashings.

**Describe the solution you'd like**
The same way doppelganger detection returns a unique `129` exit code, so should the node the node exit - possibly with another code, like 130 - on first detected slashing. This is done in order to prevent any additional slashings.

**Describe alternatives you've considered**
None.

**Additional context**
The recent wave of slashings on mainnet that spanned a period of 10 hours shows that even bigger operators can fail to detect a slashing even in their setup and fail to prevent additional slashing if no engineer is available to stop the nodes.

I see no benefit in the node continuing to run after the first slashing, only great risk.

cheatfate:
How beacon node should detect the slashing?
Does it need to exit process if it going to send slashing to the network or when it received new slashings in the block?

tersec:
Multiple possible approaches:
- monitor the [`attester_slashing`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.1/specs/phase0/p2p-interface.md#attester_slashing) and [`proposer_slashing`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.1/specs/phase0/p2p-interface.md#proposer_slashing) topics directly; and/or
- monitor the `proposer_slashings` and `attester_slashings` [fields of BeaconBlockBody](https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.1/specs/deneb/beacon-chain.md#extended-containers); and/or
- hook into [`slash_validator()`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.1/specs/bellatrix/beacon-chain.md#modified-slash_validator).

Hooking into `slash_validator` directly isn't great because it lives in the state trandition function and architecturally doesn't have direct access to the information it would need to to do this, but mentioned/acknowledged for completeness.

[...]

PR #7165, opened 2025-05-15
Prevent systemd restart on slashing exit code

ahshum:
- https://github.com/status-im/nimbus-eth2/pull/7091#issuecomment-2882785318

For information, `RestartPreventExitStatus` is separated by spaces. https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#RestartPreventExitStatus=

### DIFF
--- a/scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service
+++ b/scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service
@@ -57,8 +57,8 @@ WorkingDirectory=/var/lib/nimbus
 TimeoutSec=1200
 Restart=always
 
-# Don't restart when Doppelganger detection has been activated
-RestartPreventExitStatus=129
+# Don't restart when Doppelganger or slashing detection has been activated
+RestartPreventExitStatus=129 198
 
 ExecStart=/usr/bin/nimbus_beacon_node \
   --network=${NETWORK} \


jakubgs (approved):

jakubgs:
One question tho, does this also apply to validator client?
https://github.com/status-im/nimbus-eth2/blob/5624e5dfbf8d830c87e72efa49c082668e99fda4/scripts/package_src/nimbus_validator_client/image/lib/systemd/system/nimbus_validator_client.service#L52

tersec:
> One question tho, does this also apply to validator client?
> 
> https://github.com/status-im/nimbus-eth2/blob/5624e5dfbf8d830c87e72efa49c082668e99fda4/scripts/package_src/nimbus_validator_client/image/lib/systemd/system/nimbus_validator_client.service#L52

No, the validator client is not directly involved in this one, unlike doppelganger. This is mainly a BN show.

jakubgs:
But validator client also has a doppelganger detection flag. Does this functionality not apply to it as well? Am I misunderstanding something or was it simply not implemented for VC yet?

ahshum:
I was thinking the same but may be I am missing something about the difference between doppelganger and slashing. Would be great to know more.

jakubgs:
Well, if VC can also have this functionality then it should, because we are making more and more use of VCs in our setup, and we are planning to introduce it to our LIDO fleet, so it's kinda important.

tersec:
> No, the validator client is not directly involved in this one, unlike doppelganger. This is mainly a BN show.

Means precisely what it says and not more. It does not imply, for example, that VC _validators_ can't take advantage of this functionality, just that it's not the Nimbus validator client delivering it:
> Well, if VC can also have this functionality then it should

The actual detection algorithm here lives 100% in the BN, that's why the VC doesn't have it, but because the implementation in the BN uses `ActionTracker`, it covers all the validators, VC and BN.

While https://ethereum.github.io/beacon-APIs/#/Events/eventstream does have `proposer_slashing` and `attester_slashing` events, and the VC can subscribe to them, https://github.com/status-im/nimbus-eth2/issues/6754#issuecomment-2614469428 describes one issue, which the VC would face just as much as the BN using those:
> This works fine for gossip messages, and is slightly quicker to pick up slashings before they get bundled into blocks, but doesn't work on backfilled or potentially reorged blocks which might not be processed by gossip at all.

So it's a question of whether the additional marginal complexity for the VC to also, separately, implement this is worthwhile. For a setup running both Nimbus VC and BN, I'd say definitely no. For a setup running a Nimbus VC with a non-Nimbus BN, there's at least an argument, but it's not what:
> because we are making more and more use of VCs in our setup, and we are planning to introduce it to our LIDO fleet, so it's kinda important.

is describing.

jakubgs:
>The actual detection algorithm here lives 100% in the BN, that's why the VC doesn't have it, but because the implementation in the BN uses ActionTracker, it covers all the validators, VC and BN.

I'm not sure how that is supposed to work. Lets take for example a situation in which we have two VCs, each with two BNs.
If there's a single validator overlap between the two VCs and that validator gets slashed, what happens?

- All four BNs stop?
- Just the BN involved in given action that triggered the slashing stops?
- Some other option I'm not seeing?

tersec:
> > The actual detection algorithm here lives 100% in the BN, that's why the VC doesn't have it, but because the implementation in the BN uses ActionTracker, it covers all the validators, VC and BN.
> 
> I'm not sure how that is supposed to work. Lets take for example a situation in which we have two VCs, each with two BNs. If there's a single validator overlap between the two VCs and that validator gets slashed, what happens?
> 
>     * All four BNs stop?
> 
>     * Just the BN involved in given action that triggered the slashing stops?
> 
>     * Some other option I'm not seeing?

All four BNs stop.

Which is in this scenario overkill, but any feature of this sort is intrinsically heuristic, but the premise is that slashing is such a bad outcome for an Ethereum node that it's better to stop and compromise liveness to prevent it.

In any one-slashable-validator-overlap case, including this one, the optimal approach is not to stop on slashing at all, because once that slashing happens, it's done, and if one actually knew that the rest of the validators didn't have that problem, one could avoid any liveness issues by not stopping the node at all.

The entire point of this heuristic is that mass slashings appear to occur not when there's accidental overlap of a single validator, but in fleet setups where there are dozens, hundreds, or thousands of validators per node, and the implicit inference is that a single detected overlap signals a configuration error which the user should correct regardless, and might be much more widespread than the single detected validator. It's only a guess, it only ever was a guess, it doesn't matter whether it's only the BN's validators, the BN's validators and a VC's validator's, et cetera.

But when there's even a two-validator overlap that was not detected by doppelganger (e.g., because both nodes started at precisely the same time), that guess pays off, because then while one of those validators get slashed, it wouldn't turn into a mass slashing event.

There's nothing particularly unique about this 2 VC/4 BN case, imagine an even simpler case, 1 BN with 1 slashable validator and an arbitrary number of VC with some non-slashable validators attached to them, well, that BN shuts down in any plausible version of this feature. If it knew this, overkill, but the point it that it isn't and can't in any automatic, reliable way.

There are some interesting edge cases especially around "public" BNs, where one should look at how problematic for someone to burn currently approximately 1 ETH to attach to a public BN to compromise its liveness. While this is one somewhat intriguing argument splitting out the VC side, there are much simpler attacks against unprotected liveness attacks "public" BNs, mainly attacking some of the expensive available APIs (e.g., [`getStateV2`](https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2), especially the JSON variation, and if that's too easy to cache, go for preferably JSON [`getStateValidators`](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidators) or [`postStateValidators`](https://ethereum.github.io/beacon-APIs/#/Beacon/postStateValidators)), so practically "public" BNs need some kind of filtering to be useful anyway.

Furthermore, there's a tradeoff in this approach: a Nimbus BN with non-Nimbus VC, which might or might not do this on its own. Maybe worth caring about, maybe not, but that applies to any/all of this, Nimbus functions perfectly fine without this feature, it's not some completely obviously correct/better approach, it's a subjective value judgement about the likelihood of node operators screwing up their deployment configurations and what it's worth to protect against that compared with potential liveness costs.

There are some options here:
- back out this feature altogether. It's currently not particularly invasive or extensive, and hasn't shipped. It only protects against operator error/misconfiguation to begin with, it's pretty external to any code Nimbus functionality, VC or BN;
- keep it as is, it's a bit overzealous at times, but that's the nature of any version of this feature, and it catches all the cases at the BN;
- add it to the VC, but _this still does not remove the heuristic nature of this issue, it just shifts it around_, there's no exact approach here by definition;
- and some others, but if it starts getting to the point of adding random settings people mostly won't discover to work around a edge cases of a feature which doesn't actually need to exist because they're workarounds for operator error, I rapidly start thinking it's not worthwhile to have any of this.

But to be clear, this feature does conceptually work as shipped, by focusing on the BN. It just makes a slightly different selectiveness tradeoff, but (a) in your one-overlap scenario that's arguing for having none of this implemented, and (b) even less contrived scenarios can then be made up to show that it will sometimes compromise liveness more than strictly necessary. But the operating theory of any such thing requires this, so the mere existence of such a scenario can't be an argument against it.

tersec:
I want especially to add that the "each with two BNs" in your scenario doesn't matter at all, they legitimately both need to shut down. It doesn't matter which "triggered the slashing", that's a conceptual error. If they're holding a validator which got slashed, it doesn't matter whether they think they triggered it or not, nor should it.

Even if one wants to imagine a VC-based version (which has some of the noted gaps -- it's relatively expensive for it to monitor certain things which are cheap for a BN, and this monitoring would be on all the time, even in the 99% of time slashings aren't happening, so focusing on the inexpensive `proposer_slashing` and `attester_slashing` topics, it can't practically require the same approach the BN can use), both VCs still would shut down, and then it doesn't matter whether the BNs do or not.

jakubgs:
>All four BNs stop.
>
>Which is in this scenario overkill, but any feature of this sort is intrinsically heuristic, but the premise is that slashing is such a bad outcome for an Ethereum node that it's better to stop and compromise liveness to prevent it.

I don't think it's overkill, I think this is the correct behavior considering the impact of continuing to cause slashings.

>In any one-slashable-validator-overlap case, including this one, the optimal approach is not to stop on slashing at all, because once that slashing happens, it's done, and if one actually knew that the rest of the validators didn't have that problem, one could avoid any liveness issues by not stopping the node at all.

Lol obviously, I was just simplifying the scenario for the sake of a simpler explanation :D.

>I want especially to add that the "each with two BNs" in your scenario doesn't matter at all, they legitimately both need to shut down. It doesn't matter which "triggered the slashing", that's a conceptual error. If they're holding a validator which got slashed, it doesn't matter whether they think they triggered it or not, nor should it.

Yes, that's what I was expecting and hoping for, and that's what I got. Another victory in a never ending string of victories.

>add it to the VC, but this still does not remove the heuristic nature of this issue, it just shifts it around, there's no exact approach here by definition

I don't think this is necessary, the current behavior is sufficient to avoid mass slashing events and that's good enough for me.

Thanks for the explanation.

PR #7187, opened 2025-05-21
integrate new column quarantine into the column flow

agnxsh:


beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -328,6 +328,19 @@ func hasSidecars*(
   ## ``blck`` with block root ``blockRoot``.
   hasSidecars(quarantine, blck.root, blck)
 
+func peekSidecars*(
```
cheatfate, line 331: Could you please explain why you need this function, because in my scenario:
Code looks for sidecars and call for `popSidecars`, if there is enough sidecars for your node or supernode it will return `Opt.some()`, but if its not enough it will return you `Opt.none()`. So if you received `Opt.none()` it means that you can do nothing at this point, but if it returns `Opt.some()` and you are supernode - you are trying to rebuild... 
I'm not sure if rebuilding process could fail or not, but the idea is to keep single owner for sidecars.

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -516,6 +529,74 @@ func fetchMissingSidecars*(
           res.add(DataColumnIdentifier(block_root: blockRoot, index: column))
   res
 
+func fetchMissingColumnsByRoot*(
```
cheatfate, line 532: I think there is no need to keep `fetchMissingColumns()` function, or its better to move your code inside and adopt tests for it.

agnxsh: Okay

beacon_chain/consensus_object_pools/blob_quarantine.nim
agnxsh: so as a supernode, we can check DA at 65 and make block addition decisions based on that, but before hitting the database, at some time during that flow we will have to rebuild the rest of the columns, if i rebuild inside the block processor, yes, i won't really tweak the quarantine. but it would be an essential step for column serveability

agnxsh: so do you think, rebuilding rest of the columns before persisting column in the database is reasonable? maybe yes, because that might lighten the storage load if you attempt to take backups of the column quarantine, technically, we can do everything with 65 columns that we can with 128 columns, but the converse isn't true. additionally, if we ever choose to serve columns from the backed up quarantine itself, then we might have to rebuild the rest before serving to a remote peer.

beacon_chain/consensus_object_pools/blob_quarantine.nim
agnxsh: > I think there is no need to keep `fetchMissingColumns()` function, or its better to move your code inside and adopt tests for it.

keeping it for now, in case they change specs in the meantime, can stay a couple weeks, will adopt tests for the new method

beacon_chain/consensus_object_pools/blob_quarantine.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7187/commits/edfc956ab20e44f76aa8dc737a6b8aa098e53d41

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -33,8 +33,8 @@ type
   SidecarQuarantine[A, B] = object
     maxSidecarsCount: int
     maxSidecarsPerBlockCount: int
-    sidecarsCount: int
-    custodyColumns: seq[ColumnIndex]
+    sidecarsCount*: int
```
cheatfate: This variables should not be exported, because both of them are for internal use.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -374,41 +377,27 @@ proc processDataColumnSidecar*(
...
-        if not self.dataColumnQuarantine[].supernode:
-          if self.dataColumnQuarantine[].hasExactDataColumns(forkyBlck, self.dag.cfg):
-            let gathered_columns =
-              self.dataColumnQuarantine[].gatherDataColumns(forkyBlck.root)
-            for gdc in gathered_columns:
-              self.dataColumnQuarantine[].put(newClone(gdc))
+        let cres =
+          self.dataColumnQuarantine[].popSidecars(block_root, forkyBlck)
+        if cres.isSome():
+          if cres.get().lenu64 >= (self.dag.cfg.NUMBER_OF_COLUMNS div 2):
```
cheatfate: Why comparison is not strict? Is 64 columns is enough to reconstruct?

beacon_chain/rpc/rest_config_api.nim
```diff
@@ -299,7 +299,7 @@ proc installConfigApiHandlers*(router: var RestRouter, node: BeaconNode) =
           VALIDATOR_CUSTODY_REQUIREMENT:
             Base10.toString(cfg.VALIDATOR_CUSTODY_REQUIREMENT.uint64),
           BALANCE_PER_ADDITIONAL_CUSTODY_GROUP:
-            Base10.toString(cfg.BALANCE_PER_ADDITIONAL_CUSTODY_GROUP),
+            Base10.toString(cfg.BALANCE_PER_ADDITIONAL_CUSTODY_GROUP.uint64),
           MAX_BLOBS_PER_BLOCK_FULU:
```
tersec: Not a thing anymore.

beacon_chain/consensus_object_pools/blob_quarantine.nim
agnxsh: will fix that, but just curious why do you say so?

beacon_chain/rpc/rest_config_api.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7187/commits/bba30d0915bd2d3a271cda3442683b903894899f

beacon_chain/gossip_processing/eth2_processor.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7187/commits/bba30d0915bd2d3a271cda3442683b903894899f

beacon_chain/consensus_object_pools/blob_quarantine.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7187/commits/bba30d0915bd2d3a271cda3442683b903894899f

beacon_chain/beacon_node.nim
```diff
@@ -56,6 +56,7 @@ type
     phase0AttSlashQueue*: AsyncEventQueue[phase0.AttesterSlashing]
     electraAttSlashQueue*: AsyncEventQueue[electra.AttesterSlashing]
     blobSidecarQueue*: AsyncEventQueue[BlobSidecarInfoObject]
+    columnSidecarQueue*: AsyncEventQueue[DataColumnSidecar]
```
tersec, line 59: What is this whole mechanism used for?

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -516,6 +529,78 @@ func fetchMissingSidecars*(
...
+      columns =
+        if len(peerCustodyColumns) > 0:
+          @peerCustodyColumns
+        else:
+          quarantine.custodyColumns
+    if len(record.sidecars) == 0:
+      var columnsRequested = 0
+      for column in columns:
+        if columnsRequested >= columnsCount:
+          # We don't need to request more than (NUMBER_OF_COLUMNS div 2 + 1)
```
tersec, line 567: the nimbus_beacon_node check in this same PR is for the + 0 version?

agnxsh: not anymore, it's essentially `>`, so its kinda the same thing

Issue #7160, opened 2025-05-14
Non-Spec Fields Present in `/eth/v1/node/peers/{peer_id}` response


Alleysira:
Hello developers,

I am conducting API testing on a Electra PoS devnet chain and noticed inconsistencies in the JSON-RPC responses among Prysm, Lighthouse, Nimbus, and Teku. Specifically, the endpoint response of`/eth/v1/node/peers/{peer_id}` from nimbus have 2 extra fields. 

```shell
# nimbus
{"data": {"agent": "nimbus", "direction": "inbound", "enr": "enr:-Le4QDGhdT_15C6FqCBh1QJaQIERrCiY_Fb1bh-JBBXRf9-sStCfUA8hUS8JjFN7vsFs1iy9Hjq5gPJGzpn9f1bFldQCh2F0dG5ldHOIAAAAAAAAAACDY2djBIRldGgykJVYPQVgAAA4__________-CaWSCdjSCaXCErBAEJolzZWNwMjU2azGhA8ux81ss52i76O2ZltTPiz1UQ0h4OZQ_flibs1rtCHrzg3RjcIIjKIN1ZHCCIyg", "last_seen_p2p_address": "/ip4/172.16.4.38/tcp/46446/p2p/16Uiu2HAmSN8Vb3WDnwF7wb7cjtJehqp7JgjSbGYkA1na2xcgybJ2", "peer_id": "16Uiu2HAmSN8Vb3WDnwF7wb7cjtJehqp7JgjSbGYkA1na2xcgybJ2", "proto": "ipfs/0.1.0", "state": "connected"}}
# expected
[...]

PR #7216, opened 2025-06-09
Remove non spec api fields

ahshum:
Addresses
- https://github.com/status-im/nimbus-eth2/issues/7160

Checked the fields are not being used elsewhere. Some reference PRs
- https://github.com/status-im/nimbus-eth2/pull/2459
- https://github.com/status-im/nimbus-eth2/pull/2756


cheatfate:
This fields was added for a reason, to be able to understand what client is actually connected and what version of protocol it was using. You should not delete it, but create proposal to modify specification. Or if you do not want to do it, please make same call in `/eth/v1/nimbus/` endpoints to keep this important functionality.

ahshum:
Yes true I understand what they for. I think a better alternative is to add these information into headers instead of custom fields on top of the specification.

Regards to `/nimbus` endpoints, I am not quite following how they are related to the Beacon spec, but I think we are the spec owner so we should be good to keep the fields if they are useful to our usage on these endpoints.

cheatfate:
Headers? What headers you talking about?
I have already responded to issue author. We are providing all the required information per specification, but specification does not forbids us to add information to responses.

ahshum:
Great thank you. Closing it.

PR #7219, opened 2025-06-10
Fix VC should handle properly BLOB_SCHEDULE value.

cheatfate:


beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -2858,7 +2858,13 @@ proc readValue*(reader: var JsonReader[RestJson],
                 value: var VCRuntimeConfig) {.
      raises: [SerializationError, IOError].} =
   for fieldName in readObjectFields(reader):
-    let fieldValue = reader.readValue(string)
+    let fieldValue =
+      case toLowerAscii(fieldName)
+      of "blob_schedule":
+        string(reader.readValue(JsonString))
```
tersec, line 2864: is this guaranteed to be JSON?

cheatfate: It just reads whatever is under that field into `JsonString` (validates Json) but did not unpack it to the object. Because VC does not need this data we just store it as string, or we can just replace it with "N/A" string, it does not matter.

Issue #7151, opened 2025-05-12
Missing implementation of `postStateValidatorIdentities` in Beacon-APIs


Alleysira:
Please check the [issue](https://github.com/Consensys/teku/issues/9417) in teku for details.

Thanks for your attention.

Alleysira:
- #7223

PR #7223, opened 2025-06-11
Add beacon api postStateValidatorIdentities

ahshum:
- https://github.com/status-im/nimbus-eth2/issues/7151


beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -223,6 +223,11 @@ type
     status*: string
     validator*: Validator
 
+  RestValidatorIdentity* = object
+    index*: ValidatorIndex
+    pubkeyData*{.serializedFieldName: "pubkey".}: HashedValidatorPubKey
```
tersec: why can't this field just be called `pubkey`?

ahshum: Yeah both work for me. I think it mainly helps the naming consistency as other types using `HashedValidatorPubKey` are `pubkeyData`.

cheatfate: There is no reason to call it `pubkeyData`, to keep naming consistency rename it please to `pubkey`. Even if some types internally using `pubkeyData` all the REST types using `pubkey`.

cheatfate: Also i have not seen `HashedValidatorPubKey` serialization/deserialization procedure, so it will not working because its impossible to automatically decode this type, so i think proper type for this object would be `ValidatorPubKey`.

cheatfate:
This change require implementation of additional tests in `ncli/resttest-rules.json`

beacon_chain/spec/eth2_apis/rest_types.nim
ahshum: What does it mean? Serialization is working fine for `Validator` in the response of other APIs.

cheatfate: It means that it could provide whatever `json-serialization` will prefer, which probably could be totally incompatible with what specification means.

tersec:
> This change require implementation of additional tests in `ncli/resttest-rules.json`

https://github.com/status-im/nimbus-eth2/pull/7223/commits/3327c4c2b39965edde5cc1a856237a862a70e765

beacon_chain/spec/eth2_apis/rest_types.nim
tersec: > There is no reason to call it `pubkeyData`, to keep naming consistency rename it please to `pubkey`. Even if some types internally using `pubkeyData` all the REST types using `pubkey`.

https://github.com/status-im/nimbus-eth2/pull/7223/commits/80cf2365d66da58bce7cb7df1c5bf3feffdd3023

cheatfate (approved):

beacon_chain/spec/eth2_apis/rest_types.nim
ahshum: Confirmed with @cheatfate that using `ValidatorPubKey` the test is not needed. Removing the last commit.

PR #7237, opened 2025-06-15
Add /eth/v1/debug/beacon/data_column_sidecars/{block_id}

etan-status:
Implement new endpoint:

- https://github.com/ethereum/beacon-APIs/pull/537

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -132,7 +132,7 @@ proc toString*(kind: ValidatorFilterKind): string =
   of ValidatorFilterKind.WithdrawalDone:
     "withdrawal_done"
 
-func handleDataSidecarRequest*[
+proc handleDataSidecarRequest*[
     InvalidIndexValueError: static string,
     DataSidecarsType: typedesc[List];
     getDataSidecar: static proc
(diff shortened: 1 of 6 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -132,6 +132,49 @@ proc toString*(kind: ValidatorFilterKind): string =
   of ValidatorFilterKind.WithdrawalDone:
     "withdrawal_done"
 
+func handleDataSidecarRequest*[
```
tersec, line 135: Typically the `rest_beacon_api` module handles specifically what's needed for the https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon namespace of the Beacon API -- there's an overloaded use here. It's not really the "base" module for the overall-REST-beacon-API, so unless `handleDataSidecarRequest` is otherwise relevant to specifically the https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon endpoints, it's probably better to have this elsewhere (either in a proper helper module, which might already exist, or directly in `rest_debug_api`, but either way not `rest_beacon_api`).

tersec:
- https://github.com/nim-lang/Nim/issues/25000

PR #7241, opened 2025-06-15
Add vanity log for Fulu transition and BPO update

etan-status:
Trigger 🐅 art on Fulu upgrade and whenever a BPO activates.
Credits https://beatscribe.com


etan-status:
image

image

PR #7243, opened 2025-06-17
automated consensus spec URL updating to v1.6.0-alpha.2

tersec:
```
git reset --hard && make update && sh <(scripts/find_unchanged_consensus_spec_files.sh v1.5.0 v1.6.0-alpha.2 .) && make update
```
Unusually for these, it also has a couple of small cleanups stemming from the EIP-6110 effort, because it's so small otherwise.

It appears that the consensus-specs repo has been mass-reformatted.

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -1243,6 +1243,3 @@ proc testWeb3Provider*(
 
   discard request "Sync status":
     web3.provider.eth_syncing()
-
-  discard request "Latest block":
-    web3.provider.eth_getBlockByNumber(blockId("latest"), false)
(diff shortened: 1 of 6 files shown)


beacon_chain/el/el_manager.nim
```diff
@@ -1243,6 +1243,3 @@ proc testWeb3Provider*(
 
   discard request "Sync status":
     web3.provider.eth_syncing()
-
-  discard request "Latest block":
-    web3.provider.eth_getBlockByNumber(blockId("latest"), false)
```
tersec, line 1248: Nimbus under no circumstances calls this endpoint anymore, so no use to test it either. It did during Eth1Data scanning.

PR #7279, opened 2025-07-02
update nimbus eth1 documentation

advaita-saha:


kdeme (approved):
I think this can get merged already? These are useful updates of how to currently run the EL.

Only the "pre-synced database" could be contested, but then we should move it outside of this PR and get the docs up-to-date.

tersec:
I'd be ok merging it, but it's still marked as draft, so.

PR #7287, opened 2025-07-10
use EF consensus spec v1.6.0-alpha.3 test vectors

tersec:
https://github.com/ethereum/consensus-specs/releases/tag/v1.6.0-alpha.3
https://github.com/ethereum/consensus-spec-tests/releases/tag/v1.6.0-alpha.3


tersec:
- https://github.com/nim-lang/Nim/issues/25038
is the `version-2-2` regression here.

PR #7298, opened 2025-07-15
ci: disable restart from stage

siddarthkay:
related issue: https://github.com/status-im/infra-ci/issues/202

### DIFF
--- a/ci/Jenkinsfile
+++ b/ci/Jenkinsfile
@@ -31,6 +31,7 @@ pipeline {
   }
 
   options {
+    disableRestartFromStage()
     timestamps()
     ansiColor('xterm')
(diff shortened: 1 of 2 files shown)


ci/Jenkinsfile
```diff
@@ -9,6 +9,7 @@
 library 'status-jenkins-lib@v1.9.2'
 
 pipeline {
+
```
jakubgs: What is this newline for?

jakubgs (approved):

ci/Jenkinsfile
siddarthkay: ah must be a leftover, i've removed it now

tersec (approved):

PR #7299, opened 2025-07-15
Make nimbus beacon node accessible for threading and as a module.

pedromiguelmiranda:
- Add a public entry point to nimbus beacon node by changing 'handleStartUpCmd'
 scope to public.

- protect database file lock concurrent creation/deletion


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2166,11 +2166,15 @@ proc run(node: BeaconNode) {.raises: [CatchableError].} =
...
 
+var shouldCreatePidFile*: Atomic[bool]
+shouldCreatePidFile.store(true)
+
 var gPidFile: string
 proc createPidFile(filename: string) {.raises: [IOError].} =
-  writeFile filename, $os.getCurrentProcessId()
-  gPidFile = filename
-  addExitProc proc {.noconv.} = discard io2.removeFile(gPidFile)
+  if shouldCreatePidFile.load():
```
arnetheduck, line 2196: shouldn't this compareandswap(true)? this way, the variable doesn't have to be public and it's guaranteed to run once only

also, this should be moved to nimbus_binary_common if it's going to be shared between binaries

pedromiguelmiranda: needs to be public, otherwise a different thread can not read/write. 

The purpose of this change is to avoid the the error triggered by `addExitProc` on a threaded environment, when it tries to delete something that does not exist anymore.

Agree on moving this to `nimbus_binary_common`

arnetheduck:
Superseded by #7526

PR #7304, opened 2025-07-17
add BLOB_SCHEDULE to /config

agnxsh:


beacon_chain/rpc/rest_config_api.nim
```diff
@@ -7,6 +7,7 @@
 
 {.push raises: [].}
 
+import std/algorithm, sequtils, json
```
tersec:
```suggestion
import std/[algorithm, json, sequtils]
```

beacon_chain/rpc/rest_config_api.nim
```diff
@@ -16,11 +17,20 @@ export rest_utils
 
 logScope: topics = "rest_config"
 
+func cmpBPOconfig(x, y: BlobParameters): int =
+  cmp(x.EPOCH.distinctBase, y.EPOCH.distinctBase)
```
tersec, line 21: This is subtly different than the sorting function used elsewhere because it's ascending, not descending; maybe note that explicitly to make it clear it's intentionally designed this way and not a potentially latent bug.

agnxsh: 0k

tersec: https://github.com/status-im/nimbus-eth2/pull/7304/commits/389b5f97f1c9c0ac8b5ff01906d949beb5bd5b5d

beacon_chain/rpc/rest_config_api.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7304/commits/4907ead90958a9aab604034793f56de1eee1a7ff

PR #7306, opened 2025-07-18
Fulu v1.6.0-alpha.3 helpers

ahshum:


beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -290,7 +290,24 @@ func get_extended_sample_count*(samples_per_slot: int,
 
   NUMBER_OF_COLUMNS
 
-# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/p2p-interface.md#verify_data_column_sidecar_inclusion_proof
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.3/specs/fulu/p2p-interface.md#verify_data_column_sidecar
+proc verify_data_column_sidecar*(sidecar: DataColumnSidecar):
+                                 Result[void, cstring] =
+  # Verify if the data column sidecar is valid.
```
tersec: This should be a [documentation comment](https://nim-lang.org/docs/manual.html#lexical-analysis-comments) starting with `##` because it documents what the function does as a whole:
```nim
  ## Verify if the data column sidecar is valid.
```

These get used by [documentation generation tools](https://nim-lang.org/docs/docgen.html#documentation-generation-tools_1) in a way ordinary `#` comments aren't.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -307,23 +324,11 @@ proc verify_data_column_sidecar_inclusion_proof*(sidecar: DataColumnSidecar):
 
   ok()
 
-# https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.10/specs/fulu/p2p-interface.md#verify_data_column_sidecar_kzg_proofs
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.3/specs/fulu/p2p-interface.md#verify_data_column_sidecar_kzg_proofs
 proc verify_data_column_sidecar_kzg_proofs*(sidecar: DataColumnSidecar):
                                             Result[void, cstring] =
   ## Verify if the KZG proofs are correct.
 
-  # Check if the data column sidecar index < NUMBER_OF_COLUMNS
```
tersec, line 315: These removals are correct, but then would need to adjust
https://github.com/status-im/nimbus-eth2/blob/94fb81248b404d75a07c6b42d1c8768097ba3154/beacon_chain/gossip_processing/gossip_validation.nim#L579-L705
because it doesn't do that first step from https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.3/specs/fulu/p2p-interface.md#data_column_sidecar_subnet_id of
> [REJECT] The sidecar is valid as verified by `verify_data_column_sidecar(sidecar)`.

Because Nimbus didn't have that function.

This PR adds it, but it should then replace, e.g.
https://github.com/status-im/nimbus-eth2/blob/94fb81248b404d75a07c6b42d1c8768097ba3154/beacon_chain/gossip_processing/gossip_validation.nim#L589-L592
and anything else not in the current list of validations, which has been subsumed into `verify_data_column_sidecar` check which should be added in its place.

Because right now, it's handled by some other functions in the gossip handler which by this spec alignment now lose that.

ahshum: Thanks for the comments. Sorry that the change wasn't included as I was checking to make sure the validations match with the spec.

Just updated.

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7306/commits/0f785752c990c88c2f46e85ae530d191c7e7b85f

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7306/commits/296998913d064fa17449da183796092a9fea4dd0

PR #7307, opened 2025-07-21
log: formatting for PubKey0x

arnetheduck:
In anticipation of https://github.com/status-im/nim-chronicles/pull/166

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -1,4 +1,3 @@
-
 # beacon_chain
 # Copyright (c) 2018-2025 Status Research & Development GmbH
 # Licensed and distributed under either of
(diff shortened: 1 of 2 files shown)


beacon_chain/conf.nim
```diff
@@ -41,7 +41,7 @@ export
   defs, parseCmdArg, completeCmdArg, network_metadata,
   el_conf, network, BlockHashOrNumber,
   confTomlDefs, confTomlNet, confTomlUri,
-  LightClientDataImportMode
+  LightClientDataImportMode, slashing_protection_common
```
tersec, line 44: Why is this needed?

arnetheduck: it declares the `PubKey0x` type (which is used in configuration) and above all its `formatIt` which later is used in logging - previously it was logged as a plain `object` but next chronicles logs fields with their individual formatIt

PR #7309, opened 2025-07-21
use status v1/v2 on a priority basis

agnxsh:


beacon_chain/networking/peer_protocol.nim
```diff
@@ -211,9 +216,20 @@ p2pProtocol PeerSync(version = 1,
         discard await peer.handleStatusV2(peer.networkState, theirStatus.get())
         peer.updateAgent()
       else:
-        debug "Status response not received in time",
-              peer, errorKind = theirStatus.error.kind
-        await peer.disconnect(FaultOrError)
+        # Mark status v2 of remote peer as None.
+        peer.state(PeerSync).setStatusV2Msg(Opt.none(StatusMsgV2))
+        let
```
cheatfate: I think fallback should not be used in this case, peers should not use V1 status message after `peer.networkState.cfg.FULU_FORK_EPOCH` and should be disconnected with an error.

beacon_chain/networking/peer_protocol.nim
```diff
@@ -211,9 +216,20 @@ p2pProtocol PeerSync(version = 1,
         discard await peer.handleStatusV2(peer.networkState, theirStatus.get())
         peer.updateAgent()
       else:
-        debug "Status response not received in time",
-              peer, errorKind = theirStatus.error.kind
-        await peer.disconnect(FaultOrError)
+        # Mark status v2 of remote peer as None.
+        peer.state(PeerSync).setStatusV2Msg(Opt.none(StatusMsgV2))
```
tersec, line 220: In theory this is default, but nice to make intent clear like this

cheatfate (changes requested):

beacon_chain/networking/peer_protocol.nim
agnxsh: alright ok

beacon_chain/networking/peer_protocol.nim
```diff
@@ -348,8 +359,9 @@ proc getHeadRoot*(peer: Peer): Eth2Digest =
     state = peer.networkState(PeerSync)
     pstate = peer.state(PeerSync)
     remoteFork = state.getBeaconTime().slotOrZero.epoch()
-  if remoteFork >= state.cfg.FULU_FORK_EPOCH:
-    pstate.statusMsgV2.headRoot
+  if remoteFork >= state.cfg.FULU_FORK_EPOCH and
```
cheatfate: I think this condition is unnecessary and just `pstate.statusMsgV2.isSome()` would be enough.

beacon_chain/networking/peer_protocol.nim
```diff
@@ -358,8 +370,9 @@ proc getHeadSlot*(peer: Peer): Slot =
     state = peer.networkState(PeerSync)
     pstate = peer.state(PeerSync)
     remoteFork = state.getBeaconTime().slotOrZero.epoch()
-  if remoteFork >= state.cfg.FULU_FORK_EPOCH:
-    pstate.statusMsgV2.headSlot
+  if remoteFork >= state.cfg.FULU_FORK_EPOCH and
```
cheatfate: Same comment as above.

beacon_chain/networking/peer_protocol.nim
```diff
@@ -368,8 +381,9 @@ proc getFinalizedEpoch*(peer: Peer): Epoch =
     state = peer.networkState(PeerSync)
     pstate = peer.state(PeerSync)
     remoteFork = state.getBeaconTime().slotOrZero.epoch()
-  if remoteFork >= state.cfg.FULU_FORK_EPOCH:
-    pstate.statusMsgV2.finalizedEpoch
+  if remoteFork >= state.cfg.FULU_FORK_EPOCH and
```
cheatfate: Same comment as above.

agnxsh: https://github.com/status-im/nimbus-eth2/pull/7309/commits/a225d318cf2aa4297a31b838bd3ba27fff52c5a3

beacon_chain/networking/peer_protocol.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7309/commits/a225d318cf2aa4297a31b838bd3ba27fff52c5a3

beacon_chain/networking/peer_protocol.nim
agnxsh: https://github.com/status-im/nimbus-eth2/pull/7309/commits/a225d318cf2aa4297a31b838bd3ba27fff52c5a3

PR #7312, opened 2025-07-22
optimize slashing protection interchange import

tomasandroil:
This PR addresses a TODO in `slashing_protection_common.nim` related to the interchange import logic.

### Changes
- Replaced inefficient `O(n log n)` sort of `signed_blocks` with a linear `O(n)` scan to select the block with the highest slot.
- Removed redundant sorting of `signed_attestations`. The following loop already determines the max epochs.
- Eliminated unnecessary imports (`algorithm`) to reduce compile scope.

### Rationale
These changes improve performance and reduce complexity with **no changes to functionality**.

### Testing
All slashing protection tests pass:
`tests/slashing_protection/test_slashing_protection_db.nim`

---

> This is a performance-only cleanup and has been verified locally with full test suite.


tomasandroil:
Quick reminder about PR #7312 (“optimize slashing protection interchange import”):

- Switched signed_blocks selection from O(n log n) sort to an O(n) scan for max slot  
- Removed redundant signed_attestations sort (max epochs already determined)  
- Eliminated the unused `algorithm` import  

All tests in `tests/slashing_protection/` still pass. Let me know if you need anything else! 🚀

beacon_chain/validators/slashing_protection_common.nim
```diff
@@ -305,83 +305,54 @@ proc importInterchangeV5Impl*(
...
-    # TODO: with minification sorting is unnecessary, cleanup
-    # Sort by ascending minimum slot so that we don't trigger MinSlotViolation
-    spdir.data[v].signed_blocks.sort do (a, b: SPDIR_SignedBlock) -> int:
-      result = cmp(a.slot.int, b.slot.int)
-
-    spdir.data[v].signed_attestations.sort do (a, b: SPDIR_SignedAttestation) -> int:
-      result = cmp(a.source_epoch.int, b.source_epoch.int)
-      if result == 0: # Same epoch
-        result = cmp(a.target_epoch.int, b.target_epoch.int)
+    # --- Replace the original sorting + selection logic ---
```
tersec: It doesn't really make sense to have this as inline commentary in the result.

beacon_chain/validators/slashing_protection_common.nim
```diff
@@ -390,26 +361,20 @@ proc importInterchangeV5Impl*(
...
+      if a.target_epoch.int > maxValidTargetEpochSeen:
+        maxValidTargetEpochSeen = a.target_epoch.int
 
     if maxValidSourceEpochSeen < 0 or maxValidTargetEpochSeen < 0:
       doAssert maxValidSourceEpochSeen == -1 and maxValidTargetEpochSeen == -1
       notice "No attestation found in slashing interchange file for validator",
         pubkey = spdir.data[v].pubkey.PubKeyBytes.toHex()
       continue
 
-    # See formal proof https://github.com/michaelsproul/slashing-proofs
```
tersec, line 407: Why remove this? It seems unrelated to the rest, and the link appears to still be valid/live

beacon_chain/validators/slashing_protection_common.nim
```diff
@@ -305,83 +305,54 @@ proc importInterchangeV5Impl*(
...
+        maxValidSlotSeen = int latestBlock.slot
 
-    # Now prune everything that predates
-    # this DB or interchange file max slot
-    # Even if the block is not imported, pruning will keep the latest one.
     db.pruneBlocks(parsedKey, Slot maxValidSlotSeen)
 
     # Attestations
     # ---------------------------------------------------
-    # After import we need to prune the DB from everything
```
tersec, line 380: This explanatory comment still seems valid?

beacon_chain/validators/slashing_protection_common.nim
```diff
@@ -305,83 +305,54 @@ proc importInterchangeV5Impl*(
...
+            candidateBlock = latestBlock,
             conflict = status.error()
           result = siPartial
 
-      if B.slot.int > maxValidSlotSeen:
-        maxValidSlotSeen = int B.slot
+      if latestBlock.slot.int > maxValidSlotSeen:
+        maxValidSlotSeen = int latestBlock.slot
 
-    # Now prune everything that predates
```
tersec, line 373: As with the other comment removals, this seems separate from the stated intent of this PR?

beacon_chain/validators/slashing_protection_common.nim
```diff
@@ -305,83 +305,54 @@ proc importInterchangeV5Impl*(
...
+          latestBlock = b
+
       let
         signing_root =
-          if B.signing_root.isSome:
-            B.signing_root.get.Eth2Digest
+          if latestBlock.signing_root.isSome:
+            latestBlock.signing_root.get.Eth2Digest
           else:
-            # https://eips.ethereum.org/EIPS/eip-3076#advice-for-complete-databases
```
tersec, line 339: This is pretty useful context about why `ZeroDigest` is a good idea here. Not sure, as with other comments, it's removed.

tersec:
The core idea of replacing O(n log n) sorting with O(n) selection is sound, but this particular PR has lots of other unwarranted comment changes.

It's true that not using `std/algorithms` might slightly compile time, but those `std/` imports are fairly fast, and not in themselves much concern, and practically, it's likely imported anyway by the time `beacon_chain/validators/slashing_protection_common.nim
` is imported due to other modules in Nimbus, so it itself isn't that compelling on its own.

tomasandroil:
@tersec 
Thanks for the review! I’ve addressed the feedback:
- Removed the stray inline commentary.
- Restored the pruning explanations for blocks/attestations.
- Re-added the formal proof link and the genesis special-case in attestation checks (slashing-proofs: https://github.com/michaelsproul/slashing-proofs).
- Restored the EIP-3076 advice context next to the ZeroDigest use (EIP-3076: https://eips.ethereum.org/EIPS/eip-3076#advice-for-complete-databases).
- Kept the O(n) selection (no sorting) and left std/algorithm out since it’s no longer needed.

I also restored the chronicles.formatIt PubKey0x for log parity; Please let me know if you’d like any other comments tweaked.

PR #7316, opened 2025-07-22
Adjust for authList v field change to yParity name

kdeme:
Draft as depends on https://github.com/status-im/nim-eth/pull/806


kdeme:
Will fail as still needs a web3 update:

Either just this part of change:  https://github.com/status-im/nim-web3/blob/dev/etan/ff-yparity/web3/conversions.nim#L307-L313

Or the full blown PR that adds `U8Quantity` and `AuthorizationObject`: https://github.com/status-im/nim-web3/pull/209

arnetheduck:
> Will fail as still needs a web3 update:

can't a compat shim be added to web3 btw so that old code continues to work when bumping it?

kdeme:
@jangko ok to merge now I assume?

jangko:
> @jangko ok to merge now I assume?

Yup, both nimbuses and nim-web3 tests are green now.

PR #7321, opened 2025-07-23
Move Electra fork from VC optional to required forks.

cheatfate:


etan-status:
There is a copy paste of this list in `beacon_chain/deposits.nim` from #6811.

tersec:
> There is a copy paste of this list in `beacon_chain/deposits.nim` from #6811.

- https://github.com/status-im/nimbus-eth2/pull/7322

etan-status:
image

I experimented with a solution a bit back to prevent this getting out of sync. It would need `staticExec` or `--experimental:vmopsDanger` though, because otherwise time cannot be accessed at compile time.

I didn't proceed with that because it seems that the underlying approach is flawed; in essence, what this check does is that it requires the underlying network to have a feature set of ≥ EF mainnet. Maybe, the checks could simply be relaxed so that any forks scheduled in the future are ignored (if BN xor VC knows about it), and that only forks in the present or past are required (if BN or VC knows about a fork but the other doesn't, then mark as incompatible). Further, the BPO forks also would have to be compared, as they may have different fork digests / parameters.

etan-status:
Reason for `--experimental:vmopsDanger` is that a handful of functions just isn't available at compile time. It's weird that `staticExec` doesn't need that flag, considering it allows even more "dangerous" things.

image

tersec:
> I didn't proceed with that because it seems that the underlying approach is flawed; in essence, what this check does is that it requires the underlying network to have a feature set of ≥ EF mainnet. Maybe, the checks could simply be relaxed so that any forks scheduled in the future are ignored (if BN xor VC knows about it), and that only forks in the present or past are required (if BN or VC knows about a fork but the other doesn't, then mark as incompatible). Further, the BPO forks also would have to be compared, as they may have different fork digests / parameters.

The static-at-compiletime approach is flawed because it shouldn't be a function of when the binary was built, but when it's being run, anyway.

But yes, the main important aspect is that future forks should not (with some constraints) bind current behavior too tightly. Up through the current fork has to match, beyond that, no.

It can be a nice thing, maybe, to note to the user that the BN they're using will stop working with the next fork or BPO. But if it works at that moment, at runtime, because the fork schedule through the current beacontime covers the same fork schedule on BN and VC, those seem compatible.

Ultimately I'm not sure why to have this static special case at all.

etan-status:
Exactly, I was looking into static-at-compiletime because the current approach does that, just removing the maintenance requirement from it as it has been proven that the manual maintenance doesn't work. But I didn't proceed because I'm not convinced that it is correct.

PR #7330, opened 2025-07-25
Avoid CStringConv warning when returning LC error

etan-status:
Fixes warning

```
/light_client_processor.nim(542, 17) Warning: implicit conversion to
'cstring' from a non-const location: $error(r); this will become a
compile time error in the future [CStringConv]
```

> a pointer (`cstring`) to a local temporary is being returned here
  and it can get collected before it's used


beacon_chain/gossip_processing/light_client_processor.nim
```diff
@@ -539,15 +539,15 @@ func toValidationError(
     of VerifierError.Invalid:
       # [REJECT] The `finality_update` is valid.
       # [REJECT] The `optimistic_update` is valid.
-      errReject($r.error)
+      errReject(typeof(obj).name & ": invalid")
```
tersec, line 542: this loses the actual reason it was rejected though. It would be ideal to keep that somehow.

etan-status: $r.error is "Invalid", just dynamic instead of static string

PR #7334, opened 2025-07-28
NJS: Add option to disable automatic serialization of primitive types

jangko:


jangko:
close in favor of #7316

PR #7340, opened 2025-07-29
fix fulu builder api changes

agnxsh:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -920,14 +920,45 @@ proc getBuilderBid[
 proc proposeBlockMEV(
     node: BeaconNode, payloadBuilderClient: RestClientRef,
     blindedBlock:
-      electra_mev.SignedBlindedBeaconBlock |
+      electra_mev.SignedBlindedBeaconBlock):
+    Future[Result[Opt[BlockRef], string]] {.async: (raises: [CancelledError]).} =
+  let unblindedBlockRef = await node.unblindAndRouteBlockMEV(
+    payloadBuilderClient, blindedBlock)
+  if unblindedBlockRef.isOk and unblindedBlockRef.get.isSome:
+    beacon_blocks_proposed.inc()
```
tersec, line 928: Should add a comment here to make it obvious to remove this when Electra MEV support is removed that this case can only happen in Electra, not Fulu.

beacon_chain/el/engine_api_conversions.nim
```diff
@@ -233,11 +233,11 @@ func asConsensusTypeFulu*(
     # The `mapIt` calls below are necessary only because we use different distinct
     # types for KZG commitments and Blobs in the `web3` and the `deneb` spec types.
     # Both are defined as `array[N, byte]` under the hood.
-    blobsBundle: deneb.BlobsBundle(
+    blobsBundle: fulu.BlobsBundleV2(
```
tersec, line 235: This isn't accurate; `GetPayloadV4` uses `BlobsBundleV1`: https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/prague.md#engine_getpayloadv4

beacon_chain/spec/state_transition.nim
```diff
@@ -570,7 +570,7 @@ proc makeBeaconBlockWithRewards*(
                hash_tree_root(validator_changes.proposer_slashings),
                hash_tree_root(validator_changes.electra_attester_slashings),
                hash_tree_root(
-                 List[electra.Attestation, Limit MAX_ATTESTATIONS](
+                 List[electra.Attestation, Limit MAX_ATTESTATIONS_ELECTRA](
```
tersec, line 573: This is the actual fix.

beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -550,6 +550,7 @@ type
   GetHeaderResponseElectra* = DataVersionEnclosedObject[electra_mev.SignedBuilderBid]
   GetHeaderResponseFulu* = DataVersionEnclosedObject[fulu_mev.SignedBuilderBid]
   SubmitBlindedBlockResponseElectra* = DataVersionEnclosedObject[electra_mev.ExecutionPayloadAndBlobsBundle]
+  SubmitBlindedBlockResponseFulu* = DataVersionEnclosedObject[fulu_mev.ExecutionPayloadAndBlobsBundle]
```
tersec, line 553: Per https://github.com/ethereum/builder-specs/pull/123 the `v2` API which Nimbus now uses for Fulu instead of the deprecated `v1` API "does not return the execution payload and blobs".

There's no decoding to do, unlike Pectra. So this `SubmitBlindedBlockResponseFulu` type is neither necessary nor useful.

beacon_chain/spec/eth2_apis/eth2_rest_serialization.nim
```diff
@@ -270,6 +270,7 @@ RestJson.useDefaultSerializationFor(
   fulu_mev.BlindedBeaconBlock,
   fulu_mev.BlindedBeaconBlockBody,
   fulu_mev.BuilderBid,
+  fulu_mev.ExecutionPayloadAndBlobsBundle,
```
tersec, line 273: Because `SubmitBlindedBlockResponseFulu` isn't used, neither is this.

beacon_chain/el/engine_api_conversions.nim
```diff
@@ -233,11 +233,11 @@ func asConsensusTypeFulu*(
...
       commitments: KzgCommitments.init(
         payload.blobsBundle.commitments.mapIt(
           kzg_abi.KzgCommitment(bytes: it.data))),
-      proofs: KzgProofs.init(
+      proofs: KzgProofsV2.init(
```
tersec, line 239: Also not accurate.

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1042,7 +1042,8 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
             doAssert strictVerification notin node.dag.updateFlags
             return RestApiResponse.jsonError(Http400, InvalidBlockObjectError)
 
-          when consensusFork >= ConsensusFork.Deneb:
+          when consensusFork >= ConsensusFork.Deneb and
```
tersec, line 1045: Two points:
- https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock shows " Warning: Deprecated". That is, I'll remove (`Http410`) this entire handler as soon as feasible come Fulu. It doesn't matter if it works under Fulu; it will never have to.

- Even otherwise, this endpoint isn't about blinded blocks, but non-MEV blocks, and those should be sending their blob/column sidecars somewhere.

But the first point is more important. There's no reason to change this to work in Fulu, nor should it be changed.

beacon_chain/spec/forks.nim
```diff
@@ -489,7 +489,8 @@ template kind*(
       fulu.MsgTrustedSignedBeaconBlock |
       fulu.TrustedSignedBeaconBlock |
       fulu_mev.SignedBlindedBeaconBlock |
-      fulu_mev.SignedBuilderBid]): ConsensusFork =
+      fulu_mev.SignedBuilderBid |
+      fulu_mev.ExecutionPayloadAndBlobsBundle]): ConsensusFork =
```
tersec, line 512: Shouldn't be used either.

I removed all this, for good reason, back in https://github.com/status-im/nimbus-eth2/pull/7308

There's _still_ no actual block/blob return from `submitBlindedBlock` in Fulu, so all of this is pointless. The stateroot bug was related, as far as it seems so far, in the construction of the shim block in `state_transition`, and that fix is elsewhere. There's still nothing to decode from the relay for Fulu, aside from checking the 202-or-not-204 HTTP return.

beacon_chain/spec/mev/fulu_mev.nim
```diff
@@ -69,11 +70,17 @@ type
     message*: BlindedBeaconBlock
     signature*: ValidatorSig
 
+  # https://github.com/ethereum/builder-specs/blob/ae1d97d080a12bfb7ca248b58fb1fc6b10aed02e/specs/fulu/builder.md#executionpayloadandblobsbundle
+  ExecutionPayloadAndBlobsBundle* = object
```
tersec, line 74: See elsewhere, not useful/necessary.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -2007,15 +2053,27 @@ proc makeMaybeBlindedBeaconBlockForHeadAndSlotImpl[ResultType](
...
+  elif consensusFork >= ConsensusFork.Deneb:
     doAssert engineBid.blobsBundle.commitments ==
       forkyBlck.body.blob_kzg_commitments
     ResultType.ok((
       blck: consensusFork.MaybeBlindedBeaconBlock(
         isBlinded: false,
         data: consensusFork.BlockContents(
           `block`: forkyBlck,
-          kzg_proofs: engineBid.blobsBundle.proofs,
+          kzg_proofs: KzgProofs(engineBid.blobsBundle.proofs),
```
tersec, line 2076:
```
/github-runner/github-runner-node-01/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/validators/beacon_validators.nim(2076, 32) Hint: conversion from KzgProofs to itself is pointless [ConvFromXtoItselfNotNeeded]
```

beacon_chain/validators/beacon_validators.nim
```diff
@@ -2007,15 +2053,27 @@ proc makeMaybeBlindedBeaconBlockForHeadAndSlotImpl[ResultType](
...
-  when consensusFork >= ConsensusFork.Deneb:
+  when consensusFork >= ConsensusFork.Fulu:
+    doAssert engineBid.blobsBundleV2.commitments ==
+      forkyBlck.body.blob_kzg_commitments
+    ResultType.ok((
+      blck: consensusFork.MaybeBlindedBeaconBlock(
+        isBlinded: false,
+        data: consensusFork.BlockContents(
+          `block`: forkyBlck,
+          kzg_proofs: KzgProofsV2(engineBid.blobsBundleV2.proofs),
```
tersec, line 2064:
```
/github-runner/github-runner-node-01/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/validators/beacon_validators.nim(2064, 34) Hint: conversion from KzgProofsV2 to itself is pointless [ConvFromXtoItselfNotNeeded]
```

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1194,48 +1239,48 @@ proc proposeBlockAux(
...
-            return head
+            return Opt.some(head)
           res.get()
       signedBlock = consensusFork.SignedBeaconBlock(
         message: forkyBlck, signature: signature, root: blockRoot)
       blobsOpt =
         when consensusFork >= ConsensusFork.Deneb:
           Opt.some(signedBlock.create_blob_sidecars(
-            engineBid.blobsBundle.proofs, engineBid.blobsBundle.blobs))
+            KzgProofs(engineBid.blobsBundle.proofs), engineBid.blobsBundle.blobs))
```
tersec, line 1249:
```
/github-runner/github-runner-node-01/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/validators/beacon_validators.nim(1249, 22) Hint: conversion from KzgProofs to itself is pointless [ConvFromXtoItselfNotNeeded]
```

tersec:
Overtaken by events.

PR #7352, opened 2025-08-03
fork cleanups

arnetheduck:
* avoid extraneous typedesc when select fork - while it works most of the time, the extra layer of typedesc is trickier to support in macros and templates sometimes
* add back some pre-electra types in forks - writing type-generic code is somewhat easier with them there _until_ they are removed wholesale - ie removing some of the support but not all leads to unnecessary maintenance overhead in other parts of the code where sometimes these types can pop up as side effects of other generic usage which makes it harder to write fork-agnostic code
* move rest mev calls to fork-neutral module - the builder api is really not fork-bound that way but rather operates on its own versioning schedule - while some particular version might be guaranteed to exist after some particular fork, writing fork-specific versions of it causes increased maintenance as dependent code must now also be made fork-dependent to call the specially named functions instead of simply following type
* prefer Forky types where code is fork-agnostic
* clean up block signing code - it's fork-agnostic mostly since it sends the header and a proof whose core logic is fork-independent and just needs an index adjustment


beacon_chain/spec/mev/rest_mev_calls.nim
```diff
@@ -54,15 +54,34 @@ proc submitBlindedBlockPlain*(
...
+    extraHeaders = @[("eth-consensus-version", toString(typeof(body).kind))]
+  )
+
+proc submitBlindedBlock*(
+    client: RestClientRef,
+    body: fulu_mev.SignedBlindedBeaconBlock
+): Future[RestPlainResponse] {.
+  async: (raises: [CancelledError, RestEncodingError, RestDnsResolveError,
+                   RestCommunicationError], raw: true).} =
+  # TODO will everyone have upgraded by the time of v2?
```
tersec: yes

beacon_chain/validators/validator_pool.nim
```diff
@@ -535,192 +534,92 @@ proc signData(v: AttachedValidator,
...
+  when fork < ConsensusFork.Deneb:
+    static: raiseAssert "Unsupported fork " & $fork
+  elif fork == ConsensusFork.Deneb:
+    prop.denebIndex
+  elif fork == ConsensusFork.Electra:
+    prop.electraIndex
+  elif fork == ConsensusFork.Fulu:
+    prop.fuluIndex
+  else:
+    static: raiseAssert "Unkown fork " & $fork
```
tersec: `Unknown`

arnetheduck: [065d640](https://github.com/status-im/nimbus-eth2/pull/7352/commits/065d6404ea0a1d2f411a6c4733a9ed2c2e539614)

beacon_chain/spec/mev/rest_mev_calls.nim
arnetheduck: [065d640](https://github.com/status-im/nimbus-eth2/pull/7352/commits/065d6404ea0a1d2f411a6c4733a9ed2c2e539614)

beacon_chain/spec/forks.nim
```diff
@@ -710,10 +747,10 @@ template BlindedBlockContents*(
   elif kind == ConsensusFork.Electra:
     typedesc[electra_mev.BlindedBeaconBlock]
   else:
-    {.error: "BlindedBlockContents does not support " & $kind.}
+    {.error: "BlindedBlockContents unsupported in " & $kind.}
 
 template PayloadAttributes*(
-    kind: static ConsensusFork): auto =
+    kind: static ConsensusFork): typedesc =
```
etan-status, line 753: `forks_light_client.nim` has more of this style (but the PR is already quite big, so I guess fine to do them separately)

arnetheduck: [ea6193a](https://github.com/status-im/nimbus-eth2/pull/7352/commits/ea6193a179559b8357737fb81e03eda9f81d8188)

PR #7353, opened 2025-08-03
Block production revamp

arnetheduck:
The block production flow has grown a bit hairy over the years and at the same time is starting to perform poorly on underpowered hardware when external builders are enabled.

The main architectural change in the new flow is that instead of constructing two full consensus blocks - one for each of the builder and engine payloads - we perform the block value comparison earlier in the flow, just after getting the payload and use that payload to construct a full consensus block.

* move payload creation and selection logic to its own `block_payloads` module, shared by beacon validators and the REST API (for client validators) - signing and routing logic for beacon validators remains in `beacon_validators`
* avoid constructing two consensus blocks when builder is enabled - in particular, avoid running attestation packing and proposal state allocation twice for each block - instead, filter payloads first then create the consensus block
* Log engine/builder payload requests/responses at info level as they happen to aid users in timing and block value research
* Avoid instantiating generic code for both payload and consensus block types at the same time as this de facto results in O(fork^2) instantiations, most of which don't get used


tersec (approved):

tersec:
There's a small merge conflict

PR #7383, opened 2025-08-13
builder: use GENESIS_FORK_VERSION throughout, fix web3signer registration

arnetheduck:
https://github.com/ethereum/builder-specs/blob/v0.4.0/specs/bellatrix/builder.md#signing

>  with domain given by compute_domain(DOMAIN_APPLICATION_BUILDER,
fork_version=None, genesis_validators_root=None)

Ditto web3signer, where we should not be sending a fork info according to spec:

https://github.com/ethereum/remote-signing-api/blob/87a392deb4e43209ca896dde6b4ec40bef7ee02c/signing/schemas.yaml#L312

Per https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.4/specs/phase0/beacon-chain.md#compute_domain, we have to pass in the genesis fork version - since this is not available from the remote signing API, `nimbus_signing_node` gains a `--network` parameter from which the genesis fork version is read.

### DIFF
--- a/beacon_chain/spec/signatures.nim
+++ b/beacon_chain/spec/signatures.nim
@@ -373,27 +373,24 @@ proc verify_contribution_and_proof_signature*(
 
 # https://github.com/ethereum/builder-specs/blob/v0.4.0/specs/bellatrix/builder.md#signing
 func compute_builder_signing_root(
-    fork: Fork,
     msg: electra_mev.BuilderBid | fulu_mev.BuilderBid |
          ValidatorRegistrationV1): Eth2Digest =
-  # Uses genesis fork version regardless
-  doAssert fork.current_version == fork.previous_version
-
+  # Fork = none, per spec
   let domain = get_domain(
-    fork, DOMAIN_APPLICATION_BUILDER, GENESIS_EPOCH, ZERO_HASH)
+    Fork(), DOMAIN_APPLICATION_BUILDER, GENESIS_EPOCH, ZERO_HASH)
   compute_signing_root(msg, domain)
 
 proc get_builder_signature*(
-    fork: Fork, msg: ValidatorRegistrationV1, privkey: ValidatorPrivKey):
+    msg: ValidatorRegistrationV1, privkey: ValidatorPrivKey):
     CookedSig =
-  let signing_root = compute_builder_signing_root(fork, msg)
+  let signing_root = compute_builder_signing_root(msg)
   blsSign(privkey, signing_root.data)
 
 proc verify_builder_signature*(
-    fork: Fork, msg: electra_mev.BuilderBid |
+    msg: electra_mev.BuilderBid |
           fulu_mev.BuilderBid | ValidatorRegistrationV1,
     pubkey: ValidatorPubKey | CookedPubKey, signature: SomeSig): bool =
-  let signing_root = compute_builder_signing_root(fork, msg)
+  let signing_root = compute_builder_signing_root(msg)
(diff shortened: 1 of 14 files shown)


beacon_chain/spec/signatures.nim
```diff
@@ -373,27 +373,24 @@ proc verify_contribution_and_proof_signature*(
...
-    fork: Fork,
     msg: electra_mev.BuilderBid | fulu_mev.BuilderBid |
          ValidatorRegistrationV1): Eth2Digest =
-  # Uses genesis fork version regardless
-  doAssert fork.current_version == fork.previous_version
-
+  # Fork = none, per spec
   let domain = get_domain(
-    fork, DOMAIN_APPLICATION_BUILDER, GENESIS_EPOCH, ZERO_HASH)
+    Fork(), DOMAIN_APPLICATION_BUILDER, GENESIS_EPOCH, ZERO_HASH)
```
tersec: Per https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.4/specs/phase0/beacon-chain.md#compute_domain
```python
    if fork_version is None:
        fork_version = GENESIS_FORK_VERSION
```
so semantically per spec this approach of, pass a `fork` but verify it's genesis fork is equivalent. It's not an "empty" fork it's using, it's the genesis fork.

But Nimbus's `compute_domain` doesn't have `fork_version` having these semi-optional semantics:
https://github.com/status-im/nimbus-eth2/blob/92defe0fe6c67255cef3687514c1ba80b197f3b3/beacon_chain/spec/helpers.nim#L143-L155

whereas
```python
def compute_domain(
    domain_type: DomainType, fork_version: Version = None, genesis_validators_root: Root = None
) -> Domain:
    """
    Return the domain for the ``domain_type`` and ``fork_version``.
    """
    if fork_version is None:
        fork_version = GENESIS_FORK_VERSION
    if genesis_validators_root is None:
        genesis_validators_root = Root()  # all bytes zero by default
    fork_data_root = compute_fork_data_root(fork_version, genesis_validators_root)
    return Domain(domain_type + fork_data_root[:28])
```
does.

Similarly, this `get_domain` wrapper over `compute_domain`:
https://github.com/status-im/nimbus-eth2/blob/c15ad82c0162c84e9e3c24783026e81008c0b29b/beacon_chain/spec/helpers.nim#L157-L170
takes this `Fork` parameter so it can unconditionally pass it to `compute_domain` but be customized, whereas the spec version takes it conditionally from the `state` passed (which `get_domain` here replaces effectively with the `Fork` parameter): https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.4/specs/phase0/beacon-chain.md#get_domain

So this change doesn't work in a standalone way.

arnetheduck: Hm, good catch.

I guess what motivated this change is that web3signer explicitly does not include a fork on the wire when signing the registration and we are wrong to send/require one (spotted in their conformance examples) - it would mean that our signing node needs to be made network-aware since genesis fork changes between networks. Messy.

arnetheduck: https://github.com/status-im/nimbus-eth2/pull/7383/commits/1b642913e64d9ba82a907d56d536ecd06adf268b

arnetheduck: web3signer itself takes a network parameter as well: https://docs.web3signer.consensys.io/reference/cli/subcommands#network

PR #7385, opened 2025-08-14
Speed up and simplify json serialization

arnetheduck:
* split out "json serialization" from "rest serialization" as a separate module that only implements the `RestJson` format - this way, SSZ and JSON support each has its own module while "rest" serialization combines the two
* set unknown/optional/required field options on the `RestJson` flavor directly
* switch to `Opt` throughout and make sure json_serialization `results` support is enabled - this makes the framework consistently skip `none` whem writing fields which allows simplifying the code
* stream json text directly instead of first allocating temporary strings for hex arrays/numbers - 2-3x performance improvement on state endpoint, ie down from 6-7s to 2-3s for a `curl` download - a json state is now >800mb!
* reduce code duplication, import lists and binary size by exploiting fork patterns and auto-generating serialization for more types
* read "dynamically typed" JSON fields like `data` using custom `object` with `JsonString` instead of using hand-written parser - the framework-generated parser is generally more efficient
* add Web3Signer decoding/roundtrip tests based on yaml spec examples
* print `ConsensusFork` like it's done in the beacon API spec throughout
* make SSZ `SignedBeaconBlock` root updates safer via ssz_serialization bump
* make JSON `SignedBeaconBlock` root updates safer by enabling them for all root-carrying types

The remote keystore serialization code could likely be simplified in a similar way with additional tests.


arnetheduck:
Needs https://github.com/status-im/nim-ssz-serialization/pull/116

tests/test_eth2_rest_serialization.nim
```diff
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2021-2025 Status Research & Development GmbH
+# Copyright (c) 2023-2024 Status Research & Development GmbH
```
tersec: ?

cheatfate:
It would be proof of implementation to test newly introduced serialization procedures using not-in-order scheme like:
```
{
  data: {...}
  version: ...
}
```
there probably no such tests yet, but at least previous manual deserialization procedures was able to decode it.

arnetheduck:
> using not-in-order scheme

Out-of-order fields are part of the JSON standard and therefore [tested](https://github.com/status-im/nim-json-serialization/blob/e076cb9a2dff30e13aad0bb4a13049da86d07967/tests/test_serialization.nim#L644) in the upstream json-ser library (in various constellations), including its conformance tests.

The [automatically generated code](https://github.com/status-im/nim-json-serialization/blob/master/json_serialization/reader_impl.nim#L195) for this uses a bit field and raises an exception as it should when field is double (or missing, where relevant), and is _typically_ faster than the hand-written version for two reasons:

* it uses a bit field for field presence instead of `Opt` which is more compact and sometimes can avoid a copy of the data (varies between nim versions and compiler options)
* it uses the declared field order as a hint for which field is _likely_ to come next - ie in 99% of cases, the order will be `version,data` so if you know you just matched "version", the next field is likely to be "data"

beacon_chain/spec/eth2_apis/eth2_rest_json_serialization.nim
```diff
@@ -0,0 +1,1353 @@
...
+template writeValue*(w: RestJsonWriter, value: tuple) =
+  writeRecordValue(w, value)
+
+## https://github.com/ethereum/beacon-APIs/blob/v3.1.0/types/primitive.yaml#L31
+proc writeValue*(
+    w: var RestJsonWriter, value: uint64 | uint32 | uint16 | uint8
+) {.writer.} =
+  w.streamElement(s):
+    s.write('"')
+    s.writeText(value)
```
cheatfate, line 331: > stream json text directly instead of first allocating temporary strings for hex arrays/numbers - 2-3x performance improvement on state endpoint, ie down from 6-7s to 2-3s for a curl download - a json state is now >800mb!

This statement is not true here, because procedure `writeText()` just use `$` on this intereger, which means that string object is being allocated and numbers being added to this string first. Also i think numbers is one of the most heavy used types when producing JSON.

arnetheduck: See https://github.com/status-im/nim-faststreams/blob/a9c6b884b0971c03e2251094ac88d1ecbc2c20bb/faststreams/textio.nim#L46 which does the formatting in this case.

The performance numbers are based on measuring the `/eth/v2/debug/beacon/states/finalized` REST endpoint using `curl` in a loop, also with `vtune` attached:

image

The remaining bottleneck is that we make several copies of the json output when it traverses the various layers - in the case of a beacon state, we spend ~28% of cpu time doing this - if this was fixed, we'd probably be below 2s for sending the state - it would need something like https://github.com/status-im/nim-chronos/issues/578 to fix while maintaining the ergonomy.

cheatfate:
> > using not-in-order scheme
> 
> Out-of-order fields are part of the JSON standard and therefore [tested](https://github.com/status-im/nim-json-serialization/blob/e076cb9a2dff30e13aad0bb4a13049da86d07967/tests/test_serialization.nim#L644) in the upstream json-ser library (in various constellations), including its conformance tests.
> 
> The [automatically generated code](https://github.com/status-im/nim-json-serialization/blob/master/json_serialization/reader_impl.nim#L195) for this uses a bit field and raises an exception as it should when field is double (or missing, where relevant), and is _typically_ faster than the hand-written version for two reasons:
> 
>     * it uses a bit field for field presence instead of `Opt` which is more compact and sometimes can avoid a copy of the data (varies between nim versions and compiler options)
> 
>     * it uses the declared field order as a hint for which field is _likely_ to come next - ie in 99% of cases, the order will be `version,data` so if you know you just matched "version", the next field is likely to be "data"

Previous versions of nim-json-serialization was unable to handle this type of JSON objects, so i'm a little bit suspicious about all of it.

arnetheduck:
> Previous versions of nim-json-serialization was unable to handle this type of JSON objects, so i'm a little bit suspicious about all of it.

Specifically, json-ser was (and still is) unable to deal with `case` objects, which is why the manual parsing was done.

In the new code, we still have a 2-step process to parse these objects - first, we parse `version` and `data` and then we do a second decoding on `data` - the new trick is to use `JsonString` to store `data` without interpreting it.

In fact, parsing the "raw" version/data objects is no different from parsing a `BlockBody` or any other object in which there are no `case` fields - we've always handled out-of-order fields for those "simple" objects and that's what we use here to avoid some of the hand-written code.

The optimal version of the above code would actually not use a `JsonString` at all - instead, it would remember the begin/end offsets of `data` and then revisit them when `version` is known - in particular, it would do this as soon as `version` is known so in the case where version arrives first in the json stream, we would only make one pass over `data` - this is a significant change, but also an interesting one.

cheatfate:
> > Previous versions of nim-json-serialization was unable to handle this type of JSON objects, so i'm a little bit suspicious about all of it.
> 
> Specifically, json-ser was (and still is) unable to deal with `case` objects, which is why the manual parsing was done.
> 
> In the new code, we still have a 2-step process to parse these objects - first, we parse `version` and `data` and then we do a second decoding on `data` - the new trick is to use `JsonString` to store `data` without interpreting it.
> 
> In fact, parsing the "raw" version/data objects is no different from parsing a `BlockBody` or any other object in which there are no `case` fields - we've always handled out-of-order fields for those "simple" objects and that's what we use here to avoid some of the hand-written code.
> 
> The optimal version of the above code would actually not use a `JsonString` at all - instead, it would remember the begin/end offsets of `data` and then revisit them when `version` is known - in particular, it would do this as soon as `version` is known so in the case where version arrives first in the json stream, we would only make one pass over `data` - this is a significant change, but also an interesting one.

Previous version of manual decoding was using same scheme - decoding `version` first and preserving `data` in `JsonString`. You said that this is new trick - but it is not, and its why i want to get test, at least one with case object and with changed order of fields.

arnetheduck:
> was using same scheme 

The previous code was using a hand-written scheme - here we introduce an intermediate [`VersionedData` et al](https://github.com/status-im/nimbus-eth2/pull/7385/files#diff-20d581f413fabe9c9a1808875a479b0f76d3c8b71986004b166d583bffd2492fR556) helper to remove some of the manual steps, but not all  - the code generated by json-ser is indeed semantically the same as the current manual code, but more efficient, for this step.

The code that creates the case object from version and data is also unchanged in the sense that this step is done manually.

What's different is that we [remove](https://github.com/status-im/nimbus-eth2/pull/7385/files#diff-395601d5308e54a9cf2c8120f9cfef1a6d0158437f2ac587ce2fe9dccd4e9ee6L1594) the hand-written fork selector and instead reuse the existing [`withBlck`](https://github.com/status-im/nimbus-eth2/pull/7385/files#diff-20d581f413fabe9c9a1808875a479b0f76d3c8b71986004b166d583bffd2492fR610) infrastructure to avoid repeating the same code for every fork, thus reducing the risk of copy-paste mistakes.

PR #7396, opened 2025-08-18
Block production revamp, part 2

arnetheduck:
d74a9622009a538c29b92b2d1b3aa97b4f1d5b6b reorganised the block production flow to first compute the best payload and then create the full consensus block.

However, once we had a payload, the code was still creating a "fake" non-blinded execution payload to compute the new state root causing quite a bit of complexity in the blinded pipeline.

Therefore:

* refactor the state transition to also accept blinded blocks - the state transition does not really touch execution payloads except superficially meaning that most of it doesn't have to change - this same facility could be used for replay as well, if we start loading blocks without execution payloads
* use `Forky*` consistently in block production helpers - it had grown to a mix of both forked and forky which meant extra block copies and binary bloat without any benefit
* simplify block_sim to use more of the production block flow, thus removing fork-specific code
* parse execution requests earlier in the pipeline - this is still too late, but now we can at least fall back to a builder block if local EL gives invalid requests (it really should discard invalid encodings before selecting the best EL)

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -489,7 +489,8 @@ proc getPayload*(
     withdrawals: seq[capella.Withdrawal]
 ): Future[Opt[PayloadType]] {.async: (raises: [CancelledError]).} =
   if m.elConnections.len == 0:
-    return err()
+    notice "No engine configured, using empty payload"
+    return Opt.none(PayloadType)
 
   let
     engineApiWithdrawals = toEngineWithdrawals withdrawals
--- a/beacon_chain/spec/mev/fulu_mev.nim
+++ b/beacon_chain/spec/mev/fulu_mev.nim
@@ -48,13 +48,38 @@ type
     blob_kzg_commitments*: KzgCommitments # [New in Deneb]
     execution_requests*: ExecutionRequests # [New in Electra]
 
+  SigVerifiedBlindedBeaconBlockBody* = object
+    randao_reveal*: TrustedSig
+    eth1_data*: Eth1Data
+    graffiti*: GraffitiBytes
+    proposer_slashings*: List[TrustedProposerSlashing, Limit MAX_PROPOSER_SLASHINGS]
+    attester_slashings*:
+      List[electra.TrustedAttesterSlashing, Limit MAX_ATTESTER_SLASHINGS_ELECTRA]
+    attestations*: List[electra.TrustedAttestation, Limit MAX_ATTESTATIONS_ELECTRA]
+    deposits*: List[Deposit, Limit MAX_DEPOSITS]
+    voluntary_exits*: List[TrustedSignedVoluntaryExit, Limit MAX_VOLUNTARY_EXITS]
+    sync_aggregate*: TrustedSyncAggregate
+    execution_payload_header*: fulu.ExecutionPayloadHeader
+    bls_to_execution_changes*:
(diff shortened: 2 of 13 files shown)


beacon_chain/el/el_manager.nim
```diff
@@ -489,7 +489,8 @@ proc getPayload*(
     withdrawals: seq[capella.Withdrawal]
 ): Future[Opt[PayloadType]] {.async: (raises: [CancelledError]).} =
   if m.elConnections.len == 0:
-    return err()
+    notice "No engine configured, using empty payload"
```
tersec, line 492: This only works in Bellatrix, no? Because of the withdrawals.

beacon_chain/validators/block_payloads.nim
```diff
@@ -320,21 +299,31 @@ proc getExecutionPayload*(
...
+  type PayloadType = consensusFork.ExecutionPayloadForSigning
+  let
+    eps = (
+      await node.elManager.getPayload(
+        PayloadType, beaconHead.blck.bid.root, executionHead, latestSafe,
+        latestFinalized, timestamp, random, feeRecipient, withdrawals,
+      )
+    ).valueOr:
+      if not proposalState[].is_merge_transition_complete():
+        # Per-merge, an all-zeroes execution payload is used and there are no
```
tersec: Pre-merge

beacon_chain/el/el_manager.nim
arnetheduck: Ostensibly, but this code I guess both was and is agnostic with respect to such interpretations, ie I'm not sure this is the place to check it (this particular change just adds logging, it doesn't change the previous behavior)

beacon_chain/validators/block_payloads.nim
arnetheduck:
```suggestion
        # Pre-merge, an all-zeroes execution payload is used and there are no
```

beacon_chain/spec/mev/fulu_mev.nim
```diff
@@ -48,13 +48,38 @@ type
...
+      List[electra.TrustedAttesterSlashing, Limit MAX_ATTESTER_SLASHINGS_ELECTRA]
+    attestations*: List[electra.TrustedAttestation, Limit MAX_ATTESTATIONS_ELECTRA]
+    deposits*: List[Deposit, Limit MAX_DEPOSITS]
+    voluntary_exits*: List[TrustedSignedVoluntaryExit, Limit MAX_VOLUNTARY_EXITS]
+    sync_aggregate*: TrustedSyncAggregate
+    execution_payload_header*: fulu.ExecutionPayloadHeader
+    bls_to_execution_changes*:
+      List[SignedBLSToExecutionChange,
+        Limit MAX_BLS_TO_EXECUTION_CHANGES]
+    blob_kzg_commitments*: KzgCommitments # [New in Deneb]
```
tersec, line 66: Both of these "new in" aren't salient anymore by Fulu

arnetheduck: better keep consistent with `BlindedBeaconBlockBody` from which it was copy-pasted

arnetheduck: not sure why I changed the one below though ;) copy-paste failure maybe

WTDuck2255:
Good thx

PR #7401, opened 2025-08-20
Fix `block_gossip` event did not happen when BN is optimistically synced

cheatfate:


cheatfate:
Initial assumption for this PR was wrong, closing.

PR #7413, opened 2025-08-24
create_blob_sidecars() likely doesn't work properly with Fulu/columns

tersec:
And it's not actually used in that capacity, except incidentally by imprecise fork selection.


agnxsh:
Yes, it didn't after we decided to use `engine_getBlobsV2` from v1 in Fulu

PR #7415, opened 2025-08-25
adjust fulu mev unblind condition

agnxsh:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -502,6 +502,18 @@ proc proposeBlockAux(
 
               return head
 
+            when consensusFork >= ConsensusFork.Fulu:
+              if unblindedBlockRef.isOk and unblindedBlockRef.get.isNone:
```
tersec: isn't `unblindedBlockRef.isOk` guaranteed by this point?

PR #7426, opened 2025-08-29
initial Gloas fork infrastructure

tersec:
Fairly minimal, 1000-line no-op of a commit.

Makes Gloas a copy of Fulu. Mostly fills in `spec/forks`, but wherever possible it just avoids calling anything in `withFoo` in Gloas.

From here, each step can be filled in incrementally, in parallel, without creating ramifying changes across the entire codebase.

### DIFF
--- a/beacon_chain/spec/datatypes/gloas.nim
+++ b/beacon_chain/spec/datatypes/gloas.nim
@@ -1,0 +1,646 @@
+# beacon_chain
+# Copyright (c) 2025 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [], gcsafe.}
+
+# Types specific to Fulu (i.e. known to have changed across hard forks) - see
+# `base` for types and guidelines common across forks
+
+# TODO Careful, not nil analysis is broken / incomplete and the semantics will
+#      likely change in future versions of the language:
+#      https://github.com/nim-lang/RFCs/issues/250
+{.experimental: "notnil".}
+
+import
+  std/typetraits,
+  "."/[phase0, base, bellatrix, electra, fulu],
+  chronicles,
+  json_serialization,
+  ssz_serialization/[merkleization, proofs],
+  ssz_serialization/types as sszTypes,
+  ../digest,
+  kzg4844/[kzg, kzg_abi]
+
+from stew/bitops2 import log2trunc
+from stew/byteutils import to0xHex
+from ./altair import
+  EpochParticipationFlags, InactivityScores, SyncAggregate, SyncCommittee,
+  TrustedSyncAggregate, SyncnetBits, num_active_participants
+from ./capella import
+  ExecutionBranch, HistoricalSummary, SignedBLSToExecutionChange,
+  SignedBLSToExecutionChangeList, Withdrawal, EXECUTION_PAYLOAD_GINDEX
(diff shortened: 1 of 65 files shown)


arnetheduck:
something to maybe think about before adding another fork: all the `withFoo` can essentially be replaced with a single `withData` since all the Forked* have each fork contents `forknameData`.

tersec:
This is the most incremental possible change; the point is to open Gloas development.

Once that happens, a `withData` refactoring could still happen, and it could do so incrementally.

The issue is that some PR kind of like this is necessary (with a `withData` too, by the way, it doesn't solve the issue requiring touching 60+ files, which is that otherwise random things which aren't yet defined for Gloas unsurprisingly don't build) to happen in one shot, as quickly as possible.

This means it does no extra semantic refactoring, nothing "interesting", nothing non-mechanical.

And we do need to open Gloas development, as soon as possible. This doesn't add any tech debt (no, the `deugGloasComment` stuff isn't, in that sense: the whole point is to have a "TODO"-ish thing which is an actual Nim symbol; they'll get eliminated gradually, and form a kind of roadmap of near-future Gloas PRs) overall to the current design.

beacon_chain/spec/datatypes/gloas.nim
```diff
@@ -0,0 +1,646 @@
...
+      ## Hash of execution block
+    transactions_root*: Eth2Digest
+    withdrawals_root*: Eth2Digest
+    blob_gas_used*: uint64
+    excess_blob_gas*: uint64
+
+  ExecutePayload* = proc(
+    execution_payload: ExecutionPayload): bool {.gcsafe, raises: [].}
+
+  FinalityBranch* =
```
arnetheduck, line 131: why are these here? In fact, why are they in fulu? If a type doesn't change, it shouldn't be copied (except for _maybe_ block and state that are unavoidable.

tersec: In this case it appears that it's not different since Electra, so `FinalityBranch`/`CurrentSyncCommitteeBranch`/`NextSyncCommitteeBranch` likely can be removed. But in general, these types dependent on generalized indices aren't always decided until later. For that matter, I wouldn't bet against them actually changing in Gloas, due to the depth of the tree potentially decreasing underneath a power of 2; I haven't checked.

That also means, that most/all of the types in `datatypes/gloas` will be modified or replaced, regardless, virtually everything in there now is a placeholder.

The point of this PR is to basically uncover all the `case` statements, `when`s which needed to be marked, et cetera while giving it just enough of a set of usual types that the things which should work worked. As witnessed by:
- https://github.com/status-im/nimbus-eth2/pull/7433
- https://github.com/status-im/nimbus-eth2/pull/7434
- https://github.com/status-im/nimbus-eth2/pull/7437

And there's more to go. But by doing so I already enabled:
- https://github.com/status-im/nimbus-eth2/pull/7432

Each of those separately required enabling different parts of Nimbus, and discovering which parts changes the operation-time-per-`case`-statement/expression, or per-`withFoo` (and no, it does _not_ depend on whether that's a `withData` or `withFoo`) by some constant multiplicative factor of at least 3 or 4, while being even harder to verify the correctness of.

tersec: https://github.com/status-im/nimbus-eth2/pull/7441

arnetheduck: yeah, this is more a question of cleaning up electra, then fulu, then using that cleaned-up version as a base for gloa work (and next HF), to save on the effort - ie for fulu in particular, now is probably the time to clean it up properly and remove all the cruft that was built up there needlessly as part of development.

arnetheduck:
This needs a corresponding bump on eth1, which uses some fork-specific code

tersec:
> This needs a corresponding bump on eth1, which uses some fork-specific code

- https://github.com/status-im/nimbus-eth1/pull/3631

PR #7351, opened 2025-08-02
Parallel reconstruction of columns

ahshum:

PR #7428, opened 2025-08-29
Parallel reconstruction (unstable branch)

ahshum:
- https://github.com/status-im/nimbus-eth2/pull/7351
- https://github.com/status-im/nimbus-eth2/pull/7481


tersec:
beacon_chain/gossip_processing/block_processor.nim merge conflict.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+      var
+        colInput: seq[ref DataColumnSidecar]
+        colEntries: Table[uint64, seq[MatrixEntry]]
+
+      for entry in input:
+        colEntries.mgetOrPut(entry.column_index, @[]).add(entry)
+
+      for cIdx, entries in colEntries:
+        var cells = newSeq[Cell](rowCount)
+        for entry in entries:
```
tersec: As of:
- https://github.com/ethereum/consensus-specs/pull/4519

this isn't correct anymore because it will quasi-randomly shuffle the columns. The issue is the use of `colEntries: Table[uint64, seq[MatrixEntry]]`, which doesn't really work anymore.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+
+        # convert input into column
+        var
+          colInput: seq[ref DataColumnSidecar]
+          colEntries: Table[uint64, seq[MatrixEntry]]
+
+        for entry in input:
+          colEntries.mgetOrPut(entry.column_index, @[]).add(entry)
+
+        for cIdx, entries in colEntries:
```
tersec: Similarly due to:
- https://github.com/ethereum/consensus-specs/pull/4519

These tests don't really work anymore, as written. They'll typically return that the input is invalid, regardless of whether it is, because as a first step they'll quasi-shuffle it through this `Table[uint64, seq[MatrixEntry]]`, which would cause `recover_cells_and_proofs_parallel` to fail regardless of the validity of the input data.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+      var
+        shouldSkip = false
+        invalidData: seq[MatrixEntry]
+        indices: seq[ColumnIndex]
+
+      for i in 0..<inputCells.len:
+        let
+          cellBytes = fromHex[2048](inputCells[i].getStr).valueOr:
+            # skip case of cell has length != 2048
+            # TODO: support dynamic length for future-proof
```
tersec: If the test suite changes, the test fixtures running them can change too, it's fine not to be future-proof in this instance.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+        for j in 0..<val[i].cells.len:
+          let k = i * NUMBER_OF_COLUMNS + j
+          check data[k].cell.bytes == val[i].cells[j].bytes
+          check data[k].kzg_proof.bytes == val[i].proofs[j].bytes
+
+proc runRecoverCellsAndKzgProofsParallelInvalidTest(suiteName, suitePath: string) =
+  try:
+    # Preload data of valid cases
+    let
+      validData = loadCellsAndKzgProofsValidCases(suitePath)
```
tersec, line 399: Is the name of this function really accurate, `loadCellsAndKzgProofsValidCases`, if it's used in both `runRecoverCellsAndKzgProofsParallelValidTest` and `runRecoverCellsAndKzgProofsParallelInvalidTest`?

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+      for i in 0..<inputCells.len:
+        let
+          cellBytes = fromHex[2048](inputCells[i].getStr).valueOr:
+            # skip case of cell has length != 2048
+            # TODO: support dynamic length for future-proof
+            shouldSkip = true
+            break
+          columnIdx = ColumnIndex(inputIndices[i].getInt.toUInt64.get)
+
+        # Duplicate index is not compatible with `seq[DataColumnSidecar]`
```
tersec: How is this the case?

Yes, duplicate indices indicate something wrong and the recovery should fail, but then, part of the point of the invalid tests is to test `recover_cells_and_proofs_parallel`, not to test some ad-hoc filtering code in the test runner. The `seq` itself can have duplicate indices, or indeed, entirely duplicate entries, without issue.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+            index: ColumnIndex(cIdx),
+            column: DataColumn(cells))
+          colInput.add(newClone(sidecar))
+
+        # check recovered cells and proofs
+        var tp = Taskpool.new()
+        let v = tp.recover_cells_and_proofs_parallel(colInput)
+        check v.isErr
+  except Exception:
+    debugEcho "Problem in loading KZG invalid case data"
```
tersec: Don't need `debugEcho`, since it's a `proc`. Can use `echo`.

But that's not even really the right approach. this is a test runner, not application code. If there's a problem loading the KZG invalid case data, the test has failed. Maybe it's failed because the test runner is buggy, maybe it's failed because the test data isn't available, maybe it's failed because the test data generation was botched. It doesn't really matter, swallowing the an exception [1] isn't the correct way of dealing with this.

Furthermore, `Exception` isn't the best error to catch here. If one wants as broad an umbrella as possible, `CatchableError`, but really, tests are a place it's ok for things to fail, that's the point, and any kind of broad exception handling hides/hinders that.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+        output = rowData["output"]
+      for i in 0..<output[0].len:
+        data.add(MatrixEntry(
+          cell: Cell(bytes: fromHex[2048](output[0][i].getStr).get),
+          kzg_proof: KzgProof(bytes: fromHex[48](output[1][i].getStr).get),
+          column_index: ColumnIndex(i),
+          row_index: RowIndex(rowCount)))
+      rowCount += 1
+    return data
+  except Exception:
```
tersec: See:
- https://github.com/status-im/nimbus-eth2/pull/7428/files#r2331581132

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+proc runRecoverCellsAndKzgProofsParallelValidTest(suiteName, suitePath: string) =
+  test "KZG - Recover Cells And Kzg Proofs Parallel - valid":
+    let
+      # read scenario data from valid cases
+      data = loadCellsAndKzgProofsValidCases(suitePath)
+      rowCount = data[data.len - 1].row_index.int + 1
+      # The 64 column indices
+      indices = toSeq(0 ..< (NUMBER_OF_COLUMNS div 2)).mapIt(ColumnIndex(it * 2))
+      # Minimal data for recovery
+      input = data.filterIt(it.column_index in indices)
```
tersec, line 347: This
```nim
      indices = toSeq(0 ..< (NUMBER_OF_COLUMNS div 2)).mapIt(ColumnIndex(it * 2))
      # Minimal data for recovery
      input = data.filterIt(it.column_index in indices)
```
is a roundabout way of checking whether each `it.column_index` is
- (a) even (`mod 2 == 0`),
- (b) `>= 0`, and
- (c) `< NUMBER_OF_COLUMNS`.

But (b) is already guaranteed by `ColumnIndex`, leaving (a) and (c).

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,6 +308,171 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+        continue
+
+      let
+        rowData = loadToJson(os_ops.readFile(suitePath/path/"data.yaml"))[0]
+        inputCells = rowData["input"]["cells"]
+        inputIndices = rowData["input"]["cell_indices"]
+
+      # skip case of different length of inputs as it should have covered in
+      # the single thread tests, and it is not compatible with
+      # `seq[DataColumnSidecar]`
```
tersec: The multithreaded test should be able to stand alone here, that's the point of having it.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: Sure will address this.

Regards to the test runner, it seems that try-catch is needed or it wouldn't compile when the function is out of `test:` template call. Would be great to know if there is any workaround as I prefer not to use try-catch in test as well.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: Thanks will address. Didn't realized `indices` didn't use in the following code and that made it used for filtering only.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: I think it is caused by using MatrixEntry to DataColumnSidecar conversion which makes me think it is not compatible. I think I will have to refactor the tests in order to make them more general.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: I think so as it loads only the data from valid cases. It uses in `InvalidTest` for creating the test data with only a single faulty row.

tests/consensus_spec/test_fixture_kzg.nim
tersec: - https://github.com/status-im/nimbus-eth1/pull/3658

seems to have addressed this in the `column-syncer` branch.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: Sorry for the confusion, the comment (https://github.com/status-im/nimbus-eth2/pull/7428#discussion_r2332426871) should be for this function.

Nim compiler would complain about the potential errors from `walkDir`, and it is out of the 
```
test "test case":
  # test here
```
I have to put it inside try-catch even if the function only being called inside the `test "...":` block.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: It is now used for creating the data so that the columns are sorted.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,7 +309,142 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+
+      for i in 0 ..< colCount:
+        var cells = newSeq[Cell](rowCount)
+        for j in 0 ..< rowCount:
+          let iIdx = j * colCount + i
+          cells[j] = input[iIdx].cell
+        let dataColumn = fulu.DataColumnSidecar(
+          index: indices[i],
+          column: DataColumn(cells))
+        colInput[i] = newClone(dataColumn)
```
tersec: No need to use `newClone` or first construct the object on the stack, if it's going to the heap anyway, per https://nim-lang.org/docs/manual.html#types-object-construction, e.g., something like
```nim
        colInput[i] = (ref fulu.DataColumnSidecar)(
          index: indices[i], column: DataColumn(cells))
```
The current version first constructs the stack object, then copies it. This constructs the heap object in place, without any `fulu.DataColumnSidecar` stack object having existed.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,7 +309,142 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+            # recover without any failures
+            shouldSkip = true
+            break
+          cells.add(Cell(bytes: cellBytes))
+
+        # set data column
+        let dataColumn = fulu.DataColumnSidecar(
+          index: ColumnIndex(cIdx),
+          column: DataColumn(cells))
+        colInput[i] = newClone(dataColumn)
```
tersec: No need to use `newClone` or first construct the object on the stack, if it's going to the heap anyway, per https://nim-lang.org/docs/manual.html#types-object-construction, e.g., something like
```nim
        colInput[i] = (ref fulu.DataColumnSidecar)(
          index: cIdx, column: DataColumn(cells))
```
The current version first constructs the stack object, then copies it. This constructs the heap object in place, without any `fulu.DataColumnSidecar` stack object having existed.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,7 +309,142 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+        continue
+
+      let
+        invalidData = loadToJson(os_ops.readFile(suitePath/path/"data.yaml"))[0]
+        invalidCells = invalidData["input"]["cells"]
+        invalidIndices = invalidData["input"]["cell_indices"]
+
+      # skip when there are more cells than indices as with seq[DataColumnSidecar]
+      # length of indices is always >= length of cells for each row
+      if invalidCells.len > invalidIndices.len:
```
tersec, line 417: This effectively covers https://github.com/ethereum/consensus-spec-tests/tree/v1.6.0-alpha.6/tests/general/fulu/kzg/recover_cells_and_kzg_proofs/kzg-mainnet/recover_cells_and_kzg_proofs_case_invalid_more_cells_than_cell_indices but it doesn't seem like
https://github.com/status-im/nimbus-eth2/blob/9a746492d7c08797263c658664c64b6cc4f681ca/tests/consensus_spec/test_fixture_kzg.nim#L281-L306
needs to have any special case to handle this?

In particular, while sometimes the special cases are intrinsic to the tests (arguably the non-2048-cell-bytes is, for example: in a client the serialization should filter those out automatically, or they just won't even parse if they're in any SSZ-like format), this one doesn't seem to be. But specifically for the parallel version of the test on `recover_cells_and_kzg_proofs_case_invalid_more_cells_than_cell_indices`:
```nim
      let v = tp.recover_cells_and_proofs_parallel(colInput)
      check v.isErr
```
does not, in fact, trigger `v.isErr`.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,7 +309,142 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+          for j in 0 ..< validRowCount:
+            let vIdx = NUMBER_OF_COLUMNS * j + cIdx
+            cells.add(Cell(bytes: validData[vIdx].cell.bytes))
+
+        # insert the invalid data as the last cell
+        if i < invalidCells.lenu64:
+          let cellBytes = fromHex[2048](invalidCells[i.int].getStr).valueOr:
+            # when cell is not in 2048-length, it will be in default value and
+            # recover without any failures
+            shouldSkip = true
```
tersec, line 442: https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/tests/formats/kzg_7594/recover_cells_and_kzg_proofs.md#condition documents:
> The `recover_cells_and_kzg_proofs` handler should recover missing cells and proofs, and the result should match the expected `output`. If any cell is invalid (e.g. incorrect length or one of the 32-byte blocks does not represent a BLS field element), or any `cell_index` is invalid (e.g. greater than the number of cells for an extended blob), it should error, i.e. the output should be `null`.

These aren't skip, they're intended to fail.

Also, to show/document in the code that's what it's intended to do, good to link https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/tests/formats/kzg_7594/recover_cells_and_kzg_proofs.md#condition here, i.e. to show it's not some random whim, it's how the tests work.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,7 +309,142 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+
+proc runRecoverCellsAndKzgProofsParallelInvalidTest(suiteName, suitePath: string) =
+  test "KZG - Recover Cells And Kzg Proofs Parallel - invalid":
+    # Preload data of valid cases
+    let
+      validData = loadCellsAndKzgProofsValidCases(suitePath)
+      validRowCount = validData[validData.len - 1].row_index + 1
+
+    for kind, path in walkDir(suitePath, relative = true, checkDir = true):
+      if not path.contains("_case_invalid_"):
```
tersec: The test runner shouldn't have to examine the name of the test to determine the validity or invalidity of the test. Rather, as https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/tests/formats/kzg_7594/recover_cells_and_kzg_proofs.md#condition documents:
> The `recover_cells_and_kzg_proofs` handler should recover missing cells and proofs, and the result should match the expected `output`. If any cell is invalid (e.g. incorrect length or one of the 32-byte blocks does not represent a BLS field element), or any `cell_index` is invalid (e.g. greater than the number of cells for an extended blob), it should error, i.e. the output should be `null`.

So the expected fail/not-fail condition isn't the name of the test, it's whether the expected `output` is `null` or not. These test runners shouldn't be filtering based on test name at all; those are just human-readable explanations/annotations, not semantically meaningful for the test itself.

tests/consensus_spec/test_fixture_kzg.nim
```diff
@@ -305,7 +309,142 @@ proc runRecoverCellsAndKzgProofsTest(suiteName, suitePath, path: string) =
...
+        colInput = newSeq[ref fulu.DataColumnSidecar](invalidIndices.len)
+      for i in 0 ..< colInput.lenu64:
+        let cIdx = invalidIndices[i.int].getInt.toUInt64.get
+        var cells: seq[Cell]
+
+        # insert rows from data of valid cases if it is a valid index
+        if cIdx < NUMBER_OF_COLUMNS:
+          for j in 0 ..< validRowCount:
+            let vIdx = NUMBER_OF_COLUMNS * j + cIdx
+            cells.add(Cell(bytes: validData[vIdx].cell.bytes))
```
tersec, line 431: I'm not sure I understand how this is constructed. It starts by skipping anything that's not named as an invalid test:
https://github.com/status-im/nimbus-eth2/blob/861217a45cdfb4b5d16bdcf97cf03ee79798c5f3/tests/consensus_spec/test_fixture_kzg.nim#L399-L401

But it also loads in some valid data:
https://github.com/status-im/nimbus-eth2/blob/861217a45cdfb4b5d16bdcf97cf03ee79798c5f3/tests/consensus_spec/test_fixture_kzg.nim#L394-L397

and then using that data here. Why is it mixing in some valid data to the already-constructed test case of invalid data?

tests/consensus_spec/test_fixture_kzg.nim
tersec: Yes, I see what you mean. A couple points remain though:

This is too broad an exception to catch. Specifically, one can catch instead:
```nim
  except KeyError, OSError, YamlConstructionError, YamlParserError:
```
and more importantly, `debugEcho` (or `echo`, or a log, etc) don't capture accurately what this is, which is a test runner failure. The point of a test runner is that it makes failure visible, not hiding them.

So this should signal more visibly if this happens, in a way which causes a nonzero return value for the code being run to the operating system so CI systems can detect it.

tersec: One option here is to `raises: [KeyError, OSError, YamlConstructionError, YamlParserError]`-annotate the function, to deliberately let the error leak.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: Thanks I didn't know that there is a more direct way to create ref object.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: It doesn't require special handling in `runRecoverCellsAndKzgProofsTest` because it passes cells and cell_ids directly into the function. It has to be skipped in the parallel version due to using `seq[DataColumnSidecar]`, where each column must have an index (or at least comes with a default value which is >= 0). So not able to produce the test data in this condition.

tests/consensus_spec/test_fixture_kzg.nim
ahshum: Yes, as for testing the parallel version, there should be 2 or more rows in order to effectively run parallelly. So using the valid data helps to refine the error to only the invalid row.

tests/consensus_spec/test_fixture_kzg.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7428/commits/86d534ea41ceb40ccc2982870d85e994007f32f7

tests/consensus_spec/test_fixture_kzg.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7428/commits/86d534ea41ceb40ccc2982870d85e994007f32f7

tests/consensus_spec/test_fixture_kzg.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7428/commits/86d534ea41ceb40ccc2982870d85e994007f32f7

tests/consensus_spec/test_fixture_kzg.nim
ahshum: updated in 86d534ea41ceb40ccc2982870d85e994007f32f7

tersec: How? It still seems to have exactly the same skip logic. These should register in the test summary as having been tested, and should align with an expected `output` of `null`.

ahshum: Added a check for that. I was thinking it is kind of redundant to check again as it is behind this https://github.com/status-im/nimbus-eth2/blob/db48b11df2bfd78c841faa1fd98a08fb16d72534/tests/consensus_spec/test_fixture_kzg.nim#L408

PR #7429, opened 2025-08-29
rm unused MsgTrustedSignedBeaconBlock

tersec:


beacon_chain/spec/datatypes/altair.nim
```diff
@@ -506,12 +506,6 @@ type
 
     root* {.dontSerialize.}: Eth2Digest # cached root of signed beacon block
 
-  MsgTrustedSignedBeaconBlock* = object
```
arnetheduck, line 509: how come we're not using it any more? this was for gossip-verified but otherwise untrusted, no?

tersec: The last use I could find of it historically was removed apparently incidentally, for other reasons, by:
- https://github.com/status-im/nimbus-eth2/pull/6352

Because the optimistic syncer stopped needing to handle blocks at all, whether of the `MsgTrusted` sort or otherwise.

tersec: They were introduced by:
- https://github.com/status-im/nimbus-eth2/pull/3715

i.e. the same reason. They don't appear to have ever been used for anything else.

etan-status: MsgTrusted was for blocks that were verified against block root, but where no proposer sig check was done (LC cannot do that check bc it doesn't know the pubkeys)

PR #7439, opened 2025-09-01
el_manager: clean up some engine encoding code

arnetheduck:
* encode once before making requests
* clean up obsolete error handling
* clean up web3 type usage
* get rid of fork-specific execution payloads - these were likely the result of early electra spec versions that indeed modified them and have then just been copy-pasted onwards for no particular reason


arnetheduck:
Related: https://github.com/ethereum/consensus-specs/issues/4557 - affects the spec itself also

PR #7444, opened 2025-09-03
Fix missing block hash check on engine timeout, clean up storeBlock

arnetheduck:
`ProcessingStatus` is an unused leftover from before that we were [checking block
hashes](https://github.com/status-im/nimbus-eth2/pull/4458) in the beacon node. However, we were not performing the check correctly in the case of an EL timeout.

Quarantine processing happens after block is stored -> doesn't have to live inside the forky `storeBlock` - specially since it works on forked data.


tersec:
Nim `version-2-2` CI failure is unrelated:
```
Run ${make_cmd} -j ${ncpu} NIM_COMMIT=upstream/version-2-2 ARCH_OVERRIDE=${PLATFORM} QUICK_AND_DIRTY_COMPILER=1 update
Nim Compiler Version 2.2.1 [Linux: amd64]
Compiled at 2025-02-05
Copyright (c) 2006-2025 by Andreas Rumpf

git hash: 12347eae74dbaa0e4dd99691e3a4509e8fe8b265
active boot switches: -d:release
```

PR #7445, opened 2025-09-03
rework getBlobs polling approach

agnxsh:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1929,6 +1929,46 @@ proc syncStatus(node: BeaconNode, wallSlot: Slot): string =
 when defined(windows):
   from winservice import establishWindowsService, reportServiceStatusSuccess
 
+proc attemptGetBlobs(node: BeaconNode,
+                     lastSlot: Slot) {.async.} =
+  let lastDcSidecar =
+    node.dataColumnQuarantine[].peekSidecarAtSlot(lastSlot + 1)
```
tersec: usually `.valueOr:` would go on this line, and there is room

beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -630,6 +630,41 @@ func fetchMissingSidecars*(
       res.add(BlobIdentifier(block_root: blockRoot, index: BlobIndex(bindex)))
   res
 
+proc peekSidecarAtSlot*(
```
cheatfate: I'm curious why this incredibly ineffective procedure is needed?

cheatfate:
Sorry, but this PR looks like one more crutch, could you please elaborate on what this PR fixes/introduces? Because it does use structures in most inefficient way and so introduces a lot of O(n) iterations into procedure which will be executed every slot.

agnxsh:
> Sorry, but this PR looks like one more crutch, could you please elaborate on what this PR fixes/introduces? Because it does use structures in most inefficient way and so introduces a lot of O(n) iterations into procedure which will be executed every slot.

the idea of this pr is to remove the `getBlobs` call out of the gossip processing flow, with the existing design as getBlobs is an async function, `processDataColumnSidecar` ends up creating too many futures of the getBlobs call, as much as 65 of them, occassionally in times of mev blocks or so, the blobs aren't really present in the EL pool, in those times these futures are redundant and throttle the existing column gossip processing.

there are even observations when the block is resolved, the da from the network or so is complete, however the getBlobs futures aren't really resolved.

with this change the getBlobs call would happen just once per slot, as soon as even 1 column per that block root arrives over gossip, it's expected to fetch all the data column sidecars (trusting the EL, a spec), and be able to populate quarantine, which in turn is expected to call `enqueueBlock` earlier than usual gossip validating 65 columns.

i agree `peekSidecarAtSlot` is not the best way to fetch the block root, maybe from `block_quarantine`?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2003,6 +2003,46 @@ proc syncStatus(node: BeaconNode, wallSlot: Slot): string =
...
+    let columnless = o.unsafeGet()
+    withBlck(columnless):
+      when consensusFork >= ConsensusFork.Fulu and
+           consensusFork < ConsensusFork.Gloas:
+        let
+          blobsFromElOpt=
+            await elManager.sendGetBlobsV2(forkyBlck)
+        if blobsFromElOpt.isSome():
+          let blobsEl = blobsFromElOpt.get()
+          # check lengths of array[BlobAndProofV2 with blobs
```
```suggestion
          # check lengths of array[BlobAndProofV2] with blobs
```

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2003,6 +2003,46 @@ proc syncStatus(node: BeaconNode, wallSlot: Slot): string =
...
+    return
+  let
+    elManager = node.blockProcessor[].consensusManager.elManager
+  if (let o = node.quarantine[].getColumnless(block_id.root); o.isSome):
+    let columnless = o.unsafeGet()
+    withBlck(columnless):
+      when consensusFork >= ConsensusFork.Fulu and
+           consensusFork < ConsensusFork.Gloas:
+        let
+          blobsFromElOpt=
```
```suggestion
          blobsFromElOpt =
```

beacon_chain/consensus_object_pools/block_quarantine.nim
```diff
@@ -10,7 +10,7 @@
 import
   std/tables,
   chronicles, chronos,
-  ../spec/[presets, forks]
+  ../spec/[presets, forks, block_id]
```
tersec:
```suggestion
  ../spec/[block_id, forks, presets]
```

beacon_chain/consensus_object_pools/blob_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7445/commits/02be4c464a957d375c8ec01679f2ce220b48cb16 removes it.

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7445/commits/02be4c464a957d375c8ec01679f2ce220b48cb16

PR #7450, opened 2025-09-04
Support gloas blocks in BlockRef.init

Tomi-3-0:


beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -71,25 +71,26 @@ func init*(
...
+    Opt.some blck.body.execution_payload.block_hash,
+    executionValid =
+      executionValid or blck.body.execution_payload.block_hash == ZERO_HASH,
+    blck.slot
+  )
+
+func init*(
+    T: type BlockRef, root: Eth2Digest, executionValid: bool,
+    blck: gloas.SomeBeaconBlock | gloas.TrustedBeaconBlock): BlockRef =
+  BlockRef.init(
```
tersec, line 87: debugGloasComment. This is basically broken (maybe? will have to figure out the interface we want, anyway) right now, shouldn't let it be forgotten.

Tomi-3-0: I don't think it is. The field access is now `signed_execution_payload_header.message.block_hash` instead of `execution_payload.block_hash` from previous forks.

tersec: ok

tersec: Ah, actually, I agree with you, this is likely enough not to require the `debugGloasComment`, can either merge what you have now or if you remove it, can merge that.

Tomi-3-0: Let's go with the first option, can remove it in a next PR.

PR #6682, opened 2024-10-26
make Eth2Digest isZero 8x faster

tersec:
Via profiling by @etan-status 
iszeromemory_holesky_syncing

During Holesky syncing, Nimbus is spending 25% of its time in `isZeroMemory`. This PR improves this.

```
test 1:
  PR: 2.94 microseconds/10k zero Eth2Digest comparisons
prev: 18.6 microseconds/10k zero Eth2Digest comparisons

test 2:
  PR: 2.60 microseconds/10k zero Eth2Digest comparisons
prev: 18.6 microseconds/10k zero Eth2Digest comparisons

test 3:
  PR: 1.29 microseconds/10k zero Eth2Digest comparisons
prev: 8.33 microseconds/10k zero Eth2Digest comparisons

test 4:
  PR: 1.27 microseconds/10k zero Eth2Digest comparisons
prev: 8.29 microseconds/10k zero Eth2Digest comparisons

test 5:
  PR: 1.21 microseconds/10k zero Eth2Digest comparisons
prev: 8.29 microseconds/10k zero Eth2Digest comparisons

test 6:
  PR: 2.79 microseconds/10k zero Eth2Digest comparisons
prev: 18.7 microseconds/10k zero Eth2Digest comparisons

test 7:
  PR: 2.82 microseconds/10k zero Eth2Digest comparisons
prev: 18.7 microseconds/10k zero Eth2Digest comparisons

test 8:
  PR: 2.85 microseconds/10k zero Eth2Digest comparisons
prev: 18.7 microseconds/10k zero Eth2Digest comparisons

test 9:
  PR: 1.28 microseconds/10k zero Eth2Digest comparisons
prev: 8.28 microseconds/10k zero Eth2Digest comparisons

test 10:
  PR: 1.27 microseconds/10k zero Eth2Digest comparisons
prev: 8.32 microseconds/10k zero Eth2Digest comparisons
```
using
```nim
import "."/beacon_chain/spec/digest
import std/[strformat, times]
var aZeroHash = default(Eth2Digest)

func f(d: float): string =
  fmt"{d:.3}"

template measure(s: static string, code: untyped): bool =
  let t1 = cpuTime()
  var res: bool
  for _ in 0.uint ..< 1000.uint:
    res = code
  block:
    let duration = (cpuTime() - t1)*1_000_000
    echo s, ": ", f(duration), " microseconds/10k zero Eth2Digest comparisons"
  res

let _ = measure("  PR") do:
  isZero(aZeroHash)

[...]

PR #7457, opened 2025-09-05
Simplify `forkchoiceUpdated` flow

arnetheduck:
With recent cleanups to payload handling, it becomes feasible to simplify the `forkchoiceUpdated` flow as well - the new flow is based on the following observations:

* when the beacon node is synced, it is likely that the execution client will be able to process the next payload promptly as well
* when both consensus and execution blocks are valid, we can start attesting, even though the corresponding `forkchoiceUpdated` has not yet happened
* when a proposal is due for the next slot, there is benefit in letting the execution client know ahead of time, but not so early that we send it stale information
* the proposal state - or indeed the state for the next slot - is being computed a second before the slot ends and provides a good opportunity for collecting proposal information for `forkchoiceUpdated`

Therefore:

* always update the local consensus head as soon as block verification is done, before sending out `forkchoiceUpdated` - this allows attestation work to start
* update execution head lazily while attestations (potentially) are being sent out
* in the unhappy case, keep recomputing consensus for every block that gets invalidated until the usual block processing deadline ends (instead of a fixed number of attempts)
* keep track of optimistic head status - if the execution client says the optimistic head is invalid, stop using it in fork choice updates - this can happen in the unlikely case that the sync committee votes for an invalid execution block (!) and is a defence of last resort
* introduce `OptimisticStatus` based on optimistic spec terminology, as a replacement for NewPayloadStatus
* keep track of detailed optimistic state in `BlockRef` - this allows reporting unusual state transitions more consistently than the existing "expectValid" flow that only happens with new blocks
* prevent "payload forkchoiceUpdated" from happening concurrently with "regular forkchoiceUpdated"
* reduce duplicated/overlapping logic between block processor and consensus manager, letting the latter do all consensus-related work including notifying the EL
* update clearance state more aggressively when we're synced, ie even if we have to reorg it

### DIFF
--- a/beacon_chain/consensus_object_pools/block_dag.nim
+++ b/beacon_chain/consensus_object_pools/block_dag.nim
@@ -14,6 +14,13 @@ import
 export chronicles, forks
 
 type
+  OptimisticStatus* {.pure.} = enum
+    # A simplified version of `PayloadStatusV1`
+    # https://github.com/ethereum/consensus-specs/blob/927073b0aafc958aef4689010fb4f97d22813015/sync/optimistic.md?plain=1#L55
+    notValidated
+    valid
+    invalidated
+
   BlockRef* = ref object
     ## Node in object graph guaranteed to lead back to finalized head, and to
     ## have a corresponding entry in database.
@@ -31,7 +38,7 @@ type
       ## Root that can be used to retrieve block data from database
 
     executionBlockHash*: Opt[Eth2Digest]
-    executionValid*: bool
+    optimisticStatus*: OptimisticStatus
 
     parent*: BlockRef ##\
       ## Not nil, except for the finalized head
@@ -50,48 +57,51 @@ template root*(blck: BlockRef): Eth2Digest = blck.bid.root
 template slot*(blck: BlockRef): Slot = blck.bid.slot
 
 func init*(
-    T: type BlockRef, root: Eth2Digest,
-    executionBlockHash: Opt[Eth2Digest], executionValid: bool, slot: Slot):
-    BlockRef =
+    T: type BlockRef,
+    root: Eth2Digest,
+    executionBlockHash: Opt[Eth2Digest],
+    optimisticStatus: OptimisticStatus,
+    slot: Slot,
+): BlockRef =
   BlockRef(
     bid: BlockId(root: root, slot: slot),
-    executionBlockHash: executionBlockHash, executionValid: executionValid)
+    executionBlockHash: executionBlockHash,
+    optimisticStatus: optimisticStatus,
+  )
 
 func init*(
-    T: type BlockRef, root: Eth2Digest, executionValid: bool,
+    T: type BlockRef, root: Eth2Digest, _: OptimisticStatus,
     blck: phase0.SomeBeaconBlock | altair.SomeBeaconBlock |
           phase0.TrustedBeaconBlock | altair.TrustedBeaconBlock): BlockRef =
   # Use same formal parameters for simplicity, but it's impossible for these
   # blocks to be optimistic.
-  BlockRef.init(root, Opt.some ZERO_HASH, executionValid = true, blck.slot)
+  BlockRef.init(root, Opt.some ZERO_HASH, OptimisticStatus.valid, blck.slot)
 
 func init*(
-    T: type BlockRef, root: Eth2Digest, executionValid: bool,
+    T: type BlockRef, root: Eth2Digest, optimisticStatus: OptimisticStatus,
     blck: bellatrix.SomeBeaconBlock | bellatrix.TrustedBeaconBlock |
           capella.SomeBeaconBlock | capella.TrustedBeaconBlock |
           deneb.SomeBeaconBlock | deneb.TrustedBeaconBlock |
           electra.SomeBeaconBlock | electra.TrustedBeaconBlock |
           fulu.SomeBeaconBlock | fulu.TrustedBeaconBlock): BlockRef =
   BlockRef.init(
-    root,
-    Opt.some blck.body.execution_payload.block_hash,
-    executionValid =
-      executionValid or blck.body.execution_payload.block_hash == ZERO_HASH,
-    blck.slot
+    root, Opt.some blck.body.execution_payload.block_hash, optimisticStatus, blck.slot
   )
 
 func init*(
-    T: type BlockRef, root: Eth2Digest, executionValid: bool,
+    T: type BlockRef, root: Eth2Digest, optimisticStatus: OptimisticStatus,
     blck: gloas.SomeBeaconBlock | gloas.TrustedBeaconBlock): BlockRef =
   debugGloasComment "is this right?"
   BlockRef.init(
     root,
     Opt.some blck.body.signed_execution_payload_header.message.block_hash,
-    executionValid =
-      executionValid or
-      blck.body.signed_execution_payload_header.message.block_hash ==
-      ZERO_HASH,
-    blck.slot)
+    if optimisticStatus == OptimisticStatus.valid or
+        blck.body.signed_execution_payload_header.message.block_hash == ZERO_HASH:
+      OptimisticStatus.valid
+    else:
+      optimisticStatus,
+    blck.slot,
+  )
 
 func parent*(bs: BlockSlot): BlockSlot =
   ## Return a blockslot representing the previous slot, using the parent block
@@ -242,6 +252,24 @@ func shortLog*(v: BlockSlot): string =
     shortLog(v.blck)
   else: # There was a gap - log it
     shortLog(v.blck) & "@" & $v.slot
+
+func executionValid*(blck: BlockRef): bool =
+  blck.optimisticStatus == OptimisticStatus.valid
+
+proc markExecutionValid*(blck: BlockRef, valid: bool) =
+  ## Mark a block as having a valid or invalid excecution payload
+  if not valid:
+    blck.optimisticStatus = OptimisticStatus.invalidated
+    debug "Optimistic status updated", blck = shortLog(blck), valid
+  else:
+    # Being valid implies that the ancestors are also valid
+    var cur = blck
+
+    while cur != nil and cur.optimisticStatus == OptimisticStatus.notValidated:
+      cur.optimisticStatus = OptimisticStatus.valid
+      debug "Optimistic status updated", blck = shortLog(cur), valid
+
+      cur = cur.parent
 
 chronicles.formatIt BlockSlot: shortLog(it)
 chronicles.formatIt BlockRef: shortLog(it)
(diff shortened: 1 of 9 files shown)


beacon_chain/spec/beaconstate.nim
```diff
@@ -1434,20 +1434,20 @@ template get_updated_effective_balance*(
...
   var
     withdrawal_index = state.next_withdrawal_index
-    validator_index {.inject.} = state.next_withdrawal_validator_index
+    validator_index = state.next_withdrawal_validator_index
     withdrawals: seq[Withdrawal] = @[]
   for _ in 0 ..< bound:
     let
       validator = state.validators[validator_index]
-      balance = fetch_balance
+      balance = state.balances[validator_index]
```
tersec, line 1450: Hm, this whole template injection/DI section was to support
https://github.com/status-im/nimbus-eth2/blob/7070fa82da9bbc901c9c2eaddb917451d46c5ae8/beacon_chain/spec/state_transition_epoch.nim#L1703-L1797
and related infrastructure, which was meant to support a couple of things it's been a challenge for Nimbus to hitherto support.

But while that worked in Deneb more or less, it never made it into production, and Electra and Fulu keep adding more things which make such a solution less maintainable, because it has to be a kind of shadow-epoch-transition. There were various places in the tests, of course, which also tested that, but became less relevant. It might be reasonable to simply remove it as an unviable strategy.

The issue here is that there are a few use cases of this, with the current Nimbus design:
- https://ethereum.github.io/beacon-APIs/#/Builder/getNextWithdrawals beacon API function. Not salient anymore, deprecated, will be removes shortly;
- the fast-path vs expensive fallback heuristics in https://github.com/status-im/nimbus-eth2/blob/7070fa82da9bbc901c9c2eaddb917451d46c5ae8/beacon_chain/nimbus_beacon_node.nim#L1464-L1545 could simply vanish (except for some basic ones to ensure it doesn't try operating in a too-old/unsupported-for-this fork) and enable that to always work;
- the `payload_attributes` SSE events could work without epoch transition overhead more easily even for slot 0 of each epoch; and
- it'd help with the slot 0 proposals too.

But that information (two main parts: _effective_ balances after the transition and next withdrawals) can be captured in other ways, e.g., when a block is in slot 31 of some epoch, part of what happens is that it runs that state transition function, and could capture that information then and cache it. It's ugly in its way, but I think less ugly than either restructuring or doing template/DI games across the entire epoch transition and more testably by reusing the exact same epoch transition code.

arnetheduck: oh, I didn't really look into that, just that it was unused for this particular fork and slightly simplified an error message I encountered half-way through - it made it to this pr mostly by accident so I can just remove the change altogether from here.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -14,6 +14,13 @@ import
 export chronicles, forks
 
 type
+  OptimisticStatus* {.pure.} = enum
+    # A simplified version of `PayloadStatusV1`
+    # https://github.com/ethereum/consensus-specs/blob/927073b0aafc958aef4689010fb4f97d22813015/sync/optimistic.md?plain=1#L55
```
tersec: https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/sync/optimistic.md#helpers has a nicer version of this URL's contents an doesn't point to a random-looking commithash

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -323,129 +282,259 @@ proc getFeeRecipient*(
...
+    of OptimisticStatus.notValidated:
+      if head.blck.optimisticStatus != OptimisticStatus.notValidated:
+        info "Previously validated block not accepted as new head by execution client",
+          blck = head.blck,
+          prevStatus = head.blck.optimisticStatus,
+          payloadExecutionStatus = status
+      true
+    of OptimisticStatus.invalidated:
+      if head.blck.executionValid:
+        # https://github.com/ethereum/consensus-specs/blob/927073b0aafc958aef4689010fb4f97d22813015/sync/optimistic.md?plain=1#L420
```
tersec: Nicer URL: https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/sync/optimistic.md#transitioning-from-valid---invalidated-or-invalidated---valid

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -323,129 +282,259 @@ proc getFeeRecipient*(
...
+        )
+        beaconHead = self.attestationPool[].getBeaconHead(head)
+        headBlockHash = self.dag.loadExecutionBlockHash(beaconHead.blck).valueOr:
+          return
+
+      if headBlockHash.isZero:
+        return
+
+      when consensusFork >= ConsensusFork.Deneb:
+        # https://github.com/ethereum/execution-apis/blob/90a46e9137c89d58e818e62fa33a0347bba50085/src/engine/prague.md
```
tersec: This URL finally can be replaced with a version tag rather than a commithash: https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/prague.md

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -323,129 +282,259 @@ proc getFeeRecipient*(
...
+    slot: Slot,
+    headBlockHash, safeBlockHash, finalizedBlockHash: Eth2Digest,
+    deadline: DeadlineFuture,
+    retry: bool,
+): Future[PayloadExecutionStatus] {.async: (raises: [CancelledError]).} =
+  ## Call non-proposer version of forkchoiceUpdated using the given slot to
+  ## select the correct PayloadAttributes version
+  withConsensusFork(self[].dag.cfg.consensusForkAtEpoch(slot.epoch)):
+    when consensusFork >= ConsensusFork.Bellatrix:
+      if headBlockHash.isZero:
```
tersec, line 385: Various other ` == ZERO_HASH` comparisons in this PR could probably be helped by this. That `== ZERO_HASH` is surprisingly slow, as we found in:

beacon_chain/consensus_object_pools/consensus_manager.nim
```diff
@@ -323,129 +282,259 @@ proc getFeeRecipient*(
...
+        head.finalizedExecutionBlockHash, deadline, retry,
+      )
+
+    case status.to(OptimisticStatus)
+    of OptimisticStatus.valid:
+      head.blck.markExecutionValid(true)
+      true
+    of OptimisticStatus.notValidated:
+      if head.blck.optimisticStatus != OptimisticStatus.notValidated:
+        info "Previously validated block not accepted as new head by execution client",
```
tersec, line 457: There's a theme with all these EL-returned-`INVALID` log messages: it would be good to have a single case-sensitive substring in common with them. When supporting end-users, it's convenient to check for this class of actually-the-EL-all-along failure modes by grepping in a log, and I've usually used `INVALID` (with all caps, both because the spec does but also because it has fewer false-positives) for that.

The log messages you've added here are a bit ... diverse in their phrasing, for basically a single underlying cause (it's the EL, maybe spuriously, maybe legitimately).

The code itself wasn't perfect about this:
- `Head selection failed after invalid block, using previous head`
- `updateHeadWithExecution: attempting to recover from invalid payload`

And this PR adds:
- `Previously valid execution payload turned invalid during fork choice update - check execution client for faults and restart the beacon node`

But `invalid` as a term of art can be useful to spell the spec way, so one can ask people to grep/search for `INVALID` and get low false-positive rate. It's a quick way of figuring out reasons for certain stuck syncs, etc.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -50,48 +57,51 @@ template root*(blck: BlockRef): Eth2Digest = blck.bid.root
...
   BlockRef.init(
     root,
     Opt.some blck.body.signed_execution_payload_header.message.block_hash,
-    executionValid =
-      executionValid or
-      blck.body.signed_execution_payload_header.message.block_hash ==
-      ZERO_HASH,
-    blck.slot)
+    if optimisticStatus == OptimisticStatus.valid or
+        blck.body.signed_execution_payload_header.message.block_hash == ZERO_HASH:
```
tersec: Probably `.isZero` would be better here.

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: Another: `Light execution payload invalid - the execution client or the light client data is faulty`

It doesn't have to be `INVALID`, but `invalid` might be too broad/nonselective.

arnetheduck: In this particular case, we might be coming from `VALID` as as well, ie `VALID->SYNCING`  is also odd - that said, `PayloadExecutionStatus` is already logged as upper-case and including it in the log message gives a greppable indicator - I guess the same could be done for `OptimisticStatus`, ie upper-case its string representation which then by extension would include `INVALID` in upper case as a collateral benefit ..

beacon_chain/consensus_object_pools/block_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7457/commits/f78cb7632ce9b3362046d7d40df1dcca9f1a19d6

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7457/commits/f78cb7632ce9b3362046d7d40df1dcca9f1a19d6

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7457/commits/f78cb7632ce9b3362046d7d40df1dcca9f1a19d6

beacon_chain/consensus_object_pools/block_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7457/commits/f78cb7632ce9b3362046d7d40df1dcca9f1a19d6

beacon_chain/consensus_object_pools/consensus_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7457/commits/f78cb7632ce9b3362046d7d40df1dcca9f1a19d6

beacon_chain/spec/beaconstate.nim
tersec: - https://github.com/status-im/nimbus-eth2/pull/7475

Fine to keep this change in this PR now, since I'm removing the rest of it anyway. The design was problematic, regardless.

PR #7463, opened 2025-09-07
gloas withdrawals operation

Tomi-3-0:


beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,163 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+      w.address.data[0..19] = validator.withdrawal_credentials.data[12..^1]
+      withdrawals.add w
+      withdrawal_index = WithdrawalIndex(withdrawal_index + 1)
+    
+    if len(withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
+      break
+    validator_index = (validator_index + 1) mod num_validators
+
+  (withdrawals, 
+  processed_builder_withdrawals_count,
```
tersec: Each of `processed_builder_withdrawals_count` and `processed_partial_withdrawals_count` should be indented by an additional space to line up with `withdrawals`, not the `(`.

beacon_chain/spec/helpers.nim
```diff
@@ -569,3 +569,18 @@ func hypergeom_cdf*(k: int, population: int, successes: int, draws: int):
...
+    state: gloas.BeaconState, 
+    withdrawal: BuilderPendingWithdrawal): bool =
+  ## Check if the builder is slashed and not yet withdrawable.
+  let 
+    builder = state.validators[withdrawal.builder_index]
+    current_epoch = state.slot.epoch
+  
+  builder.withdrawable_epoch >= current_epoch or not builder.slashed
+
+func is_parent_block_full*(
```
tersec: This entire `is_parent_block_full` function signature fits easily on one line 80-character line:
```nim
func is_parent_block_full*(state: gloas.BeaconState): bool =
```

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -26,33 +26,35 @@ from ../../../beacon_chain/spec/beaconstate import
...
-  OpWithdrawalsDir])
+  OpExecutionPayloadHeaderDir, OpPayloadAttestationDir, OpProposerSlashingDir,
+  OpSyncAggregateDir, OpVoluntaryExitDir, OpWithdrawalsDir
+])
 
-debugGloasComment "ensure everything is being tested"
-#doAssert toHashSet(
-#  mapIt(toSeq(walkDir(OpDir, relative = false)), it.path)) == testDirs
+doAssert toHashSet(
+ mapIt(toSeq(walkDir(OpDir, relative = false)), it.path)) == testDirs
```
tersec: indent `mapIt` by one additional space (the default, not-align-with-something-vertically indentation happens in 2-space increments), i.e. align with `A`, not `o`.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1131,6 +1131,74 @@ func process_withdrawals*(
...
+    List[capella.Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD].init(expected_withdrawals)
+  state.latest_withdrawals_root = hash_tree_root(withdrawals_list)
+  
+  for withdrawal in expected_withdrawals:
+    let validator_index = ValidatorIndex.init(withdrawal.validator_index).valueOr:
+      return err("process_withdrawals: invalid validator index")
+    decrease_balance(state, validator_index, withdrawal.amount)
+  
+  # Update the pending builder withdrawals
+  var new_builder_withdrawals: seq[BuilderPendingWithdrawal] = @[]
```
tersec: There's no need to explicitly specify `= @[]` here, because as https://nim-lang.org/docs/manual.html#statements-and-expressions-var-statement notes, `@[]` is defined to be the default value of a `seq`. So just
```nim
  var new_builder_withdrawals: seq[BuilderPendingWithdrawal]
```
is equivalent.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1131,6 +1131,74 @@ func process_withdrawals*(
 
   ok()
 
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#modified-process_withdrawals
+func process_withdrawals*(
```
tersec:
```nim
func process_withdrawals*(state: var gloas.BeaconState): Result[void, cstring] =
```
fits in 80 characters. See https://status-im.github.io/nim-style-guide/formatting.style.html for example:

So if a function signature can naturally fit in 80 characters, it's nicer to do that.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1131,6 +1131,74 @@ func process_withdrawals*(
...
+  
+  for withdrawal in expected_withdrawals:
+    let validator_index = ValidatorIndex.init(withdrawal.validator_index).valueOr:
+      return err("process_withdrawals: invalid validator index")
+    decrease_balance(state, validator_index, withdrawal.amount)
+  
+  # Update the pending builder withdrawals
+  var new_builder_withdrawals: seq[BuilderPendingWithdrawal] = @[]
+
+  for i in 0 ..< processed_builder_withdrawals_count.int:
```
tersec: I do see that this section implements:
```python


    # Update the pending builder withdrawals
    state.builder_pending_withdrawals = [
        w
        for w in state.builder_pending_withdrawals[:processed_builder_withdrawals_count]
        if not is_builder_payment_withdrawable(state, w)
    ] + state.builder_pending_withdrawals[processed_builder_withdrawals_count:]
```
so, reasonably enough, it can't be a token-by-token recreation. But this means that it probably makes sense to make the slight further adjustment/optimization to only iterate through `min(processed_builder_withdrawals_count, state.builder_pending_withdrawal.lenu64).int` once at the beginning of the loop and skip the per-iteration `i < state.builder_pending_withdrawals.len` check.

It's true, we try generally to mirror the spec functions here, but the base Nim language already can't really do that without using some libraries we don't generally use, with these Python list comprehensions, paraphrasing it slightly more is fine here.

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,163 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+    
+    if len(withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
+      break
+    validator_index = (validator_index + 1) mod num_validators
+
+  (withdrawals, 
+  processed_builder_withdrawals_count, 
+  processed_partial_withdrawals_count)
+
+template get_expected_withdrawals_with_builder_count*(
```
tersec: Unclear if/why `get_expected_withdrawals_with_builder_count` should be exported (`*`), since nothing outside `spec/beaconstate` uses it directly. As https://status-im.github.io/nim-style-guide/language.proc.html#functions-and-procedures-languageproc notes:

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,163 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
     state.balances.item(validator_index)
 
 func get_expected_withdrawals*(
-    state: electra.BeaconState | fulu.BeaconState | gloas.BeaconState):
+    state: electra.BeaconState | fulu.BeaconState):
     seq[Withdrawal] =
   get_expected_withdrawals_with_partial_count(state)[0]
 
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#modified-get_expected_withdrawals
+template get_expected_withdrawals_with_builder_count_aux*(
```
tersec: Unclear if/why `get_expected_withdrawals_with_builder_count_aux` should be exported (`*`), since nothing outside `spec/beaconstate` uses it directly. As https://status-im.github.io/nim-style-guide/language.proc.html#functions-and-procedures-languageproc notes:

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,163 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+    validator_index {.inject.} = state.next_withdrawal_validator_index
+    withdrawals: seq[Withdrawal] = @[]
+    processed_partial_withdrawals_count: uint64 = 0
+    processed_builder_withdrawals_count: uint64 = 0
+
+  # [New in Gloas:EIP7732] 
+  # Sweep for builder payments
+  for withdrawal in state.builder_pending_withdrawals:
+    if withdrawal.withdrawable_epoch > epoch or 
+       len(withdrawals) + 1 == MAX_WITHDRAWALS_PER_PAYLOAD:
```
tersec: I, too, like the `withdrawal`/`len` vertical alignment, but currently `len` therefore has a relative alignment of 3 (not a multiple of 2 spaces). What I'll usually do in this situation to retain that but also keep the multiple-of-2 indentation is:
```nim
    if  withdrawal.withdrawable_epoch > epoch or 
        len(withdrawals) + 1 == MAX_WITHDRAWALS_PER_PAYLOAD:
```
i.e. put an additional space between `if` and `withdrawal`.

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,163 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+    epoch: Epoch, fetch_balance: untyped):
+    (seq[Withdrawal], uint64, uint64) =
+  doAssert epoch - get_current_epoch(state) in [0'u64, 1'u64]
+
+  var
+    withdrawal_index = state.next_withdrawal_index
+    validator_index {.inject.} = state.next_withdrawal_validator_index
+    withdrawals: seq[Withdrawal] = @[]
+    processed_partial_withdrawals_count: uint64 = 0
+    processed_builder_withdrawals_count: uint64 = 0
```
tersec: This whole section can be simplified to:
```nim
    withdrawals: seq[Withdrawal]
    processed_partial_withdrawals_count = 0'u64
    processed_builder_withdrawals_count = 0'u64
```
due to the omitted values already being defaults  according to https://nim-lang.org/docs/manual.html#statements-and-expressions-var-statement for `seq`.

But there's a small optimization one can make here, because it's very likely there will be all `MAX_WITHDRAWALS_PER_PAYLOAD` possible withdrawals each call, and also that there cannot be more.

So to avoid the `seq` having to keep getting resized or otherwise figure out how big it's going to end up:
```nim
    withdrawals = newSeqOfCap[Withdrawal](MAX_WITHDRAWALS_PER_PAYLOAD)
    processed_partial_withdrawals_count = 0'u64
    processed_builder_withdrawals_count = 0'u64
```
using https://nim-lang.org/docs/system.html#newSeqOfCap%2CNatural

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,163 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+  processed_partial_withdrawals_count)
+
+template get_expected_withdrawals_with_builder_count*(
+    state: gloas.BeaconState):
+    (seq[Withdrawal], uint64, uint64) =
+  get_expected_withdrawals_with_builder_count_aux(
+      state, get_current_epoch(state)) do:
+    state.balances.item(validator_index)
+
+func get_expected_withdrawals*(
```
tersec: What is the purpose of `get_expected_withdrawals` here being a wrapper function, rather than calling what's currently `get_expected_withdrawals_with_builder_count` just `get_expected_withdrawals`?

tersec:
Merge conflict with `spec/helpers.nim`.

beacon_chain/spec/beaconstate.nim
Tomi-3-0: Makes sense. This way we can allocate memory upfront for the maximum withdrawals possible.
Thanks!

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1152,18 +1151,18 @@ func process_withdrawals*(
...
-      new_builder_withdrawals.add(
-        state.builder_pending_withdrawals.item(i))
-
-  for i in processed_builder_withdrawals_count.int ..< 
-      state.builder_pending_withdrawals.len:
+  var new_builder_withdrawals: seq[BuilderPendingWithdrawal]
+
+  let processed_count = min(
+    processed_builder_withdrawals_count, 
+    state.builder_pending_withdrawals.len.uint64).int
```
tersec, line 1158: `.len.uint64` is more idiomatically in this codebase spelled `.lenu64`

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/aa098ba1c13ee2ea767dcfa4991f012f799d20d9

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1131,6 +1131,73 @@ func process_withdrawals*(
...
+    let validator_index = ValidatorIndex.init(withdrawal.validator_index).valueOr:
+      return err("process_withdrawals: invalid validator index")
+    decrease_balance(state, validator_index, withdrawal.amount)
+  
+  # Update the pending builder withdrawals
+  var new_builder_withdrawals: seq[BuilderPendingWithdrawal]
+
+  let processed_count = min(
+    processed_builder_withdrawals_count, 
+    state.builder_pending_withdrawals.len.uint64).int
```
tersec: This codebase idiomatically shortens `.len.uint64` using the `.lenu64` helper function, since `uint64` and `len` are both so ubiquitously used.

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,158 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+  # [New in Gloas:EIP7732] 
+  # Sweep for builder payments
+  for withdrawal in state.builder_pending_withdrawals:
+    if  withdrawal.withdrawable_epoch > epoch or 
+        len(withdrawals) + 1 == MAX_WITHDRAWALS_PER_PAYLOAD:
+      break
+    
+    if is_builder_payment_withdrawable(state, withdrawal):
+      let
+        validator_index = withdrawal.builder_index
```
tersec: This shadow/hides the `var` `validator_index` at the beginning of the function, which isn't per se incorrect, but a bit confusing at first to read. The executable Python spec version of this function doesn't have this problem, because it has:
```python
            total_withdrawn = sum(
                w.amount for w in withdrawals if w.validator_index == withdrawal.builder_index
            )
```
i.e. there's no plain `validator_index`, it's always referred to as `w.validator_index` or `withdrawal.builder_index`. And I think here that would be better too, because in this specific instance, creating this local variable doesn't clarify things but rather create apparent lexical ambiguity. So probably better to hew more closely to the Python executable spec here and replace
```nim
          for w in withdrawals:
            if w.validator_index == validator_index:
              res += w.amount
```
with
```nim
          for w in withdrawals:
            if w.validator_index == withdrawal.builder_index:
              res += w.amount
```
it should be equivalently fast, since `withdrawal` is already effectively a small local variable with a structure which is accessed by a known memory offset at compile-time, like the C `struct` to which it compiles, and it removes this question which otherwise first strikes of, is this `validator_index = withdrawal.builder_index` related to `validator_index {.inject.} = state.next_withdrawal_validator_index` (answer: no, no it's not).

beacon_chain/spec/beaconstate.nim
```diff
@@ -1585,17 +1585,158 @@ template get_expected_withdrawals_with_partial_count_aux*(
...
+    state: electra.BeaconState | fulu.BeaconState):
     seq[Withdrawal] =
   get_expected_withdrawals_with_partial_count(state)[0]
 
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#modified-get_expected_withdrawals
+template get_expected_withdrawals_with_builder_count_aux(
+    state: gloas.BeaconState,
+    epoch: Epoch, fetch_balance: untyped):
+    (seq[Withdrawal], uint64, uint64) =
+  doAssert epoch - get_current_epoch(state) in [0'u64, 1'u64]
```
tersec: This assertion change is kind of interesting, because it used to be there to support
https://github.com/status-im/nimbus-eth2/blob/0987bdf73365d5f5c409461a7ad2f46dfca08f1d/beacon_chain/spec/state_transition_epoch.nim#L1789-L1797
i.e. at the end of an epoch, calling `checkPerValidatorBalanceCalc(...)` would trigger the `epoch - get_current_epoch(state) == 1` case.

However, nothing now calls `checkPerValidatorBalanceCalc`, which is why this assertion didn't fire in the test suite.

The spec effectively ensures it's equal (`epoch = get_current_epoch(state)`), and that it isn't is/wasn't, but rather is provided as a separate parameter, a bit of a hack to support `checkPerValidatorBalanceCalc`.

So this seems fine, but for both
```nim
  doAssert epoch - get_current_epoch(state) in [0'u64, 1'u64]
```
here, the Electra/Fulu overload, and the Gloas overload, it's clearer then to just assert, as the way of porting over the spec as much as feasible, that equivalently,
```nim
  doAssert epoch == get_current_epoch(state)
```

If `checkPerValidatorBalanceCalc` is used again, that's fine, it's a testing-only idea, so it's safe to have the stricter assertions here, it will just trip while developing the tests.

tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/0a8ecc7943889afd2f1ffd4785c00b19b81635b3

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/0a8ecc7943889afd2f1ffd4785c00b19b81635b3

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7463/commits/0a8ecc7943889afd2f1ffd4785c00b19b81635b3

PR #7473, opened 2025-09-08
Reply with Access-Control-Expose-Headers if configured with CORS

etan-status:
When CORS is used, i.e., when the Nimbus beacon API is accessed from a web browser, by default, custom headers are not accessible from the client side JavaScript code, as they might contain sensitive info.

Because beacon-API uses custom HTTP headers, we have to send a second header to actually make them available. This PR does that for:

- Eth-Consensus-Version
- Eth-Execution-Payload-Blinded
- Eth-Execution-Payload-Value
- Eth-Consensus-Block-Value

The Access-Control-Expose-Headers header is sent if Nimbus is configured with --rest-allow-origin, matching the trigger in Presto that sends the Access-Control-Allow-Origin header.

### DIFF
--- a/beacon_chain/beacon_node.nim
+++ b/beacon_chain/beacon_node.nim
@@ -138,6 +138,9 @@ template rng*(node: BeaconNode): ref HmacDrbgContext =
 proc currentSlot*(node: BeaconNode): Slot =
   node.beaconClock.now.slotOrZero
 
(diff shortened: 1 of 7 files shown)


tersec:
```
- beacon_chain/rpc/rest_light_client_api.nim
```

PR #7476, opened 2025-09-09
Gloas/process payload attestation

Tomi-3-0:


beacon_chain/spec/helpers.nim
```diff
@@ -557,3 +560,39 @@ func is_builder_payment_withdrawable*(
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-is_parent_block_full
 func is_parent_block_full*(state: gloas.BeaconState): bool =
   state.latest_execution_payload_header.block_hash == state.latest_block_hash
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-is_valid_indexed_payload_attestation
```
tersec: Belongs in `spec/beaconstate`.

beacon_chain/spec/helpers.nim
```diff
@@ -557,3 +560,39 @@ func is_builder_payment_withdrawable*(
...
+proc is_valid_indexed_payload_attestation*(
+    state: gloas.BeaconState, 
+    indexed_payload_attestation: IndexedPayloadAttestation): bool =
+  ## Check if ``indexed_payload_attestation`` is not empty, has sorted and unique indices and has
+  ## a valid aggregate signature.
+  
+  template is_sorted(s: untyped): bool =
```
tersec: Can be replaced by https://nim-lang.org/docs/algorithm.html#isSorted%2CopenArray%5BT%5D

Notably, this differs in that regard from `is_sorted_and_unique`, which is used elsewhere:
https://github.com/status-im/nimbus-eth2/blob/bc4921b2d3e343661b26acb0a8b0dcf0e5ffdcae/beacon_chain/spec/beaconstate.nim#L631-L637
by, yeah,, the `>=` vs `>` which you picked up on. But that means one can just use the Nim standard library, which doesn't test for uniqueness. This lines up with the Python code, which only checks sortedness, not uniqueness:
```python
    if len(indices) == 0 or not indices == sorted(indices):
        return False
```

The function description in the spec does say "has sorted and unique indices", but this is the rest:
```python
    # Verify indices are non-empty and sorted
    indices = indexed_payload_attestation.attesting_indices
    if len(indices) == 0 or not indices == sorted(indices):
        return False

    # Verify aggregate signature
    pubkeys = [state.validators[i].pubkey for i in indices]
    domain = get_domain(state, DOMAIN_PTC_ATTESTER, None)
    signing_root = compute_signing_root(indexed_payload_attestation.data, domain)
    return bls.FastAggregateVerify(pubkeys, signing_root, indexed_payload_attestation.signature)
```
so sortedness-only it is.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -688,6 +688,29 @@ func process_consolidation_request*(
   discard state.pending_consolidations.add(PendingConsolidation(
     source_index: source_index.uint64, target_index: target_index.uint64))
 
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#payload-attestations
+proc process_payload_attestation*(
+    state: var gloas.BeaconState, payload_attestation: PayloadAttestation,
+    cache: var StateCache): Result[void, cstring] = 
+  # Check that the attestation is for the parent beacon block
+  let data = payload_attestation.data
```
tersec:
```nim
  template data: untyped = payload_attestation.data
```

beacon_chain/spec/validator.nim
```diff
@@ -458,6 +459,79 @@ func compute_proposer_indices*(
 
   proposerIndices
 
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-compute_balance_weighted_acceptance
+func compute_balance_weighted_acceptance(
+    state: gloas.BeaconState, index: ValidatorIndex, 
+    seed: Eth2Digest, i: uint64): bool =
+  ## Return whether to accept the selection of the validator ``index``, with probability
+  ## proportional to its ``effective_balance``, and randomness given by ``seed`` and ``i``.
+  const MAX_RANDOM_VALUE = 65536 - 1
```
tersec: small thing, but if one wants to express `2**16 - 1` even more directly mapped, it's `2^16 - 1`

beacon_chain/spec/helpers.nim
```diff
@@ -557,3 +560,39 @@ func is_builder_payment_withdrawable*(
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-is_parent_block_full
 func is_parent_block_full*(state: gloas.BeaconState): bool =
   state.latest_execution_payload_header.block_hash == state.latest_block_hash
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-is_valid_indexed_payload_attestation
+proc is_valid_indexed_payload_attestation*(
+    state: gloas.BeaconState, 
+    indexed_payload_attestation: IndexedPayloadAttestation): bool =
+  ## Check if ``indexed_payload_attestation`` is not empty, has sorted and unique indices and has
```
tersec: Better to wrap this comment block to 80 characters.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2710,3 +2710,51 @@ func can_advance_slots(
...
+  let ptc = get_ptc(state, slot, cache)
+  var attesting_indices_seq = newSeqOfCap[uint64](PTC_SIZE)
+
+  for i, index in ptc:
+    if payload_attestation.aggregation_bits[i]:
+      attesting_indices_seq.add(index.uint64)
+
+  attesting_indices_seq.sort()
+
+  var attesting_indices: List[uint64, Limit PTC_SIZE]
```
tersec: Per https://github.com/status-im/nim-ssz-serialization/blob/b85850bf638cf22fecafda08375576fb2e46624f/ssz_serialization/types.nim#L126 a `List[uint64, Limit PTC_SIZE]` is a `distinct `seq[uint64]`, which means this entire separate `var attesting_indices` loop is a pointless extra copy of, for Nim, the a data type to the effectively the same data type.

The spec code here is
```python
    attesting_indices = [
        index for i, index in enumerate(ptc) if payload_attestation.aggregation_bits[i]
    ]

    return IndexedPayloadAttestation(
        attesting_indices=sorted(attesting_indices),
        data=payload_attestation.data,
        signature=payload_attestation.signature,
    )
```
so my suggestion would be:
1. rename `attesting_indices_seq` to just `attesting_indices`;
2. remove the `var attesting_indices: List[uint64, Limit PTC_SIZE]`/`for idx in attesting_indices_seq:` section which just transforms copies `seq[uint64]` to another; and
3. change
```nim
  IndexedPayloadAttestation(
    attesting_indices: attesting_indices,
    data: payload_attestation.data,
    signature: payload_attestation.signature
  )
```
to
```nim
  IndexedPayloadAttestation(
    attesting_indices: List[uint64, Limit PTC_SIZE].init(attesting_indices),
    data: payload_attestation.data,
    signature: payload_attestation.signature
  )
```
where this `List[uint64, Limit PTC_SIZE].init(attesting_indices)` becomes at runtime a safe and no-op type conversion, rather than the separate variable and copy it is now.

This also mirrors the specs more closely.

beacon_chain/spec/validator.nim
```diff
@@ -458,6 +459,79 @@ func compute_proposer_indices*(
...
+    random_value = bytes_to_uint64(random_bytes_8)
+    effective_balance = state.validators[index].effective_balance
+
+  effective_balance.uint64 * MAX_RANDOM_VALUE >= 
+    MAX_EFFECTIVE_BALANCE_ELECTRA.uint64 * random_value
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-compute_balance_weighted_selection
+func compute_balance_weighted_selection*(
+    state: gloas.BeaconState, indices: seq[ValidatorIndex], 
+    seed: Eth2Digest, size:uint64,
```
tersec: `size: uint64`

beacon_chain/spec/validator.nim
```diff
@@ -458,6 +459,79 @@ func compute_proposer_indices*(
...
+
+  let
+    random_value = bytes_to_uint64(random_bytes_8)
+    effective_balance = state.validators[index].effective_balance
+
+  effective_balance.uint64 * MAX_RANDOM_VALUE >= 
+    MAX_EFFECTIVE_BALANCE_ELECTRA.uint64 * random_value
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-compute_balance_weighted_selection
+func compute_balance_weighted_selection*(
```
tersec: `compute_balance_weighted_selection` can be an [`iterator`](https://nim-lang.org/docs/manual.html#iterators-and-the-for-statement) because its two users are
1. a `for` loop; and
2. `compute_proposer_indices` which only cares about the first value it produces (and sets `size=1`) anywy.

Then instead of
```nim
  var 
    selected = newSeqOfCap[ValidatorIndex](size)
    i = 0'u64
  while selected.lenu64 < size:
    var next_index = i mod total
    if shuffle_indices:
      next_index = compute_shuffled_index(next_index, total, seed)
    let candidate_index = indices[next_index]
    if compute_balance_weighted_acceptance(state, candidate_index, seed, i):
      selected.add(candidate_index)
    inc i

  selected
```
the body looks more like
```nim
  var i = 0'u64
  while selected.lenu64 < size:
    var next_index = i mod total
    if shuffle_indices:
      next_index = compute_shuffled_index(next_index, total, seed)
    let candidate_index = indices[next_index]
    if compute_balance_weighted_acceptance(state, candidate_index, seed, i):
      yield candidate_index
    inc i
```
with a signature of
```nim
iterator compute_balance_weighted_selection*(
    shuffle_indices: bool): ValidatorIndex =
```

And it doesn't ever (have to) accumulate its results.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2685,7 +2685,7 @@ func latest_block_id*(state: ForkedHashedBeaconState): BlockId =
 
 func matches_block(
     state: ForkyHashedBeaconState, block_root: Eth2Digest): bool =
-  ## Return true iff the latest block applied to this state matches the given
```
tersec: All of these `iff`s are https://en.wikipedia.org/wiki/If_and_only_if and not typos

beacon_chain/spec/helpers.nim
```diff
@@ -557,3 +557,4 @@ func is_builder_payment_withdrawable*(
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-is_parent_block_full
 func is_parent_block_full*(state: gloas.BeaconState): bool =
   state.latest_execution_payload_header.block_hash == state.latest_block_hash
+
```
tersec: why the added line here?

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/ca7ea81b6f08cc49e3f428c3714d53dd628aa314

beacon_chain/spec/helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/ca7ea81b6f08cc49e3f428c3714d53dd628aa314

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/ca7ea81b6f08cc49e3f428c3714d53dd628aa314

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/ca7ea81b6f08cc49e3f428c3714d53dd628aa314

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/ca7ea81b6f08cc49e3f428c3714d53dd628aa314

beacon_chain/spec/beaconstate.nim
Tomi-3-0: Got it, thanks! 🤦🏼‍♂️

tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/e9f8a88421652fced650173e6058e3e96b49bc77

beacon_chain/spec/helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/e9f8a88421652fced650173e6058e3e96b49bc77

beacon_chain/spec/beaconstate.nim
```diff
@@ -2710,3 +2710,75 @@ func can_advance_slots(
...
+  let committees_per_slot = get_committee_count_per_slot(state, epoch, cache)
+  for committee_index in get_committee_indices(committees_per_slot):
+    let committee = get_beacon_committee(state, slot, committee_index, cache)
+    indices.add(committee)
+
+  toSeq(compute_balance_weighted_selection(
+    state, indices, seed, size=PTC_SIZE, shuffle_indices=false))
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-get_indexed_payload_attestation
+proc get_indexed_payload_attestation*(
```
tersec: This can likely be a [`func`](https://nim-lang.org/docs/manual.html#procedures-func) per https://status-im.github.io/nim-style-guide/language.proc.html#functions-and-procedures-languageproc
> Prefer `func` - use `proc` when side effects cannot conveniently be avoided.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2710,3 +2710,75 @@ func can_advance_slots(
 func can_advance_slots*(
     state: ForkedHashedBeaconState, block_root: Eth2Digest, target_slot: Slot): bool =
   withState(state): forkyState.can_advance_slots(block_root, target_slot)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-get_ptc
+proc get_ptc(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
```
tersec: This runs `toSeq(compute_balance_weighted_selection(...))` when its only user is `get_indexed_payload_attestation` which just iterates over it with `for` loop already:
```nim
  let ptc = get_ptc(state, slot, cache)
  var attesting_indices = newSeqOfCap[uint64](PTC_SIZE)

  for i, index in ptc:
    if payload_attestation.aggregation_bits[i]:
      attesting_indices.add(index.uint64)
```

so if it's slightly tweaked to just do
```nim
  var attesting_indices = newSeqOfCap[uint64](PTC_SIZE)

  for i, index in get_ptc(state, slot, cache):
    if payload_attestation.aggregation_bits[i]:
      attesting_indices.add(index.uint64)
```
then this can be
```nim
iterator get_ptc(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
    ValidatorIndex =
  ...
  for candidate_index in compute_balance_weighted_selection(
      state, indices, seed, size=PTC_SIZE, shuffle_indices=false):
    yield candidate_index
```

beacon_chain/spec/beaconstate.nim
```diff
@@ -13,7 +13,7 @@ import
   chronicles,
   "."/[eth2_merkleization, forks, signatures, validator]
 
-from std/algorithm import fill, sort
+from std/algorithm import fill, sort, isSorted
```
tersec: Prefer these lexicographically sorted, i.e. `fill, isSorted, sort`

beacon_chain/spec/validator.nim
```diff
@@ -458,6 +460,81 @@ func compute_proposer_indices*(
...
+iterator compute_balance_weighted_selection*(
+    state: gloas.BeaconState, indices: seq[ValidatorIndex], 
+    seed: Eth2Digest, size: uint64, 
+    shuffle_indices: bool): ValidatorIndex =  
+  ## Return ``size`` indices sampled by effective balance, using ``indices``
+  ## as candidates. If ``shuffle_indices`` is ``True``, candidate indices
+  ## are themselves sampled from ``indices`` by shuffling it, otherwise
+  ## ``indices`` is traversed in order.
+  let total = indices.lenu64 
+  doAssert total > 0
```
tersec, line 499: Is this always true, i.e. it's a internal logic error, not reachable by any possible network data, database state, etc?

Tomi-3-0: This function is called from `get_ptc()` with indices from `get_beacon_committee()` calls, and from `compute_proposer_indices()` with indices from `get_active_validator_indices()`. Both derive from the active validator set. Zero active validators should mean protocol failure so, internal logic error only. 

`doAssert` is appropriate.

tersec: Ok.

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/45b2ca77a93443661bc2f3a343d4ddb8be9c3fd8

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/45b2ca77a93443661bc2f3a343d4ddb8be9c3fd8

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7476/commits/45b2ca77a93443661bc2f3a343d4ddb8be9c3fd8

PR #7480, opened 2025-09-09
rework getBlobs patch to unstable

agnxsh:


beacon_chain/consensus_object_pools/block_quarantine.nim
```diff
@@ -448,4 +454,4 @@ iterator peekColumnless*(
     quarantine: var Quarantine
 ): ForkedSignedBeaconBlock {.deprecated.} =
   for k, v in quarantine.sidecarless.mpairs():
-    yield v
+    yield v
```
tersec, line 457: this removes the trailing LF character, resulting in GitHub showing that red circle with a horizontal line

PR #7482, opened 2025-09-10
Store runtime SECONDS_PER_SLOT value in ValidatorClient

etan-status:
To make SECONDS_PER_SLOT configurable, the validator client needs to be able to parse it from the config endpoint and use it for validating compatibility across different beacon nodes.

- When genesis time is initialized, also fetch and store the time config
- When checking compatibility, compare against the local time config

As the rest of the code is still based on the preset constant, it is also ensured that any non-standard time configs get rejected, same as currently done, just a bit later in the logic.

### DIFF
--- a/beacon_chain/nimbus_validator_client.nim
+++ b/beacon_chain/nimbus_validator_client.nim
@@ -71,7 +71,7 @@ proc initGenesis(
                     genesis_root = resp.data.data.genesis_validators_root
               gres.add((nodes[i], resp.data.data))
             else:
-              debug "Received unsuccessful response code", endpoint = nodes[i],
+              debug "Received unexpected response code", endpoint = nodes[i],
                     response_code = resp.status
               bres.add(nodes[i])
           elif fut.failed():
@@ -159,7 +159,7 @@ proc initTimeConfig(
           debug "Received incompatible time configuration settings",
                 endpoint = nodes[i], config = resp.data.data
       else:
-        debug "Received unsuccessful time configuration settings response code",
+        debug "Received unexpected time configuration settings response code",
               endpoint = nodes[i], response_code = resp.status
     elif fut.failed():
       let error = fut.error
--- a/beacon_chain/nimbus_validator_client.nim
+++ b/beacon_chain/nimbus_validator_client.nim
@@ -116,14 +116,14 @@ proc initTimeConfig(
 ): Future[Opt[TimeConfig]] {.async: (raises: [CancelledError]).} =
   var pendingRequests: seq[Future[RestResponse[GetSpecVCResponse]]]
   for node in nodes:
-    debug "Requesting time config", node = node
+    debug "Requesting time configuration settings", node = node
     pendingRequests.add(node.client.getSpecVC())
 
   try:
     await allFutures(pendingRequests)
   except CancelledError as exc:
     var pending: seq[Future[void]]
-    debug "Time config request was interrupted"
+    debug "Time configuration settings request was interrupted"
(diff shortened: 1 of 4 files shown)


beacon_chain/nimbus_validator_client.nim
```diff
@@ -21,12 +21,16 @@ declareGauge validator_client_node_counts,
...
+
+proc initGenesis(vc: ValidatorClientRef): Future[GenesisInfo] {.
      async: (raises: [CancelledError]).} =
   info "Initializing genesis", nodes_count = len(vc.beaconNodes)
   var nodes = vc.beaconNodes
   while true:
-    var pendingRequests: seq[Future[RestResponse[GetGenesisResponse]]]
+    var
+      pendingConfigRequests: seq[Future[RestResponse[GetSpecVCResponse]]]
+      pendingGenesisRequests: seq[Future[RestResponse[GetGenesisResponse]]]
```
tersec: These contain [information](https://ethereum.github.io/beacon-APIs/#/Beacon/getGenesis) that the validator client hadn't needed from the BN before, and not directly related to `SLOTS_PER_SECOND`. Also `TimeConfig` itself doesn't need them/have them.

etan-status: We need both the SECONDS_PER_SLOT (formerly constant), and the genesis_time (from the endpoint you link), to initialize the beacon clock.

beacon_chain/spec/presets.nim
```diff
@@ -716,6 +716,10 @@ else:
 
   # createConstantsFromPreset const_preset
 
+const
+  minSecondsPerSlot* = 1'u64
+  maxSecondsPerSlot* = int64.high.uint64 div 1_000_000_000'u64
```
tersec: Is there utility in allowing 292-year long slots?

etan-status: It's the theoretical limit that is supported without triggering overflows when converting to nanoseconds. If you prefer a more opinionated limit, which one would you like?

beacon_chain/nimbus_validator_client.nim
tersec: Oh, right, it already does a `getGenesis` here, it needs `GetSpecVCResponse` now

etan-status: Yeah, only the SECONDS_PER_SLOT from there, specifically. I think it only makes sense if both the config and the genesis can be obtained, so doing one full pass to get one and then a separate pass to get the other (from potentially a different server, as majority value is taken), seems weird.

beacon_chain/nimbus_validator_client.nim
```diff
@@ -312,8 +348,9 @@ proc asyncInit(vc: ValidatorClientRef): Future[ValidatorClientRef] {.
     else:
       notice "Cannot initialize beacon node", node = node, status = node.status
 
-  vc.beaconGenesis = await vc.initGenesis()
-  info "Genesis information", genesis_time = vc.beaconGenesis.genesis_time,
+  (vc.timeCfg, vc.beaconGenesis) = await vc.initGenesis()
+  info "Genesis information", seconds_per_slot = vc.timeCfg.SECONDS_PER_SLOT,
```
cheatfate: Could you please move `seconds_per_slot` to be 4th parameter in log statement.

etan-status:
SECONDS_PER_SLOT is preferrably taken from the first successful answer, instead of from the majority, will rework it for that.

PR #7485, opened 2025-09-11
ci: set CI_CACHE to runner.temp

siddarthkay:
When building nim compiler we pass $CI_CACHE to scripts/build_nim.sh as the 4th argument but its not set anywhere.

https://github.com/status-im/nimbus-build-system/blob/e6c2c9da39c2d368d9cf420ac22692e99715d22c/makefiles/targets.mk#L80-L88

This could be the reason for weird caching we have been seeing on runner-node-02-linux-01-eu-hel1

Explicitly setting $CI_CACHE to {{runner.temp}} ensure we use a fresh directory always so that we do not get a cached version of nim commit to build.

https://github.com/status-im/nimbus-build-system/blob/e6c2c9da39c2d368d9cf420ac22692e99715d22c/scripts/build_nim.sh#L214

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,9 @@ jobs:
 
     steps:
       - name: Fix nim cache conflicts
-        run: echo "XDG_CACHE_HOME=${{ runner.temp }}/.cache" >> $GITHUB_ENV
+        run: | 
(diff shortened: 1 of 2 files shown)


jakubgs (approved):
Indeed, this is worth a try:
https://github.com/status-im/nimbus-build-system?tab=readme-ov-file#ci_cache

PR #7490, opened 2025-09-12
make GitRevision reuseable for `nimbus-eth1`

advaita-saha:
https://github.com/status-im/nimbus-eth1/issues/3661

### DIFF
--- a/beacon_chain/buildinfo.nim
+++ b/beacon_chain/buildinfo.nim
@@ -86,5 +86,3 @@ func nimBanner*(): string =
 
 declareGauge nimVersionGauge, "Nim version info", ["version", "nim_commit"], name = "nim_version"
 nimVersionGauge.set(1, labelValues=[NimVersion, getNimGitHash()])
-
-echo GitRevision
--- a/beacon_chain/buildinfo.nim
+++ b/beacon_chain/buildinfo.nim
@@ -16,7 +16,7 @@ const sourcePath = currentSourcePath.rsplit({DirSep, AltSep}, 1)[0]
 
 proc gitFolderExists(path: string): bool {.compileTime.} =
   # walk up parent folder to find `.git` folder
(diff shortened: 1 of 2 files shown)


beacon_chain/buildinfo.nim
```diff
@@ -33,34 +33,37 @@ const
...
-        strip(
-          staticExec(
-            "git -C " & strutils.escape(sourcePath) & " rev-parse --short=8 HEAD"
-          )
-        )
-      else:
-        # otherwise we use revision number given by build system.
-        # e.g. user download from release tarball, or Github zip download.
-        "00000000"
+  GitRevision* = generateGitRevision(sourcePath)
```
kdeme: I think this `GitRevision` const should be moved to the specific version file. So that you can cleanly import `buildinfo.nim` from other projects without getting that symbol. It also avoids for somebody still using it outside of nimbus-eth2 and ending up with the wrong git repo again.

Basically, this should be a buildinfo helpers only module. Actual versions should not live here imo.

advaita-saha: Makes sense yes

advaita-saha: https://github.com/status-im/nimbus-eth2/pull/7490/commits/d6c7c93c548ccad117ecc966c1ed2fa17d221a1b

kdeme: Somewhat better but my same remark basically got transferred to `sourcePath`.

advaita-saha: Yeah can do that, but `sourcePath` terminology wise seems more suited inside buildinfo

advaita-saha: https://github.com/status-im/nimbus-eth2/pull/7490/commits/879a302d1d85d9a6928b09c06a049c31cb8227e5

Issue #7394, opened 2025-08-18
Sync gets stuck after ~a day


etan-status:
**Describe the bug**
As a sync test, I tried to full sync from genesis. This works much better now that the handling of empty-response vs disconnect is properly handled. However, after a day or two, it gets stuck and no longer progresses. Restarting Nimbus is a workaround, sync progresses normally after a restart until it gets stuck again after ~a day.

**To Reproduce**
1. Platform details (OS, architecture): Ubuntu Server
2. Branch/commit used: v25.7.1-92defe-stateofus (the latest release)
3. Commands being executed: Standard RocketPool installation, with image modified from v25.7.0 to v25.7.1.
4. Relevant log lines: 

```
rocketpool_eth2        | INF 2025-08-14 18:50:35.067+00:00 Slot start                                 topics="beacnde" nextFork=Altair:74240 sync="16d04h03m (8.13%) 8.1311slots/s (DDPPQQDDDP:1005023)" finalized=31405:31ca7316 delay=68ms812us321ns slot=12364451 epoch=386389 peers=13 head=a1cb0548:1005048
rocketpool_eth2        | INF 2025-08-14 18:50:45.138+00:00 Slot end                                   topics="beacnde" slot=12364451 nextActionWait=n/a nextAttestationSlot=-1 nextProposalSlot=-1 syncCommitteeDuties=none head=83da9473:1005137
rocketpool_eth2        | INF 2025-08-14 18:50:47.137+00:00 Slot start                                 topics="beacnde" nextFork=Altair:74240 sync="16d03h58m (8.13%) 8.1330slots/s (DPQDDDPDPD:1005151)" finalized=31409:0a63f2bb delay=138ms781us384ns slot=12364452 epoch=386389 peers=12 head=0536a409:1005154
rocketpool_eth2        | INF 2025-08-14 18:50:57.164+00:00 Slot end                                   topics="beacnde" slot=12364452 nextActionWait=n/a nextAttestationSlot=-1 nextProposalSlot=-1 syncCommitteeDuties=none head=00a6396f:1005244
rocketpool_eth2        | INF 2025-08-14 18:50:59.114+00:00 Slot start                                 topics="beacnde" nextFork=Altair:74240 sync="16d03h58m (8.13%) 8.1329slots/s (DDPDQDPDDD:1005215)" finalized=31412:258ea86b delay=115ms616us670ns slot=12364453 epoch=386389 peers=10 head=0b084bb2:1005262
[...]

PR #7491, opened 2025-09-12
Disable concurrent requests in legacy sync manager

etan-status:
This gets rid of the race that led to #7394 by disabling some of the optimizations introduced in #6722. Given that sync manager will be replaced soon for Fulu, and the long time typically needed until the issue is triggered, not sure how important / appropriate.

### DIFF
--- a/beacon_chain/sync/sync_manager.nim
+++ b/beacon_chain/sync/sync_manager.nim
@@ -27,7 +27,7 @@ const
   StatusExpirationTime* = chronos.minutes(2)
     ## Time time it takes for the peer's status information to expire.
 
-  ConcurrentRequestsCount* = 3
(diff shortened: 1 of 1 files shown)


tersec:
To what extent does it slow sync? That's one potential reason to avoid doing it.

etan-status:
On that system (Raspberry Pi), it shows around 2.5 slots per second before and after, so it doesn't really slow it there. It obviously depends on the network situation and hardware, I'm not sure. With the fix, no restarts have been needed for days, so I'm confident that this workaround resolves the linked issue with sync getting stuck. One thing that's a bit worrisome is that this has nothing to do with the issue originally being reported based on a Genesis sync, there is nothing magical about "1 day", it might happen earlier.

PR #7492, opened 2025-09-12
more rman column optimizations

agnxsh:


beacon_chain/sync/request_manager.nim
```diff
@@ -393,6 +416,7 @@ proc fetchDataColumnsFromNetwork(rman: RequestManager,
     if not(isNil(peer)):
       rman.network.peerPool.release(peer)
 
+
```
tersec: ?

beacon_chain/sync/request_manager.nim
```diff
@@ -361,8 +388,20 @@ proc fetchDataColumnsFromNetwork(rman: RequestManager,
...
             filterIt(it.indices.asSeq, it in intersection))))
         .filterIt(it.indices.len > 0)
-      debug "Requesting data columns by root", peer = peer, columns = shortLog(intColIdList),
-                                                      peer_score = peer.getScore()
+
+      if intColIdList.len == 0:
+        debug "No intersecting custody columns to request",
+          peer = peer, columns = shortLog(intColIdList),
+          peer_score = peer.getScore()
+        peer.updateScore(PeerScoreIrrelevantColumnIntersection)
```
tersec: This seems to penalize a peer which is only useless due to a particular remaining set of columns required. if columns 1, 2, 3, and 4 have already been obtained, and 5, 6, 7, and 8 remain, a peer which had all of 1 through 4 would be penalized, and I'm not sure that's useful, since they'd still have 50% custody overlap.

tersec: And this applies to some extent to releasing the peer; the question is how likely the peer will be in the future, and that's more a function of general overlap than what happens this cycle.

That said, there needs to be a way to ensure relevant peers are acquired/irrelevant ones released.

beacon_chain/networking/peer_scores.nim
```diff
@@ -35,8 +35,16 @@ const
     ## Peer's answer to our request is fine.
   PeerScoreBadValues* = -1000
     ## Peer's response contains incorrect data.
-  PeerScoreBadColumnIntersection* = -2
+  PeerScoreBadColumnIntersection* = -10
     ## Peer custodies irrelevant custody columns
+  PeerScoreIrrelevantColumnIntersection* = -100
```
tersec: In general with all of these, does it makes sense to intermix the column/column-custody specific ones with general peer scores? Except for columns, it's ok for them not to overlap. Nimbus needs some of those, but not every peer needs to have column-overlap.

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7492/commits/5063fc8be1716d636df963598a2acab259613abc

tersec:
Some of the more general concerns still apply (i.e. the peer scores being affected by incidental intersections or lack thereof which don't really speak to the peer in general), but it's a clear net improvement.

PR #7507, opened 2025-09-18
attempt to solve columndb issue + fix cgc with non-validator issue

agnxsh:


tersec:
```
[2025-09-18T12:32:17.681Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7507/beacon_chain/validators/message_router.nim(183, 47) Error: invalid indentation
[2025-09-18T12:32:17.681Z] make: *** [Makefile:451: mev_mock] Error 1
[2025-09-18T12:32:17.681Z] make: *** Waiting for unfinished jobs....
```

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -758,6 +754,8 @@ proc storeBlock(
 
   # write data columns now that block has been written
   let data_columns = dataColumnsOpt.valueOr: DataColumnSidecars @[]
+  debug "Inserting columns into database ",
```
tersec: No need for trailing space in structured logging such as chronicles

tersec:
```suggestion
  debug "Inserting columns into database",
```

tersec: https://github.com/status-im/nimbus-eth2/pull/7507/commits/f2c188b7c6e490afce0d7047850f09e536c2439d

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -252,17 +251,14 @@ proc storeBackfillBlock(
...
             debug "backfill data column validation failed",
               blockRoot = shortLog(signedBlock.root),
               column_sidecar = shortLog(columns[i][]),
               blck = shortLog(signedBlock.message),
               signature = shortLog(signedBlock.signature),
               msg = r.error()
-
-      columnsOk = (dataColumnsOpt.get.lenu64 - malformed_cols.lenu64) >=
-        (self.consensusManager.dag.cfg.NUMBER_OF_COLUMNS div 2)
+          columnsOk = r.isOk()
```
ahshum, line 261: Wouldn't `columnsOk` be always the result of the last columns if without `break`?

agnxsh: that is fine actually, if it is the result of the last columns, we want the entirety of `DataColumnSidecar`to be columnsOk = true

arnetheduck: generally, these "ok" style variables with nested if conditions are pretty hard to read - avoid if possible, ie pull the logic out into its own function and return an error directly instead - see https://github.com/status-im/nimbus-eth2/pull/7518

agnxsh: the ok style variables existed before because we didn't have overloaded functions for `storeBlock` and `storeBackfillBlock`, which is why there were different decisions of block addition based on the combination of blobsOk + columnsOk. i agree, later on, once simplified cleaning up makes sense.

arnetheduck: one of the best opportunities to do cleanup work like this is when you're reading the "previous" version of the code to understand what it's doing, as part of the due diligence to add the new version - this turns the passive reading into an active endeavor which helps cement the understanding of the existing code - ie just before adding the fulu fork - this way, the existing code gets cleaned up at a regular basis and the new fork you're adding starts out clean.

PR #7516, opened 2025-09-20
add eip-7732 network topics

Tomi-3-0:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1351,6 +1351,13 @@ proc addFuluMessageHandlers(
 proc addGloasMessageHandlers(
     node: BeaconNode, forkDigest: ForkDigest, slot: Slot) =
   node.addFuluMessageHandlers(forkDigest, slot)
+  debugGloasComment "default gossipsub config"
+  node.network.subscribe(
+    getExecutionPayloadHeaderTopic(forkDigest), basicParams())
+  node.network.subscribe(
```
tersec, line 1357: This one fits in 80 characters, and can comfortably be one line.

Tomi-3-0: Yes, it does. but I wrapped it to stay consistent with the style of the other `subscribe` calls. Otherwise it breaks the visual alignment and flow (especially with it being nested) and looks a little out of place

tersec: That's a fair point, yes.

PR #7518, opened 2025-09-20
clean up `clearanceState` access

arnetheduck:
There are several assumptions in the code about the clearance state that lead to fragility - resolve some of these by moving things around a bit

* advance clearance state together with proposal fcU
* pass state in `OnBlockAdded` to emphasize the lifetime of its validity
* remove some code duplication between block_processor / sync_overseer (there's a lot more to do but it's scheduled for dedicated refactoring so do just the minimum)
* move blob/column validation to its own proc shared between backfill/head blocks

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -23,7 +23,7 @@ from ../beacon_clock import GetBeaconTimeFn, toFloatSeconds
 from ../consensus_object_pools/block_dag import
   BlockRef, OptimisticStatus, executionValid, root, shortLog, slot
 from ../consensus_object_pools/block_pools_types import
-  EpochRef, VerifierError
+  ChainDAGRef, EpochRef, OnBlockAdded, VerifierError
 from ../consensus_object_pools/block_quarantine import
   addSidecarless, addOrphan, addUnviable, pop, removeOrphan, removeSidecarless
 from ../consensus_object_pools/blob_quarantine import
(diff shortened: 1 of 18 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -171,99 +171,98 @@ proc dumpBlock[T](
...
-    signedBlock: phase0.SignedBeaconBlock | altair.SignedBeaconBlock |
-                 bellatrix.SignedBeaconBlock | capella.SignedBeaconBlock |
-                 deneb.SignedBeaconBlock | electra.SignedBeaconBlock,
-    blobsOpt: Opt[BlobSidecars]): Result[void, VerifierError] =
-  # The block is certainly not missing any more
-  self.consensusManager.quarantine[].missing.del(signedBlock.root)
+proc verifySidecars(
+    signedBlock: ForkySignedBeaconBlock,
+    blobsOpt: Opt[BlobSidecars],
+    dataColumnsOpt: Opt[DataColumnSidecars],
```
tersec: This function signature combines `blobsOpt` and `columnsOpt`, but knows (because `Forky`) statically at compile-time which it has. It's strictly an error to provide non-`none` `columnsOpt` to pre-Fulu, and likewise non-`none` `blobsOpt` to post-Fulu.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -171,99 +171,98 @@ proc dumpBlock[T](
...
-      # Track unviables so that descendants can be discarded properly
-      self.consensusManager.quarantine[].addUnviable(signedBlock.root)
-    else: discard
-    return res
+proc storeSidecars(
+    self: BlockProcessor,
+    blobsOpt: Opt[BlobSidecars],
+    dataColumnsOpt: Opt[DataColumnSidecars],
+) =
+  doAssert not (blobsOpt.isSome and dataColumnsOpt.isSome),
```
tersec: Both callers of this, `storeBlock` and `storeBackfillBlock`, know at compile time which of these is true, and this is straightforward to set up to be an error that can't occur, statically at compile-time, by only allowing the correct kind of sidecar to be provided (or not provided, per `Opt`).

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -473,6 +465,92 @@ proc enqueueQuarantine(self: var BlockProcessor, root: Eth2Digest) =
...
+          blck.message.slot, blck.root, state.current_sync_committee.pubkeys.data[i]
+        )
+
+proc verifyPayload(
+    self: ref BlockProcessor, signedBlock: ForkySignedBeaconBlock
+): Result[OptimisticStatus, VerifierError] =
+  const consensusFork = typeof(signedBlock).kind
+  # When the execution layer is not available to verify the payload, we do the
+  # required checks on the CL instead and proceed as if the EL was syncing
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/bellatrix/beacon-chain.md#verify_and_notify_new_payload
```
tersec: Nice opportunity to update this to https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/bellatrix/beacon-chain.md#verify_and_notify_new_payload

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -473,6 +465,92 @@ proc enqueueQuarantine(self: var BlockProcessor, root: Eth2Digest) =
...
+  # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/deneb/beacon-chain.md#modified-verify_and_notify_new_payload
```
tersec: And this to https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/deneb/beacon-chain.md#modified-verify_and_notify_new_payload

beacon_chain/gossip_processing/block_processor.nim
arnetheduck: > know at compile time which of these is true

sort of - they both work from two `Opt` instances whose contents are managed at runtime (also because that's how it's stored in the queue etc and the only question is where to put the `if` to pick one of them per fork (and discard / assert against the other) - if you look at https://github.com/status-im/nimbus-eth2/pull/7518/files#diff-a09c99ea4dc7b4118405b5b5c2985124f3ac10247f3d69c9611c22ca7c496f48L389, it does the same runtime->compile-time selection as this code but without the extra sanity checking only for `storeBackfillBlock`, and because it's placed at the "outer" layer, we end up duplicating all logic that does _not_ differ between the forks. It would be much worse to approach it the same way for `storeBlock` which has a lot more shared logic. I don't see the gain of doing it this way.

The aim of this PR is really to make the two `store` functions uniform and avoid the two slightly different ways of checking sidecars. 

If we want this properly fixed, what should be done is essentially `Opt[ForkedSidecar]` that only allows one of the two sidecar types to be present in the code, per fork. This is the only clean way really to achieve what you're after, ie to make the shape of the data structures match the ongoings of the protocol - that's a *much* bigger change which will be easier to do after this PR  - because once you've done that change, the code will just overload `verifySidecar` - ie the refactoring that makes `verifySidecar` a separate function is still useful in that world.

beacon_chain/gossip_processing/block_processor.nim
arnetheduck:
```suggestion
  # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/bellatrix/beacon-chain.md#verify_and_notify_new_payload
```

beacon_chain/gossip_processing/block_processor.nim
arnetheduck:
```suggestion
  # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/deneb/beacon-chain.md#modified-verify_and_notify_new_payload
```

beacon_chain/gossip_processing/block_processor.nim
arnetheduck: [401d3d5](https://github.com/status-im/nimbus-eth2/pull/7518/commits/401d3d5d19b7bd4c628a3f3fe84e962fbe21113b)

PR #7524, opened 2025-09-22
nimcrypto: bump

arnetheduck:
* optimized sha256 - not used by nimbus-eth2 yet since hashtree+blst is still selected by default - benchmarks needed


arnetheduck:
blocked by https://github.com/cheatfate/nimcrypto/issues/98

PR #7526, opened 2025-09-22
Beacon node as library

arnetheduck:
Minimal support for running the beacon node as part of another application, for example [`nimbus`](https://github.com/status-im/nimbus-eth1/pull/3646).

* move copyright out of version (set it per binary)
* avoid allocations / logging in signal handler
* expose `run` and `main` for others to call


tersec:
Merge conflict with `beacon_chain/buildinfo.nim`.

PR #7528, opened 2025-09-22
column serving issue and other fixes

agnxsh:


beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -47,7 +47,8 @@ func handle_custody_groups(cfg: RuntimeConfig, node_id: NodeId,
     custody_groups: HashSet[CustodyIndex]
     current_id = node_id
 
-  while custody_groups.lenu64 < custody_group_count:
+  let safe_count = min(custody_group_count, cfg.NUMBER_OF_CUSTODY_GROUPS)
```
tersec, line 50: https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.0/specs/fulu/das-core.md#get_custody_groups defines it as an error to do this:
```python
    assert custody_group_count <= NUMBER_OF_CUSTODY_GROUPS

    # Skip computation if all groups are custodied
    if custody_group_count == NUMBER_OF_CUSTODY_GROUPS:
        return [CustodyIndex(i) for i in range(NUMBER_OF_CUSTODY_GROUPS)]
```
but that would make things more complicated to return anything like an `Opt` or `Result`, so there needs to be some default here. But sure, it's fine to effectively set this error handling as clamping.

agnxsh: we can need the peer for blocks?

tersec: > we can need the peer for blocks?

That's compatible with not thinking it has any custody at all.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -285,7 +285,7 @@ proc storeBackfillBlock(
   # Only store data columns after successfully establishing block viability.
   let columns = dataColumnsOpt.valueOr: DataColumnSidecars @[]
   debug "Inserting columns into database (backfill)",
-    indices = columns.mapIt($it[].index)
+    indices = columns.mapIt($it[].index).len
```
arnetheduck, line 288: `mapIt` will cause the allocation of an extra seq and some extra processing - if we're just getting the length, this is wasteful

tersec: Fixed by
- https://github.com/status-im/nimbus-eth2/pull/7518

PR #7542, opened 2025-09-26
add gloas transition tests

tersec:


beacon_chain/validator_bucket_sort.nim
```diff
@@ -57,7 +58,7 @@ func sortValidatorBuckets*(validators: openArray[Validator]):
     res.bucketSorted[insertPos[]] = i.ValidatorIndex
 
   doAssert bucketInsertPositions[0] == 0
-  for i in 1 ..< NUM_BUCKETS:
+  staticFor i, 1 ..< NUM_BUCKETS:
```
arnetheduck, line 61: uh, this unrolls to ~500 lines of `doAssert`, or in other words a  pretty massive runtime blob of code with exception raising and all that does .. nothing. was that intended?

tersec: - #9086

PR #7552, opened 2025-09-27
implements process execution payload for gloas

Tomi-3-0:


beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+        return err("process_execution_payload: invalid builder index")
+      builder_pubkey = state.validators.item(builder_index).pubkey
+    if not verify_execution_payload_envelope_signature(
+        state.fork, state.genesis_validators_root, signed_envelope, state,
+        builder_pubkey, signed_envelope.signature):
+      return err("process_execution_payload: invalid envelope signature")
+  
+  # Cache latest block header state root
+  if state.latest_block_header.state_root == Eth2Digest():
+    let state_copy = new(typeof(state))
```
Tomi-3-0: The hash_tree_root computation seems to mutate the input data. To prevent corrupting the original input state, allocating a separate copy on the heap rather than the stack (which would cause stack overflow due to size of the beaconstate) seems to make sense.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+    BuilderPendingPayment()
+  
+  # Cache the execution payload hash
+  state.execution_payload_availability[
+    state.slot mod SLOTS_PER_HISTORICAL_ROOT] = true
+  state.latest_block_hash = payload.block_hash
+  
+# Verify the state root
+  if verify:
+    let state_copy = new(typeof(state))
```
Tomi-3-0: same as above

beacon_chain/spec/state_transition_block.nim
tersec: This shouldn't happen. It's worth investigating before merging this. Do you have an example of what that looks like and how to trigger it?

tersec: This shouldn't happen. It's worth investigating before merging this. Do you have an example of what that looks like and how to trigger it?

`hash_tree_root` does manage some caching (that's what `HashList` and `HashArray` exist to support), but it shouldn't corrupt the original input state.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+  if verify:
+    let 
+      builder_index = ValidatorIndex.init(envelope.builder_index).valueOr:
+        return err("process_execution_payload: invalid builder index")
+      builder_pubkey = state.validators.item(builder_index).pubkey
+    if not verify_execution_payload_envelope_signature(
+        state.fork, state.genesis_validators_root, signed_envelope, state,
+        builder_pubkey, signed_envelope.signature):
+      return err("process_execution_payload: invalid envelope signature")
+
```
tersec: All these notionally empty lines have spaces in them which shouldn't be there.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+    let state_copy = new(typeof(state))
+    state_copy[] = state
+    state.latest_block_header.state_root = hash_tree_root(state_copy[])
+
+  # Verify consistency with the beacon block
+  if envelope.beacon_block_root != hash_tree_root(state.latest_block_header):
+    return err("process_execution_payload: beacon block root mismatch")
+  if envelope.slot != state.slot:
+    return err("process_execution_payload: slot mismatch")
+
```
tersec: Here is another space

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+    return err("process_execution_payload: slot mismatch")
+  
+  # Verify consistency with the committed bid
+  let committed_bid = state.latest_execution_payload_bid
+  if envelope.builder_index != committed_bid.builder_index:
+    return err("process_execution_payload: builder index mismatch")
+  if committed_bid.blob_kzg_commitments_root != 
+      hash_tree_root(envelope.blob_kzg_commitments):
+    return err("process_execution_payload: blob KZG commitments root mismatch")
+
```
tersec: Another, etc.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+    process_withdrawal_request(cfg, state, bsv[], op, cache)
+  for op in envelope.execution_requests.consolidations:
+    process_consolidation_request(cfg, state, bsv[], op, cache)
+
+  # Queue the builder payment
+  var payment = state.builder_pending_payments.mitem(
+    (SLOTS_PER_EPOCH + (state.slot mod SLOTS_PER_EPOCH)).int)
+  let amount = payment.withdrawal.amount
+  if amount > 0.Gwei:
+    let exit_queue_epoch =
```
tersec: Trailing space which shouldn't be here.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+    cfg: RuntimeConfig, state: var gloas.BeaconState,
+    signed_envelope: SignedExecutionPayloadEnvelope,
+    notify_new_payload: deneb.ExecutePayload, cache: var StateCache,
+    verify: bool = true): Result[void, cstring] =
+  template envelope: untyped = signed_envelope.message
+  template payload: untyped = envelope.payload
+
+  # Verify signature
+  if verify:
+    let
```
tersec: Trailing space.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+  if envelope.beacon_block_root != hash_tree_root(state.latest_block_header):
+    return err("process_execution_payload: beacon block root mismatch")
+  if envelope.slot != state.slot:
+    return err("process_execution_payload: slot mismatch")
+  
+  # Verify consistency with the committed bid
+  let committed_bid = state.latest_execution_payload_bid
+  if envelope.builder_index != committed_bid.builder_index:
+    return err("process_execution_payload: builder index mismatch")
+  if committed_bid.blob_kzg_commitments_root !=
```
tersec: Trailing space

beacon_chain/spec/state_transition_block.nim
tersec: It's an artifact of `reportDiff` not understanding the `HashList`/`HashArray` htr caching mechanism:
- https://github.com/status-im/nimbus-eth2/pull/7554

Once that merges, if one updates to the resulting `unstable` branch, shouldn't need the state copy workarounds.

Tomi-3-0: Thanks. That explains a lot

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/3a88527bb32b8e5cf04419e17869b3a1eb35e38b

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/56885afaea99cd46c21b3e2a06b136529db00088

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,115 @@ proc process_execution_payload*(
...
+    state.latest_block_header.state_root = previous_state_root
+
+  # Verify consistency with the beacon block
+  if envelope.beacon_block_root != hash_tree_root(state.latest_block_header):
+    return err("process_execution_payload: beacon block root mismatch")
+  if envelope.slot != state.slot:
+    return err("process_execution_payload: slot mismatch")
+
+  # Verify consistency with the committed bid
+  let committed_bid = state.latest_execution_payload_bid
```
tersec: A whole `deneb.ExecutionPayloadHeader` is a little big to pointlessly copy here. Should use `template` instead.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,115 @@ proc process_execution_payload*(
...
+  # Verify commitments are under limit
+  let blob_params = cfg.get_blob_parameters(get_current_epoch(state))
+  if lenu64(envelope.blob_kzg_commitments) > blob_params.MAX_BLOBS_PER_BLOCK:
+    return err("process_execution_payload: too many KZG commitments")
+
+  # Verify the execution payload is valid
+  if not notify_new_payload(payload):
+    return err("process_execution_payload: execution payload invalid")
+
+  var bsv = sortValidatorBuckets(state.validators.asSeq)
```
tersec: This might be as reasonable as it's going to get, but while I pretty benchmarked this so that (at the time) it took 10-15ms to set up from scratch on the machine I used, there are a couple of tricks, one of which definitely can work and the other might:

1. If `envelope.execution_requests.withdrawals.len + envelope.execution_requests.consolidations == 0`, i.e. there are neither withdrawal requests (there will essentially always be withdrawals, but that's different) nor consolidation requests, then `bsv` isn't used at all, and shouldn't be calculated at all. It's convenient here that it's a `ref` type return because `nil` is a valid default value there, so can set it to `sortValidatorBuckets(state.validators.asSeq)` if it'll be used at least once, and `nil` otherwise:
```nim
  let bsv =
    if  envelope.execution_requests.withdrawals.len +
        envelope.execution_requests.consolidations.len > 0:
      sortValidatorBuckets(state.validators.asSeq)
    else:
      nil
``` 

It can be `let` because the what a reference points to can be modified in Nim even if the reference itself is immutable.

See
https://github.com/status-im/nimbus-eth2/blob/c934e9facc5592c84bd968e144a9d5092a29656c/beacon_chain/spec/state_transition_block.nim#L793-L803
for an example of this. It mainly creates the separate `bsv_use` variable due to the fork-dependent complications in `process_operations()`, but here that's not necessary and likely obfuscates more than anything else.

2. Ideally, see the `process_operations` usage, this resource is shared between `process_operations` and `process_execution_payload`, but it depends how the state transition is set up, now that execution payload processing is notionally separate. This amortization across the two functions will matter more as the number of validators keeps increasing.

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -196,6 +197,36 @@ suite baseDescription & "Deposit Request " & preset():
...
+        defaultRuntimeConfig, preState, signed_envelope, executePayload, cache)
+
+  for path in walkTests(OpExecutionPayloadDir):
+    let
+      testDir = OpExecutionPayloadDir / "pyspec_tests" / path
+      inputFile =
+        if fileExists(testDir/"signed_envelope.ssz_snappy"):
+          "signed_envelope"
+        else:
+          continue
```
tersec, line 231: It seems unlike there are test cases meant to be ignored entirely? https://github.com/ethereum/consensus-specs/tree/v1.6.0-beta.0/tests/formats/operations#condition doesn't, though, specify how to handle these.

I see several test cases, e.g. https://github.com/ethereum/consensus-spec-tests/tree/v1.6.0-beta.0/tests/minimal/gloas/operations/execution_payload/pyspec_tests/incorrect_transaction_length_32_extra_bytes lack this `signed_envelope.ssz_snappy` file, and they all appear to be intended to be invalid.

Even so, the correct approach isn't to silently skip them, but to figure out what the principled failure mode is supposed to be, otherwise, it's basically testing the tests, rather than how Nimbus will actually use these. It's better to make it as realistic a feasible given how the tests work (e.g., this one fundamentally is a test of `process_execution_payload()` and nothing outside that universe matters).

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,115 @@ proc process_execution_payload*(
...
+        return err("process_execution_payload: invalid builder index")
+      builder_pubkey = state.validators.item(builder_index).pubkey
+    if not verify_execution_payload_envelope_signature(
+        state.fork, state.genesis_validators_root, signed_envelope, state,
+        builder_pubkey, signed_envelope.signature):
+      return err("process_execution_payload: invalid envelope signature")
+
+  # Cache latest block header state root
+  let previous_state_root = hash_tree_root(state)
+  if state.latest_block_header.state_root == Eth2Digest():
```
tersec: Rather than dynamically construct this `Eth2Digest()`, can use the already-available constant `ZERO_HASH`.

But as we found out in:

Nim performs these operations slowly if implemented naïvely. So even better is
```nim
  if state.latest_block_header.state_root.isZero:
```

tests/consensus_spec/gloas/test_fixture_operations.nim
Tomi-3-0: I've checked with @jtraglia about the tests missing `signed_envelope.ssz_snappy`. Confirmed this is a known issue and will be fixed in the next release. These should be skipped for now

Tomi-3-0: https://github.com/ethereum/consensus-specs/issues/4545

tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/fc8836d9626ecbeddb8a1b726d99e9c47bb316fc

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/8b22881eef7fec84b594fafe026eceeec6a6a06b

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1189,7 +1189,12 @@ proc process_execution_payload*(
   if not notify_new_payload(payload):
     return err("process_execution_payload: execution payload invalid")
 
-  var bsv = sortValidatorBuckets(state.validators.asSeq)
+  let bsv =
+    if envelope.execution_requests.withdrawals.len +
+      envelope.execution_requests.consolidations.len > 0:
```
tersec, line 1194: This is part of the `if` condition and should be indented 2 spaces deeper than the `if` body (`sortValidatorBuckets(state.validators.asSeq)`)

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+  if lenu64(envelope.blob_kzg_commitments) > blob_params.MAX_BLOBS_PER_BLOCK:
+    return err("process_execution_payload: too many KZG commitments")
+
+  # Verify the execution payload is valid
+  if not notify_new_payload(payload):
+    return err("process_execution_payload: execution payload invalid")
+
+  let bsv =
+    if envelope.execution_requests.withdrawals.len +
+      envelope.execution_requests.consolidations.len > 0:
```
tersec: This is part of the `if` condition and should be indented 2 spaces deeper than the `if` body (`sortValidatorBuckets(state.validators.asSeq)`)

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/8b22881eef7fec84b594fafe026eceeec6a6a06b

(2) can wait, it's a higher-level optimization.

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/8b22881eef7fec84b594fafe026eceeec6a6a06b

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/ed80863a58d12bdad1a0093b137bc1795671db17

tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/ed80863a58d12bdad1a0093b137bc1795671db17

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1117,6 +1117,118 @@ proc process_execution_payload*(
...
+      builder_pubkey = state.validators.item(builder_index).pubkey
+    if not verify_execution_payload_envelope_signature(
+        state.fork, state.genesis_validators_root, signed_envelope, state,
+        builder_pubkey, signed_envelope.signature):
+      return err("process_execution_payload: invalid envelope signature")
+
+  # Cache latest block header state root
+  let previous_state_root = hash_tree_root(state)
+  if state.latest_block_header.state_root.isZero:
+    state.latest_block_header.state_root = previous_state_root
```
tersec: I'm aware the spec phrases it this way, but
```nim
  let previous_state_root = hash_tree_root(state)
```
is fairly heavyweight (can be hundreds of milliseconds). So better to reorder it slightly, probably with a comment, putting
```nim
  let previous_state_root = hash_tree_root(state)
```
inside the
```nim
   if state.latest_block_header.state_root.isZero:
```
because it's not otherwise used.

It would be reasonable given how closely the rest of this hews to the spec to put a comment regarding why it's slightly modified.

tersec: Another possibility is to use a `gloas.HashedBeaconState`; it depend on the context(s) in which it's going to be called. That comes built-in with `hash_tree_root(state)` being cached.

Tomi-3-0: or we could remove the assignment to the `previous_state_root` variable completely and just do it like so:
```
if state.latest_block_header.state_root.isZero:
  state.latest_block_header.state_root = hash_tree_root(state)
```

tersec: Yeah, better probably even

Tomi-3-0: > Another possibility is to use a gloas.HashedBeaconState; it depend on the context(s) in which it's going to be called. That comes built-in with hash_tree_root(state) being cached.

This looks good, and it saves one `hash_tree_root` computation:
```
  # Cache latest block header state root
  if state.data.latest_block_header.state_root.isZero:
    state.data.latest_block_header.state_root = state.root
```
So instead of two possible computations we do one

Tomi-3-0: Not so sure about production context yet

tersec: This second `hash_tree_root`, in `verify`, as you've alluded to, is another issue here.

If one uses the current approach without `gloas.HashedBeaconState`, then if either `verify` or `state.latest_block_header.state_root.isZero` (one of which is usually set), can calculate each field of the `gloas.BeaconState` hash, and then recombine them, first in original form, then with the small modifications, and cheaply recompute the overall state root:
https://github.com/status-im/nimbus-eth2/blob/9358f3354d842dbada037e6a133a39bb0a8cef94/beacon_chain/spec/state_transition_epoch.nim#L1057-L1063

but extended to the entire `gloas.BeaconState`. This is a bit tedious to do by hand, but mechanical (and, actually, is a good candidate for a Nim macro to keep it maintainable across forks, but that's another topic here).

Another is to check that if given a `gloas.HashedBeaconState`, `htr(that state)` truly is fast. It might be, but this requires empirical verification first, e.g.,
https://github.com/status-im/nimbus-eth2/blob/9358f3354d842dbada037e6a133a39bb0a8cef94/beacon_chain/fork_choice/fork_choice.nim#L177-L187
being the usual way of measuring that in the codebase. Record `Moment.now()` at the beginning and end and print the differences (if using `log` it will require this function temporarily be `proc` because of side effects, so can also use `debugEcho` which avoids that).

Then this would have to track that when it's minimally changed between the first and second `hash_tree_root()` calls, it doesn't recompute everything but knows how to reuse what it has.

Ideally, with the combination of a `gloas.HashedBeaconState` and selective rebuilding of this cache, only recomputing either changed parts of `HashList`/HashArray` or the much smaller rest of the state, the marginal cost here can be small. Certainly it's going to be possible to incur the cost only once rather than twice, but zero times is even better, and at a scale which will matter.

Conveniently, https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/spec/state_transition.nim and https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/spec/state_transition_block.nim already operate on either `ForkedHashedBeconState` or `ForkyHashedBeaconState`, so it probably is feasible to use that here. It means the test fixture will have to compute said `hash_tree_root` itself from the test data and construct such a `gloas.HashedBeaconState` before passing it to `process_execution_payload`, but it's fine for the tests to do this.

Tomi-3-0: I've added timing measurements:

```nim

proc process_execution_payload*(
    cfg: RuntimeConfig, state: var gloas.HashedBeaconState,
    signed_envelope: SignedExecutionPayloadEnvelope,
    notify_new_payload: deneb.ExecutePayload, cache: var StateCache,
    verify: bool = true): Result[void, cstring] =
  
  let startTime = Moment.now()
  
  # Verify signature
........

  # Cache latest block header state root
  if state.data.latest_block_header.state_root.isZero:
    let hashStartTime = Moment.now()
    state.data.latest_block_header.state_root = state.root
    let hashEndTime = Moment.now()
    echo "First hash_tree_root (cached): ", hashEndTime - hashStartTime

 ......

  # Verify the state root
  if verify:
    let hashStartTime = Moment.now()
    state.root = hash_tree_root(state.data)
    let hashEndTime = Moment.now()
    echo "Second hash_tree_root (recomputed): ", hashEndTime - hashStartTime
    
    if envelope.state_root != state.root:
      return err("process_execution_payload: state root mismatch")

  let endTime = Moment.now()
  echo "Total process_execution_payload() duration: ", endTime - startTime
```

test:
```nim
  proc makeApplyExecutionPayloadCb(path: string): auto =
    return proc(
        preState: var gloas.BeaconState,
        signed_envelope: SignedExecutionPayloadEnvelope):
        Result[void, cstring] =
      let payloadValid = os_ops.readFile(
          OpExecutionPayloadDir/"pyspec_tests"/path/"execution.yaml"
        ).contains("execution_valid: true")
      var
        cache: StateCache
        hashedState = gloas.HashedBeaconState(
          data: preState,
          root: hash_tree_root(preState)
        )
      func executePayload(_: deneb.ExecutionPayload): bool = payloadValid
      result = process_execution_payload(
        defaultRuntimeConfig, hashedState, signed_envelope, executePayload, cache)
      preState = hashedState.data
```
      
 Here are the results from two runs;
 first run:
```
Second hash_tree_root (recomputed): 15ms971us524ns
Total process_execution_payload() duration: 18ms59us226ns
Second hash_tree_root (recomputed): 18ms33us524ns
Total process_execution_payload() duration: 20ms26us137ns
Second hash_tree_root (recomputed): 15ms635us508ns
Total process_execution_payload() duration: 18ms658us562ns
First hash_tree_root (cached): 50ns
Second hash_tree_root (recomputed): 13ms509us937ns
Total process_execution_payload() duration: 15ms593us870ns
Second hash_tree_root (recomputed): 15ms440us776ns
Total process_execution_payload() duration: 17ms856us869ns
Second hash_tree_root (recomputed): 14ms400us667ns
Total process_execution_payload() duration: 16ms946us120ns

```
second run:
```
Second hash_tree_root (recomputed): 14ms128us960ns
Total process_execution_payload() duration: 16ms420us110ns
Second hash_tree_root (recomputed): 13ms699us630ns
Total process_execution_payload() duration: 16ms108us599ns
Second hash_tree_root (recomputed): 14ms66us975ns
Total process_execution_payload() duration: 16ms226us490ns
First hash_tree_root (cached): 70ns
Second hash_tree_root (recomputed): 17ms619us515ns
Total process_execution_payload() duration: 19ms945us729ns
Second hash_tree_root (recomputed): 14ms341us898ns
Total process_execution_payload() duration: 16ms520us207ns
Second hash_tree_root (recomputed): 14ms69us199ns
Total process_execution_payload() duration: 16ms136us942ns
```
with the second `htr` computation accounting for about ~86-90% of the total execution time. The cached read is almost free (60ns vs 15ms), confirming the optimisation works for that code path. Confirms your concern.

tersec: And these are on small test states, not on mainnet-sized states.

Tomi-3-0: Tracking field changes and selectively rebuilding is definitely worth looking into

Tomi-3-0: On the other hand caching intermediate Merkle tree nodes will require SSZ infrastructure changes beyond this PR.

 Since `process_execution_payload` is an [independent state transition check](https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.0/specs/gloas/beacon-chain.md#new-process_execution_payload) separate from beacon block processing, this additional cost happens off the critical block propagation path

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -196,6 +197,44 @@ suite baseDescription & "Deposit Request " & preset():
...
+    return proc(
+        preState: var gloas.BeaconState,
+        signed_envelope: SignedExecutionPayloadEnvelope):
+        Result[void, cstring] =
+      let payloadValid = os_ops.readFile(
+          OpExecutionPayloadDir/"pyspec_tests"/path/"execution.yaml"
+        ).contains("execution_valid: true")
+      var
+        cache: StateCache
+        hashedState = gloas.HashedBeaconState(
```
tersec: This can't go on the stack:
```
2025-10-03T02:03:18.3584322Z /github-runner/github-runner-node-01/workspace/nimbus-eth2/nimbus-eth2/tests/consensus_spec/gloas/test_fixture_operations.nim: In function ‘_ZN27makeApplyExecutionPayloadCb15colonanonymous_E3varIN5gloas11BeaconStateEEN5gloas30SignedExecutionPayloadEnvelopeE’:
2025-10-03T02:03:18.3587313Z /github-runner/github-runner-node-01/workspace/nimbus-eth2/nimbus-eth2/tests/consensus_spec/gloas/test_fixture_operations.nim:202:15: error: stack usage is 5664384 bytes [-Werror=stack-usage=]
2025-10-03T02:03:18.3588542Z   202 |     return proc(
2025-10-03T02:03:18.3588856Z       |               ^
2025-10-03T02:03:22.4021412Z lto1: some warnings being treated as errors
2025-10-03T02:03:22.4195702Z make[2]: *** [/tmp/ccZz6JHU.mk:263: /tmp/ccd0uL3J.ltrans87.ltrans.o] Error 1
2025-10-03T02:03:22.4196636Z make[2]: *** Waiting for unfinished jobs....
2025-10-03T02:03:25.7837262Z lto-wrapper: fatal error: make returned 2 exit status
2025-10-03T02:03:25.7837926Z compilation terminated.
2025-10-03T02:03:25.8356885Z /usr/bin/ld: error: lto-wrapper failed
2025-10-03T02:03:25.8411177Z collect2: error: ld returned 1 exit status
2025-10-03T02:03:25.8427367Z make[1]: *** [nimcache/release/consensus_spec_tests_mainnet/consensus_spec_tests_mainnet.makefile:2388: build] Error 1
2025-10-03T02:03:25.8462789Z make: *** [Makefile:308: consensus_spec_tests_mainnet] Error 2
```
it needs to go on the heap instead (a `ref gloas.HashedBaconState`)

tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/a642d417f1536e28901b2c879a0313edd127d432

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -196,6 +197,44 @@ suite baseDescription & "Deposit Request " & preset():
...
+    return proc(
+        preState: var gloas.BeaconState,
+        signed_envelope: SignedExecutionPayloadEnvelope):
+        Result[void, cstring] =
+      let payloadValid = os_ops.readFile(
+          OpExecutionPayloadDir/"pyspec_tests"/path/"execution.yaml"
+        ).contains("execution_valid: true")
+      var
+        cache: StateCache
+        hashedState = (ref gloas.HashedBeaconState)()
```
tersec: Can construct the `hashedState` directly, as `let` rather than `var` even because its fields then don't need to be changed after the fact directly:
```nim
    let hashedState = (ref gloas.HashedBeaconState)(
      data: preState, root: hash_tree_root(preState))
```

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -196,6 +197,44 @@ suite baseDescription & "Deposit Request " & preset():
...
+      hashedState.data = preState
+      hashedState.root = hash_tree_root(preState)
+
+      func executePayload(_: deneb.ExecutionPayload): bool = payloadValid
+      result = process_execution_payload(
```
tersec: The Status/IFT Nim style guide [recommends against](https://status-im.github.io/nim-style-guide/language.result.html) using `result` return.

The usual approach we tend to use instead is
```nim
      let res  = process_execution_payload(...)
      preState = hashedState.data
      res
```
which dodges the various `result` footguns.

tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/9835619c10db843b89833fe803f36380894009b2

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7552/commits/9835619c10db843b89833fe803f36380894009b2

beacon_chain/spec/state_transition_block.nim
tersec: Yeah, I agree, better to land this as-is, and pursue that later. Good you added the todo comment in the code at least.

PR #7560, opened 2025-10-01
Docs: Use consensus light client naming throughout the guide

kdeme:
Use consensus light client naming throughout the nimbus guide instead of just light client. At the main page this term is already being used but in the rest of the guide it is not.

It is a source of confusion when naming the page just "light client" in terms of what this light client is and what it does.

This will even more so become the case as this guide becomes the general guide of all Nimbus products (and thus other light clients such as the verified proxy and the portal client).

So I think it is best to always name it specifically consensus light client. Other option would be cl light client or beacon light client (but the latter I like less as I think it is less valid as it doesn't really do the beacon tasks).

### DIFF
--- a/docs/the_nimbus_book/src/el-light-client.md
+++ b/docs/the_nimbus_book/src/el-light-client.md
@@ -1,18 +1,18 @@
-# Light client
+# Consensus light client
 
 !!! warning
-    The light client is currently in BETA and details around running it may change.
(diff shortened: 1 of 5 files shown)


kdeme:
We might also want to change the name of the actual binary (can keep the current name in deprecation mode), and the info log print when starting.

etan-status (approved):

PR #7563, opened 2025-10-02
ci: Move linux builds to the container

markoburcul:
We are in the process of moving pipelines across organization to linux docker containers. For the nimbus pipelines we have only two relevant ones:
* main
* nimv2_2

The problem we noticed was when running tests within the nix shell in the container because the GCC doesn't come with LTO support and thus we had to disable it for tests. The **new successfull builds** don't show any degradation in the pipeline performance in terms of duration.

Referenced issue:
* https://github.com/status-im/infra-ci/issues/188

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -175,6 +175,7 @@ pipeline {
         disableDeferredWipeout: true,
         deleteDirs: true
       )
+      dir("${env.WORKSPACE}@tmp") { deleteDir() }
     }
   }
 }
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -1,0 +1,188 @@
+#!/usr/bin/env groovy
+/* beacon_chain
+ * Copyright (c) 2025 Status Research & Development GmbH
+ * Licensed and distributed under either of
+ *   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+ *   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+ * at your option. This file may not be copied, modified, or distributed except according to those terms.
+ */
+library 'status-jenkins-lib@v1.9.2'
+
+pipeline {
+  agent {
+    dockerfile {
+      label 'linuxcontainer'
+      filename 'linux.Dockerfile'
+      dir 'ci'
+    }
+  }
+
+  parameters {
+    choice(
+      name: 'VERBOSITY',
+      description: 'Value for the V make flag to increase log verbosity',
+      choices: [0, 1, 2]
+    )
+    string(
+      name: 'NIM_COMMIT',
+      description: 'Value for the NIM_COMMIT make flag to choose Nim commit',
+      defaultValue: nimCommitForJob(),
+    )
+  }
+
+  options {
+    disableRestartFromStage()
+    timestamps()
+    ansiColor('xterm')
+    /* This also includes wait time in the queue. */
+    timeout(time: 24, unit: 'HOURS')
+    /* Limit builds retained. */
+    buildDiscarder(logRotator(
+      numToKeepStr: '5',
+      daysToKeepStr: '30',
+      artifactNumToKeepStr: '3',
+    ))
+    /* Throttle number of concurrent builds. */
+    throttleJobProperty(
+      throttleEnabled: true,
+      throttleOption: 'category',
+      categories: ['nimbus-eth2'],
+      maxConcurrentPerNode: 1,
+      maxConcurrentTotal: 9
+    )
+    /* Abort old builds for non-main branches. */
+    disableConcurrentBuilds(
+      abortPrevious: !isMainBranch()
+    )
+  }
+
+  environment {
+    NPROC = Runtime.getRuntime().availableProcessors()
+    MAKEFLAGS = "V=${params.VERBOSITY} NIM_COMMIT=${params.NIM_COMMIT} -j${env.NPROC}"
+    XDG_CACHE_HOME = "${env.WORKSPACE_TMP}/.cache"
+  }
+
+  stages {
+    stage('Deps') {
+      steps { 
+        timeout(20) {
+          script {
+            /* To allow the following parallel stages. */
+            sh 'make QUICK_AND_DIRTY_COMPILER=1 update'
+            /* Allow the following parallel stages. */
+            sh 'make deps'
+            /* Download test vectors. */
+            sh './scripts/setup_scenarios.sh'
+          }
+        } 
+      }
+    }
+
+    stage('Build') {
+      steps { 
+        timeout(50) {
+          script {
+            sh 'make LOG_LEVEL=TRACE'
+          }
+        } 
+      }
+    }
+
+    stage('Check Docs') {
+      steps {
+        script {
+          sh './scripts/check_docs_help_msg.sh'
+        }
+      }
+    }
+
+    stage('Tests') {
+      parallel {
+        stage('General') {
+          steps { 
+            timeout(60) {
+              script {
+                sh 'make DISABLE_TEST_FIXTURES_SCRIPT=1 NIMFLAGS="--passC:\"-fno-lto\" --passL:\"-fno-lto\"" test'
+                sh 'git diff --exit-code --ignore-submodules=all'
+              }
+            } 
+          }
+        }
+
+        stage('REST') {
+          steps { 
+            timeout(5) {
+              script {
+                sh 'make restapi-test'
+              }
+            } 
+          }
+          post { always {
+            sh 'tar cjf restapi-test.tar.gz resttest0_data/*.txt'
+          } }
+        }
+      }
+      post { always { timeout(5) {
+        archiveArtifacts(artifacts: '*.tar.gz', allowEmptyArchive: true)
+      } } }
+    }
+
+    stage('Finalizations') {
+      stages {  /* parallel builds of minimal / mainnet not yet supported */
+        stage('minimal') {
+          steps { 
+            timeout(26) {
+              script {
+                sh 'make local-testnet-minimal'
+              }
+            } 
+          }
+          post { always {
+            sh 'tar cjf local-testnet-minimal.tar.gz local-testnet-minimal/logs/*'
+          } }
+        }
+
+        stage('mainnet') {
+          steps { 
+            timeout(62) {
+              script {
+                sh 'make local-testnet-mainnet'
+              }
+            } 
+          }
+          post { always {
+            sh 'tar cjf local-testnet-mainnet.tar.gz local-testnet-mainnet/logs/*'
+          } }
+        }
+      }
+      post {
+        always { timeout(10) {
+          /* DEBUG: Show file sizes to catch too big ones. */
+          sh 'ls -hl *.tar.gz'
+          archiveArtifacts(
+            artifacts: '*.tar.gz',
+            excludes: '**/geth-*.tar.gz',  /* `scripts/geth_binaries.sh` */
+            allowEmptyArchive: true
+          )
+        } }
+      }
+    }
+  }
+
+  post {
+    always {
+      cleanWs(
+        disableDeferredWipeout: true,
+        deleteDirs: true
+      )
+    }
+  }
+}
+
+def isMainBranch() {
+  return ['stable', 'testing', 'unstable'].contains(env.BRANCH_NAME)
+}
+
+def nimCommitForJob() {
+  return JOB_NAME.contains('nimv2_2') ? 'upstream/version-2-2' : ''
+}
--- a/ci/linux.Dockerfile
+++ b/ci/linux.Dockerfile
@@ -1,0 +1,37 @@
+FROM harbor.status.im/infra/ci-build-containers:linux-base-1.0.0
+USER root
+
+RUN apt-get update && apt-get install -yq --no-install-recommends \
+    figlet \
+    git \
+    git-lfs \
+    make \
+    openssl \
+    lsof \
+    psmisc \
+    procps \
+    curl \
+    jq \
+    openjdk-17-jre-headless \
+    python3 \
+    python3-pip \
+    python3-venv \
+    gcc-11 \
+    g++-11 \
+    lsb-release \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-11 100 \
+    && update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-11 100
+
+RUN ln -sf /usr/bin/gcc /usr/bin/cc \
+    && ln -sf /usr/bin/g++ /usr/bin/c++
+
+RUN pip3 install --no-cache-dir --break-system-packages \
+    mkdocs \
+    mkdocs-material \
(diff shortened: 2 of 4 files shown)


jakubgs:
I looked into these errors from `make test DISABLE_TEST_FIXTURES_SCRIPT=1`:
```
09:36:31  /nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: /tmp/nix-shell.DI1pTZ/ccKHH2R1.ltrans0.ltrans.o: in function `visualizeHeader':
09:36:31  <artificial>:(.text+0x3ec): undefined reference to `ETHLightClientHeaderCopyBeaconRoot'
09:36:31  /nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: <artificial>:(.text+0x426): undefined reference to `ETHRootDestroy'
09:36:31  /nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: <artificial>:(.text+0x42e): undefined reference to `ETHLightClientHeaderGetBeacon'
09:36:31  /nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: <artificial>:(.text+0x439): undefined reference to `ETHBeaconBlockHeaderGetSlot'
...
```
I tried using the target with `-j1 and V=2` as well as `NIM_PARAMS='--passL:"-Wl,--verbose"'` and found this GCC command:
https://github.com/status-im/nimbus-eth2/blob/5c454ec70d61bdaafa79cdc0c6104b78fda9db12/Makefile#L717-L725

And indeed I could reproduce the issue using it:
```
[jakubgs@caspair:~/work/nimbus-eth2]$ gcc -D__DIR__="\"beacon_chain/libnimbus_lc\"" --std=c17 -flto -pedantic -pedantic-errors -Wall -Wextra -Werror -Wno-maybe-uninitialized -Wno-unsafe-buffer-usage -Wno-unknown-warning-option -o build/test_libnimbus_lc beacon_chain/libnimbus_lc/test_libnimbus_lc.c build/libnimbus_lc.a
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: /tmp/nix-shell.fK6O0E/ccolI9dG.ltrans0.ltrans.o: in function `visualizeHeader':
<artificial>:(.text+0x3cc): undefined reference to `ETHLightClientHeaderCopyBeaconRoot'
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: <artificial>:(.text+0x406): undefined reference to `ETHRootDestroy'
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: <artificial>:(.text+0x40e): undefined reference to `ETHLightClientHeaderGetBeacon'
...
```

jakubgs:
We attempted to compare the `build/libnimbus_lc.a` built via Nix tooling and using the Ubuntu tooling.

What we found was that `nm` could not print the symbols on `build/libnimbus_lc.a` on NixOS:
```
[jakubgs@caspair:~/work/nimbus-eth2]$ nm -g --defined-only ~/libnimbus_lc.a 2>&1 | grep NimMain

[jakubgs@caspair:~/work/nimbus-eth2]$ 
```
While it could on our CI host:
```
jakubgs@linux-01.he-eu-hel1.ci.devel:~ % nm -g --defined-only libnimbus_lc.a 2>&1 | grep NimMain
00000000 T NimMain
00000000 T NimMainInner
00000000 T NimMainModule
```
And the error was:
```
@mlibnimbus_lc.nim.c.o:
nm: @mlibnimbus_lc.nim.c.o: plugin needed to handle lto object
0000000000000000 W g_mlibnimbus_lc.nim.c.2c72c415
0000000000000001 C __gnu_lto_slim
```
Which indicates that the `nm` on NixOS lacks LTO support.

jakubgs:
The we found out that LTO actually works correctly when `nm` is used from `gcc-unwrapped` instead of `gcc`:
```
[jakubgs@caspair:~/work/nimbus-eth2]$ /nix/store/lc7vdrd7l2apdjy9gzbljn6fgzj5nyz3-gcc-wrapper-11.4.0/bin/nm -Ag --defined-only build/libnimbus_lc.a 2>&1 | grep ETHRootDestroy

[jakubgs@caspair:~/work/nimbus-eth2]$ /nix/store/g3xq3b8b8gwiw7j68v5wh3lvgc4fr3yz-gcc-11.4.0/bin/gcc-nm -Ag --defined-only build/libnimbus_lc.a 2>&1 | grep ETHRootDestroy
build/libnimbus_lc.a:@mlibnimbus_lc.nim.c.o:00000000 T ETHRootDestroy
```
Which means that the Nix wrapping of GCC is causing issues with LTO plugin.

When unwrapped GCC is used it fails on missing Glibc instead:
```
[jakubgs@caspair:~/work/nimbus-eth2]$ /nix/store/g3xq3b8b8gwiw7j68v5wh3lvgc4fr3yz-gcc-11.4.0/bin/gcc build/libnimbus_lc.a beacon_chain/libnimbus_lc/test_libnimbus_lc.c -o build/test_libnimbus_lc
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: cannot find crt1.o: No such file or directory
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: cannot find crti.o: No such file or directory
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: cannot find -lgcc_s: No such file or directory
collect2: error: ld returned 1 exit status
```

jakubgs:
If we set `export NIX_DEBUG=1` we can make the wrapped GCC print debug logs which shows the flags:
```
original flags to /nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld:
  -plugin
  /nix/store/g3xq3b8b8gwiw7j68v5wh3lvgc4fr3yz-gcc-11.4.0/libexec/gcc/x86_64-unknown-linux-gnu/11.4.0/liblto_plugin.so
  -plugin-opt=/nix/store/g3xq3b8b8gwiw7j68v5wh3lvgc4fr3yz-gcc-11.4.0/libexec/gcc/x86_64-unknown-linux-gnu/11.4.0/lto-wrapper
```
```
extra flags after to /nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld:
  -rpath
  /nix/store/ddwyrxif62r8n6xclvskjyy6szdhvj60-glibc-2.39-5/lib
  -rpath
  /nix/store/p5spyl0dhqwrs98kjbhca964rcyl4pj2-gcc-11.4.0-lib/lib
/nix/store/hqvni28zpibl6jsqqimcvng6h6qm58xy-binutils-2.41/bin/ld: /tmp/nix-shell.a7vrBu/ccR3W27J.ltrans0.ltrans.o: in function `visualizeHeader':
<artificial>:(.text+0x374): undefined reference to `ETHLightClientHeaderCopyBeaconRoot'
```
Which show that the original `ld` flags included LTO plugin, while the modified flags do not.

jakubgs:
I have managed to make it work by passing `NIM_PARAMS='--passC:"-fno-lto" --passL:"-fno-lto"'`:
```
[jakubgs@caspair:~/work/nimbus-eth2]$ make test DISABLE_TEST_FIXTURES_SCRIPT=1 NIM_PARAMS='--passC:"-fno-lto" --passL:"-fno-lto"'
...
Build completed successfully: build/process_state
Build completed successfully: build/libnimbus_lc.a
Build completed successfully: build/test_libnimbus_lc
Build completed successfully: build/nimbus_signing_node
Build completed successfully: build/block_sim
Build completed successfully: build/consensus_spec_tests_minimal
Build completed successfully: build/consensus_spec_tests_mainnet
Build completed successfully: build/all_tests

Running consensus_spec_tests_minimal --xml:build/consensus_spec_tests_minimal.xml --console
```
Which confirms the issue is with wrapped GCC and LTO support when loading symbols from `build/libnimbus_lc.a`.

markoburcul:
Posts with same issue:
https://discourse.nixos.org/t/anyone-was-successful-with-setting-up-lto-pgo-with-gcc/63905
https://github.com/NixOS/nixpkgs/pull/399656
https://discourse.nixos.org/t/how-to-correctly-use-ld-nm-ar-ranlib-with-lto-in-nix-develop-shell/33220

What I've found is this line in the `ld-wrapper.sh` from nixpkgs which is potentially causing the issue:
https://github.com/NixOS/nixpkgs/blob/20c4598c84a671783f741e02bf05cbfaf4907cff/pkgs/build-support/bintools-wrapper/ld-wrapper.sh#L144C1-L146C19

The LTO does the run time optimizations, but we are using it only in the tests and not when building the node itself. I will do a test where I'll run the tests with and without the LTO enabled to see if it impacts significantly to the performance. If not, I think it's safe to disable it.

nix/shell.nix
```diff
@@ -8,14 +8,16 @@ let
...
+in pkgs.mkShellNoCC {
 
   buildInputs = with pkgs; [
     figlet
     git
     git-lfs
     gnumake
     getopt
+    gcc11
+    binutils
```
jakubgs: I think the more correct way is using:
```nix
in pkgs.mkShell.override { stdenv = pkgs.gcc11Stdenv; } {
```
Or even better, in an overlay:
```nix
  overlays = [
    (final: prev: { stdenv = prev.gcc11Stdenv; })
  ];
```

markoburcul:
In the normal build without nix, we are having an increase of 5 min because of disabling LTO in the tests step:
* [with LTO](https://ci.status.im/job/nimbus-eth2/job/platforms/job/linux/job/x86_64/job/main/view/change-requests/job/PR-7563/25/)
* [without LTO](https://ci.status.im/job/nimbus-eth2/job/platforms/job/linux/job/x86_64/job/main/view/change-requests/job/PR-7563/27/)

The Nix [build failed](https://ci.status.im/job/nimbus-eth2/job/platforms/job/linux/job/x86_64/job/main/view/change-requests/job/PR-7584/8/) even though in the shell hook I've set the NIM_PARAMS to disable LTO, but maybe the outer quotes got stripped. I've rerun [the build](https://ci.status.im/job/nimbus-eth2/job/platforms/job/linux/job/x86_64/job/main/view/change-requests/job/PR-7584/10/) with setting the variable in the test step like this:
```
nix.develop('''make DISABLE_TEST_FIXTURES_SCRIPT=1 NIM_PARAMS='--passC:"-fno-lto" --passL:"-fno-lto"' test''', pure: false)
```
it passed the step, now I'm waiting for the total time.

nix/shell.nix
```diff
@@ -38,6 +39,9 @@ in pkgs.mkShell {
     export USE_SYSTEM_GETOPT=1
     export MAKEFLAGS="-j$NIX_BUILD_CORES"
 
+    # The LTO is not included in the GCC provided by nixpkgs, causing tests to fail
```
jakubgs:
```suggestion
    # LTO does not work via GCC wrapper provided by nixpkgs stdenv, causing tests to fail.
```

ci/Jenkinsfile.linux
```diff
@@ -0,0 +1,190 @@
...
+ *   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+ * at your option. This file may not be copied, modified, or distributed except according to those terms.
+ */
+library 'status-jenkins-lib@v1.9.2'
+
+pipeline {
+  agent {
+    dockerfile {
+      label 'linuxcontainer'
+      filename 'linux.Dockerfile'
```
jakubgs, line 15: Why are you not using `harbor.status.im/infra/ci-build-containers:linux-base-1.0.0`?

nix/shell.nix
```diff
@@ -38,6 +39,9 @@ in pkgs.mkShell {
...
+    export NIMFLAGS="--passC:\"-fno-lto\" --passL:\"-fno-lto\""
```
jakubgs: I wonder if it fails only for tests wouldn't it make sense to use this only for tests too?

Is there a noticeable difference in total duration?

jakubgs (changes requested):
Why is this a separate Jenkinsfile from `ci/Jenkinsfile`?

nix/shell.nix
markoburcul: > I wonder if it fails only for tests wouldn't it make sense to use this only for tests too?

I think it is better to just use it for tests.



> Is there a noticeable difference in total duration?

There is no difference in the duration as I observed in the last successful build.

markoburcul:
> Why is this a separate Jenkinsfile from `ci/Jenkinsfile`?

Because we are using a docker container and since the same jenkinsfile is used for macos builds this would fail.

siddarthkay (approved):

nix/shell.nix
```diff
@@ -8,7 +8,7 @@ let
     pymdown-extensions
   ];
   mkdocs-python = pkgs.python3.withPackages mkdocs-packages;
-in pkgs.mkShell {
+in pkgs.mkShell.override { stdenv = pkgs.gcc11Stdenv; } {
```
jakubgs: We should also use this `gcc11Stdenv` for the derivation to be consistent.

jakubgs (approved):

tersec:
> I have managed to make it work by passing `NIM_PARAMS='--passC:"-fno-lto" --passL:"-fno-lto"'`:
> 
> ```
> [jakubgs@caspair:~/work/nimbus-eth2]$ make test DISABLE_TEST_FIXTURES_SCRIPT=1 NIM_PARAMS='--passC:"-fno-lto" --passL:"-fno-lto"'
> ...
> Build completed successfully: build/process_state
> Build completed successfully: build/libnimbus_lc.a
> Build completed successfully: build/test_libnimbus_lc
> Build completed successfully: build/nimbus_signing_node
> Build completed successfully: build/block_sim
> Build completed successfully: build/consensus_spec_tests_minimal
> Build completed successfully: build/consensus_spec_tests_mainnet
> Build completed successfully: build/all_tests
> 
> Running consensus_spec_tests_minimal --xml:build/consensus_spec_tests_minimal.xml --console
> ```
> 
> Which confirms the issue is with wrapped GCC and LTO support when loading symbols from `build/libnimbus_lc.a`.

In general, Nimbus ships an LTO build, so it'd be better to test with it too.

nix/shell.nix
jakubgs: Not really related to this PR, but if you want to change this change it in separate PR and also for derivation.

ci/linux.Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    && ln -sf /usr/bin/g++ /usr/bin/c++
+
+RUN pip3 install --no-cache-dir --break-system-packages \
+    mkdocs \
+    mkdocs-material \
+    mkdocs-material-extensions \
+    pymdown-extensions
+
+ENV CFLAGS=""
+ENV LDFLAGS=""
```
jakubgs: Not sure what the point of this is.

jakubgs (approved):
It's really really really dumb how we can't reuse the same Jenkinsfile because of the `dockerfile` block being in a static `agent` block parsed early on.

But what we could do is define stages in a `common.groovy` file to not repeat the same thing twice. But that can be tried separately.

PR #7566, opened 2025-10-02
Pass SECONDS_PER_SLOT to BeaconClock module

etan-status:
Retain time config when initializing `BeaconClock` as the next step to make SECONDS_PER_SLOT runtime configurable.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -778,6 +778,7 @@ AllTests-mainnet
 + Voluntary exit signatures                                                                  OK
 + execution payload bid signatures                                                           OK
 + execution payload envelope signatures                                                      OK
++ payload attestation message signatures                                                     OK
 ```
 ## Network metadata
 ```diff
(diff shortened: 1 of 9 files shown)


beacon_chain/nimbus_light_client.nim
```diff
@@ -73,7 +73,7 @@ proc main() {.noinline, raises: [CatchableError].} =
         raiseAssert "Invalid baked-in state: " & err.msg
 
     genesisTime = getStateField(genesisState[], genesis_time)
-    beaconClock = BeaconClock.init(genesisTime).valueOr:
+    beaconClock = BeaconClock.init(cfg.time, genesisTime).valueOr:
```
etan-status, line 76: Interesting how this one had enough foresight to put the cfg as a parameter into the API, despite not having been used before.

AllTests-mainnet.md
```diff
@@ -778,6 +778,7 @@ AllTests-mainnet
 + Voluntary exit signatures                                                                  OK
 + execution payload bid signatures                                                           OK
 + execution payload envelope signatures                                                      OK
++ payload attestation message signatures                                                     OK
```
etan-status, line 781: It's weird that somehow this adds a new test

tersec: It doesn't, that was an oversight from another PR which wasn't able to trigger Jenkins.

PR #7569, opened 2025-10-05
add gloas validation for attestation and aggregates

Tomi-3-0:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2276,172 +2278,172 @@ proc installMessageValidators(node: BeaconNode) =
...
-                toValidationResult(
-                  node.processor[].processAttesterSlashing(
-                    MsgSource.gossip, attesterSlashing)))
-
-          # proposer_slashing
-          # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/phase0/p2p-interface.md#proposer_slashing
+        when consensusFork >= ConsensusFork.Gloas:
+          debugGloasComment "add processExecutionPayload"
+          debugGloasComment "add processExecutionPayloadBid"
+          debugGloasComment "add processPayloadAttestation"
```
ahshum: I think we could leave the placeholders out here.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1108,6 +1109,9 @@ proc validateAttestation*(
       return pool.checkedReject(v.error())
     v.get()
 
+  let currentFork =
```
tersec: Most correctly, this isn't quite time-triggered, though it should correspond fairly closely, but which fork's gossip channel the attestation (or aggregation, all this will apply similarly there) it's received on. This has the added benefit that while the current eth2 processor, etc call chain doesn't convey this information, it is known statically at compile time which fork a particular gossip validation concerns by virtue of which callback has been set up for it by `installMessageValidators`.

Tomi-3-0: I see.
I can potentially implement this by threading a `consensusFork` parameter through the call chain.

By adding the `consensusFork: static ConsensusFork` parameter to `validateAttestation` and `validateAggregate` then use the parameter directly in validation checks along with updating the processor methods to pass it through

tersec: Exactly, yeah.

Tomi-3-0: Added a fork parameter to attestation validators so they know whether to apply `Electra` or `Gloas` rules, used the `withConsensusFork` template to dispatch to the correct validator based on fork, changed gossip validators to pass fork information from the gossip topic instead of checking the current time, and made API routes calculate the current fork from wall time and pass it to validators.

tersec: https://github.com/status-im/nimbus-eth2/pull/7569/commits/50360377d713c7c18a3ed7fb9988bde87ab4a8d6

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -441,9 +442,15 @@ proc processAttestation*(
...
-  let v = await self.attestationPool.validateAttestation(
-    self.batchCrypto, attestation, wallTime, subnet_id, checkSignature)
+  let v = when attestation is phase0.Attestation:
+    await self.attestationPool.validateAttestation(
+      self.batchCrypto, attestation, wallTime, subnet_id, checkSignature)
+  else:
+    withConsensusFork(fork):
+      await self.attestationPool.validateAttestation(
+        self.batchCrypto, attestation, wallTime, subnet_id, checkSignature,
+        consensusFork)
```
ahshum, line 452: You should be able to pass `fork` without `withConsensusFork`

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1117,15 +1118,14 @@ proc validateAttestation*(
...
-      pool.dag.cfg.consensusForkAtEpoch(wallTime.slotOrZero.epoch), slot,
-      wallTime)
+    let v = check_propagation_slot_range(consensusFork, slot, wallTime)
     if v.isErr():  # [IGNORE]
       return err(v.error())
 
   # [REJECT] attestation.data.index == 0
-  if not (attestation.data.index == 0):
-    return pool.checkedReject("SingleAttestation: attestation.data.index != 0")
+  if consensusFork < ConsensusFork.Gloas:
```
ahshum, line 1126: Would be best to use `when` for runtime performance.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2276,172 +2278,174 @@ proc installMessageValidators(node: BeaconNode) =
...
-                src: PeerId
-              ): ValidationResult =
-                toValidationResult(
-                  node.processor[].processAttesterSlashing(
-                    MsgSource.gossip, attesterSlashing)))
-
-          # proposer_slashing
-          # https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.8/specs/phase0/p2p-interface.md#proposer_slashing
+        when consensusFork >= ConsensusFork.Gloas:
+          debugGloasComment " "
```
ahshum: It seems that it is not needed/helpful as there is not `else` statement or anything depends on it.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2276,172 +2278,174 @@ proc installMessageValidators(node: BeaconNode) =
...
+              src: PeerId
+            ): Future[ValidationResult] {.async: (raises: [CancelledError]).} =
+              return toValidationResult(
+                await node.processor.processBlsToExecutionChange(
+                  MsgSource.gossip, msg)))
+        # data_column_sidecar_{subnet_id}
+        # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.3/specs/fulu/p2p-interface.md#data_column_sidecar_subnet_id
+        # https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.0/specs/gloas/p2p-interface.md#data_column_sidecar_subnet_id
+        when consensusFork >= ConsensusFork.Fulu and
+             consensusFork != ConsensusFork.Gloas:
```
ahshum: Could be improved to `when consensusFork >= Gloas` then `elif consensusFork >= Fulu`.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2276,172 +2278,174 @@ proc installMessageValidators(node: BeaconNode) =
...
+                getDataColumnSidecarTopic(digest, subnet_id), proc (
+                  dataColumnSidecar: fulu.DataColumnSidecar,
+                  src: PeerId
+                ): ValidationResult =
+                  toValidationResult(
+                    node.processor[].processDataColumnSidecar(
+                      MsgSource.gossip, dataColumnSidecar, subnet_id)))
+        elif consensusFork >= ConsensusFork.Gloas:
+          debugGloasComment " "
+        when consensusFork in [ConsensusFork.Deneb, ConsensusFork.Electra]:
```
ahshum, line 2438: As a general comment, there were an empty line before each of the topic groups. Would be great to keeping them for readability.

beacon_chain/nimbus_beacon_node.nim
Tomi-3-0: This `when` is needed to ensure the code only compiles for `Gloas` and later forks. The gossip topics that will go inside this block were introduced in `Gloas`

beacon_chain/gossip_processing/gossip_validation.nim
Tomi-3-0: I'd like to use `when` for compile-time optimization, but since consensusFork is a runtime parameter (not static), I need to use `if`. Making it static would require adding runtime dispatchers throughout the call chain

beacon_chain/nimbus_beacon_node.nim
Tomi-3-0: okay

beacon_chain/gossip_processing/gossip_validation.nim
ahshum: Hm right didn't realize that rest api is calling this function. Let's leave it for now.

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7569/commits/f739dd7e388a0112490c136a4116ce7fe3399578

beacon_chain/nimbus_beacon_node.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7569/commits/f739dd7e388a0112490c136a4116ce7fe3399578

beacon_chain/nimbus_beacon_node.nim
tersec: > This `when` is needed to ensure the code only compiles for `Gloas` and later forks. The gossip topics that will go inside this block were introduced in `Gloas`

Yeah, I agree with this. the point is to be able to have as as-complete-as-reasonable map of required next steps by looking for `debugGloasComment`.

beacon_chain/gossip_processing/eth2_processor.nim
Tomi-3-0: > https://github.com/status-im/nimbus-eth2/pull/7569/commits/f739dd7e388a0112490c136a4116ce7fe3399578

ahshum: FYI the highlighted line is not outdated

tersec: Oh, indeed, something to follow up on.

Tomi-3-0: I will follow up on that

Tomi-3-0: https://github.com/status-im/nimbus-eth2/pull/7593/commits/b255d8d781b55e9a25fde73dd90d89e430f02460

PR #7617, opened 2025-10-15
add blob limit, in data column sidecar gossip

agnxsh:
https://github.com/ethereum/consensus-specs/pull/4650


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -598,6 +598,12 @@ proc validateDataColumnSidecar*(
     if v.isErr:
       return dag.checkedReject(v.error)
 
+  # Check if the sidecar respects the blob limit.
```
tersec, line 601: Good to link to:
- https://github.com/ethereum/consensus-specs/pull/4650

within the code itself in lieu of the usual spec link, which maybe can be replaced with a spec link later.

PR #7626, opened 2025-10-18
Pass TimeConfig to start_beacon_time

etan-status:
Every conversion from slot to time needs the TimeConfig.

### DIFF
--- a/tests/test_beacon_time.nim
+++ b/tests/test_beacon_time.nim
@@ -60,14 +60,9 @@ suite "Beacon time":
       check:
         counts == 2
 
-  test "basics (SECONDS_PER_SLOT=12)":
-    doBasicsTest(TimeConfig(SECONDS_PER_SLOT: 12))
-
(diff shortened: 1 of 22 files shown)


tests/test_beacon_time.nim
```diff
@@ -56,6 +60,15 @@ suite "Beacon time":
       check:
         counts == 2
 
+  test "basics (SECONDS_PER_SLOT=12)":
+    doBasicsTest(TimeConfig(SECONDS_PER_SLOT: 12))
+
+  test "basics (SECONDS_PER_SLOT=6)":
+    doBasicsTest(TimeConfig(SECONDS_PER_SLOT: 5))
```
tersec: 5?

etan-status: Gnosis

etan-status: ah, I see, yeah, refactored it

PR #7628, opened 2025-10-19
make reconstruction timeout dynamic + prevent reconstruction in nonSupernodes

agnxsh:


beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -9,7 +9,7 @@
 
 # Uncategorized helper functions from the spec
 import
-  chronos, chronicles, results, taskpools,
+  chronos, chronicles, results, taskpools, times,
```
tersec, line 12: `times` is a stdlib library. should be on a separate line and prefixed by `std/`:
```nim
import
  std/times,
  chronos, chronicles, results, taskpools,
```

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -170,13 +171,15 @@ proc recover_cells_and_proofs_parallel*(
     res = newSeq[CellsAndProofs](blobCount)
 
   let startTime = Moment.now()
-  const reconstructionTimeout = 2.seconds
+  let reconstructionTimeout =
+    (initDuration(nanoseconds = slotDuration * 100_000_000)).inNanoseconds()
```
tersec, line 175: `100_000_000` is off by a factor of 10.

More generally, `initDuration` [supports `milliseconds` too](https://nim-lang.org/docs/times.html#initDuration%2Cint64%2Cint64%2Cint64%2Cint64%2Cint64%2Cint64%2Cint64%2Cint64), so there's no particular reason to use nanoseconds here.

`inNanoSeconds` just [converts all this back into an `int64` anyway](https://nim-lang.org/docs/times.html#inNanoseconds%2CDuration), so this whole line is a roundbout method of multiplying by what's supposed to be `1_000_000_000` which invokes `initDuration`/`inNanoseconds` for no obvious reason.

That is, one should be able to do something like
```nim
  let reconstructionTimeout = slotDuration * 1000
```
or similar, rather than round-tripping through these `std/times` types. Also, as `SLOT_DURATION_MS` in milliseconds will come in, this will have to support milliseconds, so that's probably the correct granularity. Seconds will have insufficient granularity (already in some cases) and `microseconds` pointlessly large to handle, and the times library does like to operate in integers.

While the two
```nim
    if (now - startTime).nanoseconds > reconstructionTimeout
```
lines use [`nanoseconds`](https://nim-lang.org/docs/times.html#nanoseconds%2Cint), they could use [`milliseconds`](https://nim-lang.org/docs/times.html#milliseconds%2Cint) just as easily without the evidently easy-to-typo large numbers.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1713,6 +1713,16 @@ proc reconstructDataColumns(node: BeaconNode, slot: Slot) =
     warn "Failed to get the current slot head"
     return
 
+  let
+    currentCgc = node.dataColumnQuarantine.custodyColumns.lenu64
+    nonSupernode =
+      currentCgc > node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2 and
```
tersec, line 1719: This boundary condition seems to create an odd hole for `node.dataColumnQuarantine.custodyColumns.lenu64 == node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2`. It won't be rejected by the earlier
```nim
  if node.dataColumnQuarantine.custodyColumns.lenu64 <
      node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2:
    return
```
and it won't trigger `nonSupernode` status here (`64 > 64 and 64 < 128 == false`).

Maybe `>=` was meant? But see comment below for a slightly broader point, related to this.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1713,6 +1713,16 @@ proc reconstructDataColumns(node: BeaconNode, slot: Slot) =
...
+      currentCgc < node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS
+
+  if not(node.config.peerdasSupernode) or
+     nonSupernode:
```
tersec, line 1723:
```
>>> len('  if not(node.config.peerdasSupernode) or nonSupernode:')
55
```
so not really sure why it's two lines.

But, beyond the most superficial cosmetics, earlier in the function there's already a
https://github.com/status-im/nimbus-eth2/blob/d71ea30087f0781bf1eaca7cf689897491078527/beacon_chain/nimbus_beacon_node.nim#L1705-L1707
which is effectively part of this logic too. I'm not sure why these two chunks of custody column checking should exist separately, and it's simpler it they're combined: to achieve a similar effect, check once for `not(node.config.peerdasSupernode) or node.dataColumnQuarantine.custodyColumns.lenu64 < node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS`. Nothing about this function's current behavior changes between custody count of 63 and 65 (why [64](https://github.com/status-im/nimbus-eth2/pull/7628/files#r2444010287)? is that intentional?), it's just ruling everything under 128 out.

I'm not really sure in what circumstances `node.config.peerdasSupernode` could hold though and have also `node.dataColumnQuarantine.custodyColumns.lenu64 < node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS`.

tersec:
Overtaken by events.

PR #7630, opened 2025-10-20
Pass TimeConfig to slotOrZero

etan-status:
Conversion from time to slot also needs access to the TimeConfig. An alternative design based on storing TimeConfig inside BeaconTime was explored. While that reduces the amount of refactoring, such a design wouldn't be as clean, as it suggests to beacon_time module that each TimeConfig may come from a different BeaconClock. In such a setup (e.g., a hypothetical multi-network VC with various clocks), one would have to retain the entire BeaconClock not just the TimeConfig to catch time operations on times originated from different clocks. Further, the TimeConfig would have to be passed along internally on every minor operation such as adding an offset. Therefore, opted to follow existing approach to just pass TimeConfig explicitly where it will be needed.

### DIFF
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -417,7 +417,8 @@ proc addAttestation*(
   doAssert attestation.signature == signature.toValidatorSig(),
     "Deserialized signature must match the one in the attestation"
 
-  updateCurrent(pool, wallTime.slotOrZero)
+  let timeConfig = pool.dag.cfg.time
+  updateCurrent(pool, wallTime.slotOrZero(timeConfig))
 
   let candidateIdx = pool.candidateIdx(attestation.data.slot)
   if candidateIdx.isNone:
--- a/beacon_chain/beacon_clock.nim
+++ b/beacon_chain/beacon_clock.nim
@@ -75,6 +75,9 @@ proc now*(c: BeaconClock): BeaconTime =
   ## Current time, in slots - this may end up being less than GENESIS_SLOT(!)
   toBeaconTime(c, getTime())
 
+proc currentSlot*(c: BeaconClock): Slot =
+  c.now.slotOrZero(c.timeConfig)
+
 func getBeaconTimeFn*(c: BeaconClock): GetBeaconTimeFn =
   return proc(): BeaconTime = c.now()
 
--- a/beacon_chain/beacon_node.nim
+++ b/beacon_chain/beacon_node.nim
@@ -135,7 +135,7 @@ template rng*(node: BeaconNode): ref HmacDrbgContext =
   node.network.rng
 
 proc currentSlot*(node: BeaconNode): Slot =
-  node.beaconClock.now.slotOrZero
+  node.beaconClock.currentSlot
 
 func hasRestAllowedOrigin*(node: BeaconNode): bool =
   node.config.restAllowedOrigin.isSome
--- a/beacon_chain/consensus_object_pools/consensus_manager.nim
+++ b/beacon_chain/consensus_object_pools/consensus_manager.nim
@@ -511,12 +511,14 @@ proc updateExecutionHead*(
   defer:
     self.forkchoiceInflight = false
 
+  let timeConfig = self.dag.cfg.time
   var
     attempts = 0
     wallTime = getBeaconTimeFn()
     head = self.attestationPool[].getBeaconHead(self.dag.head)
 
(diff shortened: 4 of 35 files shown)


beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -417,7 +417,8 @@ proc addAttestation*(
   doAssert attestation.signature == signature.toValidatorSig(),
     "Deserialized signature must match the one in the attestation"
 
-  updateCurrent(pool, wallTime.slotOrZero)
+  let timeConfig = pool.dag.cfg.time
```
arnetheduck, line 420: since the RuntimeConfig object is called `cfg` throughout, maybe we want to call this one `tcfg`? ie as long as it's greppable ...

etan-status: I had it as timeCfg initially but was asked to change it to timeConfig.

It is greppable via `cfg.time` and `timeConfig`

etan-status: Or you mean inside cfg? `cfg.tcfg`?

arnetheduck: both inside and outside maybe..

etan-status: There are a couple exceptions where RuntimeConfig is not named `cfg`, but `...Config`, e.g., `defaultRuntimeConfig`

arnetheduck: not when used as a local / parameter

cheatfate: `tcfg` is an ugly name which could be treated as temporary config by some newcomer. If you so opposite to `config` probably we should introduce another name, but definitely not a `tcfg` or `cfgt` or any other 4-letter word.

etan-status: `timeParams` maybe?

arnetheduck: the fact that it's 4 chars is its feature - given its ubiquity, it simply becomes part of the local parlance, similar to how for loop variables are often called `i` for no particular reason. It's a shortcut for a longer name, and we use those in cases like this, when there's a crosscutting need for something that appears all over the place without in and of itself providing any particular value for understanding the code.

arnetheduck: anyway, don't let this stop the PR these discussions can get uselessly long -  `timeConfig` is not a terrible name, just makes the code unnecessarily lengthy - suffice to say that we have a few of these "standard" names in the codebase that newcomers must learn to make sense of it (cfg being one) and time config feels like it fits into that category.

cheatfate: **Question:**

```tcfg is a good name for variable?```

**Answer:**

No, "tcfg" is generally not a good variable name because it's not descriptive. Good variable names should clearly explain what the variable holds, which "tcfg" does not. While technical and concise names can be appropriate in some very specific, short-lived contexts, descriptive names are preferred for maintainable and readable code. 
Why descriptive names are better

    Clarity:
    Names like student_name or purchase_amount immediately tell you the purpose of the variable.
    Readability:
    A variable name like tcfg is cryptic and requires the reader to remember or look up what it means, making the code harder to understand.
    Maintenance:
    When you or someone else needs to revisit the code later, a descriptive name makes it much easier to understand the logic without having to decipher abbreviations. 

What to use instead

    Write out what it means: If "tcfg" stands for "text configuration," a better name would be text_configuration or textConfig (using camel case). 

Use nouns: Nouns or noun phrases are often good choices for variable names. 
Be concise but not cryptic: Find a balance between being descriptive and keeping the name from being excessively long.

etan-status: I'll make a followup PR to align them as `timeParams`, so that at least inside and outside is the same. `cfg.timeParams` works better than `cfg.timeCfg` or `cfg.tcfg` imo.

If length is a concern, I can make helper funcs to hide the interaction. e.g., define a `func timeParams` on `AttestationPool`, `ChainDAGRef`, etc, so that the various functions don't have to make the lengthy `let timeParams = ...`

etan-status: https://github.com/status-im/nimbus-eth2/pull/7635

PR #7636, opened 2025-10-21
Properly process proglist SSZ generic tests

etan-status:
Load all `ProgressiveList` tests, report folders with unexpected name.


etan-status:
Needs 1.6.0-b.1, for corrected test vectors in those folders.

PR #7644, opened 2025-10-21
Prefer runtime TimeParams over compiletime

etan-status:
Change a couple uses of compiletime constant to use runtime config, where it is local and does not span across multiple modules.

### DIFF
--- a/beacon_chain/fork_choice/fork_choice.nim
+++ b/beacon_chain/fork_choice/fork_choice.nim
@@ -196,7 +196,7 @@ func contains*(self: ForkChoiceBackend, block_root: Eth2Digest): bool =
 proc update_time*(
     self: var ForkChoice, dag: ChainDAGRef, time: BeaconTime): FcResult[void] =
   # `time` is the wall time, meaning it changes on every call typically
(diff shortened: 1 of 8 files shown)


beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -196,7 +196,7 @@ func contains*(self: ForkChoiceBackend, block_root: Eth2Digest): bool =
 proc update_time*(
     self: var ForkChoice, dag: ChainDAGRef, time: BeaconTime): FcResult[void] =
   # `time` is the wall time, meaning it changes on every call typically
-  const step_size = seconds(SECONDS_PER_SLOT.int)
+  let step_size = seconds(dag.timeParams.SECONDS_PER_SLOT.int)
```
tersec: Probably safer for this to be an `int64`, which is compatible with `seconds` and avoids 32/64-bit platform-size-dependencies. It didn't matter when it was `const`; it'd be a compile-time error.

etan-status: Done, changed to int64, and also did a quick scan for similar mistake, the others already seem to be int64.

PR #7646, opened 2025-10-22
Add execution payload bid pool

Tomi-3-0:
Implement execution payload bid pool for Gloas
 - Stores the highest-value bid per (slot, parent_block_hash) pair
 - Tracks which builders have submitted bids for each slot
 - All data stored in `slotBids`, other structures just index it


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2303,6 +2304,18 @@ proc installMessageValidators(node: BeaconNode) =
                 node.processor[].processSignedBeaconBlock(
                   MsgSource.gossip, signedBlock)))
 
+        # execution_payload_bid
+        # https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.0/specs/gloas/p2p-interface.md#execution_payload_bid
```
tersec: beta.1 is out now

beacon_chain/spec/beacon_time.nim
```diff
@@ -174,6 +174,12 @@ func lightClientOptimisticUpdateSlotOffset*(timeParams: TimeParams): TimeDiff =
   TimeDiff(nanoseconds:
     timeParams.NANOSECONDS_PER_SLOT.int64 div INTERVALS_PER_SLOT)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.0/specs/gloas/beacon-chain.md#introduction
```
tersec: beta.1

tersec: https://github.com/status-im/nimbus-eth2/pull/7646/commits/3ea39744ba7603ba3738b67e9742603a974b8fa5

beacon_chain/consensus_object_pools/execution_payload_pool.nim
```diff
@@ -0,0 +1,142 @@
...
+    Opt.none(SignedExecutionPayloadBid)
+
+proc prune*(pool: var ExecutionPayloadBidPool, finalizedSlot: Slot) =
+  var
+    toRemoveBids: seq[BidKey]
+    toRemoveHighest: seq[(Slot, Eth2Digest)]
+
+  try:
+    for key, detail in pool.bids:
+      if detail.bid.message.slot < finalizedSlot:
```
tersec: How much of a concern is a whole-pool iteration, as this pruning code does? It's conceptually inefficient, but whether it's enough to be problematic.

Tomi-3-0: Very true.
I don't have a figure/data on what the pool size for the bids will potentially look like and the O(n) iteration could very well be bad.
We could potentilaly prune by slots O(slots) instead of bids O(bids)

Tomi-3-0: The performance gain from pruning by slots was negligible for large pools. The new approach stores only the highest-value bid per slot.

tersec:
`beacon_chain/gossip_processing/gossip_validation.nim` merge conflict.

Tomi-3-0:
> `beacon_chain/gossip_processing/gossip_validation.nim` merge conflict.

12b43536d31c994a1ba426dea9c4675c7dc70a78

PR #7652, opened 2025-10-22
Derive NANOSECONDS_PER_SLOT from TimeParams

etan-status:
Have to dynamically compute NANOSECONDS_PER_SLOT, it's not a constant.


arnetheduck:
> SECONDS_PER_SLOT

how about making this one a `Duration` object in the TimeConfig - then NANOSECONDS_PER_SLOT is no longer needed..

etan-status:
Maybe separately, after the compiletime SECONDS_PER_SLOT is fully gone

etan-status:
They are making a new millisecond / percentage based approach btw. Not convinced that percentages reflect reality as e.g., network latencies are kinda assumed like that to scale with the slot time.

image

PR #7657, opened 2025-10-22
Gloas beacon block validation update

ahshum:
Cover the changes of

> There are no new validations for this topic. However, all validations with
> regards to the `ExecutionPayload` are removed:
> - If `execution_payload` verification of block's parent by an execution node is
>   *not* complete:
>   - [REJECT] The block's parent (defined by `block.parent_root`) passes all
>     validation (excluding execution node verification of the
>     `block.body.execution_payload`).
> - otherwise:
>   - [IGNORE] The block's parent (defined by `block.parent_root`) passes all
>     validation (including execution node verification of the
>     `block.body.execution_payload`).

> And instead the following validations are set in place with the alias
> `bid = signed_execution_payload_bid.message`:
> - If `execution_payload` verification of block's execution payload parent by an
>   execution node **is complete**:
>   - [REJECT] The block's execution payload parent (defined by
>     `bid.parent_block_hash`) passes all validation.
> - [REJECT] The bid's parent (defined by `bid.parent_block_root`) equals the
>   block's parent (defined by `block.parent_root`).

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -400,7 +400,7 @@ template validateBeaconBlockBellatrix(
   # cannot occur here, because Nimbus's optimistic sync waits for either
   # `ACCEPTED` or `SYNCING` from the EL to get this far.
 
-# https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.1/specs/gloas/p2p-interface.md#beacon_block
+# https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/p2p-interface.md#beacon_block
 template validateBeaconBlockDeneb(
     _: ChainDAGRef,
     _:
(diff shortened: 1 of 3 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -394,6 +392,40 @@ template validateBeaconBlockDeneb(
...
+template validateBeaconBlockGloas(
+    dag: ChainDAGRef,
+    signed_beacon_block: gloas.SignedBeaconBlock): untyped =
+  template blck: untyped = signed_beacon_block.message
+  template bid: untyped = blck.body.signed_execution_payload_bid.message
+
+  # If `execution_payload` verification of block's execution payload parent by
+  # an execution node **is complete**
+  debugGloasComment("update is_execution_block")
+  if signed_beacon_block.is_execution_block:
```
ahshum: According to https://github.com/status-im/nimbus-eth2/blob/b2276f6c527b1fdc9a9a470ad895c9a3813de4fd/beacon_chain/gossip_processing/gossip_validation.nim#L929-L934

It seems that we need some sort of refactoring for this?

ahshum: https://github.com/status-im/nimbus-eth2/blob/b2276f6c527b1fdc9a9a470ad895c9a3813de4fd/beacon_chain/gossip_processing/gossip_validation.nim#L362-L367

PR #7658, opened 2025-10-23
Remove compile-time SECONDS_PER_SLOT

etan-status:
Addresses open issue from #2710 where SECONDS_PER_SLOT required config via compile-time flag. It now loads fine from `config.yaml`.


etan-status:
Needs #7654

PR #7660, opened 2025-10-23
Re-enable tests that were botched in v1.6.0-beta.1

etan-status:
Certain light client tests no longer worked because the config.yaml was incomplete: https://github.com/ethereum/consensus-specs/pull/4652

Tests are being fixed with next spec release and can be turned on again once available.


etan-status:
Needs the next consensus-specs release

PR #7664, opened 2025-10-27
Check parent block full

ahshum:
Context 
https://discord.com/channels/595666850260713488/1479435525084545224

Gossip validation, new block and state replay since Gloas will need this to decide whether should an envelope apply or not.


beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -179,6 +189,21 @@ func get_ancestor*(
...
+func get_parent_payload_status*(blck: BlockRef): PayloadStatus =
+  if isNil(blck.parent) or
+      # When block_hash is `Opt.none`, it means it requires loading data from
+      # dag. Returns as PENDING as it is waiting the data to be filled.
+      blck.parentBlockHash.isNone or
+      blck.parent.executionBlockHash.isNone:
+    PAYLOAD_STATUS_PENDING
+  elif blck.parentBlockHash.value() != blck.parent.executionBlockHash.value() or
+      # Check either one should be enough to ensure both values are non-zero.
+      blck.parentBlockHash.value() == ZERO_HASH:
```
tersec: This is kind of slow, better to use `isZero`:

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1016,6 +1016,88 @@ proc advanceSlots*(
...
+
+      var cur = blck.parent
+      while cur != nil and cur.executionBlockHash.isNone:
+        cur.executionBlockHash = blck.executionBlockHash
+        cur = cur.parent
+
+    if blck.executionParentHash == static(Opt.some(ZERO_HASH)):
+      # Same as executionBlockHash
+      var cur = blck.parent
+      while cur != nil and cur.executionParentHash.isNone:
```
tersec: Why does this traverse the block parents twice, with overlapping termination conditions?

The difference is whether `cur.executionBlockHash.isNone` or `cur.executionParentHash.isNone` but it can accumulate both along the way, separately, and terminate when it gets both.

The other general question is, this PR replaces the `loadExecutionBlockHash` with loadExecutionAndParentBlockHash` even for users which don't need the parent block hash. This means all existing users get added overhead which causes this additional block backtracking which immediately gets thrown away.

ahshum: I think the traverse of parent hash can be removed as the new attribute should only be used when approaching to Gloas (i.e. includes a few epoch of Fulu), which it should not be useful.

For the second question, we will need the parent hash for Fulu block while both for Gloas block. As we are able to get both in one single db read, it should be easier for Fulu-Gloas handling instead of reading the db twice.

If it is a concern, how do you think if I keep the Electra and before unchanged, and add parent hash for Fulu and after?

ahshum: Sorry I was wrong. We should only need parent hash for Gloas blocks only as it is for the envelope transition, which should have included in the block transition for Fulu and before blocks.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1039,6 +1039,80 @@ proc advanceSlots*(
...
+      blck.markExecutionValid(true)
+
+      var cur = blck.parent
+      while cur != nil and cur.executionBlockHash.isNone:
+        cur.executionBlockHash = blck.executionBlockHash
+        cur = cur.parent
+
+  (blck.executionBlockHash, blck.executionParentHash)
+
+proc isParentBlockFull(blck: BlockRef): bool =
```
tersec: `func`

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1039,6 +1039,80 @@ proc advanceSlots*(
...
+  ## other envelope checks to work properly.
+
+  if blck.executionParentHash.isNone() or
+      blck.parent.executionBlockHash.isNone() or
+      blck.executionParentHash.get().isZero():
+    false
+  else:
+    blck.executionParentHash.get() == blck.parent.executionBlockHash.get()
+
+proc isParentBlockFull(blck: gloas.SignedBeaconBlock, parent: BlockRef): bool =
```
tersec: `func`

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1039,6 +1039,80 @@ proc advanceSlots*(
...
+  (blck.executionBlockHash, blck.executionParentHash)
+
+proc isParentBlockFull(blck: BlockRef): bool =
+  ## Since Gloas, we want to skip applying envelope if the envelope of its
+  ## parent is orphaned. This is particularly useful for updateState() as
+  ## orphaned envelopes, even if they are valid, should not be applied to state
+  ## in order to transition to the correct position.
+  ##
+  ## There is a helper but it uses state which is not helpful for updateState().
+  ## https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#new-is_parent_block_full
```
tersec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.3/specs/gloas/beacon-chain.md#new-is_parent_block_full

tersec:
Interesting, ok, there are side effects here. Change back to `proc`, update URL, and can merge.

beacon_chain/consensus_object_pools/blockchain_dag.nim
ahshum: Updated the comment slightly https://github.com/status-im/nimbus-eth2/pull/7664/commits/48e245bfa4f4f6291dee0cc9874b9fd3b05f63c4.

beacon_chain/consensus_object_pools/blockchain_dag.nim
ahshum: No you're right. I think I updated at wrong places.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1039,6 +1039,83 @@ proc advanceSlots*(
...
+      blck.markExecutionValid(true)
+
+      var cur = blck.parent
+      while cur != nil and cur.executionBlockHash.isNone:
+        cur.executionBlockHash = blck.executionBlockHash
+        cur = cur.parent
+
+  (blck.executionBlockHash, blck.executionParentHash)
+
+func isParentBlockFull(blck: BlockRef): bool =
```
arnetheduck, line 1084: this function is unused in `devel`

ahshum: Yes this and another one are not needed. Will remove them entirely once cleaning up the call sites.

ahshum: - https://github.com/status-im/nimbus-eth2/pull/8335

PR #7667, opened 2025-10-27
Fix spec

ahshum:


beacon_chain/spec/beaconstate.nim
```diff
@@ -2873,7 +2873,7 @@ func can_advance_slots*(
   withState(state): forkyState.can_advance_slots(block_root, target_slot)
 
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-get_ptc
-iterator get_ptc(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
+iterator get_ptc*(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
```
tersec: This is likely to need to be exported, yes, but it can be exported when it has an actual caller

tersec: https://github.com/status-im/nimbus-eth2/pull/7667/commits/ea23ae0e1a991583228a575b32d99c3b54561d25

PR #7669, opened 2025-10-27
Remove nimOldCaseObjects & Bump nim-confutils

nitely:
- https://github.com/status-im/nim-confutils/pull/117

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -179,6 +179,7 @@ if canEnableDebuggingSymbols:
 switch("warningAsError", "BareExcept:on")
 switch("warningAsError", "CStringConv:on")
 switch("warningAsError", "UnusedImport:on")
+switch("warningAsError", "CaseTransition:on")
 switch("hintAsError", "ConvFromXtoItselfNotNeeded:on")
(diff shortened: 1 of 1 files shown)


tersec:
Well, some of the serialization also uses Nim old case objects, I believe.

nitely:
@tersec I just added `switch("warningAsError", "CaseTransition:on")` to config to make sure the compilation fails if that's the case. `nim-serialization` [does not support case objects](https://github.com/status-im/nim-serialization/blob/c80ddfb1094560c5270dda2932289c5161ea9298/serialization/object_serialization.nim#L229) (yet). Json and toml serializers fail with the nim-ser error, I checked.

Only nim-json-serilizer [has it defined](https://github.com/search?q=org%3Astatus-im%20nimOldCaseObjects&type=code) but for a test that overloads a readValue case object.

Unless there is some case where the Nim compiler does not warn about CaseTransition and the nimbus test suite does not cover that case, it should be fine. Otherwise, it'll throw a runtime FieldDefect.

PR #7674, opened 2025-10-29
Optimize DataColumn processing.

cheatfate:


beacon_chain/consensus_object_pools/blob_quarantine.nim
```diff
@@ -601,6 +594,21 @@ proc popSidecars*(
...
   Opt.some(sidecars)
 
+proc popSidecars*(
+    quarantine: var ColumnQuarantine,
+    blockRoot: Eth2Digest,
+    blck: fulu.SignedBeaconBlock
+): Opt[seq[ref fulu.DataColumnSidecar]] =
+  ## Function returns sequence of column sidecars for block root ``blockRoot``
+  ## and block ``blck``.
+  ## If some of the column sidecars are missing Opt.none() is returned.
```
arnetheduck: This function should be removed - otherwise, it becomes difficult to maintain the code since the responsibility sometimes is with the quarantine and sometimes is with the calling code.

arnetheduck (approved):

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -375,22 +375,18 @@ proc processDataColumnSidecar*(
...
   self.dataColumnQuarantine[].put(block_root, newClone(dataColumnSidecar))
 
-  if (let o = self.quarantine[].popSidecarless(block_root); o.isSome):
-    withBlck(o[]):
-      when consensusFork >= ConsensusFork.Fulu and
-          consensusFork < ConsensusFork.Gloas:
-        let cres =
-          self.dataColumnQuarantine[].popSidecars(block_root, forkyBlck)
-        if cres.isSome():
+  let cres = self.dataColumnQuarantine[].popSidecars(block_root)
```
arnetheduck: actually, this is not entirely correct: consider what happens when the data columns arrive before the block - the columns will be lost because `popSidecarless` will return `none` in this case - popping should be attempted if we also have the block.

beacon_chain/consensus_object_pools/blob_quarantine.nim
arnetheduck, line 597: remove?

arnetheduck: this comment is for the `proc popSidecars` function below, but for some reason github doesn't want it there

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -378,7 +378,10 @@ proc enqueueQuarantine(self: ref BlockProcessor, parent: BlockRef) =
         const sidecarsOpt = noSidecars
       elif consensusFork == ConsensusFork.Fulu:
         let sidecarsOpt =
-          self.dataColumnQuarantine[].popSidecars(forkyBlck.root, forkyBlck)
+          if len(forkyBlck.message.body.blob_kzg_commitments) == 0:
+            Opt.some(default(fulu.DataColumnSidecars))
+          else:
+            self.dataColumnQuarantine[].popSidecars(forkyBlck.root)
       elif consensusFork in ConsensusFork.Deneb .. ConsensusFork.Electra:
         let sidecarsOpt = self.blobQuarantine[].popSidecars(forkyBlck.root, forkyBlck)
```
arnetheduck, line 386: check for sidecar count in deneb..electra too ..

beacon_chain/gossip_processing/eth2_processor.nim
arnetheduck, line 277: and here..

beacon_chain/gossip_processing/block_processor.nim
cheatfate: But BLOB version of this function will perform check inside by itself.

arnetheduck: ideally both would work the same to remove confusion

arnetheduck (approved):

beacon_chain/gossip_processing/block_processor.nim
cheatfate: Ideally `BlobSidecar` will hold information about number of kzg_commitments but it is not, so without block popSidecars() will not work, just because it does not have information on how many `BlobSidecars` should be.

beacon_chain/gossip_processing/eth2_processor.nim
cheatfate: same problem.

arnetheduck (approved):

beacon_chain/gossip_processing/block_processor.nim
arnetheduck: fair .. that said, we can simply not call pop here if there are no commitments without changing the blob quarantine code - anyway, it's not a big deal for now, it's just easier when the code has fewer special cases for each fork. This was one of the motivating reasons to rename the variables to `sidecarsOpt` as well, because once we have sidecars/columns, their "onward" processing is logically the same, even when done with different types / details.

PR #7686, opened 2025-10-30
stop using GitHub caching actions for reference tests

tersec:


etan-status:
Makes sense. There are more frequent occasions where they manually replace release assets for existing tags to not be related to the source anymore, or move tags around to different commits.

PR #7687, opened 2025-10-30
Fix Nix build without git

bbjubjub2494:
buildinfo does not accept "unknown" because it's not a 8-character hex string.

https://github.com/status-im/nimbus-eth2/blob/9839f140628ae0e2e8aa7eb055da5c4bb08171d0/beacon_chain/buildinfo.nim#L42


arnetheduck:
Thanks!

PR #7692, opened 2025-10-31
unconditionally use int64 for times

tersec:

PR #7694, opened 2025-10-31
Replace `INTERVALS_PER_SLOT` with BPS based slot timings

etan-status:
This enables configuration of:

- `PROPOSER_REORG_CUTOFF_BPS` (only used for REST API and constraints)
- `ATTESTATION_DUE_BPS` (must be ≥ reorg cutoff)
- `AGGREGATE_DUE_BPS` (must be ≥ att cutoff and ≤ 10k)
- `SYNC_MESSAGE_DUE_BPS` (must be == att cutoff)
- `CONTRIBUTION_DUE_BPS` (must be == agg cutoff)

When `INTERVALS_PER_SLOT` was retired, they changed it to be slightly off, 1/3 became 3333/10000 and 2/3 became 6667/10000, so there is a tiny bit more time to send attestations, and a tiny bit less for aggregates.

Also updated VC deadlines to be based on the dynamic config.

The constraints are just what Nimbus assumes. The spec does not define valid ranges of config values.


etan-status:
Needs

- #7692
- #7693

tersec:
> Needs
> 
>     * [unconditionally use int64 for times #7692](https://github.com/status-im/nimbus-eth2/pull/7692)
> 
>     * [Remove unused attestation v1 code #7693](https://github.com/status-im/nimbus-eth2/pull/7693)

Both of these have been merged.

PR #7701, opened 2025-11-01
Apply policy: Retain back through Mainnet's second latest fork

etan-status:
In v1.6.0 specs (#7696) the mainnet config.yaml was updated to use Fulu.

- https://github.com/ethereum/consensus-specs/blob/v1.6.0/configs/mainnet.yaml#L57

### DIFF
--- a/beacon_chain/consensus_object_pools/blockchain_dag.nim
+++ b/beacon_chain/consensus_object_pools/blockchain_dag.nim
@@ -2382,8 +2382,6 @@ proc processVanityLogs(dag: ChainDAGRef, vanityState: auto) =
         handler()
 
     # Policy: Retain back through Mainnet's second latest fork.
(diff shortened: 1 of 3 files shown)


etan-status:
Not sure if the policy should be applied based on what the EF wants (consensus-specs) or what the community wants (eth-clients/mainnet). Only the former is updated for Fulu.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -2382,8 +2382,6 @@ proc processVanityLogs(dag: ChainDAGRef, vanityState: auto) =
         handler()
 
     # Policy: Retain back through Mainnet's second latest fork.
```
arnetheduck, line 2384: still sad and boring, as far as policies go

PR #7704, opened 2025-11-02
LC: Switch to pluggable backend for LC updates

chirag-parmar:


beacon_chain/sync/light_client_manager.nim
```diff
@@ -52,6 +52,31 @@ type
...
+    async: (raises: [CancelledError])
+  .}
+  LightClientOptimisticUpdateProc = proc(
+    id: uint64
+  ): Future[NetRes[ForkedLightClientOptimisticUpdate]] {.
+    async: (raises: [CancelledError])
+  .}
+  UpdateScoreProc = proc(id: uint64, value: int) {.gcsafe, raises: [].}
+
+  EthLCBackend* = object
```
kdeme, line 73: Light client naming is already sufficiently confusing with all the types of light clients.

I would suggest to either stick here with naming scheme of the existing file: `LightClientBackend`.
Or to call it `ConsensusLightClientBackend`

PR #7711, opened 2025-11-05
signatures: reduce fork code bloat

arnetheduck:
`withState` causes unnecessary template bloat compared to `forky` - also, `items` is ok to use post 1.4.2


tersec:
Merge conflict on `beacon_chain/spec/signatures_batch.nim`.

PR #7713, opened 2025-11-05
Clean up genesis init helpers

arnetheduck:
* create testnet/test genesis starting from phase0 and upgrade to the correct fork, like the spec suggests
* reduce mem usage and number of copies while upgrading
* add gloas to `ncli_testnet` generator
* speed up mock deposit processing by skipping bls ops for most state creations and caching them for the others (attestation pool tests go from minutes to seconds)
* consolidate the various ways to create deposits and states in the test suite to a single flow
* make sure deposit creation uses the correct fork version in tests


etan-status (commented):
There's something about current/next fork fields that's different when doing sequential upgrade instead of a usual genesis (where both fields are the same)

arnetheduck:
> There's something about current/next fork fields that's different when doing sequential upgrade instead of a usual genesis (where both fields are the same)

hm, yeah, we had these checks that I don't know where they came from that the genesis fork has the same versions in current/prev version .. I think that's our own invention and not a spec requirement per se so it was removed in https://github.com/status-im/nimbus-eth2/pull/7705 - hopefully I nabbed all of them..

etan-status:
Maybe an option to check genesis states from hoodi etc which started post phase0

arnetheduck:
https://github.com/eth-clients/hoodi/blob/main/parsed/parsedConsensusGenesis.json:

```
arnetheduck@praeceps:~/Downloads$ grep _version parsedConsensusGenesis.json 
    "previous_version": "0x40000910",
    "current_version": "0x50000910",
```

ncli/ncli_testnet.nim
```diff
@@ -392,41 +404,37 @@ proc doCreateTestnet*(config: CliConfig,
...
+      when consensusFork >= ConsensusFork.Bellatrix and
+          consensusFork < ConsensusFork.Gloas:
+        debugGloasComment ""
+        forkyState.data.latest_execution_payload_header =
+          genesisBlock as consensusFork.ExecutionPayloadHeader
+      forkyState.root.reset()
+
+      doAssert forkyState.data.validators.len > 0
+
+      # let outGenesisExt = splitFile(outGenesis).ext
```
tersec, line 423: Is the commented-out code intentional?

tersec: Oh, it's in the current code too.

PR #7714, opened 2025-11-05
Block command-line supplied invalid blocks earlier

arnetheduck:
For some reason, they're only blocked in the CL fallback of EL processing, which means it doesn't work with a responding EL.

Also add initial smoke tests for non-bellatrix block processing.


tersec:
Merge conflict in `beacon_chain/gossip_processing/block_processor.nim`

tests/testblockutil.nim
```diff
@@ -73,92 +87,94 @@ proc makeInitialDeposits*(
...
-  ## Assuming a pre-state of the same slot, build a valid ExecutionPayload
-  ## without any transactions from a non-merged block.
-
-  doAssert not is_merge_transition_complete(state)
+    consensusFork: static ConsensusFork,
+    state: ForkyBeaconState,
+    blobsBundle: BlobsBundle,
+): consensusFork.ExecutionPayloadForSigning =
+  ## Construct an execution payload that is sufficiently valid to pass consensus
+  ## validations (without necessarily making sense on the execuction side, which
```
tersec: "execution side"

tests/testblockutil.nim
```diff
@@ -88,92 +102,94 @@ proc makeInitialDeposits*(
...
```
arnetheduck:
```suggestion
  ## validations (without necessarily making sense on the execution side, which
```

PR #7731, opened 2025-11-14
fix: use correct committee validator bc1[1] for attestation2

Bilogweb3:
Copy-paste bug: attestation2 created from committee bc1[1] was added to pool with validator from bc0[1] instead.
Test verifies cross-committee aggregation, so correct committee assignment is critical for test validity.


tersec:
Feel free to reopen this against `unstable`; it looks like a good fix.

PR #7733, opened 2025-11-15
rm Electra builder API support

tersec:
Don't merge until at least mainnet forks to Fusaka.

### DIFF
--- a/AllTests-mainnet.md
+++ b/AllTests-mainnet.md
@@ -902,15 +902,15 @@ AllTests-mainnet
 ```
 ## MEV calls serialization/deserialization and behavior test suite
 ```diff
-+ /eth/v1/builder/blinded_blocks [json/json] test                                            OK
-+ /eth/v1/builder/blinded_blocks [json/ssz] test                                             OK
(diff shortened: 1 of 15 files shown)


etan-status:
Gnosis still missing, right? Do they use external builders?
Note they had an EL only hard fork: https://github.com/gnosischain/specs/pull/87

tersec:
Gnosis is why I haven't merged this yet, yes. Still nothing in https://github.com/gnosischain/specs/blob/master/network-upgrades/fusaka.md#upgrade-schedule

etan-status:
Yeah, it's tricky because of their accelerated slot timings. One idea was to use a column count of 1, effectively disabling PeerDAS and using the existing everyone-is-a-supernode logic. But I think we have some round-down logic in some division cases where odd numbers wouldn't work.

tersec:
Gnosis evidently [does use MEV](https://agnostic-relay.net/).

tersec:
- https://github.com/status-im/nimbus-eth2/pull/8144

PR #7748, opened 2025-11-21
Envelope processing

ahshum:


beacon_chain/spec/helpers_el.nim
```diff
@@ -20,22 +20,38 @@ func readExecutionTransaction(
...
-  template transactions: untyped = blck.body.execution_payload.transactions
-  template commitments: untyped = blck.body.blob_kzg_commitments
+func is_valid_versioned_hashes*(
+    blck: ForkyBeaconBlock,
+    envelope: NoEnvelope | gloas.ExecutionPayloadEnvelope,
+): Result[void, string] =
+  const consensusFork = typeof(blck).kind
+  static: doAssert consensusFork >= ConsensusFork.Deneb
+
+  template forkyTransactions: untyped =
```
tersec: Typically when Nimbus uses `forkyFoo`, it means the actual `Foo` (`Transaction`, say) are different. It seems like in this case the transactions themselves are the same, just housed differently?

beacon_chain/spec/helpers_el.nim
```diff
@@ -20,22 +20,38 @@ func readExecutionTransaction(
...
+    when consensusFork >= ConsensusFork.Gloas:
+      envelope.payload.transactions
+    else:
+      blck.body.execution_payload.transactions
+  template forkyCommitments: untyped =
+    when consensusFork >= ConsensusFork.Gloas:
```
tersec, line 36: Similar point here -- are the commitments any different, or just in a different part of the block/envelope?

beacon_chain/spec/helpers_el.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/commits/b1ad20efe58add9e4d21fa323f43a27f64c0e0c8

beacon_chain/spec/helpers_el.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/commits/b1ad20efe58add9e4d21fa323f43a27f64c0e0c8

tersec:
Merge conflict in `beacon_chain/nimbus_beacon_node.nim`.

tersec:
Merge conflict with in `beacon_chain/gossip_processing/block_processor.nim`

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -459,37 +505,49 @@ proc onBlockAdded*(
...
-  when consensusFork == ConsensusFork.Gloas:
-    debugGloasComment "no exection payload field for gloas"
-    ok OptimisticStatus.valid
-  elif consensusFork >= ConsensusFork.Bellatrix:
+  when consensusFork >= ConsensusFork.Bellatrix:
+    # Since Gloas, is_execution_block should always be true.
     if signedBlock.message.is_execution_block:
-      template payload(): auto =
-        signedBlock.message.body.execution_payload
+      template paylaod(): auto =
```
tersec: misspelling/typo: should be `payload` (causes several other `paylaod`s later also to fix)

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -467,6 +467,71 @@ proc addBackfillBlock*(
...
+    info "Envelope mismatches with this block"
+    return err(VerifierError.Invalid)
+
+  # Check with the DAG head.
+  let blck = dag.head
+  if blck.root() != envelopeBlockRoot():
+    debug "Envelope is not for the current head"
+    return err(VerifierError.Invalid)
+  if blck.slot() != signedEnvelope.message.slot:
+    # TODO: check if there is any envelopes from other slots being processed.
```
tersec: Is the intent to do that in this PR or another PR?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2383,6 +2388,19 @@ proc installMessageValidators(node: BeaconNode) =
                 node.processor[].processExecutionPayloadBid(
                   MsgSource.gossip, signedBid)))
 
+        # execution_payload
+        # https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/p2p-interface.md#execution_payload
```
tersec: New spec reference URL should target and be consistent with v1.7.1-alpha.1, e.g. https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/p2p-interface.md#execution_payload

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -467,6 +467,71 @@ proc addBackfillBlock*(
...
+  ).isOkOr:
+    assign(dag.clearanceState, dag.headState)
+    info "Envelope transition failed", msg = error
+    return err(VerifierError.Invalid)
+
+  # Put the envelope into db and update optimistic status for the block.
+  dag.db.putExecutionPayloadEnvelope(signedEnvelope)
+  blck.markExecutionValid(true)
+
+  debugGloasComment("add onEnvelopeAdded ?")
```
tersec: Only if/when there's some specific purpose/utility to it.

beacon_chain/consensus_object_pools/block_clearance.nim
ahshum: I have checked that the source of the signed block is from the database, so the check over slot may not be useful here. I will revisit it later, but cleaned up for now.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/changes/b5e0470465b65e7b636da74fe315efae4ba13c34

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/changes/943738a0b416590a38582e9f360dca2790714582

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/commits/943738a0b416590a38582e9f360dca2790714582

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -36,7 +36,7 @@ from ../validators/validator_monitor import
   MsgSource, ValidatorMonitor, registerAttestationInBlock, registerBeaconBlock,
   registerSyncAggregateInBlock
 from ../beacon_chain_db import getBlobSidecar, putBlobSidecar,
-  getDataColumnSidecar, putDataColumnSidecar
+  getDataColumnSidecar, putDataColumnSidecar, putExecutionPayloadEnvelope
```
tersec: Nothing here seems to use `putExecutionPayloadEnvelope`?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -940,11 +949,29 @@ proc storePayload(
           chronos.nanoseconds((slotTime - wallTime).nanoseconds)
     deadline = sleepAsync(deadlineTime)
```
tersec: This won't do anything on its own. `deadline` here is a `Future`. To execute that future, one has to, e.g., `await` it or otherwise signal to Chronos to schedule and run it. And since nothing else in this function refers to `deadline`, it's effectively unused code.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -940,11 +949,29 @@ proc storePayload(
...
-
+  let
+    optimisticStatusRes =
+      block:
+        debugGloasComment("handle (maybe)finalized slot")
+        func shouldRetry(): bool =
+          not dag.is_optimistic(dag.head.bid)
+        await self.consensusManager.elManager.getExecutionValidity(
+          signedBlock, signedEnvelope, deadline, shouldRetry())
+    optimisticStatus =
```
tersec, line 977: `optimisticStatus` is unused here; does it have an intended use?

ahshum: Yes I think I missed this...

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/changes/24cf98fd4c0c446cee57ba7cf753897f2e68f5c0

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7748/changes/24cf98fd4c0c446cee57ba7cf753897f2e68f5c0

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -940,11 +949,36 @@ proc storePayload(
...
+      ?(optimisticStatusRes or verifyPayload(self, signedBlock, signedEnvelope))
```
tersec, line 978:
```nim
Result[OptimisticStatus, VerifierError].ok(OptimisticStatus.invalidated) or
  Result[OptimisticStatus, VerifierError].ok(OptimisticStatus.notValidated) ==
  Result[OptimisticStatus, VerifierError].ok(OptimisticStatus.invalidated)
```
which seems fine, but
```nim
Result[OptimisticStatus, VerifierError].ok(OptimisticStatus.notValidated) or
  Result[OptimisticStatus, VerifierError].ok(OptimisticStatus.invalidated) ==
  Result[OptimisticStatus, VerifierError].ok(OptimisticStatus.notValidated)
```
which seems questionable. It won't catch any of the `OptimisticStatus.invalidated` cases in https://github.com/status-im/nimbus-eth2/blob/42eeb424271ff69dd0d1966b257ca5a4ca0c0bf6/beacon_chain/gossip_processing/block_processor.nim#L535-L561

I'm not sure that https://github.com/status-im/nimbus-eth2/blob/42eeb424271ff69dd0d1966b257ca5a4ca0c0bf6/beacon_chain/gossip_processing/block_processor.nim#L688-L698 is correct, for the same reason.

ahshum: Yes I think the idea is to fallback to `verifyPayload()` unless EL fails to verify for whatever reasons, and which should also keep the payload verification behavior same as previous forks.

https://github.com/status-im/nimbus-eth2/blob/42eeb424271ff69dd0d1966b257ca5a4ca0c0bf6/beacon_chain/gossip_processing/block_processor.nim#L661-L698

ahshum: https://github.com/status-im/nimbus-eth2/blob/42eeb424271ff69dd0d1966b257ca5a4ca0c0bf6/beacon_chain/gossip_processing/block_processor.nim#L688-L698

is because there is not execution payload in Gloas block, which we could not determine whether it is invalid or valid. Yes I think it is confusing now but the status is passed/used in different places in block processing so I think this is something I need to clean up and refactor.

tersec: Alright, it's not exactly wrong currently, but a performance tradeoff. A CL isn't expected to run any analogue of `verifyPayload()` at all, it's just that Nimbus does to support some additional functionality. And at some point the EL should respond `INVALID`, and resolve this.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -940,11 +949,36 @@ proc storePayload(
...
+
+  # optimisticStatus could be valid or notValidated at this point. We will
+  # validated it by the clearance state transition.
```
tersec: `validated` -> `validate`

tersec: https://github.com/status-im/nimbus-eth2/pull/7748/commits/b22c971bb1da81728c3e262a081a8ff4d1d0bca8

tersec:

beacon_chain/gossip_processing/block_processor.nim
tersec, line 951: Nothing in the `addPayload` except `sleepAsync` currently requires it to be an async function, and the `sleepAsync` isn't even called. Will it have to be async, and why?

ahshum: I think the request to EL (`getExecutionValidity`) is async?

tersec: That's true, it is.

PR #7750, opened 2025-11-23
Add `getPayloadAttestationSignature` for local signing 

Tomi-3-0:


beacon_chain/validators/validator_pool.nim
```diff
@@ -799,3 +799,21 @@ proc getBuilderSignature*(v: AttachedValidator, genesis_fork_version: Version,
...
+  case v.kind
+  of ValidatorKind.Local:
+    SignatureResult.ok(
+      get_payload_attestation_message_signature(
+        fork, genesis_validators_root, message,
+        v.data.privateKey).toValidatorSig())
+  of ValidatorKind.Remote:
+    let request =
+      Web3SignerRequest.init(fork, genesis_validators_root, message)
+    echo "PayloadAttestationMessage JSON: ", RestJson.encode(request)
```
cheatfate: I think this should be removed.

beacon_chain/spec/eth2_apis/rest_remote_signer_calls.nim
```diff
@@ -167,6 +167,7 @@ proc signData*(client: RestClientRef, identifier: ValidatorPubKey,
         let res = decodeBytes(Web3SignerErrorResponse, response.data,
                               response.contentType)
         if res.isErr():
+          echo "Raw 400 response: ", fromBytes(string, response.data)
```
cheatfate: This also should not be present.

cheatfate (changes requested):
Please remove debugging echoes.

beacon_chain/validators/validator_pool.nim
Tomi-3-0: yes. Thanks.

cheatfate (approved):

tersec:
Where are the specs for this defined?

Tomi-3-0:
> Where are the specs for this defined?

That's correct,  the `Web3Signer` specs doesn't officially support it yet. this is getting ahead of the official specs. I will limit this to local signing

PR #7757, opened 2025-11-26
Track ptc assignment

Tomi-3-0:
https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/validator.md#payload-timeliness-committee
https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/validator.md#lookahead

### DIFF
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -2764,7 +2764,7 @@ func can_advance_slots*(
 
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-get_ptc
 iterator get_ptc*(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
-    ValidatorIndex =
+    ValidatorIndex {.closure.} =
   ## Get the payload timeliness committee for the given ``slot``
   let epoch = slot.epoch()
   var buffer {.noinit.}: array[40, byte]
--- a/beacon_chain/spec/beaconstate.nim
+++ b/beacon_chain/spec/beaconstate.nim
@@ -2763,7 +2763,7 @@ func can_advance_slots*(
   withState(state): forkyState.can_advance_slots(block_root, target_slot)
 
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-get_ptc
-iterator get_ptc(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
+iterator get_ptc*(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
     ValidatorIndex =
   ## Get the payload timeliness committee for the given ``slot``
   let epoch = slot.epoch()
--- a/beacon_chain/validators/action_tracker.nim
+++ b/beacon_chain/validators/action_tracker.nim
@@ -38,6 +38,10 @@ type
     subnet_id*: SubnetId
     slot*: Slot
 
+  PTCDuty* = object
+    slot*: Slot
+    validator_index: ValidatorIndex
+
   ActionTracker* = object
     nodeId: UInt256
 
@@ -72,6 +76,9 @@ type
 
     lastSyncUpdate*: Opt[SyncCommitteePeriod]
     syncDuties*: Table[ValidatorPubKey, Epoch]
+
+    ptcSlots: array[2, uint32]
+    ptcDuties*: HashSet[PTCDuty]
 
 func hash*(x: AggregatorDuty): Hash =
   hashAllFields(x)
@@ -116,6 +123,34 @@ func hasSyncDuty*(
     tracker: ActionTracker, pubkey: ValidatorPubKey, epoch: Epoch): bool =
   epoch < tracker.syncDuties.getOrDefault(pubkey, GENESIS_EPOCH)
 
+proc registerPTCDuty*(
+    tracker: var ActionTracker, slot: Slot, vidx: ValidatorIndex) =
+  if slot < tracker.currentSlot or
+      slot + (SLOTS_PER_EPOCH + 2) <= tracker.currentSlot:
+    debug "Irrelevant PTC duty", slot, vidx
+    return
+
+  tracker.knownValidators[vidx] = slot
+
+  let newDuty = PTCDuty(slot: slot, validator_index: vidx)
+
+  if newDuty in tracker.ptcDuties:
+    return
+
+  debug "Registering PTC duty", slot, vidx
+  tracker.ptcDuties.incl(newDuty)
+
+func hasPTCDuty*(tracker: ActionTracker, slot: Slot): bool =
+  let epoch = slot.epoch
+  if epoch > tracker.lastCalculatedEpoch + 1:
+    return false
+  ((tracker.ptcSlots[epoch mod 2] and (1'u32 shl (slot mod SLOTS_PER_EPOCH))) != 0)
+
+func getPTCDuties*(tracker: ActionTracker, slot: Slot): seq[ValidatorIndex] =
+  for duty in tracker.ptcDuties:
+    if duty.slot == slot:
+      result.add(duty.validator_index)
+
 func aggregateSubnets*(tracker: ActionTracker, wallSlot: Slot): AttnetBits =
   var res: AttnetBits
   # Subscribe to subnets for upcoming duties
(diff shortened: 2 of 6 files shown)


siddarthkay:
please re-run macos job I had to kill it to clean up `macm4-01`

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1282,6 +1282,26 @@ proc handleValidatorDuties*(node: BeaconNode, lastSlot, slot: Slot) {.async: (ra
   sendAggregatedAttestations(node, head, slot)
   sendSyncCommitteeContributions(node, head, slot)
 
+proc registerPTCDuties(node: BeaconNode, epoch: Epoch) =
+  if node.dag.cfg.consensusForkAtEpoch(epoch) < ConsensusFork.Gloas:
+    return
+
+  let validatorIndices = toHashSet(toSeq(node.attachedValidators[].indices()))
```
tersec: Can construct the `validatorIndices` `HashSet` directly. It's slightly less elegant, but there's no reason to construct the full intermediate `seq`; it's never used. Unfortunately, Nim doesn't offer great approaches here, but usually I'd just go for the straightforward
```nim
for foo in node.attachedValidators[].indices():
  validatorIndices.incl foo
```
what this loses is the `let` vs `var`. But even that is recoverable:
```nim
  let validatorIndices = block:
    var res: HashSet[ValidatorIndex]
    for ...:
      incl
    res
```

`toSeq` is sometimes required, and sometimes the least bad way of accomplishing something, but unless the actual end result desired is specifically that `seq`, it adds an extra copy/allocation for no net gain.

For tests, sure, no issue with it, but it's worth being at least cautious in the main client code with it.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1282,6 +1282,26 @@ proc handleValidatorDuties*(node: BeaconNode, lastSlot, slot: Slot) {.async: (ra
...
+
+  withState(node.dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      var cache = new StateCache
```
tersec: In general, `StateCache` can be on the stack, not the heap; it's not that large.

Also, for things which are on the heap, one can almost always use `let`, because what that protects isn't the value being pointed/referred to, but the pointer/reference itself. That is, for a reference type, `let` just means one isn't doing pointer arithmetic on it, which is already prohibited outside `cast` gimmicks.

beacon_chain/validators/action_tracker.nim
```diff
@@ -116,6 +123,33 @@ func hasSyncDuty*(
...
+      slot >= Slot(uint64(tracker.currentSlot) + (SLOTS_PER_EPOCH * 2)):
+    debug "Irrelevant PTC duty", slot, vidx
+    return
+
+  tracker.knownValidators[vidx] = slot
+
+  let newDuty = PTCDuty(slot: slot, validator_index: vidx)
+
+  if newDuty in tracker.ptcDuties:
+    return
```
tersec: https://nim-lang.org/docs/sets.html#incl%2CHashSet%5BA%5D%2CA
> This doesn't do anything if `key` is already in `s`.

So the only difference this makes is in the `debug` logging. It adds an extra lookup for non-`debug` logging setups (the usual scenario) for no gain/effect, and a redundant check. Probably better to omit this early return.

beacon_chain/validators/action_tracker.nim
```diff
@@ -116,6 +123,33 @@ func hasSyncDuty*(
...
+
+  debug "Registering PTC duty", slot, vidx
+  tracker.ptcDuties.incl(newDuty)
+
+from std/sequtils import toSeq, anyIt
```
tersec: Prefer `from std/sequtils import anyIt, toSeq` due to lexicographic ordering

beacon_chain/validators/action_tracker.nim
```diff
@@ -116,6 +123,33 @@ func hasSyncDuty*(
...
+
+func hasPTCDuty*(tracker: ActionTracker, slot: Slot): bool =
+  tracker.ptcDuties.anyIt(it.slot == slot)
+
+func getPTCDuties*(tracker: ActionTracker, slot: Slot): seq[ValidatorIndex] =
+  for duty in tracker.ptcDuties:
+    if duty.slot == slot:
+      result.add(duty.validator_index)
```
tersec: https://status-im.github.io/nim-style-guide/language.result.html

beacon_chain/spec/beaconstate.nim
```diff
@@ -2763,8 +2763,8 @@ func can_advance_slots*(
   withState(state): forkyState.can_advance_slots(block_root, target_slot)
 
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.6/specs/gloas/beacon-chain.md#new-get_ptc
-iterator get_ptc(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
-    ValidatorIndex =
+iterator get_ptc*(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
+    ValidatorIndex {.closure.} =
```
tersec, line 2767: This `{.closure.}` doesn't seem necessary?

Tomi-3-0: without it the iterator gets inlined and causes the stack overflow issue

tersec: ok

beacon_chain/validators/action_tracker.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7757/commits/0b07a77af03a26bc4126b3b72b112427300b91c1

beacon_chain/validators/action_tracker.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7757/commits/0b07a77af03a26bc4126b3b72b112427300b91c1

beacon_chain/validators/action_tracker.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7757/commits/0b07a77af03a26bc4126b3b72b112427300b91c1

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7757/commits/0b07a77af03a26bc4126b3b72b112427300b91c1

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1282,6 +1282,30 @@ proc handleValidatorDuties*(node: BeaconNode, lastSlot, slot: Slot) {.async: (ra
...
+
+  let validatorIndices = block:
+    var res: HashSet[ValidatorIndex]
+    for idx in node.attachedValidators[].indices():
+      res.incl(idx)
+    res
+
+  withState(node.dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      let cache = StateCache()
```
tersec:
```
[2025-11-30T21:09:24.446Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7757/beacon_chain/validators/beacon_validators.nim(1295, 12) template/generic instantiation of `withState` from here
[2025-11-30T21:09:24.446Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7757/beacon_chain/validators/beacon_validators.nim(1300, 39) Error: type mismatch
[2025-11-30T21:09:24.447Z] Expression: get_ptc(node.dag.headState.gloasData.data, slot, cache)
[2025-11-30T21:09:24.447Z]   [1] node.dag.headState.gloasData.data: BeaconState
[2025-11-30T21:09:24.447Z]   [2] slot: Slot
[2025-11-30T21:09:24.447Z]   [3] cache: StateCache
[2025-11-30T21:09:24.447Z] 
[2025-11-30T21:09:24.447Z] Expected one of (first mismatch at [position]):
[2025-11-30T21:09:24.447Z] [3] iterator get_ptc(state: gloas.BeaconState; slot: Slot; cache: var StateCache): ValidatorIndex
[2025-11-30T21:09:24.447Z]   expression 'cache' is immutable, not 'var'
```

Without the `ref` type this actually does have to be `var`:
```nim
      var cache: StateCache
```

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7757/commits/0b07a77af03a26bc4126b3b72b112427300b91c1

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/compare/0b07a77af03a26bc4126b3b72b112427300b91c1..c985abdfdc98ac4daf5ca55bc82f984693d6a80b

PR #7761, opened 2025-11-28
Refactor conf with flatten

nitely:
Changes:

- Pack metric opts into `MetricsConf`
- Pack keymanager opts into `keyManagerApiConf`
- Pack `restRequestTimeout`, `restMaxRequestBodySize`, `restMaxRequestHeadersSize` into `RestApiConf`. But this could be passed individually instead.
- Small initKeymanagerServer change to avoid passing rest opts.
- Clean up no longer needed `defaultValueDesc`
- Bumped nim-confutils to https://github.com/status-im/nim-confutils/pull/118

Beware this contains breaking changes; needs https://github.com/status-im/nimbus-eth1/pull/3848 to be merged after to solve them.

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -45,15 +45,10 @@ export
 declareGauge network_name, "network name", ["name"]
 
 const
-  # TODO: How should we select between IPv4 and IPv6
-  # Maybe there should be a config option for this.
-  defaultAdminListenAddress* = (static parseIpAddress("127.0.0.1"))
   defaultSigningNodeRequestTimeout* = 60
   defaultBeaconNode* = "http://127.0.0.1:" & $defaultEth2RestPort
   defaultBeaconNodeUri* = parseUri(defaultBeaconNode)
   defaultGasLimit* = 60_000_000
-  defaultAdminListenAddressDesc* = $defaultAdminListenAddress
-  defaultBeaconNodeDesc = $defaultBeaconNode
 
 when defined(windows):
   {.pragma: windowsOnly.}
@@ -305,13 +300,11 @@ type
       tcpPort* {.
         desc: "Listening TCP port for Ethereum LibP2P traffic"
         defaultValue: defaultEth2TcpPort
-        defaultValueDesc: $defaultEth2TcpPortDesc
         name: "tcp-port" .}: Port
 
       udpPort* {.
         desc: "Listening UDP port for node discovery"
         defaultValue: defaultEth2TcpPort
-        defaultValueDesc: $defaultEth2TcpPortDesc
         name: "udp-port" .}: Port
 
       maxPeers* {.
@@ -403,21 +396,7 @@ type
         defaultValue: 0
         name: "stop-at-synced-epoch" .}: uint64
 
-      metricsEnabled* {.
-        desc: "Enable the metrics server"
-        defaultValue: false
-        name: "metrics" .}: bool
-
-      metricsAddress* {.
-        desc: "Listening address of the metrics server"
-        defaultValue: defaultAdminListenAddress
-        defaultValueDesc: $defaultAdminListenAddressDesc
-        name: "metrics-address" .}: IpAddress
-
-      metricsPort* {.
-        desc: "Listening HTTP port of the metrics server"
-        defaultValue: 8008
-        name: "metrics-port" .}: Port
+      metrics* {.flatten.}: MetricsConf
 
       statusBarEnabled* {.
         posixOnly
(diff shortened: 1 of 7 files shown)


beacon_chain/nimbus_binary_common.nim
```diff
@@ -283,13 +289,35 @@ proc sleepAsync*(t: TimeDiff): Future[void] {.
...
+            "will be received"
+      name: "rest-request-timeout" .}: Natural
+
+    maxBodySize* {.
+      defaultValue: 16_384
+      desc: "Maximum size of REST request body (kilobytes)"
+      name: "rest-max-body-size" .}: Natural
+
+    maxHeadersSize* {.
+      defaultValue: 128
```
nitely, line 309: Note BeaconNodeconf had this set to 64.

beacon_chain/nimbus_binary_common.nim
```diff
@@ -352,41 +414,71 @@ proc initKeymanagerServer*(
...
+      defaultValue: defaultAdminListenAddress
+      defaultValueDesc: $defaultAdminListenAddressDesc
+      name: "metrics-address" .}: IpAddress
+
+    port* {.
+      desc: "Listening HTTP port of the metrics server"
+      defaultValue: 8008
+      name: "metrics-port" .}: Port
+
+  MetricsBetaConf* = object
```
nitely: this exists because of the port difference and the description adding "BETA". But maybe can just use MetricsConf

cheatfate:
If you going to refactor configs it would be nice if you split all the configs between parts, because when VC builds it imports whole BN codebase just because of the `conf.nim`. So it would be nice to have some module with common procedures (which does not import whole BN) and separate `conf.nim` for every binary we have.

scutuatua-crypto (approved):

beacon_chain/nimbus_binary_common.nim
```diff
@@ -283,13 +289,35 @@ proc sleepAsync*(t: TimeDiff): Future[void] {.
   sleepAsync(nanoseconds(
     if t.nanoseconds < 0: 0'i64 else: t.nanoseconds))
 
+type
+  RequestConf* = object
```
kdeme: Little nit, but if we pack this perhaps the naming should be a bit more clear that this is about REST API configuration settings. `RequestConf` seems a bit too general.

example: `RestRequestConf` or `RestApiConf` ?

beacon_chain/conf.nim
```diff
@@ -494,51 +476,9 @@ type
...
-        desc: "Listening port for the REST keymanager API"
-        defaultValue: defaultAdminListenAddress
-        defaultValueDesc: $defaultAdminListenAddressDesc
-        name: "keymanager-address" .}: IpAddress
-
-      keymanagerAllowedOrigin* {.
-        desc: "Limit the access to the Keymanager API to a particular hostname " &
-              "(for CORS-enabled clients such as browsers)"
-        name: "keymanager-allow-origin" .}: Option[string]
+      request* {.flatten.}: RequestConf
```
```suggestion
      restApiConf* {.flatten.}: RestApiConf
```

beacon_chain/conf.nim
```diff
@@ -494,51 +476,9 @@ type
...
 
-      keymanagerTokenFile* {.
-        desc: "A file specifying the authorization token required for accessing the keymanager API"
-        name: "keymanager-token-file" .}: Option[InputFile]
+      keymanager* {.flatten.}: KeymanagerConf
```
arnetheduck: keyManagerApiConf

beacon_chain/nimbus_binary_common.nim
arnetheduck: :+1: we can go with 128 consistently

beacon_chain/nimbus_binary_common.nim
arnetheduck: hm, defaults - forgot that they're different. The (BETA) tag can (and should) have been removed long ago, but we can't change the port defaults easily.

I can think of two options: `flatten: (a: b)`-style overrides that create a tuple or `MetricsConf[defaultPort: static Port]`, thoughts or other options to allow the default to be customized? Not sure if it makes sense to customize anything else?

arnetheduck:
>  nice if you split all the configs between parts

It's not a bad idea in general but I suspect the gains might be smaller than one would think since the VC already needs a significant portion of `spec` which is the main part of what's being imported "redundantly". One would also need to introduce additional structure for parsing those options that are common, so it's probably outside of the scope of this PR.

beacon_chain/nimbus_binary_common.nim
nitely: I think `flatten: (theFieldName: theDefaultValue)` can work.

> Not sure if it makes sense to customize anything else?

probably the `defaultValueDesc`. But it may not be needed, if I get the help message to print const values...

nitely: I implemented `flatten: (a: b)`. The default shows in the help message if it's a lit or a const.

Support for `defaultValueDesc` can be added later if needed. Something like: `flatten: (a: (defaultValue: b, defaultValueDesc: "b"))`.

beacon_chain/nimbus_binary_common.nim
nitely: renamed to `RestApiConf`

beacon_chain/conf.nim
nitely: done

beacon_chain/conf.nim
nitely: done

tersec:
`nim-confutils` bump done externally to this PR.

Issue #7778, opened 2025-12-05
Broken config if BPO is scheduled to genesis.


barnabasbusa:
**Describe the bug**
Nimbus currently cannot handle BPO to be scheduled to genesis.

**To Reproduce**
```yaml
participants:
  - cl_type: lighthouse
    cl_image: sigp/lighthouse:v8.0.1
    el_type: geth
    el_image: ethereum/client-go:v1.16.7
  - cl_type: nimbus
    cl_image: statusim/nimbus-eth2:multiarch-v25.11.1
    el_type: geth
    el_image: ethereum/client-go:v1.16.7

network_params:
  fulu_fork_epoch: 0
  bpo_1_epoch: 0
```

`/eth/v1/config/spec` beacon api should return
```json
    "BLOB_SCHEDULE": [
      {
        "EPOCH": "0",
        "MAX_BLOBS_PER_BLOCK": "15"
      }
    ],
```

But instead, it returns:
```json
    "BLOB_SCHEDULE": []
```

tersec:
- https://github.com/status-im/nimbus-eth2/pull/7788

PR #7780, opened 2025-12-05
fix: bpo spec initialization for genesis

barnabasbusa:
Closes: https://github.com/status-im/nimbus-eth2/issues/7778

The code was checking if `currentBPO.EPOCH.uint64 != 0.uint64` before adding blob schedule entries.
This incorrectly filtered out epoch 0 (genesis), causing the API to return an empty `BLOB_SCHEDULE` array when BPO was scheduled for epoch 0. Probably the check was supposed to be for `MAX_BLOBS_PER_BLOCK` ?

This PR results in nimbus returning for BPO_1 scheduled to epoch 0:
```json
    "BLOB_SCHEDULE": [
      {
        "EPOCH": "0",
        "MAX_BLOBS_PER_BLOCK": "15"
      }
    ],
```


beacon_chain/spec/presets.nim
```diff
@@ -920,7 +920,7 @@ proc readRuntimeConfig*(
     if inBlobSchedule:
       let entry = strip(noComment, leading=true, trailing=false)
       if entry.startsWith("- EPOCH:"):
-        if currentBPO.EPOCH.uint64 != 0.uint64:
+        if currentBPO.MAX_BLOBS_PER_BLOCK.uint64 != 0.uint64:
```
agnxsh, line 923: > Probably the check was supposed to be for MAX_BLOBS_PER_BLOCK

imo it wasn't intended to use `MAX_BLOBS_PER_BLOCK`, it was kept in mind that the BPO epoch must not be 0/Genesis epoch

also you can see the `startsWith("- EPOCH:")` on 923, refactoring to check for `currentBPO.MAX_BLOBS_PER_BLOCK.uint64`, seems incorrect.

barnabasbusa: How can we start testing with a high BPO number on a new devnet then?

barnabasbusa:
Fixed in #7778

PR #7783, opened 2025-12-08
Integrate payload attestations into validator duties

Tomi-3-0:
- Fix: Sign PayloadAttestationData instead of full message per spec
- Integrate payload attestations into validator duty cycle

### DIFF
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -1087,9 +1087,11 @@ proc getDataColumnSidecarSZ*(db: BeaconChainDB, root: Eth2Digest,
     assign(dataPtr[], data)
   db.columns.get(columnkey(root, index), decode).expectDb()
 
-proc getDataColumnSidecar*(db: BeaconChainDB, root: Eth2Digest, index: ColumnIndex,
-                           value: var fulu.DataColumnSidecar): bool =
-  if db.columns == nil:  # Fulu has not been scheduled; DB table does not exist
+proc getDataColumnSidecar*[
+    T: fulu.DataColumnSidecar | gloas.DataColumnSidecar](
+    db: BeaconChainDB, root: Eth2Digest,
+    index: ColumnIndex, value: var T): bool =
+  if db.columns == nil:  # Gloas has not been scheduled; DB table does not exist
     return false
   db.columns.getSZSSZ(columnkey(root, index), value) == GetResult.found
 
--- a/beacon_chain/beacon_chain_db.nim
+++ b/beacon_chain/beacon_chain_db.nim
@@ -117,7 +117,7 @@ type
 
     blobs: KvStoreRef # (BlockRoot -> BlobSidecar)
 
-    columns: KvStoreRef # (BlockRoot -> DataColumnSidecar)
+    columns: array[ConsensusFork, KvStoreRef] # (BlockRoot -> DataColumnSidecar)
 
     stateRoots: KvStoreRef # (Slot, BlockRoot) -> StateRoot
 
@@ -602,9 +602,20 @@ proc new*(T: type BeaconChainDB,
 
   var blobs = kvStore db.openKvStore("deneb_blobs").expectDb()
 
-  var columns: KvStoreRef
-  if cfg.FULU_FORK_EPOCH != FAR_FUTURE_EPOCH:
-    columns = kvStore db.openKvStore("fulu_columns").expectDb()
+  var columns = [
+    nil, # Phase0
+    nil, # Altair
+    nil, # Bellatrix
+    nil, # Capella
+    nil, # Deneb
+    nil, # Electra
+    if cfg.FULU_FORK_EPOCH != FAR_FUTURE_EPOCH:
+      kvStore db.openKvStore("fulu_columns").expectDb()
+    else: nil,
+    if cfg.GLOAS_FORK_EPOCH != FAR_FUTURE_EPOCH:
+      kvStore db.openKvStore("gloas_columns").expectDb()
+    else: nil
+  ]
 
   var envelopes: KvStoreRef
   if cfg.GLOAS_FORK_EPOCH != FAR_FUTURE_EPOCH:
@@ -801,8 +812,9 @@ proc close*(db: BeaconChainDB) =
 
   # Close things roughly in reverse order
   db.quarantine.close()
-  if not isNil(db.columns):
-    discard db.columns.close()
+  for kv in db.columns:
+    if kv != nil:
+      discard kv.close()
   if not isNil(db.blobs):
     discard db.blobs.close()
   db.lcData.close()
@@ -860,18 +872,25 @@ proc delBlobSidecar*(
 proc putDataColumnSidecar*(
     db: BeaconChainDB,
     value: fulu.DataColumnSidecar) =
+  const consensusFork = ConsensusFork.Fulu
+  doAssert db.columns[consensusFork] != nil
   let block_root = hash_tree_root(value.signed_block_header.message)
-  db.columns.putSZSSZ(columnkey(block_root, value.index), value)
+  db.columns[consensusFork].putSZSSZ(columnkey(block_root, value.index), value)
 
 proc putDataColumnSidecar*(
     db: BeaconChainDB,
     value: gloas.DataColumnSidecar) =
-  db.columns.putSZSSZ(columnkey(value.beacon_block_root, value.index), value)
+  const consensusFork = ConsensusFork.Gloas
+  doAssert db.columns[consensusFork] != nil
+  db.columns[consensusFork].putSZSSZ(columnkey(
+    value.beacon_block_root, value.index), value)
 
 proc delDataColumnSidecar*(
-    db: BeaconChainDB,
+    db: BeaconChainDB, consensusFork: ConsensusFork,
     root: Eth2Digest, index: ColumnIndex): bool =
-  db.columns.del(columnkey(root, index)).expectDb()
+  if db.columns[consensusFork] == nil:
+    return false
+  db.columns[consensusFork].del(columnkey(root, index)).expectDb()
 
 proc putExecutionPayloadEnvelope*(
     db: BeaconChainDB, value: SignedExecutionPayloadEnvelope) =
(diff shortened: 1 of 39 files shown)


beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,118 @@ proc sendSyncCommitteeContributions(
...
+          beacon_block_root, envelope):
+        return false
+
+      template commitments: untyped = envelope.message.blob_kzg_commitments
+      if commitments.len == 0:
+        return true # No blobs required for this slot
+
+      for i in 0..<commitments.len:
+        var blob: BlobSidecar
+        if not node.dag.db.getBlobSidecar(beacon_block_root, BlobIndex(i), blob):
```
tersec: It would be better to just check for the existence of the key, less database I/O and less memory usage, similar to how
https://github.com/status-im/nimbus-eth2/blob/b186c999147d48ba8c967576d336a1fe2624f348/beacon_chain/beacon_chain_db.nim#L1277-L1323
work, i.e. a `contains` function in the database layer.

These also end up being conveniently testable insofar as there are tests for `put`/`get`/`del`.

That said, I'm not sure checking for blobs in the database at all is the right approach. For example, blobs can't in general be reconstructed unless one has at least half the columns, but nodes which have fewer than half the columns are also expected to provide payload attestations, because neither https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/validator.md#payload-timeliness-committee nor https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/validator.md#constructing-a-payload-attestation specifies anything about the PTC needing to be made entirely of supernodes or light supernodes.

So I don't know if `checkBlobDataAvailable` is the right function to have, here.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,118 @@ proc sendSyncCommitteeContributions(
...
+  let blokData = node.dag.getForkedBlock(beacon_block_root).valueOr:
+    return false
+
+  withBlck(blokData):
+    when consensusFork >= ConsensusFork.Gloas:
+      var envelope: TrustedSignedExecutionPayloadEnvelope
+      node.dag.db.getExecutionPayloadEnvelope(beacon_block_root, envelope) and
+        envelope.message.beacon_block_root == beacon_block_root
+    else:
+      return true
```
tersec: Can just be `true` (no `return`).

beacon_chain/validators/beacon_validators.nim
Tomi-3-0: Exactly. I agree

beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,118 @@ proc sendSyncCommitteeContributions(
...
+      for i in 0..<commitments.len:
+        var blob: BlobSidecar
+        if not node.dag.db.getBlobSidecar(beacon_block_root, BlobIndex(i), blob):
+          return false
+      return true
+    else:
+      return true
+
+proc createAndSendPayloadAttestation(node: BeaconNode,
+                                   fork: Fork,
```
tersec: If using this indentation style, `fork:`, `genesis_validators_root:`, etc should align with `node: `, i.e. 2 more spaces indentation.

tersec:
```
[2025-12-09T00:14:52.688Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7783/beacon_chain/rpc/rest_debug_api.nim(31, 7) Error: type mismatch: got 'None' for 'getDataColumnSidecar' but expected 'proc'
[2025-12-09T00:14:52.688Z] make: *** [Makefile:453: nimbus_beacon_node] Error 1
```

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7783/commits/7779b07af97d0863557e5e4cfc74b83abd34580a

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7783/commits/7779b07af97d0863557e5e4cfc74b83abd34580a

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7783/commits/7779b07af97d0863557e5e4cfc74b83abd34580a changes it to custody columns, not blobs, consistent with current spec interpretation.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,119 @@ proc sendSyncCommitteeContributions(
...
+    node: BeaconNode, beacon_block_root: Eth2Digest): bool = 
+  ## Check if the blob sidecars for this slot and envelope are available
+
+  let blckData = node.dag.getForkedBlock(beacon_block_root).valueOr:
+    return false
+
+  withBlck(blckData):
+    when consensusFork >= ConsensusFork.Gloas:
+      var envelope: TrustedSignedExecutionPayloadEnvelope
+      if not node.dag.db.getExecutionPayloadEnvelope(
```
tersec: The actual `envelope` here isn't used, so retrieving it from the database, allocating and copying it around, and returning it is pointless. Better to have a `containsExecutionPayloadEnvelope` function in `beacon_chain_db` and query that.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,119 @@ proc sendSyncCommitteeContributions(
...
+        return false
+
+      template commitments: untyped = envelope.message.blob_kzg_commitments
+      if commitments.len == 0:
+        return true # No blobs required for this slot
+
+      # check that our custody columns are available
+      for columnIdx in node.dataColumnQuarantine.custodyColumns:
+        var column: gloas.DataColumnSidecar
+        if not node.dag.db.getDataColumnSidecar(
```
tersec: The `column` here isn't used, so retrieving it from the database, allocating and copying it around, and returning it is pointless. Better to have a `containsDataColumnSidecar` function in `beacon_chain_db` and query that.

beacon_chain/validators/beacon_validators.nim
Tomi-3-0: I've removed the envelope check entirely and just check custody columns directly.
eliminating the envelope deserialization completely. The trade-off is checking custody columns even when there are no blobs, but since those are just fast SQL key lookups versus the expensive deserialisation

beacon_chain/beacon_chain_db.nim
```diff
@@ -1322,6 +1324,18 @@ proc containsState*(db: BeaconChainDB, key: Eth2Digest, legacy: bool = true): bo
 
   (legacy and db.v0.containsState(key))
 
+proc containsExecutionPayloadEnvelope*(
+    db: BeaconChainDB, root: Eth2Digest): bool =
+  if db.envelopes == nil:
+    return false
+  db.envelopes.contains(root.data).expectDb()
+
+proc containsDataColumnSidecar*(
```
tersec, line 1376: Needs to be tested as part of https://github.com/status-im/nimbus-eth2/blob/8078bf01725163abb5f8e7f01d002d41f2845d61/tests/test_beacon_chain_db.nim#L464-L560

i.e. check invariants: before being `put`, `contains` should be false`. after being `put`, it should be `true`, after being deleted, it should be `false`.

beacon_chain/beacon_chain_db.nim
```diff
@@ -1322,6 +1324,18 @@ proc containsState*(db: BeaconChainDB, key: Eth2Digest, legacy: bool = true): bo
 
   (legacy and db.v0.containsState(key))
 
+proc containsExecutionPayloadEnvelope*(
```
tersec, line 1370: There don't appear to be any database tests at all for execution payload envelopes. Those should be added, with the same basic flow tested as for other database-stored objects.

It's a bigger change so it doesn't have to be part of this PR, to keep this PR cleaner. But when it's added, it too should test `containsExecutionPayloadEnvelope` like it tests the rest.

tersec: https://github.com/status-im/nimbus-eth2/pull/7783/commits/0480608366d3b08afdf53a5b2a3613681c51a30b

beacon_chain/beacon_chain_db.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7783/commits/0480608366d3b08afdf53a5b2a3613681c51a30b

beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,102 @@ proc sendSyncCommitteeContributions(
...
+  withState(node.dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      var cache: StateCache
+      for vidx in get_ptc(forkyState.data, slot, cache):
+        let validator = node.getValidatorForDuties(vidx, slot).valueOr:
+          continue
+
+        asyncSpawn createAndSendPayloadAttestation(
+          node, fork, genesis_validators_root, validator, vidx, slot,
+          target.blck.root )
```
tersec: Extra space before `)`?

beacon_chain/validators/beacon_validators.nim
```diff
@@ -838,6 +838,102 @@ proc sendSyncCommitteeContributions(
...
+      attestationTarget = shortLog(target),
+      head = shortLog(head)
+  
+  let
+    fork = node.dag.forkAtEpoch(slot.epoch)
+    genesis_validators_root = node.dag.genesis_validators_root
+    
+  withState(node.dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      var cache: StateCache
```
tersec, line 928: Using an empty `StateCache` requires recomputing the validator shufflings from scratch, which requires a significant fraction of a second with 2M+ validators (not only active validators, all validators).

https://github.com/status-im/nimbus-eth2/blob/1d32fc547a8cf0bb01c98ea9bca1a94aca447386/beacon_chain/validators/beacon_validators.nim#L1285-L1307 has the same issue.

The reason for this in the codebase is `get_ptc()`, which per https://github.com/status-im/nimbus-eth2/blob/1d32fc547a8cf0bb01c98ea9bca1a94aca447386/beacon_chain/spec/beaconstate.nim#L2765-L2787 needs it for `get_committee_count_per_slot()` and `get_beacon_committee()`.

To avoid this reshuffling, https://github.com/status-im/nimbus-eth2/blob/1d32fc547a8cf0bb01c98ea9bca1a94aca447386/beacon_chain/consensus_object_pools/spec_cache.nim#L30-L32 and https://github.com/status-im/nimbus-eth2/blob/1d32fc547a8cf0bb01c98ea9bca1a94aca447386/beacon_chain/consensus_object_pools/spec_cache.nim#L43-L67 allow both `get_committee_count_per_slot()` and `get_beacon_committee()` to be done with `getEpochRef()`, which provides the necessary `ShufflingRef`.

For an example of how this works, see https://github.com/status-im/nimbus-eth2/blob/1d32fc547a8cf0bb01c98ea9bca1a94aca447386/beacon_chain/validators/beacon_validators.nim#L678-L684

So there should be a version (also tested) of `get_ptc()` in https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/consensus_object_pools/spec_cache.nim which can be used with a `ShufflingRef` rather than a `StateCache`, because the `ShufflingRef` will already exist. It's likely the existing version of `get_ptc()` will continue to need to exist too, as long as there's a use case for the `StateCache` version.

tersec: https://github.com/status-im/nimbus-eth2/pull/7783/changes/c9bc364084600298ff2ab03c42d03bba18e88e09

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7783/commits/907edfc8473b01df8b469de4c03745a5e00ab5c3

PR #7788, opened 2025-12-09
fix blob schedule parsing at genesis

agnxsh:
succeeds #7780

### DIFF
--- a/beacon_chain/spec/presets.nim
+++ b/beacon_chain/spec/presets.nim
@@ -900,6 +900,7 @@ proc readRuntimeConfig*(
     blobScheduleEntries: seq[BlobParameters]
     inBlobSchedule = false
     currentBPO: BlobParameters
+    hasCurrentEntry = false
 
   for rawLine in splitLines(fileContent):
     inc lineNum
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/presets.nim
```diff
@@ -954,12 +957,15 @@ proc readRuntimeConfig*(
...
 
   # Final BLOB_SCHEDULE entry
-  if inBlobSchedule and currentBPO.EPOCH.uint64 != 0.uint64:
+  if inBlobSchedule and hasCurrentEntry:
     blobScheduleEntries.add(currentBPO)
 
   # BPO entries must be sorted in reverse epoch order
   blobScheduleEntries.sort(cmp = cmpBlobParameters)
 
+  # BPO entries must be sorted in reverse epoch order
```
tersec: Why is this duplicated?

agnxsh: https://github.com/status-im/nimbus-eth2/pull/7788/commits/3a626c5eb126c702947bec70b5fde848139f7b69

PR #7791, opened 2025-12-10
libbacktrace: bump

arnetheduck:
* fix memory leak in stack collector
* nicer demangled stack traces


tersec:
In [CI / Developer builds (pull_request)](https://github.com/status-im/nimbus-eth2/actions/runs/20109080575/job/57701150784?pr=7791):
```
/bin/sh: 1: g++: Permission denied
```

PR #7800, opened 2025-12-15
handle deadlocks/blocked tasks in parallel reconstruction

agnxsh:


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1771,9 +1771,9 @@ proc reconstructDataColumns(node: BeaconNode, slot: Slot) =
       node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2:
     return
 
-  # Currently, this logic is broken
-  if true:
-    return
+  # # Currently, this logic is broken
+  # if true:
```
tersec: Can just remove this `if` entirely, rather than commenting it out.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +154,170 @@ proc recover_cells_and_proofs_parallel*(
...
+      localIndices = newSeq[CellIndex](columnCount)
+      localCells = newSeq[Cell](columnCount)
+    let
+      idxArr = cast[ptr UncheckedArray[CellIndex]](idxPtr)
+      cellsArr = cast[ptr UncheckedArray[Cell]](cellsPtr)
+    for j in 0 ..< columnCount:
+      localIndices[j] = idxArr[j]
+      localCells[j] = cellsArr[j]
+    # use the task wrapper which maps string errors to void
+    recoverCellsAndKzgProofsTask(localIndices, localCells)
```
arnetheduck: there is no need for this seq as far as I can tell - just change the argument type of ...Task to `openArray`, then use `idxArr.toOpenArray(0, columnCount - 1)`

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +154,170 @@ proc recover_cells_and_proofs_parallel*(
...
+      return err("Data column reconstruction timed out")
 
-    var
-      cellIndices = newSeq[CellIndex](columnCount)
-      cells = newSeq[Cell](columnCount)
+    # Allocate unmanaged C buffers and copy data into them
+    let
+      idxBytes = csize_t(columnCount) * csize_t(sizeof(CellIndex))
+      cellsBytes = csize_t(columnCount) * csize_t(sizeof(Cell))
+      idxPtr = cast[ptr CellIndex](c_malloc(idxBytes))
```
arnetheduck: there is no need to use `malloc` here - you can use a seq as long as you pass `addr myseq[0]` to the task since both Cell and CellIndex are simple types. 

The important rule is that no garbage-collection-related operations must happen cross-thread.. for example, you must not allocate in one thread and release in the other. But if you pass a pointer to something, both allocation and deallocation happens in the same thread and the other thread is just using the allocated space without considering when it will be released.

This also means that the calling thread must guarantee that the memory does not get deallocated until the task has finished.

So the pattern is more or less:

```nim
proc work(src: openArray[SimpleType], tgt: var openArray[SimpleType]) = ...
proc worker(src: ptr SimpleType, srclen: int, tgt: ptr ..., sig: ThreadsignalPtr) =
  work(makeUncheckedArray(src).toOpenArray(0, srclen-1), tgt ...)
  # let the other thread know we're done and will not be needing src/tgt any more
  sig.notify()

proc workAsync(src: seq[SimpleType]) {.async.} =
  # Normally should not be allocated for each task, but only one async function should notify/wait or more synchronization is needed
  let sig = ThreadSignalPtr.new()
  # Allocate memory for outcome
  var tgt = newSeq[SimpleType](src.len)

  tp.spawn(worker(addr src[0],src.len, addr tgt[0], tgt.len)

  # Wait for task to finish
  await sig.wait()

  sig.close()

  # tgt now contains outcome and it's safe to "free" src and tgt

```

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +153,112 @@ proc recover_cells_and_proofs_parallel*(
...
+      idxArr = cast[ptr UncheckedArray[CellIndex]](idxPtr)
+      cellsArr = cast[ptr UncheckedArray[Cell]](cellsPtr)
+    # Use toOpenArray to create views without copying
+    recoverCellsAndKzgProofsTask(
+      idxArr.toOpenArray(0, columnCount - 1),
+      cellsArr.toOpenArray(0, columnCount - 1))
+
   var
-    pendingFuts: seq[Flowvar[Result[CellsAndProofs, void]]]
+    pendingFuts: seq[Flowvar[Result[CellsAndProofs, void]]] = @[]
```
tersec: The ` = @[]` doesn't do anything here.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +153,112 @@ proc recover_cells_and_proofs_parallel*(
...
+    # Store actual data sequences instead of C pointers
+    pendingIndices: seq[seq[CellIndex]] = @[]
+    pendingCells: seq[seq[Cell]] = @[]
```
tersec: Likewise either of these places. https://nim-lang.org/docs/manual.html#statements-and-expressions-var-statement
> Variables are always initialized with a default value if there is no initializing expression. The default
value depends on the type and is always a zero in binary.

where for a sequence, this default is documented to be `@[]`.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +153,112 @@ proc recover_cells_and_proofs_parallel*(
...
     res = newSeq[CellsAndProofs](blobCount)
 
+  # pre-size sequences so we can index-assign without reallocs
+  pendingFuts.setLen(blobCount)
+  pendingIndices.setLen(blobCount)
+  pendingCells.setLen(blobCount)
```
tersec: If these are going to be `setLen`'d immediately, why not `newSeq[Flowvar[Result[CellsAndProofs, void]]](blobCount)` (respectively, each `seq` type) them to begin with?

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +153,112 @@ proc recover_cells_and_proofs_parallel*(
...
 
+    # Use regular Nim sequences
     var
-      cellIndices = newSeq[CellIndex](columnCount)
+      indices = newSeq[CellIndex](columnCount)
       cells = newSeq[Cell](columnCount)
+    
     for i in 0 ..< dataColumns.len:
-      cellIndices[i] = dataColumns[i][].index
+      indices[i] = dataColumns[i][].index
```
tersec: The redundant/indirect assignment/copying via `indices` and `cells` is fairly expensive. Can directly ensure that `pendingIndices[spawned]` and `pendingCells[spawned]` are the right side, the have this loop assign into `pendingIndices[spawned]` item by item directly (respectively, `pendingCells[spawned]`)

Can also use `addr pendingIndices[spawned]`/`addr pendingCells[spawned]` can avoid `dataColumns.len` lookups/bound checks/etc and allow only dereferencing two pointers each loop iteration to do so, basically caching the address of each `seq` item.

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7800/commits/e5f124304dc90198a954c76d3dc8837a61ee0ab7

beacon_chain/spec/peerdas_helpers.nim
agnxsh: uh yeah

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -153,58 +153,112 @@ proc recover_cells_and_proofs_parallel*(
...
+  # ---- Spawn + bounded-await loop ----
   for blobIdx in 0 ..< blobCount:
     let now = Moment.now()
     if (now - startTime) > reconstructionTimeout:
-      debug "PeerDAS reconstruction timed out while preparing columns",
-        spawned = pendingFuts.len, total = blobCount
-      break  # Stop spawning new tasks
+      trace "PeerDAS reconstruction timed out while preparing columns",
+        spawned = spawned, total = blobCount
+      return err("Data column reconstruction timed out")
```
arnetheduck: what happens to the tasks that are still pending here? presumably, they still reference the memory allocated by `pendingIndices` and `pendingCells`?

agnxsh: in that case, we need to drain all the pending tasks, one more thing we can probably do instead of an early return is to use a timeout flag, and check if we have timed out before spawning newer tasks, that way the number of pending tasks _may_ get lower, however, timeout won't be strict in that sense, as we would wait to drain the pending tasks, but should be memory safe, most likely.

(i have made some more changes)

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7800/changes/1781d9f1201ef5c9e95be7aa6051dd98fa91c35a

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7800/changes/1781d9f1201ef5c9e95be7aa6051dd98fa91c35a

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7800/changes/1781d9f1201ef5c9e95be7aa6051dd98fa91c35a

tersec:
```
- beacon_chain/nimbus_beacon_node.nim
- beacon_chain/spec/peerdas_helpers.nim
```

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7800/changes/11a9879516d1866c7baf884d1d703ab0574b628d

beacon_chain/conf.nim
```diff
@@ -231,6 +231,11 @@ type
       desc: "Subscribe to the first half of column subnets"
       name: "light-supernode" .}: bool
 
+    debugDisableReconstruction* {.
```
arnetheduck: let's avoid the double negative here .. ie `--debug-enable-reconstruction` instead

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1777,7 +1777,7 @@ proc reconstructDataColumns(node: BeaconNode, slot: Slot) =
     return
 
   # Currently, this logic is broken
-  if true:
+  if node.config.debugEnableReconstruction:
```
tersec: Inverted check condition.

PR #7801, opened 2025-12-15
Only check isomorphism of types in debug compiles

arnetheduck:
The VM is slow at instantiating large types like the beacon state at compile time, hence it takes >10s to perform this check.

A macro similar to nim-serialization could be developed instead to replace `fieldPairs` which here is only used to reason about type, but still needs an actual instance.

### DIFF
--- a/beacon_chain/libnimbus_lc/libnimbus_lc.nim
+++ b/beacon_chain/libnimbus_lc/libnimbus_lc.nim
@@ -9,6 +9,7 @@
 
 import
   std/[json, sequtils, times],
+  stew/objects,
   eth/common/eth_types_rlp,
   eth/common/keys,
   eth/p2p/discoveryv5/random2,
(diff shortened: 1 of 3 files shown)


etan-status:
How did you profile this? Could be a useful technique for others as well.

arnetheduck:
* `nim c --profileVM`
* compile `nim` with `--debugger:native` then profile compilation with `vtune` - there, you see that zeroing is the major source of bad perf, which is true also at runtime - then look for sources of lots of zeroing, default being one, then reverse-engineer from there.
* compile with `--processing:filenames` and just look at which modules take time.

The problem with this code in particular is that the VM has to actually evaluate `default` to call `getSizeofSig` - it cannot know that the latter works with types only.

It's a problem of `fieldPairs` in general, that require an actual instance - the alternative is to do what nim-serialization does and work with type information in a macro instead, that would have been the proper fix here.

PR #7803, opened 2025-12-15
Use `debug` instead of `hidden` in debug conf options

nitely:
Changes:

- Change `debug-` prefixed conf options from `hidden` to `debug`. They will show when setting `--help:debug`

Note there are more `hidden` options. I only changed the ones that have `debug-` prefix.

Bump confutils:

- https://github.com/status-im/nim-confutils/pull/126
- https://github.com/status-im/nim-confutils/pull/127
- https://github.com/status-im/nim-confutils/pull/128

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -242,7 +242,7 @@ type
       name: "subscribe-all-subnets" .}: bool
 
     debugPeerdasSupernode* {.
-      hidden
+      debug
       defaultValue: false,
       name: "debug-peerdas-supernode" .}: bool
 
@@ -397,13 +397,13 @@ type
         name: "verify-finalization" .}: bool
 
       stopAtEpoch* {.
-        hidden
+        debug
         desc: "The wall-time epoch at which to exit the program. (for testing purposes)"
         defaultValue: 0
         name: "debug-stop-at-epoch" .}: uint64
 
       stopAtSyncedEpoch* {.
-        hidden
+        debug
         desc: "The synced epoch at which to exit the program. (for testing purposes)"
         defaultValue: 0
         name: "stop-at-synced-epoch" .}: uint64
@@ -557,7 +557,7 @@ type
         name: "light-client-data-max-periods" .}: Option[uint64]
 
       longRangeSync* {.
-        hidden
+        debug
         desc: "Enable long-range syncing (genesis sync)",
         defaultValue: LongRangeSyncMode.Lenient,
         name: "debug-long-range-sync".}: LongRangeSyncMode
@@ -601,7 +601,7 @@ type
       #     `--debug-invalidate-block-root` flag present until finality has
       #     advanced beyond the problematic chain segment
       invalidBlockRoots* {.
-        hidden
+        debug
         desc: "List of beacon block roots that, if the EL responds with SYNCING/ACCEPTED, are treated as if their execution payload was INVALID"
         name: "debug-invalidate-block-root" .}: seq[Eth2Digest]
 
@@ -685,14 +685,14 @@ type
         name: "history".}: HistoryMode
 
       trustedSetupFile* {.
-        hidden
+        debug
         desc: "Alternative EIP-4844 trusted setup file"
         defaultValue: none(string)
         defaultValueDesc: "Baked in trusted setup"
         name: "debug-trusted-setup-file" .}: Option[string]
 
       bandwidthEstimate* {.
-        hidden
+        debug
         desc: "Bandwidth estimate for the node (bits per second)"
         name: "debug-bandwidth-estimate" .}: Option[Natural]
 
@@ -1035,6 +1035,7 @@ type
       name: "graffiti" .}: Option[GraffitiBytes]
 
     stopAtEpoch* {.
+      debug
       desc: "A positive epoch selects the epoch at which to stop"
       defaultValue: 0
       name: "debug-stop-at-epoch" .}: uint64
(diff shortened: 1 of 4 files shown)


beacon_chain/conf.nim
```diff
@@ -1035,7 +1035,6 @@ type
       name: "graffiti" .}: Option[GraffitiBytes]
 
     stopAtEpoch* {.
-      debug
       desc: "A positive epoch selects the epoch at which to stop"
       defaultValue: 0
       name: "debug-stop-at-epoch" .}: uint64
```
nitely, line 1040: should this contain the debug pragma?
note it's not hidden.

beacon_chain/conf.nim
```diff
@@ -397,7 +397,7 @@ type
         name: "verify-finalization" .}: bool
 
       stopAtEpoch* {.
-        hidden
+        debug
         desc: "The wall-time epoch at which to exit the program. (for testing purposes)"
         defaultValue: 0
         name: "debug-stop-at-epoch" .}: uint64
```
nitely, line 403: There is a hidden `stopAtSyncedEpoch` below this option. Should that be switched to debug? (also adding debug- prefix).

tersec:
Generally speaking, `hidden` is intentional in many-to-all of those cases; they're not meant to be documented at all.

And, no, `stopAtSyncedEpoch` _definitely_ isn't meant to be documented. If people want to wander through source code, that's up to them, but it's not in any way a supported or generally applicable option. It's a specific artifact of running CI and CI-like setups.

beacon_chain/conf.nim
tersec: It's a mistaken/an oversight it's not `hidden`. But the documentation already says it's not supported, effectively, so the better thing to do is mark it `hidden`.

beacon_chain/conf.nim
```diff
@@ -685,14 +685,14 @@ type
         name: "history".}: HistoryMode
 
       trustedSetupFile* {.
-        hidden
+        debug
```
tersec, line 688: This is/was something to help Ethpandaops run their devnets. That's ... I think? the only use it ever got, and better that it would not be visible at all.

beacon_chain/conf.nim
```diff
@@ -557,7 +557,7 @@ type
         name: "light-client-data-max-periods" .}: Option[uint64]
 
       longRangeSync* {.
-        hidden
+        debug
```
tersec, line 560: This isn't really a "debugging" option, it's a feature that we never ended up enforcing, for various reasons.

etc etc etc

All of these have similar stories. `hidden`, not `debug` insofar as it would make them visible with a `--help:debug` option.

beacon_chain/conf.nim
```diff
@@ -242,7 +242,7 @@ type
       name: "subscribe-all-subnets" .}: bool
 
     debugPeerdasSupernode* {.
-      hidden
+      debug
```
tersec, line 245: This one is going away soon anyway, as announced in a recent release notes. It's `--debug` and that's what `--debug-foo` options do.

So definitely not something to add visibility to now.

arnetheduck:
> intentional

the idea behind `debug` is that it's also hidden by default, but for debugging, testing and similar special cases they can be brought forward - "things that pandaops wants to try" qualify for this, for example, or things we might suggest to users to get out of a bind

it's important that all these options can be discovered using `--help` _somehow_ - ie in order to understand their syntax _when_ using them. Their main difference is that they're not part of the stable flag set

tersec:
Special cases become "supported" cases.

beacon_chain/conf.nim
tersec: Also this one has been replaced by a supported option, anyway. Turned out there was nontrivial demand due to the blob use cases. So especially useless as a "debug" option.

arnetheduck:
> Special cases become "supported" cases.

For the cases where this happens, they are migrated to a supported case, with the old option left with a deprecation warning allowing users to migrate too, without breaking their setups unnecessarily - the key is to manage the migration path so that those that use these options understand what's happening.

beacon_chain/conf.nim
tersec: Also regarding this one, I'm not sure it's been used since the Dencun devnets, or has any remaining usecases now that the KZG setup is confirmed/set.

For theoretical-other-KZG setups, they'd be coming with other changes realistically anyway. A parameter to tweak this specifically is likely obsolete even for testing.

beacon_chain/conf.nim
arnetheduck: again, the key is to set expectations and document the change: 'In release XYZ, --debug-frob has migrated to --frob - the old name will be removed in the 25.x release'

tersec:
My actual tendencies for these:
- I'd check with pandaops if there's any actual usecase left, and if not, remove even `trustedSetupFile` configuration
- `longRangeSync` was a half-baked artifact of a sync method which never quite made it, remove option (i.e. allow long-range sync, as its default value is already "Lenient")
- keep various `debug-stop-at-epoch`s hidden. They're very CI-focused; they're not actually what pandaops and other testing tends to use, because they usually want tests which run indefinitely
- for the debug peerdas supernode setting, per https://github.com/status-im/nimbus-eth2/releases/tag/v25.11.1 remove it, but give it another release or so

beacon_chain/conf.nim
tersec: See elsewhere. The point is for this PR that it's a useless thing to expose.

PR #7822, opened 2025-12-24
Refactor database schema to seperate fulu and gloas data column tables

Tomi-3-0:


tersec (approved):

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -158,9 +157,14 @@ proc handleDataSidecarRequest*[
   for dataIndex in 0'u64 ..< maxDataSidecars:
     if indexFilter.len > 0 and dataIndex notin indexFilter:
       continue
-    let dataSidecar = new DataSidecarsType.T
-    if getDataSidecar(node.dag.db, bid.root, dataIndex, dataSidecar[]):
-      discard data[].add dataSidecar[]
+    var dataSidecar: DataSidecarsType.T
+    when DataSidecarsType.T is BlobSidecar:
```
agnxsh: i think we can build a better layer of abstraction here for sidecars, maybe using something like `Forky`, what do you recommend @tersec ? (i have accidentally started a review, it should just be a comment)

agnxsh: or maybe along with an overloaded `getDataSidecar`

agnxsh: or an adapter reader function for sidecars where we can pass the `ConsensusFork`, the idea is to look at sidecars as general sidecars from a given module view, ofc this change in other places will come in gradually, but it's just a great to have for changes that are landing fresh.

we can anticipate something related to cell sidecars, in that case, this place would look like an if/else/if/... trail

agnxsh (approved):

beacon_chain/beacon_chain_db.nim
```diff
@@ -602,9 +602,20 @@ proc new*(T: type BeaconChainDB,
...
+    nil, # Altair
+    nil, # Bellatrix
+    nil, # Capella
+    nil, # Deneb
+    nil, # Electra
+    if cfg.FULU_FORK_EPOCH != FAR_FUTURE_EPOCH:
+      kvStore db.openKvStore("fulu_columns").expectDb()
+    else: nil,
+    if cfg.GLOAS_FORK_EPOCH != FAR_FUTURE_EPOCH:
+      kvStore db.openKvStore("gloas_columns").expectDb()
```
tersec, line 616: As a general point, this is fine for now, but assuming, as is likely, that H* will also use `gloas` columns, that
- the naming scheme is reasonable (sure, it's just the fork in which the columns were introduced; it does mean that multiple forks' columns might map to a single table, but that's fine, so long as they're the same format); and
- the initialization can be set up not to repeat the `kvStore db.openKvStore("gloas_columns").expectDb()` (also, fairly straightforward)

But both of these are compatible with the setup presented here and are a small refactoring away.

Tomi-3-0: That's a good point.  I could easily just make that refactoring now

Tomi-3-0: Okay. Yes. when H comes we can refactor to avoid opening the table multiple times:
```
if cfg.GLOAS_FORK_EPOCH != FAR_FUTURE_EPOCH or 
   cfg.H_FORK_EPOCH != FAR_FUTURE_EPOCH:
  kvStore db.openKvStore("gloas_columns").expectDb()
```

tersec: Indeed, though it can be simpler than that even: because Nimbus will check that `cfg.H_FORK_EPOCH >= cfg.GLOAS_FORK_EPOCH` just as it check that, say, `cfg.GLOAS_FORK_EPOCH >= cfg.FULU_FORK_EPOCH`, only have to check that `cfg.GLOAS_FORK_EPOCH != FAR_FUTURE_EPOCH` no matter how many forks using Gloas columns end up existing.

beacon_chain/beacon_chain_db.nim
```diff
@@ -1322,6 +1364,19 @@ proc containsState*(db: BeaconChainDB, key: Eth2Digest, legacy: bool = true): bo
...
   (legacy and db.v0.containsState(key))
 
+proc containsExecutionPayloadEnvelope*(
+    db: BeaconChainDB, root: Eth2Digest): bool =
+  if db.envelopes == nil:
+    return false
+  db.envelopes.contains(root.data).expectDb()
+
+proc containsDataColumnSidecar*(
+    db: BeaconChainDB, consensusFork: ConsensusFork,
```
tersec: Is this `consensusFork` ever called in a position where it's not known at compiletime? In general the code operating on this operates where it already knows the fork it's in.

beacon_chain/rpc/rest_beacon_api.nim
tersec: > i think we can build a better layer of abstraction here for sidecars, maybe using something like `Forky`, what do you recommend @tersec ? (i have accidentally started a review, it should just be a comment)

Well `Forky` is used when there's some unification of code that can occur, some commonality across possible sidecar types.

For example, overloading `getDataSidecar` might not be helped by `Forky` because in one case it needs to call `getBlobSidecar` and in the other case `getDataColumnSidecar`.

There is an argument to unify things further, but at some point the beacon chain db code is written in a concrete way which needs to resolve into a specific thing to fetch via the `root, index, sidecar` parameters.

One possible API design is, yeah, to have basically one `getSidecar` and one `putSidecar` function each which depending on whether its `var` parameter is a blob, fulu column, gloas column, or whatever future sidecar type (e.g. cell) it gets the right thing from the database without calling through a more concrete function. The issue becomes consistency with the `del` and `contains` parts: those don't have a "natural" parameter, either input (no `var`) or output (`var`) to derive the type from if not otherwise specified, e.g., if they only have a block root, or block root and index.

So to keep API consistency with `del`/`contains`, it's arguably worth keeping separate concrete functions for different sidecar types.

beacon_chain/beacon_chain_db.nim
```diff
@@ -1080,18 +1099,41 @@ proc getBlobSidecar*(db: BeaconChainDB, root: Eth2Digest, index: BlobIndex,
...
+  if db.columns[consensusFork] == nil:
     return false
-  db.columns.getSZSSZ(columnkey(root, index), value) == GetResult.found
+  db.columns[consensusFork].getSZSSZ(columnkey(root, index), value) ==
+    GetResult.found
+
+proc getSidecar*(
+    db: BeaconChainDB, root: Eth2Digest,
+    index: auto, sidecar: var auto): bool =
+  when typeof(sidecar) is BlobSidecar:
```
tersec: Rather than use the `when` construction, can split this into two functions and overload on types:
```nim
proc getSidecar*(
    db: BeaconChainDB, root: Eth2Digest,
    index: auto, sidecar: var BlobSidecar): bool =
  db.getBlobSidecar(root, index, sidecar)

proc getSidecar*(
    db: BeaconChainDB, root: Eth2Digest,
    index: auto,
    sidecar: var (fulu.DataColumnSidecar | gloas.DataColumnSidecar)): bool =
  db.getDataColumnSidecar(root, index, sidecar)
```

In particular, this lets Nim's overload matching machinery work automatically rather than recreating overload matching manually via `when/else` and tends to improve error messages and other diagnostics as a result, as well as automatically requiring a specific match, rather than otherwise trying `else` (e.g., if someone tries to call this `getSidecar` with anything besides a `var BlobSidecar`, it will return a misleading error message about how that doesn't match any `getDataColumnSidecar` overload, and of course it doesn't, but it's not the root cause, because the `when`/`else` is hiding half the overload resolution from Nim).

beacon_chain/rpc/rest_beacon_api.nim
tersec: > or an adapter reader function for sidecars where we can pass the `ConsensusFork`, the idea is to look at sidecars as general sidecars from a given module view, ofc this change in other places will come in gradually, but it's just a great to have for changes that are landing fresh.
> 
> we can anticipate something related to cell sidecars, in that case, this place would look like an if/else/if/... trail

Well, beacon chain DB is ultimately where they have to be routed to specific tables in the SQLite. This translation happens somewhere, and it makes sense that it's in this module. What API the beacon chain DB module presents is a different question.

Separately, likely rather than a `if`/`else` (or `when`/`else`) this can be done more like https://github.com/status-im/nimbus-eth2/pull/7822/changes#r2647572245 by using Nim's overload resolution to match types, with better ergonomics and generally better legibility to "go to definition"-like tools.

beacon_chain/beacon_chain_db.nim
Tomi-3-0: No, actually. It will be used in payload attestation code, which already operates within a when `consensusFork >= ConsensusFork.Gloas` context.  I can hardcode `ConsensusFork.Gloas` in the function, since it's still needed to index into the appropriate columns table

tersec: Another option is to make the parameter `consensusFork: static ConsensusFork` to avoid misuse.

Tomi-3-0: This makes more sense, keeps it flexible enough

beacon_chain/beacon_chain_db.nim
```diff
@@ -1322,6 +1367,19 @@ proc containsState*(db: BeaconChainDB, key: Eth2Digest, legacy: bool = true): bo
 
   (legacy and db.v0.containsState(key))
 
+proc containsExecutionPayloadEnvelope*(
+    db: BeaconChainDB, root: Eth2Digest): bool =
+  if db.envelopes == nil:
+    return false
+  db.envelopes.contains(root.data).expectDb()
```
arnetheduck, line 1374: why are envelopes not segregated by fork?

PR #7828, opened 2026-01-03
update copyright years for `beacon_chain/*`

agnxsh:
It is highly likely that most files under the `beacon_chain` folder shall be edited in 2026, which will in future trigger the linter.

Might be reasonable to martyr a few files that we won't touch in beacon chain for 2026.


agnxsh:
if we want to have an exclusion list of files, which via probability or deterministically _won't_ be edited this year, feel free to comment them down.

tersec:
> if we want to have an exclusion list of files, which via probability or deterministically _won't_ be edited this year, feel free to comment them down.

Seems reasonable to update the years on all of these.

PR #7842, opened 2026-01-11
Expose (json) log format, hide and deprecate obsolete options

arnetheduck:
The `json` log format is documented in the guide but did not, until now, appear in the `--help` text.

Take the opportunity to clean up the options and docs:

* prefer `--el=<url>` throughout (`--web3-url` remains as an availble and equivalent name but no longer appears in `--help`)
* expose `--era` option that is used to distribute historical/archived blocks on the beacon network
* document when options were obsoleted and provide migration docs where reasonable
* If `jwt.hex` exists in the data directory, use it when starting the node - when running Nimbus as both execution and beacon client sharing a data directory, this "just works" since one of them will write while the other reads the file.
* Simplify JWT implmenentation by hardcoding some of the JSON messages
* Add a search bar to the docs (didn't we have one before??)
* Use the unified binary in the execution client docs
* Note that you can now stake more than 32 ETH per validator
* Remove obsolete gitcoin donation link
* Remove BETA tag for validator client metrics - no significant changes are currently planned

### DIFF
--- a/beacon_chain/spec/engine_authentication.nim
+++ b/beacon_chain/spec/engine_authentication.nim
@@ -1,5 +1,5 @@
 # beacon_chain
-# Copyright (c) 2022-2024 Status Research & Development GmbH
+# Copyright (c) 2022-2026 Status Research & Development GmbH
 # Licensed and distributed under either of
 #   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
 #   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
--- a/beacon_chain/spec/engine_authentication.nim
+++ b/beacon_chain/spec/engine_authentication.nim
@@ -23,7 +23,7 @@ export rand, results
 
 const jwtSecretFile* = "jwt.hex"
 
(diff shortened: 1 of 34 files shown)


beacon_chain/spec/engine_authentication.nim
```diff
@@ -8,28 +8,33 @@
...
+from std/os import `/`, fileExists
 
 export rand, results
 
-const
-  JWT_SECRET_LEN = 32
+const jwtSecretFile* = "jwt.hex"
 
-func base64urlEncode(x: auto): string =
+# https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.3/src/engine/authentication.md#key-distribution
```
tersec: https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/authentication.md#key-distribution

tersec:
```
- tests/test_engine_authentication.nim
```

beacon_chain/spec/engine_authentication.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7842/commits/2fb934c197864cf4f8fd6f17d813e523eedb7a92

PR #7826, opened 2025-12-29
Use confutils `obsolete` for deprecated beacon node conf options

nitely:
Changes:

- `obsoleteCmdOpt` overload to output obsolete/deprecated options. Keep using chronicles (otherwise the default write to stderr would be used).
- Pass confutils `load` callback to setup chronicles before outputting deprecation warnings. Adds `setupLogger` param to `loadWithBanners` .

Output example:

```text
WRN 2025-12-23 20:17:44.631-03:00 Ignoring deprecated configuration option   topics="beacnde" opt=web3-force-polling
INF 2025-12-23 20:17:44.632-03:00 Launching beacon node                      topics="beacnde" version=v25[...]
```

Bump confutils:

- https://github.com/status-im/nim-confutils/pull/135
- https://github.com/status-im/nim-confutils/pull/136
- https://github.com/status-im/nim-confutils/pull/126
- https://github.com/status-im/nim-confutils/pull/127
- https://github.com/status-im/nim-confutils/pull/128

PR #7848, opened 2026-01-12
Follow-up #7826; confutils onLoaded logger setup

nitely:
Follow-up #7826

Changes:

- Use `onLoaded` to setup the logger
- bump confutils https://github.com/status-im/nim-confutils/pull/144

### DIFF
--- a/beacon_chain/nimbus_binary_common.nim
+++ b/beacon_chain/nimbus_binary_common.nim
@@ -241,7 +241,7 @@ proc loadWithBanners*(
           if config.configFile.isSome:
             sources.addConfigFile(Toml, config.configFile.get)
         ,
-        loggerSetup = if setupLogger: loggerSetup(ConfType) else: nil
+        onLoaded = if setupLogger: loggerSetup(ConfType) else: nil
       )
     except CatchableError as exc:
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_binary_common.nim
```diff
@@ -238,7 +239,7 @@ proc loadWithBanners*(
           if config.configFile.isSome:
             sources.addConfigFile(Toml, config.configFile.get)
         ,
-        loggerSetup = if setupLogger: loggerSetup(ConfType) else: nil
+        onLoaded = onLoadedConf(ConfType, setupLogger)
```
arnetheduck: the `else: nil` solution is less wasteful though it's not a blocker really

nitely: I know, it creates a closure. But I think I like the idea of the overload better.

nitely: I reverted to `else: nil`. In case you prefer this to the overload. I'm not entirely sure about the overload being better.

arnetheduck:
ok, let's roll with this approach for now, evaluating the generics can be done separately

PR #7852, opened 2026-01-14
Move Fulu kzg check out of column quarantine

ahshum:


beacon_chain/consensus_object_pools/blob_quarantine.nim
cheatfate: I think this type declaration is not needed too. Because it is declared as private and never mentioned in module anymore.

cheatfate (changes requested):

cheatfate (approved):

PR #7856, opened 2026-01-15
Re-enable progressive shape SSZ tests

etan-status:
Adopt changes from https://github.com/ethereum/EIPs/pull/11032/files


etan-status:
Needs:

1. consensus-spec release that contains https://github.com/ethereum/consensus-specs/pull/4813
2. https://github.com/status-im/nim-ssz-serialization/pull/133

PR #7858, opened 2026-01-16
ci: patch hashtree for MacOS Release

siddarthkay:
## Summary
- Patch `hashtree` to skip  `-fno-integrated-as` for macOS arm64
- because `osxcross` doesn't have an arm64 GNU assembler, so we must use clang's integrated assembler which works well with llvm older than v21.

### DIFF
--- a/docker/dist/entry_point.sh
+++ b/docker/dist/entry_point.sh
@@ -111,6 +111,15 @@ elif [[ "${PLATFORM}" == "macOS_arm64" ]]; then
   ${CC} --version
   echo
 
+  # Patch hashtree to not use -fno-integrated-as for macOS arm64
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
Not pretty but works. But shouldn't we open an issue or PR in `nim-ssz-serialization` to make a permanent fix?

tersec:
> Not pretty but works. But shouldn't we open an issue or PR in `nim-ssz-serialization` to make a permanent fix?

This seems like a `hashtree` thing to upstream.

siddarthkay:
> PR in `nim-ssz-serialization` to make a permanent fix

here is the proper fix: https://github.com/status-im/nimbus-eth2/pull/7860

Issue #7767, opened 2025-11-30
v25.11.1 crash


b0a7:
**Describe the bug**
I believe this is caused by having one of the nodes connected to the validator client offline when the validator client is starting up

**To Reproduce**
Steps to reproduce the behavior: start validator client v25.11.1 with one of the connected nodes offline
1. Platform details (OS, architecture): Ubuntu
2. Branch/commit used: statusim/nimbus-validator-client:multiarch-v25.11.1
3. Commands being executed: '...'
4. Relevant log lines:
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-serialization/serialization/errors.nim(453) main
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-serialization/serialization/errors.nim(441) NimMain
rocketpool_validator  | /home/user/nimbus-eth2/beacon_chain/nimbus_validator_client.nim(660) _ZN23nimbus_validator_client4mainE
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/internal/asyncfutures.nim(518) _ZN12asyncfutures7waitForE20InternalRaisesFutureI4void4voidE
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/internal/asyncfutures.nim(635) _ZN12asyncfutures7pollForE20InternalRaisesFutureI4void4voidE
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/internal/asyncengine.nim(150) _ZN11asyncengine4pollE
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-chronos/chronos/internal/asyncfutures.nim(382) _ZN12asyncfutures14futureContinueE3refIN7futures26FutureBasecolonObjectType_EE
rocketpool_validator  | /home/user/nimbus-eth2/vendor/nim-chronicles/chronicles/formats.nim(20) _ZN13runVCSlotLoop13runVCSlotLoopE3refIN7futures26FutureBasecolonObjectType_EE
rocketpool_validator  | /home/user/nimbus-eth2/beacon_chain/validator_client/common.nim(923) _ZN6common28getDurationToNextAttestationE3refIN6common15ValidatorClientEEN9constants4SlotE
[...]

PR #7862, opened 2026-01-19
VC: Fix assertion crash len(vc.forks) > 0.

cheatfate:
Add more comprehensive log statements for fork_service.
Should address #7767

### DIFF
--- a/beacon_chain/validator_client/block_service.nim
+++ b/beacon_chain/validator_client/block_service.nim
@@ -657,17 +657,33 @@ proc runBlockPollMonitor(service: BlockServiceRef,
       await noCancel allFutures(pending)
       raise exc
 
-proc runBlockMonitor(service: BlockServiceRef) {.
-     async: (raises: [CancelledError]).} =
+proc runBlockMonitor(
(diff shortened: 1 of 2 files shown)


tersec:
The more comprehensive logging is useful, but how does it address https://github.com/status-im/nimbus-eth2/issues/7767 ? The issue there is also that it crashes with an assertion, which should be for internal logic errors only, not something a BN can trigger no matter what it does.

cheatfate:
> The more comprehensive logging is useful, but how does it address #7767 ? The issue there is also that it crashes with an assertion, which should be for internal logic errors only, not something a BN can trigger no matter what it does.

So usually all services which are going to use any forks functions should wait for `forksAvailable` event. I have checked all the services and found that block monitoring loops missing this feature. So it starts using getForks before internal sequence was populated with data and got the crash. 

This is actual fix https://github.com/status-im/nimbus-eth2/pull/7862/changes#diff-65584646c3da967b357dcf060fac5423042b2f938f3079fcf1a0a2c67004a01cR671-R674

PR #7864, opened 2026-01-20
bump nim-libp2p to v1.15.1

tersec:


config.nims
```diff
@@ -88,6 +88,35 @@ if defined(windows):
   # toolchain: https://github.com/status-im/nimbus-eth2/issues/3121
   switch("define", "nimRawSetjmp")
 
+# QUIC
+block:
+  if defined(windows):
+    # ls-quic requires windows specific pthread otherwise the compiler will complain
+    # about missing pthread functions
+    switch("passL", "-lwinpthread")
```
arnetheduck: does `-pthread` also work here? this seems to be what lsquic itself uses (along with -m` for maths): https://github.com/litespeedtech/lsquic/blob/9e7c2ba0f86594ea00f69140b6d52afaf2722781/CMakeLists.txt#L389

config.nims
```diff
@@ -88,6 +88,35 @@ if defined(windows):
...
+        "d57acbf46e010707864a5e75887d1d01", "5e1562ddbee31f225865870baadb7606"]:
+      writeFile(path, anglebracket_include)
+
+    if defined(windows):
+      let asmFiles = readFile(
+        basepath / "scripts" / "boringssl_win_nasm.list").splitLines().filterIt(
+          it.len > 0)
+      # https://github.com/vacp2p/nim-lsquic/blob/main/lsquic.nimble
+      for asmPath in asmFiles:
+        exec "nasm -f win64 " & quoteShell(basepath / asmPath) & " -o " &
```
arnetheduck: in lsquic, *I think* it should be possible to do the following:

* call nasm with `staticExec`, preferably gated to be called only when the object file is missing
* use `{.passl: "asmoutput.o".}` for linking it

the aim would be to avoid the need for these customizations of the config.nims file.

arnetheduck: we do something similar with `rcc` and resource compilations in qt: https://github.com/arnetheduck/ngui/blob/master/src/ngui.nim#L13

config.nims
richard-ramos: I'll try and see what up.

richard-ramos: https://github.com/vacp2p/nim-lsquic/pull/60 yup, that works.

arnetheduck: :+1: in general, it's a useful trick when translating build systems - just do what they do ;)

config.nims
richard-ramos: Testing this approach in https://github.com/status-im/nimbus-eth2/pull/7898

PR #7866, opened 2026-01-20
tie `getBlobs` calling with block gossip callback

agnxsh:
Fairly deterministic way to chalk out the right moment to ask for blobs from the blobpool

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -328,6 +328,7 @@ proc initFullNode(
       EventBeaconBlockGossipObject.init(data))
 
     withBlck(data):
+      debugGloasComment ""
       when consensusFork == ConsensusFork.Fulu:
         asyncSpawn node.attemptGetBlobs(forkyBlck.root)
 
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -283,6 +283,8 @@ proc isSlotWithinWeakSubjectivityPeriod(dag: ChainDAGRef, slot: Slot): bool =
   is_within_weak_subjectivity_period(dag.cfg, slot,
                                      dag.headState, checkpoint)
 
+proc attemptGetBlobs(node: BeaconNode, root: Eth2Digest) {.async.}
+
 proc initFullNode(
     node: BeaconNode,
     rng: ref HmacDrbgContext,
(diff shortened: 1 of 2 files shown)


tersec:
If all the required (custodied) columns have arrived over gossip, is it still worth doing this? Does that depend on whether the node would in theory need to reconstruct as a supernode?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -324,6 +326,11 @@ proc initFullNode(
   proc onBlockGossipAdded(data: ForkedSignedBeaconBlock) =
     node.eventBus.blockGossipQueue.emit(
       EventBeaconBlockGossipObject.init(data))
+
+    withBlck(data):
+      when consensusFork == ConsensusFork.Fulu:
```
tersec, line 332: Is this actually Fulu-specific, or does it work on Fulu and later forks too? [`getBlobsV2`](https://github.com/ethereum/execution-apis/blob/c4988b1c427645d1b861c8c12488975f5a2f8cb9/src/engine/osaka.md#engine_getblobsv2) and [`getBlobsV3`](https://github.com/ethereum/execution-apis/blob/c4988b1c427645d1b861c8c12488975f5a2f8cb9/src/engine/osaka.md#engine_getblobsv3) seem to return the same (potentially nullable) `BlobAndProofV2`s regardless of potentially being in Gloas.

agnxsh:
> If all the required (custodied) columns have arrived over gossip, is it still worth doing this? Does that depend on whether the node would in theory need to reconstruct as a supernode?

if all the required columns have arrived over gossip within say the first 2 seconds into the slot, then it's not as useful to do this, however, after looking at mainnet data, with higher blob counts, sidecars come in much later than expected

image

as late as 3.25 seconds into the slot, so it's safe to have this optimization in place, the other advantage being, it greatly eases out latency of verifying kzg proofs/signatures, yes not the most trustless approach, but it's still a plus in terms of gettting to faster DA with higher blob count (atleast before we have partial messages).

beacon_chain/nimbus_beacon_node.nim
agnxsh: i'm not entirely convinced that Gloas payload types are not subject to change in the near future, hence i kept it that way, before launching a real Gloas devnet we can change it.

tersec: Ok. good to leave a `debugGloasComment` then.

tersec:
> > If all the required (custodied) columns have arrived over gossip, is it still worth doing this? Does that depend on whether the node would in theory need to reconstruct as a supernode?
> 
> if all the required columns have arrived over gossip within say the first 2 seconds into the slot, then it's not as useful to do this, however, after looking at mainnet data, with higher blob counts, sidecars come in much later than expected
> image
> 
> as late as 3.25 seconds into the slot, so it's safe to have this optimization in place, the other advantage being, it greatly eases out latency of verifying kzg proofs/signatures, yes not the most trustless approach, but it's still a plus in terms of gettting to faster DA with higher blob count (atleast before we have partial messages).

Interesting/useful data. Somewhat unexpected given that MEV relays/builders have no incentive to delay blobs/columns but they do blobs.

beacon_chain/nimbus_beacon_node.nim
agnxsh: ok

PR #7870, opened 2026-01-21
nix: use Nix Flake from NBS repo to provide Nim

jakubgs:
This way we can avoid rebuilding Nim every time we build packages.

Compilation with Nim: 12 minutes
Compilation without Nim: 8.5 minutes

Depends on:
- https://github.com/status-im/nimbus-build-system/pull/112


flake.nix
```diff
@@ -1,9 +1,15 @@
 {
   description = "nimbus-eth2";
 
-  inputs.nixpkgs.url = github:NixOS/nixpkgs/master;
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs?rev=ff0dbd94265ac470dda06a657d5fe49de93b4599";
+    nimbusBuildSystem = {
+      url = "git+file:./vendor/nimbus-build-system?submodules=1";
```
markoburcul, line 7: Why do we use git+file and not reference with git+https? How would this behave if we reference the nimbus-eth2 flake from another repo? Would we need to override the inputs of this flake in that case and use git+https or git+file just works even in that case?

markoburcul (approved):
Left some questions, but it looks way cleaner and easier to understand. It can serve us as an example for other repos dependent on the NBS and NIm, so thank you for this!

siddarthkay:
I tested this, perhaps the nix develop shell needs to force nim from nix instead of picking from homebrew in my case?
```
siddarthkumar in ~/code/status-im/nimbus-eth2 on nix/use-nbs-flake λ nix develop --experimental-features 'nix-command flakes'                                                                        1.165s
__        __   _                            _
\ \      / /__| | ___ ___  _ __ ___   ___  | |_ ___
 \ \ /\ / / _ \ |/ __/ _ \| '_ ` _ \ / _ \ | __/ _ \
  \ V  V /  __/ | (_| (_) | | | | | |  __/ | || (_) |
   \_/\_/ \___|_|\___\___/|_| |_| |_|\___|  \__\___/

 _   _ _           _                          _   _     ____
| \ | (_)_ __ ___ | |__  _   _ ___        ___| |_| |__ |___ \
|  \| | | '_ ` _ \| '_ \| | | / __|_____ / _ \ __| '_ \  __) |
| |\  | | | | | | | |_) | |_| \__ \_____|  __/ |_| | | |/ __/
|_| \_|_|_| |_| |_|_.__/ \__,_|___/      \___|\__|_| |_|_____|

bash-5.2$ nim -v
Nim Compiler Version 2.2.6 [MacOSX: arm64]
Compiled at 2025-10-31
Copyright (c) 2006-2025 by Andreas Rumpf

active boot switches: -d:release -d:nimUseLinenoise
bash-5.2$ which nim
/opt/homebrew/bin/nim
bash-5.2$
```

tersec:
As a general point, we're not likely to merge this until we switch to Nim v2.2.8 (hopefully, within the next ~1-1.5 months), because the CI linter points out a real issue:
```
Submodule 'vendor/nimbus-build-system': 'ac5b51bc6a91a2884d0a6b339d216b7e5a1b4865' (2026-01-22 16:23:39 +0000) is more recent than latest 'master' (2026-01-13 14:59:16 +0000) (branch config: '.gitmodules')
```
i.e. we intentionally use trunk branches of submodules.

Right now, the Nim version in `n-b-s` default is v2.2.6, which Nimbus can't use. So we haven't bumped n-b-s in a while, even for non-Nim-version things.

One option is to move n-b-s back to v2.2.4, then pick up other updates such as this one more incrementally, but it depends on the priority here.

Another question is, presumably there's some other process somewhere else which builds/caches the flake? Is this automatic?

flake.nix
jakubgs: I use `git+file:` because it means if `vendor/nimbus-build-system` is updated we just need to call `nix flake update` and the `flake.lock` will be updated with the commit used by that submodule. That's the main advantage. Ease of use.

jakubgs:
@tersec I don't see why we can't cherry pick my changes on top of whatever we are currently using in `unstable` for `vendor/nimbus-build-system` and use that. I don't change much in NBS itself.

As for caching, that's a separate thing, but we can certainly do that, the only issue is that even if we supply our own cache via something like this in the flake:
```nix
  nixConfig = {
    extra-substituters = [ "https://nix-cache.status.im/" ];
    extra-trusted-public-keys = [ "nix-cache.status.im-1:x/93lOfLU+duPplwMSBR+OlY4+mo+dCN7n0mr4oPwgY=" ];
  };
```
Anyone with multi-user setup of Nix will need to either be in trusted user(not great for security as it gives your user passwordless sudo) or add our Nix cache directly to their `/etc/nix/nix.conf`. Which we could of course recommend to the user at `nix develop` shell setup for example. Or even add a script for doing that.

jakubgs:
NBS PR was merged, but I'll put this into draft mode since we have to wait for a proper upgrade to Nim 2.2.8 for this to go in.

tersec:

jakubgs:
@tersec please review when you have time.

tersec (approved):

Issue #7838, opened 2026-01-08
VC sends multiple submit attestation requests


KaloyanTanev:
**Describe the bug**
Post-electra committee index was taken outside of the signed attestation data. Beacon API requests were modified and now VCs managing multiple keys do need to send only 1 request to `/eth/v1/validator/attestation_data` for all of its keys, specifying committee_index as 0. Respectively we also expect only 1 call with all of the signed attestations to `/eth/v2/beacon/pool/attestations`.

Currently Nimbus VC sends multiple signed attestation requests to the beacon node. While this is not incorrect per se, it creates bigger overhead in cases with many keys. In our situation of distributed validators, this is even more prominent, as each of those signatures is a partial signature and the distributed validator needs to propagate this to the other distributed validator nodes.

**To Reproduce**
Applicable to any Nimbus VC with high enough amount of keys that multiple keys need to attest in a slot.

**Screenshots**
Adding screenshots of comparison between Nimbus VC's `/eth/v2/beacon/pool/attestations` requests and another VC in a DV cluster with close to 2000 active keys.

<img width="1410" height="811" alt="Image" src="https://github.com/user-attachments/assets/badf6d12-3b20-4dde-8fc5-0d3cb2ae10e2" />
<img width="1416" height="807" alt="Image" src="https://github.com/user-attachments/assets/13c4aca4-6b70-4c5f-b3b9-525551be1ec0" />

tersec:
- https://github.com/status-im/nimbus-eth2/issues/7838

PR #7874, opened 2026-01-22
VC: Add option to publish all attestations for single slot in one call.

cheatfate:
This PR should address #7838

### DIFF
--- a/beacon_chain/conf.nim
+++ b/beacon_chain/conf.nim
@@ -1054,6 +1054,11 @@ type
       defaultValue: BlockMonitoringType.Event
       name: "block-monitor-type".}: BlockMonitoringType
 
+    batchAttestations* {.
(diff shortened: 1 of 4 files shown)


beacon_chain/conf.nim
```diff
@@ -1054,6 +1054,11 @@ type
       defaultValue: BlockMonitoringType.Event
       name: "block-monitor-type".}: BlockMonitoringType
 
+    batchAttestations* {.
+      desc: "Wait until every attestation signed and send all of them in one call (BETA)"
```
tersec: "until every attestation is signed", or "until every attestation has been signed"

I'm not sure if it's worth explicitly framing this as a DVT flag; is there a use case otherwise?

tersec:
Merge conflict in `beacon_chain/validator_client/attestation_service.nim`

PR #7876, opened 2026-01-23
Bump nim-web3 and related changes

kdeme:
Note that this is not a blind increase of the number of fields. But no additional check can be done on that timestamp field as one would need access to the relevant header.

see change https://github.com/status-im/nim-web3/commit/6868584a2f473eb1da04d5a3a7270d8f6795acf0#diff-eaf5eeecddea56125909e7daaeee2ed63d867c91957a90539f5c4cabd2f58dbeR217-R220


kdeme:
I'll go ahead and merge this as it is rather trivial

PR #7878, opened 2026-01-23
Block production for Gloas

Tomi-3-0:
Implements block production (self-building) for the Gloas fork with Engine API V6 support.


beacon_chain/el/engine_api_conversions.nim
```diff
@@ -127,6 +127,32 @@ func asConsensusType*(rpcExecutionPayload: ExecutionPayloadV3):
     blob_gas_used: rpcExecutionPayload.blobGasUsed.uint64,
     excess_blob_gas: rpcExecutionPayload.excessBlobGas.uint64)
 
+func asConsensusType*(rpcExecutionPayload: ExecutionPayloadV4):
```
tersec: This function appears to have an identical body to
https://github.com/status-im/nimbus-eth2/blob/341fe477d8f5e1a993174fd0f8081aabd3826296/beacon_chain/el/engine_api_conversions.nim#L104-L128
so could instead allow
```nim
func asConsensusType*(rpcExecutionPayload: ExecutionPayloadV3):
    deneb.ExecutionPayload =
```
to instead become
```nim
func asConsensusType*(
    rpcExecutionPayload: ExecutionPayloadV3 | ExecutionPayloadV4):
    deneb.ExecutionPayload =
```

tersec: https://github.com/status-im/nimbus-eth2/pull/7878/commits/d740302b02c41035af6bb3607507b0c96cf05bb5

PR #7883, opened 2026-01-24
Add Gloas data column sidecar support to block production/proposal

Tomi-3-0:
https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/builder.md#constructing-the-datacolumnsidecars
This adds Gloas specific `assemble_data_column_sidecars` overload and acccepts `kzg_commitments` as explicit parameter (from blobs bundle, not beacon block body) and uses `beacon_block_root` and `slot` fields instead of `signed_block_header`.
Also, it removes inclusion proofs as commitments are no longer in the block body for Gloas

### DIFF
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -267,10 +267,10 @@ proc assemble_data_column_sidecars*(
     signed_beacon_block: gloas.SignedBeaconBlock,
     blobs: seq[KzgBlob], kzg_commitments: deneb.KzgCommitments,
     cell_proofs: seq[KzgProof]): seq[gloas.DataColumnSidecar] =
-  var sidecars = newSeqOfCap[gloas.DataColumnSidecar](CELLS_PER_EXT_BLOB)
-
   if kzg_commitments.len == 0 or blobs.len == 0:
-    return sidecars
+    return static(default(seq[gloas.DataColumnSidecar]))
+
+  var sidecars = newSeqOfCap[gloas.DataColumnSidecar](CELLS_PER_EXT_BLOB)
 
   if blobs.len != kzg_commitments.len:
     return sidecars
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -273,13 +273,9 @@ proc assemble_data_column_sidecars*(
     return sidecars
 
   if blobs.len != kzg_commitments.len:
-    warn "Blob count mismatch with commitments",
-      blobs = blobs.len, commitments = kzg_commitments.len
     return sidecars
 
   if cell_proofs.len != blobs.len * CELLS_PER_EXT_BLOB:
-    warn "Cell proofs count mismatch",
-      expected = blobs.len * CELLS_PER_EXT_BLOB, actual = cell_proofs.len
     return sidecars
 
   var
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -208,63 +208,112 @@ proc recover_cells_and_proofs_parallel*(
   ok(res)
 
 proc assemble_data_column_sidecars*(
-    signed_beacon_block: fulu.SignedBeaconBlock | gloas.SignedBeaconBlock,
+    signed_beacon_block: fulu.SignedBeaconBlock,
     blobs: seq[KzgBlob], cell_proofs: seq[KzgProof]): seq[fulu.DataColumnSidecar] =
   template blck(): auto = signed_beacon_block.message
   var sidecars = newSeqOfCap[fulu.DataColumnSidecar](CELLS_PER_EXT_BLOB)
 
-  when signed_beacon_block is gloas.SignedBeaconBlock:
-    debugGloasComment "kzg_commitments removed from beaconblock in gloas"
+  template kzg_commitments: untyped =
+    signed_beacon_block.message.body.blob_kzg_commitments
+  if kzg_commitments.len == 0:
     return sidecars
-  else:
-    template kzg_commitments: untyped =
-      signed_beacon_block.message.body.blob_kzg_commitments
-    if kzg_commitments.len == 0:
-      return sidecars
-    let
-      beacon_block_header =
-        BeaconBlockHeader(
-          slot: blck.slot,
-          proposer_index: blck.proposer_index,
-          parent_root: blck.parent_root,
-          state_root: blck.state_root,
-          body_root: hash_tree_root(blck.body))
+  let
+    beacon_block_header =
+      BeaconBlockHeader(
+        slot: blck.slot,
+        proposer_index: blck.proposer_index,
+        parent_root: blck.parent_root,
+        state_root: blck.state_root,
+        body_root: hash_tree_root(blck.body))
 
-      signed_beacon_block_header =
-        SignedBeaconBlockHeader(
-          message: beacon_block_header,
-          signature: signed_beacon_block.signature)
+    signed_beacon_block_header =
+      SignedBeaconBlockHeader(
+        message: beacon_block_header,
+        signature: signed_beacon_block.signature)
 
+  var
+    cells = newSeq[CellBytes](blobs.len)
+    proofs = newSeq[ProofBytes](blobs.len)
+
+  for i in 0 ..< blobs.len:
+    cells[i] = computeCells(blobs[i]).get
+    let proofElem = addr proofs[i]
+    staticFor j, 0 ..< CELLS_PER_EXT_BLOB:
+      assign(proofElem[][j], cell_proofs[i * CELLS_PER_EXT_BLOB + j])
+
+  for columnIndex in 0..<CELLS_PER_EXT_BLOB:
     var
-      cells = newSeq[CellBytes](blobs.len)
-      proofs = newSeq[ProofBytes](blobs.len)
+      column = newSeqOfCap[KzgCell](blobs.len)
+      kzgProofOfColumn = newSeqOfCap[KzgProof](blobs.len)
+    for rowIndex in 0..<blobs.len:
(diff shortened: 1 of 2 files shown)


beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -208,63 +208,112 @@ proc recover_cells_and_proofs_parallel*(
...
-        sidecar.kzg_commitments_inclusion_proof).expect("Valid gindex")
-      sidecars.add(sidecar)
+  if blobs.len != kzg_commitments.len:
+    warn "Blob count mismatch with commitments",
+      blobs = blobs.len, commitments = kzg_commitments.len
+    return sidecars
 
-    sidecars
+  if cell_proofs.len != blobs.len * CELLS_PER_EXT_BLOB:
+    warn "Cell proofs count mismatch",
```
agnxsh: just debug should be fine here, infact, i think just returning the empty sidecar array should be okay too.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -208,63 +208,112 @@ proc recover_cells_and_proofs_parallel*(
...
-        column: DataColumn.init(column),
-        kzg_commitments: blck.body.blob_kzg_commitments,
-        kzg_proofs: deneb.KzgProofs.init(kzgProofOfColumn),
-        signed_block_header: signed_beacon_block_header)
-      blck.body.build_proof(
-        KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH_GINDEX.GeneralizedIndex,
-        sidecar.kzg_commitments_inclusion_proof).expect("Valid gindex")
-      sidecars.add(sidecar)
+  if blobs.len != kzg_commitments.len:
+    warn "Blob count mismatch with commitments",
```
agnxsh: similar comment regarding warn, debug is fine

beacon_chain/validators/beacon_validators.nim
```diff
@@ -574,11 +576,16 @@ proc proposeBlockAux(
...
     )
 
-  when consensusFork == ConsensusFork.Gloas:
-    # using noSidecars for Phase0 -> Capella blocks for now
-    let sidecarsOpt = Opt.none(seq[gloas.DataColumnSidecar])
-  elif consensusFork >= ConsensusFork.Fulu and
-      consensusFork < ConsensusFork.Gloas:
+  when consensusFork >= ConsensusFork.Gloas:
+    let sidecarsOpt =
+      if engineBid[].eps.blobsBundle.commitments.len > 0:
```
agnxsh: we are already checking for commitments length in the `assemble` function

agnxsh:
I think there are a few more places where on hitting the Gloas consensusFork, we are temporarily assuming it be Opt.none gloas.DataColumnSidecars, they can be adapted later on, or on this PR itself, whatever's fine.

beacon_chain/validators/beacon_validators.nim
Tomi-3-0: Yes, that's right. I kept it because it avoids the function call overhead altogether and also prevents the allocation of the empty sequence.

agnxsh: instead of wrapping the sequence in a result, you can make the `assemble` function for Gloas itself in Result too, i'm not sure how expensive in terms of code generation is `.get`, cc @tersec 

thereby, directly making the function return Opt.none

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7883/changes/a16099313fc6749c86425fc877e3bdd35795e062

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7883/changes/a16099313fc6749c86425fc877e3bdd35795e062

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7883/changes/a16099313fc6749c86425fc877e3bdd35795e062#diff-cca36aeef08bfe3cdc2248acaebf0e00bc741f252a8dacf3dae4735c3b1dded8 addresses this as stated

Tomi-3-0:
> I think there are a few more places where on hitting the Gloas consensusFork, we are temporarily assuming it be Opt.none gloas.DataColumnSidecars, they can be adapted later on, or on this PR itself, whatever's fine.

The other place where we have `Opt.none(seq[gloas.DataColumnSidecar])` are in the REST API / VC paths: https://github.com/status-im/nimbus-eth2/blob/a8ad70fcc43440e917b8723047e0465991f39542/beacon_chain/rpc/rest_beacon_api.nim#L1054

and i think it's better left out of this PR

beacon_chain/validators/beacon_validators.nim
tersec: > instead of wrapping the sequence in a result, you can make the `assemble` function for Gloas itself in Result too, i'm not sure how expensive in terms of code generation is `.get`, cc @tersec
> 
> thereby, directly making the function return Opt.none

`.get`/`.value` is a complete copy, plus almost double stack space usage for that return value, as a result. It does not simply access the correct member of the `Result`.
https://github.com/arnetheduck/nim-results/blob/df8113dda4c2d74d460a8fa98252b0b771bf1f27/results.nim#L870-L879
```nim
func value*[T: not void, E](self: Result[T, E]): maybeLent T {.inline.} =
  ## Fetch value of result if set, or raise Defect
  ## Exception bridge mode: raise given Exception instead
  ## See also: Option.get
  withAssertOk(self):
    when T isnot void:
      # TODO: remove result usage.
      # A workaround for nim VM bug:
      # https://github.com/nim-lang/Nim/issues/22216
      result = self.vResultPrivate
```

It's `.unsafeGet`/`.unsafeValue` which do the efficient thing:
https://github.com/arnetheduck/nim-results/blob/df8113dda4c2d74d460a8fa98252b0b771bf1f27/results.nim#L902-905
```nim
template unsafeValue*[T: not void, E](self: Result[T, E]): T =
  ## Fetch value of result if set, undefined behavior if unset
  ## See also: `unsafeError`
  self.vResultPrivate
```

arnetheduck: So - there's nuance here, and things have changed in recent Nim versions .. 

> .get/.value is a complete copy, 

`get` and `value` don't perform any copies - they return a `lent` which effectively is a pointer to the value - in this regard, they are actually equivalent to `unsafeGet` except for the `Defect`. If you have a `var Result`, you get a `var T` back from `get` which you can mutate / move / etc.

What performs the copy is typically what happens next, namely an assignment:
```nim
let v = xxx.get()
```

In the above snippet, it's not `get` causing the copy but the fact that it's being assigned to `x` - if instead we did:
```nim
f(xxx.get())
```
... no copies would happen.

With that out of the way, there's other nuance too - if you have a mutable `Result`, the value can be _moved_ out of it -  this will happen in ORC and hopefully soon with refc as well - ie _in theory_, something like `let v = f().get()` should in the future _move_ the value out of the temporary returned by f(). We're not there yet though.

```nim
# This should avoid copies, and in future Nim versions, move should not be needed...
let v = move(f().get())

# ...and with a mutable Result, one can move things out of it  - the nim compiler is sometimes, but rarely, smart enough to perform this move automatically for the last usage of `r`

var r: Result[seq[...], ...]
let r2 = move(r.get())
```

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -208,63 +208,108 @@ proc recover_cells_and_proofs_parallel*(
...
+    cell_proofs: seq[KzgProof]): seq[gloas.DataColumnSidecar] =
+  var sidecars = newSeqOfCap[gloas.DataColumnSidecar](CELLS_PER_EXT_BLOB)
 
-    for i in 0 ..< blobs.len:
-      cells[i] = computeCells(blobs[i]).get
-      let proofElem = addr proofs[i]
-      staticFor j, 0 ..< CELLS_PER_EXT_BLOB:
-        assign(proofElem[][j], cell_proofs[i * CELLS_PER_EXT_BLOB + j])
+  if kzg_commitments.len == 0 or blobs.len == 0:
+    return sidecars
```
tersec: One way to keep the no-heap-allocations-if-0 property but also avoid the extra length check at the callsite is to rearrange these:
```nim
proc assemble_data_column_sidecars*(
    signed_beacon_block: gloas.SignedBeaconBlock,
    blobs: seq[KzgBlob], kzg_commitments: deneb.KzgCommitments,
    cell_proofs: seq[KzgProof]): seq[gloas.DataColumnSidecar] =
  if kzg_commitments.len == 0 or blobs.len == 0:
    return static(default(seq[gloas.DataColumnSidecar]))
  var sidecars = newSeqOfCap[gloas.DataColumnSidecar](CELLS_PER_EXT_BLOB)

  ...
```

tersec: https://github.com/status-im/nimbus-eth2/pull/7883/commits/48ca73558fdb94064067d03c1f273ff1194b9192

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -208,63 +208,108 @@ proc recover_cells_and_proofs_parallel*(
...
+      SignedBeaconBlockHeader(
+        message: beacon_block_header,
+        signature: signed_beacon_block.signature)
 
+  var
+    cells = newSeq[CellBytes](blobs.len)
+    proofs = newSeq[ProofBytes](blobs.len)
+
+  for i in 0 ..< blobs.len:
+    cells[i] = computeCells(blobs[i]).get
```
arnetheduck, line 239: how much time does this typically take? is it worth parallellising?

PR #7887, opened 2026-01-26
refactor: remove duplicate validator index assignment

strmfos:
setValidatorIndex() already sets validator.index internally, so setting it  again right after is redundant. Removed the duplicate line.


strmfos:
Sorry about that, my bad

PR #7889, opened 2026-01-27
Envelope gossip update

ahshum:
Part of #7748

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -31,7 +31,7 @@ from ../consensus_object_pools/block_quarantine import
 from ../consensus_object_pools/blob_quarantine import
   BlobQuarantine, ColumnQuarantine, GloasColumnQuarantine, popSidecars, put
 from ../consensus_object_pools/envelope_quarantine import
-  EnvelopeQuarantine
+  EnvelopeQuarantine, addMissing, addOrphan, popOrphan
 from ../validators/validator_monitor import
   MsgSource, ValidatorMonitor, registerAttestationInBlock, registerBeaconBlock,
   registerSyncAggregateInBlock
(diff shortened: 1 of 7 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -2387,6 +2387,19 @@ proc installMessageValidators(node: BeaconNode) =
                 node.processor[].processExecutionPayloadBid(
                   MsgSource.gossip, signedBid)))
 
+        # execution_payload
+        # https://github.com/ethereum/consensus-specs/blob/v1.6.1/specs/gloas/p2p-interface.md#execution_payload
```
tersec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/p2p-interface.md#execution_payload

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -852,3 +852,74 @@ proc addBlock*(
...
     of VerifierError.Duplicate:
       err(res.error())
+
+proc storePayload(
+    self: ref BlockProcessor,
+    signedBlock: gloas.SignedBeaconBlock,
+    signedEnvelope: gloas.SignedExecutionPayloadEnvelope,
+    sidecarsOpt: Opt[gloas.DataColumnSidecars],
+): Future[Result[void, VerifierError]] {.async: (raises: [CancelledError]).} =
+  ok()
```
tersec, line 884: This doesn't seem to do anything, so most of the rest of the `block_processor` changes don't either. Is the intent to fill this in? If so, `debugGloasComment`.

ahshum: Yeah the logic is there but those require changes from other PRs so removing the inner logic is easier for merging. I just added more changes to address this.

PR #7894, opened 2026-01-28
Gloas clearance state transition

ahshum:
Part of #7748

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -467,6 +467,68 @@ proc addBackfillBlock*(
 
   ok()
 
+proc addHeadExecutionPayload*(
+    dag: ChainDAGRef,
+    signedBlock: gloas.SignedBeaconBlock,
+    signedEnvelope: gloas.SignedExecutionPayloadEnvelope,
+): Result[BlockRef, VerifierError] =
+  ## Try adding the execution payload envelope to the head block, which should
+  ## usually be invoked after the call of addHeadBlockWithParent()
+  ##
+  ## First check that the block and envelope are matched with the DAG block.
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -467,6 +467,68 @@ proc addBackfillBlock*(
...
+  # Check with the DAG head.
+  let blck = dag.head
+  if not (
+    blck.root() == envelopeBlockRoot() and
+    blck.slot() == signedEnvelope.message.slot
+  ):
+    debug "Envelope is not for the current head"
+    return err(VerifierError.Invalid)
+
+  var cache: StateCache
```
tersec, line 511: Building a `StateCache` from scratch is expensive (a sizeable fraction of a second) because it has to shuffle all the validators. 

The easiest approach is, if the caller of `addHeadExecutionPayload` has/will have access to an already-constructed `StateCache`, use that and pass it in as a parameter.

Another possibility is to use https://github.com/status-im/nimbus-eth2/blob/00429942de066d58494c38061108891696744d04/beacon_chain/consensus_object_pools/blockchain_dag.nim#L663-L702

An additional option is to use `EpochRef` and https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/consensus_object_pools/spec_cache.nim versions of relevant functions.

These are listed in approximate order, most readily implemented first, of preference, i.e. if a pre-existing `StateCache` is available, just reuse it.

ahshum: I have reviewed the logic of the caller (`block_processor`) side. It could construct a `StateCache` on its own but I believe it isn't very useful as the only possible second use may be on the head state transition which a clean cache would be preferred.

I modified to use `loadStateCache` to preload cache instead.

PR #7898, opened 2026-01-28
fix: lsquic nasm file compilation

richard-ramos:


arnetheduck:
> D:\a\nimbus-eth2\nimbus-eth2\vendor\nimbus-build-system\vendor\Nim\lib\pure\os.nim(298, 7) Error: cannot 'importc' variable at compile time; findClose

hmm right might not be able to do the modification check unless there's some way to do it in windows with `gorgeEx` and looking at the return code

richard-ramos:
I'm trying now with a md5 checksum.

richard-ramos:
Merging it 'cause it worked https://github.com/status-im/nimbus-eth2/actions/runs/21482931343/job/61884008252

PR #7903, opened 2026-01-29
ForkedHashedBeaconState helpers

arnetheduck:
Using `func` avoids a template expansion at every usage site and just looks nicer .. unlike earlier nim versions, good code  is generated for `lent`.

An additional benefit is that mutability semantics are preserved and the resulting value is immutable even when being accessed indirectly (unlike `addr`+`[]` which results in a `var`).

No matter what, when iterating over validators we continue to be affected by:

* https://github.com/nim-lang/Nim/issues/25470
* https://github.com/nim-lang/Nim/issues/25469

### DIFF
--- a/beacon_chain/beacon_node_light_client.nim
+++ b/beacon_chain/beacon_node_light_client.nim
@@ -19,7 +19,7 @@ func shouldSyncOptimistically*(node: BeaconNode, wallSlot: Slot): bool =
     when lcDataFork > LightClientDataFork.None:
       shouldSyncOptimistically(
         optimisticSlot = forkyHeader.beacon.slot,
-        dagSlot = getStateField(node.dag.headState, slot),
+        dagSlot = node.dag.headState.slot,
         wallSlot = wallSlot)
     else:
       false
--- a/beacon_chain/consensus_object_pools/attestation_pool.nim
+++ b/beacon_chain/consensus_object_pools/attestation_pool.nim
@@ -169,9 +169,8 @@ proc init*(T: type AttestationPool, dag: ChainDAGRef,
     doAssert status.isOk(), "Error in preloading the fork choice: " & $status.error
 
   info "Fork choice initialized",
-    justified = shortLog(getStateField(
-      dag.headState, current_justified_checkpoint)),
-    finalized = shortLog(getStateField(dag.headState, finalized_checkpoint))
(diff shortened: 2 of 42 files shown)


tersec:
One nice aspect of the slight ugliness of `getStateField` is that it is somewhat involved:
https://github.com/status-im/nimbus-eth2/blob/f896feb674ea5e0fa7ec88d6ded6d3705fb57068/beacon_chain/spec/forks.nim#L987-L1020

i.e. being a kind of explicit call helps discourage a bunch of `getStateField`s near each other on the same state, while it would be better to have one `withState` covering them all. This PR hides the ugliness but it's a factory for spawning lots of unnecessary separate `case` statements/expressions on the fork.

arnetheduck:
> factory for spawning lots

yeah I thought about this but I think it's negligible - specially given the fact that all fields are at the same offset meaning that effectively, the case _should_ go away, or almost at least. If there was any significant computation involved, I'd certainly agree.

arnetheduck:
also, we don't use Forked* in tight loops anyway, typically .. ie everywhere where performance matters, we go `Forked`

arnetheduck:
@jakubgs nix not found?

arnetheduck:
@siddarthkay what's with the mac?

siddarthkay:
`macos` error is this : 
```
[2026-01-29T11:50:32.966Z] LLVM ERROR: Unsupported stack probing method
[2026-01-29T11:50:32.966Z] clang: error: unable to execute command: Abort trap: 6
[2026-01-29T11:50:32.966Z] clang: error: linker command failed due to signal (use -v to see invocation)
[2026-01-29T11:50:32.966Z] make[2]: *** [build] Error 1
[2026-01-29T11:50:32.966Z] make[1]: *** [ncli_testnet] Error 2
[2026-01-29T11:50:32.967Z] make[1]: *** Waiting for unfinished jobs....
```
let me replay the pipeline with `-v` to learn more

siddarthkay:
@arnetheduck : could be an llvm version issue, I see that macos job passes now. @jakubgs was applying it yesterday.

jakubgs:
>@jakubgs nix not found?

That job is not a mandatory check for merging PRs. I re-enabled it to test my PR:

- https://github.com/status-im/nimbus-eth2/pull/7870

And added a branch filter, but I guess I did it wrong because it still runs for other PRs.

PR #7906, opened 2026-01-30
fix: OPENSSL_free warning

richard-ramos:


config.nims
```diff
@@ -231,3 +231,4 @@ put("sqlite3.always", "-fno-lto") # -Wno-return-local-addr
 put("lsquic_enc_sess_ietf.always", "-fno-lto -Wno-stack-usage")
 put("lsquic_handshake.always", "-fno-lto -Wno-stack-usage")
 put("lsquic_hkdf.always", "-fno-lto -Wno-stack-usage")
+put("mem.always", "-fno-lto -Wno-stack-usage")
```
tersec: Shouldn't need `-Wno-stack-usage` here, only `-fno-lto`.

tersec: Shouldn't need `-Wno-stack-usage` here, only `-fno-lto`. Also, would be useful to put in the comments adjacent to the `put` the warning it's fixing. The issue is that over time, these essentially superstitions build up -- some exception that isn't obviously necessary, but a year from now, who know why.

config.nims
```diff
@@ -231,3 +231,11 @@ put("sqlite3.always", "-fno-lto") # -Wno-return-local-addr
...
 put("lsquic_handshake.always", "-fno-lto -Wno-stack-usage")
 put("lsquic_hkdf.always", "-fno-lto -Wno-stack-usage")
+
+# ############################################################
+#
+#  Required for BoringSSL warning about frees on offset pointers
+#
+# ############################################################
+
+put("mem.always", "-Wno-free-nonheap-object")
```
tersec, line 241: This only disables the warning, not the underlying root cause. If we do this, we should know confidently that it's a false positive.

With `-fno-lto`, its at least plausible it genuinely avoids some real issue, because it changes how the translation units are processed by the C compiler.

There are some cases where we disable warnings outright due to false positives; it happens. But it needs more reasoning/evidence than this.

config.nims
tersec: https://github.com/status-im/nimbus-eth2/pull/7906/changes/302d683a45d62993dbbc25de8f56c5ff2a2fd795

config.nims
tersec: This looks at least intentional:
```c

void OPENSSL_free(void *orig_ptr) {
  if (orig_ptr == nullptr) {
    return;
  }

  if (OPENSSL_memory_free != nullptr) {
    OPENSSL_memory_free(orig_ptr);
    return;
  }

  void *ptr = ((uint8_t *)orig_ptr) - OPENSSL_MALLOC_PREFIX;
  __asan_unpoison_memory_region(ptr, OPENSSL_MALLOC_PREFIX);

  size_t size = *(size_t *)ptr;
  OPENSSL_cleanse(ptr, size + OPENSSL_MALLOC_PREFIX);

// ASan knows to intercept malloc and free, but not sdallocx.
#if defined(OPENSSL_ASAN)
  (void)sdallocx;
  free(ptr);
#else
  if (sdallocx) {
    sdallocx(ptr, size + OPENSSL_MALLOC_PREFIX, 0 /* flags */);
  } else {
    free(ptr);
  }
#endif
}
```

PR #7907, opened 2026-02-01
use `func`s for `getForkedBlockField`

arnetheduck:
Similar to its state counterpart, this avoids some ever growing template expansions and looks nicer.


tersec:
Same broad concern as last time: these are (somewhat, comparatively) expensive operations which should look that way. But sure, ok.

PR #7918, opened 2026-02-02
Option -> Opt in fork choice files

etan-status:
Use the more modern `results` library instead of stdlib in fork choice.


tersec:
Merge conflict in `beacon_chain/fork_choice/proto_array.nim`.

PR #7922, opened 2026-02-02
Broadcast execution payload envelope after block proposal for self-building

Tomi-3-0:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -620,8 +620,38 @@ proc proposeBlockAux(
...
+    else:
+      let signedEnvelope = gloas.SignedExecutionPayloadEnvelope(
+        message: envelope,
+        signature: signatureRes.get()
+      )
+
+      discard await node.router.routeExecutionPayloadEnvelope(
+        signedEnvelope, checkValidator = false)
+
+    notice "Payload Envelope proposed",
```
arnetheduck, line 650: should this be logged even when signing fails?

Tomi-3-0: No. the log is intended for when signing succeeds in the `else` block.

PR #7926, opened 2026-02-03
Update gloas spec to v.1.7.0-alpha.2

Tomi-3-0:
https://github.com/ethereum/consensus-specs/releases/tag/v1.7.0-alpha.2


beacon_chain/spec/beaconstate.nim
tersec: These should be updated to show the latest specs they're known to comply with. So it's being updated here for alpha.2, that will be https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#new-get_builder_withdrawals

beacon_chain/spec/beaconstate.nim
```diff
@@ -1772,7 +1773,10 @@ func get_builder_withdrawals(
     state: gloas.BeaconState, withdrawal_index: WithdrawalIndex,
     prior_withdrawals: seq[Withdrawal]):
     (seq[Withdrawal], WithdrawalIndex, uint64) =
-  const withdrawals_limit = MAX_WITHDRAWALS_PER_PAYLOAD
+  const withdrawals_limit = MAX_WITHDRAWALS_PER_PAYLOAD - 1
+
+  doAssert len(prior_withdrawals) <= withdrawals_limit
```
tersec, line 1779: Generally speaking, conditions that the spec identifies as `assert`, as here:
```python
def get_builder_withdrawals(
    state: BeaconState,
    withdrawal_index: WithdrawalIndex,
    prior_withdrawals: Sequence[Withdrawal],
) -> Tuple[Sequence[Withdrawal], WithdrawalIndex, uint64]:
    withdrawals_limit = MAX_WITHDRAWALS_PER_PAYLOAD - 1
    assert len(prior_withdrawals) <= withdrawals_limit
```
don't translate into `doAssert`, or any kind of `assert`, in Nimbus's state transition function, because that would crash the client. Instead, they're signal that this is a condition under which `get_builder_withdrawals()`, in this case, should signal an error.

However, this does appear to be a special case, where the internal logic of `get_builder_withdrawals()`/`get_expected_withdrawals()` implies this assertion to be true: the only place `get_builder_withdrawals()` is called is from
```python
def get_expected_withdrawals(state: BeaconState) -> ExpectedWithdrawals:
    withdrawal_index = state.next_withdrawal_index
    withdrawals: List[Withdrawal] = []

    # [New in Gloas:EIP7732]
    # Get builder withdrawals
    builder_withdrawals, withdrawal_index, processed_builder_withdrawals_count = (
        get_builder_withdrawals(state, withdrawal_index, withdrawals)
    )
    withdrawals.extend(builder_withdrawals)
```
i.e. this assertion is not only true, but `len(prior_withdrawals)` should always be 0.

But it's worth commenting why this is a safe assertion to add, because typically, it's worth being cautious about adding assertions to this section of codebase unless they're clearly detecting internal logic errors/inconsistencies rather than external input.

beacon_chain/spec/beaconstate.nim
tersec: Since `get_validator_sweep_withdrawals()` is being updated to alpha.2 and being checked for compliance for this, its URL should be updated to https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#modified-get_validators_sweep_withdrawals

beacon_chain/spec/beaconstate.nim
```diff
@@ -1911,6 +1920,8 @@ func get_validators_sweep_withdrawals(
     validators_limit = min(len(state.validators), MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP)
   const withdrawals_limit = MAX_WITHDRAWALS_PER_PAYLOAD
 
+  doAssert len(prior_withdrawals) <= withdrawals_limit
```
tersec: The [spec condition](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#modified-get_validators_sweep_withdrawals) is slightly different:
```python
    # There must be at least one space reserved for validator sweep withdrawals
    assert len(prior_withdrawals) < withdrawals_limit
```

But for this to remain a `doAssert` rather than some other error-handling condition, it's again necessary to show that it cannot occur regardless of external input to `get_expected_withdrawals()`, i.e. regardless of the `BeaconState` it's processing.

It occurs in two places: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#modified-get_expected_withdrawals and https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#modified-get_expected_withdrawals

Practically, it's true that Nimbus's Electra `get_expected_withdrawals()` was not changed to use this function, but it might be, so the reasoning should hold based on the spec and not just Nimbus's current deliberately incomplete refactoring.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2651,7 +2790,7 @@ func upgrade_to_next*(
       setBit(res, i)
     res
 
-  gloas.BeaconState(
+  var post = gloas.BeaconState(
```
tersec: This puts the `gloas.BeaconState` on the stack, which won't or shouldn't fit, due to the (intentional) check at https://github.com/status-im/nimbus-eth2/blob/d5bc16fa45b8114a1fdbdef29dd8d33dc68e775b/config.nims#L64-L70

The trick which makes this work in the Altair and Electra update functions is:
https://github.com/status-im/nimbus-eth2/blob/d5bc16fa45b8114a1fdbdef29dd8d33dc68e775b/beacon_chain/spec/beaconstate.nim#L2153-L2154
and
https://github.com/status-im/nimbus-eth2/blob/d5bc16fa45b8114a1fdbdef29dd8d33dc68e775b/beacon_chain/spec/beaconstate.nim#L2450-L2451
i.e. it's not really creating a new variable `post`, but rather assigning ultimately via a `post` template to the space already allocated (hopefully, in the heap) by the caller using [return value optimization](https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization).

More details at
- https://github.com/status-im/nimbus-eth2/pull/7713

beacon_chain/spec/beaconstate.nim
tersec: If a spec function is getting updated, its URL should too: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#new-get_pending_partial_withdrawals

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/ffcb1b85e47630d1fb18d9952e90ae45fa86c632

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/ffcb1b85e47630d1fb18d9952e90ae45fa86c632

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/ffcb1b85e47630d1fb18d9952e90ae45fa86c632

tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/ffcb1b85e47630d1fb18d9952e90ae45fa86c632

tersec:
They added (probably) some new operations directory (in phase0 though?)
```
[2026-02-04T11:31:03.818Z] Error: unhandled exception: /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7926/tests/consensus_spec/phase0/test_fixture_operations.nim(36, 1) `toHashSet(mapIt(toSeq(walkDir(OpDir, relative = false)), it.path)) ==
[2026-02-04T11:31:03.818Z]     toHashSet([OpAttestationsDir, OpAttSlashingDir, OpBlockHeaderDir,
[2026-02-04T11:31:03.818Z]                OpDepositsDir, OpProposerSlashingDir, OpVoluntaryExitDir])`  [AssertionDefect]
[2026-02-04T11:31:03.818Z] [Summary] 0 tests run (0.00s): 0 OK, 0 FAILED, 0 SKIPPED
```
this is why they're all failing.

tersec:
Oh, it's because this PR needs to include the `nim-eth2-scenarios` bump to https://github.com/status-im/nim-eth2-scenarios/commit/99ba940767aa20c1d217439578bad5e02940a56c

beacon_chain/spec/beaconstate.nim
```diff
@@ -2138,6 +2152,134 @@ func translate_participation(
...
+    state.builders.mitem(opt_validator_index.get).balance += amount
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/fork.md#new-onboard_builders_from_pending_deposits
+func onboard_builders_from_pending_deposits*(
+    cfg: RuntimeConfig,
+    state: var gloas.BeaconState) =
+  ## Applies any pending deposit for builders, effectively
+  ## onboarding builders at the fork.
+
+  var validator_pubkeys = initHashSet[ValidatorPubKey]()
```
tersec: https://nim-lang.org/docs/sets.html#initHashSet documents that
> Starting from Nim v0.20, sets are initialized by default and it is not necessary to call this function explicitly.

i.e. one can just write
```nim
  var validator_pubkeys: HashSet[ValidatorPubKey]
```

beacon_chain/spec/beaconstate.nim
```diff
@@ -2138,6 +2152,134 @@ func translate_participation(
...
+
+    # If the pubkey is associated with a builder that was created in a
+    # previous iteration or it is a builder deposit, try to apply the
+    # deposit to the new/existing builder. Note that the function
+    # apply_deposit_for_builder can mutate the state and may add a builder
+    # to the registry. For this reason, the list of builder pubkeys must
+    # be recomputed each iteration.
+    var builder_pubkeys = initHashSet[ValidatorPubKey]()
+    for builder in state.builders:
+      builder_pubkeys.incl(builder.pubkey)
```
tersec: Similar comment here about `initHashSet` but more saliently, it already does
```nim
  var
    bucket_sorted_builders = sortValidatorBuckets(state.builders.asSeq)
```
so it already has this information. It doesn't need to rebuild it.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2138,6 +2152,134 @@ func translate_participation(
...
+
+    let
+      is_existing_builder = deposit.pubkey in builder_pubkeys
```
tersec: Can use
https://github.com/status-im/nimbus-eth2/blob/8f368effd5eedc69453f95757a21f93e30f05528/beacon_chain/validator_bucket_sort.nim#L73-L90
and check for `Opt.none` (`isNone`, etc).

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/4bb6a0cb169ef79ba712596465449fc551a8e6ae

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/4bb6a0cb169ef79ba712596465449fc551a8e6ae

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/4bb6a0cb169ef79ba712596465449fc551a8e6ae

beacon_chain/spec/beaconstate.nim
```diff
@@ -2138,6 +2152,131 @@ func translate_participation(
...
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/fork.md#new-onboard_builders_from_pending_deposits
+func onboard_builders_from_pending_deposits*(
+    cfg: RuntimeConfig,
+    state: var gloas.BeaconState) =
+  ## Applies any pending deposit for builders, effectively
+  ## onboarding builders at the fork.
+
+  var validator_pubkeys: HashSet[ValidatorPubKey]
+  for validator in state.validators:
+    validator_pubkeys.incl(validator.pubkey)
```
tersec: Better to use https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/validator_bucket_sort.nim for this, as it's both faster and uses an order of magnitude less memory. in particular, `sortValidatorBuckets(state.validator.asSeq)`. It's both faster and uses less memory than this approach:
- https://github.com/status-im/nimbus-eth2/pull/6469

By contrast to the less than 10MB usage of one of these sorted validator buckets, there are 2M+ validators, and each pubkey is 48 bytes, so even the most efficient (but completely storing pubkeys is going to take 100MB or so RAM minimum, and Nim's `HashSet` isn't especially efficient with millions of medium-size items, so it seems to consume 200+ MB for this data structure alone.

Then the check for `if deposit.pubkey in validator_pubkeys` runs again through `findValidatorIndex`.

And instead of `validator_pubkeys.incl(deposit.pubkey)`, to track changes to this, rather than regenerate the whole thing (it's a mix of semi-immutable and small mutable structure) can use https://github.com/status-im/nimbus-eth2/blob/8f368effd5eedc69453f95757a21f93e30f05528/beacon_chain/validator_bucket_sort.nim#L68-L71

Otherwise it will behave incorrectly with two builder deposits close in time to each other, in the same `
onboard_builders_from_pending_deposits()` call.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2138,6 +2152,131 @@ func translate_participation(
...
+    state.builders.mitem(index) = builder
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-apply_deposit_for_builder
+func apply_deposit_for_builder*(
+    cfg: RuntimeConfig, state: var gloas.BeaconState,
+    bucket_sorted_builders: var BucketSortedValidators,
+    pubkey: ValidatorPubKey, withdrawal_credentials: Eth2Digest,
+    amount: Gwei, signature: ValidatorSig) =
+  let opt_validator_index =
+      findValidatorIndex(state.builders.asSeq, bucket_sorted_builders, pubkey)
```
tersec: Indent 2 fewer spaces

tersec:
```
Build completed successfully: build/consensus_spec_tests_minimal
/github-runner/github-runner-node-03/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim: In function ‘_ZN7runTest21runTestX60gensym7504_E6string6string’:
/github-runner/github-runner-node-03/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim:1112:15: error: stack usage is 5664944 bytes [-Werror=stack-usage=]
 1112 |   proc runTest(suiteName, testName: string): TestStatus {.raises: [], gensym.} =
      |               ^
lto1: some warnings being treated as errors
make[2]: *** [/tmp/ccuqLJRV.mk:269: /tmp/ccBd0vuY.ltrans89.ltrans.o] Error 1
make[2]: *** Waiting for unfinished jobs....
lto-wrapper: fatal error: make returned 2 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
make[1]: *** [nimcache/release/consensus_spec_tests_mainnet/consensus_spec_tests_mainnet.makefile:1300: build] Error 1
make: *** [Makefile:310: consensus_spec_tests_mainnet] Error 2
```
is why the GitHub Action Linux CIs are failing.

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -314,12 +318,12 @@ suite baseDescription & "Voluntary Exit " & preset():
...
-      executionPayload: deneb.ExecutionPayload): Result[void, cstring] =
+  proc applyWithdrawals(
+      preState: var gloas.BeaconState, _: gloas.BeaconState):
+      Result[void, cstring] =
     process_withdrawals(preState)
 
   for path in walkTests(OpWithdrawalsDir):
-    runTest[deneb.ExecutionPayload, typeof applyWithdrawals](
-      OpWithdrawalsDir, suiteName, "Withdrawals", "execution_payload",
+    runTest[gloas.BeaconState, typeof applyWithdrawals](
```
tersec: This change is the cause of
```
Build completed successfully: build/consensus_spec_tests_minimal
/github-runner/github-runner-node-03/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim: In function ‘_ZN7runTest21runTestX60gensym7504_E6string6string’:
/github-runner/github-runner-node-03/workspace/nimbus-eth2/nimbus-eth2/vendor/nim-unittest2/unittest2.nim:1112:15: error: stack usage is 5664944 bytes [-Werror=stack-usage=]
 1112 |   proc runTest(suiteName, testName: string): TestStatus {.raises: [], gensym.} =
      |               ^
lto1: some warnings being treated as errors
make[2]: *** [/tmp/ccuqLJRV.mk:269: /tmp/ccBd0vuY.ltrans89.ltrans.o] Error 1
make[2]: *** Waiting for unfinished jobs....
lto-wrapper: fatal error: make returned 2 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
make[1]: *** [nimcache/release/consensus_spec_tests_mainnet/consensus_spec_tests_mainnet.makefile:1300: build] Error 1
make: *** [Makefile:310: consensus_spec_tests_mainnet] Error 2
```
in GitHub Actions because it provides the `T` type parameter in
https://github.com/status-im/nimbus-eth2/blob/6d509e09546cb3bca61b529288f3e85191982b2f/tests/consensus_spec/gloas/test_fixture_operations.nim#L60-L75
which due to `parseTest` in
https://github.com/status-im/nimbus-eth2/blob/6d509e09546cb3bca61b529288f3e85191982b2f/tests/consensus_spec/gloas/test_fixture_operations.nim#L74-L75
results in an attempt to put an object of type `T` (i.e. `gloas.BeaconState` here) on the stack, which is detected by
https://github.com/status-im/nimbus-eth2/blob/6d509e09546cb3bca61b529288f3e85191982b2f/config.nims#L64-L70

A more correct approach, which also would not result in this stack usage, would be to more precisely model that unlike the other operations tests, Gloas withdrawal tests have no separate change to apply, exactly because they depend only on the `gloas.BeaconState`. So `runTest`'s current interface assumes too much for Gloas operations tests, and there should be a variation which performs a test but doesn't run `applyProc` in https://github.com/status-im/nimbus-eth2/blob/6d509e09546cb3bca61b529288f3e85191982b2f/tests/consensus_spec/gloas/test_fixture_operations.nim#L74-L75.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2138,6 +2152,138 @@ func translate_participation(
...
+    # apply_deposit_for_builder can mutate the state and may add a builder
+    # to the registry.
+    let
+      is_existing_builder = findValidatorIndex(
+        state.builders.asSeq, bucket_sorted_builders[], deposit.pubkey).isSome
+      has_builder_credentials = 
+        is_builder_withdrawal_credential(deposit.withdrawal_credentials)
+
+    if is_existing_builder or (
+        has_builder_credentials and not is_pending_validator):
```
tersec: `is_pending_validator` doesn't seem to be in https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/fork.md#new-onboard_builders_from_pending_deposits

It presumably comes from
- https://github.com/ethereum/consensus-specs/pull/4897

and solves a real issue in the specs. It's just not part of alpha.2 and it's not separately listed in https://notes.ethereum.org/@ethpandaops/epbs-devnet-0 either.

So as-is, it might pass the test vectors because they don't test this particular scenario (precisely because it hadn't been realized), but if someone does trigger it, intentionally or otherwise, on `epbs-devnet-0`, Nimbus would diverge because it wouldn't be implementing the specified specs.

Tomi-3-0: The issue is without this fix the  test `fork_validator_deposit_followed_by_builder_credentials` fails  but passes with it.
If I implement exactly alpha.2 without `is_pending_validator` the test fails. The alternative might be to deactivate this particular test case for now and wait until the new fix is merged and released

tersec: Oh, that's interesting, that's worth interrogating with them: is the mismatch in their setup or ours? i.e. is it a bug-working-around-another bug in Nimbus's implementation or are they shipping a spec which doesn't correspond to the test vectors nominally for that spec?

tersec: > The alternative might be to deactivate this particular test case for now and wait until the new fix is merged and released

This would be potentially very reasonable, yes.

Tomi-3-0: the mismatch was in my implementation. I was checking `is_existing_validator` against validator list at the start, but it was not been updated as it processed valid validator deposits. So when the second deposit (with builder credentials) came for the same pubkey, it didn't know that pubkey had already been claimed by a pending validator deposit.

this fixes it:
```
var pending_validator_pubkeys: HashSet[ValidatorPubKey]

for deposit in state.pending_deposits:
  let is_existing_validator = findValidatorIndex(
    state.validators.asSeq, bucket_sorted_validators[], deposit.pubkey).isSome
  
  if is_existing_validator or deposit.pubkey in pending_validator_pubkeys:
    pending_deposits.add(deposit)
    continue
  
  .....
  
  if verify_deposit_signature(...):
    pending_validator_pubkeys.incl(deposit.pubkey)  # We Only add new ones here
    pending_deposits.add(deposit)
```

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/commits/578d746e869c762daa06197d23fb882277521139

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -314,12 +314,37 @@ suite baseDescription & "Voluntary Exit " & preset():
       applyVoluntaryExit, path)
 
 suite baseDescription & "Withdrawals " & preset():
-  func applyWithdrawals(
-      preState: var gloas.BeaconState,
-      executionPayload: deneb.ExecutionPayload): Result[void, cstring] =
+  proc applyWithdrawals(
+      preState: var gloas.BeaconState, _: gloas.BeaconState):
```
tersec: Why have this ignored (`_`) parameter at all?

It ends up getting called with `done = applyWithdrawals(preState[], preState[])`, but there's just no reason for second parameter.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -470,6 +466,31 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: fulu.DataColumnSidecar |
...
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#modified-verify_data_column_sidecar_kzg_proofs
+proc verify_data_column_sidecar_kzg_proofs*(sidecar: gloas.DataColumnSidecar,
+                                            kzg_commitments: KzgCommitments):
+                                            Result[void, cstring] =
+  ## Verify if the KZG proofs are correct.
+
+  # The column index also represents the cell index
+  var cellIndices = newSeqOfCap[CellIndex](sidecar.column.len)
+  for _ in 0..<sidecar.column.len:
+    cellIndices.add(CellIndex(sidecar.index))
```
tersec: [`repeat()`](https://nim-lang.org/docs/sequtils.html#repeat%2CT%2CNatural) already does this.

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -314,12 +314,37 @@ suite baseDescription & "Voluntary Exit " & preset():
...
+    test prefix & baseDescription & "Withdrawals - " & path:
+      let testDir = OpWithdrawalsDir / "pyspec_tests" / path
+      let preState = newClone(
+        parseTest(testDir/"pre.ssz_snappy", SSZ, gloas.BeaconState))
+      
+      var done: Result[void, cstring]
+      try:
+        done = applyWithdrawals(preState[], preState[])
+      except Defect:
+        done = Result[void, cstring].err("Index out of bounds")
```
tersec: From
https://github.com/status-im/nimbus-eth2/blob/578d746e869c762daa06197d23fb882277521139/beacon_chain/spec/state_transition_block.nim#L1493-L1519
it doesn't look like `Index out of bounds` would be the most likely cause of `Defect`s, so it's assuming quite a bit to use that as the error string.

But more generally, it's probably not worth doing this `try`/`except` at all, because https://github.com/status-im/nim-unittest2/blob/1efa65bb037185f0983ac4fa65932e8d3450825d/unittest2.nim#L1117-L1154 already handles exceptions. It would just be a different reason for the test to fail, and `nim-unittest2` would be able to show the whole exception, rather than an inevitably incomplete attempt to interpret it.

Furthermore, to the extent that one is going to specialize this for withdrawals, there's not even really a need for an `applyWithdrawals()` wrapper function; just call `process_withdrawals` directly.

beacon_chain/spec/beaconstate.nim
tersec: The logic in https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/fork.md#new-onboard_builders_from_pending_deposits doesn't seem to need any `pending_validator_pubkeys` though?

So the question remains, if the executable spec doesn't need this to pass the reference tests, where is the inconsistency coming from?

tests/consensus_spec/gloas/test_fixture_operations.nim
Tomi-3-0: The Python spec crashes with IndexError on some invalid test cases that fails without the try/except, e.g:
https://github.com/ethereum/consensus-specs/blob/2938e1ad74cea54f1a24508a85704d5bd87837ad/tests/core/pyspec/eth2spec/test/gloas/block_processing/test_process_withdrawals.py#L1286
https://github.com/ethereum/consensus-specs/blob/2938e1ad74cea54f1a24508a85704d5bd87837ad/tests/core/pyspec/eth2spec/test/gloas/block_processing/test_process_withdrawals.py#L1319

adding bounds checking at the start of process_withdrawals and in apply_withdrawals seems like a better approach.

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/bf239f83d1bd1119d5f7504039b3e81f49ccdec1

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/changes/bf239f83d1bd1119d5f7504039b3e81f49ccdec1

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: Indeed, three tests fail without the safeguards:
- [`invalid_validator_index_pending_partial`](https://github.com/ethereum/consensus-specs/blob/2938e1ad74cea54f1a24508a85704d5bd87837ad/tests/core/pyspec/eth2spec/test/gloas/block_processing/test_process_withdrawals.py#L1348-L1376) in [`get_pending_partial_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#new-get_pending_partial_withdrawals) because it tries to access invalid validator index 256 when only 256 validators exist
- [ `invalid_builder_index_sweep`](https://github.com/ethereum/consensus-specs/blob/2938e1ad74cea54f1a24508a85704d5bd87837ad/tests/core/pyspec/eth2spec/test/gloas/block_processing/test_process_withdrawals.py#L1317-L1345) in [`get_builders_sweep_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#new-get_builders_sweep_withdrawals) due to trying to access an invalid builder index 8 when only 8 builder indices exist; and
- [`invalid_validator_index_sweep`](https://github.com/ethereum/consensus-specs/blob/2938e1ad74cea54f1a24508a85704d5bd87837ad/tests/core/pyspec/eth2spec/test/gloas/block_processing/test_process_withdrawals.py#L1379-L1407) in [`get_validators_sweep_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#modified-get_validators_sweep_withdrawals) on validator index 256 with 256 validators

`get_pending_partial_withdrawals()` straightforwardly is the first part of Gloas [`get_expected_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#modified-get_expected_withdrawals) which touches that part of the state, which means that the corruption must have existed, in the attack scenario, before `process_withdrawals()` was called.

The question becomes, how would this have been achieved, when [`process_withdrawal_request()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#new-process_withdrawal_request) is the only way to add an entry to `state.pending_partial_withdrawals`. And
```python
    index = ValidatorIndex(validator_pubkeys.index(request_pubkey))
    validator = state.validators[index]

    # ...

    # Only allow partial withdrawals with compounding withdrawal credentials
    if (
        has_compounding_withdrawal_credential(validator)
        and has_sufficient_effective_balance
        and has_excess_balance
    ):
        to_withdraw = min(
            state.balances[index] - MIN_ACTIVATION_BALANCE - pending_balance_to_withdraw, amount
        )
        exit_queue_epoch = compute_exit_epoch_and_update_churn(state, to_withdraw)
        withdrawable_epoch = Epoch(exit_queue_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY)
        state.pending_partial_withdrawals.append(
            PendingPartialWithdrawal(
                validator_index=index,
                amount=to_withdraw,
                withdrawable_epoch=withdrawable_epoch,
            )
        )
```

I don't see how this scenario in `invalid_validator_index_pending_partial` can occur, because `state.validators` can only grow.

For the second case, [`get_builders_sweep_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#new-get_builders_sweep_withdrawals), it results from
```python
    builder_index = state.next_withdrawal_builder_index
    for _ in range(builders_limit):
        all_withdrawals = prior_withdrawals + withdrawals
        has_reached_limit = len(all_withdrawals) >= withdrawals_limit
        if has_reached_limit:
            break

        builder = state.builders[builder_index]
        if builder.withdrawable_epoch <= epoch and builder.balance > 0:
            withdrawals.append(
                Withdrawal(
                    index=withdrawal_index,
                    validator_index=convert_builder_index_to_validator_index(builder_index),
                    address=builder.execution_address,
                    amount=builder.balance,
                )
            )
            withdrawal_index += WithdrawalIndex(1)

        builder_index = BuilderIndex((builder_index + 1) % len(state.builders))
```
i.e. `state.next_withdrawal_builder_index` needs to be corrupt on entering the function, because otherwise after the first loop iteration, `builder_index = BuilderIndex((builder_index + 1) % len(state.builders))`. Only [`update_next_withdrawal_builder_index()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#new-update_next_withdrawal_builder_index) updates `state.next_withdrawal_builder_index`, and it disallows this case:
```python
def update_next_withdrawal_builder_index(
    state: BeaconState, processed_builders_sweep_count: uint64
) -> None:
    if len(state.builders) > 0:
        # Update the next builder index to start the next withdrawal sweep
        next_index = state.next_withdrawal_builder_index + processed_builders_sweep_count
        next_builder_index = BuilderIndex(next_index % len(state.builders))
        state.next_withdrawal_builder_index = next_builder_index
```
So since, like `state.validators`, `state.builders` cannot shrink between `update_next_withdrawal_builder_index()` and `get_builders_sweep_withdrawals()`, it's not possible to hit this case.

Finally, the third failing test, [`get_validators_sweep_withdrawals()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/electra/beacon-chain.md#modified-get_validators_sweep_withdrawals) involves
```python
    validator_index = state.next_withdrawal_validator_index
    for _ in range(validators_limit):
        all_withdrawals = prior_withdrawals + withdrawals
        has_reached_limit = len(all_withdrawals) >= withdrawals_limit
        if has_reached_limit:
            break

        validator = state.validators[validator_index]
```
so one needs to track the provenance of `state.next_withdrawal_validator_index`. This is solely done by [`update_next_withdrawal_validator_index()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/capella/beacon-chain.md#new-update_next_withdrawal_validator_index):
```python
def update_next_withdrawal_validator_index(
    state: BeaconState, withdrawals: Sequence[Withdrawal]
) -> None:
    # Update the next validator index to start the next withdrawal sweep
    if len(withdrawals) == MAX_WITHDRAWALS_PER_PAYLOAD:
        # Next sweep starts after the latest withdrawal's validator index
        next_validator_index = ValidatorIndex(
            (withdrawals[-1].validator_index + 1) % len(state.validators)
        )
        state.next_withdrawal_validator_index = next_validator_index
    else:
        # Advance sweep by the max length of the sweep if there was not a full set of withdrawals
        next_index = state.next_withdrawal_validator_index + MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP
        next_validator_index = ValidatorIndex(next_index % len(state.validators))
        state.next_withdrawal_validator_index = next_validator_index
```
where all assignments to `state.next_withdrawal_validator_index` involve `% len(state.validators)`. As with the other two cases, this cannot actually happen without having deliberately constructed a corrupt `BeaconState`.

I'm unconvinced those tests make sense, because they assume a corrupted `BeaconState`. The question is, is the corruption accessible, and if so, should it be, i.e. are the engineering tradeoffs which allow it overall beneficial. All three require either disabling `validate_validator_indices` or `validate_builder_indices` to "bypass helper validation".

More broadly, if the state transition function should have a goal of handling arbitrarily corrupted `BeaconState`s validly without crashing, that would be, if questionable, worth articulating, but generally speaking it has not, and these ad-hoc tests aren't otherwise well-motivated.

tersec: @Tomi-3-0 feel free to add skips/ignores for all three tests. They're not worth accommodating.

Tomi-3-0: okay

beacon_chain/spec/beaconstate.nim
tersec: Looked into this more, and yeah, it's a fairly reasonable way of addressing the issue here.

tersec:
Can rerun `make test` to regenerate summaries to avoid:
```

[2026-02-07T18:37:21.665Z] + git diff --exit-code --ignore-submodules=all
[2026-02-07T18:37:21.665Z] diff --git a/ConsensusSpecPreset-mainnet.md b/ConsensusSpecPreset-mainnet.md
[2026-02-07T18:37:21.665Z] index 04052c145..3f0514e42 100644
[2026-02-07T18:37:21.665Z] --- a/ConsensusSpecPreset-mainnet.md
[2026-02-07T18:37:21.665Z] +++ b/ConsensusSpecPreset-mainnet.md
[2026-02-07T18:37:21.665Z] @@ -4106,10 +4106,6 @@ ConsensusSpecPreset-mainnet
[2026-02-07T18:37:21.665Z]  ```
[2026-02-07T18:37:21.665Z]  ## EF - Gloas - Operations - Withdrawals  [Preset: mainnet]
[2026-02-07T18:37:21.665Z]  ```diff
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_builder_index_pending            OK
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_builder_index_sweep              OK
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_validator_index_pending_partial  OK
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_validator_index_sweep            OK
[2026-02-07T18:37:21.665Z]  + [Valid]   EF - Gloas - Operations - Withdrawals - all_builder_withdrawals_zero_balance     OK
[2026-02-07T18:37:21.665Z]  + [Valid]   EF - Gloas - Operations - Withdrawals - builder_and_pending_leave_room_for_sweep OK
[2026-02-07T18:37:21.665Z]  + [Valid]   EF - Gloas - Operations - Withdrawals - builder_max_minus_one_plus_one_regular   OK
[2026-02-07T18:37:21.665Z] diff --git a/ConsensusSpecPreset-minimal.md b/ConsensusSpecPreset-minimal.md
[2026-02-07T18:37:21.665Z] index 1bf2ca8d2..53fbfdf01 100644
[2026-02-07T18:37:21.665Z] --- a/ConsensusSpecPreset-minimal.md
[2026-02-07T18:37:21.665Z] +++ b/ConsensusSpecPreset-minimal.md
[2026-02-07T18:37:21.665Z] @@ -4368,10 +4368,6 @@ ConsensusSpecPreset-minimal
[2026-02-07T18:37:21.665Z]  ```
[2026-02-07T18:37:21.665Z]  ## EF - Gloas - Operations - Withdrawals  [Preset: minimal]
[2026-02-07T18:37:21.665Z]  ```diff
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_builder_index_pending            OK
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_builder_index_sweep              OK
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_validator_index_pending_partial  OK
[2026-02-07T18:37:21.665Z] -+ [Invalid] EF - Gloas - Operations - Withdrawals - invalid_validator_index_sweep            OK
[2026-02-07T18:37:21.665Z]  + [Valid]   EF - Gloas - Operations - Withdrawals - all_builder_withdrawals_zero_balance     OK
[2026-02-07T18:37:21.665Z]  + [Valid]   EF - Gloas - Operations - Withdrawals - builder_and_pending_leave_room_for_sweep OK
[2026-02-07T18:37:21.665Z]  + [Valid]   EF - Gloas - Operations - Withdrawals - builder_max_minus_one_plus_one_regular   OK
```
from https://ci.status.im/blue/organizations/jenkins/nimbus-eth2%2Fplatforms%2Flinux%2Fx86_64%2Fmain/detail/PR-7926/15/pipeline

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7926/commits/46b4db4b49fa6bc97f4cd80e7c357fe598c88881

PR #7929, opened 2026-02-04
Add on_slot_start_after_past_attestations_applied hook

etan-status:
Fast Confirmation Rule needs to re-check that the confirmed block is still confirmed at the start of a new epoch, after unrealized justifications / finalizations are realized, and after processing any deferred attestations from the queue. Add this hook, while retaining current safe block behaviour.


etan-status:
https://github.com/ethereum/consensus-specs/pull/4747/files#r2767755610

May get away using the existing hook for that. Will try and simplify.

PR #7937, opened 2026-02-05
Track fork choice votes by slot instead of epoch

etan-status:
FCR needs access to the latest vote of each validator, to estimate the share of votes coming from empty/non-canonical slots, vs canonical ones. We can simply track the latest slot, then we don't need access to a historical shuffling, as the slot implies that they were part of a committee that was allowed to contribute during that slot.

### DIFF
--- a/beacon_chain/fork_choice/fork_choice.nim
+++ b/beacon_chain/fork_choice/fork_choice.nim
@@ -651,14 +651,14 @@ when isMainModule:
     votes.add VoteTracker(
       current_root: fakeHash(1),
       next_root: default(Eth2Digest),
-      slot: Epoch(0))
+      slot: Slot(0))
 
     # One validator moves their vote from the block to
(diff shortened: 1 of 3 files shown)


tersec:
```
[2026-02-05T11:23:11.476Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7937/beacon_chain/fork_choice/fork_choice.nim(654, 11) Error: type mismatch: got 'Epoch' for 'Epoch(0)' but expected 'Slot = distinct uint64'
[2026-02-05T11:23:11.476Z] make: *** [Makefile:335: fork_choice] Error 1
[2026-02-05T11:23:11.476Z] make: *** Waiting for unfinished jobs....
```

etan-status:
Ah gotcha, was one of the isMainModule tests, so didn't catch it.

PR #7938, opened 2026-02-05
Switch to the intops-powered version of stint.

moigagoo:
Similarly to https://github.com/status-im/nimbus-eth1/pull/3956, this PR introduces a new dependency ([intops](https://github.com/vacp2p/nim-intops/)) and updates stint to point to the forked version that uses intops internally.

Since nimbus-eth2 doesn't use bncurve, only stint is affected.

Whether the benchmarks should be verified and how to do that are open questions but the tests to pass locally and so far no performance regression was observed in benchmarks for nimbus-eth1, stint, or bncurve.


tersec:
Looks reasonable. `nimbus-eth2` doesn't use `nim-stint` for anything that performance-sensitive, unlike `nimbus-eth1`, and e.g. doesn't depend on the slowest features (modular exponentiation, et cetera) at all.

PR #7949, opened 2026-02-09
cache sidecar signatures

agnxsh:
fast path for verifying block signatures, in too many sidecars

### DIFF
--- a/beacon_chain/consensus_object_pools/block_clearance.nim
+++ b/beacon_chain/consensus_object_pools/block_clearance.nim
@@ -34,7 +34,7 @@ proc verifyBlockProposer*(
     slot: Slot,
     proposer_index: uint64,
     blockRoot: Eth2Digest,
-    signature: ValidatorSig,
+    signature: ValidatorSig
 ): Result[void, tuple[msg: cstring, invalid: bool]] =
   ## Verify block proposer and signature, making sure to check that the proposer
   ## was indeed elected for the given slot and that the signature checks out.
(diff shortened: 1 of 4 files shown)


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -51,6 +85,10 @@ proc verifyBlockProposer*(
   if uint64(proposer) != proposer_index:
     return err(("verifyBlockProposer: unexpected proposer", true))
 
+  # If signature matches the latest valid signature, skip verification
+  if signature == latestValidSig:
```
tersec: This needs to verify it's a correct signature for this specific `proposer_index`.  That is, if verifyBlockProposer` is initially intended to be for block `B_1` with signature `S_1`, and there's a valid block `B_2` which should have signature `S_2`, but instead it's called on `B_2` with signature `S_1`, it should properly reject it.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -28,13 +28,47 @@ export results, signatures_batch, block_dag, blockchain_dag
...
+  ## Because the signature only covers the block body (via its root), it's
+  ## possible that the block itself is valid while the signature is not: in
+  ## this case false is returned for "invalid".
+  let proposer = dag.getProposer(parent, slot).valueOr:
+    warn "cannot compute proposer for block", parent, slot, proposer_index, blockRoot
+    return err(("verifyBlockProposer: cannot compute proposer", false)) # internal issue
+
+  # `getProposer` returns a trusted proposer index, while `proposer_index` may
+  # be invalid -> convert the former to the latter
+  if uint64(proposer) != proposer_index:
```
tersec: Using `distinctBase` instead of `uint64` here statically guarantees the result is typesafe.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -28,13 +28,47 @@ export results, signatures_batch, block_dag, blockchain_dag
 logScope:
   topics = "clearance"
 
+proc verifyBlockProposer*(
```
tersec: This entire function is almost, but not quite, identical to the `RecentSidecarSignatureLru`. Are both useful? Right now it has to be here because none of the 4 `beacon_chain` callsites have been adjusted, but presumably using this one is the slow path, and I'm not sure why one would want that.

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7949/commits/dd9f226c8086d826db4b4bbac3e0b9a3b87afb09

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7949/changes/dc07ada784de01246459cd5eb6aeb4883c41ffb5

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7949/changes/dc07ada784de01246459cd5eb6aeb4883c41ffb5

PR #7952, opened 2026-02-10
chronos: v4.2.0

arnetheduck:
https://github.com/status-im/nim-chronos/releases/tag/v4.2.0

* The continous mode for the rate limiter means requests will be served more evenly instead of in bursts, when others are syncing from the node


tersec:
Not going to merge until after the next release comes out.

PR #7953, opened 2026-02-10
Request missing envelopes

ahshum:


beacon_chain/sync/request_manager.nim
```diff
@@ -137,6 +155,22 @@ func checkResponse(roots: openArray[Eth2Digest],
...
 
+func checkResponse(
+    roots: openArray[Eth2Digest],
+    envelopes: openArray[ref SignedExecutionPayloadEnvelope],
+): bool =
+  ## Ensure there is the requested envelope as per each root.
+  var checks = @roots
+  if len(envelopes) > len(roots):
+    return false
+  for envelope in envelopes:
```
agnxsh: would it be useful to make `envelopes` a HashSet? how frequently can rman for enevelopes get triggered for missing envelopes? also mostly depends on the max missing envelopes

beacon_chain/sync/request_manager.nim
```diff
@@ -742,6 +893,7 @@ proc requestManagerDataColumnLoop(
 proc start*(rman: var RequestManager) =
   ## Start Request Manager's loops.
   rman.blockLoopFuture = rman.requestManagerBlockLoop()
+  rman.envelopeLoopFuture = rman.requestManagerEnvelopeLoop()
```
agnxsh: This needs to start only at >= `GLOAS_FORK_EPOCH`, several ways of doing this, easiest is to `getBeaconTime`

beacon_chain/sync/request_manager.nim
```diff
@@ -505,6 +600,61 @@ proc requestManagerBlockLoop(
...
+
+    var workers:
+      array[PARALLEL_REQUESTS, Future[void].Raising([CancelledError])]
+
+    for i in 0 ..< PARALLEL_REQUESTS:
+      workers[i] = self.fetchEnvelopesFromNetwork(blockRoots)
+
+    await allFutures(workers)
+
+    let finish = SyncMoment.now(uint64(len(blockRoots)))
```
tersec: `uint64(len(x)` is `lenu64(x)`

beacon_chain/sync/request_manager.nim
```diff
@@ -505,6 +600,61 @@ proc requestManagerBlockLoop(
...
+          blockRoots.add blockRoot
+          continue
+        debug "Loaded orphaned envelope from storage", blockRoot
+        verifiers.add self.envelopeVerifier(envelope.asSigned())
+      try:
+        await allFutures(verifiers)
+      except CancelledError as exc:
+        var futs = newSeqOfCap[Future[void].Raising([])](verifiers.len)
+        for verifier in verifiers:
+          futs.add verifier.cancelAndWait()
```
tersec: These three lines can be replaced by `let futs = verifiers.mapIt(it.cancelAndWait())`

beacon_chain/sync/request_manager.nim
```diff
@@ -753,10 +903,20 @@ proc switchToColumnLoop*(rman: var RequestManager) =
     rman.dataColumnLoopFuture =
       rman.requestManagerDataColumnLoop()
 
+proc switchToEnvelopeLoop*(self: var RequestManager) =
```
tersec: `switchToColumnLoop` is named as such because it's switching from a blob loop. But what is this switching from? It's not really switching at all.

agnxsh: exactly, it's not really a _switch_, you can just name it as `startEnvelopeLoop`, or just another overload of `start`

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/4cffcbdfa66a8896a71480f4df2ef62dae4c1e58

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/4cffcbdfa66a8896a71480f4df2ef62dae4c1e58

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/9c17235230a9f8e91a9c1f5728b36e3f00ea02c6

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/4cffcbdfa66a8896a71480f4df2ef62dae4c1e58 sort of

beacon_chain/sync/request_manager.nim
```diff
@@ -137,6 +155,21 @@ func checkResponse(roots: openArray[Eth2Digest],
...
+  ## Ensure there is the requested envelope as per each root.
+  if len(envelopes) > len(roots):
+    return false
+  var checks = roots.toHashSet()
+  for envelope in envelopes:
+    if envelope[].message.beacon_block_root in checks:
+      checks.excl(envelope[].message.beacon_block_root)
+    else:
+      return false
+  true
```
```nim
  var checks = roots.toHashSet()
  for envelope in envelopes:
    if envelope[].message.beacon_block_root in checks:
      checks.excl(envelope[].message.beacon_block_root)
    else:
      return false
  true
```
can be replaced (with caveats on duplicate items; see below) with
```nim
  let checks = roots.toHashSet()
  envelopes.allIt(it[].message.beacon_block_root in checks)
```
Since it's O(1) access anyway, there's not much point in removing items along the way.

Unless one wants to catch duplicate `envelope`s? But I'm not sure that'd be correct anyway, to say the second copy of an envelope isn't actually in `root`s, or at least it's unintuitive that a duplication check would live as as side-effect here. Is that the intent?

beacon_chain/sync/request_manager.nim
tersec: Only one of `envelope`s and `checks`/`root`s needs to be a `HashSet` to get O(1); see https://github.com/status-im/nimbus-eth2/pull/7953/changes#r2810831624

beacon_chain/sync/request_manager.nim
```diff
@@ -260,6 +293,68 @@ proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async:
...
+    peer = await self.network.peerPool.acquire()
+    debug "Requesting envelopes by root",
+      peer = peer, envelopes = shortLog(roots),
+      peer_score = peer.getScore()
+
+    let envelopes = await executionPayloadEnvelopesByRoot(
+      peer, BlockRootsList roots)
+
+    if envelopes.isOk:
+      var uenvelopes = envelopes.get().asSeq()
```
tersec: This doesn't look like it needs to be `var`, since it's never mutated?

beacon_chain/sync/request_manager.nim
```diff
@@ -260,6 +293,68 @@ proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async:
...
+          let res = await self.envelopeVerifier(envelope[])
+
+          # Envelope is marked as missing when we got a valid block. As in
+          # Gloas, both valid block and envelope are required in order to
+          # proceed to the next slot. So in theory we should only be able to
+          # notice at most one missing envelope per slot.
+          #
+          # If there is a good way to find missing envelopes other than the
+          # head, the response, which may contains 2 or more envelopes, may
+          # cause verifier error as the order matters.
```
tersec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#execution_payload
> - _[IGNORE]_ The node has not seen another valid `SignedExecutionPayloadEnvelope` for this block root from this builder.

i.e.:
- already a fairly arbitrary order-dependence for a builder, and it's on them if one builder sends multiple envelopes for one block root. Both should be valid if they're going to play that game; and

I'm not certain if the same reasoning applies across builders though.

So yes, order matters, but verifier error?

beacon_chain/sync/request_manager.nim
```diff
@@ -260,6 +293,68 @@ proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async:
...
+          #
+          # TODO improve/investigate way of figuring out missing envelope
+          #      efficiently (across different slots).
+          #
+          # TODO verify multiple envelopes in the chain's order.
```
tersec: On chain, only one can be valid ultimately per block root.

beacon_chain/sync/request_manager.nim
```diff
@@ -505,6 +600,59 @@ proc requestManagerBlockLoop(
...
+    {.async: (raises: [CancelledError]).} =
+  while true:
+    # TODO This polling could be replaced with an AsyncEvent that is fired
+    #      from the quarantine when there's work to do
+    await sleepAsync(POLL_INTERVAL)
+
+    if self.inhibit():
+      continue
+
+    let missingBlockRoots = self.envelopeQuarantine[].peekMissing().toSeq()
```
tersec: A couple points.
- `peekMissing` is defined as iterating over `self.envelopeQuarantine[].missing`, which this then reconstitutes a `HashSet` into a `seq`. And this a reasonably fine way to do that as far as it goes, and currently that copy would be necessary anyway when `self.envelopeLoader == nil`, but otherwise, it's an extra copy/allocation/etc because
```nim
      for blockRoot in missingBlockRoots:
        let envelope = self.envelopeLoader(blockRoot).valueOr:
          blockRoots.add blockRoot
          continue
```
only iterates and doesn't need access to a reified data structure.

My question is, to accommodate this flexibility, what is the use case of `self.envelopeLoader == nil`, which is effectively as a special case where `self.envelopeLoader` hasn't been defined and asserts that no envelopes are currently loaded? Because it's structured as special case in the code, it creates these additional complications.

Alternatively, one could align `blockRoots` to also be a `HashSet`, to allow `blockRoots = missingBlockRoots` to work. For the most part, this works fine already, except that `fetchEnvelopesFromNetwork(blockRoots)` needs to handle a `HashSet`, which would imply that `executionPayloadEnvelopesByRoot()` needs to take a `HashSet`.

Other approaches exist as well.

Either way, it's nicer to consider the types holistically in these situations, to avoid pointless conversions back and forth between `seq[Eth2Digest]` and `HashSet[Eth2Digest]`.

beacon_chain/sync/request_manager.nim
ahshum: Yes that is true for one block root, but I think with the spec, we can only have at most one _missing_ envelope at the same time, due to we need the valid block to mark an envelope as missing.

So it sounds like _multiple missing_ envelopes is for backfilling which should be out of scope here?

beacon_chain/sync/request_manager.nim
ahshum: Yes I think the response from peer may not be trustworthy so there can be duplicate envelopes. This is also what I get from the request logic of blocks:

```nim
func checkResponse(roots: openArray[Eth2Digest],
                   blocks: openArray[ref ForkedSignedBeaconBlock]): bool =
  ## This procedure checks peer's response.
  var checks = @roots
  if len(blocks) > len(roots):
    return false
  for blk in blocks:
    let res = checks.find(blk[].root)
    if res == -1:
      return false
    else:
      checks.del(res)
  true
```

If we think it is not worth or does not need to catch the duplicates, I am happy to modify to use `allIt` which is much cleaner and simpler.

beacon_chain/sync/request_manager.nim
ahshum: Thanks for the comment.

I thought `self.envelopeLoader == nil` is a common practice as similar logic exists in block and blob, etc. so I am not sure in what situation it is useful as the `rman.*Loader()` are always defined.

I think `EnvelopeQuarantine.missing` can be modified into `seq[Eth2Digest]` as if the comment above stays true (at most one missing envelope at a time), the cost of `addMissing` or `delMissing`, with checking of `if root in missing` should be negligible.

beacon_chain/sync/request_manager.nim
ahshum: Will clean it up if we have a conclusion from the discussion above.

beacon_chain/sync/request_manager.nim
tersec: Yeah, I'd agree with that. Or forward syncing, but that's also another issue, and that flow doesn't intrinsically/ideally involve request manager.

beacon_chain/sync/request_manager.nim
tersec: Indeed, the peer's responses are untrustworthy, so they could send duplicate envelopes. The question is, are those in/out of spec and how should they be handled.

Nothing in https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#executionpayloadenvelopesbyroot-v1 prohibits duplicate envelopes. The one requirement is:
> The response is a list of `SignedExecutionPayloadEnvelope` whose length is less than or equal to the number of requested execution payload envelopes. It may be less in the case that the responding peer is missing payload envelopes.

So it would be legal to request [A, B, C] and get [B, B] back. Useless, but legal. `checkResponse` should therefore not reject this response.

beacon_chain/sync/request_manager.nim
ahshum: The issue that I raised would only be valid when requesting 2 or more envelopes, that all the envelopes in the response could be valid, there could still be `VerifierError.invalid` due to the order how we apply to the chain.

I will clean it up if this is not a concern, or needs any improvement.

beacon_chain/sync/request_manager.nim
ahshum: Does it mean the current implementation of, e.g. missing block request, is incorrect? Peer score would be updated by `PeerScoreBadResponse` if there is duplicates or any mismatches.

beacon_chain/sync/request_manager.nim
ahshum: For clarity, these functions are refactored in https://github.com/status-im/nimbus-eth2/pull/7953/changes/2f7f63c33000906afe674478ad0874b65a9f69e0

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/9796ab38c5f502053f61775ad28cba26a0b4116e

beacon_chain/sync/request_manager.nim
tersec: > Does it mean the current implementation of, e.g. missing block request, is incorrect? Peer score would be updated by `PeerScoreBadResponse` if there is duplicates or any mismatches.

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/phase0/p2p-interface.md#beaconblocksbyroot-v1 doesn't seem to explicitly prohibit duplicate blocks either (yes, it's the v1 endpoint, and yes, it's deprecated in favor of the v2 endpoint in Altair and newer which specifies the forks. But the newer forks don't seem to change its core semantics, so that's where the documentation seems to live.

So similarly yeah, it doesn't specify anything about duplicates in the specs, so I think that code for blocks might be wrong. I wouldn't adjust it in this PR, no, but it's worth checking/looking at. At least it explains where you got your implementation approach from, indeed.

In general, though, `checkResponse` is binary, it's either a valid response or not. Scoring can potentially be more fine-grained, e.g., valid but useless, etc.

ahshum: Yes sure it is out of scope for this PR but good to confirm the misalignment between the implementation and specs.

beacon_chain/consensus_object_pools/envelope_quarantine.nim
```diff
@@ -31,7 +31,11 @@ template root(v: SignedExecutionPayloadEnvelope): Eth2Digest =
 func addMissing*(
     self: var EnvelopeQuarantine,
     root: Eth2Digest) =
-  self.missing.incl(root)
+  if root notin self.missing:
+    self.missing.add(root)
```
tersec: If one's going to use a `seq`, there's already a [stdlib function](https://nim-lang.org/docs/sequtils.html#addUnique%2Cseq%5BT%5D%2CsinkT) for this.

tersec:

beacon_chain/consensus_object_pools/envelope_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/commits/19ca8d7a4df03a28804d9783f867704286bc6573

beacon_chain/consensus_object_pools/envelope_quarantine.nim
tersec, line 14: Can add
```nim
from std/sequtils import addUnique
```
between the existing `import` section and the `type` section to fix
```
[2026-02-20T15:38:48.309Z] Building: build/mev_mock
[2026-02-20T15:39:49.743Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7953/beacon_chain/consensus_object_pools/envelope_quarantine.nim(34, 15) Error: attempting to call undeclared routine: 'addUnique'
[2026-02-20T15:39:49.743Z] make: *** [Makefile:451: mev_mock] Error 1
```

ahshum: Oh sorry that I wasn't careful enough. Also the `remove()` needs to be fixed.

tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/d1236f2e72bba5990f4459064725651fbed39f3d

beacon_chain/sync/request_manager.nim
```diff
@@ -260,6 +285,69 @@ proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async:
...
+              peer.updateScore(PeerScoreUnviableFork)
+              return
+            of VerifierError.Invalid:
+              debugGloasComment("log level to notice")
+              debug "Received invalid envelope",
+                peer = peer, envelopes = shortLog(roots)
+              peer.updateScore(PeerScoreBadValues)
+              return
+          else:
+            gotGoodEnvelope = true
```
tersec, line 327: These boolean flag variables aren't ideal. The flow here appears to be either that `res.isErr()` and `gotGoodEnvelope == false`  it figures out what kind of error (`case res.error():`), or `res.isOk()` and `gotGoodEnvelope == true` and it runs `if gotGoodEnvelope:`.

But it's simpler and clearer to just put the entire `if gotGoodEnvelope:` in the `else` block currently occupied by `gotGoodEnvelope = true` and remove the then-unnecessary `gotGoodEnvelope` altogether.

beacon_chain/sync/request_manager.nim
```diff
@@ -260,6 +285,69 @@ proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async:
     if not(isNil(peer)):
       rman.network.peerPool.release(peer)
 
+proc fetchEnvelopesFromNetwork(self: RequestManager, roots: seq[Eth2Digest])
+    {.async: (raises: [CancelledError]).} =
+  var peer: Peer
+
+  try:
+    peer = await self.network.peerPool.acquire()
```
tersec: One approach to avoid the not-fully-initialized window for `peer`:
```nim
diff --git i/beacon_chain/sync/request_manager.nim w/beacon_chain/sync/request_manager.nim
index 14865121b..e1fe29a86 100644
--- i/beacon_chain/sync/request_manager.nim
+++ w/beacon_chain/sync/request_manager.nim
@@ -287,14 +287,12 @@ proc requestBlocksByRoot(rman: RequestManager, items: seq[Eth2Digest]) {.async:
 
 proc fetchEnvelopesFromNetwork(self: RequestManager, roots: seq[Eth2Digest])
     {.async: (raises: [CancelledError]).} =
-  var peer: Peer
+  let peer = await self.network.peerPool.acquire()
+  debug "Requesting envelopes by root",
+    peer = peer, envelopes = shortLog(roots),
+    peer_score = peer.getScore()
 
   try:
-    peer = await self.network.peerPool.acquire()
-    debug "Requesting envelopes by root",
-      peer = peer, envelopes = shortLog(roots),
-      peer_score = peer.getScore()
-
     let envelopes = await executionPayloadEnvelopesByRoot(
       peer, BlockRootsList roots)
```
The `finally` will run anyway to clean up/release it and this ensures `peer` is in the scope of `finally`.

beacon_chain/sync/request_manager.nim
ahshum: `uenvelopes` is a seq and we discard the result for some errors. There is a chance that all of them go into `VerifierError.Duplicate` or `VerifierError.MissingParent` (Probably we should use `discard` for `VerifierError.UnviableFork`).

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/f6ada173a423d80e4529bed8262745f58847c102

beacon_chain/sync/request_manager.nim
tersec: Ah, you're right.

beacon_chain/sync/request_manager.nim
```diff
@@ -504,6 +590,59 @@ proc requestManagerBlockLoop(
...
+    {.async: (raises: [CancelledError]).} =
+  while true:
+    # TODO This polling could be replaced with an AsyncEvent that is fired
+    #      from the quarantine when there's work to do
+    await sleepAsync(POLL_INTERVAL)
+
+    if self.inhibit():
+      continue
+
+    let missingBlockRoots = self.envelopeQuarantine[].getMissing()
```
tersec, line 605: Does it know which roots are finalized and not? [ExecutionPayloadEnvelopesByRange](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#executionpayloadenvelopesbyroot-v1) is only specified "since the latest finalized epoch".

beacon_chain/sync/request_manager.nim
```diff
@@ -504,6 +590,59 @@ proc requestManagerBlockLoop(
...
+    if missingBlockRoots.len() == 0:
+      continue
+
+    var blockRoots: seq[Eth2Digest]
+    if self.envelopeLoader == nil:
+      blockRoots = missingBlockRoots
```
tersec: If one is going to do this, then likely better to use `assign(blockRoots, missingBlockRoots)`, since Nim's built-in `genericAssignAux` is rather slow.

beacon_chain/sync/request_manager.nim
ahshum: We may know the slot of missing envelope as they are added when there is a valid block. So we should be able to check root with DAG in theory.

I will include the finality check in this PR if that makes sense.

tersec: Probably better to do this in a separate PR, but it would be good to have this check.

ahshum: Sure, will create a PR to clean up the envelope quarantine on slot as those calls are not triggered/missing at the moment.

beacon_chain/sync/request_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7953/changes/30be2ce8e5e86122e0a2affe363283ae36f42307

PR #7954, opened 2026-02-10
Use getPayloadV5 response for Gloas

Tomi-3-0:


tersec:
```

[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/validators/beacon_validators.nim(670, 20) template/generic instantiation of `withConsensusFork` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/validators/beacon_validators.nim(672, 17) template/generic instantiation of `proposeBlockAux` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/validators/beacon_validators.nim(655, 15) template/generic instantiation of `setResult` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/validators/beacon_validators.nim(411, 19) template/generic instantiation of `getExecutionPayload` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/validators/block_payloads.nim(412, 59) template/generic instantiation of `setResult` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/validators/block_payloads.nim(391, 27) template/generic instantiation of `getPayload` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/el/el_manager.nim(594, 3) template/generic instantiation of `setResult` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/el/el_manager.nim(589, 56) template/generic instantiation of `setResult` from here
[2026-02-10T18:57:59.958Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/vendor/nim-results/results.nim(454, 41) Error: type mismatch:
[2026-02-10T18:57:59.958Z]  got 'ExecutionPayloadForSigning' for 'asConsensusType(value(requests[get(bestPayloadIdx)]))' [object declared in /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/spec/datatypes/fulu.nim(148, 3)]
[2026-02-10T18:57:59.958Z]  but expected 'ExecutionPayloadForSigning = Alias' [object declared in /home/jenkins/workspace/tforms_linux_x86_64_main_PR-7954/beacon_chain/spec/datatypes/gloas.nim(75, 3)]
```

The error is because https://github.com/status-im/nimbus-eth2/blob/e709b33afb983c687e6870230906e9d62b4d849d/beacon_chain/el/engine_api_conversions.nim#L171-L189
i.e. the consensus and execution types don't losslessly round-trip, so it ends up mapping `gloas.ExecutionPayloadForSigning` to `engine_api.GetPayloadV5Response` which gets transformed back to `fulu.ExecutionPayloadForSigning`, which doesn't match the original `ExecutionPayloadForSigning` type.

It's unfortunate, because ePBS will absolutely ship with `GetPayloadV6Response`, so this would naturally disappear. However, it won't naturally disappear soon enough, so it has to be worked around.

The only caller of this `asConsensusType` is https://github.com/status-im/nimbus-eth2/blob/e709b33afb983c687e6870230906e9d62b4d849d/beacon_chain/el/el_manager.nim#L477-L594 and in particular https://github.com/status-im/nimbus-eth2/blob/e709b33afb983c687e6870230906e9d62b4d849d/beacon_chain/el/el_manager.nim#L588-L589 where it attempts to recover the original `gloas.ExecutionPayloadForSigning` type and fails.

The general principle here is to adjust that to call the correct `asConsensusType` function to recover a `gloas.ExecutionPayloadForSigning`. There are at least a couple of obvious ways, noting that `getPayload` has as one of its parameters
```nim
    PayloadType: type ForkyExecutionPayloadForSigning,
```
and therefore can directly access either the `PayloadType` or the fork (`ConsensusFork.Fulu` / `ConsensusFork.Gloas`) as required.

One is a more general, but also in the medium-term unnecessary approach where all the `asConsensusType` functions take an additional `static ConsensusType` parameter for Nim to dispatch on. Someday, that might be necessary, if we end up with a true case of a fork without a new  `GetPayload` version on the EL, but so far that has not happened, and it will not happen for Glamsterdam. It's possible ePBS will be dropped, but practically at this point not BALs or `SLOTNUM`. So this more general approach is more orthogonal but kind of ugly in that it adds redundant information to every `asConsensusType` definition and callsite to work around a situation which will last only as long as the split between BAL-only devnets and ePBS-only devnets.

Another approach is to put the special casing in `getPayload`'s call to `asConsensusType`, statically (`when`, using `PayloadType` or `Payload.kind`) detect the Gloas/V5 case, and call a renamed-to-`asConsensusTypeGloas` or siimlarly named function which is just the current `asConsensusType` for `GetPayloadV5` specialized to construct an `gloas.ExecutionPayloadForSigning` instead of a `fulu.ExecutionPayloadForSigning`. It can't be named `asConsensusFork` because then (without the additional `static ConsensusType parameter which this approach does not use) Nim has no way of disambiguating which the callsite wants. Then
```nim
  when Payload.kind == ConsensusType.Gloas
    return ok(requests[bestPayloadIdx.get()].value().asConsensusTypeGloas)
  else:
    return ok(requests[bestPayloadIdx.get()].value().asConsensusType)
```

And once the BAL/eBPS devnets merge into `glamsterdam` devnets, drop this special case and the special `asConsensusTypeGloas` and return to the current design with is fairly clean in its type handling but does require a one-to-one mapping between `GetPayload` types and `ExecutionPayloadForSigning` types.

Once there is such a fork which doesn't change `GetPayload` type, that 1:1 mapping is gone anyway, but so far it's not, so it's nicer to keep it as a design.

PR #7987, opened 2026-02-17
fix: self-built execution payload envelope processing

Tomi-3-0:
When a validator proposes a Gloas block and builds the execution payload locally (self-build), the envelope's state_root field is set at the creation time. This is the root after the beacon state root after `process_execution_payload_bid` runs but before `process_execution_payload` runs.

When `addHeadExecutionPayload` later calls `process_execution_payload` the additional state transitions inside `process_execution_payload` updates the `latest_block_hash`. The resulting state root no longer matches the `state_root` stored in the envelope that causes verification to fail with : 

`process_execution_payload: state root mismatch`

this passes the verify=false to `process_execution_payload` for self-built envelopes


beacon_chain/validators/beacon_validators.nim
```diff
@@ -602,47 +602,85 @@ proc proposeBlockAux(
...
+        envelope = shortLog(signedEnvelope.message),
+        error = envelopeRouteRes.error
+    else:
       notice "Payload Envelope proposed",
         blockRoot = shortLog(blockRoot),
         blck = shortLog(signedBlock.message),
         signature = shortLog(signature),
         validator = shortLog(validator)
 
+    # Process envelope locally - skip atate root verification
```
tersec: typo: `atate` -> `state`

beacon_chain/validators/beacon_validators.nim
```diff
@@ -602,47 +602,85 @@ proc proposeBlockAux(
...
+        envelope,
+        sleepAsync(NEWPAYLOAD_TIMEOUT),
+        retry = false)
+    
+    if payloadStatus.isNone:
+      warn "Failed to send self-built execution payload to local EL",
+        blockRoot = shortLog(blockRoot)
+      return head
+    elif payloadStatus.get != PayloadExecutionStatus.valid:
+      warn "Local EL rejected self-built execution payload",
```
tersec: This isn't necessarily true.

The engine API defines [five possible payload processing results](https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/paris.md#payloadstatusv1), which group naturally into:
- `VALID`;
- `SYNCING` and `ACCEPTED`, which are largely identical; and
- `INVALID` and `INVALID_BLOCK_HASH`, which can similar be treated identically

The consensus specs [canonically define these](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/sync/optimistic.md#helpers) groupings:
> - Alias `NOT_VALIDATED` to:
>   - `SYNCING`
>   - `ACCEPTED`
> - Alias `INVALIDATED` to:
>   - `INVALID`
>   - `INVALID_BLOCK_HASH`

Nimbus implements precisely this mapping:
https://github.com/status-im/nimbus-eth2/blob/a18b9461822122f4a1cecfa50a26b95090178b35/beacon_chain/consensus_object_pools/consensus_manager.nim#L92-L99

One cannot soundly infer from that an EL did not state that a payload was valid that it was, instead, `INVALIDATED`; it might have rather been `NOT_VALIDATED`, in which case it the EL did not fully process it; it did not, in this case, reject it either.

tersec: https://github.com/status-im/nimbus-eth2/pull/7987/changes/aca5e969b73f8e54eaf5b1eaaaf95d95d31ce31a

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7987/changes/aca5e969b73f8e54eaf5b1eaaaf95d95d31ce31a

beacon_chain/validators/beacon_validators.nim
```diff
@@ -602,47 +602,89 @@ proc proposeBlockAux(
...
+    let
+      signedEnvelope = gloas.SignedExecutionPayloadEnvelope(
         message: envelope,
         signature: signatureRes.get()
       )
 
-      discard await node.router.routeExecutionPayloadEnvelope(
-        signedEnvelope, checkValidator = false)
+      # Send payload to local EL before broadcasting
+      payloadStatus = await node.elManager.newExecutionPayload(
```
tersec: Generally, we don't necessarily want this in the critical path for getting the payload out at all, i.e. right now this sequences events as:
1. gossip-broadcast block
2. newPayload the envelope/payload
3. gossip-broadcast the payload

in that order.

Effectively, if the payload isn't accepted, that's a bug, either in Nimbus or in the EL, but it's better to try and get the payload out than to wait on the EL to do so in case. If it's not valid, it's not as if there's some fallback here. Better to just broadcast.

`newExecutionPayload` can [have a timeout of 8s](https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/engine/paris.md#engine_newpayloadv1). Nimbus should not wait 8s before proceeding with envelope broadcast, and probably the `newExecutionPayload` shouldn't occur in `proposeBlockAux` at all, or indeed `beacon_validators.nim`, but rather https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/gossip_processing/block_processor.nim

In particular, `block_processor` centralizes as much as feasible all block validation, from untrusted -> trusted. So what should happen is that we broadcast/route the envelope, and the routing also triggers the block processor to assess the changed block + envelope.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -602,47 +602,89 @@ proc proposeBlockAux(
...
         blockRoot = shortLog(blockRoot),
         blck = shortLog(signedBlock.message),
         signature = shortLog(signature),
         validator = shortLog(validator)
 
+    # Process envelope locally - skip state root verification
+    # for self-builds since the envelope's state_root is 
+    # computed before `process_execution_payload` runs,
+    # so it wouldn't match the post-envelope state root
+    discard node[].dag.addHeadExecutionPayload(
```
tersec: Is this special case necessary? I'd note that there's no `addHeadBlock` here either, broadly for similar reasons as the `newPayload` call: that's the domain of `block_processor`/`block_clearance`. `routeSignedBeaconBlock` and `routeExecutionPayloadEnvelope` alone should ensure `block_processor`/`block_clearance` pick this up.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -514,13 +514,17 @@ proc addHeadExecutionPayload*(
...
   process_execution_payload(
     dag.cfg,
     dag.clearanceState.forky(consensusFork),
     signedEnvelope,
     func(_: deneb.ExecutionPayload): bool = true,
     cache,
+    verify = not isSelfBuild
   ).isOkOr:
     assign(dag.clearanceState, dag.headState)
     info "Envelope transition failed", msg = error
```
ahshum, line 535: When envelope is self-build, I believe they could be from ourselves or others. And in devnet-0, all blocks should be self-build. It seems we would not be able to verify with clearance state effectively?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -941,12 +942,55 @@ proc storePayload(
     deadline = sleepAsync(deadlineTime)
 
   debugGloasComment("optimisticStatusRes")
```
tersec: Presumably filling this in should address/remove this `debugGloasComment`.

Tomi-3-0: yes

beacon_chain/consensus_object_pools/block_clearance.nim
Tomi-3-0: I believe "self-build" in this context refers to an envelope that was built locally by this node (i.e., we were the proposer and constructed the payload ourselves)

ahshum: Yes so my question is - how do we set `verify = true` for the received envelopes that are self-build by others?

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7987/changes/b669cfeeed09f5015ba006cc2c765a602a0166a8

tersec:

Tomi-3-0:
this: https://github.com/status-im/nimbus-eth2/pull/8006 and this: https://github.com/status-im/nimbus-eth2/pull/7748/ adresses the purpose of this PR

PR #7989, opened 2026-02-17
Envelope syncing

ahshum:


beacon_chain/sync/sync_manager.nim
```diff
@@ -93,11 +94,14 @@ type
 
   BeaconBlocksRes =
     NetRes[List[ref ForkedSignedBeaconBlock, Limit MAX_REQUEST_BLOCKS]]
+  EnvelopesRes =
+    NetRes[List[ref gloas.SignedExecutionPayloadEnvelope, Limit MAX_REQUEST_BLOCKS]]
```
tersec: Probably don't need `gloas.` here, as there shouldn't be any other `SignedExecutionPayloadEnvelope`s to disambiguate.

beacon_chain/sync/sync_queue.nim
```diff
@@ -22,8 +22,11 @@ type
   GetSlotCallback* = proc(): Slot {.gcsafe, raises: [].}
   GetBoolCallback* = proc(): bool {.gcsafe, raises: [].}
   ProcessingCallback* = proc() {.gcsafe, raises: [].}
-  BlockVerifier* =  proc(signedBlock: ForkedSignedBeaconBlock,
-                         blobs: Opt[BlobSidecars], maybeFinalized: bool):
+  BlockVerifier* =
+    proc(
+      signedBlock: ForkedSignedBeaconBlock,
+      singedEnvelope: Opt[gloas.SignedExecutionPayloadEnvelope],
```
tersec: typo: `signedEnvelope`

beacon_chain/sync/sync_manager.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7989/changes/ecbe1fc781d6149e71d603a1e34503edaf694941

beacon_chain/sync/sync_queue.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/7989/changes/ecbe1fc781d6149e71d603a1e34503edaf694941

beacon_chain/sync/sync_queue.nim
```diff
@@ -782,8 +797,12 @@ proc process[T](
...
     return SyncProcessingResult.init(SyncProcessError.Empty)
 
-  for blk, blb in blocks(sq.kind, blcks, blobs):
-    let res = await sq.blockVerifier(blk[], blb, maybeFinalized)
+  for blk, evl, blb in blocks(sq.kind, blcks, envelopes, blobs):
+    let evlval = evl.map(
+      func(x: ref gloas.SignedExecutionPayloadEnvelope):
+           gloas.SignedExecutionPayloadEnvelope =
+        x[])
+    let res = await sq.blockVerifier(blk[], evlval, blb, maybeFinalized)
```
tersec: The `map` bridges between https://github.com/status-im/nimbus-eth2/blob/80a09e88b6cab20e6f545db8a3475533810b6bb5/beacon_chain/sync/sync_queue.nim#L755-L762 and https://github.com/status-im/nimbus-eth2/blob/ecbe1fc781d6149e71d603a1e34503edaf694941/beacon_chain/sync/sync_queue.nim#L25-L30 i.e. the former yields `Opt[ref gloas.SignedExecutionPayloadEnvelope]` while the latter expects a dereferenced `Opt[SignedExecutionPayloadEnvelope]`. 

This creates a pointless and inefficient copy/allocation for each envelope, and envelopes will be fairly sizable, currently up to 10MB according to protocol limits. It would be better to align whether to uniformly use `ref`-based types here or non-`ref`-based types to avoid this.

`iterator blocks()` in gets this  `Opt[ref gloas.SignedExecutionPayloadEnvelope]` from https://github.com/status-im/nimbus-eth2/blob/ecbe1fc781d6149e71d603a1e34503edaf694941/beacon_chain/sync/sync_queue.nim#L747-L753
However, in this case, either all of the `envelopes` are effectively `Opt.some` or they're all `Opt.none`; it's not envelope-by-envelope.

This implies that the caller of `blocks` iterator, i.e. https://github.com/status-im/nimbus-eth2/blob/ecbe1fc781d6149e71d603a1e34503edaf694941/beacon_chain/sync/sync_queue.nim#L800-L805 can know ahead of time whether all of the blocks it receives back are going to be `Opt.some` or all of them are going to be `Opt.none` based solely on whether `envelopes` in the `for blk, evl, blb in blocks(sq.kind, blcks, envelopes, blobs)` iteration is `Opt.some` or `Opt.none`.

Because the dereferencing itself from `ref gloas.SignedExecutionPayloadEnvelope` to `SignedExecutionPayloadEnvelope` for the purposes of calling a function with it (i.e. `f(envelope[])` doesn't cost significantly more than `f(envelope)` and doesn't, in general, incur additional copy/allocation costs), the issue here arises from the `map`/monad-ish `bind` (these are lists, they're monads if one wants them to be).

One observation here is that the signature of the `blocks()` iterator itself forces this to some extent. So I'm not sure `iterator blocks()` is the best abstraction. It doesn't provide that much, and it pretty much has to do this `Opt` unwrapping/rewrapping of both `envelopes` and `sidecars`, which are both all-`Opt.some` or `Opt.none`.

So proposed alternative design: `iterator blocks()` doesn't really make sense with these levels of optional sidecars, so instead refactor that to just contain the check of `SyncQueueKind.Forward` or `SyncQueueKind.Backward`, do `countup`/`countdown` as appropriate, and then have `process()` do the actual lookups in into `envelopes`/`blobs` if/as needed (i.e. they _as a whole_ are `Opt.some`).

tersec: https://github.com/status-im/nimbus-eth2/pull/7989/commits/e74754dc821fa1e744cbaf75d44eeaf84e05e0a8 and https://github.com/status-im/nimbus-eth2/pull/7989/commits/21cb33b96db7f59aeed5daf9ce258fef03e7e3e8 address this to some extent, checking again

tersec:
Merge conflict with block_processor.

tersec:
Another merge conflict with `beacon_chain/gossip_processing/block_processor.nim`.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -609,32 +609,55 @@ proc initFullNode(
...
+          # Disable sidecars processing at block time.
+          const sidecarsOpt = noSidecars
+        elif consensusFork == ConsensusFork.Fulu:
           # TODO document why there are no columns here
-          when consensusFork == ConsensusFork.Gloas:
-            # Disable sidecars processing at block time.
-            const sidecarsOpt = noSidecars
-          else:
-            let sidecarsOpt = Opt.none(fulu.DataColumnSidecars)
+          let sidecarsOpt = Opt.none(fulu.DataColumnSidecars)
```
tersec, line 623: `const`

ahshum: Just would like to confirm, `const sidecarsOpt = Opt.none(fulu.DataColumnSidecars)` seems not working on my end. Does that require a different syntax?

tersec: No, the syntax is right.

```
Error: invalid type: 'ref DataColumnSidecar' in this context: 'Opt[DataColumnSidecars]' for const
```
Can see the same thing with `const foo = Opt.none(ref int)`. Anyway, it's fine, then, as far as this PR goes.

PR #8006, opened 2026-02-21
Fix envelope proposal

ahshum:
Fix as per 
- ~https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/builder.md#constructing-the-signedexecutionpayloadenvelope~
- ~https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.5/specs/gloas/builder.md#constructing-the-signedexecutionpayloadenvelope~
- https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/gloas/builder.md#constructing-the-signedexecutionpayloadenvelope


beacon_chain/validators/beacon_validators.nim
```diff
@@ -628,14 +649,18 @@ proc proposeBlockAux(
...
+      return head
     else:
       let signedEnvelope = gloas.SignedExecutionPayloadEnvelope(
         message: envelope,
         signature: signatureRes.get()
       )
 
-      discard await node.router.routeExecutionPayloadEnvelope(
-        signedEnvelope, checkValidator = false)
+      let res = await node.router.routeExecutionPayloadEnvelope(
```
agnxsh, line 645: why is this a separate method? can it not be a part of `routeSignedBeaconBlock`? i'm just trying to understand

tersec:
Updating to an `unstable` release incorporating

looks likely to fix the jenkins/nimbus-eth2/linux/x86_64/nix CI failure.

beacon_chain/validators/beacon_validators.nim
ahshum: This is one of the duties of builder (https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/builder.md) now.

Tomi-3-0 (approved):

beacon_chain/validators/beacon_validators.nim
```diff
@@ -611,12 +611,33 @@ proc proposeBlockAux(
...
+      func(_: deneb.ExecutionPayload): bool = true,
+      cache[],
+      verify = false,
+    ).isOkOr:
+      rollbackState()
+      debug "Proposed envelope failed to verify with transition", msg = error
+      return head
+
+    envelope.state_root = hash_tree_root(
+      node.dag.clearanceState.forky(consensusFork).data)
```
tersec: `clearanceState` is a `ForkedHashedBeaconState`, which means one shouldn't have to recompute the hash; it's cached:
https://github.com/status-im/nimbus-eth2/blob/82bb3f44f67c66eb5215be515a554cdb950aa7b6/beacon_chain/spec/forks.nim#L80-L89
and
https://github.com/status-im/nimbus-eth2/blob/82bb3f44f67c66eb5215be515a554cdb950aa7b6/beacon_chain/spec/datatypes/gloas.nim#L392-L394

beacon_chain/validators/beacon_validators.nim
```diff
@@ -611,12 +611,33 @@ proc proposeBlockAux(
...
       execution_requests = engineBid[].execution_requests,
       beacon_block_root = blockRoot,
       slot = slot,
-      state_root = signedBlock.message.state_root)
+      state_root = ZERO_HASH)
+
+    template rollbackState() =
+      assign(node.dag.clearanceState, node.dag.headState)
+
+    process_execution_payload(
```
tersec: This seems a bit odd to verify here, in general, because this whole `handleProposal()`/`proposeBlock()`/`proposeBlockAux()` callstack only triggers for BN-controlled validators, while as much as possible should be architected to support (even though, yes, Nimbus doesn't have the Beacon API support in place yet, it will have to by hoodi/sepolia testnet shipping) the VC/Beacon REST API approach as symmetrically/identically as possible, so they're guaranteed the same logic and have as much as feasible the same codepaths.

This is why `routeFoo()` combined with `block_processor` centralizes things: it means that both `beacon_validators` and the `rpc/rest_foo.nim` can call `routeFoo()` and get the same checks, the same codepaths. As is, to the extent this `process_execution_payload()` call is important, it's not reached at all here a REST version of Gloas block proposal.

I'd note that similarly the code above delegates to `routeSignedBeaconBlock()` to handle everything involved e.g.,
https://github.com/status-im/nimbus-eth2/blob/82bb3f44f67c66eb5215be515a554cdb950aa7b6/beacon_chain/validators/message_router.nim#L293-L313

and then all these callers (aside from the exported definition) run the same validation:
```
$ rg routeSignedBeaconBlock
beacon_chain/validators/message_router_mev.nim
133:      (await node.router.routeSignedBeaconBlock(
136:        return err("routeSignedBeaconBlock error") # Errors logged in router

beacon_chain/validators/message_router.nim
293:proc routeSignedBeaconBlock*(

beacon_chain/validators/beacon_validators.nim
593:      node.router.routeSignedBeaconBlock(signedBlock, sidecarsOpt,

beacon_chain/rpc/rest_beacon_api.nim
1052:            await node.router.routeSignedBeaconBlock(
1059:            await node.router.routeSignedBeaconBlock(
1064:            await node.router.routeSignedBeaconBlock(
1069:            await node.router.routeSignedBeaconBlock(
1207:            await node.router.routeSignedBeaconBlock(
```
with no code duplication and no possibility for different validation variations to drift apart, because they all live in that one `routeSignedBeaconBlock`.

So whatever validation needs to happen should ideally be part of `routeExecutionPayloadEnvelope()`.

beacon_chain/validators/beacon_validators.nim
ahshum: The main reason is with `verify = false`, state.root wouldn't be updated due to https://github.com/status-im/nimbus-eth2/blob/82bb3f44f67c66eb5215be515a554cdb950aa7b6/beacon_chain/spec/state_transition_block.nim#L1259-L1262

And follows the spec, `envelope.state_root` is created by using `process_execution_payload` with `verify=false` so I believe I have to manually compute the state root after that.

beacon_chain/validators/beacon_validators.nim
ahshum: Good shout. Will address this.

ahshum: Instead of message router, I move the logic to `makeExecutionPayloadEnvelope` which I think it would also help on testing.

beacon_chain/validators/message_router.nim
```diff
@@ -300,7 +300,14 @@ proc routeSignedBeaconBlock*(
   await router.publishRouteBlock(blck)
 
   # 3. Publish sidecars
-  let finalSidecars = await publishSidecars(router, blck, someSidecarsOpt)
+  const consensusFork = typeof(blck).kind
+  when consensusFork >= ConsensusFork.Gloas:
+    # Disable column processing at block time.
+    const finalSidecars = noSidecars
+  elif consensusFork in ConsensusFork.Deneb .. ConsensusFork.Fulu:
```
tersec: `routeSignedBeaconBlock()` won't ever see a pre-Electra block: https://github.com/status-im/nimbus-eth2/blob/186cfe0e2443f1926b2f68c13a8cc821066e72de/beacon_chain/validators/message_router.nim#L287-L293

tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes/b7cf4a08ff030432b6886246bc9ae103a8ff2528

beacon_chain/validators/message_router.nim
```diff
@@ -704,31 +709,29 @@ proc routePayloadAttestationMessage*(
 
 proc routeExecutionPayloadEnvelope*(
     router: ref MessageRouter,
+    signedBlock: gloas.SignedBeaconBlock | heze.SignedBeaconBlock,
     signedEnvelope: gloas.SignedExecutionPayloadEnvelope,
+    sidecarsOpt: Opt[seq[gloas.DataColumnSidecar]],
     checkValidator: bool
```
tersec: This looks unused?

beacon_chain/validators/block_payloads.nim
```diff
@@ -185,22 +185,41 @@ func decodePayloadRequests(
...
+    func(_: deneb.ExecutionPayload): bool = true,
+    cache,
+    verify = false,
+  ).isOkOr:
+    # Return the assigned envelope here due to some incorrect data. One
+    # should verify the state_root with zero value to confirm the envelope is
+    # successfully created.
+    return envelope
+
+  envelope.state_root = hash_tree_root(state.data)
```
tersec: Probably good to put a `debugHezeComment` here because this is a recomputation which should avoid most of the cost currently potentially required by a blind `hash_tree_root()`, or at least benchmarked to show that it's not the performance bottleneck full-`BeaconState` `hash_tree_root`s typically become. Adding a new one might be the expedient thing for the moment but it shouldn't be lost track of.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -610,12 +610,24 @@ proc proposeBlockAux(
...
-      slot = slot,
-      state_root = signedBlock.message.state_root)
+      slot = slot)
+
+    # Rollback clearanceState as it is modified.
+    assign(node.dag.clearanceState, node.dag.headState)
+
+    if envelope.state_root.isZero():
+      debug "Proposed envelope failed to verify with transition"
+      return head
```
tersec: Well, the head has still moved, no? Just as an empty block.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -627,14 +639,18 @@ proc proposeBlockAux(
     if signatureRes.isErr:
       error "Failed to sign sign execution payload envelope",
         slot, validator = shortLog(validator), err = signatureRes.error
+      return head
```
tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes#r3095206951

beacon_chain/validators/beacon_validators.nim
```diff
@@ -627,14 +639,18 @@ proc proposeBlockAux(
...
         signature: signatureRes.get()
       )
 
-      discard await node.router.routeExecutionPayloadEnvelope(
-        signedEnvelope, checkValidator = false)
+      let res = await node.router.routeExecutionPayloadEnvelope(
+        signedBlock, signedEnvelope, sidecarsOpt, checkValidator = false)
+      if res.isErr():
+        error "Failed to propose envelope", reason = res.error(), slot = slot
+        return head
```
tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes#r3095206951

beacon_chain/validators/message_router.nim
ahshum: I guess you mean the Heze block here?

beacon_chain/validators/block_payloads.nim
ahshum: Sure, but this will no longer be needed as https://github.com/ethereum/consensus-specs/pull/5094/files#diff-221ca7d5a71aa06ecab51f0872030eb8ba2217afd79b1dd24f4987e0282b9004L243-L251
is merged

tersec: ok, then skip

beacon_chain/validators/message_router.nim
tersec: `checkValidator` doesn't seem used in https://github.com/status-im/nimbus-eth2/blob/15b0756935d0ee05d5c1a02a046fd59a628cec7f/beacon_chain/validators/message_router.nim#L710-L734

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes/15b0756935d0ee05d5c1a02a046fd59a628cec7f

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes/15b0756935d0ee05d5c1a02a046fd59a628cec7f

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes/15b0756935d0ee05d5c1a02a046fd59a628cec7f

beacon_chain/validators/message_router.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8006/changes/8b655ccd5132a51f2c99dc68fce783485a98c805

beacon_chain/validators/beacon_validators.nim
```diff
@@ -631,14 +631,18 @@ proc proposeBlockAux(
     if signatureRes.isErr:
       error "Failed to sign sign execution payload envelope",
         slot, validator = shortLog(validator), err = signatureRes.error
+      return newBlockRef.get()
```
arnetheduck:
```suggestion
```

this looks redundant?

beacon_chain/validators/message_router.nim
```diff
@@ -299,7 +299,11 @@ proc routeSignedBeaconBlock*(
   await router.publishRouteBlock(blck)
 
   # 3. Publish sidecars
-  let finalSidecars = await publishSidecars(router, blck, someSidecarsOpt)
+  when typeof(blck).kind >= ConsensusFork.Gloas:
+    # Disable column processing at block time.
```
arnetheduck: wouldn't it make more sense to pass in `noSidecars` in the callers instead? https://github.com/status-im/nimbus-eth2/blob/unstable/beacon_chain/rpc/rest_beacon_api.nim#L1058 does not have a sidecar and the local flow is easier to follow if it's obvious in `proposeBlockAux` where the sidecar is routed/saved

ahshum: Created overloading and fork-based caller. Hope this looks better now.

PR #8009, opened 2026-02-23
Fix envelope processing and add logging

ahshum:


tersec:
Updating to an `unstable` release incorporating

looks likely to fix the jenkins/nimbus-eth2/linux/x86_64/nix CI failure.

PR #8014, opened 2026-02-23
Use [] on balances/validators instead of item

etan-status:
#7903 forces `balances` / `validators` lookup through a func returning `lent`, implying readonly access. So, we don't need to use `.item` anymore, which is only needed on `var HashList|...` to distinguish read/write access.


etan-status:
Context here is that for 7688 the type of the validators/balances field changes so it would return `lent seq[T]` instead of the HashList wrapper. assuming `lent` is readonly and given we don't manually hash_tree_root through that accessor anyway, that should not mess with caches etc

PR #8042, opened 2026-02-26
fix builder_index check against validators len

nethoxa:
This PR fixes an issue with the gloas implementation of `validateExecutionPayloadBid`, which incorrectly checked the `builder_index` against the validator set instead of the builder set

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -1973,7 +1973,7 @@ proc validateLightClientOptimisticUpdate*(
   pool.latestForwardedOptimisticSlot = attested_slot
   ok()
 
-# https://github.com/ethereum/consensus-specs/blob/v1.6.0-beta.1/specs/gloas/p2p-interface.md#execution_payload_bid
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#execution_payload_bid
 proc validateExecutionPayloadBid*(
     dag: ChainDAGRef,
     executionPayloadBidPool: ref ExecutionPayloadBidPool,
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -1983,26 +1983,17 @@ proc validateExecutionPayloadBid*(
 
   withState(dag.headState):
     when consensusFork >= ConsensusFork.Gloas:
-      # [REJECT] bid.builder_index is a valid, active, and non-slashed builder index
-      # Check builder index is valid
+      # [REJECT] bid.builder_index is a valid, active builder index
       if bid.builder_index >= forkyState.data.builders.lenu64:
         return dag.checkedReject("ExecutionPayloadBid: invalid builder index")
 
-      let validator = forkyState.data.builders.item(bid.builder_index)
-
-      # Check builder is active
-      let currentEpoch = get_current_epoch(forkyState.data)
-      if not is_active_validator(validator, currentEpoch):
+      if not is_active_builder(forkyState.data, bid.builder_index):
         return dag.checkedReject("ExecutionPayloadBid: builder not active")
 
-      # Check builder is not slashed
-      if validator.slashed:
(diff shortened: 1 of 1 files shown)


tersec:
```
2026-02-27T11:54:52.7275842Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/gossip_processing/gossip_validation.nim(1984, 12) template/generic instantiation of `withState` from here
2026-02-27T11:54:52.7279350Z /github-runner/github-runner-node-02/workspace/nimbus-eth2/nimbus-eth2/beacon_chain/gossip_processing/gossip_validation.nim(1995, 33) Error: type mismatch
2026-02-27T11:54:52.7280623Z Expression: is_active_validator(validator, currentEpoch)
2026-02-27T11:54:52.7281251Z   [1] validator: Builder
2026-02-27T11:54:52.7281927Z   [2] currentEpoch: Epoch
2026-02-27T11:54:52.7282187Z 
2026-02-27T11:54:52.7282354Z Expected one of (first mismatch at [position]):
2026-02-27T11:54:52.7282907Z [1] func is_active_validator(validator: Validator; epoch: Epoch): bool
2026-02-27T11:54:52.7284213Z 
2026-02-27T11:54:53.0515086Z make: *** [Makefile:449: mev_mock] Error 1
```

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1985,10 +1985,10 @@ proc validateExecutionPayloadBid*(
     when consensusFork >= ConsensusFork.Gloas:
       # [REJECT] bid.builder_index is a valid, active, and non-slashed builder index
```
tersec: First, https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#execution_payload_bid updated the text on this to make it clearer. It now reads:
> [REJECT] `bid.builder_index` is a valid/active builder index -- i.e. `is_active_builder(state, bid.builder_index)` returns `True`.

The issue is that the check builder is active section uses `is_active_validator`, but it should call https://github.com/status-im/nimbus-eth2/blob/82bb3f44f67c66eb5215be515a554cdb950aa7b6/beacon_chain/spec/beaconstate.nim#L3084-L3092 at which point it should be both consistent with the gossip validation specs and type-compatible.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1985,10 +1985,10 @@ proc validateExecutionPayloadBid*(
...
       # Check builder index is valid
-      if bid.builder_index >= forkyState.data.validators.lenu64:
+      if bid.builder_index >= forkyState.data.builders.lenu64:
         return dag.checkedReject("ExecutionPayloadBid: invalid builder index")
 
-      let validator = forkyState.data.validators.item(bid.builder_index)
+      let validator = forkyState.data.builders.item(bid.builder_index)
```
tersec: This variable name is misleading. It's not a `validator` anymore. It's a `builder`. Which is why...

beacon_chain/gossip_processing/gossip_validation.nim
tersec: This `is_active_validator()` call doesn't work, but one has to use `is_active_builder()`.

nethoxa:
On it, sorry for the delay, did not receive the GH ping

beacon_chain/gossip_processing/gossip_validation.nim
nethoxa: Changed to:

```nim
if not is_active_builder(forkyState.data, bid.builder_index):
        return dag.checkedReject("ExecutionPayloadBid: builder not active")
```

nethoxa:
I pushed another fix regarding the need of using `can_builder_cover_bid` for `[IGNORE] bid.value is less or equal than the builder's excess balance`, please take a look, thanks

beacon_chain/gossip_processing/gossip_validation.nim
tersec: This URL should be updated to https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#execution_payload_bid to reflect the new specs it implements.

nethoxa:
done @tersec

PR #8043, opened 2026-02-26
implement execution_payload_available event

Tomi-3-0:


beacon_chain/consensus_object_pools/envelope_quarantine.nim
```diff
@@ -11,6 +11,9 @@ import std/tables
 import ../spec/[digest, forks]
 
 type
+  OnExecutionPayloadCallback* = proc(
+    data: ExecutionPayloadInfoObject) {.gcsafe, raises: [].}
```
tersec: In theory, the `{.push raises: [], gcsafe.}` at the top of the module should render that part of this declaration redundant.

Tomi-3-0: Not exactly. The compiler complains
```
 Error: 'validateExecutionPayload' is not GC-safe as it calls 'onExecutionPayloadCallback'
make: *** [Makefile:370: all_tests] Error 1
```

tersec: Interesting. I guess `push` only pushes onto actual `proc`s/`func`s, not type declarations.

PR #8060, opened 2026-03-06
fix state replay crash

Tomi-3-0:


tersec:
Merge conflicts in `blockchain_dag`, `envelope_quarantine`, and `message_router`

PR #8066, opened 2026-03-11
draft: branch for partial messages

agnxsh:
tasks:
- ~make a partial column quarantine to cache headers as they come in via gossip~
- ~that quarantine should also be able to handle `PartialColumnDataSidecar` until the adequate no. of cells are hit, then assemble them into `DataColumnSidecar`~
- ~write tests for both of these cases~
- ~write the new datatypes and gossip validation conditions~
- ~make these assemble sidecars ingest into `ColumnQuaratine`~
- ~write a backward compatible broadcast mechanism to handle partial and data column sidecar, can add a flag maybe~
- ~investigate if there's anything left in the Nimbus Eth2 <-> Nim Libp2p layer in terms of orchestration/compatibility~
- ~manage eth2processor to deal with partial columns and then on requisite no of them populate data column quarantine~
- ~handle the partial to full column translation in the getBlobs service as well~


beacon_chain/consensus_object_pools/partial_column_quarantine.nim
```diff
@@ -0,0 +1,285 @@
...
+  MaxPartialHeaders* = 3 * int(SLOTS_PER_EPOCH)
+    ## Maximum number of validated headers to cache.
+  MaxPartialEntries* = 3 * int(SLOTS_PER_EPOCH) * NUMBER_OF_COLUMNS
+    ## Maximum number of (block_root, column_index) entries to cache.
+
+type
+  PartialColumnEntry* = object
+    ## Tracks accumulated cells for a single (block_root, column_index) pair.
+    headerValidated*: bool
+    cellsReceived*: seq[bool]
```
tersec: This is quite inefficient. See https://github.com/status-im/nim-stew/blob/master/stew/bitseqs.nim

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
```diff
@@ -0,0 +1,285 @@
...
+  minilru, results,
+  kzg4844/[kzg, kzg_abi],
+  ../spec/[digest, forks, helpers, presets]
+
+from ../spec/datatypes/fulu import
+  PartialDataColumnHeader, PartialDataColumnSidecar, ColumnIndex,
+  DataColumn, DataColumnSidecar,
+  MAX_BLOB_COMMITMENTS_PER_BLOCK
+
+from ../spec/datatypes/deneb import KzgCommitments, KzgProofs
```
tersec, line 24: If one imports `spec/forks`, there's no reason to import either `../spec/datatypes/deneb` or `../spec/datatypes/fulu` explicitly. But it's not obvious me it needs/benefits from `../spec/forks` to begin with -- what does it use it for?

beacon_chain/conf.nim
```diff
@@ -231,6 +231,11 @@ type
       desc: "Subscribe to the first half of column subnets"
       name: "light-supernode" .}: bool
 
+    partialColumns* {.
+      defaultValue: false,
+      desc: "Backward compatible partial data column sidecar support",
+      name: "partial-columns" .}: bool
```
tersec: Better to leave this `hidden` and `debug-foo` for now.

If we need to make it visible and part of the official set of options, can do that easily, can't go back the other way as easily.

In particular, once eventually deployed, it should not need this option at all. It should just be on, by default, because that's how Ethereum will work.

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
```diff
@@ -0,0 +1,286 @@
+# beacon_chain
+# Copyright (c) 2025-2026 Status Research & Development GmbH
```
tersec: This module didn't exist in 2025.

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
```diff
@@ -0,0 +1,286 @@
...
+  ## Ingest cells and proofs from a validated PartialDataColumnSidecar into
+  ## the quarantine entry for the given (block_root, column_index) pair.
+  ## Assumes the sidecar has already passed validatePartialDataColumnSidecar.
+  let key = PartialColumnKey(blockRoot: blockRoot, columnIndex: columnIndex)
+  var entry = quarantine.entries.get(key).valueOr:
+    return
+
+  var cellIdx = 0
+  for blobIdx in 0 ..< int(MAX_BLOB_COMMITMENTS_PER_BLOCK):
+    if sidecar.cells_present_bitmap[Natural(blobIdx)]:
```
tersec, line 249: Is `Natural` necessary here?

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -204,6 +217,7 @@ proc new*(T: type Eth2Processor,
           quarantine: ref Quarantine,
           blobQuarantine: ref BlobQuarantine,
           dataColumnQuarantine: ref ColumnQuarantine,
+          partialColumnQuarantine: ref PartialColumnQuarantine,
```
tersec: Ultimately it doesn't really make sense to have two different quarantines here, or?

If a full data column is present in the data column quarantine, then it should be representable in the more general `partialColumnQuarantine`. And right now, this leads to two possible quarantines to check.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -638,6 +650,201 @@ proc validateBlobSidecar*(
...
+      let r = check_partial_data_column_sidecar_kzg_proofs(
+        p_data_column_sidecar,
+        header.kzg_commitments,
+        column_index)
+      if r.isErr:
+        return dag.checkedReject(r.error)
+
+  ok()
+
+
```
tersec: extra empty line between functions?

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
+# beacon_chain
+# Copyright (c) 2025-2026 Status Research & Development GmbH
```
tersec: This module didn't exist in 2025.

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+  let tmp = uint64(index).toBytesLE()
+  copyMem(addr res.bytes[0], unsafeAddr tmp[0], sizeof(uint64))
+  res
+
+func genPartialDataColumnHeader(
+    slot: int, proposerIndex: int, numCommitments: int
+): PartialDataColumnHeader =
+  var commitments: KzgCommitments
+  for i in 0 ..< numCommitments:
+    discard commitments.add(genKzgCommitment(i))
```
tersec: Tests generally shouldn't `discard` return values but verify whatever they're supposed to be.

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+  # --- markCellReceived with cell data and proof ---
+
+  test "markCellReceived with data stores cell and proof":
+    var quarantine = PartialColumnQuarantine.init()
+    let
+      root = genBlockRoot(1)
+      colIdx = ColumnIndex(0)
+      numBlobs = 3
+
+    discard quarantine.getOrCreateEntry(root, colIdx, numBlobs)
```
tersec: Similar comment about `discard` in tests.

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+  test "markCellReceived with data on non-existent entry is no-op":
+    var quarantine = PartialColumnQuarantine.init()
+    quarantine.markCellReceived(
+      genBlockRoot(99), ColumnIndex(0), 0, genKzgCell(1), genKzgProof(1))
+    check not quarantine.hasCellReceived(genBlockRoot(99), ColumnIndex(0), 0)
+
+  test "markCellReceived with data out-of-bounds is no-op":
+    var quarantine = PartialColumnQuarantine.init()
+    let root = genBlockRoot(1)
+    discard quarantine.getOrCreateEntry(root, ColumnIndex(0), numBlobs = 2)
```
tersec: More `discard`

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+  # --- addCells ---
+
+  test "addCells ingests cells from a PartialDataColumnSidecar":
+    var quarantine = PartialColumnQuarantine.init()
+    let
+      root = genBlockRoot(1)
+      colIdx = ColumnIndex(5)
+      numBlobs = 4
+
+    discard quarantine.getOrCreateEntry(root, colIdx, numBlobs)
```
tersec: More `discard`.

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+      entry.cells[3].isNone()
+
+  test "addCells accumulates across multiple sidecars":
+    var quarantine = PartialColumnQuarantine.init()
+    let
+      root = genBlockRoot(1)
+      colIdx = ColumnIndex(0)
+      numBlobs = 3
+
+    discard quarantine.getOrCreateEntry(root, colIdx, numBlobs)
```
tersec: `discard` in a test

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+    quarantine.addCells(genBlockRoot(99), ColumnIndex(0), sidecar)
+    check not quarantine.hasEntry(genBlockRoot(99), ColumnIndex(0))
+
+  test "addCells with overlapping bitmap overwrites existing cells":
+    var quarantine = PartialColumnQuarantine.init()
+    let
+      root = genBlockRoot(1)
+      colIdx = ColumnIndex(0)
+
+    discard quarantine.getOrCreateEntry(root, colIdx, numBlobs = 3)
```
tersec: `discard` in a test

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,973 @@
...
+      genPartialDataColumnSidecar([1], startCellId = 99))
+    let entry2 = quarantine.getEntry(root, colIdx).get()
+    check entry2.cells[1].get() == genKzgCell(99)
+
+  test "addCells is independent across columns":
+    var quarantine = PartialColumnQuarantine.init()
+    let root = genBlockRoot(1)
+
+    discard quarantine.getOrCreateEntry(root, ColumnIndex(0), numBlobs = 3)
+    discard quarantine.getOrCreateEntry(root, ColumnIndex(1), numBlobs = 3)
```
tersec: two `discard`s in a test

tests/test_block_processor.nim
```diff
@@ -403,5 +404,207 @@ suite "Block processor" & preset():
...
+          KZG_COMMITMENTS_INCLUSION_PROOF_DEPTH_GINDEX.GeneralizedIndex,
+          pHeader.kzg_commitments_inclusion_proof).expect("Valid gindex")
+
+        # Use quarantine to ingest partials and reassemble
+        var pcq = PartialColumnQuarantine.init()
+        pcq.putPartialHeader(blockRoot, pHeader)
+
+        # Pick column index 0 for the round-trip test
+        let colIdx = ColumnIndex(0)
+        discard pcq.getOrCreateEntry(blockRoot, colIdx, numBlobs)
```
tersec: A `discard` here too

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -638,6 +650,201 @@ proc validateBlobSidecar*(
...
 
+proc validatePartialDataColumnSidecar*(
+    dag: ChainDAGRef, quarantine: ref Quarantine,
+    dataColumnQuarantine: ref ColumnQuarantine,
+    partialColumnQuarantine: ref PartialColumnQuarantine,
+    p_data_column_sidecar: fulu.PartialDataColumnSidecar,
+    wallTime: BeaconTime, subnet_id: uint64):
+    Result[void, ValidationError] =
+
+  # === For all partial messages ===
```
tersec, line 471: are there other partial messages? Also, this seems to be still fairly partial column-specific. Not really for "all partial messages"

beacon_chain/validators/beacon_validators.nim
```diff
@@ -587,14 +633,15 @@ proc proposeBlockAux(
...
 
-  let
-    newBlockRef = await(
-      node.router.routeSignedBeaconBlock(signedBlock, sidecarsOpt,
-        checkValidator = false)
-    ).valueOr:
-      # TODO Is this an error?
-      beacon_block_production_errors.inc()
-      return head # Errors logged in router
+  when consensusFork != ConsensusFork.Fulu:
```
tersec: Presumably the partial cell messaging would continue in some form into Gloas? I'm not sure if/how that's specified though.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -416,22 +416,23 @@ proc assemble_partial_data_column_sidecars*(
...
+    if sidecar.cells_present_bitmap[Natural(i)]:
+      blobIndices.add(i)
+
+  # The cell index is the column index for all cells in this column
+  let cellIndices = repeat(CellIndex(column_index), blobIndices.len)
+
+  # Batch verify that the cells match the corresponding commitments and proofs
+  var commitments = newSeqOfCap[KzgCommitment](blobIndices.len)
+  for i in blobIndices:
+    commitments.add(all_commitments[i])
```
tersec: This loop can be more clearly expressed as a mapIt:
```nim
  let commitments = blobIndices.mapIt(all_commitments[it])
```

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -638,6 +650,201 @@ proc validateBlobSidecar*(
...
+    return dag.checkedReject(
+      "PartialDataColumnSidecar: message is semantically empty")
+
+  # [REJECT] There are the same number of cells and proofs in the message.
+  if p_data_column_sidecar.partial_columns.len !=
+      p_data_column_sidecar.kzg_proofs.len:
+    return dag.checkedReject(
+      "PartialDataColumnSidecar: cells and proofs count mismatch")
+
+  let column_index = ColumnIndex(subnet_id)
```
tersec, line 495: Subnet indices can't be uniquely mapped back to data column indices, e.g., see
https://github.com/status-im/nimbus-eth2/blob/6aa946f8220089b691fc6ae8aa382d9e2de3d081/beacon_chain/spec/network.nim#L139-L144 or [`compute_subnet_for_data_column_sidecar`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.3/specs/fulu/p2p-interface.md#compute_subnet_for_data_column_sidecar):
```python
def compute_subnet_for_data_column_sidecar(column_index: ColumnIndex) -> SubnetID:
    return SubnetID(column_index % DATA_COLUMN_SIDECAR_SUBNET_COUNT)
```

tersec: To the extent this is true, as it currently is, this needs another one of those `static: doAssert DATA_COLUMN_SIDECAR_SUBNET_COUNT == NUMBER_OF_COLUMNS` assertions to verify that it's currently safe (but that might change in the future, and it should not compile the moment it isn't).

beacon_chain/validators/message_router.nim
tersec: Is there ever going to be a reason to route a `fulu.SignedBeaconBlock` (or a `gloas.SignedBeaconBlock`, when the partial cell dissemination is defined for that fork) without the `partialSidecars` of the overload added in this PR? It seems like it should just not compile, insofar as it allows silently neglecting to send the partial sidecars.

Or is there a use case for this?

beacon_chain/networking/eth2_network.nim
```diff
@@ -2856,6 +2904,26 @@ proc broadcastDataColumnSidecar*(
       node.forkDigestAtEpoch(contextEpoch), subnet_id)
   node.broadcast(topic, data_column)
 
+proc broadcastPartialDataColumnHeader*(
```
tersec: Nothing ever seems to call this function?

beacon_chain/networking/eth2_network.nim
```diff
@@ -2606,6 +2608,52 @@ func addValidator*[MsgType](
...
+      Future[ValidationResult] {.raises: [].} =
+    inc nbc_gossip_messages_received
+    trace "Validating incoming gossip message (dual)",
+      len = message.data.len, topic
+
+    let maxSize = max(gossipMaxSize(MsgType1), gossipMaxSize(MsgType2))
+    var decompressed = snappy.decode(message.data, maxSize)
+    let res = if decompressed.len > 0:
+      block:
+        var result = ValidationResult.Reject
```
tersec: `result` has special meaning in Nim. Better to use `res` or any other non-literally-`result` name for this kind of variable.

tersec: Here there's already a `res`, so could be a different variation or etc of `res`.

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
```diff
@@ -0,0 +1,286 @@
...
+  MaxPartialHeaders* = 3 * int(SLOTS_PER_EPOCH)
+    ## Maximum number of validated headers to cache.
+  MaxPartialEntries* = 3 * int(SLOTS_PER_EPOCH) * NUMBER_OF_COLUMNS
+    ## Maximum number of (block_root, column_index) entries to cache.
+
+type
+  PartialColumnEntry* = object
+    ## Tracks accumulated cells for a single (block_root, column_index) pair.
+    headerValidated*: bool
+    cellsReceived*: BitSeq
```
tersec, line 44: Is this redundant with the `isSome`/`isNone`-ness of each `cell`/`proof` entry? Can they diverge?

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -410,6 +425,127 @@ proc processBlobSidecar*(
...
+      (Moment.now() - validationStart).toFloatSeconds())
+
+    if v.isErr():
+      debug "Dropping partial data column", error = v.error()
+      partial_data_column_sidecars_dropped.inc(1, [$v.error[0]])
+      return v
+
+    let
+      block_root = hash_tree_root(block_header)
+      column_index = ColumnIndex(subnet_id)
```
tersec, line 432: This is only true if
```nim
  static: doAssert DATA_COLUMN_SIDECAR_SUBNET_COUNT == NUMBER_OF_COLUMNS
```
so if relying on this assertion implicitly, should add it explicitly, so if that ever changes, the compilation will immediately fail and the fix will be more obvious.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -241,6 +241,18 @@ proc check_data_column_sidecar_kzg_proofs(
 
   ok()
 
+proc check_partial_data_column_sidecar_kzg_proofs(
+    p_data_column: fulu.PartialDataColumnSidecar,
+    all_commitments: deneb.KzgCommitments,
+    column_index: ColumnIndex):
+    Result[void, ValidationError] =
+  let res = p_data_column.verify_partial_data_column_sidecar_kzg_proofs(
+      all_commitments, column_index)
```
tersec: Reduce indentation by 2 spaces.

beacon_chain/validators/message_router.nim
```diff
@@ -304,6 +330,33 @@ proc routeSignedBeaconBlock*(
   # 4. Add block to DAG
   return await router.addRoutedBlock(blck, finalSidecars)
 
+proc routeSignedBeaconBlock*(
+    router: ref MessageRouter,
+    blck: fulu.SignedBeaconBlock,
+    someSidecarsOpt: seq[fulu.DataColumnSidecar],
+    partialSidecars: seq[fulu.PartialDataColumnSidecar],
```
tersec, line 339: Generally speaking, it seems like the `someSidecarsOpt` and `partialSidecars` should be constructible from each other, or?

beacon_chain/networking/eth2_network.nim
agnxsh: yeah i thought earlier there's a function to publish the header separately, since we have atomic header management in gossip validation function

beacon_chain/validators/beacon_validators.nim
agnxsh: yes, but the experimental spec branch does not suggest that, i've tried to keep it that way

beacon_chain/gossip_processing/gossip_validation.nim
agnxsh: technically there's a cell only situation and a cells + header situation

beacon_chain/validators/message_router.nim
agnxsh: yes we are passing both to broadcast one and adding the other to block processor without wasting much time and resources converting

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/8192c964b1666041462b68dbd2b83f862a59b8b3

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tests/test_partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/34dae10ed0c9409270a28fe3782a6a75f2241e8c

tersec:
Merge conflicts in `AllTests-mainnet.md` and `tests/test_block_processor.nim`.

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,234 @@
...
+
+method partsMetadata*(m: DataColumnPartialMessage): PartsMetadata {.gcsafe, raises: [].} =
+  ## Returns metadata indicating which cells are available in this partial
+  ## message.
+  var metadata: fulu.PartialDataColumnPartsMetadata
+  metadata.available = m.sidecar.cells_present_bitmap
+  # We don't set requests here -- this describes what we *have*
+  encodePartsMetadata(metadata)
+
+method materializeParts*(
```
tersec: Why the 3 `method`s?

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,234 @@
...
+  ok(encodePartsMetadata(merged))
+
+# -----------------------------------------------------------------------------
+# DataColumnPartialMessage: concrete PartialMessage
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/fulu/p2p-interface.md#encoding-and-decoding-responses
+# All responses MUST be encoded and decoded with the
+# PartialDataColumnSidecar container.
+# -----------------------------------------------------------------------------
+
+proc newDataColumnPartialMessage*(
```
tersec: `func`

tersec:
Should add tests for
https://github.com/status-im/nimbus-eth2/blob/940d6907eac347ca79efc04967e47ca5d2dfa59e/tests/consensus_spec/fulu/test_fixture_ssz_consensus_objects.nim#L197-L199

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,234 @@
...
+
+  # Determine which cells the peer needs (they request) that we have
+  var
+    bitmap: BitArray[int(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
+    cells = newSeqOfCap[KzgCell](m.sidecar.partial_columns.len)
+    proofs = newSeqOfCap[KzgProof](m.sidecar.kzg_proofs.len)
+    ourIdx = 0  # index into our sparse arrays
+
+  for i in 0 ..< int(MAX_BLOB_COMMITMENTS_PER_BLOCK):
+    let n = Natural(i)
```
tersec: Can the loop iteration itself be over a `Natural`? `for i in 0.Natural ..< MAX_BLOB_COMMITMENTS_PER_BLOCK.Natural:`?

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,234 @@
...
+    return err("unsupported groupId version: " & $rpc.groupID[0])
+  # partsMetadata can be empty (no metadata update, just data)
+  # partialMessage can be empty (metadata-only update)
+  ok()
+
+# -----------------------------------------------------------------------------
+# Config factory
+# -----------------------------------------------------------------------------
+
+proc makePartialMessageExtensionConfig*(
```
tersec, line 219: `func`

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,234 @@
...
+  ## - available bits are OR'd together
+  ## - requests bits are OR'd together, then any bits already available
+  ##   are cleared
+  let
+    metaA = ? decodePartsMetadata(a)
+    metaB = ? decodePartsMetadata(b)
+
+  var merged: fulu.PartialDataColumnPartsMetadata
+  for i in 0 ..< int(MAX_BLOB_COMMITMENTS_PER_BLOCK):
+    let n = Natural(i)
```
tersec: Similar question as elsewhere, it would be nicer if the loop were natively over `Natural` since that's all it uses.

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,234 @@
+# beacon_chain
+# Copyright (c) 2026 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [].}
```
tersec: Several of the functions here are defined as `{.raises: [], gcsafe.}`. Would be cleaner to declare the whole module as such here and avoid repeating that on `unionPartsMetadata`, `groupId`, etc.

beacon_chain/validators/message_router.nim
```diff
@@ -246,6 +266,36 @@ proc publishSidecars(
...
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/fulu/p2p-interface.md#partial-columns-for-cell-dissemination
+proc publishPartialSidecars(
+    router: ref MessageRouter,
+    blck: fulu.SignedBeaconBlock,
+    partials: seq[fulu.PartialDataColumnSidecar]
+): Future[void] {.async: (raises: [CancelledError]).} =
+  ## Publish partial data column sidecars via the gossipsub Partial Message
+  ## Extension. Parts metadata and materialized cells are sent to peers that
+  ## opted into partial messages on the relevant data column sidecar topics.
+  let blockRoot = hash_tree_root(blck.message)
```
tersec, line 257: This could be relatively heavyweight; in Fulu, blocks can be up to 10MB.

beacon_chain/networking/gossip_partial_columns.nim
tersec: Also this one in particular doesn't seem to be used anywhere?
```
$ rg materializeP
beacon_chain/networking/gossip_partial_columns.nim
148:method materializeParts*(
```

tersec:
https://ci.status.im/blue/rest/organizations/jenkins/pipelines/nimbus-eth2/pipelines/platforms/pipelines/linux/pipelines/x86_64/pipelines/main/branches/PR-8066/runs/19/nodes/77/log/?start=0
```
[2026-04-06T10:31:58.326Z] + git diff --exit-code --ignore-submodules=all
[2026-04-06T10:31:58.326Z] diff --git a/AllTests-mainnet.md b/AllTests-mainnet.md
[2026-04-06T10:31:58.326Z] index c6862485f..cab5b7baf 100644
[2026-04-06T10:31:58.326Z] --- a/AllTests-mainnet.md
[2026-04-06T10:31:58.326Z] +++ b/AllTests-mainnet.md
[2026-04-06T10:31:58.326Z] @@ -151,7 +151,6 @@ AllTests-mainnet
[2026-04-06T10:31:58.326Z]  + Partial column quarantine round-trip assembly [Preset: mainnet]                            OK
[2026-04-06T10:31:58.326Z]  + Process Deneb block with blob sidecars [Preset: mainnet]                                   OK
[2026-04-06T10:31:58.326Z]  + Process Deneb block without blob sidecars [Preset: mainnet]                                OK
[2026-04-06T10:31:58.326Z] -+ Process Fulu block via partial-assembled data columns [Preset: mainnet]                    OK
[2026-04-06T10:31:58.326Z]  + Process Fulu block with data column sidecars [Preset: mainnet]                             OK
[2026-04-06T10:31:58.326Z]  + Process Fulu block without data column sidecars [Preset: mainnet]                          OK
[2026-04-06T10:31:58.326Z]  + Process a block from each fork (without blobs) [Preset: mainnet]                           OK
script returned exit code 1[2026-04-06T10:32:00.520Z] [WS-CLEANUP] Deleting project workspace...
```

beacon_chain/networking/eth2_network.nim
```diff
@@ -2672,6 +2702,52 @@ func addValidator*[MsgType](
...
+        var v = ValidationResult.Reject
+        # Try primary type first
+        try:
+          let decoded = SSZ.decode(decompressed, MsgType1)
+          v = primaryValidator(decoded, message.fromPeer)
+        except SerializationError:
+          # Try fallback type
+          try:
+            let decoded = SSZ.decode(decompressed, MsgType2)
+            v = fallbackValidator(decoded, message.fromPeer)
```
tersec: It's not really correct to just try to SSZ decode one, then the other. It's encoded, according to Raul, differently on the libp2p side, and only one decoding is correct per message delivery approach. One suggestion from @richard-ramos was to use https://github.com/status-im/nimbus-eth2/blob/72c86e243ab4ad75d3eb9ce30a26eeda960fed2a/beacon_chain/nimbus_beacon_node.nim#L879-L894

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -14,12 +14,21 @@ import
...
   ssz_serialization/types,
   ../el/el_manager,
+<<<<<<< HEAD
+  ../spec/[helpers, forks, network],
+  ../consensus_object_pools/[
+    attestation_pool, blob_quarantine, block_clearance, block_quarantine,
+    blockchain_dag, envelope_quarantine, execution_payload_pool,
+    partial_column_quarantine, payload_attestation_pool, light_client_pool,
+    sync_committee_msg_pool, validator_change_pool],
+=======
```
tersec: Merge artifacts

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -14,12 +14,21 @@ import
...
   ../spec/[column_map, helpers, forks],
   ../consensus_object_pools/[
     attestation_pool, block_clearance, block_quarantine, blockchain_dag,
     column_quarantine, envelope_quarantine, execution_payload_pool,
     payload_attestation_pool, light_client_pool, sync_committee_msg_pool,
     validator_change_pool],
+>>>>>>> dc71cf782a4e1fb82ad274a8568748dcec94eff8
```
tersec: More merge artifacts

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -420,22 +420,21 @@ proc assemble_partial_data_column_sidecars*(
 
 proc verify_partial_data_column_sidecar_kzg_proofs*(
```
tersec, line 566: Link to https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/fulu/p2p-interface.md#verify_partial_data_column_sidecar_kzg_proofs (and it should match semantics)

beacon_chain/validators/beacon_validators.nim
```diff
@@ -577,6 +577,56 @@ proc proposeBlockAux(
       engineBid[].eps.blobsBundle.blobs.mapIt(kzg.KzgBlob(bytes: it)),
       engineBid[].eps.blobsBundle.proofs.mapIt(kzg.KzgProof(it))))
   elif consensusFork == ConsensusFork.Fulu:
+    let
+      blobs = engineBlock.blobsBundle.blobs.mapIt(kzg.KzgBlob(bytes: it))
+      proofs = @(engineBlock.blobsBundle.proofs.mapIt(kzg.KzgProof(it)))
```
tersec: Where are these `@(...mapIt(...))` coming from? See
- https://github.com/status-im/nimbus-eth2/pull/8251

They're pointless.

beacon_chain/networking/eth2_network.nim
```diff
@@ -2927,6 +3003,86 @@ proc broadcastDataColumnSidecar*(
       node.forkDigestAtEpoch(contextEpoch), subnet_id)
   node.broadcast(topic, data_column)
 
+proc broadcastPartialDataColumnSidecar*(
+    node: Eth2Node, subnet_id: uint64,
+    p_data_column: fulu.PartialDataColumnSidecar):
+    Future[SendResult] {.async: (raises: [CancelledError], raw: true).} =
+  let header = p_data_column.header[0]
```
tersec: There's no reason to copy this at all. It's copying at https://github.com/status-im/nimbus-eth2/blob/72c86e243ab4ad75d3eb9ce30a26eeda960fed2a/beacon_chain/spec/datatypes/fulu.nim#L109-L114 which isn't huge but isn't tiny and no purpose to doing so.

Usual options here are:
- inlining `p_data_column.header[0]` into `contextEpoch = header.signed_block_header.message.slot.epoch`; and
- `template header ...` (instead of `let`)

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
agnxsh: BitArray's [] / []= accessors require a Natural (non-negative integer)

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/9053d629303d6bed6171b982dcb966ae6c5aa63b

beacon_chain/spec/peerdas_helpers.nim
tersec: alpha.8 now

tests/test_block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/a86d9ebb4f41905d34288c794c6239f1a0738c6b

beacon_chain/conf.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/14b5d2437514aa43fc9bd4771df643f7b945c9c6

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/14b5d2437514aa43fc9bd4771df643f7b945c9c6

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/34dae10ed0c9409270a28fe3782a6a75f2241e8c

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/34dae10ed0c9409270a28fe3782a6a75f2241e8c

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/34dae10ed0c9409270a28fe3782a6a75f2241e8c

beacon_chain/networking/gossip_partial_columns.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/77e4c5e15e25bd4bad045f383a428cff2654eedc

beacon_chain/networking/gossip_partial_columns.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/77e4c5e15e25bd4bad045f383a428cff2654eedc

beacon_chain/networking/gossip_partial_columns.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/77e4c5e15e25bd4bad045f383a428cff2654eedc

beacon_chain/networking/gossip_partial_columns.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/77e4c5e15e25bd4bad045f383a428cff2654eedc

beacon_chain/networking/gossip_partial_columns.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/77e4c5e15e25bd4bad045f383a428cff2654eedc

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/9be094da4ce3c623d20677b45a200de4823d1acb

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/9be094da4ce3c623d20677b45a200de4823d1acb

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/9be094da4ce3c623d20677b45a200de4823d1acb

beacon_chain/networking/eth2_network.nim
tersec: https://github.com/status-im/nimbus-eth2/commit/9be094da4ce3c623d20677b45a200de4823d1acb

tests/test_block_processor.nim
```diff
@@ -20,8 +20,9 @@ import
     peerdas_helpers, state_transition],
   ../beacon_chain/gossip_processing/block_processor,
   ../beacon_chain/consensus_object_pools/[
-    blockchain_dag, block_quarantine, block_clearance, column_quarantine,
-    consensus_manager, envelope_quarantine
+    attestation_pool, blockchain_dag, block_quarantine, block_clearance,
```
tersec, line 23:
```
/Users/runner/work/nimbus-eth2/nimbus-eth2/tests/test_block_processor.nim(31, 45) Error: duplicate import of 'attestation_pool'; previous import here: /Users/runner/work/nimbus-eth2/nimbus-eth2/tests/test_block_processor.nim(23, 5) [DuplicateModuleImport]
```
Another option is to add the relevant symbols to
```nim
from ../beacon_chain/consensus_object_pools/attestation_pool import
  AttestationPool, init
```

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -655,6 +667,228 @@ proc validateBlobSidecar*(
...
+      return errIgnore("PartialDataColumnSidecar: slot already finalized")
+
+    let block_root = hash_tree_root(block_header)
+
+    # [REJECT] If a valid header was previously received, the received header
+    # MUST equal the previously valid header.
+    let existingHeader =
+      partialColumnQuarantine[].getPartialHeader(block_root)
+    if existingHeader.isSome():
+      let existing = existingHeader.get()
```
tersec, line 541: `isErrOr` to avoid the explicit/error-prone `isSome`/`get` pair (here it's obvious, but we have had a case recently where an unprotected `get` could cause a `Defect`. Better to avoid if possible)

tests/test_block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8066/changes/8479170b57297479868c07d5fc25a8edf04f0c60

tests/test_gossip_partial_columns.nim
```diff
@@ -0,0 +1,280 @@
...
+
+  # --- materializeParts ---
+
+  test "materializeParts with empty metadata returns the full sidecar":
+    let
+      sidecar = genSidecar([0, 2], startCellId = 100)
+      m = newDataColumnPartialMessage(
+        genDigest(1), ColumnIndex(0), sidecar)
+      parts = m.materializeParts(@[]).get()
+      roundTrip = SSZ.decode(seq[byte](parts), fulu.PartialDataColumnSidecar)
```
tersec:
```
/github-runner/github-runner-node-01/workspace/nimbus-eth2/nimbus-eth2/tests/test_gossip_partial_columns.nim(191, 39) Error: conversion from seq[byte] to itself is pointless [ConvFromXtoItselfNotNeeded]
```

beacon_chain/networking/eth2_network.nim
```diff
@@ -2929,6 +2959,96 @@ proc broadcastDataColumnSidecar*(
...
+    pdc: fulu.PartialDataColumnSidecar]
+
+  if rpc.partialMessage.len == 0:
+    return Opt.none(R)
+
+  let blockRoot = parseGroupId(rpc.groupID).valueOr:
+    debug "Invalid partial column RPC groupId", error, peer
+    return Opt.none(R)
+
+  let subnetId = parseSubnetIdFromTopic(rpc.topicID).valueOr:
```
tersec, line 3036: This whole setup reflects a not-ideal API. It's worth figuring out if this can be done in a way more generally useful. In particular, unlike [the subscribe API](https://github.com/vacp2p/nim-libp2p/blob/36717ff7a404ad25fb32b560eb599e81b04cdd37/libp2p/protocols/pubsub/pubsub.nim#L568-L609), it appears to ["always call" the extension RPC handlers](https://github.com/vacp2p/nim-libp2p/blob/36717ff7a404ad25fb32b560eb599e81b04cdd37/libp2p/protocols/pubsub/gossipsub/extensions.nim#L96-L164), and then [always call partial message extension RPC handlers](https://github.com/vacp2p/nim-libp2p/blob/36717ff7a404ad25fb32b560eb599e81b04cdd37/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim#L228-L255) specifically. 

`PartialMessageExtensionConfig` [does have a `nodeTopicOpts` field](https://github.com/vacp2p/nim-libp2p/blob/36717ff7a404ad25fb32b560eb599e81b04cdd37/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim#L20-L55), which already tells `nim-libp2p` which topics are of interest, but it only appears to check this on publish, not the receive RPC. So right now, at least this `parseSubnetIdFromTopic(...)` isn't being repeated, but addressing the apparent asymmetry of [`publishPartial` checking `nodeTopicOpts`](https://github.com/vacp2p/nim-libp2p/blob/36717ff7a404ad25fb32b560eb599e81b04cdd37/libp2p/protocols/pubsub/gossipsub/extension_partial_message.nim#L333-L346) but the receive callbacks not checking it would also mean duplicating work in, e.g., this `handlePartialColumnRPC` as it would have to check the topics yet again and maintain basically its own subscription list.

Furthermore, this isn't even really correct, because now it potentially means that any partial message on column subnet topics Nimbus doesn't subscribe to via https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/nimbus_beacon_node.nim#L1352-L1360 and https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/sync/validator_custody.nim#L299-L308 still make it through libp2p, and would have to be in theory redundantly checked in `handlePartialColumnRPC`/`parseSubnetIdFromTopic`.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -671,6 +704,14 @@ proc proposeBlockAux(
         when fork >= ConsensusFork.Gloas:
           await node.router.routeSignedBeaconBlock(
```
tersec, line 725: Maybe worth a `debugGloasComment` here, because while this PR doesn't have to address the Gloas partial columns, `nimbus-eth2` in general at some point will.

beacon_chain/spec/datatypes/fulu.nim
```diff
@@ -116,7 +116,7 @@ type
   # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/fulu/p2p-interface.md#encoding-and-decoding-responses
```
tersec: This URL does not define the data structure.

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.8/specs/fulu/partial-columns/p2p-interface.md#partialdatacolumnsidecar does

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,232 @@
...
+  var id = newSeqOfCap[byte](PartialGroupIdLen)
+  id.add(PartialGroupIdVersion)
+  id.add(blockRoot.data)
+  id
+
+func parseGroupId*(groupId: GroupId): Result[Eth2Digest, string] =
+  if groupId.len != PartialGroupIdLen:
+    return err("invalid groupId length: " & $groupId.len)
+  if groupId[0] != PartialGroupIdVersion:
+    return err("unsupported groupId version: " & $groupId[0])
```
vladopajic, line 76: this is unnecessary having in mind that anything that comes via rpc will be validated by `validatePartialMessageRPC`. if you really want to duplicate validation, do not duplicate code - extract this in template

beacon_chain/networking/gossip_partial_columns.nim
```diff
@@ -0,0 +1,232 @@
...
+  var root: Eth2Digest
+  for i in 0 ..< 32:
+    root.data[i] = groupId[1 + i]
+  ok(root)
```
vladopajic, line 80: random suggestion: 

`ok(Eth2Digest.init(groupId, 1))`  here last argument `offset: int = 0` so it can be reused 
or 
`ok(Eth2Digest.init(groupId.toOpenArray(1, 32))`

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -447,6 +459,225 @@ template validateBeaconBlockGloas(
   if not (bid.parent_block_root == blck.parent_root):
     return dag.checkedReject("validateBeaconBlockGloas: parent block mismatch")
 
+proc validatePartialDataColumnSidecar*(
```
tersec, line 462: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/fulu/partial-columns/p2p-interface.md#partial-messages-on-data_column_sidecar_subnet_id

PR #8068, opened 2026-03-11
rid the column q check now that q updation is stabler

agnxsh:


tersec:
This has blown up fairly impressively in in the past. Is the testing there so it's less likely to in the future?

PR #8072, opened 2026-03-11
Restart confirmation chain as needed by fast confirmation conditions

etan-status:
`get_latest_confirmed` defines a combination of conditions under which the confirmation chain is restarted from the current epoch observed justified checkpoint:

- https://github.com/ethereum/consensus-specs/pull/4747

Add the checks, both for when epoch advances (first check), and when head changes. Should not trigger under finalizing (safe) conditions, and if it does, gets overwritten by current mainnet logic of simply tracking the justified checkpoint anyway.


tersec:
Merge conflict in `beacon_chain/fork_choice/fork_choice.nim`

PR #8075, opened 2026-03-12
Unify Fulu fork digest handling into BPO list

Tomi-3-0:


beacon_chain/spec/forks.nim
```diff
@@ -1662,21 +1657,15 @@ func forkVersionAtEpoch*(cfg: RuntimeConfig, epoch: Epoch): Version =
...
-    res
-  of ConsensusFork.Fulu:
+  of ConsensusFork.Fulu, ConsensusFork.Gloas:
     var res = FAR_FUTURE_EPOCH
     for entry in cfg.BLOB_SCHEDULE:
       if epoch >= entry.EPOCH:
         break
       res = entry.EPOCH
+    if epoch < cfg.GLOAS_FORK_EPOCH:
+      res = min(res, cfg.GLOAS_FORK_EPOCH)
```
tersec, line 1668: This `min` is basically necessary because it's directly reading off `cfg.BLOB_SCHEDULE`, which is for configuration reasons by necessity constructed as a fairly raw representation of the actual BPO schedule, non-unified. But it clearly suffers from the special casing required and will create problems in Heze due to this. But fine to merge for now, this PR is still a net improvement which unified more of the forkdigest mechanism regarding BPOs.

Tomi-3-0: this: https://github.com/status-im/nimbus-eth2/pull/8133 should solve the special casing required for each subsequent fork

PR #8085, opened 2026-03-14
fix for trusted node sync

Tomi-3-0:


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -306,15 +294,90 @@ proc addHeadBlockWithParent*(
...
+            dag.updateFlags + {skipLastEnvelope}):
+          error "Unable to reload clearance state without envelope",
+            clearanceBlock = shortLog(parentBlock)
+          return err(VerifierError.MissingParent)
+
+        # Advance slots to the incoming block's slot
+        var info: ForkedEpochInfo
+        dag.advanceSlots(
+          dag.clearanceState, signedBlock.message.slot, true, cache, info,
+          dag.updateFlags)
```
ahshum: This is the non-finalized route of handling block, which both block and envelope are strongly required in order to verify the block and envelope in the next slot. So I don't see why we should `{skipLastEnvelope}` here.

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -306,15 +294,90 @@ proc addHeadBlockWithParent*(
...
+
+        info "Applied parent envelope from DB",
+          parentRoot = shortLog(parent.bid.root),
+          slot = signedBlock.message.slot
+
+        # Advance slots to the incoming block's slot
+        var info: ForkedEpochInfo
+        dag.advanceSlots(
+          dag.clearanceState, signedBlock.message.slot, true, cache, info,
+          dag.updateFlags)
```
ahshum: This is handled by updateState() and we should not call this directly. Same to that in case 1.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1033,11 +1033,34 @@ proc applyBlock(
...
-    cache: var StateCache): Result[void, cstring] =
+    cache: var StateCache,
+    nextBid: Opt[BlockId] = Opt.none(BlockId)): Result[void, cstring] =
   withConsensusFork(dag.cfg.consensusForkAtEpoch(bid.slot.epoch)):
     when consensusFork >= ConsensusFork.Gloas:
       let data = dag.db.getExecutionPayloadEnvelope(bid.root).valueOr:
-        return err("Envelope load failed")
+        # Envelope not in DB — builder may not have delivered the payload.
+        # This is normal in ePBS; skip and continue.
+        return ok()
```
ahshum, line 1044: When updateState() fails, `MissingParent` is returned so it is the responsibility of call site to handle it instead of this critical path.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1033,11 +1033,34 @@ proc applyBlock(
...
+              return err("Next block load failed")
+            let envelopeBlockHash =
+              data.message.payload.block_hash
+            let nextParentHash =
+              nextBlock.message.body.signed_execution_payload_bid.message.parent_block_hash
+            if nextParentHash != envelopeBlockHash:
+              # The next block doesn't reference this envelope's block_hash,
+              # meaning the builder didn't deliver. Skip the envelope.
+              return ok()
+
```
ahshum: It doesn't quite making sense that when applying an envelope, it needs to know the block/bid of its descendants. Also there could be more than one descendant.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -773,9 +774,9 @@ proc storeBlock(
     validationDur, queueDur, newPayloadDur, addHeadBlockDur, updateHeadDur
 
   when consensusFork >= ConsensusFork.Gloas:
-    # Enqueue payload here instead of `addBlock` for the consistency of payload
-    # processing with backfilling.
-    self.enqueuePayload(signedBlock)
+    # Process envelope synchronously before returning — the next block in the
+    # queue needs the clearance state to have this block's envelope applied.
+    await self.enqueuePayload(signedBlock)
```
ahshum: Use `storePayload()` if it needs `await` or simply `enqueuePayload()`.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -1034,7 +1032,7 @@ proc addPayload*(
     signedEnvelope: gloas.SignedExecutionPayloadEnvelope,
     sidecarsOpt: Opt[gloas.DataColumnSidecars],
 ): Future[Result[void, VerifierError]] {.async: (raises: [CancelledError]).} =
-  if signedBlock.message.slot <= self.consensusManager.dag.finalizedHead.slot:
+  if signedBlock.message.slot < self.consensusManager.dag.finalizedHead.slot:
```
ahshum: This is hack to devnet or trusted node sync as in Gloas we are updating finalized head in envelope yet https://github.com/status-im/nimbus-eth2/blob/0e0d1fc94f11ed0527681afc8dedefd72830b8bf/beacon_chain/consensus_object_pools/blockchain_dag.nim#L2698

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -1047,21 +1045,38 @@ proc addPayload*(
...
+    envelope = block:
+      var env = self.envelopeQuarantine[].popOrphan(blck)
+      if env.isNone():
+        # Envelope may arrive via gossip shortly after the block; we poll briefly
+        # here (up to ~2s in 100ms steps) to give it time.
+        for i in 0 ..< 20:
+          await sleepAsync(chronos.milliseconds(100))
+          env = self.envelopeQuarantine[].popOrphan(blck)
+          if env.isSome():
+            break
```
ahshum: When envelope arrives from whatever channels it should start processing as long as all required components are there. Can you elaborate why it is needed?

ahshum (commented):
For
- beacon_chain/nimbus_beacon_node.nim
- beacon_chain/trusted_node_sync.nim
etc.

Would be good to have a PR, _based on_ `unstable` and _merge into_ `unstable`

beacon_chain/gossip_processing/block_processor.nim
Tomi-3-0: this was a hack intended to make the clearance state correct for the next block, because withouth the envelope already applied the `clearanceState` will be incomplte.  this is trying to ensure the envelope gets applied to `clearanceState` during `storeBlock` so by the time that `addHeadBlock` returns the state is already fully transitioned. but, i agree with your point that this should happen independently when the envelope arrives.

beacon_chain/gossip_processing/block_processor.nim
Tomi-3-0: one thing to note also is that this isn't intended to be merged as it is. just workarounds that will be cleaned up

beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -234,22 +234,11 @@ proc checkHeadBlock*(
...
-        debug "Parent envelope missing",
-          head = shortLog(dag.head),
-          blckSlot = signedBlock.message.slot,
-          blckRoot = signedBlock.root
-        return err(VerifierError.MissingParent)
+  # In ePBS, the parent's envelope may legitimately be missing if the builder
+  # didn't deliver. The state transition (process_execution_payload_bid) will
+  # handle this correctly by checking bid.parent_block_hash against the actual
+  # state.latest_block_hash, which should reflect only delivered
+  # envelopes, all things being equal.
```
ahshum, line 241: Could you elaborate why is this needed? From what I have tested, syncing and missing do not have issue with this. Probably there is something I missed?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -633,10 +633,13 @@ proc initFullNode(
...
             forkyBlck.message.body.signed_execution_payload_bid
-          if bres.isErr():
+          if bres.isErr() and bres.error() != VerifierError.Duplicate:
             bres
           elif signedEnvelope.isNone():
-            err(VerifierError.Invalid)
+            # No envelope: either Duplicate block, or envelope was not
+            # delivered on the canonical chain (builder didn't fulfill bid).
+            # Both cases are OK — just return the block result.
+            bres
```
ahshum, line 642: It is definitely an `Invalid` as Gloas block requires an envelope. For backfilling, we should add both in once, invalidate the block if envelope is invalid. I think we will need to revisit this but it should not be the issue in devnet.

The validation of envelope in syncing is not complete. I only added a simple check to ensure they have the same length, but they could be from different slot.

tests/testblockutil.nim
```diff
@@ -176,8 +176,9 @@ func makeExecutionPayloadForSigning*(
         calcEip1599BaseFee(latest.gas_limit, latest.gas_used, latest.base_fee_per_gas)
       else:
         EIP1559_INITIAL_BASE_FEE
-    payload.withdrawals =
-      List[capella.Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD](get_expected_withdrawals(state))
+    when consensusFork >= ConsensusFork.Capella:
+      payload.withdrawals =
+        List[capella.Withdrawal, MAX_WITHDRAWALS_PER_PAYLOAD](get_expected_withdrawals(state))
```
ahshum, line 182: I have taken out my changes to `testblockutil`. It should work without this changes if you could check again.

beacon_chain/consensus_object_pools/block_clearance.nim
Tomi-3-0: this previously rejected a Gloas block with `Missing parent` if the payload envelope wasn't already in the DB.  when the builder fails to deliver the envelope, `state.latest_block_hash` doesn't get updated...stays at whatever the last delivered payload hash was. If for example the envelope was never delivered/seen, the `latest_block_hash` should still point to the parent's parent payload, so the state transition should handle this naturally.
if we say the `dag.db.containsExecutionPayloadEnvelope(parent.root())` check should return `MissingParent` for a block whose parent didn't have a payload legitimately, then we send the block to quarantine, and trigger the request to download an nvelope that doesn't exist and the block probably stays in quarantine for a long time

Tomi-3-0: to trigger this you would need a slot were a block was proposed but the envelope wasn't delivered

ahshum: Yes when the parent's envelope of a Gloas block is missing, we shouldn't verify it until we have it. Until we have the parent's envelope, we would not know the correct `latest_block_hash` is so we cannot verify this block yet.

beacon_chain/nimbus_beacon_node.nim
Tomi-3-0: you are right that if an envelope was deliveered for a given block we definitely need it.  but Gloas blocks don't always require an envelope.

Tomi-3-0: the thing here is during syncing, how do we know if the payload for the particular block was actually delivered or doesn't exist anywhere

tests/testblockutil.nim
Tomi-3-0: okay

beacon_chain/nimbus_beacon_node.nim
ahshum: We request data from the same peer. I would say this is a dishonest node in this case and we should discard all of them.

ahshum: These are finalized slots. Block and envelope should either both exist or both not exist. Not only block.

beacon_chain/consensus_object_pools/block_clearance.nim
Tomi-3-0: we should probably have a `MissingParentEnvelope` error for this

beacon_chain/nimbus_beacon_node.nim
Tomi-3-0: why can't a finalised block exist without an envelope? I might have missed something

beacon_chain/consensus_object_pools/block_clearance.nim
ahshum: Refer to the TODO in comments

beacon_chain/nimbus_beacon_node.nim
ahshum: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#beacon_block

Beacon block requires `execution_payload` to be valid. If envelope can be absent, the `block_hash` in bid will be empty which means no EL updates while CL accepts a new block.

PR #8087, opened 2026-03-15
Init proto_array with correct finalized slot on gap

etan-status:
During initialization, we put finalized_checkpoint.epoch.start_slot instead of finalized_blck into the proto_array, which isn't correct when the initial epoch slot was empty. We actually have access to the correct slot in fork_choice, just pass it through.


alialobidm:
just

PR #8090, opened 2026-03-16
fix applying envelope

ahshum:
This PR is depended on
- https://github.com/status-im/nimbus-eth2/pull/7664


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1185,10 +1185,17 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
 
     while headBlocks.len > 0:
+      let blck = headBlocks.pop()
       dag.applyBlock(
-        dag.headState, headBlocks.pop().bid, cache,
+        dag.headState, blck.bid, cache,
         info, dag.updateFlags).expect("head blocks should apply")
 
+      dag.applyExecutionPayloadEnvelope(dag.headState, blck.bid, cache).isOkOr:
+        # Since Gloas, envelopes can only be missing at the head block.
```
tersec: Before Gloas, envelopes didn't even exist. The "Since gloas" qualifier is redundant.

tersec:
Merge conflict with `beacon_chain/consensus_object_pools/envelope_quarantine.nim`.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -1315,11 +1323,27 @@ proc init*(T: type ChainDAGRef, cfg: RuntimeConfig, db: BeaconChainDB,
...
+        elif i > 0:
+          let child = headBlocks[i - 1]
+          isParentBlockFull(dag, child)
+        else:
+          dag.db.containsExecutionPayloadEnvelope(headBlocks[i].root())
+
+      if wantsPayload:
+        dag.applyExecutionPayloadEnvelope(
+          dag.headState, headBlocks[i].bid,
+          cache).expect("head envelopes shoudl apply")
```
tersec: `should`

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -423,22 +423,14 @@ template validateBeaconBlockGloas(
...
-            "validateBeaconBlockGloas: invalid execution parent")
-        if not (bid.parent_block_hash == parentHash):
-          return dag.checkedReject(
-            "validateBeaconBlockGloas: invalid execution parent")
+  let parent = dag.getBlockRef(bid.parent_block_root).valueOr:
+    return dag.checkedReject("validateBeaconBlockGloas: invalid execution parent")
+  debugGloasComment("request missing envelope if not found in db")
+  if not (
+      isParentBlockFull(dag, signed_beacon_block, parent) or
+      isParentBlockFull(dag, signed_beacon_block, parent.parent)
```
tersec, line 431: `parent.parent` can be `nil`, e.g., if `parent` is the finalized head. But in this case, `isParentBlockFull` includes `let (parentBlockHash, _) = dag.loadExecutionAndParentBlockHash(parent)` which
https://github.com/status-im/nimbus-eth2/blob/277e831edc5cfee52726d7ae8aafdefc5dc113e8/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1094-L1096 will try to dereference. It should crash in this situation.

tersec: This does mean the `isParentBlockFull(dag, signed_beacon_block, parent)` has to return false, but it looks like an empty block under the right circumstances should do it.

ahshum: I agree (and left a comment https://github.com/status-im/nimbus-eth2/blob/277e831edc5cfee52726d7ae8aafdefc5dc113e8/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1125-L1127) 

I think it should be handled at the call site as finalized/genesis blocks should be allowed in all of the calls of `isParentBlockFull`, but it isn't reflected precisely by returning false and using `Result` seems to make the function a bit complicated.

I would like to address it in a separate PR if that makes sense.

tersec: Ok, sure.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8090/changes/04f1c7e3c190b51ed5765d387440540904cacfde

PR #8104, opened 2026-03-17
Only update proposer boost if proposer is same as on canonical chain

etan-status:
From https://github.com/ethereum/consensus-specs/pull/4807:

> There's a few changes. Firstly, there's a fix to the mechanism for
> setting the proposer boost root: we only update it if
> `block.proposer_index` matches the expected proposer of your current
> `head_state`. This way you can't nullify proposer boost by just
> creating forks with a divergent proposer sequence. Also small refactor
> with the updating logic moved to `update_proposer_boost_root`, and the
> block timeliness logic in `record_block_timeliness`. All of this is
> essentially unrelated to epbs.

Having to pull a fresh EpochRef is unfortunate, the existing one given to `process_block` is not necessarily compatible with the canonical one. Should be cached, though.


alialobidm (commented):
[](url)

PR #8134, opened 2026-03-19
Reintroduce support to initialize Nimbus in single command

etan-status:
#7888 introduced a regression where the `--external-beacon-api-url` state (combined with `--trusted-state-root` / `--trusted-block-root`) was discarded if not also using Era files. Re-add support.

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -213,7 +213,7 @@ proc setupDatabase(
         if tmp[].slot > checkpointState[].slot:
           checkpointState = tmp
       else:
-        discard  # State from era is newer than the sync target
+        discard  # Remain on state from era (if any)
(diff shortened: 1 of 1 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -203,15 +203,17 @@ proc setupDatabase(
...
+        config.externalBeaconApiUrl.get(), syncTarget[], genesisState)
+      if checkpointState == nil:
+        checkpointState = tmp
+      elif tmp != nil:
         # Special case: we loaded a checkpoint state (for example from era
         # files) and then the remote beacon api gave us a newer one!
         if tmp[].slot > checkpointState[].slot:
           checkpointState = tmp
+      else:
+        discard  # State from era is newer than the sync target
```
```suggestion
      else:
        discard  # Remain on state from era (if any)
```

PR #8137, opened 2026-03-20
find_latest_confirmed_descendant

etan-status:
Implement the main FCR spec function for advancing the EL safe block:

- https://github.com/ethereum/consensus-specs/pull/4747

Note this replaces previous behaviour where the justified checkpoint was tracked as the "safe" block. That was never safe, as justified does not come with reorg guarantees, but it's what the spec mandated.


etan-status:
Needs #8130

PR #8146, opened 2026-03-21
add results to peer lookup and tests

agnxsh:


tests/test_peer_pool.nim
```diff
@@ -1301,3 +1303,138 @@ suite "PeerPool testing suite":
...
+    check:
+      res.status == PeerCgcStatus.Found
+      res.value == CUSTODY_REQUIREMENT
+
+  test "No metadata, no ENR - returns default CUSTODY_REQUIREMENT":
+    let peer = createMinimalPeer()
+    let res = peer.lookupCgcFromPeer()
+    check:
+      res.status == PeerCgcStatus.Found
+      res.value == CUSTODY_REQUIREMENT
```
tersec: The last line does not appear to end with a `\n` (that is, just the newline character itself). This makes reading diffs inconsistent in GitHub's viewer, so it's nicer to consistently make sure that every file ends with the newline character:
Image

beacon_chain/sync/request_manager.nim
```diff
@@ -399,7 +311,13 @@ proc checkPeerCustody(rman: RequestManager,
...
   ## with the peer. Also applies peer scoring.
   var intersection: DataColumnIndices
-  let remoteCustodyGroupCount = peer.lookupCgcFromPeer()
+  let remoteCgcResult = peer.lookupCgcFromPeer()
+  if remoteCgcResult.status != PeerCgcStatus.Found:
+    debug "Failed to lookup cgc from peer",
+      peer = peer, status = remoteCgcResult.status
+    peer.updateScore(PeerScoreNoValues)
+    return intersection
+  let remoteCustodyGroupCount = remoteCgcResult.value
```
tersec: This pattern is partly why `Result` exists though.

That is, yes, on inspection this is correct, but it's still not encapsulating the in-principle-unsafe case object access. `Result` has the `?`, `valueOr`, etc helpers which, yes, ultimately do something similar, but are tested separately to be correct in this capacity. This code contains the equivalent of a direct `foo.get` or `foo.unsafeGet`, even if  it is guarded appropriately.

The purpose of `Result`, then, is partly do allow ultimately something like:
```nim
  let remoteCustodyGroupCount = remoteCgcResult.valueOr:
    debug "Failed to lookup cgc from peer",
      peer = peer, status = remoteCgcResult.status
    peer.updateScore(PeerScoreNoValues)
    return intersection
```
to be written instead.

Ideally, the Nimbus codebase can, consistently with broad efficiency, minimize direct/statically-unsafe (treating `nim-results` as safe, for this purpose, because it is at least tested), avoid this kind of code as much as feasible.

beacon_chain/sync/request_manager.nim
```diff
@@ -454,8 +372,11 @@ proc checkPeerCustody(rman: RequestManager,
 
 func matchIntersection(rman: RequestManager): PeerCustomFilterCallback[Peer] =
   return proc(peer: Peer): bool =
+    let remoteCgcResult = peer.lookupCgcFromPeer()
+    if remoteCgcResult.status != PeerCgcStatus.Found:
+      return false
     let
-      remoteCustodyGroupCount = peer.lookupCgcFromPeer()
+      remoteCustodyGroupCount = remoteCgcResult.value
```
tersec: Similarly to the other comment, this is another equivalent-to-`.get`/`.unsafeGet` code style which could cause a runtime `Defect` and would be improved by using `Result` or `Opt` (i.e. `Result` with an error type of `void`).

tests/test_peer_pool.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8146/commits/1ac610958d88ef23818a5a4e9865695b6cdf2282

beacon_chain/networking/eth2_network.nim
tersec: It seems like `Found` isn't used anymore?

PR #8154, opened 2026-03-23
lsquic: don't link c++ explicitly

arnetheduck:


arnetheduck:
Works with nightlies too: https://github.com/status-im/nimbus-eth2/actions/runs/23442363709

tersec:
Holding this until after the next release, which we might try to do soon, if the current setup works. So sequence is test status quo, if works, then wait on this. If not, merge this. Want to avoid disruptions for the moment.

tersec:
Still relevant?

tersec:
`nim-lsquic` bumped separately.

PR #8168, opened 2026-03-25
bump gloas spec to alpha.4

Tomi-3-0:


tersec:
Merge conflict in the submodule `vendor/nim-ssz-serialization`.

tersec:
```
Run excluded_files="config.yaml|config.nims|beacon_chain.nimble|Makefile|Product.wxs"
- tests/consensus_spec/fulu/test_fixture_ssz_consensus_objects.nim
- tests/consensus_spec/gloas/test_fixture_state_transition_epoch.nim
```

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2283,71 @@ func onboard_builders_from_pending_deposits*(
...
+  let
+    epoch = slot.epoch()
+    state_epoch = get_current_epoch(state)
+    index =
+      if epoch < state_epoch:
+        doAssert epoch + 1 == state_epoch
+        slot mod SLOTS_PER_EPOCH
+      else:
+        doAssert epoch <= state_epoch + MIN_SEED_LOOKAHEAD
+        (epoch - state_epoch + 1) * SLOTS_PER_EPOCH + (slot mod SLOTS_PER_EPOCH)
```
tersec: In general, Nimbus typically uses https://github.com/status-im/nimbus-eth2/blob/9bc62ee6f123becfc8bc0dac61e51d86150383ca/beacon_chain/spec/beacon_time.nim#L281-L286 to express this `* SLOTS_PER_EPOCH` computation when it appears in the specs.

beacon_chain/spec/state_transition_block.nim
tersec: Looks like https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#modified-process_deposit_request and https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md#modified-process_deposit_request are different?

beacon_chain/spec/validator.nim
```diff
@@ -498,6 +497,10 @@ iterator compute_balance_weighted_selection*(
   let total = indices.lenu64
   doAssert total > 0
 
+  var effective_balances = newSeqOfCap[Gwei](total)
+  for idx in indices:
+    effective_balances.add(state.validators[idx].effective_balance)
```
tersec: i.e. this is a translation of
```python
    effective_balances = [state.validators[index].effective_balance for index in indices]
```
It's arguably simpler and corresponds more closely to the list comprehension of the spec to use
```nim
  let effective_balances = indices.mapIt(state.validators[it].effective_balance)
```
And, `mapIt` does preallocate a sequence of the correct size, just like `newSeqOfCap`, so it's also memory-allocation-efficient.

I do wonder if it doesn't make more sense given to use instead of
```nim
    if compute_balance_weighted_acceptance(effective_balances[next_index], seed, i):
```
somewhere where it looks up that particular `effective_balances[next_index]` as it needs it.

In terms of performance, this isn't trivial to benchmark properly, and the intermediate sequence doesn't take enormous amounts of memory either, but would tend to bias towards not pointlessly allocating where not necessary (i.e. the thing which is happening) unless there is a demonstrated memory access speed/locality  improvement.

`compute_proposer_indices` only calls this with `size=1`, so most of those lookups are just going to end up not even used. The other callers/users request `PTC_SIZE` (i.e. 512) of them, but that still is often going to be a subset of the entire set of indices provided, so most of the lookups will be wasted in a sufficiently large network.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -509,12 +535,37 @@ proc check_voluntary_exit*(
...
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md#modified-process_voluntary_exit
 proc process_voluntary_exit*(
     cfg: RuntimeConfig,
     state: var ForkyBeaconState,
     signed_voluntary_exit: SomeSignedVoluntaryExit,
     flags: UpdateFlags, exit_queue_info: ExitQueueInfo,
     cache: var StateCache): Result[ExitQueueInfo, cstring] =
+
+  when typeof(state).kind >= ConsensusFork.Gloas:
+    let voluntary_exit = signed_voluntary_exit.message
```
tersec: Better to use the `template voluntary_exit ...` approach here to avoid a pointless copy.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1396,11 +1448,11 @@ func process_withdrawals*(
   ok()
 
 # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-is_builder_index
-func is_builder_index(validator_index: uint64): bool =
+func is_builder_index*(validator_index: uint64): bool =
```
tersec: This doesn't appear used outside this module?
```
$ rg is_builder_index
beacon_chain/spec/state_transition_block.nim
548:    if is_builder_index(voluntary_exit.validator_index):
1450:# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-is_builder_index
1451:func is_builder_index*(validator_index: uint64): bool =
1464:    if is_builder_index(withdrawal.validator_index):
```

Not sure, therefore, why it should be exported.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -1396,11 +1448,11 @@ func process_withdrawals*(
...
   (validator_index and BUILDER_INDEX_FLAG) != 0
 
 # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-convert_validator_index_to_builder_index
-func convert_validator_index_to_builder_index(validator_index: uint64): BuilderIndex =
+func convert_validator_index_to_builder_index*(validator_index: uint64): BuilderIndex =
```
tersec: Not used outside this module?
```
$ rg convert_validator_index_to_builder_index
beacon_chain/spec/state_transition_block.nim
552:        convert_validator_index_to_builder_index(
1454:# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-convert_validator_index_to_builder_index
1455:func convert_validator_index_to_builder_index*(validator_index: uint64): BuilderIndex =
1467:          convert_validator_index_to_builder_index(withdrawal.validator_index)
```

So unsure of reason for exporting.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -496,7 +522,7 @@ proc check_voluntary_exit*(
         voluntary_exit_fork, state.genesis_validators_root, voluntary_exit,
         validator[].pubkey, signed_voluntary_exit.signature):
       return err("Exit: invalid signature")
-
+
```
tersec: ?

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -1389,6 +1389,30 @@ func process_builder_pending_payments*(
...
+  let
+    current_epoch = state.get_current_epoch()
+    new_epoch = current_epoch + MIN_SEED_LOOKAHEAD + 1
+
+  # Shift all epochs forward by one
+  for i in 0 ..< (1 + MIN_SEED_LOOKAHEAD) * SLOTS_PER_EPOCH:
+    state.ptc_window[i] = state.ptc_window.item(i + SLOTS_PER_EPOCH)
+
+  # Fill in the last epoch
+  let base_index = (1 + MIN_SEED_LOOKAHEAD) * SLOTS_PER_EPOCH
```
tersec: `const`

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -308,11 +309,16 @@ suite baseDescription & "Voluntary Exit " & preset():
...
-    runTest[SignedVoluntaryExit, typeof applyVoluntaryExit](
-      OpVoluntaryExitDir, suiteName, "Voluntary Exit", "voluntary_exit",
-      applyVoluntaryExit, path)
-
+  for path in walkTests(OpVoluntaryExitDir):                                                                                                                              
+    if not fileExists(
+        OpVoluntaryExitDir / "pyspec_tests" / path / "voluntary_exit.ssz_snappy"):                                                                                        
+      test "[Skip]    " & baseDescription & "Voluntary Exit - " & path:                                                                                                   
+        skip()                                                                                                                                                            
+      continue
```
tersec: It seems like there might be a 1:1 matching here?
```
$ find mainnet/gloas/operations/voluntary_exit/pyspec_tests/ -maxdepth 1 -mindepth 1 | wc -l
31
$ find mainnet/gloas/operations/voluntary_exit/pyspec_tests/ -name voluntary_exit.ssz_snappy | wc -l
31
$ find minimal/gloas/operations/voluntary_exit/pyspec_tests/ -maxdepth 1 -mindepth 1 | wc -l
27
$ find minimal/gloas/operations/voluntary_exit/pyspec_tests/ -name voluntary_exit.ssz_snappy | wc -l
27
```
Each of mainnet and minimal at least have the same number of `voluntary_exit.ssz_snappy` files as they do tests, and nothing in the preset-dependent test summaries indicated that this `skip()` codepath activated.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2283,71 @@ func onboard_builders_from_pending_deposits*(
...
+
+  var res: seq[ValidatorIndex]
+  for idx in state.ptc_window[index]:
+    res.add(ValidatorIndex(idx))
```
tersec: It seems like every caller of this uses it directly in a `for` loop, i.e. it can be an `iterator` and avoid this `seq` accumulation entirely, with no change in any of its callers.
```
$ rg get_ptc
tests/test_payload_attestation_pool.nim
24:from ../beacon_chain/spec/beaconstate import get_ptc
97:        for validator_index in get_ptc(forkyState.data, slot):
133:        for validator_index in get_ptc(forkyState.data, slot):
166:        for validator_index in get_ptc(forkyState.data, slot):
213:        for validator_index in get_ptc(forkyState.data, slot):
239:        for validator_index in get_ptc(forkyState.data, slot):
269:        for validator_index in get_ptc(forkyState.data, slot):
318:  test "get_ptc with ShufflingRef matches StateCache version" & preset():
326:          for validator_index in get_ptc(forkyState.data, slot):
334:          for validator_index in get_ptc(

AllTests-mainnet.md
967:+ get_ptc with ShufflingRef matches StateCache version [Preset: mainnet]                     OK

beacon_chain/validators/beacon_validators.nim
955:      for vidx in get_ptc(forkyState.data, slot):
1434:        for validator_index in get_ptc(

beacon_chain/consensus_object_pools/payload_attestation_pool.nim
19:from ../spec/beaconstate import get_ptc
98:      for ptc_validator_index in get_ptc(forkyState.data, slot):

beacon_chain/gossip_processing/gossip_validation.nim
2129:  # `get_ptc(state, data.slot)`. The `state` is the head state corresponding to
2136:      for idx in get_ptc(forkyState.data, data.slot):

beacon_chain/consensus_object_pools/spec_cache.nim
232:iterator get_ptc*(

research/block_sim.nim
42:  get_committee_indices, get_ptc
257:          for validator_index in get_ptc(forkyState.data, slot):

beacon_chain/spec/beaconstate.nim
2310:# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md#new-get_ptc
2311:func get_ptc*(state: gloas.BeaconState, slot: Slot):
3086:  for index in get_ptc(state, slot):
```

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2283,71 @@ func onboard_builders_from_pending_deposits*(
...
+    let committee = get_beacon_committee(state, slot, committee_index, cache)
+    indices.add(committee)
+
+  for candidate_index in compute_balance_weighted_selection(
+      state, indices, seed, size=PTC_SIZE, shuffle_indices=false):
+    yield candidate_index
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md#new-get_ptc
+func get_ptc*(state: gloas.BeaconState, slot: Slot):
+    seq[ValidatorIndex]=
```
tersec: Missing space between `seq[ValidatorIndex]` and `=`

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2283,71 @@ func onboard_builders_from_pending_deposits*(
...
+# {.closure.} prevents stack overflow from inline expansion.
+# See: https://github.com/nim-lang/Nim/issues/25287
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md#new-compute_ptc
+iterator compute_ptc*(state: gloas.BeaconState, slot: Slot, cache: var StateCache):
+    ValidatorIndex {.closure.} =
+  ## Get the payload timeliness committee for the given ``slot``.
+  let epoch = slot.epoch()
+  var buffer {.noinit.}: array[40, byte]
+  buffer[0..31] = get_seed(state, epoch, DOMAIN_PTC_ATTESTER).data
+  buffer[32..39] = uint_to_bytes(slot.uint64)
```
tersec: It's slightly more typesafe to use `distinctBase` than `uint64` here, because that guarantees in a static way that the Nim compiler believes that converting from `slot` to a `uint64` version of `slot` is a lossless, safe, free conversion.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2283,71 @@ func onboard_builders_from_pending_deposits*(
...
+func get_ptc*(state: gloas.BeaconState, slot: Slot):
+    seq[ValidatorIndex]=
+  ## Get the payload timeliness committee for the given ``slot``
+  let
+    epoch = slot.epoch()
+    state_epoch = get_current_epoch(state)
+    index =
+      if epoch < state_epoch:
+        doAssert epoch + 1 == state_epoch
+        slot mod SLOTS_PER_EPOCH
```
tersec: Well, 5 lines up one already has this value computed.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1387,6 +1396,16 @@ proc validateAggregate*(
...
+
+    # [REJECT] If attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block passes validation.
+    # [IGNORE] When attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block has been seen
+    if aggregate.data.index == 1:
+      if not pool.dag.db.containsExecutionPayloadEnvelope(
+        aggregate.data.beacon_block_root):
+          return errIgnore(
+            "Aggregate: execution payload not yet seen")
```
ahshum: indent

```suggestion
    if aggregate.data.index == 1:
      if not pool.dag.db.containsExecutionPayloadEnvelope(
          aggregate.data.beacon_block_root):
        return errIgnore(
          "Aggregate: execution payload not yet seen")
```

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1187,6 +1187,15 @@ proc validateAttestation*(
...
         )
+    # [REJECT] If attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block passes validation.
+    # [IGNORE] When attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block has been seen
+    if attestation.data.index == 1:
+      if not pool.dag.db.containsExecutionPayloadEnvelope(
+        attestation.data.beacon_block_root):
+        return errIgnore(
+         "SingleAttestation: execution payload not yet seen")
```
ahshum: indent

```suggestion
    if attestation.data.index == 1:
      if not pool.dag.db.containsExecutionPayloadEnvelope(
          attestation.data.beacon_block_root):
        return errIgnore(
          "SingleAttestation: execution payload not yet seen")
```

beacon_chain/spec/state_transition_block.nim
```diff
@@ -443,7 +461,15 @@ func process_deposit_request*(
...
-    ok()
+      if verify_deposit_signature(
+          cfg.GENESIS_FORK_VERSION,
+          DepositData(
+            pubkey: deposit_request.pubkey,
+            withdrawal_credentials: deposit_request.withdrawal_credentials,
+            amount: deposit_request.amount,
+            signature: deposit_request.signature)):
+        pending_validators.incl(deposit_request.pubkey)
+      ok()
```
ahshum: should these 2 spaces left?

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -308,11 +309,11 @@ suite baseDescription & "Voluntary Exit " & preset():
...
-  for path in walkTests(OpVoluntaryExitDir):
-    runTest[SignedVoluntaryExit, typeof applyVoluntaryExit](
-      OpVoluntaryExitDir, suiteName, "Voluntary Exit", "voluntary_exit",
-      applyVoluntaryExit, path)
-
+  for path in walkTests(OpVoluntaryExitDir):                                                                                                                                                        
+    runTest[SignedVoluntaryExit, typeof applyVoluntaryExit](                                                                                                              
+      OpVoluntaryExitDir, suiteName, "Voluntary Exit", "voluntary_exit",                                                                                                  
+      applyVoluntaryExit, path)                                                                                                                                           
+
```
ahshum: pure whitespace change?

Tomi-3-0: thanks

beacon_chain/spec/validator.nim
Tomi-3-0: Tried benchmarking the three approaches (pre-allocated sequence, mapIt, and the on-demand lookup) with  a 200k validators and the on-demand lookup is about 18x faster.
 `Scenario: PTC (size=512, no shuffle)  | Pre-allocated: 2814us  | mapIt: 2577us  | On-demand: 156us`

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1187,6 +1187,15 @@ proc validateAttestation*(
...
         )
+    # [REJECT] If attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block passes validation.
+    # [IGNORE] When attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block has been seen
+    if attestation.data.index == 1:
+      if not pool.dag.db.containsExecutionPayloadEnvelope(
+          attestation.data.beacon_block_root):
+        return errIgnore(
+          "SingleAttestation: execution payload not yet seen")
```
ahshum, line 1201: When an envelope exists in the db, the payload passes validation. So this is rejected.

Regarding the payload has been seen, it should check with the envelope quarantine.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1387,6 +1396,16 @@ proc validateAggregate*(
...
+
+    # [REJECT] If attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block passes validation.
+    # [IGNORE] When attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block has been seen
+    if aggregate.data.index == 1:
+      if not pool.dag.db.containsExecutionPayloadEnvelope(
+          aggregate.data.beacon_block_root):
+        return errIgnore(
+          "Aggregate: execution payload not yet seen")
```
ahshum, line 1413: ^ same as above

beacon_chain/gossip_processing/gossip_validation.nim
Tomi-3-0: my main reason for this approach is simplicity.  In the common case, if the envelope is in the db, it passed validation. if it's not anywhere, it gets `IGNORED`  whether it has truly not been seen or just still in the quarantine. 

for the envelope quarantine check  we might have to pass something like envelopeQuaranitne: ref EnvelopeQuarantine` to the function signature or pass it as a field to the AttestationPool. I should probably add a TODO and address it in a follow up

ahshum: I see but by reading from db it should cover only the rejected case, isn't?

Tomi-3-0: yes. correct

Tomi-3-0: https://github.com/status-im/nimbus-eth2/pull/8168/commits/6e60d6518f4f0770200f9f0b5b8a967d10d707d3
                                                  
 added the quarantine check  because at least fixes the IGNORE case properly: envelope seen but not yet in DB we don't ignore and let it through.

ahshum: `Eth2Processor` should have all the deps so I think it makes more sense to pass envelope quarantine into the `validate*` function if attestation pool doesn't need it.

And should there be two separate errors, one for the reject and one for the ignore?

Tomi-3-0: the spec does say to REJECT if the payload didn't pass validation. we just ignore everything that's not in the DB. There's no tracking of "envelope was seen and failed validation" to distinguish REJECT from IGNORE.

ahshum: Oh yeah you're right

Tomi-3-0: also, i think the field approach is simpler since that way we avoid updating all the callers and since there's no signifcant performance benefit from passing the env quarantine

ahshum: Not sure if I understand correctly, for example `validateAttestation`, is only being used in `eth2_processor` and tests. 

```
$ rg -i 'validateAttestation\('
docs/attestation_flow.dot
11:  GossipSub -> Eth2Processor [label="getAttestationTopic: attestationValidator->attestationPool.validateAttestation()"];
14:  Eth2Processor -> AttestationAggregation [label="validateAttestation() validateAggregate()"];
18:  Eth2Processor -> AsyncQueueAttestationEntry [label="Move to queue after P2P validation via validateAttestation()"];

tests/test_gossip_validation.nim
116:      validateAttestation(pool, batchCrypto, att_1_0, beaconTime, subnet, true).waitFor().isOk
119:      validateAttestation(pool, batchCrypto, att_1_0, beaconTime, subnet, true).waitFor().error()[0] ==
125:      validateAttestation(
131:      validateAttestation(
137:      validateAttestation(
148:        validateAttestation(
158:        fut_1_0 = validateAttestation(
160:        fut_1_1 = validateAttestation(
174:        fut_1_0 = validateAttestation(
176:        fut_1_1 = validateAttestation(
196:        fut_1_0 = validateAttestation(
198:        fut_1_1 = validateAttestation(

docs/attestation_flow.md
53:The attestations are then validated by `validateAttestation()` or `validateAggregate()` in either `attestationValidator()` or `aggregateValidator()` according to the P2P specs.

beacon_chain/gossip_processing/eth2_processor.nim
577:    await self.attestationPool.validateAttestation(
```

Regards to the point of performance benefit, would you mind further elaborate please?

Tomi-3-0: my point is we would have to update all this callers if we decide to pass the envelope quarantine parameter: `11:  GossipSub -> Eth2Processor [label="getAttestationTopic: attestationValidator->attestationPool.validateAttestation()"];
14:  Eth2Processor -> AttestationAggregation [label="validateAttestation() validateAggregate()"];
18:  Eth2Processor -> AsyncQueueAttestationEntry [label="Move to queue after P2P validation via validateAttestation()"];

tests/test_gossip_validation.nim
116:      validateAttestation(pool, batchCrypto, att_1_0, beaconTime, subnet, true).waitFor().isOk
119:      validateAttestation(pool, batchCrypto, att_1_0, beaconTime, subnet, true).waitFor().error()[0] ==
125:      validateAttestation(
131:      validateAttestation(
137:      validateAttestation(
148:        validateAttestation(
158:        fut_1_0 = validateAttestation(
160:        fut_1_1 = validateAttestation(
174:        fut_1_0 = validateAttestation(
176:        fut_1_1 = validateAttestation(
196:        fut_1_0 = validateAttestation(
198:        fut_1_1 = validateAttestation(`

> Regards to the point of performance benefit, would you mind further elaborate please?

meaning I don't see any performance gain that would justify passing the envelope  quarantine through all these call sites.

ahshum: I would say that is the pattern of `gossip_validation` for dependency injection.

Tomi-3-0: https://github.com/status-im/nimbus-eth2/pull/8168/commits/51ad833f1a3326281cc0301dd8a920fe1ddb7e1c

beacon_chain/consensus_object_pools/attestation_pool.nim
```diff
@@ -120,7 +120,8 @@ func init(T: type AttestationData, entry: AttestationEntry): T =
   )
 
 proc init*(T: type AttestationPool, dag: ChainDAGRef,
-           quarantine: ref Quarantine, wallTime = default(BeaconTime),
+           quarantine: ref Quarantine,
+           wallTime = default(BeaconTime),
```
ahshum: I think this should be reverted as part of previous changes?

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1187,6 +1188,17 @@ proc validateAttestation*(
...
+    # the execution payload for block passes validation.
+    # [IGNORE] When attestation.data.index == 1 (payload present for a past block),
+    # the execution payload for block has been seen
+    if attestation.data.index == 1:
+      template block_root: untyped = attestation.data.beacon_block_root
+      if not pool.dag.db.containsExecutionPayloadEnvelope(block_root) and
+          (envelopeQuarantine.isNil or
+            block_root notin envelopeQuarantine[].orphans):
+        return errIgnore(
+          "SingleAttestation: execution payload not yet seen")
```
ahshum, line 1201: If `envelopeQuarantine.isNil` is only useful in tests, why not passing a non-nil envelopeQuarantine instead?

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -557,7 +557,8 @@ proc initFullNode(
       Quarantine.init(dag.cfg))
     envelopeQuarantine = newClone(EnvelopeQuarantine.init(onEnvelopeAdded))
     attestationPool = newClone(AttestationPool.init(
-      dag, quarantine, getBeaconTime(), onSingleAttestationReceived))
+      dag, quarantine,
+      getBeaconTime(), onSingleAttestationReceived))
```
ahshum: I think this should be reverted as part of previous changes?

Tomi-3-0: you mean formatting?

ahshum (approved):

beacon_chain/spec/state_transition_block.nim
```diff
@@ -456,7 +482,7 @@ proc check_voluntary_exit*(
     signed_voluntary_exit: SomeSignedVoluntaryExit,
     flags: UpdateFlags): Result[ValidatorIndex, cstring] =
 
-  let voluntary_exit = signed_voluntary_exit.message
+  template voluntary_exit:untyped = signed_voluntary_exit.message
```
ahshum: space missing in `:untyped`

beacon_chain/nimbus_beacon_node.nim
tersec: Yeah, I'm not sure why this was changed.

beacon_chain/consensus_object_pools/attestation_pool.nim
tersec: And not sure why this was changed either, yeah.

beacon_chain/spec/state_transition_block.nim
```diff
@@ -312,6 +312,22 @@ proc process_attester_slashing*(
 from ".."/validator_bucket_sort import
   BucketSortedValidators, add, findValidatorIndex, sortValidatorBuckets
 
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.3/specs/gloas/beacon-chain.md#new-is_pending_validator
```
tersec: alpha.4 is current/target version.

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -1389,6 +1389,30 @@ func process_builder_pending_payments*(
...
+  const base_index = (1 + MIN_SEED_LOOKAHEAD) * SLOTS_PER_EPOCH
+  for slot_offset in 0 ..< SLOTS_PER_EPOCH:
+    let slot = new_epoch.start_slot() + slot_offset
+    var
+      ptcArray: HashArray[Limit PTC_SIZE, uint64]
+      i = 0
+    for idx in compute_ptc(state, slot, cache):
+      ptcArray[i] = uint64(idx)
+      inc i
+    state.ptc_window.mitem(base_index + slot_offset) = ptcArray
```
tersec: This currently does an extra copy, i.e. it constructs each subset `HashArray` of the PTC lookahead then copies that. But it could fill in the target directly.

The challenge is to avoid constant hash recomputation, but that can be done manually.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2291,76 @@ func onboard_builders_from_pending_deposits*(
...
+      base_index = (1 + epoch_offset) * SLOTS_PER_EPOCH
+    for slot_offset in 0'u64 ..< SLOTS_PER_EPOCH:
+      let slot = epoch.start_slot() + slot_offset
+      var
+        ptcArray: HashArray[Limit PTC_SIZE, uint64]
+        i = 0
+      for idx in compute_ptc(state, slot, cache):
+        ptcArray[i] = uint64(idx)
+        inc i
+      state.ptc_window.mitem(base_index + slot_offset) = ptcArray
```
tersec: See:
- https://github.com/status-im/nimbus-eth2/pull/8168/changes#r3031997501

regarding how it does extra copies/allocations.

beacon_chain/spec/beaconstate.nim
```diff
@@ -1757,6 +1768,14 @@ func convert_builder_index_to_validator_index(builder_index: BuilderIndex):
     uint64 =
   builder_index or BUILDER_INDEX_FLAG
 
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-convert_validator_index_to_builder_index
+func convert_validator_index_to_builder_index*(validator_index: uint64): BuilderIndex =
+  validator_index and not BUILDER_INDEX_FLAG
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/gloas/beacon-chain.md#new-is_builder_index
```
tersec: newly added URLs/functions (or sufficiently refactored ones) should be checked for alpha.4 and the URLs reflecting such; this applies to both `convert_validator_index_to_builder_index` and `is_builder_index`.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: One general point about the quarantine is that things in quarantine are not fully validated/checked. So it can be useful, but also, one has to be intentional/conscious of the security posture

beacon_chain/consensus_object_pools/attestation_pool.nim
Tomi-3-0: I previously added the envelope quarantine field to the attestation pool, and this was an artifact from the change. reverting it

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/61a00bda69dd271f9151c37936448bbba2e3fd21

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/state_transition_block.nim
```diff
@@ -312,6 +312,22 @@ proc process_attester_slashing*(
 from ".."/validator_bucket_sort import
   BucketSortedValidators, add, findValidatorIndex, sortValidatorBuckets
 
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md#new-is_pending_validator
+func get_pending_validators*(cfg: RuntimeConfig, state: gloas.BeaconState):
+    HashSet[ValidatorPubKey] =
+  ## Check if a pending deposit with a valid signature is in the queue for the given pubkey.
```
tersec: That's not really what this function does anymore. It's what `is_pending_validator(...)` does, yes, but it's not what `get_pending_validators(...)` does.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2291,73 @@ func onboard_builders_from_pending_deposits*(
...
+    slot_in_epoch = slot mod SLOTS_PER_EPOCH
+
+  if epoch < state_epoch and epoch + 1 != state_epoch:
+    return
+  if epoch >= state_epoch and epoch > state_epoch + MIN_SEED_LOOKAHEAD:
+    return
+
+  let index =
+    if epoch < state_epoch:
+      slot_in_epoch
```
tersec: The spec only has the
```python
    offset = (epoch - state_epoch + 1) * SLOTS_PER_EPOCH
```
case. Why this case here, is it some case of Python using effectively signed arithmetic, and that creating different results, or?

beacon_chain/spec/state_transition_block.nim
```diff
@@ -509,12 +535,37 @@ proc check_voluntary_exit*(
...
+      if not is_active_builder(state, builder_index):
+        return err("Exit: builder not active")
+      if get_pending_balance_to_withdraw_for_builder(
+          state, builder_index) != 0.Gwei:
+        return err("Exit: builder has pending withdrawals")
+      let voluntary_exit_fork = typeof(state).kind.voluntary_exit_signature_fork(
+        state.fork, cfg.CAPELLA_FORK_VERSION)
+      if not verify_voluntary_exit_signature(
+          voluntary_exit_fork, state.genesis_validators_root, voluntary_exit,
+          state.builders.mitem(builder_index).pubkey,
```
tersec: Why `mitem` here? It shouldn't be modifying any of this.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2291,73 @@ func onboard_builders_from_pending_deposits*(
...
+    yield ValidatorIndex(idx)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/fork.md#new-initialize_ptc_window
+proc initialize_ptc_window(
+    state: var gloas.BeaconState, cache: var StateCache) =
+  ## Return the cached PTC window starting from the current epoch.
+  ## Used to initialize the ``ptc_window`` field in the beacon state
+  ## at genesis and after forks.
+  let current_epoch = state.get_current_epoch()
+  for epoch_offset in 0'u64 .. MIN_SEED_LOOKAHEAD:
```
tersec: This is a good candidate for [`staticFor`](https://github.com/status-im/nim-stew/blob/master/stew/staticfor.nim), which unrolls `for` loops to skip some runtime bounds checks and enable the C compiler to perform more optimizations. See [its tests](https://github.com/status-im/nim-stew/blob/master/tests/test_staticfor.nim) for how to use it.

It only works where the loop bounds are statically known at compile-time, and it tends not to make a lot of sense with nontrivially large bounds (e.g., I wouldn't necessarily use it on `SLOTS_PER_EPOCH` loops, when that can be 32). But `MIN_SEED_LOOKAHEAD` loops are ideal: static and only two iterations, while Nim's codegen for them is inefficient and somewhat C-compiler-hostile.

beacon_chain/spec/beaconstate.nim
Tomi-3-0: Yes, it's the unsigned arithmetic case.

beacon_chain/spec/state_transition_epoch.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/e43d30b5f13238d898fc12644d40223a2412b3f6

beacon_chain/spec/beaconstate.nim
Tomi-3-0: But I could actually avoid the extra case altogether since adding 1 to epoch first ensures the subtraction - state_epoch never underflows

tests/consensus_spec/gloas/test_fixture_operations.nim
```diff
@@ -186,13 +186,14 @@ suite baseDescription & "Deposit " & preset():
       OpDepositsDir, suiteName, "Deposit", "deposit", applyDeposit, path)
 
 suite baseDescription & "Deposit Request " & preset():
-  func applyDepositRequest(
+  proc applyDepositRequest(
```
tersec: `func` still seems to work here, why changed?

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/cd5f7ee29c68e219aed99cec80eb71af6b514c2e

tests/consensus_spec/gloas/test_fixture_operations.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/cd5f7ee29c68e219aed99cec80eb71af6b514c2e

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/cd5f7ee29c68e219aed99cec80eb71af6b514c2e

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/cd5f7ee29c68e219aed99cec80eb71af6b514c2e

beacon_chain/spec/state_transition_block.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/cd5f7ee29c68e219aed99cec80eb71af6b514c2e

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2291,70 @@ func onboard_builders_from_pending_deposits*(
...
+proc initialize_ptc_window(
+    state: var gloas.BeaconState, cache: var StateCache) =
+  ## Return the cached PTC window starting from the current epoch.
+  ## Used to initialize the ``ptc_window`` field in the beacon state
+  ## at genesis and after forks.
+  let current_epoch = state.get_current_epoch()
+  staticFor epoch_offset, 0 .. MIN_SEED_LOOKAHEAD.int:
+    let
+      epoch = current_epoch + epoch_offset
+      base_index = (1 + epoch_offset) * SLOTS_PER_EPOCH
```
tersec: This is a `const` now, because of `staticFor`.

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2291,70 @@ func onboard_builders_from_pending_deposits*(
...
+    return
+
+  let index =
+    (epoch + 1 - state_epoch).Epoch.start_slot.uint64 + slot_in_epoch
+
+  for idx in state.ptc_window[index]:
+    yield ValidatorIndex(idx)
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/fork.md#new-initialize_ptc_window
+proc initialize_ptc_window(
```
tersec: `func`

beacon_chain/spec/beaconstate.nim
tersec: It doesn't look like this still needs this `cache` parameter at all.

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/1d10dd4c2c7ec7c2f04fd5b8ea98e53c009175c5

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/commits/406a540032cae837d028a447d184078e0653fa29

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/commits/406a540032cae837d028a447d184078e0653fa29

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/commits/406a540032cae837d028a447d184078e0653fa29

beacon_chain/spec/state_transition_block.nim
tersec: In a cascading way due to the removal of it as a parameter from `get_indexed_payload_attestation(...)`, `process_payload_attestation(...)` doesn't need a `StateCache` parameter either anymore.

tersec: https://github.com/status-im/nimbus-eth2/pull/8168/commits/18cbbf4517635a77b03d4a46eb386009983d0709

beacon_chain/spec/beaconstate.nim
```diff
@@ -2272,6 +2291,69 @@ func onboard_builders_from_pending_deposits*(
...
+  ## at genesis and after forks.
+  let current_epoch = state.get_current_epoch()
+  staticFor epoch_offset, 0 .. MIN_SEED_LOOKAHEAD.int:
+    let epoch = current_epoch + epoch_offset
+    const base_index = (1 + epoch_offset) * SLOTS_PER_EPOCH
+    for slot_offset in 0'u64 ..< SLOTS_PER_EPOCH:
+      let slot = epoch.start_slot() + slot_offset
+      var i = 0
+      for idx in compute_ptc(state, slot, cache):
+        state.ptc_window.mitem(base_index + slot_offset)[i] = uint64(idx)
```
tersec: https://github.com/status-im/nim-ssz-serialization/blob/f422a3b47d89642983a6c2dcda5d91836b3dbb17/ssz_serialization/types.nim#L728-L731
```nim
func mitem*(a: var HashArray, b: auto): var a.T =
  # Access mutable item clearing its cache
  clearCaches(a, b.Limit)
  a.data[b]
```
i.e. every `mitem` call triggers a `clearCaches()` every inner `for` loop iteration, pointlessly. It should happen exactly once, across each changed, `state.ptc_window`, before the loop. Each `state.ptc_window` entry which gets replaced gets replaced fully, so that part at least straightforward.

beacon_chain/spec/state_transition_epoch.nim
```diff
@@ -1389,6 +1389,27 @@ func process_builder_pending_payments*(
...
+  for i in 0 ..< (1 + MIN_SEED_LOOKAHEAD) * SLOTS_PER_EPOCH:
+    state.ptc_window[i] = state.ptc_window.item(i + SLOTS_PER_EPOCH)
+
+  # Fill in the last epoch
+  const base_index = (1 + MIN_SEED_LOOKAHEAD) * SLOTS_PER_EPOCH
+  for slot_offset in 0 ..< SLOTS_PER_EPOCH:
+    let slot = new_epoch.start_slot() + slot_offset
+    var i = 0
+    for idx in compute_ptc(state, slot, cache):
+      state.ptc_window.mitem(base_index + slot_offset)[i] = uint64(idx)
```
tersec: See https://github.com/status-im/nimbus-eth2/pull/8168/changes#r3035876724 for basically the same observation about `mitem` triggering a lot of extra `clearCaches(...)`.

beacon_chain/spec/beaconstate.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/aa6c908fbd73203fecf5466d9bba7e2dcc2ece9c

beacon_chain/spec/state_transition_epoch.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8168/changes/aa6c908fbd73203fecf5466d9bba7e2dcc2ece9c

Issue #1743, opened 2020-09-24
Enable LTO builds for Windows


arnetheduck:
https://github.com/status-im/nim-beacon-chain/pull/1661 enables LTO for linux and mac, but omits windows due to linker-related difficulties - these should be resolved, either by setting up the right paths or switching to clang.

tersec:
- https://github.com/status-im/nimbus-eth2/pull/8529

PR #8172, opened 2026-03-26
set lto based on compiler, not platform

arnetheduck:
ie clang on linux should also use clang lto syntax


arnetheduck:
fixes https://github.com/status-im/nimbus-eth2/issues/1743

PR #8204, opened 2026-03-31
Early reject data with incorrect fork digest

etan-status:
After reading response data, check that the given ForkDigest actually matches the expected ConsensusFork, same as how these checks are already done for light client data. For example, a Fulu block mislabeled as an Electra block. Note: This still allows forkdigests from conflicting blob schedules through, which is intentional, as a block can still be valid according to our local forkdigest, we can just relabel it when others request it from us.


etan-status:
Example: Someone relabels a Fulu block as Electra --> it parses fine, then state_transition_block does a `when consensusFork == type(signedBlock).kind` and returns error. This just catches it earlier.

PR #8205, opened 2026-03-31
BN: Dynamic number of columns count which depends on current sync status.

cheatfate:
With this change i have also made `ValidatorCustodyRef` as main source for number of custody groups and columns.

Algorithm:

1. Node starts with minimal amount of columns  `CUSTODY_REQUIREMENT` (state: `Init`).
2. When node is not in sync it will going to use minimal amount of columns `CUSTODY_REQUIREMENT` (state: `LimitedSet`).
3. When node become optimistically synced it will enter period of time where we monitor node sync status stability, so if node kept synced it will be switched to number of columns based on validators total balance (state: `StabilityPeriod`).
4. When node is in sync it will use number of columns based on validators total balance (**NOTE**: current version does not fix bug that we did not count validators managed by VC).

`syncDistance()` is main sync check function. As soon as result of function is bigger than `MaxSyncDistanceDeviationSlots = 5` node considered as not in sync.

### DIFF
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -17,7 +17,7 @@ import
     types],
   stew/assign2,
   ./crypto,
-  ./[helpers, digest],
+  ./[helpers, digest, column_map],
   ./datatypes/[fulu, deneb]
 
 from std/algorithm import sort
@@ -85,6 +85,18 @@ func resolve_columns_from_custody_groups*(cfg: RuntimeConfig, node_id: NodeId,
   for group in custody_groups:
     for index in compute_columns_for_custody_group(cfg, group):
       columns.incl index
+  columns
+
+func resolve_column_map_from_custody_groups*(
+    cfg: RuntimeConfig,
+    node_id: NodeId,
+    custody_group_count: CustodyIndex
+): ColumnMap =
+  let custody_groups = cfg.get_custody_groups(node_id, custody_group_count)
+  var columns: ColumnMap
+  for group in custody_groups:
+    for index in compute_columns_for_custody_group(cfg, group):
+      columns.incl(index)
   columns
 
 # https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.4/specs/fulu/das-core.md#compute_matrix
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -80,7 +80,7 @@ func resolve_columns_from_custody_groups*(cfg: RuntimeConfig, node_id: NodeId,
                                           custody_group_count: CustodyIndex):
                                           HashSet[ColumnIndex] =
   ## Returns a set of unique columns for the custody groups of a node.
-  let custody_groups = cfg.get_custody_groups(node_id, custody_group_count)
+  let custody_groups = cfg.handle_custody_groups(node_id, custody_group_count)
   var columns: HashSet[ColumnIndex]
   for group in custody_groups:
     for index in compute_columns_for_custody_group(cfg, group):
@@ -92,7 +92,7 @@ func resolve_column_map_from_custody_groups*(
     node_id: NodeId,
     custody_group_count: CustodyIndex
 ): ColumnMap =
-  let custody_groups = cfg.get_custody_groups(node_id, custody_group_count)
+  let custody_groups = cfg.handle_custody_groups(node_id, custody_group_count)
   var columns: ColumnMap
   for group in custody_groups:
     for index in compute_columns_for_custody_group(cfg, group):
--- a/beacon_chain/sync/validator_custody.nim
+++ b/beacon_chain/sync/validator_custody.nim
@@ -12,60 +12,257 @@ import chronicles
 import ssz_serialization/[proofs, types]
 import
   ../validators/action_tracker,
-  ../spec/[beaconstate, forks, network, helpers, peerdas_helpers],
+  ../spec/[beaconstate, forks, network, helpers, peerdas_helpers, column_map],
   ../networking/eth2_network,
-  ../consensus_object_pools/blockchain_dag,
-  ../consensus_object_pools/block_dag,
-  ../consensus_object_pools/blob_quarantine
+  ../conf,
+  ../consensus_object_pools/[blockchain_dag, block_dag, blob_quarantine]
 
-from std/sequtils import toSeq
 from ../beacon_clock import GetBeaconTimeFn
 
 logScope: topics = "validator_custody"
 
+const
+  MaxSyncDistanceDeviationSlots = 5'u64
+    ## Distance from wall slot after which Validator Custody will think that
+    ## node is out of sync.
+  StabilityDistanceSlots = 32'u64
+    ## Distance which Validator Custody will wait until it switches back to
+    ## full custody.
 
 type
+  ValidatorCustodyState* {.pure.} = enum
+    Init,
+      ## Period when node has been started, but validators balance is not yet
+      ## known.
+    FullCustody,
+      ## When node is in sync it uses full validator custody columns set.
+    LimitedCustody,
+      ## When node is out of sync it uses `CUSTODY_REQUIREMENT` validator
+      ## custody columns set
+    StabilityPeriod
+      ## When node recently become synced it should stay synced for some period
+      ## of time before switch from `LimitedCustody` to `FullCustody`.
+
+  ## Init -> FullCustody -> LimitedCustody -> StablePeriod -> FullCustody
+  ## Init -> LimitedCustody -> StablePeriod -> FullCustody
+
   ValidatorCustody* = object
     network: Eth2Node
-    dag*: ChainDAGRef
-    older_column_set: HashSet[ColumnIndex]
-    newer_column_set*: HashSet[ColumnIndex]
-    diff_set*: seq[ColumnIndex]
-    dataColumnQuarantine: ref ColumnQuarantine
+    dag: ChainDAGRef
+    config: BeaconNodeConf
+    curGroupsCount: CgcCount
+    curColumnMap: ColumnMap
+    fuluColumnQuarantine: ref ColumnQuarantine
+    gloasColumnQuarantine: ref GloasColumnQuarantine
+    state: ValidatorCustodyState
+    stabilitySlot: Opt[Slot]
 
   ValidatorCustodyRef* = ref ValidatorCustody
 
-func init*(T: type ValidatorCustodyRef, network: Eth2Node,
-           dag: ChainDAGRef,
-           older_column_set: HashSet[ColumnIndex],
-           dataColumnQuarantine: ref ColumnQuarantine): ValidatorCustodyRef =
-  (ValidatorCustodyRef)(
-    network: network,
-    dag: dag,
-    older_column_set: older_column_set,
-    dataColumnQuarantine: dataColumnQuarantine)
+  ColumnQuarantines* = ref ColumnQuarantine | ref GloasColumnQuarantine
 
-proc detectNewValidatorCustody*(vcus: ValidatorCustodyRef,
-                                current_slot: Slot,
-                                total_node_balance: Gwei) =
-  debug "Total node balance before applying validator custody",
-    total_node_balance = total_node_balance
-  let
-    vcustody =
-      vcus.dag.cfg.get_validators_custody_requirement(total_node_balance)
-    newer_columns =
-      vcus.dag.cfg.resolve_columns_from_custody_groups(
-        vcus.network.nodeId,
-        max(vcus.dag.cfg.CUSTODY_REQUIREMENT.uint64,
-        vcustody))
+func getGroupsCount(
+  state: ValidatorCustodyState,
+  config: BeaconNodeConf,
+  network: Eth2Node,
+  dag: ChainDAGRef,
+  nodeBalance: Opt[Gwei]
+): CgcCount =
+  case state
+  of ValidatorCustodyState.Init:
+    if config.peerdasSupernode:
+      CgcCount(dag.cfg.NUMBER_OF_CUSTODY_GROUPS)
+    elif config.lightSupernode:
+      CgcCount((dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1)
+    else:
+      CgcCount(dag.cfg.CUSTODY_REQUIREMENT)
+  of ValidatorCustodyState.FullCustody:
+    if config.peerdasSupernode:
+      CgcCount(dag.cfg.NUMBER_OF_CUSTODY_GROUPS)
+    elif config.lightSupernode:
+      CgcCount((dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1)
+    else:
+      if nodeBalance.isSome():
+        CgcCount(
+          dag.cfg.get_validators_custody_requirement(nodeBalance.get()))
+      else:
+        CgcCount(dag.cfg.CUSTODY_REQUIREMENT)
+  of ValidatorCustodyState.LimitedCustody:
+    CgcCount(dag.cfg.CUSTODY_REQUIREMENT)
+  of ValidatorCustodyState.StabilityPeriod:
+    CgcCount(dag.cfg.CUSTODY_REQUIREMENT)
 
-  # check which custody set is larger
-  if newer_columns.len >= vcus.older_column_set.len:
-    vcus.diff_set = toSeq(newer_columns.difference(vcus.older_column_set))
-  # else declare the difference to be 0
+func getColumnMap(
+    config: BeaconNodeConf,
+    network: Eth2Node,
+    dag: ChainDagRef,
+    groupsCount: CgcCount
+): ColumnMap =
+  if uint64(groupsCount) == dag.cfg.NUMBER_OF_CUSTODY_GROUPS:
+    var res: ColumnMap
+    for i in 0 ..< dag.cfg.NUMBER_OF_CUSTODY_GROUPS:
+      res.incl(ColumnIndex(i))
+    res
+  elif uint64(groupsCount) == (dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1:
+    var res: ColumnMap
+    for i in 0 ..< (dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1:
+      res.incl(ColumnIndex(i))
+    res
   else:
-    vcus.diff_set = @[]
+    dag.cfg.resolve_column_map_from_custody_groups(
+      network.nodeId, CustodyIndex(groupsCount))
 
-  vcus.older_column_set = vcus.newer_column_set
-  vcus.newer_column_set = newer_columns
-  vcus.dag.eaSlot = max(vcus.dag.eaSlot, current_slot)
+func getGroupsCount(
+    vcus: ValidatorCustodyRef,
+    nodeBalance: Opt[Gwei]
+): CgcCount =
+  getGroupsCount(vcus.state, vcus.config, vcus.network, vcus.dag, nodeBalance)
+
+func getColumnMap(
+    vcus: ValidatorCustodyRef,
+    groupsCount: CgcCount
+): ColumnMap =
+  getColumnMap(vcus.config, vcus.network, vcus.dag, groupsCount)
+
+func syncDistance(
+  vcus: ValidatorCustodyRef,
+  currentSlot: Slot
+): uint64 =
+  let headSlot = vcus.dag.head.slot
+  if currentSlot <= headSlot:
+    0'u64
+  else:
+    currentSlot - headSlot
+
+func updateState(vcus: ValidatorCustodyRef, currentSlot: Slot) =
+  let distance = vcus.syncDistance(currentSlot)
+  case vcus.state
+  of ValidatorCustodyState.Init:
+    if distance <= MaxSyncDistanceDeviationSlots:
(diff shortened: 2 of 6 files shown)


beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+      vcus.stabilitySlot = Opt.some(currentSlot)
+  of ValidatorCustodyState.StabilityPeriod:
+    if distance > MaxSyncDistanceDeviationSlots:
+      vcus.state = ValidatorCustodyState.LimitedCustody
+      vcus.stabilitySlot = Opt.none(Slot)
+    else:
+      doAssert(vcus.stabilitySlot.isSome(),
+        "Stability start slot should be set at this moment")
+      doAssert(vcus.stabilitySlot.get() <= currentSlot,
+        "Invalid time? Current slot is less than stability start slot")
```
tersec: `vcus.stabilitySlot.isSome()` might be assertable, but anything dependent on wall clock time when clocks do, actually, run backwards sometimes is a bit risky as an intentionally triggered crash.

As with the other cases in the codebase of handling pathologically wrong or inconsistent clocks, sometimes it's true that Nimbus has no better option than to shut down and simply stop trying to follow a chain or validate, but the way to express that isn't `doAssert`, but something more controlled.

I'm also not sure it makes sense here to begin with.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
 
-from std/sequtils import toSeq
 from ../beacon_clock import GetBeaconTimeFn
 
 logScope: topics = "validator_custody"
 
+const
+  MaxSyncDistanceDeviationSlots = 5'u64
+    ## Distance from wall slot after which Validator Custody will think that
+    ## node is out of sync.
```
tersec, line 26: How important is it that Nimbus has a consistent view of what counts as "synced"? Different parts of Nimbus already have each their own ad-hoc definitions of what counts as synced or not, including but not limited to:
- what triggers the sync manager rather than request manager (roughly, being one epoch out of date)
- cosmetic display (even a single missed slot on the network, which happens regularly on mainnet and frequently on Hoodi, results in "almost synced". Which is wrong.); and
- a measure of when to turn on gossip traffic, which during longer-range syncing is turned off for efficiency.

So in particular, what will happen here is that validator custody will think the node is out of sync when the only other indication of that (and not a reliable one) is the "almost synced" indication which has regular false positives.

That said, sure, the gap between `MaxSyncDistanceDeviationSlots` and `StabilityDistanceSlots` is necessary to avoid hysteresis.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+  StabilityDistanceSlots = 32'u64
+    ## Distance which Validator Custody will wait until it switches back to
+    ## full custody.
 
 type
+  ValidatorCustodyState* {.pure.} = enum
+    Init,
+      ## Period when node has been started, but validators balance is not yet
+      ## known.
```
tersec, line 35: Isn't there always some defined validator balance? It might not be the final one, but that's dynamic in principle anyway, both due to the REST beacon API  and the keymanager API. I'm not sure what's special about this period.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+  network: Eth2Node,
+  dag: ChainDAGRef,
+  nodeBalance: Opt[Gwei]
+): CgcCount =
+  case state
+  of ValidatorCustodyState.Init:
+    if config.peerdasSupernode:
+      CgcCount(dag.cfg.NUMBER_OF_CUSTODY_GROUPS)
+    elif config.lightSupernode:
+      CgcCount((dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1)
```
tersec: Especially given the push to change this to some other exact formula, it's not great to keep adding more ad-hoc (x/2)+1 calculations which will have to be 100% found and fixed later, or else one of the lovely assertions in code like this fires.

Probably a good first step is to centralize these somewhere so there's a single access method that canonically provides this calculation based on the `RuntimeConfig`.

There aren't a huge (almost no) parts of Nimbus I'm that hesitant to merge changes on, because for the most part things are reasonably factored and I'm confident that subsystems are architecturally reviewable. This one is, realistically, not, and that's kind of bad. Too easy to miss one of the (x/2)+1 places. First step is to stop adding to it, next step is to fix (_without_ changing the number itself, just refatoring the calculations) then looking at modifying it.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+
+proc init*(
+    T: type ValidatorCustodyRef,
+    config: BeaconNodeConf,
+    network: Eth2Node,
+    dag: ChainDAGRef
+): ValidatorCustodyRef =
+  let
+    localGroupsCount = getGroupsCount(
+      ValidatorCustodyState.Init, config, network, dag, Opt.none(Gwei))
```
tersec: Why is this `Opt.none` warranted? Even if the answer is that the node hasn't loaded either its in-BN validators yet nor have the VC(s) had time to register their validators with the BN through the beacon API, there doesn't seem to be anything conceptually special about this time period, i.e. I don't see how `Opt[Gwei]` models this properly. It might be that it starts at, e.g., 0, it might be that it starts at in-BN validators only, it might be that one can massage this by putting this initialization specifically after BN validators have been initialized.

But the code has to handle dynamic balance anyway, so this one-off initialization state doesn't really buy anything?

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
-proc detectNewValidatorCustody*(vcus: ValidatorCustodyRef,
-                                current_slot: Slot,
-                                total_node_balance: Gwei) =
-  debug "Total node balance before applying validator custody",
-    total_node_balance = total_node_balance
+    curColumnMap: columnMap,
+    state: ValidatorCustodyState.Init
+  )
+
+proc setQuarantine*[T: ColumnQuarantines](
```
tersec: Can be `func`

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+func getMap*(vcus: ValidatorCustodyRef): ColumnMap =
+  vcus.curColumnMap
+
+func getSet*(vcus: ValidatorCustodyRef): HashSet[ColumnIndex] =
+  var res: HashSet[ColumnIndex]
+  for index in vcus.curColumnMap:
+    res.incl(index)
+  res
+
+func getCustodyGroupSubnets*(vcus: ValidatorCustodyRef): uint64 =
```
tersec: Doesn't seem to be used?

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+func isSupernode*(vcus: ValidatorCustodyRef): bool =
+  uint64(vcus.curGroupsCount) == vcus.dag.cfg.NUMBER_OF_CUSTODY_GROUPS
+
+func isLightSupernode*(vcus: ValidatorCustodyRef): bool =
+  uint64(vcus.curGroupsCount) ==
+    (vcus.dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1
+
+func getCustodyGroups*(vcus: ValidatorCustodyRef): seq[CustodyIndex] =
+  if vcus.isLightSupernode():
+    let custodyGroups = (vcus.dag.cfg.NUMBER_OF_CUSTODY_GROUPS div 2) + 1
```
tersec: Yeah even this PR has at least 4 disparate instances of this (x/2)+1 pattern randomly embedded in places. Last time it was attempted to be changed things promptly imploded. It's not sufficiently readily maintainable.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+  )
+
+proc setQuarantine*[T: ColumnQuarantines](
+    vcus: ValidatorCustodyRef,
+    q: T
+) =
+  when T is ColumnQuarantine:
+    vcus.fuluColumnQuarantine = q
+  elif T is GloasColumnQuarantine:
+    vcus.gloasColumnQuarantine = q
```
tersec: It's arguably cleaner to use Nim's type system to address the different `T` types, e.g., one can also have a `setQuarantine` which takes a `q: ColumnQuarantine` overload and a `q: GloasColumnQuarantine` overload, as one possible approach.

This avoids the gap here of, the maintainability point of what if someone calls `setQuarantine` with a new type in `ColumnQuarantines` type that this isn't updated for yet (e.g., maybe a partial column cell quarantine, either for Fulu or Gloas).

Right now this silently does nothing and compiles without errors, which is a tricky failure mode to notice. Typically code which does manual compile-time type dispatching with `when` in `nimbus-eth2` has a catch-all `else:` clause which uses some `{.error}`/`static: doAssert`/`static: raiseAssert` etc to ensure that whatever it is is properly updated.

Sometimes that's the best one can do, but here one can just use Nim's type-based overload resolution mechanisms directly.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+  res
+
+func getCustodyGroupSubnets*(vcus: ValidatorCustodyRef): uint64 =
+  uint64(vcus.curGroupsCount)
+
+func isSupernode*(vcus: ValidatorCustodyRef): bool =
+  uint64(vcus.curGroupsCount) == vcus.dag.cfg.NUMBER_OF_CUSTODY_GROUPS
+
+func isLightSupernode*(vcus: ValidatorCustodyRef): bool =
+  uint64(vcus.curGroupsCount) ==
```
tersec: All of these `uint64(someDistinctUint64Var)` and `uint8(someDistinctUint8Var)` would be statically more typesafe with `distinctBase`.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,257 @@ import chronicles
...
+  let distance = vcus.syncDistance(currentSlot)
+  case vcus.state
+  of ValidatorCustodyState.Init:
+    if distance <= MaxSyncDistanceDeviationSlots:
+      vcus.state = ValidatorCustodyState.FullCustody
+    else:
+      vcus.state = ValidatorCustodyState.LimitedCustody
+  of ValidatorCustodyState.FullCustody:
+    if distance > MaxSyncDistanceDeviationSlots:
+      vcus.state = ValidatorCustodyState.LimitedCustody
```
agnxsh, line 159: what happens if quarantine gets updated to `LimitedCustody` and soon after that existing entry of (block+blob) set in the block processor looks up the column quarantine and finds much lesser columns than it was expecting? Does that entire entry get discarded? I think this is a decent risk here?

i understand that `StabilityPeriod` is something made to tackle the reverse of such a situation, that is `LimitedCustody -> StablePeriod -> FullCustody
`

cheatfate: If quarantine gets updated to `LimitedCustody` it will have `CUSTODY_REQUIREMENT` columns, it means that if there is pending `block` which is going to be processed it will query quarantine and obtains all the columns it needs, because at this moment only `CUSTODY_REQUIREMENT` columns will be needed, so if `CUSTODY_REQUIREMENT` columns was present  before transition to `LimitedCustody` there will be no problems after transition... Block will be processed with `CUSTODY_REQUIREMENT` number of columns.

beacon_chain/sync/validator_custody.nim
cheatfate: I see only critique in this post, but i do not see any  proposal. Currently we do not have any long range syncing working, and this PR improves situation in case when we lost sync. 5 slots difference from head is already a problem no? It is definitely not a "missed" slot and if we switch to `LimitedCustody` it can just improve our back to sync capabilities.

beacon_chain/sync/validator_custody.nim
cheatfate: It is how it works now, it first initialized with `CUSTODY_REQUIREMENT`, after we loading validators, after we perform slot actions and at the end of first slot we update our CUSTODY with "balances".

beacon_chain/sync/validator_custody.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8205/commits/3202483ba804035dacc9ff0cd45004496f93e9c9

beacon_chain/sync/validator_custody.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8205/commits/3202483ba804035dacc9ff0cd45004496f93e9c9

beacon_chain/sync/validator_custody.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8205/commits/3202483ba804035dacc9ff0cd45004496f93e9c9

beacon_chain/sync/validator_custody.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8205/commits/3202483ba804035dacc9ff0cd45004496f93e9c9

beacon_chain/sync/validator_custody.nim
cheatfate: There is a short gap in time between when `ValidatorCustody` is being initialized and first slot being finished (onSlotEnd called), in this gap we use `CUSTODY_REQUIREMENT` number of columns which is now (4).
Also in this gap we loading validators, so when first `onSlotEnd` we will have `total_balance` which will adjust our current state to `FullCustody` and so update number of columns node will serve.

beacon_chain/sync/validator_custody.nim
cheatfate: I could call it different without using term `sync`. Because what means if node is behind wall clock by 5 slots:
* The worst case would be that we just missed 5 slots and nothing more, in this case we still consider switch to `LimitedCustody` but i assume that this case is the worst possible scenario and probability of it is pretty low, but we could increase number of slots to 10 to make probability even lower.
* Next case is that we are in period of non-finality and in this case switching to `LimitedCustody` will just help us to follow multiple forks, because it much better to follow forks with 4 columns and not with 128.
* And last case is that our node is out of sync just because our CPU is overloaded  and we are unable to follow network, this is just what we need - 4 columns to get synced as fast as possible.

beacon_chain/sync/validator_custody.nim
cheatfate: The problem is that `ValidatorCustody` will become single source of columns data for multiple entities:
1. Blob quarantine.
2. Eth2 network metadata.
3. Request manager
So it should have something for this data structures to become initialized.

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8205/commits/ffe5f13df0425cc53a5b38e392564b87aafc62ee

beacon_chain/sync/validator_custody.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8205/commits/ffe5f13df0425cc53a5b38e392564b87aafc62ee

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -87,6 +87,18 @@ func resolve_columns_from_custody_groups*(cfg: RuntimeConfig, node_id: NodeId,
       columns.incl index
   columns
 
+func resolve_column_map_from_custody_groups*(
+    cfg: RuntimeConfig,
+    node_id: NodeId,
+    custody_group_count: CustodyIndex
+): ColumnMap =
+  let custody_groups = cfg.get_custody_groups(node_id, custody_group_count)
```
tersec: https://github.com/status-im/nimbus-eth2/blob/5e676c4a1359935a7bc2926abcd439d0196ae61f/beacon_chain/spec/peerdas_helpers.nim#L68-L77 is a wrapper over https://github.com/status-im/nimbus-eth2/blob/5e676c4a1359935a7bc2926abcd439d0196ae61f/beacon_chain/spec/peerdas_helpers.nim#L39-L66 which adds no value for this purpose and (pointlessly) sorts it. Slightly more efficient to use `handle_custody_groups` directly.

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,272 @@ import chronicles
...
+    info "New validator custody count detected"
+    vcus.setValidatorCustody(currentSlot, newGroupsCount, newMap)
+
+func getMap*(vcus: ValidatorCustodyRef): ColumnMap =
+  vcus.curColumnMap
 
-  vcus.older_column_set = vcus.newer_column_set
-  vcus.newer_column_set = newer_columns
-  vcus.dag.eaSlot = max(vcus.dag.eaSlot, current_slot)
+func getSet*(vcus: ValidatorCustodyRef): HashSet[ColumnIndex] =
```
tersec: This is only used in one place, which is a `for` loop; not sure why it reconstructs a whole `HashSet` when apparently the underlying `vcus.curColumnMap` already iterates over precisely the things that request manager will want to iterate over.
```
$ rg getSet beacon_chain -C4
beacon_chain/sync/validator_custody.nim
252-
253-func getMap*(vcus: ValidatorCustodyRef): ColumnMap =
254-  vcus.curColumnMap
255-
256:func getSet*(vcus: ValidatorCustodyRef): HashSet[ColumnIndex] =
257-  var res: HashSet[ColumnIndex]
258-  for index in vcus.curColumnMap:
259-    res.incl(index)
260-  res

beacon_chain/sync/request_manager.nim
424-            remoteNodeId,
425-            max(rman.network.cfg.SAMPLES_PER_SLOT,
426-                remoteCustodyGroupCount))
427-
428:      for local_column in rman.vcus.getSet():
429-        if local_column in remoteCustodyColumns:
430-          discard intersection.add(local_column)
431-      # Apply scoring logic + logs
432-      if intersection.len == 0:
```

beacon_chain/sync/validator_custody.nim
```diff
@@ -12,60 +12,272 @@ import chronicles
...
+  vcus.config.peerdasSupernode and
+    vcus.curGroupsCount == vcus.supernodeGroupsCount()
+
+func isLightSupernode*(vcus: ValidatorCustodyRef): bool =
+  ## This function returns current value, based on current state, so if we are
+  ## not in `LimitedCustody` state it will returns ``true``.
+  vcus.config.lightSupernode and
+    vcus.curGroupsCount == vcus.lightSupernodeGroupsCount()
+
+func getCustodyGroups*(vcus: ValidatorCustodyRef): seq[CustodyIndex] =
```
tersec: Similar question here:
```
$ rg getCustodyGroups -C3
beacon_chain/nimbus_beacon_node.nim
1330-proc updateDataColumnSidecarHandlers(node: BeaconNode, gossipEpoch: Epoch) =
1331-  let
1332-    forkDigest = node.dag.forkDigests[].atEpoch(gossipEpoch, node.dag.cfg)
1333:    custody = node.validatorCustody.getCustodyGroups()
1334-
1335-  for i in custody:
1336-    let topic = getDataColumnSidecarTopic(forkDigest, i)
--
1432-  node.removeCapellaMessageHandlers(forkDigest)
1433-
1434-  # Due to the dynamic column change, we need to unsubscribe from all the
1435:  # subnets we subscribed to. Because getCustodyGroups() may return a different
1436-  # set than the one we subscribed to previously.
1437-  for i in node.lastColumnCustodyIndices:
1438-    let topic = getDataColumnSidecarTopic(forkDigest, i)

beacon_chain/sync/validator_custody.nim
271-  vcus.config.lightSupernode and
272-    vcus.curGroupsCount == vcus.lightSupernodeGroupsCount()
273-
274:func getCustodyGroups*(vcus: ValidatorCustodyRef): seq[CustodyIndex] =
275-  ## Returns current dynamic state of custody groups.
276-  if vcus.isLightSupernode():
277-    let custodyGroups = vcus.lightSupernodeGroupsCount()
```

This appears to be used only in one place, which iterates over it. So there's no need to actually construct a `seq[CustodyIndex]`, only provide that method of iteration.

tersec:
```
[2026-04-08T16:24:15.378Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-8205/beacon_chain/nimbus_beacon_node.nim(1333, 36) Error: type mismatch
[2026-04-08T16:24:15.378Z] Expression: getCustodyGroups(node.validatorCustody)
[2026-04-08T16:24:15.378Z]   [1] node.validatorCustody: ValidatorCustodyRef
[2026-04-08T16:24:15.378Z] 
[2026-04-08T16:24:15.378Z] Expected one of (first mismatch at [position]):
[2026-04-08T16:24:15.378Z] [1] func get_custody_groups(cfg: RuntimeConfig; node_id: NodeId;
[2026-04-08T16:24:15.378Z]                         custody_group_count: CustodyIndex): seq[CustodyIndex]
[2026-04-08T16:24:15.378Z] 
[2026-04-08T16:24:15.378Z] make: *** [Makefile:449: nimbus_beacon_node] Error 1
script returned exit code 2[2026-04-08T16:24:17.573Z] [WS-CLEANUP] Deleting project workspace...
[2026-04-08T16:24:17.573Z] [WS-CLEANUP] Deferred wipeout is disabled by the job configuration...
[2026-04-08T16:24:39.244Z] [WS-CLEANUP] done
```

PR #8213, opened 2026-04-01
added lookup to last persisted column indices at boot

agnxsh:


agnxsh:
it's a bit conflicting with #8205 , will have to refactor after either merges

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -388,6 +388,20 @@ proc getForkedBlock*(
...
+  ## Returns an empty set if no columns are found.
+  let blckOpt = dag.getForkedBlock(dag.head.bid)
+  if blckOpt.isSome():
+    withBlck(blckOpt.get()):
+      when consensusFork >= ConsensusFork.Fulu:
+        for i in 0..<NUMBER_OF_COLUMNS.uint64:
+          var colData: fulu.DataColumnSidecar
+          if dag.db.getDataColumnSidecar(
+              forkyBlck.root, ColumnIndex(i), colData):
+            result.incl ColumnIndex(i)
```
tersec, line 403: https://status-im.github.io/nim-style-guide/language.result.html

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1363,6 +1368,22 @@ func readCustodyGroupSubnets(node: BeaconNode): uint64 =
...
+    custody_groups = node.dag.cfg.NUMBER_OF_CUSTODY_GROUPS
+    targetSubnets = node.readCustodyGroupSubnets()
+    persisted = node.dag.getPersistedCustodyColumns()
+  if persisted.len > 0:
+    for col in persisted:
+      let group = CustodyIndex(col mod custody_groups)
+      if group notin result:
+        result.add group
+  else:
+    result = node.dag.cfg.get_custody_groups(
```
tersec, line 1384: https://status-im.github.io/nim-style-guide/language.result.html

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -388,6 +388,20 @@ proc getForkedBlock*(
           result.err()
           return
 
+proc getPersistedCustodyColumns*(
+    dag: ChainDAGRef): HashSet[ColumnIndex] =
+  ## Returns the set of column indices persisted in the DB for the head block.
+  ## Returns an empty set if no columns are found.
+  let blckOpt = dag.getForkedBlock(dag.head.bid)
+  if blckOpt.isSome():
+    withBlck(blckOpt.get()):
```
tersec, line 397: It's suboptimal to use manual `.get()`s. Yes, this one should be safe, but the general issue is that checking this requires manually checking each case in a contextual way. Ideally, we'd evolve the Nimbus codebase towards treating either `get`/`unsafeGet` as largely deprecated, and be able to therefore more automatically spot potential `Defect` triggers.

This is particularly easy case to use `nim-results``'s various useful abstractions on without losing efficiency or code ergonomics.

tersec:
`result` return nitpicks aside, this conceptually seems a bit odd: there's supposed to be a deterministic mapping from (node ID, number of columns) tuple to (custody set) which other nodes can know.

In particular, https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/fulu/p2p-interface.md#peer-scoring
> Due to the deterministic custody functions, a node knows exactly what a peer should be able to respond to. In the event that a peer does not respond to samples of their custodied rows/columns, a node may downscore or disconnect from a peer.

The details of this mapping are specified by [get_custody_groups()](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/fulu/das-core.md#get_custody_groups); in general, it means that this PR seems to hit an odd, sort of not optimally useful middle ground because it only ensures consistency with its own previous instances, not with its current node id.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -388,6 +388,20 @@ proc getForkedBlock*(
...
+      when consensusFork >= ConsensusFork.Fulu:
+        for i in 0..<NUMBER_OF_COLUMNS.uint64:
+          var colData: fulu.DataColumnSidecar
+          if dag.db.getDataColumnSidecar(
```
cheatfate, line 401: I think query database minimum 4 times for  CUSTODY_REQUIREMENT and 128 times for supernode is not efficient way. Probably database needs one more API call which will return you all columns which belongs to specific block root.

agnxsh: fair enough, but read tersec's above comment of it might not be the best way, since, other peers use our _current_ node id to decipher the columns we custody, if they see a mismatch in those, there might downscoring, i think the best solution would be to do a combination of this + fixing the nodeid or updating eas once there new node id.

cheatfate (changes requested):

agnxsh:
this can be closed in favor of #8205 since 2 of these prs were 2 different approaches to deal with the same problem, #8205 happens to be a more plausible one.

PR #8227, opened 2026-04-07
Replace historical summary GINDEX constants with get_generalized_index

etan-status:
Instead of doing the computation of gindex manually (which is sometimes preset dependent), use the new `get_generalized_index` API from nim-ssz and statically assert that the computed value matches the spec values.

### DIFF
--- a/beacon_chain/spec/eth2_apis/rest_types.nim
+++ b/beacon_chain/spec/eth2_apis/rest_types.nim
@@ -57,10 +57,10 @@ type
   # https://github.com/ethereum/beacon-APIs/blob/v2.4.2/apis/eventstream/index.yaml
   EventTopic* {.pure.} = enum
     Head, Block, BlockGossip, VoluntaryExit, BLSToExecutionChange,
-    ProposerSlashing, AttesterSlashing, BlobSidecar, DataColumnSidecar, SingleAttestation,
-    FinalizedCheckpoint, ChainReorg, ContributionAndProof,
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/eth2_apis/rest_types.nim
```diff
@@ -1200,16 +1209,31 @@ template init*(
...
+  for consensusFork in ConsensusFork:
+    withConsensusFork(consensusFork):
+      const historicalSummariesFork =
+        historicalSummariesForkAtConsensusFork(consensusFork)
+      when historicalSummariesFork.isSome:
+        template check(gindex, T: untyped, path: varargs[untyped]): untyped =
+          doAssert gindex == consensusFork.T.get_generalized_index(path)
+
+        check historicalSummariesFork.get.historical_summaries_gindex,
+          BeaconState, "historical_summaries"
```
kdeme, line 1233: I didn't expect this to just work without specifying which `BeaconState`, but apparently it does...

kdeme (approved):

beacon_chain/spec/eth2_apis/rest_types.nim
etan-status: the "check" function prepends `consensusFork.` before `T` ;-)

etan-status: it's a template within `forks.nim`

kdeme: ow right, of course. Thanks

PR #8228, opened 2026-04-07
Replace verifying signer GINDEX constants with get_generalized_index

etan-status:
Instead of doing the computation of gindex manually (which is sometimes preset dependent), use the new `get_generalized_index` API from nim-ssz and statically assert that the computed value matches the spec values.

Gloas flow for fee recipient was broken before and still is.

### DIFF
--- a/beacon_chain/spec/keystore.nim
+++ b/beacon_chain/spec/keystore.nim
@@ -21,7 +21,12 @@ import
   nimcrypto/[sha2, rijndael, pbkdf2, bcmode, hash, scrypt],
   # Local modules
   libp2p/crypto/crypto as lcrypto,
-  ./datatypes/base, ./signatures
+  ./datatypes/base, ./[ssz_codec, signatures]
+
+from ssz_serialization/proofs import get_generalized_index
+from ./datatypes/electra import BeaconBlockBody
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/keystore.nim
```diff
@@ -719,26 +724,49 @@ template writeValue*(w: var JsonWriter,
...
-      fuluIndex: GeneralizedIndex(18),
-      gloasIndex: GeneralizedIndex(18))
+      electraIndex: electra.BeaconBlockBody.gindexOrZero(pathArgs),
+      fuluIndex: fulu.BeaconBlockBody.gindexOrZero(pathArgs),
+      gloasIndex: gloas.BeaconBlockBody.gindexOrZero(pathArgs))
+
+  case propertyPath
+  of ".execution_payload.fee_recipient":
+    let res = initProvenProperty("execution_payload", "fee_recipient")
+    ok res
```
etan-status, line 746: For some reason, this doesn't work inline :) whatever

tersec: Workaround isn't needed by sometime in Nim 2.2.9, i.e. the branch will become 2.2.10. No new bugs to report upstream.

PR #8231, opened 2026-04-07
proposer preferences implementation

Tomi-3-0:
Implement proposer preferences according to:
https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/validator.md#broadcasting-signedproposerpreferences
https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/p2p-interface.md#proposer_preferences
https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/p2p-interface.md#new-proposerpreferences

### DIFF
--- a/beacon_chain/spec/validator.nim
+++ b/beacon_chain/spec/validator.nim
@@ -614,38 +614,32 @@ func get_beacon_proposer_indices*(
 
 # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/p2p-interface.md#proposer_preferences
 func is_valid_proposal_slot(
-    state: ForkyBeaconState, slot: Slot, validator_index: uint64): bool =
+    state: gloas.BeaconState | heze.BeaconState,
+    slot: Slot, validator_index: uint64): bool =
   ## Check if the validator is the proposer for the given slot in the current or
   ## next epoch.
-  when typeof(state).kind < ConsensusFork.Fulu:
-    false
-  else:
-    let
-      current_epoch = state.get_current_epoch()
-      start_slot = current_epoch.start_slot()
-    if slot < start_slot or
-        slot - start_slot >= state.proposer_lookahead.lenu64:
-      return false
-    state.proposer_lookahead.item(slot - start_slot) == validator_index
+  let start_slot = state.get_current_epoch().start_slot()
+  if slot < start_slot or
+      slot - start_slot >= state.proposer_lookahead.lenu64:
+    return false
+  state.proposer_lookahead.item(slot - start_slot) == validator_index
 
 # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/validator.md#broadcasting-signedproposerpreferences
 iterator get_upcoming_proposal_slots(
-    state: ForkyBeaconState, validator_index: uint64): Slot =
+    state: gloas.BeaconState | heze.BeaconState,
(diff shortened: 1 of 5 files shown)


beacon_chain/spec/validator.nim
```diff
@@ -612,6 +612,41 @@ func get_beacon_proposer_indices*(
     # function does not require shuffled indices post Fulu
     get_beacon_proposer_indices(state, epoch)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/p2p-interface.md#proposer_preferences
+func is_valid_proposal_slot(
```
tersec, line 616: Will there be a reason to call this pre-Gloas? If not, can just check for Gloas and Heze `BeaconState`s as the typeclass, rather than try to semi-arbitrarily handle pre-Gloas calls.

tersec: i.e. instead of `ForkyBeaconState`

beacon_chain/spec/validator.nim
```diff
@@ -612,6 +612,41 @@ func get_beacon_proposer_indices*(
...
+      current_epoch = state.get_current_epoch()
+      start_slot = current_epoch.start_slot()
+    if slot < start_slot or
+        slot - start_slot >= state.proposer_lookahead.lenu64:
+      return false
+    state.proposer_lookahead.item(slot - start_slot) == validator_index
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/validator.md#broadcasting-signedproposerpreferences
+iterator get_upcoming_proposal_slots(
+    state: ForkyBeaconState, validator_index: uint64): Slot =
```
tersec: Similar question about `get_upcoming_proposal_slots`: it seems like one could restrict the `state` parameter to just `gloas.BeaconState | heze.BeaconState`.

beacon_chain/spec/validator.nim
```diff
@@ -612,6 +612,41 @@ func get_beacon_proposer_indices*(
...
+  ## Yield the future slots in the current epoch and the slots in the next
+  ## epoch for which ``validator_index`` is proposing.
+  when typeof(state).kind >= ConsensusFork.Gloas:
+    let
+      current_epoch = state.get_current_epoch()
+      current_slot = state.slot
```
tersec: Both of these variables are used just once, and, at some level, integers are free, but `state.get_current_epoch()` and `state.slot` are also pretty descriptive already. Not sure it's worth having these separately declared/defined rather than used directly in their each-one-use-cases.

tersec: https://github.com/status-im/nimbus-eth2/pull/8231/changes/0d460984e3cc876a92ae0fe607b2f8426f7302a7

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8231/changes/0d460984e3cc876a92ae0fe607b2f8426f7302a7

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8231/changes/0d460984e3cc876a92ae0fe607b2f8426f7302a7

PR #8251, opened 2026-04-12
fix 32-bit builds, rm pointless @(mapIt(...)) wrapping, rm unused code

tersec:
- `const maxBodyBytes = 3 * 1024 * 1024 * 1024` overflows on 32-bit build and targets https://github.com/status-im/nimbus-eth2/blob/83aaf6dca139f94b37d0f5d489b3f8fcca1676cc/beacon_chain/spec/eth2_apis/rest_common.nim#L37-L43 which then uses `int` per se (e.g., https://github.com/status-im/nim-chronos/blob/3bc3d80256f08a6e4d87eb3b3e4afc8bd5f0e8b9/chronos/streams/asyncstream.nim#L531 and https://github.com/status-im/nim-chronos/blob/3bc3d80256f08a6e4d87eb3b3e4afc8bd5f0e8b9/chronos/transports/stream.nim#L2776) and
- [`@(...)`](https://nim-lang.org/docs/system.html#%40%2CopenArray%5BT%5D) "is not as efficient as turning a fixed length array into a sequence as it always copies every element of `a`", while [`mapIt`](https://nim-lang.org/docs/sequtils.html#mapIt.t%2Ctyped%2Cuntyped) already returns a `seq`. `@(mapIt(...))` is redundant and wasteful


tersec:
- https://github.com/status-im/nimbus-eth2/pull/6682 shows the optimization around `isZero`

PR #8263, opened 2026-04-13
Add discovery custody filter ability.

cheatfate:


beacon_chain/networking/eth2_discovery.nim
```diff
@@ -124,19 +130,21 @@ func isCompatibleForkId*(discoveryForkId: ENRForkID, peerForkId: ENRForkID): boo
 
 proc queryRandom*(
```
tersec, line 131:
```
[2026-04-13T22:48:43.855Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-8263/tests/test_discovery.nim(71, 33) Error: type mismatch
[2026-04-13T22:48:43.855Z] Expression: queryRandom(node1, enrForkId, attnetsSelected, noSyncnetsPreference,
[2026-04-13T22:48:43.855Z]             noCgcnetsPreference, 1)
[2026-04-13T22:48:43.855Z]   [1] node1: Eth2DiscoveryProtocol
[2026-04-13T22:48:43.855Z]   [2] enrForkId: ENRForkID
[2026-04-13T22:48:43.855Z]   [3] attnetsSelected: AttnetBits
[2026-04-13T22:48:43.855Z]   [4] noSyncnetsPreference: SyncnetBits
[2026-04-13T22:48:43.855Z]   [5] noCgcnetsPreference: CgcBits
[2026-04-13T22:48:43.855Z]   [6] 1: int literal(1)
[2026-04-13T22:48:43.855Z] 
[2026-04-13T22:48:43.855Z] Expected one of (first mismatch at [position]):
[2026-04-13T22:48:43.855Z] [2] proc queryRandom(d: Eth2DiscoveryProtocol; cfg: RuntimeConfig;
[2026-04-13T22:48:43.855Z]                  forkId: ENRForkID; wantedAttnets: AttnetBits;
[2026-04-13T22:48:43.855Z]                  wantedSyncnets: SyncnetBits; localMap: ColumnMap;
[2026-04-13T22:48:43.855Z]                  minScore: QueryScore): InternalRaisesFuture[seq[Node],
[2026-04-13T22:48:43.855Z]     (CancelledError,)]
[2026-04-13T22:48:43.855Z] [2] proc queryRandom(d: Protocol): InternalRaisesFuture[seq[Node], (CancelledError,)]
[2026-04-13T22:48:43.855Z]   extra argument given
[2026-04-13T22:48:43.855Z] [2] proc queryRandom(d: Protocol; enrField: (string, seq[byte])): InternalRaisesFuture[
[2026-04-13T22:48:43.855Z]     seq[Node], (CancelledError,)]
[2026-04-13T22:48:43.855Z] 
[2026-04-13T22:48:43.855Z] make: *** [Makefile:363: all_tests] Error 1
```

tersec: https://github.com/status-im/nimbus-eth2/pull/8263/commits/2c1f03a05867346acc9a519cfd011ef95f8839f0

beacon_chain/networking/eth2_network.nim
```diff
@@ -1623,29 +1619,27 @@ proc runDiscoveryLoop(node: Eth2Node) {.async: (raises: [CancelledError]).} =
...
-        minScore =
-          if wantedAttnetsCount > 0 or wantedSyncnetsCount > 0 or
-              wantedCgcnetsCount > 0:
-            1
-          else:
-            0
+        minScore = [
+          (if wantedAttnetsCount > 0: 1 else: 0),
+          (if wantedSyncnetsCount > 0: 10 else: 0),
+          100
```
tersec, line 1635: This would disallow peers with no overlapping columns at all, even if, e.g., they do share a sync committee subnet?

beacon_chain/networking/eth2_discovery.nim
```diff
@@ -162,8 +170,11 @@ proc queryRandom*(
             peer = n.record.toURI(), exception = e.name, msg = e.msg
           continue
 
-      if wantedCgcnets.countOnes().uint8 == cgcCountNode:
-        score += 1
+      let peerMap =
+        cfg.resolve_column_map_from_custody_groups(
+          n.id, CustodyIndex(cgcCountNode))
+
+      score[ColumnNetScore] = 100 * len(peerMap and localMap)
```
tersec, line 177: One aspect this doesn't address is overall coverage of columns. In theory, the way custody set and therefore column custody is constructed, it should be somewhat uniformly randomly distributed, and there's some evidence this is the case. But this algorithm depends on this uniform randomness to avoid this metric pinging on the same subset of locally custodied columns repeatedly, without making progress towards overall completion of overlap of custody between local and remote peers.

In particular, I'd be concerned about, with increasing number of columns required, that without such bias towards not only overlap with local custody but local custody not covered by existing peers, it falls into a https://en.wikipedia.org/wiki/Coupon_collector%27s_problem situation where collecting all of them becomes increasingly unlikely (https://en.wikipedia.org/wiki/File:Coupon_collector_problem.svg) unless encountering supernodes.

The exact math in that Wikipedia article would need to be adapted for this situation, but the general idea of, sampling with replacement covering a set fairly slowly, still applies (there are lots of variations on this distribution).

beacon_chain/networking/eth2_network.nim
cheatfate: First of all this filter is only applied outgoing connection peers, and yes minimal score 100 disallow all the peers which do not share at least one column with us.

beacon_chain/networking/eth2_discovery.nim
cheatfate: Current version is still improvement over the original version,  but i can add comments about this problem in code.

tersec: I agree, it's an improvement on the current version, and it's worth merging such an improvement regardless of that there are further possible improvements one might make.

PR #8268, opened 2026-04-14
don't require safe/finalized block hashes to match when checking cached payload requests

tersec:
This was causing consistent misses, and additional 500ms delays when proposing.

This also disables fulu in minimal local testnets for now because it needs further fixes to be fast enough.


etan-status (approved):
nice catch!

PR #8275, opened 2026-04-15
add tests and function to batch kzg verify

agnxsh:


beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -576,6 +576,83 @@ proc verify_data_column_sidecar_kzg_proofs*(sidecar: gloas.DataColumnSidecar,
...
+  let res = verifyCellKzgProofBatch(
+      commitments, cellIndices, cells, proofs).valueOr:
+    return err("DataColumnSidecar: validation error")
+
+  if not res:
+    return err("DataColumnSidecar: validation failed")
+
+  ok()
+
+proc verify_data_column_sidecar_kzg_proofs*(
```
tersec, line 604: The fulu and gloas functions are almost identical, except that one takes a separate `kzg_commitments: KzgCommitments` parameter. The logic contained within them though aside from that is the same, so it would be better to refactor this so the Fulu version is a simple wrapper function which extracts `sidecar.kzg_commitments`, passes it along as an argument, and then calls a function which takes `openArray[fulu.DataColumnSidecar] | openArray[gloas.DataColumnSidecar]`, i.e. a generic function which handles both the same way.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -534,17 +534,20 @@ func verify_partial_data_column_header_inclusion_proof*(
   ok()
 
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.3/specs/fulu/p2p-interface.md#verify_data_column_sidecar_kzg_proofs
-proc verify_data_column_sidecar_kzg_proofs*(sidecar: fulu.DataColumnSidecar):
-                                            Result[void, cstring] =
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/p2p-interface.md#modified-verify_data_column_sidecar_kzg_proofs
+proc verify_data_column_sidecar_kzg_proofs*[
+    T: fulu.DataColumnSidecar | gloas.DataColumnSidecar](
+    sidecar: T,
+    kzg_commitments: KzgCommitments): Result[void, cstring] =
```
tersec: `sidecar: T, ` can be on the same line as the rest

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8275/commits/a493a3b0b4c603cfe48a60a285ea75fe6d098131

PR #8280, opened 2026-04-17
use justified rather than fcr for engine API safe block hash

tersec:
Otherwise Nimbus triggers
- https://github.com/NethermindEth/nethermind/issues/11185


etan-status (approved):
Makes sense, and shouldn't change EF tests behaviour.

For users interested in FCR, can direct them to v26.3.1. Maybe worth an entry in release notes.

khanbilal732 (approved):

PR #8271, opened 2026-04-14
add proposer preferences gossip validation and publishing

Tomi-3-0:

PR #8163, opened 2026-03-24
test(ci): check if macos build is stable on nix-darwin host

markoburcul:
Adjustments to the scripts to also work on the MacOS hosts where we configure them using nix-darwin.

Reason for this change is that we are in the process of removing homebrew from our MacOS CI hosts and replacing it with nix-darwin.

PR #8284, opened 2026-04-17
implement proposer preferences sending

Tomi-3-0:
completes this https://github.com/status-im/nimbus-eth2/pull/8231 and https://github.com/status-im/nimbus-eth2/pull/8271

### DIFF
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -972,6 +972,44 @@ proc sendPayloadAttestations(
           node, fork, genesis_validators_root, validator, vidx, slot,
           target.blck.root)
 
+proc sendProposerPreferences(
+    node: BeaconNode, head: BlockRef,
+    slot: Slot) {.async: (raises: [CancelledError]).} =
+
+  if node.dag.cfg.consensusForkAtEpoch(slot.epoch) < ConsensusFork.Gloas:
+    return
+
+  let
+    fork = node.dag.forkAtEpoch(slot.epoch)
+    genesis_validators_root = node.dag.genesis_validators_root
+
+  withState(node.dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      for validator in node.attachedValidators[].items:
+        let validator_index =
+          validator.index.valueOr:
+            continue
+
+        for proposal_slot in get_upcoming_proposal_slots(
+            forkyState.data, validator_index.uint64):
+          let
+            data = ProposerPreferences(
+              validator_index: validator_index.uint64,
+              proposal_slot: proposal_slot)
+            signatureRes = await validator.getProposerPreferencesSignature(
+              fork, genesis_validators_root, data)
+
+          if signatureRes.isErr:
+            warn "Unable to sign proposer preferences",
+              validator = shortLog(validator),
+              error_msg = signatureRes.error
+            continue
+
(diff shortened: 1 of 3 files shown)


tersec:
Update to current `unstable` to get
- https://github.com/status-im/nimbus-eth2/pull/8163

 to fix `jenkins/nimbus-eth2/macos/aarch64/main` CI.

beacon_chain/validators/beacon_validators.nim
```diff
@@ -972,6 +972,44 @@ proc sendPayloadAttestations(
...
+          if signatureRes.isErr:
+            warn "Unable to sign proposer preferences",
+              validator = shortLog(validator),
+              error_msg = signatureRes.error
+            continue
+
+          let signed = SignedProposerPreferences(
+            message: data, signature: signatureRes.get)
+          
+          discard await node.router.routeProposerPreferences(signed)
```
tersec: https://github.com/status-im/nimbus-eth2/blob/6fff96c7a9cf4cd290b49ffaa2ccdbc6a1a5936e/beacon_chain/validators/message_router.nim#L733-L758 returns a `SendResult` but it's unclear this is ever used, if it's `discard`ed here?

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1405,6 +1443,9 @@ proc handleValidatorDuties*(node: BeaconNode, lastSlot, slot: Slot) {.async: (ra
     await sleepAsync(payloadAttestationCutOff.offset)
 
   sendPayloadAttestations(node, head, slot)
+  
+  if slot.is_epoch:
```
tersec: What happens if the node joins mid-epoch and already gets a proposal before the next epoch boundary?

tersec: https://github.com/status-im/nimbus-eth2/pull/8284/changes/0b677129b9b8740d1c31cf857c330d47a6a05403

beacon_chain/validators/beacon_validators.nim
Tomi-3-0: this way it matches the existing pattern of the other `route` calls, keeps it consistent with the rest?

tersec: It looks like `routeExecutionPayloadEnvelope()` and `routePayloadAttestationMessage()` have the same issue though, of returning a never-consumed value. Fixing them, per se, is outside of the scope of this PR, but it's not worth maintaining a pattern for this sake, and it's misleading: it suggests to the reader of `routeProposerPreferences()` (or `routeExecutionPayloadEnvelope()` or `routePayloadAttestationMessage()`) that errors or other anomalous conditions can be handled elsewhere, when in fact they can't. It's useful to know the dataflow in this regard.

So consistency on this point isn't the priority, but enabling someone reading the code to understand what can and cannot happen.

Separately, it's worth considering if, instead, an error condition should be handled and surfaced to logs and/or metrics.

Tomi-3-0: I've updated all three to not return a `SendResult` in this PR. The beacon-APIs spec defines `[/eth/v1/beacon/pool/payload_attestations`](https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/pool/payload_attestations.yaml) and when that endpoint is implemented `sendResult should be needed for the REST handlers, but it's trivial readding it

PR #8295, opened 2026-04-20
Use updated engineAPI endpoints for gloas

Tomi-3-0:


tests/test_el_manager.nim
```diff
@@ -125,6 +125,17 @@ proc setupMockEngineAPI*(server: RpcHttpServer, state: MockEngineState) =
 
     return GetPayloadV4Response()
```
Tomi-3-0, line 126: Once gloas.ExecutionPayloadForSigning is fully updated from with alpha.5 changes, the `getPayload` tests can switch from `electra.ExecutionPayloadForSigning` to the gloas type, and the duplicate V4 mock handler can just be removed

PR #8299, opened 2026-04-20
implement proposer preferences sending

etan-status:
completes this https://github.com/status-im/nimbus-eth2/pull/8231 and https://github.com/status-im/nimbus-eth2/pull/8271

-------

Replay #8284, as `unstable` was force-pushed


etan-status:
Same .patch as #8284.

PR #8302, opened 2026-04-20
speedup block processing with batch verification

agnxsh:
we always get >= 65 columns as a `BlockEntry` in our block processor, the batch kzg verification should significantly speed up block processing

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -202,17 +202,15 @@ proc verifySidecars(
     template kzgCommits(): auto =
       bid.message.blob_kzg_commitments.asSeq
     if columns.len > 0 and kzgCommits.len > 0:
-      for i in 0 ..< columns.len:
-        let r = verify_data_column_sidecar_kzg_proofs(
-          columns[i][], bid.message.blob_kzg_commitments)
-        if r.isErr():
-          debug "data column validation failed",
-            blockRoot = shortLog(signedBlock.root),
-            column_sidecar = shortLog(columns[i][]),
-            blck = shortLog(signedBlock.message),
-            signature = shortLog(signedBlock.signature),
-            msg = r.error()
-          return err(VerifierError.Invalid)
+      let r = verify_data_column_sidecar_kzg_proofs(
+        columns, bid.message.blob_kzg_commitments)
+      if r.isErr():
+        debug "data column validation failed",
+          blockRoot = shortLog(signedBlock.root),
+          blck = shortLog(signedBlock.message),
+          signature = shortLog(signedBlock.signature),
+          msg = r.error()
+        return err(VerifierError.Invalid)
   ok()
 
 proc verifySidecars(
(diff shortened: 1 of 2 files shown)


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -202,17 +202,15 @@ proc verifySidecars(
...
-          return err(VerifierError.Invalid)
+      let r = verify_data_column_sidecar_kzg_proofs(
+        columns, bid.message.blob_kzg_commitments)
+      if r.isErr():
+        debug "data column validation failed",
+          blockRoot = shortLog(signedBlock.root),
+          blck = shortLog(signedBlock.message),
+          signature = shortLog(signedBlock.signature),
+          msg = r.error()
+        return err(VerifierError.Invalid)
```
tersec, line 213: Conveniently, though this doesn't have a fallback for detecting which column failed, none of its callers ever noticed/cared before, so it's not a regression.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -601,6 +602,13 @@ proc verify_data_column_sidecar_kzg_proofs*(
     return ok()
   verify_data_column_sidecar_kzg_proofs(sidecars, sidecars[0].kzg_commitments)
 
+proc verify_data_column_sidecar_kzg_proofs*(
+    sidecars: openArray[ref fulu.DataColumnSidecar]): Result[void, cstring] =
+  ## Batch verify KZG proofs across multiple fulu DataColumnSidecars (ref variant).
+  if sidecars.len == 0:
+    return ok()
```
tersec: This wraps a function which already, as the first thing it does, has this 0-length check. It's redundant here.

beacon_chain/gossip_processing/block_processor.nim
agnxsh: Yes, usually if this fails we try to make re entry of the block to the processor.

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -601,6 +602,13 @@ proc verify_data_column_sidecar_kzg_proofs*(
...
+  verify_data_column_sidecar_kzg_proofs(sidecars, sidecars[0].kzg_commitments)
```
tersec: How problematic is it if the different sidecars have different KZG commitments? Invalid KZG commitments formerly would have lead to those particular columns being invalid, but not anymore unless it's the 0-index one.

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8302/changes/62868253b1b5e59c7babb0fe5186a754e17c24b7

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8302/changes/62868253b1b5e59c7babb0fe5186a754e17c24b7

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -592,14 +586,27 @@ proc verify_data_column_sidecar_kzg_proofs*[
   if not res:
     return err("DataColumnSidecar: validation failed")
 
-  ok()
+  return ok()
```
tersec: ?

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -577,10 +571,10 @@ proc verify_data_column_sidecar_kzg_proofs*[
     cells = newSeqOfCap[KzgCell](totalCells)
     proofs = newSeqOfCap[KzgProof](totalCells)
 
-  for sidecar in sidecars:
+  for sidecar {.inject.} in sidecarsArg:
     let idx = CellIndex(sidecar.index)
-    for i in 0 ..< sidecar.column.len:
-      commitments.add(kzg_commitments[i])
+    for i {.inject.} in 0 ..< sidecar.column.len:
```
tersec: This loop variable is now effectively part of `verifyKzgBatchBody`'s interface and warrants a more descriptive name than the usual throwaway `i`, `j`, `k`, etc loop variable names.

PR #8303, opened 2026-04-20
store full proposer preferences for bid gossip validation

Tomi-3-0:


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -152,7 +152,7 @@ type
     lightClientPool: ref LightClientPool
     executionPayloadBidPool*: ref ExecutionPayloadBidPool
     payloadAttestationPool*: ref PayloadAttestationPool
-    seenProposerPreferences*: array[2, BitArray[int SLOTS_PER_EPOCH]]
+    seenProposerPreferences*: array[2, array[SLOTS_PER_EPOCH, Opt[ProposerPreferences]]]
```
ahshum, line 155: So after 2 epochs, how do I know if it is unseen?

Tomi-3-0: After 2 epochs it should be useless

ahshum: ~My question is there is no data pruning, how do I know if it is have-not-been-seen case?~

In terms of usefulness, I think preference messages would be useless when they are in past slots. It is used for builders for part of the block proposal. After the proposal, either it was a good or bad block, why do we still need them and keep for 1 epoch?

ahshum: cc @tersec

Tomi-3-0: Yes, preferences are useless after their slot passes. Also, they could be cleaned per-slot but is there any benefit? we would need to track past-slots and clear those entries. the epoch boundary reset is simpler and achieves basically the same thing (and only carried out once every epoch) with at the very most 32 stale enntries sitting for a few extra slots before wiping them off.

PR #8308, opened 2026-04-21
fix ImplicitTemplateRedefinition warning and prevent them from reoccurring

tersec:
```
nimbus-eth2/beacon_chain/spec/helpers_el.nim(37, 7) Warning: template 'bid' is implicitly redefined; this is deprecated, add an explicit .redefine pragma [ImplicitTemplateRedefinition]
```


ahshum (approved):
Thanks I was also fixing this.

PR #8314, opened 2026-04-21
add batch validation path for column to speedup gossip validaiton

agnxsh:


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -414,6 +427,77 @@ proc processBlobSidecar*(
...
+  let cres = self.dataColumnQuarantine[].popSidecars(block_root)
+  if cres.isNone():
+    return
+
+  let blckOpt = self.quarantine[].popSidecarless(block_root)
+  if blckOpt.isNone():
+    return
+
+  withBlck(blckOpt.get()):
+    when (consensusFork >= ConsensusFork.Fulu) and
```
tersec: This is just exactly `consensusFork == ConsensusFork.Fulu`.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -414,6 +427,77 @@ proc processBlobSidecar*(
...
+
+  var pending: seq[ref fulu.DataColumnSidecar]
+  swap(pending, self.pendingColumnSidecars)
+
+  var sidecars = newSeqOfCap[fulu.DataColumnSidecar](pending.len)
+  for s in pending:
+    sidecars.add(s[])
+
+  data_column_kzg_batch_verifications.inc()
+  let batchRes = verify_data_column_sidecar_kzg_proofs(sidecars)
```
```nim
  let batchRes = verify_data_column_sidecar_kzg_proofs(pending.mapIt(it[]))
```
and can remove all of
```nim
  var sidecars = newSeqOfCap[fulu.DataColumnSidecar](pending.len)
  for s in pending:
    sidecars.add(s[])
```
which clarifies the logic. But I don't understand why it needs this at all `ref fulu.DataColumnSidecar` to `fulu.DataColumnSidecar` conversion at all, since
- https://github.com/status-im/nimbus-eth2/pull/8302 appears to deliberately add overloads for both `openArray[fulu.DataColumnSidecar]` and `openArray[ref fulu.DataColumnSidecar]`: https://github.com/status-im/nimbus-eth2/blob/62868253b1b5e59c7babb0fe5186a754e17c24b7/beacon_chain/spec/peerdas_helpers.nim#L591-L609

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -451,27 +537,100 @@ proc processDataColumnSidecar*(
...
-    let cres = self.dataColumnQuarantine[].popSidecars(block_root)
-    if cres.isSome():
-      let blck = self.quarantine[].popSidecarless(block_root).expect("checked above")
-      withBlck(blck):
-        when (consensusFork >= ConsensusFork.Fulu) and
-          (consensusFork < ConsensusFork.Gloas):
-          self.blockProcessor.enqueueBlock(MsgSource.gossip, forkyBlck, cres)
-        else:
-          raiseAssert "Wrong fork for columns: " & $consensusFork
+  let batchThreshold = max(self.dataColumnQuarantine[].custodyColumns.len, 1)
```
tersec: When is `self.dataColumnQuarantine[].custodyColumns.len < 4`? But if it's not, then this `max` doesn't do anything. What is it accomplishing?

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -488,21 +647,27 @@ proc processDataColumnSidecar*(
...
-  self.gloasColumnQuarantine[].put(
-    dataColumnSidecar.beacon_block_root, newClone(dataColumnSidecar))
-  self.blockProcessor.enqueuePayload(dataColumnSidecar.beacon_block_root)
+  # Non-KZG checks passed — park the sidecar in the pending buffer. KZG
+  # proofs are verified in a single batched call once the buffer has grown
+  # to the number of columns this node custodies.
+  self.pendingGloasColumnSidecars.add(newClone(dataColumnSidecar))
+
+  let batchThreshold =
+    max(self.gloasColumnQuarantine[].custodyColumns.len, 1)
```
tersec: Same question about the purpose of `max` here: this only matters if `self.gloasColumnQuarantine[].custodyColumns.len` is 0, but that should never happen. If it does, there are bigger problems.

tersec: i.e. a network with 0 custody columns also means not doing anything with `gloasColumnQuarantine` or `columnQuarantine` at all, not just tweaking around the edges.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -451,27 +537,100 @@ proc processDataColumnSidecar*(
...
+
+  var batchOk = false
+  if commitmentsOpt.isSome():
+    var sidecars = newSeqOfCap[gloas.DataColumnSidecar](pending.len)
+    for s in pending:
+      sidecars.add(s[])
+
+    data_column_kzg_batch_verifications.inc()
+    let batchRes = verify_data_column_sidecar_kzg_proofs(
+      sidecars, commitmentsOpt.get())
```
```nim
    let batchRes = verify_data_column_sidecar_kzg_proofs(
      pending.mapIt(it[]), commitmentsOpt.get())
```
makes the logic clearer here by connecting it to its inputs more directly and simply. But it raises the same question of why do this `ref` to non-`ref` conversion at all, per https://github.com/status-im/nimbus-eth2/pull/8314/changes#r3121047564

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -414,6 +427,77 @@ proc processBlobSidecar*(
...
+    return
+
+  withBlck(blckOpt.get()):
+    when (consensusFork >= ConsensusFork.Fulu) and
+        (consensusFork < ConsensusFork.Gloas):
+      self.blockProcessor.enqueueBlock(MsgSource.gossip, forkyBlck, cres)
+    else:
+      raiseAssert "Wrong fork for columns: " & $consensusFork
+
+proc flushPendingColumnSidecars*(self: var Eth2Processor) =
```
tersec: Would be clearer to name this `flushFuluPendingColumnSidecars`. For legacy reasons a number of column-related functions were written before any support for Gloas existed, and something with `functionName` became `functionNameWithGloas` without renaming the original function, and there's an argument for that, but there isn't a viable one for newly written functions.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -451,27 +537,100 @@ proc processDataColumnSidecar*(
     data_column_sidecars_dropped.inc(1, [$v.error[0]])
     return v
 
-  let block_root = hash_tree_root(block_header)
+  # Non-KZG checks passed — park the sidecar in the pending buffer. KZG
+  # proofs are verified in a single batched call once the buffer has grown
+  # to the number of columns this node custodies, i.e. the largest batch
+  # we can reasonably expect to assemble per slot.
```
tersec: This doesn't suffice for [gossip validation](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.5/specs/fulu/p2p-interface.md#data_column_sidecar_subnet_id):
> - _[REJECT]_ The sidecar's `kzg_commitments` field inclusion proof is valid as verified by `verify_data_column_sidecar_inclusion_proof(sidecar)`.
> - _[REJECT]_ The sidecar's column data is valid as verified by `verify_data_column_sidecar_kzg_proofs(sidecar)`.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -488,21 +647,27 @@ proc processDataColumnSidecar*(
...
     data_column_sidecars_dropped.inc(1, [$v.error[0]])
     return v
 
-  debug "Data column validated"
-  self.gloasColumnQuarantine[].put(
-    dataColumnSidecar.beacon_block_root, newClone(dataColumnSidecar))
-  self.blockProcessor.enqueuePayload(dataColumnSidecar.beacon_block_root)
+  # Non-KZG checks passed — park the sidecar in the pending buffer. KZG
+  # proofs are verified in a single batched call once the buffer has grown
+  # to the number of columns this node custodies.
```
tersec: This doesn't suffice for [gossip validation](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.5/specs/gloas/p2p-interface.md#data_column_sidecar_subnet_id):
> - _[REJECT]_ The sidecar is valid as verified by `verify_data_column_sidecar(sidecar, bid.blob_kzg_commitments)`.
> - _[REJECT]_ The sidecar's column data is valid as verified by `verify_data_column_sidecar_kzg_proofs(sidecar, bid.blob_kzg_commitments)`.
> - _[IGNORE]_ The sidecar is the first sidecar for the tuple `(sidecar.beacon_block_root, sidecar.index)` with valid kzg proof.

tersec:
Merge conflict(s) in `beacon_chain/gossip_processing/eth2_processor.nim` mean CI can't properly run.

beacon_chain/gossip_processing/batch_kzg_validation.nim
```diff
@@ -0,0 +1,225 @@
...
+proc lookupCommitments(
+    dag: ChainDAGRef, sidecar: gloas.DataColumnSidecar
+): Opt[KzgCommitments] =
+  let
+    bsi = dag.getBlockIdAtSlot(sidecar.slot).valueOr:
+      return Opt.none(KzgCommitments)
+    forkedBlock = dag.getForkedBlock(bsi.bid).valueOr:
+      return Opt.none(KzgCommitments)
+  withBlck(forkedBlock):
+    when consensusFork == ConsensusFork.Gloas:
```
tersec, line 114: Heze looks also to use Gloas columns.

beacon_chain/gossip_processing/batch_kzg_validation.nim
```diff
@@ -0,0 +1,225 @@
...
+      notice "KZG batch queue pruned, skipping column validation",
+        batches = batchKzg.batches.len
+      batchKzg.pruneTime = startTick
+    batch[].skip()
+    batch_kzg_batches_skipped.inc()
+    return
+
+  var sidecars = newSeqOfCap[T](batch[].items.len)
+  for it in batch[].items:
+    sidecars.add(it.sidecar[])
```
tersec, line 146: `let sidecars = mapIt(...)` (or `foo.mapIt(...)`, either way).

beacon_chain/gossip_processing/batch_kzg_validation.nim
```diff
@@ -0,0 +1,225 @@
...
+  if ok:
+    batch[].completeAll(BatchResult.Valid)
+    batch_kzg_verifications.inc()
+  else:
+    # Whole-batch verify failed; re-run per-sidecar to identify the bad
+    # apple(s). Items that pass individually still ACCEPT.
+    batch_kzg_fallbacks.inc()
+    debug "kzg batch - failure, falling back per-sidecar",
+      items = batch[].items.len
+    for it in batch[].items.mitems():
```
tersec, line 162: Why `mitems()`?

Yes, it returns effectively a reference, but semantically these aren't supposed to be modified here. This requires checking Nim's codegen.

beacon_chain/gossip_processing/batch_kzg_validation.nim
```diff
@@ -0,0 +1,225 @@
...
+  let fut = newFuture[BatchResult]("scheduleColumnKzgCheck")
+
+  let blockRoot = sidecarBlockRoot(sidecar[])
+  let commitmentsOpt = lookupCommitments(batchKzg.dag, sidecar[])
+  if commitmentsOpt.isNone():
+    # Block not (yet) available — we can't verify KZG. IGNORE so the caller
+    # neither relays nor descores; the sidecar may arrive again once the
+    # block is in the dag.
+    fut.complete(BatchResult.Timeout)
+    return fut
```
tersec, line 220: `let commitments = lookupCommitments(batchKzg.dag, sidecar[]).valueOr: ...`

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -430,36 +483,33 @@ proc processDataColumnSidecar*(
...
     return v
 
-  let block_root = hash_tree_root(block_header)
-
-  debug "Data column validated, putting data column in quarantine"
-  if dataColumnSidecar.index notin self.dataColumnQuarantine[].custodyMap:
+  # Non-KZG checks passed — schedule batched KZG verification and wait for
+  # the verdict before returning to libp2p. Returning ACCEPT prior to KZG
+  # would relay possibly invalid sidecars (strict reject under the gossip
+  # rules), which is what the legacy flush-after-the-fact path did.
```
tersec, line 489: Does the "legacy" path even exist anymore in this branch/PR? If not, it's a bit odd for this comment to discuss it per se. It can discuss the design motivations of the current design, yeah, and even what a conceptually serial implementation might do as a model but referring to a specific implementation which doesn't exist anymore in this context isn't useful.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -430,36 +483,33 @@ proc processDataColumnSidecar*(
...
+  let sidecarRef = newClone(dataColumnSidecar)
```
tersec, line 490: Another random copy of a potentially nontrivially-sized object, especially with 21 or more rows/blobs.

It's worth looking through these designs with an overall eye towards keeping consistency so these broadly pointless copies don't have to happen (it's not some intrinsic need to have a `ref` because it's being stored outside the caller's original lifetime: that's taken care of by the `seq` it gets stored into).

Not unique to this PR, but this PR introduces more of it, with a wholly new subsystem.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -472,28 +522,34 @@ proc processDataColumnSidecar*(
...
     return errIgnore("Data column before genesis")
 
-  debug "Data column received (Gloas - quarantine not implemented)"
+  debug "Data column received (Gloas)"
 
   let v = self.dag.validateDataColumnSidecar(
     self.quarantine, self.gloasColumnQuarantine, self.executionPayloadBidPool,
-    dataColumnSidecar, wallTime, subnet_id)
+    dataColumnSidecar, wallTime, subnet_id,
+    checkKzgProofs = false)
```
tersec, line 530: Not sure why this isn't part of the previous line?

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -472,28 +522,34 @@ proc processDataColumnSidecar*(
...
 
   if v.isErr():
     debug "Dropping data column", error = v.error()
     data_column_sidecars_dropped.inc(1, [$v.error[0]])
     return v
 
-  debug "Data column validated"
-  if dataColumnSidecar.index notin self.gloasColumnQuarantine[].custodyMap:
+  let sidecarRef = newClone(dataColumnSidecar)
```
tersec, line 537: See:
- https://github.com/status-im/nimbus-eth2/pull/8314/changes#r3223665759

beacon_chain/gossip_processing/batch_kzg_validation.nim
```diff
@@ -0,0 +1,225 @@
...
+): Opt[KzgCommitments] =
+  let
+    bsi = dag.getBlockIdAtSlot(sidecar.slot).valueOr:
+      return Opt.none(KzgCommitments)
+    forkedBlock = dag.getForkedBlock(bsi.bid).valueOr:
+      return Opt.none(KzgCommitments)
+  withBlck(forkedBlock):
+    when consensusFork == ConsensusFork.Gloas:
+      if forkyBlck.root != sidecar.beacon_block_root:
+        return Opt.none(KzgCommitments)
```
tersec, line 116: Does this work across forks? That is, if the node receiving this things that the block at `sidecar.slot` has some `forkyBlck.root` because it's on a different for than the sending node, it will presumably not ever match the root.

`getBlockIdAtSlot` is specifically for the canonical fork, and it has its uses, but one of them can't really be validating blocks on a noncanonical fork.

beacon_chain/gossip_processing/batch_kzg_validation.nim
```diff
@@ -0,0 +1,225 @@
...
+    batchKzg: ref BatchKzg[T], batch: ref KzgBatch[T]
+) {.async: (raises: [CancelledError]).} =
+  if batch[].items.len == 0:
+    return
+
+  let
+    startTick = Moment.now()
+    slotDuration = batchKzg.dag.cfg.timeParams.SLOT_DURATION
+
+  # Stale batches are dropped: callers receive Timeout (→ IGNORE) so we
```
tersec, line 133: Not all editors, viewers, or fonts handle these `→` Unicode characters well. Nicer to avoid them when they don't add real value (apply this comment to all `→` in this PR.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -658,7 +658,8 @@ proc validateDataColumnSidecar*(
     dag: ChainDAGRef, quarantine: ref Quarantine,
     dataColumnQuarantine: ref ColumnQuarantine,
     data_column_sidecar: fulu.DataColumnSidecar,
-    wallTime: BeaconTime, subnet_id: uint64):
+    wallTime: BeaconTime, subnet_id: uint64,
+    checkKzgProofs = true):
```
tersec, line 662: More substantively, what is this parameter even for and why is it optional?

There appears to be, in the `unstable` branch, exactly one caller of each of the fulu and gloas versions:
https://github.com/status-im/nimbus-eth2/blob/767d79b1be32ba6a82831b34158045cf607cf165/beacon_chain/gossip_processing/eth2_processor.nim#L396-L423 and https://github.com/status-im/nimbus-eth2/blob/767d79b1be32ba6a82831b34158045cf607cf165/beacon_chain/gossip_processing/eth2_processor.nim#L459-L479

This PR calls both, unconditionally, with this `checkKzgProofs = false`. So it's unclear why the parameter exists or why the codepaths still exist. They're dead code now.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -472,28 +522,34 @@ proc processDataColumnSidecar*(
...
+    ok()
+  of BatchResult.Invalid:
+    debug "Dropping Gloas data column after KZG verification failed",
+      index = dataColumnSidecar.index
+    data_column_sidecars_dropped.inc(1, ["DataColumnSidecarKzgInvalid"])
+    errReject("DataColumnSidecar: KZG verification failed")
+  of BatchResult.Timeout:
+    debug "Gloas data column KZG batch timed out",
+      index = dataColumnSidecar.index
+    errIgnore("DataColumnSidecar: KZG batch timeout")
```
tersec, line 552: All of this `BatchResult` handling, `scheduleColumnKzgCheck`, etc should be in the gossip validation function itself, not here.

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -430,36 +483,33 @@ proc processDataColumnSidecar*(
...
+    ok()
+  of BatchResult.Invalid:
+    debug "Dropping data column after KZG verification failed",
+      index = dataColumnSidecar.index
+    data_column_sidecars_dropped.inc(1, ["DataColumnSidecarKzgInvalid"])
+    errReject("DataColumnSidecar: KZG verification failed")
+  of BatchResult.Timeout:
+    debug "Data column KZG batch timed out",
+      index = dataColumnSidecar.index
+    errIgnore("DataColumnSidecar: KZG batch timeout")
```
tersec, line 506: - https://github.com/status-im/nimbus-eth2/pull/8314/changes#r3223810358

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8314/commits/0940ec93383f18f31b40b17830f9a8d6033af684

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -790,7 +793,8 @@ proc validateDataColumnSidecar*(
     gloasColumnQuarantine: ref GloasColumnQuarantine,
     executionPayloadBidPool: ref ExecutionPayloadBidPool,
     data_column_sidecar: gloas.DataColumnSidecar,
-    wallTime: BeaconTime, subnet_id: uint64):
+    wallTime: BeaconTime, subnet_id: uint64,
+    checkKzgProofs = true):
```
tersec, line 797: - https://github.com/status-im/nimbus-eth2/pull/8314#discussion_r3223804472

PR #8335, opened 2026-04-25
Update block gossip validation; Remove isParentBlockFull

ahshum:
- Update to get execution parent safely, and validations with it.
- Added unviable envelope storage for addressing execution invalid cases.


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -422,23 +423,49 @@ template validateBeaconBlockGloas(
...
+    else:
+      # Parent's payload status is EMPTY, i.e. check with grandparent.
+      let res = block:
+        if not isNil(parent.parent):
+          # Grandparent is non-finalized yet.
+          let gpBhash = dag.loadExecutionBlockHash(parent.parent).valueOr:
+            return errIgnore("validateBeaconBlockGloas: cannot load block hash")
+          (parent.parent.bid, gpBhash)
+        else:
+          # Grandparent should be either finalized or not existed.
```
tersec: "not existed" -> "nonexistent"

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -422,23 +423,49 @@ template validateBeaconBlockGloas(
...
-  debugGloasComment("request missing envelope if not found in db")
-  debugGloasComment("revisit the naive parent.parent.isNil guard")
-  if not (
-      isParentBlockFull(dag, signed_beacon_block, parent) or
-      parent.parent.isNil or
-      isParentBlockFull(dag, signed_beacon_block, parent.parent)
-  ):
-    # REJECT only once EL verification complete and parent doesn't validate.
-    return errIgnore("validateBeaconBlockGloas: parent execution payload not yet verified")
+  debugGloasComment("execution valid")
```
tersec: Is the premise that this is covered by the previous validation now or that it's still missing?

ahshum: Yes I would say so. By "execution_payload verification" it should refer to the result of 
https://github.com/status-im/nimbus-eth2/blob/41d4f812078df82d80901e3144feac1ac3ef61a7/beacon_chain/gossip_processing/block_processor.nim#L517-L521
or `engine_newPayload` request, which we don't store the result persistently. I think we might need a new storage or, for example, `loadExecutionValid` for `BlockRef` so that we can use 
https://github.com/status-im/nimbus-eth2/blob/41d4f812078df82d80901e3144feac1ac3ef61a7/beacon_chain/consensus_object_pools/block_dag.nim#L247-L248
for this.

ahshum: Just would like to add that envelope should be EL verified when it is stored into the database, so it is cheap to check. The problem is with the EIP8237. I will bring up a discussion regarding this and the EIP.

ahshum: I think adding unviable to envelope quarantine would help. Will include that here.

beacon_chain/consensus_object_pools/envelope_quarantine.nim
```diff
@@ -17,7 +18,13 @@ const
   MaxMissingItems* = 1024
     ## Revisit the setting and same as block quarantine for now
 
+const
```
tersec: Don't need this `const`, can be part of the same `const` block as precedes it.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8335/changes/3c0ea10aa624418d81397f6217f49a2d3ac68f09

beacon_chain/consensus_object_pools/envelope_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8335/changes/a4833f4b12e7927b63f96e779a7bc61c797b3b60

tersec:
> Edit: we could check if an envelope exists in the db but we may not be able to handle the case if EIP8237 (range sync without envelopes) is merged.

It was not chosen to be merged, so this is a case that does not have to be handled.

ahshum:
> It was not chosen to be merged, so this is a case that does not have to be handled.

Sure updated. It is not included in the PR.

tersec:
Does `skipLastEnvelope` have any semantics?

```
$ rg skipLastEnvelope -C3
beacon_chain/extras.nim
38-    ## When process_slots() is being called as part of a state_transition(),
39-    ## the hash_tree_root() from the block will fill in the state.root so it
40-    ## should skip calculating that last state root.
41:    skipLastEnvelope ##\
42-    ## As a full beacon block consists of a beacon block and envelope since
43-    ## Gloas, the last envelope may not be there yet. This is used to toggle the
44-    ## strict state transition of the last envelope in updateState().

beacon_chain/consensus_object_pools/block_clearance.nim
526-  let blckBsi = BlockSlotId.init(blck.bid, envelopeSlot)
527-  if not updateState(
528-      dag, dag.clearanceState, blckBsi, false, cache,
529:      dag.updateFlags + {skipLastEnvelope}):
530-    # If updateState() fails, it means there may be some missing blocks and
531-    # envelopes of its parents, or the database is corrupted.
532-    error "Unable to load clearance state for envelope, database corrupt?",
--
541-  ):
542-    debug "Envelope is not for the current head"
543-    return err(VerifierError.Invalid)
544:  # With skipLastEnvelope flag and containsExecutionPayloadEnvelope() check
545-  # above, the envelope should have not been applied but double check.
546-  elif dag.clearanceState.forky(consensusFork).data.latest_block_hash ==
547-       signedEnvelope.message.payload.block_hash:

beacon_chain/consensus_object_pools/blockchain_dag.nim
2549-  var cache: StateCache
2550-  if not updateState(
2551-      dag, dag.headState, newHead.bid.atSlot(), false, cache,
2552:      dag.updateFlags + {skipLastEnvelope}):
2553-    # Advancing the head state should never fail, given that the tail is
2554-    # implicitly finalised, the head is an ancestor of the tail and we always
2555-    # store the tail state in the database, as well as every epoch slot state in
```

It seems to be only set, not checked.

ahshum:
> Does `skipLastEnvelope` have any semantics?
>
> It seems to be only set, not checked.

Yes it is no longer needed. There are updates and cleaning up required to happen around envelope processing. Working on a separate PR for this.

PR #7808, opened 2025-12-16
Gloas ForkChoice

Tomi-3-0:

PR #8337, opened 2026-04-25
implement verify_execution_payload_envelope

Tomi-3-0:
adds `verify_execution_payload_envelope` implementing just the spec's pure verification checks needed for forkchoice envelope validation

### DIFF
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -1840,6 +1840,67 @@ proc process_block*(
 
   ok(operations_rewards)
 
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.5/specs/gloas/fork-choice.md#new-verify_execution_payload_envelope
+proc verify_execution_payload_envelope*(
+    timeParams: TimeParams,
+    fork: Fork,
+    state: gloas.BeaconState | heze.BeaconState,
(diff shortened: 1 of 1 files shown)


beacon_chain/spec/state_transition_block.nim
```diff
@@ -1840,6 +1840,67 @@ proc process_block*(
...
+
+  if not verify_execution_payload_envelope_signature(
+      fork, genesis_validators_root,
+      payload.slot_number.epoch,
+      envelope, pubkey, signed_envelope.signature):
+    return err("verify_execution_payload_envelope: invalid signature")
+
+  # Verify consistency with the beacon block
+  var header = state.latest_block_header
+  header.state_root = hash_tree_root(state)
```
tersec: Can this be avoided? `hash_tree_root(BeaconState)` is very expensive.

Tomi-3-0: well, yes.  the proposed caller ( `on_execution_payload` ) alreasy checks if the beacon block root is in the proto array before calling it. that's the way it's actually implemented [here](https://github.com/status-im/nimbus-eth2/pull/7808/changes#diff-88df9240ff1f433c6740570051a2ecfc5c7cf1f4d6faa67160d711131a640ebaR1844-R1894)

PR #8338, opened 2026-04-26
spec-align Gloas gossip validation REJECT/IGNORE; properly handle builder exit signatures

tersec:


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -411,28 +411,34 @@ template validateBeaconBlockGloas(
...
   # If execution_payload verification of block's execution payload parent by an
   # execution node is complete:
   #
   # - [REJECT] The block's execution payload parent (defined by
   #   bid.parent_block_hash) passes all validation.
   let parent = dag.getBlockRef(bid.parent_block_root).valueOr:
-    return dag.checkedReject("validateBeaconBlockGloas: invalid execution parent")
+    return errIgnore("validateBeaconBlockGloas: parent not yet seen")
   debugGloasComment("request missing envelope if not found in db")
   debugGloasComment("revisit the naive parent.parent.isNil guard")
```
ahshum, line 434: I think ~`The block's parent execution payload` means parent envelope while `The block's execution payload parent` is the envelope of the parent block by block hash.~ the validations are around the parent envelope by block hash.

Both don't seem that we have checked precisely.

PR #8183, opened 2026-03-27
add execution_payload events

ahshum:
https://github.com/ethereum/beacon-APIs/pull/588

PR #8340, opened 2026-04-27
fix proposal execution head at genesis; rm non-spec SSE; ensure envelopse SSE has envelope in database

tersec:
For the proposal head at genesis, even though
- https://github.com/ethereum/consensus-specs/pull/5067

states
```python
        state.latest_block_hash = spec.Hash32()
        empty_requests_root = spec.hash_tree_root(spec.ExecutionRequests())
        state.latest_execution_payload_bid = spec.ExecutionPayloadBid(
            block_hash=spec.Hash32(eth1_block_hash),
```
the current generator does the opposite of this. So, work around this mismatch.

Should be removed when the generator is fixed.

According to https://github.com/ethereum/beacon-APIs/blob/v5.0.0-alpha.1/apis/eventstream/index.yaml `execution_payload_available` and `execution_payload_bid` exist but plain `execution_payload` does not.

### DIFF
--- a/beacon_chain/beacon_node.nim
+++ b/beacon_chain/beacon_node.nim
@@ -66,7 +66,6 @@ type
     optUpdateQueue*: AsyncEventQueue[
       RestVersioned[ForkedLightClientOptimisticUpdate]]
     optFinHeaderUpdateQueue*: AsyncEventQueue[ForkedLightClientHeader]
-    execPayloadAddedQueue*: AsyncEventQueue[EventExecutionPayloadObject]
     execPayloadGossipAddedQueue*: AsyncEventQueue[EventExecutionPayloadGossipObject]
     execPayloadAvlQueue*: AsyncEventQueue[EventExecutionPayloadAvailableObject]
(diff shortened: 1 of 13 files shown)


ahshum:
Is the eventstream topic causing the issue? Yes it is right that the topics not exists because the beacon-API PR has not been merged yet https://github.com/status-im/nimbus-eth2/pull/8183

But seems that other clients have merged the changes according to the references in the beacon-API PR.

tersec:
> Is the eventstream topic causing the issue? Yes it is right that the topics not exists because the beacon-API PR has not been merged yet #8183
> 
> But seems that other clients have merged the changes according to the references in the beacon-API PR.

Ah,
- https://github.com/ethereum/beacon-APIs/pull/588

merged 4 hours ago, after I opened this PR and enabled auto-merge.

It wasn't per se causing problems, no, but when I was looking into where to put the `onEnvelopeAvailable` callback, it proved redundant, and at the time, it was... not in specs.

It can be added back, in the correct place.

alialobidm:
``

alialobidm:
````

PR #8344, opened 2026-04-27
Fix Gloas data column quarantine crash on non-custody columns

Tomi-3-0:
Adds a custody check before putting a data column sidecar into the quarantine. When a node receives an already gossip-validated data column for a column index it doesn't have custody of  `getIndex()` returns -1, and put() hits `doAssert(index >= 0)`. This skips non-custody columns so they are never inserted into the quarantine

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -14,7 +14,7 @@ import
   kzg4844/kzg,
   ssz_serialization/types,
   ../el/el_manager,
-  ../spec/[helpers, forks],
+  ../spec/[column_map, helpers, forks],
   ../consensus_object_pools/[
     attestation_pool, blob_quarantine, block_clearance, block_quarantine,
     blockchain_dag, envelope_quarantine, execution_payload_pool,
@@ -500,9 +500,10 @@ proc processDataColumnSidecar*(
     return v
 
   debug "Data column validated"
-  self.gloasColumnQuarantine[].put(
-    dataColumnSidecar.beacon_block_root, newClone(dataColumnSidecar))
-  self.blockProcessor.enqueuePayload(dataColumnSidecar.beacon_block_root)
+  if dataColumnSidecar.index in self.gloasColumnQuarantine[].custodyMap:
+    self.gloasColumnQuarantine[].put(
+      dataColumnSidecar.beacon_block_root, newClone(dataColumnSidecar))
+    self.blockProcessor.enqueuePayload(dataColumnSidecar.beacon_block_root)
 
   data_column_sidecars_received.inc()
   v
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -454,6 +454,11 @@ proc processDataColumnSidecar*(
   let block_root = hash_tree_root(block_header)
 
   debug "Data column validated, putting data column in quarantine"
+  if dataColumnSidecar.index notin self.dataColumnQuarantine[].custodyMap:
+    data_column_sidecars_received.inc()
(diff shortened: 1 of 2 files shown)


agnxsh:
would you not subscribe to just custody columns, unless you also subscribe to non-custody columns, i don't one should be reaching this crash?

Tomi-3-0:
> would you not subscribe to just custody columns, unless you also subscribe to non-custody columns, i don't one should be reaching this crash?

don't we currently subscribe to all custody groups: https://github.com/status-im/nimbus-eth2/blob/b99895a46081812d4c2590b2e319dc55e3086fb5/beacon_chain/nimbus_beacon_node.nim#L2488 and so non custodied ones also arrive via gossip

agnxsh:
> > would you not subscribe to just custody columns, unless you also subscribe to non-custody columns, i don't one should be reaching this crash?
> 
> 
> 
> dont we currently subscribe to all custody groups: https://github.com/status-im/nimbus-eth2/blob/b99895a46081812d4c2590b2e319dc55e3086fb5/beacon_chain/nimbus_beacon_node.nim#L2488 and so non custodied ones also arrive via gossip
> 
> 

okay I'm wondering how we have never witnessed this quarantine crash in Fulu yet for quite long, it would then be reasonable to extend this condition to Fulu column processing as well

cc @tersec

tersec:
> > would you not subscribe to just custody columns, unless you also subscribe to non-custody columns, i don't one should be reaching this crash?
> 
> don't we currently subscribe to all custody groups:
> 
> https://github.com/status-im/nimbus-eth2/blob/b99895a46081812d4c2590b2e319dc55e3086fb5/beacon_chain/nimbus_beacon_node.nim#L2488
> and so non custodied ones also arrive via gossip

These don't mean noncustodied columns arrive via gossip -- that's https://github.com/status-im/nimbus-eth2/blob/102be79c06b408c4901fc73d92cbc8397c39425f/beacon_chain/nimbus_beacon_node.nim#L1366-L1384 which does only subscribe to custodied columns.

Tomi-3-0:
> > > would you not subscribe to just custody columns, unless you also subscribe to non-custody columns, i don't one should be reaching this crash?
> > 
> > 
> > don't we currently subscribe to all custody groups:
> > https://github.com/status-im/nimbus-eth2/blob/b99895a46081812d4c2590b2e319dc55e3086fb5/beacon_chain/nimbus_beacon_node.nim#L2488
> > 
> > and so non custodied ones also arrive via gossip
> 
> These don't mean noncustodied columns arrive via gossip -- that's
> 
> https://github.com/status-im/nimbus-eth2/blob/102be79c06b408c4901fc73d92cbc8397c39425f/beacon_chain/nimbus_beacon_node.nim#L1366-L1384
> 
> which does only subscribe to custodied columns.

I still think this is a good defensive fix, regardless of the underlying cause

tersec:
Generally prefer merging which which are mergeable into `unstable` into that, then keeping (as long as feasible) `glamsterdam-devnet-0` as a mirror of `unstable`, until it must diverge.

But to give one possible specific example of how this might arise: https://github.com/status-im/nimbus-eth2/blob/d8297d268de717dd264686f67b59073b55ce804c/beacon_chain/sync/validator_custody.nim#L148-L175 switches from `FullCustody` to`LimitedCustody`, but https://github.com/status-im/nimbus-eth2/blob/d8297d268de717dd264686f67b59073b55ce804c/beacon_chain/nimbus_beacon_node.nim#L1352-L1363 only subscribes to, but does not unsubscribe from, data column column topics.

Then something like this could happen: it tries to put a non-custodied column into the existing quarantine which can't handle it. But the bug is really that it should be unsubscribing from the data column sidecar topics.

tersec:
Ok, merging this. But something is still going wrong, and would be good to understand why.

PR #8346, opened 2026-04-27
avoid blocking main thread on reconstruction

arnetheduck:
When running column reconstruction, the current approach blocks until timeout or the reconstruction is finished, delaying other processing that should be happening at the same time.

With a TSP, we can allow async processing to continue while the reconstruction is done in the background also taking care to not stuff the thread pool with tasks (since it's shared).

Further work is needed to enable reconstruction, in particular:
* Reconstruction should likely not run in onSlotEnd, ie it should probably start somewhere close to the arrival of 50+% of the columns
* As reconstruction is running, we might receive columns from the network - we should top up the known colums before creating new tasks so as to avoid unnecessary recomputations
* We should be distributing the computed columns incrementally on the network instead of just saving them to the DB - this of course relies on computing them earlier - care must be taken to not start this process "too" early, ie before a new head has been established by sufficient attestations..


agnxsh:
the changes look reasonable to me, however, with some discussions in the last month or so, we have decided to use reconstruction in a different way,

1. with some optimizations such as signature caching, and batched kzg verifications at the time of gossip we have decided to expand the column quarantine to a 128 columns, thereby nullifying the need to go for slot by slot reconstruction of columns

2. in times of syncing, for the ease of syncing we have come up with a few stability modes, where we downshift our custody columns to something as low as minimum (especially in times of a turbulent sync), accordingly change what we can serve in enr and metadata, also updating our `earliest available slot` from which we can reliably serve data, this would enable us to sync a part of history as a lightweight full node and then upshift as a supernode (or as per validator balance) once we have synced

3. as per current data right now, once we are stable, most of 128 columns are provided by getBlobsV2, that's almost a 9/10 slots on Hoodi and 7/10 slots on Mainnet.

coming to real world use cases, one would need reconstruction to be able to serve blob data to L2s, for that we always need the first half of 128 columns, that should enable us serve blob data without having to reconstruct. we currently do this in our lightsupernode flag, which does custody the first half columns.


with all of these things in place, we've decided to make reconstruction a sort of lazy backfill mechanism where we can maintain a slot-keyed recovery matrix in the backfilling service, this would essentially take multiple slots to slowly reconstruct data column sidecars for every slot, and do it upto blob retention period backwards, till then we adjust our earliest_available_slot to the last stability period that we achieved (i.e, the slot from which we could reliably use getBlobs/gossip validate/short-range sync 128 columns, in case we actually do custody all 128)

agnxsh:
> lazy backfill mechanism where we can maintain a slot-keyed recovery matrix in the backfilling service, this would essentially take multiple slots to slowly reconstruct data column sidecars for every slot

so instead of spawning task for every blob idx (based on availability and other factors ofc), we can actually do 1 task for 1 blob idx per slot and keep filling the recovery matrix buffer (and i think that we can take the main thread, it won't be too harmful), so in every 20-ish odd seconds we are gradually able to revive 1 non reconstructed slot

with the onset of slot availability advertisements, reconstruction has almost turned out altruistic in today's network, in mainnet 3/10 slots are worth reconstructing, if getBlobs is well utilised, so we might as well take decently long to reconstruct for these 3 slots?

arnetheduck:
> expand the column quarantine to a 128 columns,

If this is about waiting for the columns to arrive from the network, I'd assume we need to run a reconstruction anyway in case some columns are not available. More importantly, we must start attesting as soon as we have 64 and not wait for all to arrive, since the full 128 may take time to arrive, if at all (and attestations can go out as soon as we have 64).

> and i think that we can take the main thread,

I don't think we should be running any computation of this nature on the main thread, ie it's perfectly runnable on a separate thread with negligible overhead -> perfect candidate.

> so we might as well take decently long to reconstruct for these 3 slots?

from a network health perspective, if we're going to be reconstructing we might as well do it right and publish the results on the network - the additional effort is negligible. 

This indeed implies running the reconstruction incrementally and continuously but there's no reason not to start sooner rather than layer, and using the technique in this PR which interleaves reconstruction work with other taskpool work .. importantly, we should _never_ block the main thread with these kinds of loads.

agnxsh:
> We should be distributing the computed columns incrementally on the network instead of just saving them to the DB - this of course relies on computing them earlier - care must be taken to not start this process "too" early, ie before a new head has been established by sufficient attestations..

i do agree to this, there needs to be more focus on redistributing the columns we are downloading from gossip, EL blobpool, or sync as well as reconstructing, will look into as soon as this lands (this existing approach can be used to create a lazy reconstruction backfiller as well, like i was talking about in the above thread)


> we must start attesting as soon as we have 64 and not wait for all to arrive, since the full 128 may take time to arrive

this is what we do today as well, we process the block when we about 50%+ columns (whatever comes in the `BlockEntry` batch, provided it's atleast half)

agnxsh:
but @arnetheduck to timely redistribute the columns, we need to be sufficiently performant in reconstructing columns, say <= 2 seconds into the slot (before attestation, i think a few seconds after attestation is also reasonable) and I'm not sure if we can achieve that even with the taskpool interleaving, unless sufficiently beefy hardware, i'm almost _sure_ our hardware cannot get there, the one we use for fleet.

i agree this should be the only way of doing this, however, i think the finish line goal should be have an efficient enough column retrieval from the EL and/or fast enough gossip processing strategies to entirely _mitigate_ the need to reconstruct, keeping it as low as possible per epoch.

that's because everytime we choose to reconstruct for a slot, unless we can't redistribute them via gossip within the early quarter of the slot, i don't think it's of much help.

overall, i'm not sure about one thing, does it make sense to redistribute columns of slot N-5 at slot N over gossip? 5 is just an example number here. redistributing for others' syncning needs is definitely useful.

arnetheduck:
> unless sufficiently beefy hardware

this is fine - we would likely limit redistribution to column topics we subscribe to, meaning that if the column was received by network _before_ computation finished, we would not resend it anyway.

In fact, the point of this feature is to let users with beefy hardware "help" those with weaker hardware.

> efficient enough column retrieval from the EL

the two methods are complementary and meant as safety nets on top of each other - column retrieval from the el is also not possible for private blobs so there are cases where the EL fallback doesn't work systematically - in these cases, we want reconstruction to do its magic - however, due to hardware differences, we _also_ want to make sure we don't start reconstruction for columns that arrived over the network while we were busy - this way, the beefy hardware naturally helps the less beefy hardware - this is also a core reason why we should be reconstructing a _random_ column rather than the "next" one.

agnxsh:
> if the column was received by network before computation finished, we would not resend it anyway.

makes sense, thank you.

> column retrieval from the el is also not possible for private blobs so there are cases where the EL fallback doesn't work systematically 

for mainnet the getBlobs retrival rate is roughly 60%

> this way, the beefy hardware naturally helps the less beefy hardware - this is also a core reason why we should be reconstructing a random column rather than the "next" one.

this is a reasonable approach ok

> Reconstruction should likely not run in onSlotEnd, ie it should probably start somewhere close to the arrival of 50+% of the columns

IMO in that case an event driven service could be reworked on, which triggers on the 50% callback.

agnxsh:
@tersec i would advocate for merging this, whatever service reworking design we take, the TSP approach should help. also this just helps me work on the next steps mentioned above.

PR #8347, opened 2026-04-27
Recover envelope sse events

ahshum:
- https://github.com/status-im/nimbus-eth2/pull/8340#issuecomment-4327696265


beacon_chain/consensus_object_pools/block_clearance.nim
```diff
@@ -563,12 +563,8 @@ proc addHeadExecutionPayload*(
   # Put the envelope into db and update optimistic status for the block.
   dag.db.putExecutionPayloadEnvelope(signedEnvelope)
 
-  # https://github.com/ethereum/beacon-APIs/blob/v5.0.0-alpha.1/apis/eventstream/index.yaml
```
tersec, line 548: Why remove this comment? The idea is to identify which of the now-3 envelope SSE events triggers on specific conditions, and the `dag.onEnvelopeFoo` name doesn't make it immediately obvious what the mapping to/from the SSE event name is.

ahshum: Beacon block also has 3 SSE events and there is not references/comments so I thought it is fine to not have comment. Are you suggesting to add them to all 3 places and also to block's?

ahshum: It is also confusing to add on trigger instead of the event topic, likes how we define the data types of forks.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -960,6 +960,13 @@ proc storePayload(
           chronos.nanoseconds((slotTime - wallTime).nanoseconds)
     deadline = sleepAsync(deadlineTime)
 
+  # https://github.com/ethereum/beacon-APIs/blob/31f7d04f869d40a643b68ac22e10fb27644d20e7/apis/eventstream/index.yaml
+  # execution_payload_available: The node has verified that the execution
+  # payload and blobs for a block are available and ready for payload
+  # attestation
+  if not isNil(dag.onEnvelopeAvailable):
+    dag.onEnvelopeAvailable(signedEnvelope)
```
tersec: https://github.com/status-im/nimbus-eth2/blob/c4474aaa46312f3377ed3d567a6cfca479a108af/beacon_chain/gossip_processing/block_processor.nim#L963-L986

This fires before it checks blobs.

tersec: https://github.com/status-im/nimbus-eth2/pull/8347/changes/9888ac55df6fe16128dd39be2bd986f0ad446f00

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -965,6 +965,13 @@ proc storePayload(
     blck = shortLog(blck),
     slot = signedBlock.message.slot
 
+  # https://github.com/ethereum/beacon-APIs/blob/31f7d04f869d40a643b68ac22e10fb27644d20e7/apis/eventstream/index.yaml
+  # execution_payload: The node has received a `SignedExecutionPayloadEnvelope`
+  # (from P2P or API) that is successfully imported on the fork-choice
+  # `on_execution_payload_envelope` handler
+  if not isNil(dag.onEnvelopeAdded):
+    dag.onEnvelopeAdded(signedEnvelope)
```
tersec: According to
- https://github.com/ethereum/beacon-APIs/pull/588#discussion_r2923313275

> `execution_payload_available` is emitted when the execution payload + all data (based on custody) is available, it does not require the execution payload to be valid, while `execution_payload` does since it is emitted after successful import into fork choice via `on_execution_payload` handler, so it includes the time it takes for the EL to execute and verify the execution payload

So `execution_payload` shouldn't fire at all unless the EL signals VALID, which isn't checked here.

tersec: Which is consistent with https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/gloas/fork-choice.md#new-on_execution_payload_envelope
```python
def on_execution_payload_envelope(
    store: Store, signed_envelope: SignedExecutionPayloadEnvelope
) -> None:
    # ...

    # Verify the execution payload envelope
    verify_execution_payload_envelope(state, signed_envelope, EXECUTION_ENGINE)

    # ...
```

beacon_chain/consensus_object_pools/block_clearance.nim
tersec: It's partly because getting the triggers right is the most subtle part for these, and the part where the spec verbiage most directly applies.

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8347/commits/cd33408824503d28574eab08b86a8307ff425b91

PR #8353, opened 2026-04-28
optimize data column pruning and prune payload envelopes along blocks

Tomi-3-0:
Instead of loading every block from disk in a loop just to get the consensus fork using `getForkedBlock`,  compute it once with consensusForkAtEpoch and pass it directly to `delDataColumnSidecars`.
 
When pruning post Gloas blocks also delete its executionpayload envelope if present

### DIFF
--- a/beacon_chain/nimbus_beacon_node.nim
+++ b/beacon_chain/nimbus_beacon_node.nim
@@ -1766,20 +1766,17 @@ proc pruneDataColumns(node: BeaconNode, slot: Slot) =
   let dataColumnPruneEpoch = (slot.epoch -
                               node.dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS - 1)
   if slot.is_epoch() and dataColumnPruneEpoch >= node.dag.cfg.FULU_FORK_EPOCH:
+    let consensusFork = node.dag.cfg.consensusForkAtEpoch(dataColumnPruneEpoch)
+    if consensusFork < ConsensusFork.Fulu:
+      return
+
     var blocks: array[SLOTS_PER_EPOCH.int, BlockId]
     var count = 0
     let startIndex = node.dag.getBlockRange(
       dataColumnPruneEpoch.start_slot, blocks.toOpenArray(0, SLOTS_PER_EPOCH - 1))
     for i in startIndex..<SLOTS_PER_EPOCH:
-      let blck = node.dag.getForkedBlock(blocks[int(i)]).valueOr: continue
-      withBlck(blck):
-        when consensusFork < ConsensusFork.Fulu: continue
-        else:
-          # Iterate the full column space rather than just the local custody
-          # set so late-arriving or reconstructed columns outside of this
-          # node's custody groups are also cleaned up.
-          count += node.db.delDataColumnSidecars(
-            consensusFork, blocks[int(i)].root)
+      count += node.db.delDataColumnSidecars(
+        consensusFork, blocks[int(i)].root)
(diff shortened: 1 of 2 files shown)


beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1766,20 +1766,20 @@ proc pruneDataColumns(node: BeaconNode, slot: Slot) =
   let dataColumnPruneEpoch = (slot.epoch -
                               node.dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS - 1)
   if slot.is_epoch() and dataColumnPruneEpoch >= node.dag.cfg.FULU_FORK_EPOCH:
+    let consensusFork = node.dag.cfg.consensusForkAtEpoch(dataColumnPruneEpoch)
+    if consensusFork < ConsensusFork.Fulu:
+      return
+
```
ahshum: Not sure how does this work - Both using `dataColumnPruneEpoch` and so the early return is a dead code?

Tomi-3-0: you are right.

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1766,20 +1766,17 @@ proc pruneDataColumns(node: BeaconNode, slot: Slot) =
   let dataColumnPruneEpoch = (slot.epoch -
                               node.dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS - 1)
   if slot.is_epoch() and dataColumnPruneEpoch >= node.dag.cfg.FULU_FORK_EPOCH:
+    let consensusFork = node.dag.cfg.consensusForkAtEpoch(dataColumnPruneEpoch)
     var blocks: array[SLOTS_PER_EPOCH.int, BlockId]
     var count = 0
     let startIndex = node.dag.getBlockRange(
       dataColumnPruneEpoch.start_slot, blocks.toOpenArray(0, SLOTS_PER_EPOCH - 1))
```
arnetheduck:
```suggestion
let startIndex = node.dag.getBlockRange(dataColumnPruneEpoch.start_slot, blocks)
```

beacon_chain/nimbus_beacon_node.nim
```diff
@@ -1766,20 +1766,17 @@ proc pruneDataColumns(node: BeaconNode, slot: Slot) =
...
-        else:
-          # Iterate the full column space rather than just the local custody
-          # set so late-arriving or reconstructed columns outside of this
-          # node's custody groups are also cleaned up.
-          count += node.db.delDataColumnSidecars(
-            consensusFork, blocks[int(i)].root)
+      # Iterate the full column space rather than just the local custody
+      # set so late-arriving or reconstructed columns outside of this
+      # node's custody groups are also cleaned up.
+      count += node.db.delDataColumnSidecars(
```
arnetheduck, line 1777: what happens with empty slots here?

Tomi-3-0: sorry, I missed this. 
empty slots should already be excluded from the loop since `getBlockRange`  only writes to the output when `isProposed` is `true`: https://github.com/status-im/nimbus-eth2/blob/d3c18c795a9f037e3220f756369b284dfb641e41/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1819-L1824 so, `blocks[i] for i in startIndex..<SLOTS_PER_EPOCH` always contains a real `BlockId` since results are packed without gaps

agnxsh (approved):
LGTM

agnxsh:
@tersec might want to merge this

PR #8354, opened 2026-04-28
Add Gloas light client containers

etan-status:
Define the types and helpers needed for Gloas light client specs.

- https://github.com/ethereum/consensus-specs/pull/5178


beacon_chain/spec/datatypes/gloas.nim
```diff
@@ -700,3 +702,318 @@ static:
...
+      EXECUTION_BLOCK_HASH_GINDEX_DENEB, header.beacon.body_root)
+
+  # [Modified in Gloas:EIP7732]
+  if epoch >= cfg.CAPELLA_FORK_EPOCH:
+    return is_valid_normalized_merkle_branch(
+      header.execution_block_hash, header.execution_branch,
+      EXECUTION_BLOCK_HASH_GINDEX, header.beacon.body_root)
+
+  # [Modified in Gloas:EIP7732]
+  header.execution_block_hash == ZERO_HASH and
```
tersec: == ZERO_HASH is a bit faster as `.isZero`

PR #8364, opened 2026-04-28
Fix blob_data_available for blocks without blobs

Tomi-3-0:


beacon_chain/validators/beacon_validators.nim
```diff
@@ -889,7 +889,16 @@ proc checkPayloadPresent(node: BeaconNode, blck: BlockRef): bool =
 proc checkBlobDataAvailable(node: BeaconNode, blck: BlockRef): bool =
   withConsensusFork(node.dag.cfg.consensusForkAtEpoch(blck.slot.epoch)):
     when consensusFork >= ConsensusFork.Gloas:
-      # check that our custody columns are available
+      let forkedBlock = node.dag.getForkedBlock(blck.bid).valueOr:
```
tersec: Can avoid the `Forked`/case object aspect of this, at least, with https://github.com/status-im/nimbus-eth2/blob/8202770d8c894ef480301e9ee1b9b1f36caa7490/beacon_chain/consensus_object_pools/blockchain_dag.nim#L344-L351. The fork is statically known here, can use `consensusFork.TrustedSignedBeaconBlock` via https://github.com/status-im/nimbus-eth2/blob/8202770d8c894ef480301e9ee1b9b1f36caa7490/beacon_chain/spec/forks.nim#L642-L662

PR #8372, opened 2026-04-29
Limit missing envelope requests

ahshum:


beacon_chain/consensus_object_pools/envelope_quarantine.nim
```diff
@@ -10,16 +10,24 @@
...
+    tries*: int
+
   EnvelopeQuarantine* = object
     orphans*: Table[Eth2Digest, Table[uint64, SignedExecutionPayloadEnvelope]]
       ## Envelopes that we have received but did not have a block yet. In the
       ## ideal scenario, block should arrive before envelope but that is not
       ## guaranteed.
 
-    missing*: seq[Eth2Digest]
+    missing*: Table[Eth2Digest, MissingEnvelope]
```
arnetheduck, line 30: in a follow-up PR, it would make sense to export this missing logic to a separate type and use it in all the quarantines (ie `type MissingTable` that deals with the exponential backoff, limits etc)

arnetheduck (approved):

beacon_chain/consensus_object_pools/envelope_quarantine.nim
ahshum: Sure, will refactor these to reduce repeated code.

PR #8385, opened 2026-05-01
Implement  /eth/v1/beacon/execution_payload_bid REST endpoint (POST)

Tomi-3-0:
https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/execution_payload/bid.yaml
https://github.com/ethereum/beacon-APIs/blob/master/beacon-node-oapi.yaml

### DIFF
--- a/beacon_chain/rpc/rest_beacon_api.nim
+++ b/beacon_chain/rpc/rest_beacon_api.nim
@@ -1657,24 +1657,20 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
     if contentBody.isNone():
       return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
 
-    template decodeBidAndRoute(BidType: untyped) =
-      let dres = decodeBodyJsonOrSsz(BidType, contentBody.get())
-      if dres.isErr():
-        return RestApiResponse.jsonError(dres.error())
-      let res = await node.router.routeExecutionPayloadBid(dres.get())
-      if res.isErr():
-        return RestApiResponse.jsonError(Http400,
-                                         ExecutionPayloadBidValidationError,
-                                         $res.error)
-      return RestApiResponse.jsonMsgResponse(
-        ExecutionPayloadBidValidationSuccess)
+    if consensusVersion.get() < ConsensusFork.Gloas:
+      return RestApiResponse.jsonError(Http400,
+                                       SlotFromTheIncorrectForkError)
 
-    case consensusVersion.get():
-      of ConsensusFork.Phase0 .. ConsensusFork.Fulu:
-        RestApiResponse.jsonError(Http400,
-                                  SlotFromTheIncorrectForkError)
-      of ConsensusFork.Gloas .. ConsensusFork.Heze:
-        decodeBidAndRoute(gloas.SignedExecutionPayloadBid)
+    let dres = decodeBodyJsonOrSsz(
+      gloas.SignedExecutionPayloadBid, contentBody.get())
+    if dres.isErr():
+      return RestApiResponse.jsonError(dres.error())
+    let res = await node.router.routeExecutionPayloadBid(dres.get())
+    if res.isErr():
+      return RestApiResponse.jsonError(Http400,
+                                       ExecutionPayloadBidValidationError,
+                                       $res.error)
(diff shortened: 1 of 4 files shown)


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1643,6 +1643,41 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+                                         $res.error)
+      return RestApiResponse.jsonMsgResponse(
+        ExecutionPayloadBidValidationSuccess)
+
+    case consensusVersion.get():
+      of ConsensusFork.Phase0 .. ConsensusFork.Fulu:
+        return RestApiResponse.jsonError(Http400,
+                                         SlotFromTheIncorrectForkError)
+      of ConsensusFork.Gloas .. ConsensusFork.Heze:
+        decodeBidAndRoute(gloas.SignedExecutionPayloadBid)
```
tersec: One shouldn't have to have this asymmetry of one branch of the `case` doing a `return` and the other at least notionally doing a `decodeBidAndRoute`. `decodeBidAndRoute` is itself a template which `returns`, so here right now both `return. But also, both could implicit-return.

beacon_chain/validators/message_router.nim
```diff
@@ -774,3 +774,25 @@ proc routeProposerPreferences*(
...
+
+  let res = await router[].network.broadcastExecutionPayloadBid(signedBid)
+  if res.isOk():
+    notice "Execution payload bid sent",
+      bid = shortLog(signedBid.message)
+  else:
+    notice "Execution payload bid not sent",
+      bid = shortLog(signedBid.message), error = res.error()
+
+  return ok()
```
tersec: Can use implicit return (just `ok()`)

beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1643,6 +1643,35 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
+
+    if contentBody.isNone():
+      return RestApiResponse.jsonError(Http400, EmptyRequestBodyError)
+
+    if consensusVersion.get() < ConsensusFork.Gloas:
+      return RestApiResponse.jsonError(Http400,
+                                       SlotFromTheIncorrectForkError)
+
+    let dres = decodeBodyJsonOrSsz(
+      gloas.SignedExecutionPayloadBid, contentBody.get())
```
tersec, line 1666: In general, how is this endpoint expected to handle that Heze [has a different `ExecutionPayloadBid` / `SignedExecutionPayloadBid`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/heze/beacon-chain.md#executionpayloadbid) due to FOCIL?

https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/publishExecutionPayloadBid doesn't seem to currently indicate either way, but presumably, it would be pointless to broadcast a Gloas bid in Heze.

The code should either explicitly fail and `debugHezeComment` this in Heze, or do something plausible (e.g., detect fork and use the correct payload bid type); this code will silently fail in Heze.

beacon_chain/rpc/rest_beacon_api.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8385/changes/9b047347e13a3f829e0a01caca5e01a49f3dbcfe

beacon_chain/validators/message_router.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8385/changes/9b047347e13a3f829e0a01caca5e01a49f3dbcfe

ahshum (commented):
LGTM, but would be good to put a `debugHezeComment` here.

beacon_chain/rpc/rest_beacon_api.nim
Tomi-3-0: Added a `debugHezeComment` and restricted the endpoint to just Gloas. Full Heze support will require downstream changes across multiple files to make them fork aware, which is probably better handled in a seperate PR

tersec (approved):

ahshum (approved):

PR #8388, opened 2026-05-02
fix payload signature verification in execution payload gossip validation

Tomi-3-0:


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -1112,19 +1112,25 @@ proc validateExecutionPayload*(
...
-        dag.genesis_validators_root,
-        envelope.slot.epoch,
-        signed_execution_payload_envelope.message,
-        builderKey,
-        signed_execution_payload_envelope.signature):
-      return dag.checkedReject("ExecutionPayload: invalid builder signature")
-  else:
-    return dag.checkedReject("ExecutionPayload: invalid fork")
+  let builderKey =
+    withState(dag.headState):
```
arnetheduck, line 1139: is the head state at gloas when the first block in the new fork arrives?

Tomi-3-0: For the envelope to reach this code, the block should already be in the dag when we retrieve it by [`getForkedBlock`](https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/gossip_processing/gossip_validation.nim#L1081).  and since the block is added via [`storeBlock`](https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/gossip_processing/block_processor.nim#L618) which calls [`addHeadBlockWithtParent`](https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/gossip_processing/block_processor.nim#L707) followed by [`updateHead`](https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/gossip_processing/block_processor.nim#L737)  which then calls [updateState](https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/consensus_object_pools/block_clearance.nim#L301) which should have advanced the head state by then

arnetheduck: hm .. it looks fragile since time passes between the head being updated on block arrival and here. not sure what the options are though.

ahshum: Agree with @arnetheduck, should we save builder keys instead of getting from headState?

https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/consensus_object_pools/block_clearance.nim#L111-L114

arnetheduck: hm, this ordering / guarantee used to be true when we had eth1deposits, but it might be the case that we should re-evaluate these guarantees based on the latest deposit handling which I don't quite remember how it goes cc @tersec

Tomi-3-0: > should we save builder keys instead of getting from headState?

the builder registry isn't append only unlike validators'. [`get_index_for_new_builder`](https://github.com/status-im/nimbus-eth2/blob/2f4c900311d2a5876822d77d0884ca4e606e8950/beacon_chain/spec/beaconstate.nim#L2249) assigns exited builders' indices to new builder registratns

Tomi-3-0: >hm .. it looks fragile since time passes between the head being updated on block arrival and here. not sure what the options are though.

the thing is imo, a gossip rejection isn't a fatal as is visible on the current devnet, the request manager fallback still goes ahead to fetch the envelope and it goes through the clearance path which doesn't have this bug. also, this rare case should only happen at the fork boundary.

PLus my understanding of this is that there is zero yield points between calling `addHeadBlockWithParent` and `updateHead`.  there should be some form of guarantee that the headState should already be at Gloas

arnetheduck: so I had a look at the rest of the function and there are a few issues:

* the seen check should not look at `missing` (if it's missing, we haven't seen it!) - it's also odd that it looks in the quarantine - if the block is in the quarantine but we can't get a blockref for it, we will ignore no matter what so the logic doesn't make sense (it's also interesting to consider the case where the block is in-flight in the async loop while receiving the payload)
* a payload may be received for _any_ block - this also means a block that is not always the head block - apart from the fork transition, it opens the question of risks related to non-head-block equivalence during a period of long non-finality - ie in a 50/50 split network, consider what happens when we receive a block from the "other" fork - again, I don't know the gloas guarantees surrounding this scenario, specially with the builder id but it's worth documenting how it works / why it's safe if it is safe
* we might receive the payload out-of-order - if we do, we should redistribute the payload once the block is received (in case they arrive out-of-order)

In any case, we can go ahead and merge this PR assuming we document the options (either with todo or answers)

tersec: > hm, this ordering / guarantee used to be true when we had eth1deposits, but it might be the case that we should re-evaluate these guarantees based on the latest deposit handling which I don't quite remember how it goes cc [@tersec](https://github.com/tersec)

It's based on finalized in EIP-6110. It doesn't get reordered.

PR #8392, opened 2026-05-04
ci: verify jobs on host with macos,xcode 26

siddarthkay:
## made to check how CI jobs perform on the host that has newer macos and xcode


siddarthkay:
verification was successful, PR no longer needed

PR #8393, opened 2026-05-04
catch the first column gossip addition event for getBlobs

agnxsh:
improves overall getBlobs efficiency even if the first column sidecar arrives before the designated block

### DIFF
--- a/beacon_chain/el/el_getblobs_service.nim
+++ b/beacon_chain/el/el_getblobs_service.nim
@@ -37,6 +37,7 @@ declareGauge beacon_engine_getblobs_slot_hit_rate,
 type
   GetBlobsService* = object
     blockGossipBus*: AsyncEventQueue[EventBeaconBlockGossipPeerObject]
+    columnSidecarBus*: AsyncEventQueue[DataColumnSidecarInfoObject]
     blockProcessor*: ref BlockProcessor
     dataColumnQuarantine*: ref ColumnQuarantine
     validatorCustody*: ValidatorCustodyRef
@@ -52,12 +53,14 @@ type
 proc new*(
     t: typedesc[GetBlobsServiceRef],
     blockGossipBus: AsyncEventQueue[EventBeaconBlockGossipPeerObject],
+    columnSidecarBus: AsyncEventQueue[DataColumnSidecarInfoObject],
     blockProcessor: ref BlockProcessor,
     dataColumnQuarantine: ref ColumnQuarantine,
     validatorCustody: ValidatorCustodyRef
 ): GetBlobsServiceRef =
   GetBlobsServiceRef(
     blockGossipBus: blockGossipBus,
+    columnSidecarBus: columnSidecarBus,
     blockProcessor: blockProcessor,
     dataColumnQuarantine: dataColumnQuarantine,
     validatorCustody: validatorCustody,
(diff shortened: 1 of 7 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -782,6 +782,14 @@ proc validateDataColumnSidecar*(
       slot: data_column_sidecar.signed_block_header.message.slot,
       kzg_commitments: data_column_sidecar.kzg_commitments)
 
+  # Fire the path-2 (column-first) hook with the full sidecar so the EL
```
tersec: path-2? Is there a designated path-1 and path-2 now?

I don't think this is how they're otherwise referred to.

agnxsh: I've added some clarifications, overall the idea is, we would need to extract headers/commitments from the event, hence the need for an out-of-spec callback which returns this data as well, apart from just `DataColumnInfoObject` which is in spec and doesn't provide all this.

tersec: https://github.com/status-im/nimbus-eth2/pull/8393/changes/4366aa7b919d41cd5625dfc8d64eb52f67c55281

beacon_chain/el/el_getblobs_service.nim
```diff
@@ -37,6 +38,7 @@ declareGauge beacon_engine_getblobs_slot_hit_rate,
 type
   GetBlobsService* = object
     blockGossipBus*: AsyncEventQueue[EventBeaconBlockGossipPeerObject]
+    columnSidecarBus*: AsyncEventQueue[ref fulu.DataColumnSidecar]
```
tersec: Maybe explicitly call this `fuluColumnSidecarBus` because Gloas would need something similar (does not have to be in this PR).

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -782,6 +782,14 @@ proc validateDataColumnSidecar*(
       slot: data_column_sidecar.signed_block_header.message.slot,
       kzg_commitments: data_column_sidecar.kzg_commitments)
 
+  # Notify with the full sidecar so the EL (out of spec)
```
tersec, line 785: How does "out of spec" apply here?

agnxsh: The gossip validation conditions don't ask one to notify full column sidecars they only expect to notify via data column info objects

beacon_chain/el/el_getblobs_service.nim
```diff
@@ -151,9 +175,75 @@ proc attemptGetBlobs*(
...
+    block_root = hash_tree_root(sidecar[].signed_block_header.message)
+    slot = sidecar[].signed_block_header.message.slot
+
+  # Dedup: only fire EL fetch once per block_root. Subsequent column
+  # arrivals for the same block are no-ops on this path.
+  if block_root in self.columnFirstFetched:
+    return
+
+  # If the sidecarless block is already in the block quarantine, the
+  # block-first path (consumeBlockGossip → attemptGetBlobs) owns this
```
tersec: `→`

agnxsh: i will get rid of all arrows no issue

PR #8401, opened 2026-05-05
fix seen block for envelope gossip

ahshum:
Seen block shouldn't include missing.

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -1030,7 +1030,6 @@ proc validateExecutionPayload*(
     block:
       var seen =
         envelope.beacon_block_root in quarantine.unviable or
-        envelope.beacon_block_root in quarantine.missing or
         dag.getBlockRef(envelope.beacon_block_root).isSome()
       if not seen:
         for k, _ in quarantine.orphans:


Tomi-3-0:
might be nice to also handle the in-flight case (case 3): https://github.com/status-im/nimbus-eth2/pull/8388#discussion_r3181955949 in this PR

ahshum:
> might be nice to also handle the in-flight case (case 3): [#8388 (comment)](https://github.com/status-im/nimbus-eth2/pull/8388#discussion_r3181955949) in this PR

There is no guarantees on using `headState` without state replay. I think it is in the scope of your PR.

Tomi-3-0:
> > might be nice to also handle the in-flight case (case 3): [#8388 (comment)](https://github.com/status-im/nimbus-eth2/pull/8388#discussion_r3181955949) in this PR
> 
> There is no guarantees on using `headState` without state replay. I think it is in the scope of your PR.

I meant this: 

> we might receive the payload out-of-order - if we do, we should redistribute the payload once the block is received (in case they arrive out-of-order)

this is likely a quarantine logic issue. when `getForkedBlock` fails but the block is known, the envelope can be quarantined instead of rejected.

ahshum:
> this is likely a quarantine logic issue. when `getForkedBlock` fails but the block is known, the envelope can be quarantined instead of rejected.

https://github.com/status-im/nimbus-eth2/blob/894378842fa94dd6696564b4614c8a4060f19901/beacon_chain/gossip_processing/gossip_validation.nim#L1040-L1043

Block processor will try processing envelope on the valid block is added.

PR #8406, opened 2026-05-05
chore: use latest nim-libp2p and try to use boringssl as a dependency separate from lsquic

richard-ramos:


tersec:
Has since occurred.

PR #8410, opened 2026-05-06
add partial data column quarantine for partial column gossip management, add tests for the same

agnxsh:


tests/all_tests.nim
```diff
@@ -46,6 +46,7 @@ import # Unit test
   ./test_payload_attestation_pool,
   ./test_peer_pool,
   ./test_peerdas_helpers,
+  ./test_partial_column_quarantine,
```
tersec: Lexicographic order

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,993 @@
...
+  copyMem(addr res.bytes[0], unsafeAddr tmp[0], sizeof(uint64))
+  res
+
+func genKzgCell(index: int): KzgCell =
+  var res: KzgCell
+  let tmp = uint64(index).toBytesLE()
+  copyMem(addr res.bytes[0], unsafeAddr tmp[0], sizeof(uint64))
+  res
+
+func genKzgProof(index: int): KzgProof =
```
tersec: `genBlockRoot`, `genKzgCommitment`, `genKzgCell`, and `genKzgProof` are all basically the same function. Can have one generic functioning parameterized by return type to cover all of them.

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
```diff
@@ -0,0 +1,286 @@
...
+func getOrCreateEntry*(
+    quarantine: var PartialColumnQuarantine,
+    blockRoot: Eth2Digest,
+    columnIndex: ColumnIndex,
+    numBlobs: int): PartialColumnEntry =
+  ## Get or create a PartialColumnEntry for the given (block_root, column_index).
+  let key = PartialColumnKey(blockRoot: blockRoot, columnIndex: columnIndex)
+  let existing = quarantine.entries.get(key)
+  if existing.isSome():
+    return existing.get()
```
tersec: Can use `isErrOr` here to avoid the `get` (which, yes, is largely safe, by inspection here, but one goal I have is to reduce the overall amount of them, and use the `nim-results` abstractions as much as feasible, to concentrate/centralize where random `Results`-based `Defect`s might need to be investigated). It will work quite neatly in this case.

agnxsh: Ok putting

tests/test_partial_column_quarantine.nim
```diff
@@ -0,0 +1,976 @@
...
+  unittest2,
+  kzg4844/kzg_abi,
+  ssz_serialization/[types as sszTypes, bitseqs],
+  ../beacon_chain/spec/datatypes/[fulu, deneb],
+  ../beacon_chain/spec/presets,
+  ../beacon_chain/consensus_object_pools/partial_column_quarantine
+
+func gen[T](index: int): T =
+  let tmp = uint64(index).toBytesLE()
+  when T is Eth2Digest:
```
tersec: Oops, missed that `Eth2Digest` was slightly different. Better to keep that split out into its own function.

tersec: https://github.com/status-im/nimbus-eth2/pull/8410/changes/17e1926d817a2d4834a9d91683e1ba79390e4ee3

beacon_chain/consensus_object_pools/partial_column_quarantine.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8410/changes/17e1926d817a2d4834a9d91683e1ba79390e4ee3

PR #8415, opened 2026-05-07
fix: use correct condition for withdrawal computation

Tomi-3-0:
When a parent block had a bid but its payload was never delivered, the old check computed fresh withdrawals instead of using                                                               the cached `payload_expected_withdrawals` causing "withdrawals mismatch"  when preparing payload attributes to send to the EL which later fails when `verify_execution_payload_envelope` is called on the built payload

### DIFF
--- a/beacon_chain/validators/block_payloads.nim
+++ b/beacon_chain/validators/block_payloads.nim
@@ -380,10 +380,11 @@ proc getExecutionPayload*(
         # - If `should_extend_payload(store, parent_root)`:
         #     `withdrawals = get_expected_withdrawals(state).withdrawals`.
         # - else `withdrawals = state.payload_expected_withdrawals`.
-        if forkyState.data.latest_execution_payload_bid.block_hash.isZero():
-          forkyState.data.payload_expected_withdrawals.asSeq()
-        else:
+        if forkyState.data.latest_block_hash ==
+            forkyState.data.latest_execution_payload_bid.block_hash:
           get_expected_withdrawals(forkyState.data).withdrawals
+        else:
+          forkyState.data.payload_expected_withdrawals.asSeq()
       elif consensusFork >= ConsensusFork.Capella:
         get_expected_withdrawals(forkyState.data)
       else:


beacon_chain/validators/block_payloads.nim
```diff
@@ -380,10 +380,11 @@ proc getExecutionPayload*(
         # - If `should_extend_payload(store, parent_root)`:
         #     `withdrawals = get_expected_withdrawals(state).withdrawals`.
         # - else `withdrawals = state.payload_expected_withdrawals`.
-        if forkyState.data.latest_execution_payload_bid.block_hash.isZero():
-          forkyState.data.payload_expected_withdrawals.asSeq()
-        else:
+        if forkyState.data.latest_block_hash ==
+            forkyState.data.latest_execution_payload_bid.block_hash:
           get_expected_withdrawals(forkyState.data).withdrawals
+        else:
```
ahshum, line 386: might be good to check with `proposalExecutionHead`. I think isZero is for genesis but not sure have it fixed yet.

Tomi-3-0: `proposalExecutionHead` handles which EL parent hash to pass to the engine, which is a different issue from the withdrawals branch selection.

tersec: Agree, this is a pure CL issue.

ahshum: Good to know.

ahshum (approved):

beacon_chain/validators/block_payloads.nim
tersec: Is it worth ensuring this works well at genesis, both the current EF genesis state/block generators and the spec. According to the spec:
- https://github.com/ethereum/consensus-specs/pull/5067
- https://github.com/ethereum/consensus-specs/issues/5149
- https://github.com/ethereum/consensus-specs/issues/5168
- https://github.com/ethereum/consensus-specs/pull/5172

Neither is in https://github.com/ethereum/consensus-specs/commits/v1.7.0-alpha.7/ but presumably both will be in alpha.8. As of the next spec release, at genesis:
```
state.latest_block_hash = blockhash
state.latest_execution_payload_bid.block_hash = 0x00
state.latest_execution_payload_bid.parent_block_hash = blockhash
```

i.e. `forkyState.data.latest_block_hash == forkyState.data.latest_execution_payload_bid.block_hash` is meant to stand in for [`should_extend_payload(store, parent_root)`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/gloas/fork-choice.md#new-should_extend_payload). At gloas genesis, this should be always false due to [`is_payload_verified(...)`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/gloas/fork-choice.md#new-is_payload_verified) being false.

In an ongoing chain, this captures that, and it also matches at genesis. So it matches the desired behavior in alpha.8 spec. It should also match the current genesis generator, where they also don't match at genesis.

ahshum: Right, so for genesis we may still need the isZero check?

tersec: As long as `forkyState.data.latest_block_hash != forkyState.data.latest_execution_payload_bid.block_hash` captures this case, it doesn't have to be separately checked. It would be good to add a comment in a subsequent PR to note this reasoning though. In alpha.8, this inequality holds, so it's fine, and from what I last saw in the EF's genesis generator, it also holds.

PR #8417, opened 2026-05-07
Fix ColumnQuarantine to start  use correct amount of supernode columns.

cheatfate:
Changes:

1. Its enough to obtain only `NUMBER_OF_COLUMNS div 2` columns for `hasSidecars`  returning `true` and `popSidecars` returning you list of columns (all columns will be retrieved, not only `NUMBER_OF_COLUMNS div 2`).
2. If remote peer is `supernode` - `fetchMissingSidecars()` returns `NUMBER_OF_COLUMNS` columns request.

### DIFF
--- a/beacon_chain/consensus_object_pools/column_quarantine.nim
+++ b/beacon_chain/consensus_object_pools/column_quarantine.nim
@@ -102,6 +102,13 @@ func maxSidecars(maxSidecarsPerBlock: uint64): int =
   # blobs may arrive before an orphan is tagged `blobless`
   3 * int(SLOTS_PER_EPOCH) * int(maxSidecarsPerBlock)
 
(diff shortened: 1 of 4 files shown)


beacon_chain/consensus_object_pools/column_quarantine.nim
```diff
@@ -102,6 +102,13 @@ func maxSidecars(maxSidecarsPerBlock: uint64): int =
   # blobs may arrive before an orphan is tagged `blobless`
   3 * int(SLOTS_PER_EPOCH) * int(maxSidecarsPerBlock)
 
+func enoughColumns*[A, B](q: SidecarQuarantine[A, B], count: int): bool =
```
agnxsh, line 105: we don't really have to export this, it's all within column quarantine

tersec: - #8689

PR #8422, opened 2026-05-08
fix: gloas block gossip validation

ahshum:
- On checking parent envelope, we should also ensure that the parent is a Gloas or later block.

### DIFF
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -462,11 +462,11 @@ template validateBeaconBlockGloas(
   #
   # - [REJECT] The block's execution payload parent (defined by
   #   bid.parent_block_hash) passes all validation.
-  if dag.cfg.consensusForkAtEpoch(executionParent.slot.epoch()) >=
+  if executionParent.root in envelopeQuarantine.unviable:
+    return dag.checkedReject("validateBeaconBlockGloas: invalid parent payload")
+  elif dag.cfg.consensusForkAtEpoch(executionParent.slot.epoch()) >=
       ConsensusFork.Gloas:
-    if executionParent.root in envelopeQuarantine.unviable:
-      return dag.checkedReject("validateBeaconBlockGloas: invalid parent payload")
-    elif not (
+    if not (
         dag.db.containsExecutionPayloadEnvelope(executionParent.root) or
         executionParent.root in envelopeQuarantine.orphans
     ):
@@ -474,10 +474,8 @@ template validateBeaconBlockGloas(
       discard quarantine[].addOrphan(dag.finalizedHead.slot, signed_beacon_block)
       return errIgnore("validateBeaconBlockGloas: parent payload not yet seen")
   else:
-    # We don't have a reliable source of truth of execution validity for
-    # pre-Gloas blocks. Probably we could check with EL.
-    #
-    # TODO: execution validity of blocks
+    # The execution parent is a pre-Gloas block. It has been validated and
+    # imported to DAG so we could assume it is execution valid.
     discard
 
   # [REJECT] The bid's parent (defined by `bid.parent_block_root`) equals the
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -474,7 +474,7 @@ template validateBeaconBlockGloas(
       discard quarantine[].addOrphan(dag.finalizedHead.slot, signed_beacon_block)
       return errIgnore("validateBeaconBlockGloas: parent payload not yet seen")
   else:
-    # We don't have a reliable source of truth of the execution validity for
+    # We don't have a reliable source of truth of execution validity for
     # pre-Gloas blocks. Probably we could check with EL.
     #
     # TODO: execution validity of blocks
--- a/beacon_chain/gossip_processing/gossip_validation.nim
+++ b/beacon_chain/gossip_processing/gossip_validation.nim
@@ -423,34 +423,26 @@ template validateBeaconBlockGloas(
   template bid: untyped = blck.body.signed_execution_payload_bid.message
 
   let executionParent = block:
-    let
-      parent = dag.getBlockRef(bid.parent_block_root).valueOr:
+    var
+      cur = dag.getBlockRef(bid.parent_block_root).valueOr:
         return errIgnore("validateBeaconBlockGloas: parent not yet seen")
-      pBhash = dag.loadExecutionBlockHash(parent).valueOr:
+      i = 0
+      found = false
+
+    # Search execution parent up to 2 ancestors. Also stop searching if there is
+    # not any parents, that could be a finalized block or genesis.
+    while not isNil(cur.parent) and i < 2:
+      let pBhash = dag.loadExecutionBlockHash(cur).valueOr:
         return errIgnore("validateBeaconBlockGloas: cannot load block hash")
-    if pBhash == bid.parent_block_hash:
-      # Parent's payload status is FULL
-      parent.bid
-    else:
-      # Parent's payload status is EMPTY, i.e. check with grandparent.
-      let res = block:
-        if not isNil(parent.parent):
-          # Grandparent is non-finalized yet.
-          let gpBhash = dag.loadExecutionBlockHash(parent.parent).valueOr:
-            return errIgnore("validateBeaconBlockGloas: cannot load block hash")
-          (parent.parent.bid, gpBhash)
-        else:
-          # Grandparent should be either finalized or nonexistent.
-          let
-            gpBid = dag.parent(parent.bid).valueOr:
-              return errIgnore("validateBeaconBlockGloas: invalid execution parent")
-            gpBhash = dag.loadExecutionBlockHash(gpBid).valueOr:
-              return errIgnore("validateBeaconBlockGloas: cannot load block hash")
-          (gpBid, gpBhash)
-      if res[1] == bid.parent_block_hash:
-        res[0]
+      if pBhash == bid.parent_block_hash:
+        found = true
+        break
       else:
-        return errIgnore("validateBeaconBlockGloas: invalid execution parent")
+        cur = cur.parent
+      inc i
+    if not found:
+      return errIgnore("validateBeaconBlockGloas: invalid execution parent")
+    cur
 
   # - [IGNORE] The block's parent execution payload (defined by
   #   bid.parent_block_hash) has been seen (via gossip or non-gossip sources)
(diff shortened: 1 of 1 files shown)


beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -462,15 +462,23 @@ template validateBeaconBlockGloas(
...
+    elif not (
+        dag.db.containsExecutionPayloadEnvelope(executionParent.root) or
+        executionParent.root in envelopeQuarantine.orphans
+    ):
+      envelopeQuarantine[].addMissing(executionParent.root)
+      discard quarantine[].addOrphan(dag.finalizedHead.slot, signed_beacon_block)
+      return errIgnore("validateBeaconBlockGloas: parent payload not yet seen")
+  else:
+    # We don't have a reliable source of truth of execution validity for
+    # pre-Gloas blocks. Probably we could check with EL.
```
tersec: The gossip validation should not check with EL. It needs to be able to respond synchronously and quickly.

tersec:
> * On checking parent envelope, we should also ensure that the parent is a Gloas or later block.

If there is a parent envelope at all, that implies the parent block is Gloas or later. Or?

beacon_chain/gossip_processing/gossip_validation.nim
ahshum: The information of execution validity is not persistent even in memory, as we could get finalized blocks here. Probably we would need a new store if not checking with EL?

ahshum:
> If there is a parent envelope at all, that implies the parent block is Gloas or later. Or?

By getting `executionParent` (code above), there is no checks around slot/fork. We could be getting Fulu block here but falsely check if an envelope exists, which should always be false.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_dag.nim#L24-L46 already has this in `optimisticStatus`:
https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_dag.nim#L17-L22

ahshum: Yes I think `optimisticStatus` would be helpful if we assume execution parent must be non-finalized block. Currently `typeof(executionParent) == BlockId`.

Anyway, I updated the validation for pre-Gloas blocks hope that works simpler now.

ahshum: The problem of `optimisticStatus` is that it is in-memory only and only set when we receive a new block that is built on it and is valid. In the ideal path it works perfectly. However, if the data lose for some reasons, e.g. a node restart, we have to trust that it is execution valid if it is `notValidated`.

tersec: > The problem of `optimisticStatus` is that it is in-memory only and only set when we receive a new block that is built on it and is valid. In the ideal path it works perfectly. However, if the data lose for some reasons, e.g. a node restart, we have to trust that it is execution valid if it is `notValidated`.

Yes, immediately on startup blocks get marked at optimistic, regardless of how the EL decided previously. But https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_dag.nim#L250-L263 kicks in the first time the EL responds on the canonical chain and marks all the relevant `BlockRef`s nonoptimistic/`VALID` again.

> Yes I think `optimisticStatus` would be helpful if we assume execution parent must be non-finalized block.

The current finalized block is still in the `blockchain_dag`. It's its parents/ancestors which are pruned: https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/blockchain_dag.nim#L2644-L2674

And `dag.finalizedHead` is exactly this finalization point, in the DAG, in `BlockRef` form: https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_pools_types.nim#L105-L115 and https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_pools_types.nim#L182-L184

tersec:
> > If there is a parent envelope at all, that implies the parent block is Gloas or later. Or?
> 
> By getting `executionParent` (code above), there is no checks around slot/fork. We could be getting Fulu block here but falsely check if an envelope exists, which should always be false.

If
> The executionParent exists in DAG,

then https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_dag.nim#L24-L38 means one can check the `bid` field of the `BlockRef` for the `slot` and whether it's before or after Gloas.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -462,15 +462,19 @@ template validateBeaconBlockGloas(
...
-  elif not (
-      dag.db.containsExecutionPayloadEnvelope(executionParent.root) or
-      executionParent.root in envelopeQuarantine.orphans
-  ):
-    envelopeQuarantine[].addMissing(executionParent.root)
-    discard quarantine[].addOrphan(dag.finalizedHead.slot, signed_beacon_block)
-    return errIgnore("validateBeaconBlockGloas: parent payload not yet seen")
+  if dag.cfg.consensusForkAtEpoch(executionParent.slot.epoch()) >= ConsensusFork.Gloas:
+    # The executionParent exists in DAG, so we should check unviable envelope
+    # and the database for the validation rules.
```
ahshum, line 459: @tersec Here is the check of slot/fork. 

The idea is that `executionParent` is retrieved from `getBlockRef` or database so it is a valid block (i.e. must not be unviable/orphaned/missing). So it should be execution valid for pre-Gloas blocks while need additional checks over envelope for Gloas blocks.

The empty `else` for pre-Gloas forks is for comment only.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -462,15 +462,19 @@ template validateBeaconBlockGloas(
...
+    if executionParent.root in envelopeQuarantine.unviable:
+      return dag.checkedReject("validateBeaconBlockGloas: invalid parent payload")
+    elif not dag.db.containsExecutionPayloadEnvelope(executionParent.root):
+      envelopeQuarantine[].addMissing(executionParent.root)
+      discard quarantine[].addOrphan(dag.finalizedHead.slot, signed_beacon_block)
+      return errIgnore("validateBeaconBlockGloas: parent payload not yet seen")
+  else:
+    # The execution parent is a pre-Gloas block. It has been validated and
+    # imported to DAG so we could assume that it is execution valid.
```
tersec: This isn't necessarily the case. The DAG contains execution-optimistic blocks too: https://github.com/status-im/nimbus-eth2/blob/f44c70f869f971313cc6a28c3d167d5a0bf7b418/beacon_chain/consensus_object_pools/block_dag.nim#L247-L248

How that interacts with Gloas block validation is a different question, but this assumption does not hold.

beacon_chain/gossip_processing/gossip_validation.nim
```diff
@@ -462,15 +454,27 @@ template validateBeaconBlockGloas(
...
+      return dag.checkedReject("validateBeaconBlockGloas: unviable execution parent")
+    # The genesis block would not have an envelope. Otherwise, we should have
+    # the envelope for the execution parent.
+    elif not (executionParent.slot != GENESIS_SLOT and
+        dag.db.containsExecutionPayloadEnvelope(executionParent.root)):
+      envelopeQuarantine[].addMissing(executionParent.root)
+      discard quarantine[].addOrphan(dag.finalizedHead.slot, signed_beacon_block)
+      return errIgnore("validateBeaconBlockGloas: parent payload not yet seen")
+  else:
+    # The executionParent is found from DAG, which is a validated Fulu block. We
```
tersec: It's a validated pre-Gloas block. It's very likely a Fulu block but it it might be an earlier fork too.

tersec: Also, regarding checking for whether it's execution-enabled, while Nimbus kind of supports (mainly for e.g., `block_sim` and some `tests/` purposes) running non-merged all the way through Heze, as https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/capella/beacon-chain.md#modified-process_execution_payload indicates the spec itself does not:
> The function `process_execution_payload` is modified to use the new `ExecutionPayloadHeader` type and removed the `is_merge_transition_complete` check.

That is, according to spec, if this block is from Capella or later, if the parent is in the DAG at all, it's guaranteed to be `is_execution_enabled.

`getForkedBlock` pulls in an entire block (could be up to 10MB) from the database and decodes it, which be somewhat expensive, so it's nice to avoid when feasible. But with a `BlockId.slot` can provide this information without ever accessing the database at all, with only what's already in memory and a single comparison, so it can be readily avoided.

But even that check isn't really necessary, because https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.7/specs/gloas/p2p-interface.md#beacon_block documents that:
> And instead the following validations are set in place with the alias `bid = block.body.signed_execution_payload_bid.message`:
>
> ...
>
> - If `execution_payload` verification of block's execution payload parent by an execution node **is complete**:
>   - [REJECT] The block's execution payload parent (defined by `bid.parent_block_hash`) passes all validation.

So pre-execution (phase0 through part of Bellatrix) blocks vacuously satisfy this. They do pass all the validation they need to pass, because there is no verification of block's execution payload parent by an  execution node to complete.

So if it's in the DAG and it's pre-Gloas it's either:
- premerge (see above, no REJECT)
- optimistic (not verification not completed, nothing to do, no REJECT)
- valid (no REJECT)

It should not ever be invalid, because those are addressed elsewhere. So this condition should be automatically satisfied by any pre-Gloas block in the DAG, without the need to check `getForkedBlock` and incur a database read/snappy/SSZ decode.

ahshum: Regards to pre-merge, would we have pre-merge Gloas in theory?

tersec: Nimbus specifically can have pre-merge Gloas in the state transition function. This is how `block_sim` runs. However, the official EF consensus specs say that no, pre-merge Gloas is not possible in theory, and none of the networks Nimbus support involve such as pre-merge Electra/Fulu/Gloas network.

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8422/changes/0dd757247345c8cebc461521d8505a3b2dea1041

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8422/changes/751242ad9df5a9b1a6fcd1d40fdc7cbec739dd43

beacon_chain/gossip_processing/gossip_validation.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8422/commits/5c9f64b7b7cd69e9efceb460264ebb6aca89600d

PR #8424, opened 2026-05-08
nix: add NixOS BN and VC service definitions

jakubgs:
This adds NixOS service definitions that have been used for more than a year on our production fleets for ease of re-use in multiple fleet repos.

### DIFF
--- a/ci/nix.Jenkinsfile
+++ b/ci/nix.Jenkinsfile
@@ -55,6 +55,12 @@ pipeline {
         sh 'result/bin/nimbus_beacon_node --version'
       } }
     }
+
+    stage('Service check') {
+      steps { script {
+        sh 'nix run ".#checks.x86_64-linux.beacon-node.driver"'
+      } }
+    }
   }
 
   post {
--- a/flake.nix
+++ b/flake.nix
@@ -56,5 +56,18 @@
       devShells = forAllSystems (system: {
         default = pkgsFor.${system}.callPackage ./nix/shell.nix { };
       });
+
+      nixosModules = rec {
+        beacon-node = import ./nix/services/beacon-node.nix { inherit (self) packages; };
+        validator-client = import ./nix/services/validator-client.nix { inherit (self) packages; };
+        default = { imports = [ beacon-node validator-client ]; };
+      };
+
+      checks = forAllSystems (system: let
+        inherit (nixpkgs.legacyPackages.${system}) callPackage;
+      in {
+        beacon-node = callPackage ./nix/services/checks/beacon-node.nix { inherit self; };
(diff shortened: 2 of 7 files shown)


nix/README.md
```diff
@@ -28,3 +28,27 @@ This is [a known issue with `github:` schema](https://github.com/NixOS/nix/issue
 ```sh
 nix run 'git+https://github.com/status-im/nimbus-eth2'
 ```
+
+## Flakes
+
+You can use our packages and NixOS service definitions by adding `nimbus-eth2` as a flake input:
+```nix
+inputs.nimbus-eth2.url = "git+https://github.com/status-im/nimbus-eth2?submodules=1#";
```
tersec: This would seem to create a novel dependency on the specific structure of the `nimbus-eth2` repository as an external service?

apentori (approved):

nix/README.md
jakubgs: Well, yes, or to be more exact, on specific structure of the Nix flake. User can include it via `inputs` and use it as part of their NixOS config, but it's not as strict as you might think. Pretty much any part of the service definition shown here can be overridden by the user.

For example, lets say I don't like the `Restart = "on-failure";` setting. I can either just copy the service definition to my own repo and edit it, or simply override that setting in your config:
```nix
  systemd.services.nimbus-validator-client.serviceConfig.Restart = lib.mkForce "always";
```
I'd say this is not much different than someone fetching [`nimbus_beacon_node.service`](https://github.com/status-im/nimbus-eth2/blob/stable/scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service) from this repo as part of their system setup. But actually way more customizable.

tersec: > I'd say this is not much different than someone fetching [`nimbus_beacon_node.service`](https://github.com/status-im/nimbus-eth2/blob/stable/scripts/package_src/nimbus_beacon_node/image/lib/systemd/system/nimbus_beacon_node.service) from this repo as part of their system setup. But actually way more customizable.

This only offers the self-contained version once downloaded, no further dependency on the upstream repo. In that crucial way, it does differ and this is novel. It might be worth supporting, but right now, there's no equivalent in the non-Nix part of the repo; the supported aspects of the repro are solely the set of binary artifacts, nimbus.guide pages, etc it triggers the creation of.

jakubgs: That's false already, since people use the Nix flake to build the package as part of Flake inputs already. It's the same thing. So no, this is not novel in any way.

jakubgs: I can remove readme entry if you insist.

jakubgs:
I have tested this setup in both our Hoodi EthCTV cluster and my personal setup and it works without issues.

PR #8429, opened 2026-05-11
nix: introduce our nix cache to flake and ci

markoburcul:
## Changes
- adapt the nix CI so it builds VC as well
- push the build artefacts to our nix cache
- add the usage of nix cache in flake.nix settings - speed up the builds

## Note
Before merging this I need to merge the:
* https://github.com/status-im/status-jenkins-lib/pull/138

### DIFF
--- a/ci/nix.Jenkinsfile
+++ b/ci/nix.Jenkinsfile
@@ -6,7 +6,9 @@
  *   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
  * at your option. This file may not be copied, modified, or distributed except according to those terms.
  */
-library 'status-jenkins-lib@v1.9.41'
+library 'status-jenkins-lib@v1.9.44'
+
+def result = ''
 
 pipeline {
   agent {
@@ -23,6 +25,11 @@ pipeline {
       name: 'VERBOSITY',
       description: 'Value for the V make flag to increase log verbosity',
       choices: [0, 1, 2]
+    )
+    choice(
+      name: 'NIX_TARGET',
+      description: 'Nix flake target to build',
+      choices: ['beacon_node', 'validator_client']
     )
   }
 
@@ -44,15 +51,32 @@ pipeline {
   }
 
   stages {
-    stage('Beacon Node') {
+    stage('Deps') {
       steps { script {
-        nix.flake('beacon_node')
+        sh 'git submodule update --init --recursive'
+      } }
+    }
+
+    stage('Build') {
+      steps { script {
+        result = nix.flake(params.NIX_TARGET)
       } }
     }
 
     stage('Version check') {
       steps { script {
-        sh 'result/bin/nimbus_beacon_node --version'
+        sh "${result}/bin/nimbus_${params.NIX_TARGET} --version"
+      } }
+    }
+
+    stage('Push to Nix cache') {
+      when {
+        expression {
+          env.JOB_NAME.toLowerCase().contains('nightly')
+        }
+      }
+      steps { script {
+        nix.copyToCache(derivations: [result])
       } }
     }
   }
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,12 @@
 {
   description = "nimbus-eth2";
 
+  nixConfig = {
+    extra-substituters = [ "https://nix-cache.status.im/" ];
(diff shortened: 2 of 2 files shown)


ci/nix.Jenkinsfile
```diff
@@ -44,17 +47,43 @@ pipeline {
...
+    stage('Deps') {
+      steps { script {
+        sh 'git submodule update --init --recursive'
+      } }
+    }
+
     stage('Beacon Node') {
       steps { script {
-        nix.flake('beacon_node')
+        bn = nix.flake('beacon_node')
```
jakubgs: Eh, why not parametrize this Jenkinsfile with `MAKE_TARGET` and have two jobs with different targets?

markoburcul: We could I guess

jakubgs (approved):

ci/nix.Jenkinsfile
jakubgs: That way we can more easily add more targets, since this repo also contains portal client and light client too.

tersec:
Well, looks plausible, but hard to say unless it can actually pass CI.

markoburcul:
> Well, looks plausible, but hard to say unless it can actually pass CI.

I've made it now to skip push to our Nix cache when it is a PR build. We are gonna have that step run in the nightly pipeline which is gonna run on the release Jenkins instance, where we would build all the relevant branches and push to the Nix cache (testing, unstable, stable,libp2p) both BN and VC,

One question though: do we want PR builds for both BN and VC? This would add one more check here in GitHub PR.

tersec:
> I've made it now to skip push to our Nix cache when it is a PR build.

Doesn't this just push the issue to the non-PR builds, which will still fail? That is, if this PR is made to pass CI by skipping the tests which currently fails, but that test is not skipped once it runs in `unstable`, that's an even worse outcome.

> One question though: do we want PR builds for both BN and VC? This would add one more check here in GitHub PR.

In general PRs should test what `unstable` or other parts of CI test, yeah. The point isn't really to have it be in PR or not for me, it's to have it be consistent. Right now, today, it is: there's no Nix CI in either the PRs or non-PRs in `nimbus-eth2`.

With this PR as initially posted, it would have presumably been similarly consistent: there would have been such a CI test. But it was failing. I'm not going to merge a known-failing-CI PR modulo exigent circumstances.

But hiding that and having it still fail but only fail after merge (and requiring immediate reverting as a result) would be the worst outcome.

markoburcul:
> Doesn't this just push the issue to the non-PR builds, which will still fail? That is, if this PR is made to pass CI by skipping the tests which currently fails, but that test is not skipped once it runs in `unstable`, that's an even worse outcome.

The step that failed was to push nix artefacts to our Nix cache, which has been tested in the release CI as well and the issue was mitigated. Here we don't want to push to the cache because we don't wanna keep the credentials to access our nix cache in the development CI. Why would this push the issue to the non-PR builds?

jakubgs:
>Doesn't this just push the issue to the non-PR builds, which will still fail? That is, if this PR is made to pass CI by skipping the tests which currently fails, but that test is not skipped once it runs in unstable, that's an even worse outcome.

I think you're confusing things here @tersec . The stage we are skipping in PR builds is the stage that pushes the built packages and dependencies to our binary cache. It's not testing anything, it's just caching things.

The primary reason why PR builds cannot push to our caches is because I prefer to keep credentials that can do that in release CI where nightly builds can run for security reasons. It is simply safer to compartmentalize what CI has what permissions. In case of non-release CI being compromised due to a malicious PR the attacker can't extract credentials for pushing to our cache since they are not available there in the first place. Hope that explains it.

ci/nix.Jenkinsfile
```diff
@@ -44,15 +51,32 @@ pipeline {
...
+    stage('Build') {
+      steps { script {
+        result = nix.flake(params.NIX_TARGET)
       } }
     }
 
     stage('Version check') {
       steps { script {
-        sh 'result/bin/nimbus_beacon_node --version'
+        sh "result/bin/nimbus_${params.NIX_TARGET} --version"
```
jakubgs: Why not use `${result}/bin/...`?

jakubgs (approved):

markoburcul:
@tersec do we have a verdict on this change? Please look at our comments above.

PR #8430, opened 2026-05-11
add fork choice types for gloas

Tomi-3-0:
Context:
Implements EIP-7732 fork choice changes. Beacon blocks no longer contain execution payloads directly; a proposer commits to a builder's bid, the builder (self-built or external) delivers the payload seperately, and a Payload Timeliness Committee  (PTC) votes on whether it arrived timely.

In this implementation the forkchoice now handles blocks in two states (with and without a verified payload). Using a dual-node architecture where each block can expand into `EMPTY` and `FULL` variants.

Proposed implementation will follow the following steps:
  1. Fork choice types (this PR)
    - New fields on ProtoNode, VoteTracker, QueuedAttestation, ForkChoiceBackend
    - New PtcVotes, ForkChoiceNode types; ProtoArray object gains fullBlockIndices
  2. Dual-node proto_array
    - FULL virtual nodes as siblings to EMPTY, parent routing, head payload status, sibling resolution, prune cleanup. ~2x nodes worst case (assuming no empty blocks), bounded by pruning; no special-casing in tree traversal
  3. Payload-aware vote delta computation
    - Route vote weight to FULL or EMPTY node based on payload preference with one extra table lookup per vote change.
  4. PTC vote recording, block timeliness tracking with indexed equivocation lookup, payload envelope verification
  5. Attestation pipeline
    - Slot-based vote tracking, full-payload vote filtering, committee index as
  payload preference signal. The epoch-level dedup is replaced with slot-level dedup.
  6. Proposer boost gating
    - Conditional boost based on weak-head detection and equivocation checks with constant-time per head computation with no tree walks
  7. process_block integration
    - Parent payload status computation, batch PTC processing from in-block payload attestations.

### DIFF
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -173,7 +173,7 @@ type
     balances*: seq[ForkChoiceBalance]
     execution_payload_states*: HashSet[Eth2Digest]
     ptc_vote*: Table[Eth2Digest, PtcVotes]
-    ptc_data_availability*: Table[Eth2Digest, PtcVotes]
+    ptc_data_availability_vote*: Table[Eth2Digest, PtcVotes]
     block_timeliness*: Table[Eth2Digest, array[2, bool]]
 
   QueuedAttestation* = object
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -184,10 +184,6 @@ type
     checkpoints*: Checkpoints
     queuedAttestations*: seq[QueuedAttestation]
 
-  ForkChoiceNode* = object
-    root*: Eth2Digest
-    payloadStatus*: PayloadStatus
-
 func shortLog*(vote: VoteTracker): auto =
   (
     slot: vote.slot,
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -146,14 +146,13 @@ type
     current_root*: Eth2Digest
     next_root*: Eth2Digest
     slot*: Slot
-    next_epoch*: Epoch
-    next_slot*: Slot
     payload_present*: bool
     next_payload_present*: bool
 
   PtcVotes* = object
     voted*: BitArray[int(PTC_SIZE)]
-    value*: BitArray[int(PTC_SIZE)]
+    payload_present*: BitArray[int(PTC_SIZE)]
+    data_available*: BitArray[int(PTC_SIZE)]
 
   BalanceSource* = object
     # Effective balances / slashings in `info` based on historical checkpoint.
@@ -175,7 +174,6 @@ type
     balances*: seq[ForkChoiceBalance]
     execution_payload_states*: HashSet[Eth2Digest]
     ptc_vote*: Table[Eth2Digest, PtcVotes]
-    ptc_data_availability_vote*: Table[Eth2Digest, PtcVotes]
     block_timeliness*: Table[Eth2Digest, array[2, bool]]
 
   QueuedAttestation* = object
@@ -198,8 +196,6 @@ func shortLog*(vote: VoteTracker): auto =
     slot: vote.slot,
     current_root: shortLog(vote.current_root),
     next_root: shortLog(vote.next_root),
-    next_epoch: vote.next_epoch,
-    next_slot: vote.next_slot,
     payload_present: vote.payload_present,
     next_payload_present: vote.next_payload_present
   )
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -147,11 +147,6 @@ type
     payload_present*: bool
     next_payload_present*: bool
 
-  PtcVotes* = object
-    voted*: BitArray[int(PTC_SIZE)]
-    payload_present*: BitArray[int(PTC_SIZE)]
-    data_available*: BitArray[int(PTC_SIZE)]
-
   BalanceSource* = object
     # Effective balances / slashings in `info` based on historical checkpoint.
     # The `assigned_slots` (`fast_confirmation.nim`) are based on `dag.head`
@@ -170,7 +165,6 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
-    ptc_vote*: Table[Eth2Digest, PtcVotes]
     block_timeliness*: Table[Eth2Digest, array[2, bool]]
 
   QueuedAttestation* = object
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -9,7 +9,7 @@
 
 import
   # Standard library
-  std/tables,
+  std/[sets, tables],
   # Status
   results,
   chronicles,
@@ -121,6 +121,9 @@ type
     invalid*: bool
     bestChild*: Opt[Index]
     bestDescendant*: Opt[Index]
+    parentPayloadStatus*: PayloadStatus
+    bidBlockHash*: Eth2Digest
+    proposerIndex*: uint64
 
   BalanceCheckpoint* = object
     checkpoint*: Checkpoint
@@ -142,6 +145,13 @@ type
     current_root*: Eth2Digest
     next_root*: Eth2Digest
     slot*: Slot
+    next_epoch*: Epoch
+    next_slot*: Slot
+    payload_present*: bool
+
+  PtcVotes* = object
+    voted*: BitArray[int(PTC_SIZE)]
+    value*: BitArray[int(PTC_SIZE)]
 
   BalanceSource* = object
     # Effective balances / slashings in `info` based on historical checkpoint.
@@ -161,22 +171,34 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
+    execution_payload_states*: HashSet[Eth2Digest]
+    ptc_vote*: Table[Eth2Digest, PtcVotes]
+    ptc_data_availability*: Table[Eth2Digest, PtcVotes]
+    block_timeliness*: Table[Eth2Digest, array[2, bool]]
 
   QueuedAttestation* = object
     attesting_indices*: seq[ValidatorIndex]
     block_root*: Eth2Digest
     slot*: Slot
+    committee_index*: CommitteeIndex
 
   ForkChoice* = object
     backend*: ForkChoiceBackend
     checkpoints*: Checkpoints
     queuedAttestations*: seq[QueuedAttestation]
 
+  ForkChoiceNode* = object
+    root*: Eth2Digest
+    payloadStatus*: PayloadStatus
+
 func shortLog*(vote: VoteTracker): auto =
   (
     slot: vote.slot,
     current_root: shortLog(vote.current_root),
     next_root: shortLog(vote.next_root),
+    next_epoch: vote.next_epoch,
+    next_slot: vote.next_slot,
+    payload_present: vote.payload_present
   )
 
 chronicles.formatIt VoteTracker: it.shortLog
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -108,6 +108,7 @@ type
     checkpoints*: FinalityCheckpoints
     nodes*: ProtoNodes
     indices*: Table[Eth2Digest, Index]
+    fullBlockIndices*: Table[Eth2Digest, Index]
     unrealized*: Table[Index, FinalityCheckpoints]
     previousProposerBoostRoot*: Eth2Digest
     previousProposerBoostScore*: Gwei
@@ -148,6 +149,7 @@ type
     next_epoch*: Epoch
     next_slot*: Slot
     payload_present*: bool
+    next_payload_present*: bool
 
   PtcVotes* = object
     voted*: BitArray[int(PTC_SIZE)]
@@ -198,7 +200,8 @@ func shortLog*(vote: VoteTracker): auto =
     next_root: shortLog(vote.next_root),
     next_epoch: vote.next_epoch,
     next_slot: vote.next_slot,
-    payload_present: vote.payload_present
+    payload_present: vote.payload_present,
+    next_payload_present: vote.next_payload_present
   )
 
 chronicles.formatIt VoteTracker: it.shortLog
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -165,7 +165,6 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
-    block_timeliness*: Table[Eth2Digest, array[2, bool]]
 
   QueuedAttestation* = object
     attesting_indices*: seq[ValidatorIndex]
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -122,7 +122,6 @@ type
     invalid*: bool
     bestChild*: Opt[Index]
     bestDescendant*: Opt[Index]
-    proposerIndex*: uint64
 
   BalanceCheckpoint* = object
     checkpoint*: Checkpoint
--- a/beacon_chain/fork_choice/fork_choice_types.nim
+++ b/beacon_chain/fork_choice/fork_choice_types.nim
@@ -9,7 +9,7 @@
 
 import
   # Standard library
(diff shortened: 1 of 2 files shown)


beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -142,6 +146,14 @@ type
     current_root*: Eth2Digest
     next_root*: Eth2Digest
     slot*: Slot
+    next_epoch*: Epoch
```
tersec: In https://github.com/status-im/nimbus-eth2/pull/7808 this seems to be simply defined as `vote.next_epoch = slot.epoch`, i.e. it's the same thing as `next_slot.epoch`, but why have it as a separate field? It's very easy and efficient to shift right by 5 bits (or 3 bits, for 8-slot epochs) to go from slots to epochs, and we generally don't need to store them separately.

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -142,6 +146,14 @@ type
...
+    next_slot*: Slot
```
tersec: I don't fully understand what this is doing? In https://github.com/status-im/nimbus-eth2/pull/7808 it seems to sort of take the place of `vote.slot`, looking at `process_attestation`. Which might be fine, but then I don't know why it `VoteTracker` would need both `slot` and `next_slot`.

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -161,22 +173,35 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
+    execution_payload_states*: HashSet[Eth2Digest]
+    ptc_vote*: Table[Eth2Digest, PtcVotes]
+    ptc_data_availability_vote*: Table[Eth2Digest, PtcVotes]
```
tersec: `ptc_vote` and `ptc_data_availability_vote` in #7808 appear to be generally/always (?) set at the same time, sometimes with the same value (resetting to defaults), sometimes to different values. It's generally more memory-efficient and table-lookup efficient, in that case, to have something like a single `Table[Eth2Digest, ...]` which treats them at the same time.

beacon_chain/fork_choice/fork_choice_types.nim
Tomi-3-0: The `next_slot`  field was to mirror the `current_root/next_root` pattern where (current = applied, next = pending), but the existing slot field already can serve this purpose. You are right, it's redundant.

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/a0a723773ad506d46350f9a84df1bef69e744182

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/a0a723773ad506d46350f9a84df1bef69e744182

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/a0a723773ad506d46350f9a84df1bef69e744182

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -161,22 +172,32 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
+    execution_payload_states*: HashSet[Eth2Digest]
```
tersec: In #7808 the lifecycle of this is
```nim
  self.backend.execution_payload_states.incl(beacon_block_root)
  ? self.backend.proto_array.onPayloadVerified(beacon_block_root)
```
in `on_execution_payload()`,
```nim
  # [New in Gloas:EIP7732]
  # If attesting for a full node, the payload must be known
  if dag.isGloasEnabled(attestation_slot) and
      attestation_committee_index == CommitteeIndex(1) and
      beacon_block_root notin self.backend.execution_payload_states:
    return ok()
```
in `on_attestation`, and
```nim
  when typeof(blck).kind >= ConsensusFork.Gloas:
    blkBidBlockHash = blck.body.signed_execution_payload_bid.message.block_hash
    let parentNode = self.getNode(blck.parent_root)
    if parentNode != nil and
        blck.body.signed_execution_payload_bid.message.parent_block_hash ==
          parentNode.bidBlockHash:
      parentPayloadStatus = PAYLOAD_STATUS_FULL
  else:
    if blck.parent_root in self.backend.execution_payload_states:
      parentPayloadStatus = PAYLOAD_STATUS_FULL
```
in `process_block`.

But it's unclear to me
- (a) whether the `process_block` check for `if blck.parent_root in self.backend.execution_payload_states` makes sense. That is, pre-Gloas and post-merge, all blocks have payloads, so this seems like it would be tautological, leaving
- (b) `on_attestation`, which had and still has access to `dag: ChainDAGRef` in #7808, and presumably could use that to check the existence of the payload directly in the `BlockRef` without creating and maintaining a separate fork choice-specific data structure.

This would leave no use case left of `execution_payload_states`, or is there a gap in this reasoning (could well be)?

(yes, this is based on #7808 but it's hard to evaluate this PR in isolation. But I am focusing on the data structures here, and aspects specifically relevant to the data structures)

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -121,6 +122,9 @@ type
     invalid*: bool
     bestChild*: Opt[Index]
     bestDescendant*: Opt[Index]
+    parentPayloadStatus*: PayloadStatus
+    bidBlockHash*: Eth2Digest
```
tersec: The one place this appears to be read in #7808 is in `ForkChoice.process_block`:
```nim
  when typeof(blck).kind >= ConsensusFork.Gloas:
    blkBidBlockHash = blck.body.signed_execution_payload_bid.message.block_hash
    let parentNode = self.getNode(blck.parent_root)
    if parentNode != nil and
        blck.body.signed_execution_payload_bid.message.parent_block_hash ==
          parentNode.bidBlockHash:
```
But it's already doing a `let parentNode = self.getNode(blck.parent_root)` and the DAG's `BlockRef` already should have the execution payload hashes: https://github.com/status-im/nimbus-eth2/blob/cee0fb994b6b90f828edb2e80b75f0a7aa0ce1ce/beacon_chain/consensus_object_pools/block_dag.nim#L24-L42

These sometimes are not initialized, but https://github.com/status-im/nimbus-eth2/blob/cee0fb994b6b90f828edb2e80b75f0a7aa0ce1ce/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1102-L1119 (the `BlockId` overload always hits `getForkedBlock`, it's actually not ideal as an API design, looking at it now, having two similar-looking functions with similar-looking arguments have such extremely different costs) addresses this case, and presumably the database load shouldn't be firing newly in that approach vs here.

beacon_chain/fork_choice/fork_choice_types.nim
Tomi-3-0: The field was an attempt to make the payload_states self contained in forkchoice without needing access to the dag

tersec: Which I appreciate, but in this case it had and has access to the DAG already: https://github.com/status-im/nimbus-eth2/blob/cee0fb994b6b90f828edb2e80b75f0a7aa0ce1ce/beacon_chain/fork_choice/fork_choice.nim#L413-L466 in `unstable` today, from which it calls `get_attesting_indices()`, plus various other helpers which need it (`update_proposer_boost_root`). Since this doesn't involve passing the `dag` to any other functions than it already has, just using that existing datastore, that seems preferable.

In general, I would be wary of creating new structural dependencies which aren't there, but this isn't one.

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/3801cb113cf63904cb15cf7c2b3d4f650034c871

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/3801cb113cf63904cb15cf7c2b3d4f650034c871

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -161,22 +170,31 @@ type
...
     block_root*: Eth2Digest
     slot*: Slot
+    committee_index*: CommitteeIndex
 
   ForkChoice* = object
     backend*: ForkChoiceBackend
     checkpoints*: Checkpoints
     queuedAttestations*: seq[QueuedAttestation]
 
+  ForkChoiceNode* = object
```
tersec: How is this intended to be used? I don't see an equivalent in #7808.

Tomi-3-0: I removed it and it's redundant now. must have forgotten to update it here

tersec: https://github.com/status-im/nimbus-eth2/pull/8430/commits/48d1aafda6a3eefd8519a38f16afc2119e39fb70

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -142,6 +144,13 @@ type
     current_root*: Eth2Digest
     next_root*: Eth2Digest
     slot*: Slot
+    payload_present*: bool
+    next_payload_present*: bool
+
+  PtcVotes* = object
+    voted*: BitArray[int(PTC_SIZE)]
```
tersec: This field is modified in https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice.nim#L520 but is not read out in the fork choice function itself, only in the tests: https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/tests/consensus_spec/test_fixture_fork_choice.nim#L437-L446

The fork choice function itself shouldn't be carrying this information. If the test fixtures wants to verify this assertion condition, it can track the information separately. It set up the scenario to begin with, after all. But it has no utility in a running node.

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -142,6 +144,13 @@ type
...
+    payload_present*: BitArray[int(PTC_SIZE)]
+    data_available*: BitArray[int(PTC_SIZE)]
```
tersec: In #7808, this field is indeed written to by https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice.nim#L523-L524, but nothing in the production codepaths read it. It's only read in the tests, by https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/tests/consensus_spec/test_fixture_fork_choice.nim#L442-L445

As with `PtcVotes.voted`, this is overhead the fork choice code itself doesn't need/benefit from. It does enable some assertions in the test, but if they're sufficiently important, the test can construct this data structure on its own.

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -142,6 +144,13 @@ type
     current_root*: Eth2Digest
     next_root*: Eth2Digest
     slot*: Slot
+    payload_present*: bool
+    next_payload_present*: bool
+
+  PtcVotes* = object
+    voted*: BitArray[int(PTC_SIZE)]
+    payload_present*: BitArray[int(PTC_SIZE)]
```
tersec: #7808 writes but only in the tests reads this field: https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice.nim#L521-L522 writes and https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/tests/consensus_spec/test_fixture_fork_choice.nim#L442-L445 reads.

This for testing only but creates permanent overhead for the beacon node fork choice code.

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -142,6 +144,13 @@ type
     current_root*: Eth2Digest
     next_root*: Eth2Digest
     slot*: Slot
+    payload_present*: bool
+    next_payload_present*: bool
+
+  PtcVotes* = object
```
tersec: The entire `PtcVotes` object appears to be an affordance for https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/tests/consensus_spec/test_fixture_fork_choice.nim#L437-L445 and shouldn't be in the fork choice code itself.

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -161,11 +170,14 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
+    ptc_vote*: Table[Eth2Digest, PtcVotes]
```
tersec: Nothing in `PtcVotes` appears to be read in `beacon_chain/fork_choice/`, only `tests/consensus_spec/test_fixture_fork_choice.nim`.

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/f1bc67adb65021374bc46f9e034ec82d4ac19d74

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/f1bc67adb65021374bc46f9e034ec82d4ac19d74

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/f1bc67adb65021374bc46f9e034ec82d4ac19d74

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/f1bc67adb65021374bc46f9e034ec82d4ac19d74

beacon_chain/fork_choice/fork_choice_types.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8430/changes/f1bc67adb65021374bc46f9e034ec82d4ac19d74

beacon_chain/fork_choice/fork_choice_types.nim
```diff
@@ -161,11 +165,13 @@ type
     previous_slot_head*, current_slot_head*: Eth2Digest
     votes*: seq[VoteTracker]
     balances*: seq[ForkChoiceBalance]
+    block_timeliness*: Table[Eth2Digest, array[2, bool]]
```
tersec: `block_timeliness` is both written to and then read from within 6 lines of `ForkChoice.process_block()`, on the same `blckRef.root`:
https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice.nim#L457-L462

In particular, it's read from in `record_block_timeliness(..., ..., blckRef.root, ..., ...)`:
https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice_gloas.nim#L116-L126

And written in `update_proposer_boost_root`: https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice_gloas.nim#L112-L114

But it's the same `root`, all occurring under the aegis of those 6 lines of `ForkChoice.process_block()`, so using this `block_timeliness` data structure as an intermediary doesn't provide any added value. Can still use `record_block_timeliness` to compute  https://github.com/status-im/nimbus-eth2/blob/33a032a11d66a9b1d35ae4b4e050d4c9ab98ff09/beacon_chain/fork_choice/fork_choice_gloas.nim#L104-L105 which is what the fork choice cares about, return and store that single `bool` as a local in `ForkChoice.process_block()`, the pass it to `update_proposer_boost_root`.

tersec: https://github.com/status-im/nimbus-eth2/pull/8430/commits/b731e315c634af466767618febd7fcb4fc6c6c40

PR #8434, opened 2026-05-12
deprecate blob sidecar support in block publishing paths

agnxsh:


beacon_chain/rpc/rest_beacon_api.nim
```diff
@@ -1567,22 +1563,6 @@ proc installBeaconApiHandlers*(router: var RestRouter, node: BeaconNode) =
...
-    # The merkleization limit of the list is `MAX_BLOB_COMMITMENTS_PER_BLOCK`,
-    # the serialization limit is configurable and is:
-    # - `MAX_BLOBS_PER_BLOCK` from Deneb onward
-    # - `MAX_BLOBS_PER_BLOCK_ELECTRA` from Electra.
-    handleDataSidecarRequest[
-      InvalidBlobSidecarIndexValueError,
-      List[BlobSidecar, Limit MAX_BLOB_COMMITMENTS_PER_BLOCK]
-    ](
-      node, preferredContentType(jsonMediaType, sszMediaType),
-      block_id, indices, node.dag.cfg.MAX_BLOBS_PER_BLOCK_ELECTRA)
```
tersec, line 1584: This [has been deprecated already](https://github.com/ethereum/beacon-APIs/blob/v4.0.0/CHANGES.md), but will be removed [once Glamsterdam forks on mainnet](https://github.com/ethereum/beacon-APIs/blob/v5.0.0-alpha.1/CHANGES.md), not quite yet.

beacon_chain/rpc/rest_validator_api.nim
```diff
@@ -453,43 +453,6 @@ proc installValidatorApiHandlers*(router: var RestRouter, node: BeaconNode) =
             node.hasRestAllowedOrigin)
         else:
           raiseAssert "preferredContentType() returns invalid content type"
-      elif consensusFork == ConsensusFork.Electra:
```
tersec, line 456: This deprecates more than just blobs in Electra? This deprecates Electra block publishing support entirely from REST.

agnxsh: yes, so technically, the blob sidecar support is kind of knit together with the electra block publishing route, i think it's better to proceed with this once glamsterdam is mainnet, as you pointed out above.

i was planning to deprecate it in the favor of getting rid of some of the generics and union types that we use, have used previously, but trying to deprecate say just parts of message router breaks the overall engine block flow, we can think about this later.

tersec:
Got rather merge-conflicted. Superseded by
- #9070

PR #8436, opened 2026-05-13
add beacon blocks by head

agnxsh:
https://github.com/ethereum/consensus-specs/pull/5181

### DIFF
--- a/beacon_chain/sync/sync_protocol.nim
+++ b/beacon_chain/sync/sync_protocol.nim
@@ -435,14 +435,21 @@ p2pProtocol BeaconSync(version = 1,
       if found >= count:
         break
 
-      # `dag.parent` walks the in-memory `BlockRef` tree while above the
-      # finalized head and falls through to the canonical chain in the
-      # database (via `getBlockIdAtSlot`) below it - so the walk spans the
-      # full available history, not just the head-to-finalized window.
-      let parentBid = dag.parent(cur).valueOr:
-        # No further ancestor is resolvable. On a checkpoint-synced node
-        # this typically means we hit `dag.backfill.slot` (the lower bound
-        # of locally available history), not genesis.
+      if cur.slot == GENESIS_SLOT:
+        break
+
+      # Walk the parent chain via the block summary table (root-indexed),
+      # not via `dag.parent` - `dag.parent` for finalized blocks goes
+      # through `getBlockIdAtSlot` (slot-indexed `dag.db.finalizedBlocks`),
+      # which can run out before the summary table does. Every block we
+      # have stored has a summary, so this walk spans the full available
+      # history regardless of which side of the finalized head we're on.
+      let summary = dag.db.getBeaconBlockSummary(cur.root).valueOr:
+        break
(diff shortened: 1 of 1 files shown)


beacon_chain/sync/sync_protocol.nim
```diff
@@ -373,6 +373,68 @@ p2pProtocol BeaconSync(version = 1,
...
+        uncompressedLen, bytes,
+        peer.network.forkDigestAtEpoch(cur.slot.epoch).data)
+
+      inc found
+
+      if found >= count:
+        break
+
+      # Walk to the parent; stop at genesis or when the next ancestor falls
+      # outside the epoch range we are required to serve.
```
tersec: https://github.com/ethereum/consensus-specs/blob/945a70ae5bf1e6d8d4835c5bce4e4675b5fcec85/specs/fulu/p2p-interface.md#beaconblocksbyhead-v1 says
> The walk stops as soon as the response contains `min(count, MAX_REQUEST_BLOCKS_DENEB)` blocks or the next ancestor falls outside the epoch range that clients are required to serve (see below).
>
> ...
>
> Clients MUST support requesting blocks on the epoch range `[current_epoch - compute_min_epochs_for_block_requests(), current_epoch]`. If `beacon_root` references a block earlier than this range, peers MAY respond with error code `3: ResourceUnavailable` or with no blocks in the response.

It enforces the `MAX_REQUEST_BLOCKS_DENEB` constraint but does not necessarily serve the full epoch range it should. Because `MAX_REQUEST_BLOCKS_DENEB` is 128 on mainnet/Hoodi/Sepolia, when they're finalizing after 2 epochs, the head-to-finalized slot range will typically be from 2-3 epochs, less than `MAX_REQUEST_BLOCKS_DENEB`, so just walking the DAG doesn't suffice.

The comment is also incorrect in describing what the code does, because the DAG typically does not reach back to genesis.

tersec: https://github.com/status-im/nimbus-eth2/pull/8436/commits/7c65af872af307dced368360bc95646d98d190bb

beacon_chain/sync/sync_protocol.nim
```diff
@@ -373,6 +373,88 @@ p2pProtocol BeaconSync(version = 1,
...
+
+      # `dag.parent` walks the in-memory `BlockRef` tree while above the
+      # finalized head and falls through to the canonical chain in the
+      # database (via `getBlockIdAtSlot`) below it - so the walk spans the
+      # full available history, not just the head-to-finalized window.
+      let parentBid = dag.parent(cur).valueOr:
+        # No further ancestor is resolvable. On a checkpoint-synced node
+        # this typically means we hit `dag.backfill.slot` (the lower bound
+        # of locally available history), not genesis.
+        break
```
tersec, line 453: This still `break`s the entire loop once it runs out of DAG.

tersec: https://github.com/status-im/nimbus-eth2/pull/8436/changes/fc273cdf6f0d7f608a6f4b8deb22afc03e372428

PR #8446, opened 2026-05-14
Preserve negative peer score across reconnections

etan-status:
When peer with negative score reconnects, disconnect immediately instead of resetting the score to 0.


tersec:
How likely is this to cause it to gradually blacklist enough peers not to be able to find others?

The current equilibrium practically is balanced between "forgiving" peers (intentionally or otherwise) but disconnecting them and not in the moment reconnecting them, creating a bias but not a ban.

This also means it adapts to circumstances where enough time has passed that maybe those peers have updated client versions or restarted, since most CLs keep their network identities across updates and restarts.

etan-status:
The LowPeerScore ban is ~1 hour. What's your thought about just preserving the current entry (so that it eventually expires), rather than extending it with a new ban? As it is now, bans don't have any effect, as they can just immediately reconnect to us.

etan-status:
With latest commit, the re-disconnect preserves the existing ban timer rather than extend it.

PR #8449, opened 2026-05-15
feat(rest): expose peer score and disconnect reason on /eth/v1/node/peers

barnabasbusa:
## Summary

Extends `/eth/v1/node/peers` (and `/eth/v1/node/peers/{peer_id}`) with two optional fields per a simplified beacon-API spec extension currently under discussion:

- `score` — from `Peer.score: int` (Nimbus's application-level reputation score, [0, 1000] with default 300)
- `disconnect_reason` — from a new `lastDisconnectReason` field on `Peer`, mapped to the spec controlled vocab

`agent_version` is already exposed in Nimbus's current response (under field name `agent`).

`downscore_reasons` is **deliberately omitted** — Nimbus doesn't currently track per-event reason history. Adding it would require instrumenting the ~30 call sites that adjust `Peer.score`. Happy to follow up if there's interest.

## Spec proposal

ethereum/beacon-APIs#606

### Commit 1: expose `score`
- `beacon_chain/spec/eth2_apis/rest_types.nim` — add `score*: Opt[int]` to `RestNodePeer`.
- `beacon_chain/rpc/rest_node_api.nim` — populate from `peer.score` in both the list and single-peer handlers.

### Commit 2: expose `disconnect_reason`
- `beacon_chain/networking/eth2_network.nim` — add `lastDisconnectReason*: Opt[DisconnectionReason]` + `lastDisconnectAt*: Opt[chronos.Moment]` to the `Peer` object. Instrument `proc disconnect*` at the top, so all ~20 disconnect call sites are caught at the single choke point.
- `beacon_chain/spec/eth2_apis/rest_types.nim` — add `disconnect_reason*: Opt[string]` to `RestNodePeer`.
- `beacon_chain/rpc/rest_node_api.nim` — add `mapDisconnectReason()` translating Nimbus's 4-value `DisconnectionReason` enum to the spec controlled vocab (`ClientShutDown → client_shutdown`, `IrrelevantNetwork → irrelevant_network`, `FaultOrError → io_error`, `PeerScoreLow → bad_score`).

## Coordinated implementations

Part of a coordinated multi-client effort — see ethereum/beacon-APIs#606 for the other five client PRs.

## Status

**Draft.** `nim check` on all three patched files: identical error counts vs. pre-patch baseline (only pre-existing version.nim / forks.nim issues, none introduced by this patch).

### DIFF
--- a/beacon_chain/rpc/rest_node_api.nim
+++ b/beacon_chain/rpc/rest_node_api.nim
@@ -221,8 +221,12 @@ proc installNodeApiHandlers*(router: var RestRouter, node: BeaconNode) =
           agent: node.network.switch.peerStore[AgentBook][peer.peerId],
           proto: node.network.switch.peerStore[ProtoVersionBook][peer.peerId],
           score: Opt.some(peer.score),
+          # Per beacon-API spec, `disconnect_reason` MUST only be populated
+          # when `state` is `disconnected` or `disconnecting`.
           disconnect_reason:
-            if peer.lastDisconnectReason.isSome():
+            if peer.connectionState in
+                {ConnectionState.Disconnected, ConnectionState.Disconnecting} and
+                peer.lastDisconnectReason.isSome():
               Opt.some(mapDisconnectReason(peer.lastDisconnectReason.get()))
             else:
               Opt.none(string)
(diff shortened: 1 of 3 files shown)


beacon_chain/rpc/rest_node_api.nim
```diff
@@ -109,6 +109,19 @@ proc toString(direction: PeerType): string =
   of PeerType.Outgoing:
     "outbound"
 
+proc mapDisconnectReason(reason: DisconnectionReason): string =
```
tersec, line 112: `func`

PR #8450, opened 2026-05-15
nix: add multiple gcc versioned BN and VC package outputs to flake

markoburcul:
Goal of this improvement is to be able to test BN and VC built with different GCC version in the CI and fleet.

Referenced issue:
* https://github.com/status-im/infra-nimbus/issues/276

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,12 +3,18 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs?rev=2a777ace4b722f2714cc06d596f2476ee628c04a";
+
+    # Backfill source for gcc11/gcc12, pinned to nixos-25.05.
+    # Intentionally does not follow nixpkgs since we need the older toolchains.
+    nixpkgsGccCompat.url = "github:NixOS/nixpkgs?rev=ac62194c3917d5f474c1a844b6fd6da2db95077d";
+
     # WARNING: Do not use relative path, it breaks caching of NBS.
     # Remember to call 'nix flake update' when NBS is updated.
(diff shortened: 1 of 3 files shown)


nix/default.nix
```diff
@@ -4,6 +4,8 @@
   self ? {},
   # Nimbus-build-system package.
   nim ? null,
+  # GCC version to compile the target with.
+  gcc ? null,
```
jakubgs, line 8: If you leave it without default it will just use `gcc` package from `nixpkgs` if not provided explicitly:

```suggestion
  gcc,
```

flake.nix
```diff
@@ -18,40 +24,71 @@
...
+          build = targets: gcc:
+            buildTarget.override { inherit targets nim gcc; };
+
+        in rec {
+          beacon_node_gcc11 = build ["nimbus_beacon_node"] pkgs.gcc11;
+          beacon_node_gcc12 = build ["nimbus_beacon_node"] pkgs.gcc12;
+          beacon_node_gcc13 = build ["nimbus_beacon_node"] pkgs.gcc13;
+          beacon_node_gcc14 = build ["nimbus_beacon_node"] pkgs.gcc14;
+          beacon_node_gcc15 = build ["nimbus_beacon_node"] pkgs.gcc15;
+          beacon_node       = beacon_node_gcc14;
```
jakubgs, line 75: Worth explaining why we use 14 here and below.
```suggestion
          # Current default in latest Ubuntu LTS release.
          beacon_node       = beacon_node_gcc14;
```

jakubgs (approved):

PR #8458, opened 2026-05-16
Apply KZG check to columns fetched via Req/Resp

etan-status:
- #4732 KZG primitives added
- #4785 added gossip validation for BlobSidecar but did not check KZG
- #4844 added req/resp for BlobSidecar but did not validate responses
- #5445 fixed KZG validate_blobs in state transition (Opt[bool] jank)
- https://github.com/ethereum/consensus-specs/pull/3531 new spec hits
- #5565 new spec, + fixed gossip missing inclusion proof + KZG checks
- #6109 goerli hardening, + fixed Req/Resp missing inclusion proof check
- #6741 copies the remaining req/resp missing KZG check gap to Fulu
- #8254 removes the Deneb blob_quarantine
- This one now finally adds the missing KZG checks to req/resp

### DIFF
--- a/beacon_chain/sync/request_manager.nim
+++ b/beacon_chain/sync/request_manager.nim
@@ -153,6 +153,16 @@ func cmpColumnIndex(x: ColumnIndex, y: ref fulu.DataColumnSidecar): int =
 func checkColumnResponse(idList: seq[DataColumnsByRootIdentifier],
                          columns: openArray[ref fulu.DataColumnSidecar]):
                          Opt[seq[DataColumnResponseRecord]] =
+  # The response is a list of DataColumnSidecar whose length is
+  # less than or equal to requested_columns_count, where
+  # requested_columns_count = sum(len(r.columns) for r in request).
+  # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.8/specs/fulu/p2p-interface.md#datacolumnsidecarsbyroot-v1
+  var expectedMax = 0
+  for id in idList:
+    expectedMax += id.indices.len
+  if columns.len > expectedMax:
+    return Opt.none(seq[DataColumnResponseRecord])
+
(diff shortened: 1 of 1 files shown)


agnxsh:
This just simply double kzg verifies the entire thing, since we are already kzg verfying things from this pathway inside the block processor, the kzg check was indeed added at a point in 2025 to the req man column response checking flow, but honestly with the sheer number of columns to validate, we don't want to waste this time double verfying kzg proofs

agnxsh:
there might be a latency improvement now, since we have done kzg batching, plus a precompute optimization but, inspite of these, i'd still say double checking this is pointless, perhaps we can create a source parameter if it comes from rman and then if the rman source parameter matches inside the block processor we can straightaway bypass kzg verification of the same data in block processor

we do double verify in the gossip flow as per my understanding, simply because there's a network obligation of timely redistribution of these columns. but this case is different.

etan-status:
I indeed missed the flow in the block processor!

It's also interesting, there's #7301 which _moved_ it to ingress (similar to the PR here), but that was merged into a different branch, not unstable.

Cleanest would probably be to adopt the same type-safety solution that we use for blocks (to distinguish untrusted, sig-verified, and trusted). The source parameters don't come with type safety and can introduce slight bugs. But maybe it's tricky to get the type safety in properly.

I still think it's correct to do the check in the request_manager:

1. We downscore the correct peer that sent us the bad columns, instead of the one that gave us the good block
2. Quarantine only ever hits kzg-verified data, less scenarios where someone can evict the good data by just sending us bad proofs.

But, argument for block_processor is that maybe it's faster due to batching benefits? i.e., we can delay the check until we got all the columns, and then batch-verify them all in one go. I'm not sure how that one scales. 

For now, will revert this PR, as indeed, double verification shouldn't happen.

PR #7623, opened 2025-10-17
Create execution payload gossip validation

ahshum:

PR #8462, opened 2026-05-17
Add bounds to envelope quarantine

etan-status:
Apply same max capacity to envelope quarantine, as is also applied to beacon blocks. This should retain the same memory bounds as pre-Glaos. Beacon blocks will get smaller with Gloas, envelopes take over. Original envelope quarantine was introduced without bounds in #7623, with the cleanupOrphans function left unused.

### DIFF
--- a/beacon_chain/sync/request_manager.nim
+++ b/beacon_chain/sync/request_manager.nim
@@ -270,7 +270,6 @@ proc fetchEnvelopesFromNetwork(self: RequestManager, roots: seq[Eth2Digest])
           gotUnviableEnvelope = false
 
         for envelope in uenvelopes:
-          self.envelopeQuarantine[].addOrphan(envelope[])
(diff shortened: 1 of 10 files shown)


etan-status:
Roughly tried to match block_quarantine patterns.

beacon_chain/sync/request_manager.nim
```diff
@@ -270,7 +270,6 @@ proc fetchEnvelopesFromNetwork(self: RequestManager, roots: seq[Eth2Digest])
           gotUnviableEnvelope = false
 
         for envelope in uenvelopes:
-          self.envelopeQuarantine[].addOrphan(envelope[])
```
etan-status, line 273: Moved to the envelopeVerifier in nimbus_beacon_node as rm doesn't have dag.

PR #8464, opened 2026-05-18
Revert "Apply KZG check to columns fetched via Req/Resp (#8458)"

etan-status:
This reverts commit 8b56bbcfbe9b05e5ab6d85f5c0dff7db5b5ace38.

In the original PR #8458 I overlooked the flow in block_processor that already contains the KZG check (at a later stage). It's not perfect, as it penalizes the incorrect peer (the one who sent the block, instead of the one that sent us the bad column), and because it allows quarantine pollution with bad data, but the check is indeed there! Double checking just for minor benefits in peer score accounting is too expensive. Should revisit later with a proper solution to avoid double-checking, ideally with a type-system approach similar to blocks.


etan-status:
ty @agnxsh for catching the performance regression!

PR #8468, opened 2026-05-18
ci: use url when building with nix

markoburcul:
We need to push the derivation artifacts that are the same as the ones we use in the nimbus fleet so we can maximize utilization of the cache, The problem with nix is described here: https://github.com/NixOS/nix/issues/15875


markoburcul:
1.

PR #8475, opened 2026-05-19
Forward compatible consensus data structures (EIP-7688)

etan-status:
- EIP https://eips.ethereum.org/EIPS/eip-7688
- Spec https://github.com/ethereum/consensus-specs/pull/4630
- Tests https://github.com/ethereum/consensus-specs/pull/4630#issuecomment-4352744403
- Kurtosis https://github.com/ethereum/consensus-specs/pull/4630#issuecomment-4377125436


beacon_chain/nimbus_signing_node.nim
```diff
@@ -226,9 +226,17 @@ proc installApiHandlers*(node: SigningNodeRef) =
       of Web3SignerRequestKind.AggregateAndProofV2:
         let
           forkInfo = request.forkInfo.get()
-          signature = get_aggregate_and_proof_signature(forkInfo.fork,
-            forkInfo.genesis_validators_root, request.aggregateAndProofV2,
-            validator.data.privateKey).toHex()
+          forked = request.aggregateAndProofV2
+          signature =
+            if forked.kind >= ConsensusFork.Gloas:
```
tersec: It might be cleaner to instead add to the typeclass at
https://github.com/status-im/nimbus-eth2/blob/c92036f7eceef35a8bfaddf37b0b70f8df9c42fe/beacon_chain/spec/signatures.nim#L129-L137 since apparently it's `forked.data` either way, just, a different `forked.data`.

etan-status: 6872c1284

PR #8481, opened 2026-05-20
Store head payload by fork choice

ahshum:
- added `headPayload` as `BlockRef` under DAG
- defined `should_extend_payload` as `dag.head == dag.headPayload`
- benefits of doing so
    - We do not need updating `headPayload` on `head` updated, i.e. not extending a payload until we receive a valid envelope.
    - ~~Only minimal changes to DAG `init`, but no changes to `preInit`.~~ (no need for persistence, in-memory only)
- regards to invalid envelopes, `head.executionValid` should be checked against `head.executionParent` instead for proposal and in-synced checks.


beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -2526,6 +2526,23 @@ proc executionParent*(
...
+  ## on the head payload, as the payload selected by fork choice is stored in
+  ## DAG.
+  ##
+  ## Related fork choice helpers
+  ## https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-should_build_on_full
+  ## https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-should_extend_payload
+  (
+    # For either genesis or pre-Gloas block, we should always build on them.
+    head.slot == GENESIS_SLOT or
+    dag.cfg.consensusForkAtEpoch(head.slot.epoch) < ConsensusFork.Gloas or
```
tersec: https://github.com/status-im/nimbus-eth2/blob/0ee3b53911d4e34122b4a4569f8f86042184f1af/beacon_chain/spec/forks.nim#L1157-L1179

checks the entire set of possible forks this epoch could be in. This code only cares specifically about whether `head.slot.epoch < dag.cfg.GLOAS_FORK_EPOCH`, which is a single comparison.

beacon_chain/consensus_object_pools/blockchain_dag.nim
```diff
@@ -2788,37 +2805,25 @@ proc updateHead*(
...
-      dag, dag.headState, head.bid.atSlot(), false, cache,
-      dag.updateFlags):
-    # Advancing the head state should never fail, given that the tail is
-    # implicitly finalised, the head is an ancestor of the tail and we always
-    # store the tail state in the database, as well as every epoch slot state in
-    # between
-    fatal "Unable to load head state during head update, database corrupt?"
-    quit 1
-
+  dag.headPayload = headPayload
```
tersec, line 2826: `dag.headPayload` is only updated here. A couple of the edge cases are fine, or not very significant, relating to Gloas fork boundary which will generally happen exactly once. But if `dag.head` reorgs, nothing automatically adjusts or checks for whether `dag.headPayload` should change too. And that means that it will incorrectly think that `dag.head != dag.headPayload` when building on the reorg'd head.

beacon_chain/consensus_object_pools/blockchain_dag.nim
tersec:
```suggestion
    head.slot.epoch < dag.cfg.GLOAS_FORK_EPOCH or
```

beacon_chain/consensus_object_pools/blockchain_dag.nim
ahshum: This function (`updateHeadExecutionPayload`) should be called after we got the decision from fork choice, similar to beacon block. I would expect the handler in #8604 to notify fork choice. Once the fork choice of payload has made a decision for the selected head, we will update the headPayload with this function. 

> And that means that it will incorrectly think that dag.head != dag.headPayload when building on the reorg'd head.

So I believe what you mentioned is on fork choice/ how we trigger fork choice?

tersec: Ok, it just doesn't happen currently. But yeah, agree, it needs to be plumbed through separately.

PR #8486, opened 2026-05-21
fix: gloas block gossip validation (2)

etan-status:
Followup from #8422 to correctly handle the case when a new block points to the finalized / genesis as the parent payload (e.g., testnet slot 1). Old logic triggers IGNORE on these, as "found = false" is stuck and also puts the genesis payload into the quarantine as code mismatches comment.


ahshum (approved):
Thanks

PR #8496, opened 2026-05-22
Execution valid for Gloas

ahshum:
Part of #8481

Fix the false positive on skipping block proposal due to EL out-of-sync.


beacon_chain/consensus_object_pools/block_pools_types.nim
```diff
@@ -163,6 +163,18 @@ type
       ## The most recently known head, as chosen by fork choice; might be
       ## optimistic
 
+    headPayload*: BlockRef
+      ## The known payload head that is chosen by fork choice. It will be used
+      ## on the next block proposal for building payload on either the current
+      ## head (parent) or the parent of the current head (grandparent).
+      ##
+      ## Uses only since Gloas.
```
tersec:
```suggestion
      ## Used only since Gloas.
```

beacon_chain/validators/beacon_validators.nim
```diff
@@ -1499,7 +1499,7 @@ proc handleValidatorDuties*(node: BeaconNode, lastSlot, slot: Slot) {.async: (ra
 
     return
 
-  elif not head.executionValid:
+  elif not node.dag.headExecutionValid(head, node.dag.headPayload):
```
tersec: Why should Nimbus refuse to build on a gloas CL-valid CL block which happens not to have an envelope?

tersec: Of course, fork choice should descore it according to the usual specs, but it's not invalid, it's just a low-weight fork.

ahshum: > Why should Nimbus refuse to build on a gloas CL-valid CL block which happens not to have an envelope?

The changes here is not to refuse building on a valid CL block. It is to produce a new CL block based on head but not the head payload, e.g. in case of invalid envelope.

> Of course, fork choice should descore it according to the usual specs, but it's not invalid, it's just a low-weight fork.

This is needed because envelope processing updates head execution validity. Without this, when the head envelope is absent for whatever reasons, head would not be execution valid and the next proposal would not be reachable due to this.

https://github.com/status-im/nimbus-eth2/blob/408a16967443386ec20a6dff04b9722daf964079/beacon_chain/validators/beacon_validators.nim#L1520

tersec: Given a CL head `BlockRef`, at any given time, there is only one possible `headPayload` though. And that's what the beacon node has to decade based on at any given time, ultimately: the CL block head to build on, or in general, whether that CL head is optimistic, taken as a whole. So the optimisticness/lack thereof which determines whether it does so shouldn't be at that level a function of a separately determined execution block hash. It's already determined by the CL head `BlockRef`.

So yes, the head envelope might be missing or invalid, but conceptually the is-buildable query is on a single CL head. The trouble with addressing these specific callers, but not the existing API in general, is that all the other users of `executionValid`, e.g., these are all basically about whether the CL block is optimistic or valid, and right now would presumably have the same issue as this PR fixes in maybe 2 places:
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/validators/beacon_validators.nim#L1612-L1618
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/nimbus_beacon_node.nim#L1942-L1947
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/nimbus_beacon_node.nim#L2048-L2051
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/rpc/rest_validator_api.nim#L280-L283
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/rpc/rest_validator_api.nim#L395-L396
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/rpc/rest_validator_api.nim#L554-L557
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/rpc/rest_validator_api.nim#L841-L842
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/rpc/rest_node_api.nim#L271-L275
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/consensus_object_pools/blockchain_dag.nim#L2123-L2133
https://github.com/status-im/nimbus-eth2/blob/d46545bdc936038b0909d1c3f9c7017890538049/beacon_chain/fork_choice/fast_confirmation.nim#L510-L515

tersec: Deliberately separating messages for clarity.

There are more examples of callers, e.g., in the REST beacon API server both of `isExecutionValid` and its downstream `is_optimistic`, and #8481 which this PR derives from doesn't really address that aspect properly either.

The core point is that this valid-vs-optimistic distinction isn't only tied to block production, and arguably what Gloas instead changes is that certain CL blocks with missing payloads might not be "optimistic".
One can have truly "optimistic" scenarios, but they look different than pre-Gloas ones, because they require that the block point to a most recent ancestor execution result which is both part of the chain and not verified locally by the local EL.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -244,8 +249,37 @@ func shortLog*(v: BlockSlot): string =
   else: # There was a gap - log it
     shortLog(v.blck) & "@" & $v.slot
 
+func executionParent*(blck: BlockRef): Opt[BlockRef] =
+  result = Opt.none(BlockRef)
+  if isNil(blck.parent) or blck.executionParentHash.isNone():
+    return
+
+  # Parent hash of pre-Gloas blocks is zero but it could be built on the
+  # genesis. Either cases, the execution parent should be same as the block
```
tersec:
```suggestion
  # genesis. Either way, the execution parent should be same as the block
```

tersec: https://github.com/status-im/nimbus-eth2/pull/8496/changes/91ab745035545c8cbe7ce7b9bfb959eb89973e24

beacon_chain/validators/beacon_validators.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8496/changes/e381e9a0a9a15f03402ccc3bf2d0fa20672fc063

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -244,8 +249,37 @@ func shortLog*(v: BlockSlot): string =
   else: # There was a gap - log it
     shortLog(v.blck) & "@" & $v.slot
 
+func executionParent*(blck: BlockRef): Opt[BlockRef] =
+  result = Opt.none(BlockRef)
```
tersec: https://status-im.github.io/nim-style-guide/language.result.html
>
> Use expression-based return

tersec: https://github.com/status-im/nimbus-eth2/pull/8496/changes/4f5ee480e3883f02ca5fb77fccb38e81dcb22dea

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -244,8 +249,37 @@ func shortLog*(v: BlockSlot): string =
...
+      break
+    cur = cur.parent
+  Opt.none(BlockRef)
+
 func executionValid*(blck: BlockRef): bool =
-  blck.optimisticStatus == OptimisticStatus.valid
+  if blck.optimisticStatus == OptimisticStatus.valid:
+    return true
+
+  # Fallback to its execution parent if blck is not valid.
```
tersec, line 297: The block could also be `OptimisticStatus.invalidated`, in which case there's also no need to look at its parents. That shouldn't ever be true for the head fork choice provides; it's a bug if that head is invalid. But it can be true for `BlockRef`s in general.

ahshum: Yes I have raised my concerns earlier that: 

_For checking EL in-synced_, it should fallback whenever the block is not valid (by valid it mean `Optimistic.valid`). Reasoning behind is that if a payload is not valid (invalid as `Optimistic.invalidated` and missing as `Optimistic.notValidated`), fork choice should not choose it so we should be checking its execution parent instead for the EL sync status.

_This should not apply to other call sites_, where they should fallback only when it is `notValidated`.

ahshum: Would you mind further elaborate regards to this? With fallback only when `notValidated`, are we expecting that Nimbus should stop proposing on receiving an invalid payload? Would be great to know more in this particular part.

beacon_chain/consensus_object_pools/block_dag.nim
```diff
@@ -244,8 +249,37 @@ func shortLog*(v: BlockSlot): string =
...
+  if blck.executionParentHash.unsafeGet().isZero():
+    return Opt.some(blck.parent)
+
+  var cur = blck.parent
+  debugGloasComment("revisit the max depth of ancestors")
+  for _ in 0 ..< EXECUTION_PARENT_MAX_DEPTH:
+    if cur.executionBlockHash.isNone():
+      break
+    if cur.executionBlockHash.unsafeGet() ==
+        blck.executionParentHash.unsafeGet():
```
tersec: Regarding `unsafeGet`, in theory, `get`/`value` [use `lent`](https://github.com/arnetheduck/nim-results/blob/06deae1c81fd27b6c94bbc7cd0e619b6905d49da/results.nim#L870-L879):
```nim
func value*[T: not void, E](self: Result[T, E]): maybeLent T {.inline.} =
  ## Fetch value of result if set, or raise Defect
  ## Exception bridge mode: raise given Exception instead
  ## See also: Option.get
  withAssertOk(self):
    when T isnot void:
      # TODO: remove result usage.
      # A workaround for nim VM bug:
      # https://github.com/nim-lang/Nim/issues/22216
      result = self.vResultPrivate
```

So they should also avoid copying the return value. Meanwhile, they have at least a defined outcome for incorrect usage, while [`unsafeGet`/`unsafeValue`](https://github.com/arnetheduck/nim-results/blob/06deae1c81fd27b6c94bbc7cd0e619b6905d49da/results.nim#L902-L905) have in this situation undefined behavior:
```nim
template unsafeValue*[T: not void, E](self: Result[T, E]): T =
  ## Fetch value of result if set, undefined behavior if unset
  ## See also: `unsafeError`
  self.vResultPrivate
```

So it's generally preferable to use `.get` over `.unsafeGet`, unless one can show that the `lent` isn't working. But it works generally well in Nim 2.2.10. `lent` exists [exactly to make such usage efficient](https://nim-lang.org/docs/destructors.html#lent-type):
> To eliminate even more creation/copy <-> destruction pairs, a proc's return type can be annotated as `lent T`. This is useful for "getter" accessors that seek to allow an immutable view into a container.
>
> The `sink` and `lent` annotations allow us to remove most (if not all) superfluous copies and destructions.
>
> `lent T` is like `var T` a hidden pointer. It is proven by the compiler that the pointer does not outlive its origin. No destructor call is injected for expressions of type `lent T` or of type `var T`.

PR #8498, opened 2026-05-22
extend getBlobs in gloas

agnxsh:


beacon_chain/el/el_getblobs_service.nim
```diff
@@ -198,13 +201,72 @@ proc attemptGetBlobs*(
...
+  # Redistribute every reconstructed column to its subnet before any
+  # custody-based filtering — peers on subnets we don't custody still
+  # need them.
+  await self.redistributeColumns(recovered_columns)
+
+  let custodyMap = self.validatorCustody.getMap()
+  var batch = newSeqOfCap[ref gloas.DataColumnSidecar](len(custodyMap))
+  for col in recovered_columns:
+    if col[].index in custodyMap:
+      batch.add col
```
tersec: `let batch = recovered_columns.filterIt(it[].index in custodyMap)`

Yes, right now `batch` does some preallocation, but for low-custody nodes this very much overallocates, so it's not clear it's a net win.

Does it help to do this as a batch, vs `self.gloasColumnQuarantine[].put(blck.root, batch)`'ing these one by one? https://github.com/status-im/nimbus-eth2/blob/7e55a3da9a73b1f4e8821d0ba3207322b23f2930/beacon_chain/consensus_object_pools/column_quarantine.nim#L294-L366 shows the one-by-one is a wrapper over the batch version, so either is valid.

This version does a lot of whole-column copying, and the one-by-one version repeats some table lookups and checks.

My suspicion is that the latter is less expensive than the former, but this is an empirical question.

tersec: https://github.com/status-im/nimbus-eth2/pull/8498/changes/2edbdf78a4a5ca86396f07e4451311b3fd6cccad / https://github.com/status-im/nimbus-eth2/pull/8498/changes/bdfe49f4dedd703e0d74525b726d1bb3e26470bd

beacon_chain/el/el_manager.nim
```diff
@@ -773,21 +773,21 @@ proc getBlobsV2*(
 ): Future[Opt[seq[BlobAndProofV2]]] {.async: (raises: [CancelledError], raw: true).} =
   mixin getBlobsV2
 
-  when blck is gloas.SignedBeaconBlock:
-    debugGloasComment "handle correctly for Gloas?"
-    return err()
-  else:
-    let deadline = sleepAsync(GETBLOBS_TIMEOUT)
+  template kzg_commitments(): auto =
+    when blck is gloas.SignedBeaconBlock:
```
tersec: While not strictly an issue yet, this isn't correct for `heze.SignedBeaconBlock`, or presumably, I* which also won't have the `blob_kzg_commitments` directly in the `blck.message.body`. The usual way this is to use https://github.com/status-im/nimbus-eth2/blob/1cb8fd680c8048534929cd71050a5c956ea90b3d/beacon_chain/spec/forks.nim#L506-L552 to detect rather something like `when typeof(blck).kind >= ConsensusFork.Gloas:`, which doesn't end up requiring manual checks for every post-Gloas fork, including Heze already partly integrated into the codebase.

tersec: https://github.com/status-im/nimbus-eth2/pull/8498/commits/fe093d2e0d5ed308c5b1359bcc02641c2f0fa614

PR #8500, opened 2026-05-22
Bump nim-web3

etan-status:
Adds the blockAccessList field to the BlockObject so that we can obtain the correct execution block hash.

Note that EIP-7843 slot number is still missing. Either spec bumps:

- https://github.com/ethereum/execution-apis/pull/806

or we have to change the API and pass the beacon clock to compute it from the timestamp. Let's wait for spec bump, only affects ≥Amsterdam.


tersec:
Merge conflict with `vendor/nim-web3`.

etan-status:
Weird, there wasn't a conflict when syncing the branch locally.

PR #8501, opened 2026-05-22
ensure sidecarless is cleaned once enqueuePayload is successful + drive enqueuePayload from rman

agnxsh:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -1041,6 +1042,11 @@ proc enqueuePayload*(self: ref BlockProcessor, blck: gloas.SignedBeaconBlock) =
           self.envelopeQuarantine[].addOrphan(
             self.consensusManager.dag.finalizedHead.slot, envelope)
           return
+        # Sidecars are in hand — drop any sidecarless entry left over
+        # from a prior round so the request manager stops chasing columns
+        # we already have, and downstream iterators of `sidecarless` don't
+        # see a block that is already in the DAG.
+        discard self.consensusManager.quarantine[].removeSidecarless(blck.root)
```
ahshum: I think we should add `debugGloasComment` here to note that sidecarless structure is not suitable in/since Gloas.

beacon_chain/sync/request_manager.nim
```diff
@@ -506,7 +512,17 @@ template fetchDataColumnsFromNetworkImpl(
...
+            withBlck(columnless):
+              when consensusFork == ConsensusFork.Gloas:
+                # Drive the payload join with the block in hand. If the
+                # block isn't in DAG yet (rare, e.g. quarantine pressure
+                # caused an early addSidecarless), the join fails with
+                # MissingParent and re-parks envelope+sidecars for the
+                # storeBlock-trailing enqueuePayload to pick up later.
+                rman.payloadEnqueue(forkyBlck)
+              else:
+                discard await rman.blockVerifier(columnless, false)
```
ahshum: You can get the consensusFork from block in `blockVerifier` as well. Why do we need creating `payloadEnqueue`?

beacon_chain/gossip_processing/block_processor.nim
tersec: Indeed, as a general point, the entire idea of sidecarless blocks should probably be removed once we drop this aspect of Fulu gossip/block verification support. The premise of ePBS is that the sidecars are effectively part of the envelope. If anything, they're really more related to the envelope now.

beacon_chain/sync/request_manager.nim
agnxsh: That is because the current BlockVerifier behaviour is technically `addBlock` with `noSidecars` which isn't ideal for Gloas, or i might be missing something, I wanted to separate out the payload trigger especially for the Gloas blocks, but yes, circling back the columns are not really `sidecarless`

ahshum: So there should be a new one specifically for blob/columns instead of reusing `blockVerifier`. That being said, `blockVerifier` should get columns from quarantine for missing blocks pre-Gloas.

agnxsh: yeah you're right, when i said `addBlock` with `noSidecars`, i was talking only about Gloas, anyway, looks like this part needs a redesign, i can maybe touch this part if that's ok with you/ @Tomi-3-0 ? that should include a redesign such that the rman for gloas doesn't refer to sidecarless, plus we either:
a) need to alter what the `BlockVerifier` functions like in gloas
b) have some other verification callback flow and limit `BlockVerifier` pre gloas

ahshum: Another workaround would be `enqueuePayload` in 

https://github.com/status-im/nimbus-eth2/blob/83ac63e1519382aac57dae415b963b9b03dde9bb/beacon_chain/gossip_processing/block_processor.nim#L887-L888

agnxsh: yes, that's correct

PR #8505, opened 2026-05-23
Sync systemd extra flags to nix services

etan-status:
Apply the same extra flags that we set in scripts/package_src .service definitions also when running from nix.

- LimitNOFILE: "Apart from sockets and databases, a file descriptor is needed for every local validator"
- RestartPreventExitStatus: "Don't restart when Doppelganger or slashing detection has been activated"
- wants / after instead or requires: Allow restarting network without restarting BN/VC

All pre-existing for our deb setup, be consistent in the Nix setup.

Checked with

- nix eval --raw '.#checks.aarch64-linux.beacon-node.nodes.machine.systemd.units."nimbus-beacon-node.service".text'
- nix eval --raw '.#checks.aarch64-linux.validator-client.nodes.machine.systemd.units."nimbus-validator-client.service".text'


tersec:
Merge conflict in `nix/services/validator-client.nix`

PR #8512, opened 2026-05-26
Ensure confirmed chain includes greatest unrealized checkpoint

etan-status:
The spec was updated to match our implementation, the current epoch GU has to not be just any ancestor but actually be the expected checkpoint.

- https://github.com/ethereum/consensus-specs/pull/5288


etan-status:
> + Fast Confirmation - minimal/altair/fast_confirmation/reconfirmation/pyspec_tests/reconfirm OK


etc on the new test (nimbus already passed them before the pr)

PR #8513, opened 2026-05-26
partial column movement from EL to ColumnQuarantine

agnxsh:


beacon_chain/el/el_getblobs_service.nim
```diff
@@ -145,28 +158,109 @@ proc attemptGetBlobs*(
...
+              forkyBlck.root, ColumnIndex(columnIndex),
+              partialSidecars[columnIndex])
+
+          var presentCount = 0
+          for item in blobsV3:
+            if item.isSome: inc presentCount
+          debug "Added partial data columns from EL blobpool to quarantine",
+            root = forkyBlck.root,
+            slot = forkyBlck.message.slot,
+            present = presentCount,
```
tersec: Prefer to avoid incurring this cost in the usual case where debug logging isn't enabled, so inline into the logging statement itself, where it will be skipped otherwise:
```nim
            present = blobsV3.countIt(it.isSome),
```
and remove
```nim
          var presentCount = 0
          for item in blobsV3:
            if item.isSome: inc presentCount
```

beacon_chain/el/el_getblobs_service.nim
```diff
@@ -145,28 +158,109 @@ proc attemptGetBlobs*(
...
+          self.recordEngineGetBlobs(forkyBlck.message.slot, hit = false)
+          return
+
+        var allPresent = true
+        for item in blobsV3:
+          if item.isNone:
+            allPresent = false
+            break
+
+        if not allPresent:
```
tersec:
```nim
        if blobsV3.anyIt(it.isNone):
```
then one can remove
```nim
        var allPresent = true
        for item in blobsV3:
          if item.isNone:
            allPresent = false
            break
```
and this slightly clunky indicator variable.

beacon_chain/el/el_getblobs_service.nim
```diff
@@ -145,28 +158,109 @@ proc attemptGetBlobs*(
...
+          self.recordEngineGetBlobs(forkyBlck.message.slot, hit = false)
+
+          let numBlobs = kzg_commitments_count()
+          var
+            blobsOpt = newSeq[Opt[kzg.KzgBlob]](numBlobs)
+            cellProofsOpt = newSeq[Opt[kzg.KzgProof]](
+              numBlobs * fulu_preset.CELLS_PER_EXT_BLOB)
+          for i in 0 ..< numBlobs:
+            if blobsV3[i].isSome:
```
tersec:
```nim
            blobsV3[i].isErrOr:
```
to avoid the manual `get`

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -460,6 +460,81 @@ proc assemble_data_column_sidecars*(
...
+  if kzg_commitments.len == 0 or blobs.len != kzg_commitments.len:
+    return (header, sidecars)
+  if blobs.len > int(MAX_BLOB_COMMITMENTS_PER_BLOCK):
+    return (header, sidecars)
+  if cell_proofs.len != blobs.len * CELLS_PER_EXT_BLOB:
+    return (header, sidecars)
+
+  var cells = newSeq[Opt[CellBytes]](blobs.len)
+  for i in 0 ..< blobs.len:
+    if blobs[i].isSome:
```
tersec:
```nim
    blobs[i].isErrOr:
```
to avoid manual `get`

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -460,6 +460,81 @@ proc assemble_data_column_sidecars*(
...
+    var
+      bitmap: BitArray[int(MAX_BLOB_COMMITMENTS_PER_BLOCK)]
+      partialColumn = newSeqOfCap[KzgCell](blobs.len)
+      partialProofs = newSeqOfCap[KzgProof](blobs.len)
+    for rowIndex in 0 ..< blobs.len:
+      if cells[rowIndex].isNone:
+        continue
+      let proofOpt = cell_proofs[rowIndex * CELLS_PER_EXT_BLOB + columnIndex]
+      if proofOpt.isNone:
+        continue
```
tersec:
```nim
        let proof = cell_proofs[rowIndex * CELLS_PER_EXT_BLOB + columnIndex].valueOr:
          continue
```
(or similar, might have to put parens around `cell_proofs[rowIndex * CELLS_PER_EXT_BLOB + columnIndex]` to avoid manual `get`

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -460,6 +460,81 @@ proc assemble_data_column_sidecars*(
...
+        message: beacon_block_header,
+        signature: signed_beacon_block.signature)
+    inclusion_proof =
+      blck.body.build_proof(KZG_COMMITMENTS_GINDEX).expect("Valid gindex")
+    header = fulu.PartialDataColumnHeader(
+      kzg_commitments: kzg_commitments,
+      signed_block_header: signed_beacon_block_header,
+      kzg_commitments_inclusion_proof: inclusion_proof)
+
+  var sidecars = newSeqOfCap[fulu.PartialDataColumnSidecar](CELLS_PER_EXT_BLOB)
```
tersec: Prefer to allocate this `CELLS_PER_EXT_BLOB`-capacity list only if the 3 early return conditions immediately following aren't met. So they can directly return an, ideally, `const`/`static` instantiation of the empty `seq[fulu.PartialDataColumnSidecar]`

beacon_chain/spec/peerdas_helpers.nim
tersec: There's no reason to ever store/materialize this list in memory. It's only used by
```nim
        partialColumn.add(cells[rowIndex][columnIndex])
```
and that's only for that one row at a time, i.e. it can run `computeCells(...)` on that row in particular, when it needs it, and not store the rest.

tersec: Prefer to move this more expensive initialization after the
```nim
  if blobs.len == 0 or blobs.len > int(MAX_BLOB_COMMITMENTS_PER_BLOCK):
    return sidecars
  if cell_proofs.len != blobs.len * CELLS_PER_EXT_BLOB:
    return sidecars
```
early returns. And if one combines the boolean conditions into a single disjunction, then it's:
```nim
  if blobs.len == 0 or ...:
    return static(default(seq[fulu.PartialDataColumnSidecar]))

  var sidecars = newSeqOfCap[fulu.PartialDataColumnSidecar](CELLS_PER_EXT_BLOB)
```
and only have to write out that type once in the early return section.

beacon_chain/el/el_getblobs_service.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

beacon_chain/el/el_getblobs_service.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

beacon_chain/el/el_getblobs_service.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

tersec: https://github.com/status-im/nimbus-eth2/pull/8513/changes/7943c39f1ac07c5df83ad21013910ed1613afc62

beacon_chain/spec/peerdas_helpers.nim
agnxsh:
```nim
  for rowIndex in 0 ..< blobs.len:
    let blob = blobs[rowIndex].valueOr:
      continue
    let rowCells = computeCells(blob).valueOr:
      continue
    for columnIndex in 0 ..< CELLS_PER_EXT_BLOB:
      let proof = (cell_proofs[rowIndex * CELLS_PER_EXT_BLOB + columnIndex]).valueOr:
        continue
      bitmaps[columnIndex][Natural(rowIndex)] = true
      columns[columnIndex].add(rowCells[columnIndex])
      columnProofs[columnIndex].add(proof)
```
need for that has totally disappeared

PR #8523, opened 2026-05-29
Implement eip 8045

Tomi-3-0:
https://eips.ethereum.org/EIPS/eip-8045

### DIFF
--- a/beacon_chain/spec/validator.nim
+++ b/beacon_chain/spec/validator.nim
@@ -11,7 +11,7 @@
 import
   std/algorithm,
   "."/[crypto, helpers]
-from std/sequtils import mapIt
+from std/sequtils import filterIt, mapIt
 from std/math import `^`
 export helpers
 
@@ -569,8 +569,9 @@ func get_beacon_proposer_indices*(
 
   debugGloasComment "temporary workaround for Gloas"
   when typeof(state).kind >= ConsensusFork.Gloas:
-    let proposers = compute_proposer_indices(state, epoch, seed, indices)
-    proposers.mapIt(Opt.some(it))
+    # [Modified in Gloas:EIP8045]
+    let proposers = compute_proposer_indices(
+      state, epoch, seed, indices.filterIt(not state.validators[it].slashed))
   else:
     compute_proposer_indices(state, epoch, seed, indices)
 
--- a/beacon_chain/spec/validator.nim
+++ b/beacon_chain/spec/validator.nim
@@ -560,6 +560,7 @@ func get_beacon_proposer_index*(
       return res
 
 # https://github.com/ethereum/consensus-specs/blob/v1.6.0-alpha.2/specs/fulu/beacon-chain.md#new-get_beacon_proposer_indices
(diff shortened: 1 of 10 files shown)


tersec:
As a general point, not going to merge this unless it's consistent with the consensus test vectors we're currently using, which it evidently is not, but likely will be for the `glamsterdam-devnet-5` spec version.

beacon_chain/spec/validator.nim
tersec, line 562: When this change is made, it will presumably come with updated an updated `get_beacon_proposer_indices` definition in `consensus-specs`, which should be linked here in addition to the Fulu one to explain why the code that's there is there. Because it hasn't been merged yet to my knowledge, this wouldn't yet be available.

beacon_chain/spec/validator.nim
```diff
@@ -569,7 +569,9 @@ func get_beacon_proposer_indices*(
 
   debugGloasComment "temporary workaround for Gloas"
   when typeof(state).kind >= ConsensusFork.Gloas:
-    let proposers = compute_proposer_indices(state, epoch, seed, indices)
+    # [Modified in Gloas:EIP8045]
+    let proposers = compute_proposer_indices(
+      state, epoch, seed, indices.filterIt(not state.validators[it].slashed))
```
tersec: This adds an avoidable copy of a `seq` of 900k 8-byte integers, so 8MB (plus resulting garbage collection time overhead, et cetera which can be tricky to profile or attribute properly).

https://github.com/status-im/nimbus-eth2/blob/6fb05f36804d53c2e8e014cfeeea8ad7996a5efe/beacon_chain/spec/helpers.nim#L100-L106 already, unfortunately, makes what might be an avoidable copy, but certainly one doesn't have to do this twice. As a minimal, no-regression change, one can instead filter the indices and construct the required `seq` directly rather than make two passes over them, first to active, then to active and not-slashed.

More generally, various other parts of Nimbus play some tricks with using `shuffled_active_validator_indices` and "unshuffling" from https://github.com/status-im/nimbus-eth2/blob/014f59e8901a1200f930482688bb84005d41a80f/beacon_chain/spec/datatypes/base.nim#L416-L420 to avoid copies altogether. That might require more consideration, but at least the basic step already described avoids making things less efficient than they already are.

Tomi-3-0:
> As a general point, not going to merge this unless it's consistent with the consensus test vectors we're currently using, which it evidently is not, but likely will be for the `glamsterdam-devnet-5` spec version.

Agree. I implemented it assuming it would be compatible with the existing test vectors without actually verifying that. The change itself is a net positive regardless, so probably worth keeping around. I willl leave this as a draft pending the spec bump (assuming it gets merged)

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8523/changes/579a5a7a191f792bb99bd66a5a81da8ee602313f

beacon_chain/spec/validator.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8523/commits/d0fcaeb8850d3fea177ddcab35cf307c02a2b60f

PR #8533, opened 2026-06-02
Reconstruction backfiller service

agnxsh:


beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+    return
+
+  # On reconstruction failure `reconstructSlot` returns `Unknown`, leaving the
+  # slot eligible for a later retry.
+  let state =
+    case blockFork
+    of ConsensusFork.Gloas:
+      await self.reconstructSlot(slot, blockRoot, have, gloas.DataColumnSidecar)
+    else:
+      await self.reconstructSlot(slot, blockRoot, have, fulu.DataColumnSidecar)
```
tersec: Heze and newer block forks don't/won't use Fulu columns.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  else:
+    Unknown
+
+proc processSlot(
+    self: ColumnReconstructionBackfillerRef,
+    slot: Slot) {.async: (raises: [CancelledError]).} =
+  let blockFork = self.dag.cfg.consensusForkAtEpoch(slot.epoch)
+  # Data columns currently only exist for Fulu and Gloas; pre-Fulu slots carry
+  # none and later forks (Heze+) don't yet have a column store wired up.
+  if blockFork notin {ConsensusFork.Fulu, ConsensusFork.Gloas}:
```
tersec: Heze columns are identical to Gloas columns.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+
+declareGauge beacon_column_reconstruction_backfill_slot,
+  "Slot most recently selected for reconstruction by the backfiller service"
+
+declareGauge beacon_column_reconstruction_earliest_available_slot,
+  "Earliest slot from which the full column matrix can be served, as extended " &
+  "by the reconstruction backfiller"
+
+const
+  # Spec: a node "SHOULD reconstruct the full data matrix" once it holds
```
tersec: Rather than just asserting the spec says something, link it: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/fulu/das-core.md#reconstruction-and-cross-seeding

tersec: Also, it's not just on mainnet.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+    return
+  let idx = self.slotIdx(bottom - 1)
+  if idx >= 0 and idx < self.slotStates.len and self.slotStates[idx] == TooFew:
+    self.slotStates[idx] = Unknown
+
+proc countExistingColumns(
+    db: BeaconChainDB,
+    consensusFork: ConsensusFork,
+    blockRoot: Eth2Digest
+): HashSet[uint64] =
```
tersec: This and every other `HashSet[uint64]` in this PR should be a `ColumnMap`.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+# Copyright (c) 2026 Status Research & Development GmbH
+# Licensed and distributed under either of
+#   * MIT license (license terms in the root directory or at https://opensource.org/licenses/MIT).
+#   * Apache v2 license (license terms in the root directory or at https://www.apache.org/licenses/LICENSE-2.0).
+# at your option. This file may not be copied, modified, or distributed except according to those terms.
+
+{.push raises: [], gcsafe.}
+
+import
+  std/[sequtils, sets],
```
tersec: Because all of the `HashSet[uint64]`s should be `ColumnMap`s, I don't see the need for `std/sets`.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+    # Columns are stored in a per-fork table, so the right fork must be
+    # consulted: Fulu sidecars live in `fulu_columns`, Gloas in `gloas_columns`.
+    let found =
+      case consensusFork
+      of ConsensusFork.Fulu:
+        db.containsDataColumnSidecar(ConsensusFork.Fulu, blockRoot, i)
+      of ConsensusFork.Gloas:
+        db.containsDataColumnSidecar(ConsensusFork.Gloas, blockRoot, i)
+      else:
+        false
```
tersec: It seems like this entire `case consensusFork` collapses into a single `db.containsDataColumnSidecar(consensusFork, blockRoot, i)`?

https://github.com/status-im/nimbus-eth2/blob/6fb05f36804d53c2e8e014cfeeea8ad7996a5efe/beacon_chain/beacon_chain_db.nim#L1471-L1476 already does what's required.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  ## period — and `eaSlot` — extend further. A genuinely under-filled slot
+  ## simply settles back to `TooFew`, at the cost of one re-count per idle tick.
+  let bottom = self.headContiguousBottom()
+  if bottom <= self.bitmapStartSlot:
+    return
+  let idx = self.slotIdx(bottom - 1)
+  if idx >= 0 and idx < self.slotStates.len and self.slotStates[idx] == TooFew:
+    self.slotStates[idx] = Unknown
+
+proc countExistingColumns(
```
tersec: This function isn't really "count"ing existing columns; it's returning the full set of existing columns.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+import
+  std/[sequtils, sets],
+  chronicles,
+  chronos,
+  taskpools,
+  kzg4844/kzg,
+  metrics,
+  ssz_serialization/[proofs, types],
+  ../spec/[forks, helpers, peerdas_helpers, column_map],
+  ../spec/datatypes/[fulu, gloas],
```
tersec: When importing `forks`, redundant to separately import `spec/datatypes/foo`.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  let count = have.len
+
+  if count.uint64 == NUMBER_OF_COLUMNS:
+    self.markSlot(slot, Full)
+    return
+
+  if count == 0:
+    # Distinguish "block legitimately has zero blobs" from "we never
+    # received any columns" — only the latter is a reconstruction miss.
+    let forked = self.dag.getForkedBlock(bsi.bid).valueOr:
```
tersec, line 305: For pre-finalized blocks which have been pruned from the DAG, this reads some blocks from the database twice, first in https://github.com/status-im/nimbus-eth2/blob/f8262ddfb5a0e4e54b37ae922b8fc5bb6787e11f/beacon_chain/consensus_object_pools/blockchain_dag.nim#L244-L274, then again here with https://github.com/status-im/nimbus-eth2/blob/f8262ddfb5a0e4e54b37ae922b8fc5bb6787e11f/beacon_chain/consensus_object_pools/blockchain_dag.nim#L311-L342

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  # Brief yield between back-to-back reconstructions so the taskpool and
+  # other duties (sig verification, gossip) aren't fully starved.
+  WorkYieldDuration = chronos.milliseconds(50)
+
+type
+  SlotRecon = enum
+    Unknown      ## Not yet inspected by the backfiller.
+    NoColumns    ## Slot has no block, or the block has zero blob commitments.
+    TooFew       ## Block has blobs but <64 columns present — cannot recover.
+    Full         ## All 128 columns are now present (original or reconstructed).
```
tersec: Is there any behavioral difference between `NoColumns` and `Full`? `isFullyServable` appears to be the only consumer which directly checks for either `NoColumns` and `Full`, and it conflates them.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+    self.dag.eaSlot = bottom
+    beacon_column_reconstruction_earliest_available_slot.set(int64(bottom))
+    debug "Extended earliest available slot to reconstructed trail",
+      eaSlot = bottom, head = self.dag.head.slot
+
+proc retryTrailBlocker(self: ColumnReconstructionBackfillerRef) =
+  ## The slot just below the head-contiguous run is what stops `eaSlot` from
+  ## descending. If it is only `TooFew` (has some columns but fewer than the
+  ## `ColumnsRequiredForReconstruction` threshold) re-arm it as `Unknown` so a
+  ## later pass re-counts it: columns arriving via gossip/sync since we last
```
tersec: With 12 second polling, practically, gossip will never arrive with new columns for random old (even, say, 2 or 3 slot old) columns. The window for this to occur basically doesn't exist.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+    let wanted = int(head - self.bitmapStartSlot) + 1
+    if wanted > self.slotStates.len:
+      self.slotStates.setLen(wanted)
+
+  if retentionStart > self.bitmapStartSlot:
+    let drop = int(retentionStart - self.bitmapStartSlot)
+    if drop >= self.slotStates.len:
+      self.slotStates.setLen(0)
+    else:
+      self.slotStates.delete(0 .. drop - 1)
```
tersec: https://nim-lang.org/docs/system.html#delete,seq[T],Natural
> Deletes the item at index `i` by moving all `x[i+1..^1]` items by one position.
>
> This is an `O(n)` operation.

https://nim-lang.org/docs/deques.html is made for this use case.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+      # reconstruction threshold, resuming once it climbs back.
+      if self.validatorCustody.getMap().len < ColumnsRequiredForReconstruction:
+        await sleepAsync(IdleSleepDuration)
+        continue
+
+      self.syncBitmap()
+      self.updateEarliestAvailableSlot()
+
+      let target = self.pickNextSlot()
+      if target.isNone():
```
tersec: This won't progress on "trail blockers" until the entire `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` backfill is complete, even though that does nothing for the ea slot of the node, i.e. the services it advertises and offers to the network. On EF networks, that means 4,096 * 32 == 131,072 slots, and on Gnosis, that requires exhausting 16,384 * 32 = 524,288 slots.

What's the advantage of this approach over one that monotonically proceeds backwards, i.e. what's the advantage of reconstructing slots which will be guaranteed to be behind easlot?

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  # `getBlockIdAtSlot` returns the most recent block at or before `slot`;
+  # if that block is not the proposed block for this slot, the slot is
+  # empty and there is nothing to reconstruct.
+  if not bsi.isProposed():
+    self.markSlot(slot, NoColumns)
+    return
+
+  let blockRoot = bsi.bid.root
+  let have = countExistingColumns(self.dag.db, blockFork, blockRoot)
+  let count = have.len
```
tersec: `lenu64` is already available. Would avoid the separate `count.uint64` conversion and remains compatible with the other users.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  SlotRecon = enum
+    Unknown      ## Not yet inspected by the backfiller.
+    NoColumns    ## Slot has no block, or the block has zero blob commitments.
+    TooFew       ## Block has blobs but <64 columns present — cannot recover.
+    Full         ## All 128 columns are now present (original or reconstructed).
+
+  ColumnReconstructionBackfiller* = object
+    dag: ChainDAGRef
+    validatorCustody: ValidatorCustodyRef
+    beaconClock: BeaconClock
```
tersec: Nothing seems to access this field?

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,451 @@
...
+  # reconstruction. Roughly one slot so we pick up newly arrived columns
+  # without busy-waiting.
+  IdleSleepDuration = chronos.seconds(12)
+
+  # Brief yield between back-to-back reconstructions so the taskpool and
+  # other duties (sig verification, gossip) aren't fully starved.
+  WorkYieldDuration = chronos.milliseconds(50)
+
+type
+  SlotRecon = enum
```
tersec: `{.pure.}` enums are nicer to avoid namespace pollution, especially with generic identifiers such as `Unknown`, `Full`, etc.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/9f81777d096b3905770d89728dc54710c9955412

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+  else:
+    max(headSlot - retentionSlots, fuluStartSlot)
+
+proc syncBitmap(self: ColumnReconstructionBackfillerRef) =
+  ## Resize and slide the bitmap so it spans the current
+  ## [retentionStart..head] window. New top entries are `Unknown`; entries
+  ## that fall behind the retention horizon are dropped. A `Deque` is used so
+  ## both ends are amortised O(1): growth appends at the tail (head side) while
+  ## retention drops a prefix from the front, the latter being O(dropped) rather
+  ## than the O(len) of a `seq.delete`.
```
tersec: The https://en.wikipedia.org/wiki/Deque is a well-known CS concept. Don't really need an explanation of them here, and " rather ... than the O(len) of a `seq.delete`" now refers to code which doesn't exist, and from the perspective of anything but this own branch/PR's commit history, never existed.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+  present
+
+proc loadExistingColumns[T: fulu.DataColumnSidecar | gloas.DataColumnSidecar](
+    db: BeaconChainDB,
+    blockRoot: Eth2Digest,
+    indices: ColumnMap
+): seq[ref T] =
+  var columns = newSeqOfCap[ref T](indices.len)
+  for i in indices:
+    let colData = new T
```
tersec: Hoist this memory allocation outside the loop.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+    headSlot = self.dag.head.slot
+    retentionEpochs = self.dag.cfg.MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS
+    retentionSlots = retentionEpochs * SLOTS_PER_EPOCH
+    fuluStartSlot = self.dag.cfg.FULU_FORK_EPOCH.start_slot()
+  if uint64(headSlot) <= retentionSlots:
+    fuluStartSlot
+  else:
+    max(headSlot - retentionSlots, fuluStartSlot)
+
+proc syncBitmap(self: ColumnReconstructionBackfillerRef) =
```
tersec: `func`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+    if drop >= self.slotStates.len:
+      self.slotStates.clear()
+    else:
+      self.slotStates.shrink(fromFirst = drop)
+    self.bitmapStartSlot = retentionStart
+
+func slotIdx(self: ColumnReconstructionBackfillerRef, slot: Slot): int =
+  int(slot - self.bitmapStartSlot)
+
+proc markSlot(
```
tersec: `func`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+  if self.slotStates.len == 0:
+    return
+  let bottom = self.headContiguousBottom()
+  if bottom < self.dag.eaSlot:
+    self.dag.eaSlot = bottom
+    beacon_column_reconstruction_earliest_available_slot.set(int64(bottom))
+    debug "Extended earliest available slot to reconstructed trail",
+      eaSlot = bottom, head = self.dag.head.slot
+
+proc onColumnsStored*(
```
tersec: `func`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+    of ConsensusFork.Gloas:
+      # `columnFork` folds Gloas and every newer fork that shares the Gloas
+      # column format into this branch.
+      debugHezeComment "confirm Heze data columns stay identical to Gloas"
+      await self.reconstructSlot(slot, blockRoot, have, gloas.DataColumnSidecar)
+    else:
+      await self.reconstructSlot(slot, blockRoot, have, fulu.DataColumnSidecar)
+  self.markSlot(slot, state)
+
+proc frontierSlot(self: ColumnReconstructionBackfillerRef): Opt[Slot] =
```
tersec: `func`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+    taskpool: Taskpool
+    # Sliding bitmap over [bitmapStartSlot..head]. Each entry is the
+    # backfiller's last-known reconstruction status for that slot. Kept in
+    # memory only — on restart we rebuild via a fresh backward scan.
+    slotStates: Deque[SlotRecon]
+    bitmapStartSlot: Slot
+
+  ColumnReconstructionBackfillerRef* = ref ColumnReconstructionBackfiller
+
+proc new*(
```
tersec: `func`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+  chronicles,
+  chronos,
+  taskpools,
+  kzg4844/kzg,
+  metrics,
+  ssz_serialization/[proofs, types],
+  ../spec/[forks, helpers, peerdas_helpers, column_map],
+  ../sync/validator_custody,
+  ../beacon_chain_db,
+  ../beacon_clock,
```
tersec: Not needed since `BeaconClock` usage was removed.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,464 @@
...
+{.push raises: [], gcsafe.}
+
+import
+  std/[deques, sequtils],
+  chronicles,
+  chronos,
+  taskpools,
+  kzg4844/kzg,
+  metrics,
+  ssz_serialization/[proofs, types],
```
tersec: How is `ssz_serialization/proofs` relevant or useful to this module?

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/e71b6b30e47866f5e806ec08f2d5dce6060a6193

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,462 @@
...
+  ## Run the parallel cell+proof recovery and persist every column the
+  ## caller did not already have. Returns true on full success.
+  beacon_column_reconstruction_attempts_total.inc()
+  beacon_column_reconstruction_backfill_slot.set(int64(slot))
+
+  let startTime = Moment.now()
+  # `recover_cells_and_proofs_parallel` is concrete over Fulu sidecars (its
+  # taskpool `spawn` can't be instantiated generically). Recovery only reads
+  # `index`/`column`, so adapt Gloas sidecars to that shape; Fulu ones pass
+  # through unchanged.
```
tersec: This is a "generics sandwich":
```nim
diff --git a/beacon_chain/spec/peerdas_helpers.nim b/beacon_chain/spec/peerdas_helpers.nim
index d9225d8c..22242b6e 100644
--- a/beacon_chain/spec/peerdas_helpers.nim
+++ b/beacon_chain/spec/peerdas_helpers.nim
@@ -24,6 +24,9 @@ from std/algorithm import sort
 from std/sequtils import anyIt, mapIt, repeat, toSeq
 from stew/staticfor import staticFor
 
+# Generics sandwich via `recover_cells_and_proofs_parallel`
+export threadsync
+
 type
   CellBytes = array[fulu.CELLS_PER_EXT_BLOB, Cell]
   ProofBytes = array[fulu.CELLS_PER_EXT_BLOB, KzgProof]
@@ -157,7 +160,7 @@ proc recover_matrix*(partial_matrix: seq[MatrixEntry],
 
 proc recover_cells_and_proofs_parallel*(
     tp: Taskpool,
-    dataColumns: seq[ref fulu.DataColumnSidecar]):
+    dataColumns: seq[ref fulu.DataColumnSidecar] | seq[ref gloas.DataColumnSidecar]):
     Future[Result[seq[CellsAndProofs], cstring]] {.async: (raises: []).} =
   ## Recover blobs from data column sidecars in parallel.
   ## Only the `index`/`column` fields are consumed, which both the Fulu and
```
works, because it ensures the relevant `chronos/threadsync` symbols are in scope at the callsite, where due to generics, it gets instantiated.

See also, e.g.
https://github.com/status-im/nimbus-eth2/blob/0ee3b53911d4e34122b4a4569f8f86042184f1af/beacon_chain/spec/helpers.nim#L22-L27

Related to:
- https://github.com/nim-lang/Nim/issues/11225

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/f2f250b909907be2defe6e64ea84827f45f1bf1d

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,447 @@
...
+    return
+
+  if count < ColumnsRequiredForReconstruction:
+    self.markSlot(slot, SlotRecon.TooFew)
+    return
+
+  # On reconstruction failure `reconstructSlot` returns `Unknown`, leaving the
+  # slot eligible for a later retry.
+  let state =
+    case columnFork
```
tersec: Why is this a `case` rather than `if`/`else`?

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/f2f250b909907be2defe6e64ea84827f45f1bf1d

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,447 @@
...
+    let forked = self.dag.getForkedBlock(bsi.bid).valueOr:
+      self.markSlot(slot, SlotRecon.Servable)
+      return
+    var blockHadBlobs = false
+    withBlck(forked):
+      when consensusFork >= ConsensusFork.Gloas:
+        # [Modified in Gloas:EIP7732] commitments live under the payload bid.
+        blockHadBlobs = forkyBlck.message.body
+          .signed_execution_payload_bid.message.blob_kzg_commitments.len > 0
+      elif consensusFork >= ConsensusFork.Fulu:
```
tersec: There aren't any forks between Fulu and Gloas. This `>=` is just `==`.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/dd61958f1c9686c1d76edc7849c8e80be118358f

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/dd61958f1c9686c1d76edc7849c8e80be118358f

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+  ## Run the parallel cell+proof recovery and persist every column the
+  ## caller did not already have. Returns true on full success.
+  beacon_column_reconstruction_attempts_total.inc()
+  beacon_column_reconstruction_backfill_slot.set(int64(slot))
+
+  let
+    startTime = Moment.now()
+    recovered = (await recover_cells_and_proofs_parallel(
+      self.taskpool, columns)).valueOr:
+    beacon_column_reconstruction_failures_total.inc()
```
tersec, line 203:
```
[2026-06-18T23:35:56.353Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-8533@2/beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim(247, 5) Error: invalid indentation
```

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/a03f966cebea4bfeb139965230eff9cfae0b6c06

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+    slot: Slot) {.async: (raises: [CancelledError]).} =
+  let blockFork = self.dag.cfg.consensusForkAtEpoch(slot.epoch)
+  # Data columns only exist from Fulu onward; pre-Fulu slots carry none.
+  if blockFork notin {ConsensusFork.Fulu, ConsensusFork.Gloas, ConsensusFork.Heze}:
+    self.markSlot(slot, SlotRecon.Servable)
+    return
+
+  # Gloas and every newer fork (Heze, ...) share the Gloas data column format
+  let columnFork =
+    if blockFork >= ConsensusFork.Gloas: ConsensusFork.Gloas else: blockFork
```
tersec: There's only one pre-Gloas fork with columns. It's Fulu.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/5e19ef44048d94c33965b8223d8e2fb8d1086efe

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/f2f250b909907be2defe6e64ea84827f45f1bf1d

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+              ## The backfiller treats both the same: nothing left to do.
+
+  ColumnReconstructionBackfiller* = object
+    dag: ChainDAGRef
+    validatorCustody: ValidatorCustodyRef
+    taskpool: Taskpool
+    # Sliding bitmap over [bitmapStartSlot..head]. Each entry is the
+    # backfiller's last-known reconstruction status for that slot. Kept in
+    # memory only — on restart we rebuild via a fresh backward scan.
+    slotStates: Deque[SlotRecon]
```
tersec: How does this handle reorgs?

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+    retentionStart = self.retentionStartSlot()
+    head = self.dag.head.slot
+
+  if self.bitmapStartSlot == FAR_FUTURE_SLOT:
+    self.bitmapStartSlot = retentionStart
+    if head >= retentionStart:
+      let initialLen = int(head - retentionStart) + 1
+      self.slotStates = initDeque[SlotRecon](initialLen)
+      for _ in 0 ..< initialLen:
+        self.slotStates.addLast(SlotRecon.Unknown)
```
tersec: Since it monotonically moves backwards along with `frontierSlot(...)`, will it tend to use even a small fraction of this `initialLen` in a meaningful way? It gets pre-filled here with `SlotRecon.Unknown`, but in general, anything preceding the `frontierSlot(...)` will by construction be effectively `SlotRecon.Unknown`, so if a slot status before `self.bitmapStartSlot` is queried, that can default to `SlotRecon.Unknown` without storing them all explicitly.

Since it tracks `self.bitmapStartSlot` already, it doesn't have to pre-fill for some alignment reason, because
```nim
func slotIdx(self: ColumnReconstructionBackfillerRef, slot: Slot): int =
  int(slot - self.bitmapStartSlot)
```
will use the correct offset regardless.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+  if slot < self.bitmapStartSlot:
+    return
+  let idx = self.slotIdx(slot)
+  if idx >= 0 and idx < self.slotStates.len:
+    self.slotStates[idx] = state
+
+func headContiguousBottom(self: ColumnReconstructionBackfillerRef): Slot =
+  ## Lowest slot of the unbroken run of fully-servable slots ending at the
+  ## head. The continuous "we hold all NUMBER_OF_COLUMNS columns" period is
+  ## therefore `[result..head]`. Returns `head + 1` when even the head isn't
```
tersec: `result`?

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+  ## downward we lower `eaSlot` so peers learn of the longer servable history.
+  ## We anchor the walk *at the head* (not at the current `eaSlot`) so we never
+  ## extend the advertised window across a still-unfilled gap, and we only ever
+  ## lower it — a momentarily un-inspected tip, or columns still arriving for
+  ## the current slot, must not retract history we have already guaranteed.
+  if self.slotStates.len == 0:
+    return
+  let bottom = self.headContiguousBottom()
+  if bottom < self.dag.eaSlot:
+    self.dag.eaSlot = bottom
```
tersec: Once this happens, there's no need to hold statuses of any slots newer than `self.dag.eaSlot` on the chain the node believes canonical.

One nuance here involved reorgs, i.e. post-finalized blocks ranges can change slot to block mappings, so the cached database states become invalidated (or should; this PR doesn't do that). At that point, any cached backfill statuses of finalized slots by definition aren't invalidated, but some range of post-finalized blocks might be.

While I'd note that this does introduce a kind of discontiguity, it's:
1. one that the previous version of this PR didn't handle anyway; and more importantly
2. specific, bounded, and non-arbitrary in its structure: the finalized slots/blocks never invalidate, but rather invalidation would happen up to the most recent common ancestor of the previous head and the new head, before which the ancestors' cache statuses remain correct.

So it still doesn't need the block-by-block bitmap approach, since there are specifically 7 regions to consider, from least to most recent; some of these ranges might be empty/0-length, for example:
- (a) blocks for slots before the required sidecar retention window. Not covered by this PR, and not required or reasonable by default for backfill;
- (b) finalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (c) finalized blocks within the column retention window but for which the column set has been reconstructed;
- (d) nonfinalized blocks _outside_ the column retention window. This will typically be empty and arguably means the spec has hit a fatal edge case. Reasonable enough to ignore here;
- (e) nonfinalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (f) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, before this common ancestor across a reorg (i.e. not invalidated); and
- (g) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, after this common ancestor across a reorg (i.e. invalidated)

Ignoring (a) and (d), because they're listed essentially for completeness but by definition don't participate in a reconstruction backfiller:
- (b) finalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (c) finalized blocks within the column retention window but for which the column set has been reconstructed;
- (e) nonfinalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (f) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, before this common ancestor across a reorg (i.e. not invalidated); and
- (g) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, after this common ancestor across a reorg (i.e. invalidated)

Which is 5 regions to consider. Each is internally contiguous, i.e. in each case, either all or none of the blocks has been reconstructed. Furthermore, the boundary between (b) and (c) only ever monotonically goes backwards in time.

Further splitting these two cases, either the backfill has reached the finalized area or not. If it _has reached the finalized blocks_, the remaining regions are
- (b) finalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (c) finalized blocks within the column retention window but for which the column set has been reconstructed;
- (f) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, before this common ancestor across a reorg (i.e. not invalidated); and
- (g) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, after this common ancestor across a reorg (i.e. invalidated)

If it _has not reached the finalized blocks_, what remains are:
- (b) finalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (e) nonfinalized blocks within the column retention window but for which the column set hasn't been reconstructed;
- (f) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, before this common ancestor across a reorg (i.e. not invalidated); and
- (g) nonfinalized blocks within the column retention window and for which the column set has been reconstructed, after this common ancestor across a reorg (i.e. invalidated)

`dag.eaSlot` does not necessarily track the reconstructed vs nonreconstructed boundary between either (b)/(c) or (e)/(f), depending on scenario, because when the node switches to `Limited` mode if it loses sync at head, `dag.eaSlot` is allowed deliberately to slip back into something close to head. But an internal, `eaSlot`-like variable could track this similarly, waiting for the node to return to non-`Limited` capacity. From this perspective, what separates the has vs has not reached finalized blocks scenario can be reconstructed by comparing the finalized checkpoint with this internal-quasi-`eaSlot`.

The (c)/(f) boundary in the case where reconstruction backfill _has_ reached the finalized blocks doesn't have to be tracked explicitly. It effectively functions as a single coherent region for any given reorg.

The next boundary to track is the one between (f)/(g) region, which is calculated anew for each reorg based on the common ancestor. This should function as a stopping point at which point it skips over to the internal-quasi-`eaSlot` boundary.

So, that's two persistent single slot numbers, one of which effectively mirrors a kind of shadow semantics of `eaSlot` stored instead of a sliding-window bitmap which doesn't anyway work with reorgs at all right now, to also correctly handle reorgs.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,436 @@
...
+  ## threshold, so mark it `Unknown` for the next pass to re-count (and, if it
+  ## was the trail-blocker, let `eaSlot` extend further). Slots in any other
+  ## state are left untouched — `Servable` slots need no redo and `Unknown`
+  ## is already queued.
+  if slot < self.bitmapStartSlot:
+    return
+  let idx = self.slotIdx(slot)
+  if idx >= 0 and idx < self.slotStates.len and
+      self.slotStates[idx] == SlotRecon.TooFew:
+    self.slotStates[idx] = SlotRecon.Unknown
```
tersec: The purpose of this seems to be to allow reconsideration of arbitrary slots in the history, but per
- https://github.com/status-im/nimbus-eth2/pull/8533#discussion_r3441090353

and the usage of `frontierSlot` this is at any given time exactly one slot in the history that the reconstruction backfiller would consider reconstructing. Before what it can provide as the `eaSlot` and it can't advertise to the network that it's sharing it back out anyway, and clients won't tend to request it.

So I'm not sure why it needs to specifically track the slate slots to trigger reconsideration this way: in the current implementation which doesn't handle reorgs, this monotonically goes backwards, so can be just one slot number. The `self.slotStates` abstraction doesn't handle reorgs particularly well anyway, so if one wants to remain in the reorg-invalidates-all-or-nothing regime, it's at least as easy either way.

But while reorgs make the backwards/forwards movement slightly more nuanced (one possibility: reorg means reset the frontier to head, but per
- https://github.com/status-im/nimbus-eth2/pull/8533#discussion_r3441090353

skip the
> nonfinalized blocks within the column retention window and for which the column set has been reconstructed, before this common ancestor across a reorg (i.e. not invalidated)

So those aren't redundantly checked and over to the not-yet reconstructed ones, either:
> (b) finalized blocks within the column retention window but for which the column set hasn't been reconstructed;

or
> (e) nonfinalized blocks within the column retention window but for which the column set hasn't been reconstructed;

And continue with the `frontierSlot` from there.

Or just check the database redundantly and don't add the complexity to skip them. A tradeoff.

But all of this is, again, to handle reorg scenarios the current changes don't handle at all.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,417 @@
...
+
+proc processSlot(
+    self: ColumnReconstructionBackfillerRef,
+    slot: Slot): Future[SlotRecon] {.async: (raises: [CancelledError]).} =
+  ## Resolve, and where possible achieve, the servable state of `slot` on the
+  ## *current* canonical chain. The canonical block is read fresh each call, so
+  ## a reorg that re-mapped this slot is handled implicitly.
+  let blockFork = self.dag.cfg.consensusForkAtEpoch(slot.epoch)
+  # Data columns only exist from Fulu onward; pre-Fulu slots carry none.
+  if blockFork notin {ConsensusFork.Fulu, ConsensusFork.Gloas, ConsensusFork.Heze}:
```
tersec: More generally, `if blockFork < ConsensusFork.Fulu:`, unless one wants to bet on the I* fork not having any kind of reconstructable columns. It's possible, I suppose, but then the current approach still silently fails when it starts appearing in the `nimbus-eth2` codebase, just by advertising that there's nothing to reconstruct.

So, `if blockFork < ConsensusFork.Fulu:` or similar, and if one wants to be really, really sure about checking the I* fork before allowing it here, add the usual `static: doAssert high(ConsensusFork) == ConsensusFork.Heze` to ensure that this particular part of the code is checked during that transition.

But fail-open is at least as bad here as fail-closed.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -243,6 +249,14 @@ proc storeSidecars(
 ) =
   if sidecarsOpt.isSome():
```
tersec: Replacing this with `sidecarsOpt.isErrOr: ...` avoids having to repeatedly do `sidecarsOpt[]`.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,417 @@
...
+  else:
+    max(headSlot - retentionSlots, fuluStartSlot)
+
+proc updateEarliestAvailableSlot(self: ColumnReconstructionBackfillerRef) =
+  ## Lower `dag.eaSlot` to `runBottom`, the durable bottom of the servable run,
+  ## so peers learn of the longer history. We only ever lower it here: other
+  ## writers (prune, custody) own raising it. During a reorg refill
+  ## `[runBottom..head]` is briefly not fully head-anchored again; we accept
+  ## that transient over-advertisement (a peer sees a miss, never wrong data)
+  ## rather than retract history we have already guaranteed.
```
tersec: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.11/specs/fulu/p2p-interface.md#status-v2
> If the node is able to serve all blocks throughout the entire sidecars retention period (as defined by both `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` and `MIN_EPOCHS_FOR_DATA_COLUMN_SIDECARS_REQUESTS`), but is NOT able to serve all sidecars during this period, it should advertise the earliest slot from which it can serve all sidecars.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/38e993bce98629039d869aaf87501f696eba529d

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/38e993bce98629039d869aaf87501f696eba529d

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,434 @@
...
+    consensusFork: ConsensusFork,
+    blockRoot: Eth2Digest
+): ColumnMap =
+  ## The set of columns currently persisted for `blockRoot`. Columns are stored
+  ## in a per-fork table keyed by fork, which `containsDataColumnSidecar`
+  ## already dispatches on, so no per-fork branching is needed here.
+  var present: ColumnMap
+  for i in 0'u64 ..< NUMBER_OF_COLUMNS.uint64:
+    if db.containsDataColumnSidecar(consensusFork, blockRoot, i):
+      present.incl(i)
```
tersec, line 167: If the remaining columns can't possibly provide `ColumnsRequiredForReconstruction` columns, even if every remaining column is present, then an early exit from this loop to avoid pointless database lookups. The outcome will be the same, too few columns.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -241,8 +247,16 @@ proc storeSidecars(
     self: BlockProcessor,
     sidecarsOpt: Opt[fulu.DataColumnSidecars] | Opt[gloas.DataColumnSidecars]
 ) =
-  if sidecarsOpt.isSome():
-    self.consensusManager.dag.db.putDataColumnSidecars(sidecarsOpt[])
+  sidecarsOpt.isErrOr:
+    self.consensusManager.dag.db.putDataColumnSidecars(value)
+    if self.onDataColumnsStored != nil and value.len > 0:
+      let slot =
+        when sidecarsOpt is Opt[gloas.DataColumnSidecars]:
```
tersec: Alternatively, `value is gloas.DataColumnSidecars` to be consistent with other usage of `value` here, and it's clearer why the body of the `when` statement depends on the `when` statement.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -156,13 +156,19 @@ proc existingColumns(
...
+  ## `ColumnsRequiredForReconstruction` columns; below that the outcome is "too
+  ## few" regardless of the exact set, and the caller only checks the count.
+  var
+    present: ColumnMap
+    remaining = NUMBER_OF_COLUMNS  # columns not yet looked up
   for i in 0'u64 ..< NUMBER_OF_COLUMNS.uint64:
     if db.containsDataColumnSidecar(consensusFork, blockRoot, i):
       present.incl(i)
+    dec remaining
+    if present.len + remaining < ColumnsRequiredForReconstruction:
```
tersec, line 169: `remaining` is a pure function of `NUMBER_OF_COLUMNS - i`, no need to have basically two separate loop counters.

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/commits/fab418ffd9859faa95f57f0444f9293018233f27

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/8ddba724d28441f8d5a2eac78af9156ce86493d0

beacon_chain/gossip_processing/block_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/8ddba724d28441f8d5a2eac78af9156ce86493d0

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/10c7affc07c01b4417250b4c4f7852549b5f32f6

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,439 @@
...
+    finalized = self.dag.finalizedHead.slot
+    retentionStart = self.retentionStartSlot()
+
+  self.reconcileHead(head, self.dag.head.root, finalized)
+
+  # Follow the advancing retention floor: pruned slots can be neither served nor
+  # reconstructed, so drop run/frontier that fall behind it.
+  if self.runBottom != FAR_FUTURE_SLOT and self.runBottom < retentionStart:
+    self.runBottom = retentionStart
+  if self.frontier != FAR_FUTURE_SLOT and self.frontier < retentionStart:
```
tersec: `self.frontier < retentionStart` implies `self.frontier != FAR_FUTURE_SLOT`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,439 @@
...
+  let
+    head = self.dag.head.slot
+    finalized = self.dag.finalizedHead.slot
+    retentionStart = self.retentionStartSlot()
+
+  self.reconcileHead(head, self.dag.head.root, finalized)
+
+  # Follow the advancing retention floor: pruned slots can be neither served nor
+  # reconstructed, so drop run/frontier that fall behind it.
+  if self.runBottom != FAR_FUTURE_SLOT and self.runBottom < retentionStart:
```
tersec: `self.runBottom < retentionStart` implies `self.runBottom != FAR_FUTURE_SLOT`

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,439 @@
...
+      self.runTop = head
+      self.runBottom = self.frontier
+    elif self.frontier > self.runTop:
+      discard  # gap-fill above the island; merges on a later pass
+    else:
+      # extend the run into history
+      self.runBottom = self.frontier
+    self.frontier =
+      if self.frontier <= retentionStart: FAR_FUTURE_SLOT
+      else: self.frontier - 1
```
tersec: These 3 lines of code adjusting `self.frontier` occur twice, verbatim, in this function. Maybe extract to a small template defined within `step` to make it more obvious it's the same, well, `step`ing it's doing.

tests/test_reconstruction_backfiller.nim
```diff
@@ -0,0 +1,270 @@
...
+
+func outcome(chain: DummyBeaconChain, slot: Slot): SlotRecon =
+  if slot in chain.tooFew: SlotRecon.TooFew else: SlotRecon.Servable
+
+proc tick(c: var BackfillCursors, chain: var DummyBeaconChain): bool =
+  ## One pass, mirroring the service's `step`: reconcile the head, pick the next
+  ## frontier, "reconstruct" it synthetically, fold the outcome back. Returns
+  ## false when the cursors are idle (nothing to do).
+  c.reconcileHead(chain.head, chain.finalized, chain.headRoot, chain.extension)
+  let target = c.nextFrontier(chain.head, chain.finalized, chain.retentionStart)
```
tersec:
```nim
  let target = c.nextFrontier(...).valueOr:
    return false
```
and can avoid the `target.get`s

beacon_chain/spec/peerdas_helpers.nim
```diff
@@ -157,9 +164,14 @@ proc recover_matrix*(partial_matrix: seq[MatrixEntry],
...
     tp: Taskpool,
-    dataColumns: seq[ref fulu.DataColumnSidecar]):
+    dataColumns: seq[ref fulu.DataColumnSidecar] |
+                 seq[ref gloas.DataColumnSidecar]):
     Future[Result[seq[CellsAndProofs], cstring]] {.async: (raises: []).} =
   ## Recover blobs from data column sidecars in parallel.
+  ## Only the `index`/`column` fields are consumed, which both the Fulu and
+  ## Gloas sidecar layouts share, so the input is a type class over either; the
+  ## proc thus instantiates concretely per fork at each callsite, sidestepping
+  ## the taskpool `spawn` below not being instantiable from a generic context.
```
tersec: > so the input is a type class over either; the proc thus instantiates concretely per fork at each callsite

Yes, that's a literal reading of   `dataColumns: seq[ref fulu.DataColumnSidecar] | seq[ref gloas.DataColumnSidecar]` but one should be able to safely assume someone reading this code understands how basic Nim features work and as such Nim features don't have to be re-explained whenever they're used, any more than commenting that "i += 1` increments `i` makes sense as a comment.

>  sidestepping the taskpool `spawn` below not being instantiable from a generic context

Not true, self-evidently. This comment contradicts itself within a single sentence.

tests/test_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/63548d9ebb7da8bf76251709e81cf1db2c5c6efa

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/82b7117cafea35c78877defefe72b05e9e020a2a

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/82b7117cafea35c78877defefe72b05e9e020a2a

beacon_chain/consensus_object_pools/column_reconstruction_backfiller.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/82b7117cafea35c78877defefe72b05e9e020a2a

beacon_chain/spec/peerdas_helpers.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8533/changes/1665e8aafdfd8126479526ae1c52589d8436629e

PR #8535, opened 2026-06-02
nix: use TMPDIR for cache locations

jakubgs:
Hardcoding `/tmp` is fine for Linux but on macOS hosts due to lackluster sandboxing weird errors due to corrupted Nim cache can appear.
```
Error: cannot open '/tmp/nim/koch_d/@psystem@sexceptions.nim.c'
```
Updating NBS to include the same fix for Nim build.
- https://github.com/status-im/nimbus-build-system/pull/126

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -53,14 +53,8 @@ in stdenv.mkDerivation rec {
 
   enableParallelBuilding = true;
 
-  env = {
-    # Disable CPU optimizations that make binary not portable.
-    NIMFLAGS = "-d:disableMarchNative -d:git_revision_override=${revision}";
-    # Avoid errors about missing user home.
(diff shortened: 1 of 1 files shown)


etan-status:
maybe ${TMPDIR:-/tmp} so that it still works when it isn't set?

jakubgs:
Nix interpreter sets it in the sadbox itself and it is documented:

- https://github.com/NixOS/nix/blob/master/doc/manual/source/command-ref/env-common.md#env-TMPDIR
- https://github.com/NixOS/nix/blob/cbbc07c6b04f5ecfae73f1504aaa7a4bdddbd84c/src/libstore/darwin/build/sandbox-defaults.sb#L3

PR #8540, opened 2026-06-03
feat(ci): auto-select nix flake target for manual, multibranch, and nightly builds

markoburcul:
Makes the nimbus-eth2 nix CI pipeline pick the correct Nix flake target(`beacon_node` vs `validator_client`) automatically, based on which job is running, with an optional manual override.

Referenced issue:
* https://github.com/status-im/infra-nimbus/issues/276

### DIFF
--- a/ci/nix.Jenkinsfile
+++ b/ci/nix.Jenkinsfile
@@ -28,8 +28,8 @@ pipeline {
     )
     choice(
       name: 'NIX_TARGET',
-      description: 'Nix flake target to build',
-      choices: ['beacon_node', 'validator_client']
+      description: 'Flake target to build. "auto" derives it from the job name, pick a target to force it.',
+      choices: ['auto', 'beacon_node', 'validator_client']
     )
   }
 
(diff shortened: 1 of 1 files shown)


tersec (approved):

ci/nix.Jenkinsfile
```diff
@@ -100,3 +100,15 @@ pipeline {
...
+def resolveTarget() {
+  return (params.NIX_TARGET in [null, '', 'auto']) ? nixTarget() : params.NIX_TARGET
+}
+
+def nixTarget() {
+  /* Dev CI is multibranch (BRANCH_NAME set) and only ever builds the beacon node.*/
+  if (env.BRANCH_NAME) {
+    return 'beacon_node'
+  }
+  return env.JOB_NAME.toLowerCase().contains('validator-client') ? 'validator_client' : 'beacon_node'
```
jakubgs, line 113: We could make it more generic and support multiple targets:
```suggestion
    def validTargets = ['beacon_node', 'validator_client', 'signing_node', 'ncli', 'ncli_db']
    /* Check if the job path contains any of the valid targets. */
    validTargets.each {
        if (env.JOB_NAME.split('/').contains(it)) { return it }
    }
```
Kinda like we do in `jenkins.getAgentLabelFromJob()`.

jakubgs (approved):

PR #8543, opened 2026-06-03
complete EMPTY/FULL sibling support in proto_array

Tomi-3-0:
Follow-up to #8514.
This completes head selection and pruning for the EMPTY/FULL sibling nodes, matching alpha.9.

- Prune: rebuild the node array instead of chopping off the front. A block's FULL node is added whenever its payload arrives, which can be late and land after the finalized block. The old front-chop then kept that stray FULL node but deleted its index entry, crashing the `indices.len + fullBlockIndices.len == nodes.len` check.

- findHead: if a justified block's EMPTY node has no children because the chain continued on its FULL node, follow the FULL node's best descendant so head selection doesn't get stuck.

 - Tiebreak: choose between a block's EMPTY and FULL node by weight first, then by status (FULL beats EMPTY), matching [alpha.9](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.9/specs/gloas/fork-choice.md#modified-get_head), replacing the slot-based rule in #8514.
 
- applyScoreChanges takes an emptyPreferredRoot input that says when to prefer EMPTY for a block. The code that computes it (from PTC votes) lands in a later PR.

### DIFF
--- a/beacon_chain/fork_choice/proto_array.nim
+++ b/beacon_chain/fork_choice/proto_array.nim
@@ -186,7 +186,8 @@ func applyScoreChanges*(
     currentSlot: Slot,
     checkpoints: FinalityCheckpoints,
     justifiedTotalActiveBalance: Gwei,
-    proposerBoostRoot: Eth2Digest): FcResult[void] =
+    proposerBoostRoot: Eth2Digest,
+    emptyPreferredRoot: Eth2Digest = ZERO_HASH): FcResult[void] =
   ## Iterate backwards through the array, touching all nodes and their parents
   ## and potentially the best-child of each parent.
   #
@@ -216,6 +217,7 @@ func applyScoreChanges*(
     checkpoints.finalized.root == self.checkpoints.finalized.root or
     self.checkpoints.finalized.epoch == GENESIS_EPOCH
   self.checkpoints = checkpoints
+  self.emptyPreferredRoot = emptyPreferredRoot
 
   ## Alias
   # This cannot raise the IndexError exception, how to tell compiler?
@@ -384,9 +386,6 @@ func onBlock*(
 
 func onPayloadVerified*(
     self: var ProtoArray, root: Eth2Digest): FcResult[void] =
-  ## Register the FULL payload variant of a block once its execution payload
-  ## has been verified, as a sibling of the EMPTY/PENDING node sharing the same
-  ## parent. Subsequent blocks that build on the full parent attach here.
   if root in self.fullBlockIndices:
     return ok()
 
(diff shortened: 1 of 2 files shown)


beacon_chain/fork_choice/proto_array.nim
```diff
@@ -459,6 +458,15 @@ func findHead*(self: var ProtoArray, head: var Eth2Digest): FcResult[void] =
   head = bestNode.bid.root
   ok()
 
+func remapIdx(idx: Opt[Index], oldToNew: Table[Index, Index]): Opt[Index] =
+  # Remap a logical node index through the prune survivor table; drops it if
+  # the referenced node did not survive.
+  if idx.isSome:
```
tersec: `isErrOr` + `value`

beacon_chain/fork_choice/proto_array.nim
```diff
@@ -186,7 +186,8 @@ func applyScoreChanges*(
     currentSlot: Slot,
     checkpoints: FinalityCheckpoints,
     justifiedTotalActiveBalance: Gwei,
-    proposerBoostRoot: Eth2Digest): FcResult[void] =
+    proposerBoostRoot: Eth2Digest,
+    emptyPreferredRoot: Eth2Digest = ZERO_HASH): FcResult[void] =
```
tersec, line 190: Does anything ever call this with a non-default value?

Tomi-3-0: No. not in this PR. it's always `ZERO_HASH` here.
  
`emptyPreferredRoot` would eventually become the consumer half of the previous slot EMPTY/FULL tiebreak. `maybeUpdateBestChildAndDescendant` reads it so that, for a block at `currentSlot - 1`, its `EMPTY` node is preferred over its FULL node unless the payload should be extended. It's how [`should_extend_payload`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-should_extend_payload)'s decision reaches the proto-array.

 The producer will be the  `find_head` caller that computes it from the PTC votes and the proposer-boost root, as per [`get_payload_status_tiebreaker`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-get_payload_status_tiebreaker)/should_extend_payload and depends on the ptcVotes machinery, which lands in a follow-up PR.
With the default `ZERO_HASH` value here the tiebreak can stay a no-op

beacon_chain/fork_choice/proto_array.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8543/changes/b859c25d7ea164d03a2cbd32d824786a4af1eea4

PR #8544, opened 2026-06-03
route and validate gloas attestation votes in fork choice

Tomi-3-0:
Makes attestation votes actually weight a block's EMPTY vs FULL node until now every vote landed on EMPTY because payload_present was never set and validates gloas attestations, matching alpha.9.

- process_attestation records the vote's payload preference, and compute_deltas routes the vote to the FULL node when it's a payload-present vote and that node exists, else EMPTY.
- on_attestation and process_block read the payload vote from `attestation.data.index` and thread it through.
- validate_on_attestation: data.index must be 0 or 1; a same-slot vote must be EMPTY; a full vote requires the payload to be known.

### DIFF
--- a/beacon_chain/fork_choice/fork_choice.nim
+++ b/beacon_chain/fork_choice/fork_choice.nim
@@ -369,7 +369,7 @@ proc on_attestation*(
   ? self.update_time(dag,
     max(wallTime, attestation_slot.start_beacon_time(dag.timeParams)))
 
-  # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.9/specs/gloas/fork-choice.md#modified-validate_on_attestation
+  # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#modified-validate_on_attestation
   if attestation_slot.epoch >= dag.cfg.GLOAS_FORK_EPOCH:
     let index = attestation_committee_index.uint64
     if index notin [0'u64, 1'u64]:
--- a/beacon_chain/fork_choice/fork_choice.nim
+++ b/beacon_chain/fork_choice/fork_choice.nim
@@ -122,7 +122,7 @@ func process_attestation(
         new_vote = shortLog(vote)
 
 proc process_attestation_queue(
-    self: var ForkChoice, slot: Slot, dag: ChainDAGRef) =
+    self: var ForkChoice, slot: Slot, cfg: RuntimeConfig) =
   # Spec:
   # Attestations can only affect the fork choice of subsequent slots.
   # Delay consideration in the fork choice until their slot is in the past.
@@ -132,7 +132,7 @@ proc process_attestation_queue(
       for validator_index in it.attesting_indices:
         self.backend.process_attestation(
           validator_index, it.block_root, it.slot,
-          it.committee_index == CommitteeIndex(1), dag.cfg)
+          it.committee_index == CommitteeIndex(1), cfg)
       false
     else:
       true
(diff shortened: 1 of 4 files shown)


beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -347,22 +364,38 @@ proc on_attestation*(
     attestation_slot: Slot,
     beacon_block_root: Eth2Digest,
     attesting_indices: openArray[ValidatorIndex],
+    attestation_committee_index: CommitteeIndex,
     wallTime: BeaconTime): FcResult[void] =
   ? self.update_time(dag,
     max(wallTime, attestation_slot.start_beacon_time(dag.timeParams)))
 
+  # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.9/specs/gloas/fork-choice.md#modified-validate_on_attestation
```
tersec: alpha.10 was just released

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -92,21 +92,37 @@ proc init*(
...
+        vote.next_root = block_root
+        vote.slot = slot
 
       trace "Integrating vote in fork choice",
         validator_index = validator_index,
         new_vote = shortLog(vote)
 
-proc process_attestation_queue(self: var ForkChoice, slot: Slot) =
+proc process_attestation_queue(
+    self: var ForkChoice, slot: Slot, dag: ChainDAGRef) =
```
tersec: The only part of `dag` used here is `GLOAS_FORK_EPOCH`, and even more specifically, whether `slot.epoch >= cfg.GLOAS_FORK_EPOCH`. But passing the entire `ChainDAGRef` makes it tricky to look at the signature of this function and know dataflow-wise what it does/does not depend on, as well as whether that's mutable (as the data behind `ref`s are) or compiler-enforced immutable (non-`var` value types).

While fork choice is, yes, already somewhat intertwined with the DAG, most parts remain somewhat factored from it. Would be preferable to pass only the relevant subset of information to `process_attestation_queue`. This could be either `GLOAS_FORK_EPOCH` in a standalone way or the information of whether `slot` Gloas or later.

tersec: The `RuntimeConfig` part is reasonably-scoped too, being that it's also fairly focused and immutable, but still larger than it really needs to be.

beacon_chain/fork_choice/fork_choice.nim
tersec, line 363: The actual `dag` dependencies here are `dag.cfg.GLOAS_FORK_EPOCH` and `dag.timeParams`.

As
- https://github.com/status-im/nimbus-eth2/pull/8544#discussion_r3357733235

also notes, `ChainDAGRef` has the challenge that it's overarching and mutable, which creates architecturally this recursive dependency between fork choice and the DAG. This is hard to avoid for certain operations (`on_block` in particular), but here it's easier to disentangle: it's semi-accidental that the easiest way to get to the `RuntimeConfig` and `TimeParams` is via this all-encompassing, mutable-via-ref object. It is one more parameter, but significantly more tightly scoped (and immutable) to swap out the `ChainDAGRef` for `RuntimeConfig` and `TimeParams`, and any caller which can provide the former can also provide the latter.

Tomi-3-0: the first statement here: 
`? self.update_time(dag, max(wallTime, attestation_slot.start_beacon_time(dag.timeParams)))`
passes the whole dag  and `update_time` calls `on_tick` that  reads dag.head.root and drives reconfirm_fcr/update_confirmed. swappng it here isn't exactly straightforward

tersec: Ah, right. So skip this one then.

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8544/changes/56624de03be9c7673792b2a02fb26dfb3dc2ffc0

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8544/commits/c15f5cfab876791bb8814671cb5db334fca46940

PR #8549, opened 2026-06-05
remove eth1 deposit bridge in Fulu

Tomi-3-0:
- https://github.com/ethereum/consensus-specs/pull/4704
- Also remove the `skipStateRootValidation` flag in the transition fixture test which disabled `verifyStateRoot` and silently hid state divergences.

### DIFF
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -849,7 +849,7 @@ proc process_operations(
   const consensusFork = typeof(state).kind
 
   when consensusFork >= ConsensusFork.Fulu:
-    let req_deposits = 0'u64
+    const req_deposits = 0'u64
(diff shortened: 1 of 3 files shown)


beacon_chain/spec/state_transition_block.nim
```diff
@@ -845,7 +848,9 @@ proc process_operations(
   # deposits
   const consensusFork = typeof(state).kind
 
-  when consensusFork >= ConsensusFork.Electra:
+  when consensusFork >= ConsensusFork.Fulu:
+    let req_deposits = 0'u64
```
tersec: `const`

tersec: https://github.com/status-im/nimbus-eth2/pull/8549/commits/60010038b7d2a5f947820e80a7043b0e1961ba17

PR #8559, opened 2026-06-05
add fork-choice handlers for ePBS payloads

Tomi-3-0:


beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -17,10 +17,10 @@ import
   ../spec/datatypes/[phase0, altair, bellatrix],
   # Fork choice
   ../consensus_object_pools/[spec_cache, blockchain_dag],
-  "."/[fork_choice_types, proto_array, fast_confirmation]
+  "."/[fork_choice_types, proto_array, fast_confirmation, fork_choice_epbs]
```
tersec: The quotes around `.` aren't required anymore; they're a workaround for old Nim bugs.
```nim
  ./[fork_choice_types, ...]
```
is fine.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+  # Standard library
+  std/tables,
+  # Status libraries
+  results, chronicles,
+  # Internal
+  ../spec/[beaconstate, helpers],
+  ../spec/datatypes/[phase0, altair, bellatrix],
+  # Fork choice
+  ../consensus_object_pools/[blockchain_dag, spec_cache],
+  "."/[fork_choice_types, proto_array]
```
tersec: Similarly don't need `"."/[...]` here; `./[...]` suffices.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_execution_payload_envelope
+proc on_execution_payload*(
+    self: var ForkChoice, dag: ChainDAGRef,
+    signedEnvelope: SignedExecutionPayloadEnvelope): FcResult[void] =
+  ## Run ``on_execution_payload_envelope`` upon receiving a new execution
+  ## payload envelope.
+  template envelope: auto = signedEnvelope.message
+  let
+    beacon_block_root = envelope.beacon_block_root
```
tersec: Prefer this as a `template` to avoid the copy.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+  var ptcIndices: seq[int]
+  withState(dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      # Check that it's for the current slot if it is coming from the wire
+      if not is_from_block and
+          slot != self.checkpoints.time.slotOrZero(dag.timeParams):
+        return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+      # Get all positions of the attester in the PTC
+      var ptcIndex = 0
```
tersec: [`enumerate`](https://nim-lang.org/docs/enumerate.html#enumerate.m%2CForLoopStmt) already provides exactly this mechanism:
```nim
     for ptcIndex, vidx in enumerate(get_ptc(forkyState.data, slot)):
```

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+proc on_payload_attestation_message*(
+    self: var ForkChoice,
+    dag: ChainDAGRef,
+    validator_index: uint64,
+    data: PayloadAttestationData,
+    is_from_block: bool = false): FcResult[void] =
+  ## Run ``on_payload_attestation_message`` upon receiving a new ``ptc_message`` from
+  ## either within a block or directly on the wire.
+  let
+    beacon_block_root = data.beacon_block_root
```
tersec: Prefer `template` for `beacon_block_root`

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+              self.backend.votes[val].slot == FAR_FUTURE_SLOT and
+              val < balances.lenu64:
+            head_weight += balances[val].effective_balance
+
+  head_weight < reorg_threshold
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-should_apply_proposer_boost
+proc should_apply_proposer_boost*(
+    self: var ForkChoice, dag: ChainDAGRef): bool =
+  let proposer_root = self.checkpoints.proposer_boost_root
```
tersec: Prefer `template`

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -480,12 +505,23 @@ proc process_block*(
             vidx, attestation.data.beacon_block_root, attestation.data.slot,
             false, dag.cfg)
 
+  when typeof(blck).kind >= ConsensusFork.Gloas:
+    for pa in blck.body.payload_attestations:
+      let root = pa.data.beacon_block_root
```
tersec: Prefer `template`

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+        data.payload_present
+    self.backend.ptcVotes.mgetOrPut(
+      beacon_block_root, PtcVoteTally()).available[ptcIndex] =
+        data.blob_data_available
+
+  ok()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_execution_payload_envelope
+proc on_execution_payload*(
+    self: var ForkChoice, dag: ChainDAGRef,
```
tersec: `dag` here used only for `TimeParams` and `RuntimeConfig`. See discussions elsewhere for why it's preferable to provide them separately; `ChainDAGRef` is a [God object](https://en.wikipedia.org/wiki/God_object) the scope of which is difficult to bound or reason about.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+
+  # Apply proposer boost if `parent` is not weak
+  if not self.is_head_weak(parent_node.bid.root, dag):
+    return true
+
+  # If `parent` is weak and from the previous slot, apply proposer boost if
+  # there are no early equivocations
+  let parentBlck = dag.getBlockRef(parent_node.bid.root)
+  if parentBlck.isNone:
+    return true
```
tersec, line 74:
```nim
  let parentBlck = dag.getBlockRef(parent_node.bid.root).valueOr:
    return true
```

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+{.push raises: [], gcsafe.}
+
+import
+  # Standard library
+  std/tables,
+  # Status libraries
+  results, chronicles,
+  # Internal
+  ../spec/[beaconstate, helpers],
+  ../spec/datatypes/[phase0, altair, bellatrix],
```
tersec: This looks a lot like it was copied from some other module. Are these specific imports actually required?

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,184 @@
...
+  let proto_node = self.backend.proto_array.node(head_root).valueOr:
+    return true
+
+  var head_weight = proto_node.weight.Gwei
+
+  # Also count effective balance of equivocating validators (vote disabled via
+  # `vote.slot == FAR_FUTURE_SLOT`) in the head slot's committees, so the weight
+  # is monotonic in observed attestations.
+  let headBlck = dag.getBlockRef(head_root)
+  if headBlck.isSome:
```
tersec:
```nim
  dag.getBlockRef(head_root).isErrOr:
```
and then use `value` instead of `headBlck.get` inside the block.

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -424,30 +424,55 @@ func process_block*(
   self.proto_array.onBlock(
     bid, parent_root, checkpoints, unrealized, parent_payload_status)
 
-# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.5/specs/gloas/fork-choice.md#modified-update_proposer_boost_root
-proc update_proposer_boost_root(
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#modified-record_block_timeliness
+proc record_block_timeliness(
     self: var ForkChoice, dag: ChainDAGRef,
```
tersec: `dag` only used here for `TimeParams`.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,168 @@
...
+  var ptcIndices: seq[int]
+  withState(dag.headState):
+    when consensusFork >= ConsensusFork.Gloas:
+      # Check that it's for the current slot if it is coming from the wire
+      if not is_from_block and
+          slot != self.checkpoints.time.slotOrZero(dag.timeParams):
+        return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+      # Get all positions of the attester in the PTC
+      for ptcIndex, vidx in enumerate(get_ptc(forkyState.data, slot)):
```
tersec: This is triggering a compile error:
```
[2026-06-06T04:38:47.582Z] /home/jenkins/workspace/tforms_linux_x86_64_main_PR-8559/beacon_chain/fork_choice/fork_choice_epbs.nim(124, 46) Error: type mismatch
[2026-06-06T04:38:47.582Z] Expression: get_ptc(dag.headState.hezeData.data, slot)
[2026-06-06T04:38:47.582Z]   [1] dag.headState.hezeData.data: BeaconState
[2026-06-06T04:38:47.582Z]   [2] slot: Slot
[2026-06-06T04:38:47.582Z] 
[2026-06-06T04:38:47.582Z] Expected one of (first mismatch at [position]):
[2026-06-06T04:38:47.582Z] [2] iterator get_ptc(state: gloas.BeaconState | heze.BeaconState;
[2026-06-06T04:38:47.582Z]                  shufflingRef: ShufflingRef; slot: Slot): ValidatorIndex
[2026-06-06T04:38:47.582Z] 
[2026-06-06T04:38:47.582Z] make: *** [Makefile:443: mev_mock] Error 1
```

But locally, adding an `import` directly which seems like it should already be there transitively (otherwise it wouldn't have worked to begin with) isn't being picked up by `enumerate` properly:
```nim
diff --git i/beacon_chain/fork_choice/fork_choice_epbs.nim w/beacon_chain/fork_choice/fork_choice_epbs.nim
index db33c1779..2805ceeed 100644
--- i/beacon_chain/fork_choice/fork_choice_epbs.nim
+++ w/beacon_chain/fork_choice/fork_choice_epbs.nim
@@ -12,6 +12,8 @@ import
   ../consensus_object_pools/[blockchain_dag, spec_cache],
   ./[fork_choice_types, proto_array]
 
+from ../spec/beaconstate import get_ptc
+
 # https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#modified-is_head_weak
 proc is_head_weak(
     self: var ForkChoice, head_root: Eth2Digest, dag: ChainDAGRef): bool =
```

This might be a Nim issue. I'm investigating.

tersec: Ah, it's due to dropping some of the other overbroad imports. Still an improvement to swap them for a more targeted import of `spec/beaconstate`, which is what's actually required here.

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/compare/255c3434387a31f9c28e73dc3b87abe348e46375..23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/23247ecab896af3b60334140ba5eb5985cc384d6

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,169 @@
...
+        return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+      trace "Validated PTC vote",
+        validator_index,
+        ptc_positions = ptcIndices.len,
+        payload_present = data.payload_present,
+        blob_data_available = data.blob_data_available
+
+  # Update the votes for the block
+  if ptcIndices.len > 0:
```
tersec: It seems like the only reason this has to be conditional is the it's not part of the `withState(dag.headState):` / `when consensusFork >= ConsensusFork.Gloas:` block?

This can't ever be true past
```nim
if ptcIndices.len == 0:
  return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
```
in Gloas or later, but that information is lost. This conditional is really detecting the pre-Gloas case, implicitly.

The essentially nonconditional aspect of this flow might be clearer putting it the early return, within the `withState(dag.headState):` / `when consensusFork >= ConsensusFork.Gloas:` block. Then one also wouldn't have this odd unused-before-Gloas `var ptcIndices: seq[int]` having to hang around outside any possible use case and it more clearly scopes all the Gloas thing together.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,169 @@
...
+    when consensusFork >= ConsensusFork.Gloas:
+      # Check that it's for the current slot if it is coming from the wire
+      if not is_from_block and
+          slot != self.checkpoints.time.slotOrZero(dag.timeParams):
+        return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+      # Get all positions of the attester in the PTC
+      for ptcIndex, vidx in enumerate(get_ptc(forkyState.data, slot)):
+        if vidx == valIdx:
+          ptcIndices.add ptcIndex
```
tersec: Is the reason this accumulates into `ptcIndices` at all rather than filling in `tally.present` and `tally.available` incrementally that in https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_payload_attestation_message there's this whole
```python
    # Verify the signature and check that its for the current slot if it is coming from the wire
    if not is_from_block:
        # Check that the attestation is for the current slot
        assert data.slot == get_current_slot(store)
        # Verify the signature
        assert is_valid_indexed_payload_attestation(
            state,
            IndexedPayloadAttestation(
                attesting_indices=[ptc_message.validator_index],
                data=data,
                signature=ptc_message.signature,
            ),
        )
```
section not here? Because the logic there is, currently, accumulate the `ptc_indices`, then run `is_valid_indexed_payload_attestation` first, so if it fails, the counts aren't changed, then update `store.payload_timeliness_vote` and `store.payload_data_availability_vote` (here `tally`) after.

But here there's no check on `is_valid_indexed_payload_attestation` so that split isn't useful.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,169 @@
...
+    return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+  # PTC votes can only change the vote for their assigned beacon block, return
+  # early otherwise
+  let blockSlot = self.backend.proto_array.slot(beacon_block_root).valueOr:
+    return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+  if slot != blockSlot:
+    return ok()
+
+  var ptcIndices: seq[int]
```
tersec: [The spec](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_payload_attestation_message) names this `ptc_indices`. Which, there are plenty of other snake-case variables around, might be nicer to match the spec in this, yeah, superficial aspect.

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,169 @@
...
+  withState(dag.headState):
```
tersec, line 122: https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_payload_attestation_message says
```python
    # PTC attestation must be for a known block. If block is unknown, delay consideration until the block is found
    assert data.beacon_block_root in store.block_states
    state = store.block_states[data.beacon_block_root]
```
and not necessarily a head state. Is this generally correct?

Practically, this code does not actually need the full state. It needs it for `get_ptc` and `is_valid_indexed_payload_attestation`, and, see other comment, the later isn't called at all in this implementation.

So assuming that's intentional, it's for `get_ptc`, it doesn't need a full state at all, only a `ShufflingRef`:
https://github.com/status-im/nimbus-eth2/blob/1b5ed320eaaba8f16cbd61e7661990b6c9633d7e/beacon_chain/consensus_object_pools/spec_cache.nim#L232-L251

which are (a) per-epoch, not per-slot, (b) a function basically of dependent root, so a fairly big reorg needs to happen to invalidate them, and (c) Nimbus keeps a few around in `EpochRef` cache form, so it might be efficient enough to actually use.

But would need to look at the performance cliffs possible, sure.

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -424,30 +424,55 @@ func process_block*(
   self.proto_array.onBlock(
     bid, parent_root, checkpoints, unrealized, parent_payload_status)
 
-# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.5/specs/gloas/fork-choice.md#modified-update_proposer_boost_root
-proc update_proposer_boost_root(
-    self: var ForkChoice, dag: ChainDAGRef,
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#modified-record_block_timeliness
+proc record_block_timeliness(
```
tersec: `func`

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -424,30 +424,55 @@ func process_block*(
...
+  let epoch = current_slot.epoch
+  if epoch <= MIN_SEED_LOOKAHEAD:
+    return ZERO_HASH
+  # spec's `start_slot(epoch - MIN_SEED_LOOKAHEAD) - 1` == attester_dependent_slot
+  let dependent = dag.atSlot(bid, epoch.attester_dependent_slot).valueOr:
+    return ZERO_HASH
+  dependent.bid.root
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#modified-update_proposer_boost_root
+proc update_proposer_boost_root(
```
tersec: `func`

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,169 @@
...
+    let candBlck = dag.getBlockRef(root).valueOr: continue
+    if candBlck.slot != parent_node.bid.slot: continue
+    let candProposer = dag.getProposer(candBlck, candBlck.slot).valueOr: continue
+    if candProposer == parentProposer:
+      return false
+
+  true
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_payload_attestation_message
+proc on_payload_attestation_message*(
```
tersec, line 90: `func`

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,169 @@
...
+    let tally = addr self.backend.ptcVotes.mgetOrPut(
+      beacon_block_root, PtcVoteTally())
+    for ptcIndex in ptcIndices:
+      tally.present[ptcIndex] = data.payload_present
+      tally.available[ptcIndex] = data.blob_data_available
+
+  ok()
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-on_execution_payload_envelope
+proc on_execution_payload*(
```
tersec: `func`

beacon_chain/fork_choice/fork_choice_epbs.nim
Tomi-3-0: this will have to stay as a `proc` since it calls the `shufflingRef` version of `get_ptc` now
`Hint: 'on_payload_attestation_message' calls `.sideEffect` 'getShufflingRef'`

tersec: Yeah that's fine.

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/cd1be302e3c24cc0641d55a099f0f642927072e0

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/cd1be302e3c24cc0641d55a099f0f642927072e0

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/cd1be302e3c24cc0641d55a099f0f642927072e0

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/cd1be302e3c24cc0641d55a099f0f642927072e0

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/cd1be302e3c24cc0641d55a099f0f642927072e0

beacon_chain/fork_choice/fork_choice_epbs.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8559/commits/cd1be302e3c24cc0641d55a099f0f642927072e0

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,177 @@
...
+
+import
+  std/enumerate,
+  ../consensus_object_pools/[blockchain_dag, spec_cache],
+  ./[fork_choice_types, proto_array]
+
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#modified-is_head_weak
+proc is_head_weak(
+    self: var ForkChoice, head_root: Eth2Digest, dag: ChainDAGRef): bool =
+  let
```
tersec, line 21: Spec code has a useful comment
```python
  # Calculate weight threshold for weak head
```

Tomi-3-0: https://github.com/status-im/nimbus-eth2/pull/8559/commits/1d3e7e3dd937f8041d8a11d64fb2678cc4be20cd

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,190 @@
...
+    return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+  if slot != blockSlot:
+    return ok()
+
+  # The spec reads the PTC from `state = store.block_states[beacon_block_root]`.
+  # `ptc_window` is committed state, so load the attested block's post-state and
+  # read it there rather than recomputing the committee.
+  let blockRef = dag.getBlockRef(beacon_block_root).valueOr:
+    return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+  dag.withUpdatedState(dag.epochRefState, blockRef.bid.atSlot()):
```
tersec: "must only be used in `getEpochRef`:
https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/consensus_object_pools/block_pools_types.nim#L208-L210

Which indeed it is:
https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1511-L1552

And in `init`, which has to set it up to begin with:
https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/consensus_object_pools/blockchain_dag.nim#L1270-L1295

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,190 @@
...
+          return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+        # Verify the signature and check that its for the current slot if it is
+        # coming from the wire
+        if not is_from_block:
+          # Check that the attestation is for the current slot
+          if slot != self.checkpoints.time.slotOrZero(dag.timeParams):
+            return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+          # Verify the signature
+          if not is_valid_indexed_payload_attestation(
```
tersec: In
- #7808

`on_payload_attestation_message()` is only called from `processPayloadAttestationMessage`: https://github.com/status-im/nimbus-eth2/blob/5ef04b8fda4a35ec8b32e2e670007f9c52ee1a03/beacon_chain/gossip_processing/eth2_processor.nim#L951-L975 which already verifies with `validatePayloadAttestationMessage()` the gossip-received `PayloadAttestationMessage`, which checks the validity of `PayloadAttestationMessage.signature`: https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/gossip_processing/gossip_validation.nim#L2223-L2249 using https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/gossip_processing/batch_validation.nim#L577-L600 and https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/spec/signatures_batch.nim#L182-L189

i.e. it's using a pubkey based on the validator, on the `PayloadAttestationData` (which itself contains no signature field), against the signature.

But [`is_valid_indexed_payload_attestation()`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/beacon-chain.md#new-is_valid_indexed_payload_attestation) does two things:
```python
def is_valid_indexed_payload_attestation(
    state: BeaconState, attestation: IndexedPayloadAttestation
) -> bool:
    """
    Check if ``attestation`` is non-empty, has sorted indices, and has
    a valid aggregate signature.
    """
    # Verify indices are non-empty and sorted
    indices = attestation.attesting_indices
    if len(indices) == 0 or indices != sorted(indices):
        return False

    # Verify aggregate signature
    pubkeys = [state.validators[i].pubkey for i in indices]
    domain = get_domain(state, DOMAIN_PTC_ATTESTER, compute_epoch_at_slot(attestation.data.slot))
    signing_root = compute_signing_root(attestation.data, domain)
    return bls.FastAggregateVerify(pubkeys, signing_root, attestation.signature)
```
1. `Verify indices are non-empty and sorted` (true by construction here: `attesting_indices: List[uint64, Limit PTC_SIZE].init(@[validator_index])`); and
2. `Verify aggregate signature` (the same thing gossip already did, but this time in a kind of trivial/degenerate "aggregate" form of exactly one validator index)

How is this check not entirely redundant if it's already passed gossip validation?

Furthermore, the gossip validation does it with batch validation, whereas this does not, so the gossip validation check will be faster.

(Even if it is, a note should be left here why that part of the `on_payload_attestation_message()` spec is being skipped, if it can indeed be skipped).

Tomi-3-0: Yes, it is redundant. The weak reason for adding it was to get those test cases passing. These tests have no gossip layer (handlers get pre-trusted data), so I'll either mirror the check in the test directly or skip that vector. I think the former is a better option

tersec: Yeah, agree. Ordinarily I'd say that the test should be as realistic as possible, et cetera, but BLS verification is too expensive to just throw in as a why-not.

tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/8714e008d6421cd951ba8deaa48073f23474dc99

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,171 @@
...
+        if vidx == valIdx:
+          ptc_indices.add ptc_index
+
+      # Check that the attester is from the PTC
+      if ptc_indices.len == 0:
+        return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
+
+      # Check that the attestation is for the current slot.
+      if slot != self.checkpoints.time.slotOrZero(dag.timeParams):
+        return err ForkChoiceError(kind: fcInvalidPayloadAttestation)
```
tersec, line 136: If this check is hoisted above the `for ptc_index, vidx in enumerate(get_ptc(forkyState.data, slot)):` loop, then instead of `ptc_indices.add ptc_index`, that loop can directly update `tally.present` and `tally.available`. No `var ptc_indices: seq[int]` required.

Yes, going with the specs is a good default, and new temporary integers or booleans are fine, they don't matter. But unnecessary intermediate `seq`s and similar are not worth it. Similarly, when it has O(n^2) or O(n^3) algorithms (not the issue here, specifically), it typically can't be used directly.

One challenge with these "small" `seq`/etc other heap allocations is each one adds slightly to garbage collector load, and they don't create a single hotspot which can be easily profiled, so much as just make everything a bit slower.

tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/ad31b9435c4b54ecdc97c93eedbbee521a6fecd4

tests/consensus_spec/test_fixture_fork_choice.nim
```diff
@@ -471,21 +560,21 @@ template fcSuite(suiteName: static[string], testPathElem: static[string]) =
       if kind != pcDir or not os_ops.dirExists(testsPath):
         continue
       if path.contains("eip7732") or path.contains("eip7805") or
```
tersec: Are either of these exclusions still necessary?
```
nimbus-eth2/vendor/nim-eth2-scenarios/tests-v1.7.0-alpha.10/mainnet$ ls
altair  bellatrix  capella  deneb  electra  fulu  gloas  heze  phase0
```
They used to be, yes, but 7732 is now `gloas` and 7805 is now `heze`.

tersec: https://github.com/status-im/nimbus-eth2/pull/8559/commits/ea68ff695eb21ff018f571096c5cba7f666e1ada

beacon_chain/fork_choice/fork_choice_epbs.nim
```diff
@@ -0,0 +1,165 @@
...
+    self: var ForkChoice,
+    dag: ChainDAGRef,
+    validator_index: uint64,
+    data: PayloadAttestationData): FcResult[void] =
+  ## Run ``on_payload_attestation_message`` upon receiving a new ``ptc_message``
+  ## from either within a block or directly on the wire.
+  template beacon_block_root: untyped = data.beacon_block_root
+  let
+    slot = data.slot
+    valIdx = ValidatorIndex(validator_index)
```
tersec: This is, in principle, not safe: https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/spec/datatypes/base.nim#L171

For valid validator indices, it will be, because there are other constraints from the deposit contract such that reaching 2 billion validators shouldn't happen anyway. Current indices max out around 1000x less than that.

The concern, rather, is hostile-constructed payload attestations, which also would be an issue in `validatePayloadAttestationMessage(...)`: https://github.com/status-im/nimbus-eth2/blob/50b259a47a367ec8b0160284778fe1a4dfccce7a/beacon_chain/gossip_processing/gossip_validation.nim#L2203-L2227

At no point does it check `payload_attestation_message.validator_index` for being in bounds of the actual known validator set, nor does it here. So arguably this not checking for overflow is not as bad here specifically, but it's not checked anywhere else either between being external network input and crashing Nimbus.

The usual idiom to do this conversion safely is `let foo = ValidatorIndex.init(bar).valueOr:`

tersec: https://github.com/status-im/nimbus-eth2/pull/8559/changes/3204ac35707013ad7647a165400bbd870bc09055

beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -424,30 +424,55 @@ func process_block*(
...
   const consensusFork = typeof(blck).kind
-
+  let isCurrentSlot = current_slot == blck.slot
+  let ptcTimely =
+    when consensusFork >= ConsensusFork.Gloas:
+      isCurrentSlot and self.checkpoints.time <
+        blck.slot.payload_attestation_deadline(timeParams)
+    else:
+      false
+  self.backend.block_timeliness[blckRef.root] = [
```
tersec: I'm not sure that `self.backend.block_timeliness` has to exist at all.

The block timeliness, yes, must be computed, but it's only written here in `record_block_timeliness(...)` and then read back in https://github.com/status-im/nimbus-eth2/blob/3204ac35707013ad7647a165400bbd870bc09055/beacon_chain/fork_choice/fork_choice.nim#L457-L475

But `record_block_timeliness(...)` and `update_proposer_boost_root(...)` are called immediately one after another: https://github.com/status-im/nimbus-eth2/blob/3204ac35707013ad7647a165400bbd870bc09055/beacon_chain/fork_choice/fork_choice.nim#L520-L523 and on the same `blckRef`, therefore the same `blckRef.root`.

So `self.update_proposer_boost_root(dag, blckRef, slot)` never cares about some arbitrary block timeliness, it cares about the block timeliness computed by `self.record_block_timeliness(dag.timeParams, blckRef, blck, slot)`. Thus, https://github.com/status-im/nimbus-eth2/blob/3204ac35707013ad7647a165400bbd870bc09055/beacon_chain/fork_choice/fork_choice_types.nim#L182 doesn't have to exist at all, and doesn't have to be pruned, either.

Tomi-3-0: The scope of this PR isn't entirely self-contained. There's a third reader: block_timeliness is also used by should_apply_proposer_boost, which iterates it to detect proposer equivocation, i.e. to check whether the parent's proposer also produced another PTC-timely block at the same slot:
  
  https://github.com/status-im/nimbus-eth2/blob/3204ac35707013ad7647a165400bbd870bc09055/beacon_chain/fork_choice/fork_choice_epbs.nim#L79-L86

I agree that record_block_timeliness and update_proposer_boost_root run back-to-back on the same blckRef, so the attestation-timeliness doesn't need to do a round-trip through the Table. But rather than removing it entirely, the two bits can be split better:

- The attestation bit is only read right after it's written, so I can just return it instead of storing it.
- The PTC bit is different: `should_apply_proposer_boost` needs a way to look it up across many blocks, so it still has to be stored, but now it can be a 'simpler' `HashSet` of the PTC-timely block roots instead of an `array[2, bool]`, since the attestation bit no longer needs a slot there.

tersec: Ok, sure.

tersec: https://github.com/status-im/nimbus-eth2/pull/8559/commits/e14635f8327b6973dff48a25912f796375d05b95

PR #8561, opened 2026-06-06
run newPayload on Fulu blocks concurrently with DA arrival

tersec:
This change affects only Fulu, neither Electra or earlier, nor Gloas or later. Neither has the latency issue involved. Currently, Nimbus:
1. receives a Fulu beacon block
2. waits for `enoughColumns` to arrive
3. sends the EL `newPayload`
4. when that returns `VALID`, resolves the block.

(2) and (3) do not overlap, currently. Pre-Fulu, blobs tended to arrive and be verified fairly quickly. Especially when waiting for 64 columns, this can readily require 2+s seconds, which leaves little time for some longer blocks to fully execute before running out of the 4s window.

So this PR changes the flow to:
1. receives a Fulu beacon block
2. sends the EL `newPayload`, which then runs in the background
3. waits for `enoughColumns` to arrive
4. when (2) and (3) both occur, resolves the block

There isn't a safety/security/EL DoS issue here: this PR specifically targets gossip traffic, and gossip validation already guarantees that each block is the first valid one per-proposer/slot to arrive.

### DIFF
--- a/beacon_chain/gossip_processing/block_processor.nim
+++ b/beacon_chain/gossip_processing/block_processor.nim
@@ -8,7 +8,7 @@
 {.push raises: [], gcsafe.}
 
 import
-  chronicles, chronos, metrics,
+  chronicles, chronos, metrics, minilru,
   ../spec/[forks, helpers_el, signatures, signatures_batch, column_map,
            peerdas_helpers],
   ../sszdump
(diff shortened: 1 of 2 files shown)


tersec:
The Hoodi fleet node`geth-04` `unstable` decreased its attestation head miss rate, within the 48 hour window around the deployment on those nodes of this PR, from:
p50: 335 to 219 (35% decrease)
p90: 439 to 284 (35% decrease)
p95: 484 to 304 (37% decrease)
p99: 567 to 329 (42% decrease)

The Hoodi fleet node`geth-05` `unstable` decreased its attestation head miss rate, within the 48 hour window around the deployment on those nodes of this PR, from:
p50: 929 to 557 (40% decrease)
p90: 1140 to 711 (38% decrease)
p95: 1217 to 810 (33% decrease)
p99: 1282 to 909 (30% decrease)

In both charts, of geth-04 and geth-05, before/after threshold is the red node-offline/restarting spike around 2026-06-06 09:00 (UTC):
2026-06-07T01:05:04,576274840+00:00
2026-06-07T01:06:02,922303525+00:00

PR #8572, opened 2026-06-08
add epbs fork-choice handlers into get_head and gossip

Tomi-3-0:
Wires the ePBS fork-choice handlers into get_head, head selection, and gossip, so the node tracks each block's FULL vs EMPTY payload status and acts on it. get_head now returns the head together with its payload status.
  - get_head: gate proposer boost via should_apply_proposer_boost  proposer boost is applied only when `should_apply_proposer_boost` holds for the current slot; otherwise the boost root is zeroed. `get_head` returns (root, full) so callers know whether the head is the `FULL` or `EMPTY` variant.
  - find_head: compute emptyPreferredRoot ([should_extend_payload](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/fork-choice.md#new-should_extend_payload)) implements the gloas variant tie-break
  - eth2_processor: feeds PTC messages to fork choice via `on_payload_attestation_message`,  payload-attestation gossip is now routed into the fork-choice backend, populating ptc_votes so is_payload_timely / the should_extend_payload tie-break have the data they need.
  -   Track weight per node (`pendingWeight`) and subtract it in the EMPTY-vs-FULL tie-break, mirroring `is_supporting_vote`


beacon_chain/fork_choice/fork_choice.nim
```diff
@@ -558,13 +582,43 @@ func find_head(
...
+    let parentNode = self.proto_array.node(parentIdx).valueOr:
+      break maybeEmptyPreferred
+    let
+      parentRoot = parentNode.bid.root
+      fullParentIdx =
+        self.proto_array.fullBlockIndices.getOrDefault(parentRoot, -1)
+    if fullParentIdx < 0:
+      emptyPreferredRoot = parentRoot
+    elif parentIdx != fullParentIdx:
+      let tally = self.ptc_votes.getOrDefault(parentRoot)
```
tersec: `getOrDefault` always runtime-creates a `default(PtcVoteTally)`:
```nim
  PtcVoteTally* = object
    present*: BitArray[int PTC_SIZE]
    available*: BitArray[int PTC_SIZE]
```
regardless of whether the object exists in the database .`getOrDefault(...)` is fine for integers or other smaller objects, but it can get a bit much for something like a `PtcVoteTally`:
```c
#line 605 FX_335
        tyObject_PtcVoteTally__YwJ5L6hwPWRgNLYYmbByNg tally_1;
#line 606 FX_335
        NU64 present_1;
        NU64 available_1;
#line 604 FX_335
        nimln_(604);
        if (!!((parentIdx_1 == fullParentIdx_1)))
          goto LA25_;
        nimZeroMem((void *)(&tally_1),
                   sizeof(tyObject_PtcVoteTally__YwJ5L6hwPWRgNLYYmbByNg));
        _ZN6tables22getOrDefault_s256_s256E5TableI7MDigestIEN17fork_choice_types12PtcVoteTallyEE7MDigestIE(
            (*self_p0).ptc_votes, (&parentRoot_1), (&tally_1));
```
That is, every time this is called, it pointlessly re-constructs and zero-initializes  usually-unused `default(PtcVoteTally)`.

This is partly why other parts of the codebase which do similar things with `Table` use other approaches, e.g., https://github.com/status-im/nimbus-eth2/blob/f8262ddfb5a0e4e54b37ae922b8fc5bb6787e11f/beacon_chain/consensus_object_pools/attestation_pool.nim#L501-L509 or [`proc mgetOrPut[A, B](t: var Table[A, B]; key: A; val: B): var B`](https://nim-lang.org/docs/tables.html#mgetOrPut%2CTable%5BA%2CB%5D%2CA%2CB) to achieve similar goals without doing a double table lookup to first check whether a value is present, then act on that information.

tersec:
Can rerun `make test` to automatically resolve the merge conflicts in `ConsensusSpecPreset-{mainnet,minimal}.md`.

beacon_chain/fork_choice/fork_choice.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8572/changes/31c891a7d4b21505caebde57a4d209a487f66f23

tersec:
Merge conflict in `tests/consensus_spec/test_fixture_fork_choice.nim`

Issue #6059, opened 2024-03-11
Database or Era files corrupt on Mainnet


jakubgs:
**Describe the bug**
Recently we've been seeing errors like this on Mainnet nodes, in this case `libp2p-01` on `linux-07.ih-eu-mda1.nimbus.mainnet`:
```json
{"lvl":"FAT","ts":"2024-03-11 11:20:37.394+00:00","msg":"Era summaries don't lead up to backfill, database or era files corrupt?","bid":"c78a68b6:8518954"}
```
A previous instance of this happening was with `unstable-02` on `linux-03.ih-eu-mda1.nimbus.mainnet`:
```json
{"lvl":"FAT","ts":"2024-03-08 08:11:18.637+00:00","msg":"Era summaries don't lead up to backfill, database or era files corrupt?","bid":"bd80c153:8503305"}
```
Which was purged and allowed to resync. The `libp2p-01` was left in a broken state for debugging.

**Additional context**
The `unstable-02` was running `v24.2.2-3b4b1c-stateofus` when issue was discovered(probably).
The `libp2p-02` was running `v24.2.2-1fef38-stateofus` when issue was discovered.

No issues were identified by `verifyEra` command.

PR #8575, opened 2026-06-09
Fix incorrect ERA files check.

cheatfate:
This fix should finally address #6059

* Fix missing fields in logs `Block backfilled` and `Block received` because of `logScope` in generic functions.

### DIFF
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -261,9 +261,9 @@ proc processSignedBeaconBlock*(
 
   # Start of block processing - in reality, we have already gone through SSZ
   # decoding at this stage, which may be significant
-  let bid = BlockId(root: signedBlock.root, slot: signedBlock.message.slot)
   debug "Block received",
-    bid = shortLog(bid),
+    bid = shortLog(
+      BlockId(root: signedBlock.root, slot: signedBlock.message.slot)),
     blck = shortLog(signedBlock.message),
     signature = shortLog(signedBlock.signature),
     wallSlot,
--- a/beacon_chain/gossip_processing/eth2_processor.nim
+++ b/beacon_chain/gossip_processing/eth2_processor.nim
@@ -261,7 +261,13 @@ proc processSignedBeaconBlock*(
 
   # Start of block processing - in reality, we have already gone through SSZ
   # decoding at this stage, which may be significant
-  debug "Block received", delay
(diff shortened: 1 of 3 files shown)


beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -261,7 +261,13 @@ proc processSignedBeaconBlock*(
 
   # Start of block processing - in reality, we have already gone through SSZ
   # decoding at this stage, which may be significant
-  debug "Block received", delay
+  let bid = BlockId(root: signedBlock.root, slot: signedBlock.message.slot)
```
tersec: This is only used by `bid = shortLog(bid)`, but it will run even in non-`DEBUG` settings (default). Better to put it in the `debug` statement inline so it gets conditionally run:
```nim
  debug "Block received",
    bid = BlockId(root: signedBlock.root, slot: signedBlock.message.slot)
```

cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8575/commits/3fee1d85df8e1bdab26c5c10700f389f2cef90ac

beacon_chain/gossip_processing/eth2_processor.nim
```diff
@@ -261,7 +261,13 @@ proc processSignedBeaconBlock*(
 
   # Start of block processing - in reality, we have already gone through SSZ
   # decoding at this stage, which may be significant
-  debug "Block received", delay
+  debug "Block received",
+    bid = shortLog(
+      BlockId(root: signedBlock.root, slot: signedBlock.message.slot)),
```
etan-status:
```suggestion
    bid = shortLog(signedBlock.toBlockId()),
```

beacon_chain/gossip_processing/eth2_processor.nim
tersec: https://github.com/status-im/nimbus-eth2/pull/8575/commits/3fee1d85df8e1bdab26c5c10700f389f2cef90ac

beacon_chain/gossip_processing/eth2_processor.nim
cheatfate: Fixed in https://github.com/status-im/nimbus-eth2/pull/8575/commits/7fdc2fa70e5789e92091debd54431f0d22cdef1d

PR #8582, opened 2026-06-10
Honor --debug-invalidate-block-root on ChainDAG init

etan-status:
The --debug-invalidate-block-root flag allows users to prevent import of blocks that become asynchronously invalid via engine_forkchoiceUpdated. While we honor the flag during gossip validation to prevent importing blocks into fork choice optimistically (potentially getting stuck on an invalid justification), we currently don't apply it on initialization.

Update ChainDAGRef.init so that if we encounter a user invalidated block when loading the non-finalized portion of the chain, we instead reset the head to its parent block (latest valid head).


tersec:
Merge conflict in `beacon_chain/consensus_object_pools/blockchain_dag.nim`

PR #8591, opened 2026-06-11
Ensure bids are for a higher slot than their parent

etan-status:
Adopt gossip rule from v1.7.0-alpha.9:

- https://github.com/ethereum/consensus-specs/pull/5302


beacon_chain/gossip_processing/gossip_validation.nim
tersec, line 1843: Spec reference URL should be updated to reflect that it's been made consistent with a newer version

etan-status: It wasn't and still isn't consistent, though. https://github.com/ethereum/consensus-specs/pull/5190 isn't adopted, even though part of specs since a.6

PR #8606, opened 2026-06-16
Ensure verification conditions for bid are satisfied as proposer

etan-status:
Spec requires the proposer to ensure that the verification conditions in `process_execution_payload_bid` are satisfied in order to construct the `signed_execution_payload_bid` field in `BeaconBlockBody`:

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/validator.md#signed-execution-payload-bid

As gossip validation uses some arbitrary head state and also misses certain checks compared to the full verification conditions, bids that passed gossip validation may become invalid when re-verified with the correct state.

Re-verify external bid validity on proposal, and if invalid, fall back to local block building.

### DIFF
--- a/beacon_chain/consensus_object_pools/execution_payload_pool.nim
+++ b/beacon_chain/consensus_object_pools/execution_payload_pool.nim
@@ -112,6 +112,23 @@ func getHighestBidForSlotAndParent*(
   else:
     Opt.none(gloas.SignedExecutionPayloadBid)
 
+proc getHighestBidForProposalState*(
+    pool: ExecutionPayloadBidPool, state: ForkyBeaconState,
+    payloadAvailability: PayloadAvailability
+): Opt[gloas.SignedExecutionPayloadBid] =
+  if state.slot <= GENESIS_SLOT:
+    return static(Opt.none gloas.SignedExecutionPayloadBid)
+
+  result = pool.getHighestBidForSlotAndParent(
+    state.slot, state.get_block_root_at_slot(state.slot - 1),
+    payloadAvailability)
+  if result.isNone:
+    return static(Opt.none gloas.SignedExecutionPayloadBid)
+
+  pool.dag.cfg.can_process_execution_payload_bid(
+      state, result.unsafeGet, state.slot, {skipBlsValidation}).isOkOr:
+    return static(Opt.none gloas.SignedExecutionPayloadBid)
+
 func hasSeenBidFromBuilder*(
     pool: ExecutionPayloadBidPool, slot: Slot,
     builderIndex: uint64): bool =
--- a/beacon_chain/consensus_object_pools/execution_payload_pool.nim
+++ b/beacon_chain/consensus_object_pools/execution_payload_pool.nim
@@ -119,15 +119,14 @@ proc getHighestBidForProposalState*(
   if state.slot <= GENESIS_SLOT:
     return static(Opt.none gloas.SignedExecutionPayloadBid)
 
-  result = pool.getHighestBidForSlotAndParent(
+  let res = pool.getHighestBidForSlotAndParent(
     state.slot, state.get_block_root_at_slot(state.slot - 1),
     payloadAvailability)
-  if result.isNone:
-    return static(Opt.none gloas.SignedExecutionPayloadBid)
-
-  pool.dag.cfg.can_process_execution_payload_bid(
-      state, result.unsafeGet, state.slot, {skipBlsValidation}).isOkOr:
-    return static(Opt.none gloas.SignedExecutionPayloadBid)
+  res.isErrOr:
+    if pool.dag.cfg.can_process_execution_payload_bid(
+        state, value, state.slot, {skipBlsValidation}).isErr:
+      return static(Opt.none gloas.SignedExecutionPayloadBid)
+  res
 
 func hasSeenBidFromBuilder*(
     pool: ExecutionPayloadBidPool, slot: Slot,
--- a/beacon_chain/spec/state_transition_block.nim
+++ b/beacon_chain/spec/state_transition_block.nim
@@ -1276,15 +1276,18 @@ proc apply_parent_execution_payload*(
 
   ok()
 
-# copy of datatypes/gloas.nim
-type SomeGloasBeaconBlock =
-  gloas.BeaconBlock | gloas.SigVerifiedBeaconBlock | gloas.TrustedBeaconBlock
-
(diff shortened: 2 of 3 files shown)


beacon_chain/spec/state_transition_block.nim
```diff
@@ -1316,16 +1320,51 @@ proc process_execution_payload_bid*(
...
+    signed_bid:
+      gloas.SignedExecutionPayloadBid | heze.SignedExecutionPayloadBid,
+    proposal_slot: Slot, flags = default(UpdateFlags)): Result[void, cstring] =
+  debugHezeComment "accept only heze.SignedExecutionPayloadBid"
+  cfg.can_process_execution_payload_bid_impl(
+    state, signed_bid, proposal_slot, flags)
+
+# copy of datatypes/gloas.nim
+type SomeGloasBeaconBlock =
+  gloas.BeaconBlock | gloas.SigVerifiedBeaconBlock | gloas.TrustedBeaconBlock
```
tersec, line 1354: As of 2.2.10, this redefinition shouldn't be required anymore; should be able to use the datatypes/gloas typeclass directly. The underlying Nim issue has been fixed

beacon_chain/consensus_object_pools/execution_payload_pool.nim
```diff
@@ -112,6 +112,23 @@ func getHighestBidForSlotAndParent*(
...
     Opt.none(gloas.SignedExecutionPayloadBid)
 
+proc getHighestBidForProposalState*(
+    pool: ExecutionPayloadBidPool, state: ForkyBeaconState,
+    payloadAvailability: PayloadAvailability
+): Opt[gloas.SignedExecutionPayloadBid] =
+  if state.slot <= GENESIS_SLOT:
+    return static(Opt.none gloas.SignedExecutionPayloadBid)
+
+  result = pool.getHighestBidForSlotAndParent(
```
tersec: https://status-im.github.io/nim-style-guide/language.result.html

beacon_chain/spec/state_transition_block.nim
etan-status: Want me to remove these as part of this PR? I only moved it, and there are additional copies of these in this file. If you like dropping, should I drop all of them as part of this?

beacon_chain/consensus_object_pools/execution_payload_pool.nim
etan-status: worth the risk of additional bid copies, to do a "var res = ..." here? xD

etan-status: could also do a couple unwrappings / re-wrappings with .valueOr / Opt.some etc, just not sure about the tradeoff of performance / readability.

beacon_chain/consensus_object_pools/execution_payload_pool.nim
```diff
@@ -112,6 +112,23 @@ func getHighestBidForSlotAndParent*(
...
+    state.slot, state.get_block_root_at_slot(state.slot - 1),
+    payloadAvailability)
+  if result.isNone:
+    return static(Opt.none gloas.SignedExecutionPayloadBid)
+
+  pool.dag.cfg.can_process_execution_payload_bid(
+      state, result.unsafeGet, state.slot, {skipBlsValidation}).isOkOr:
+    return static(Opt.none gloas.SignedExecutionPayloadBid)
```
etan-status:
```suggestion
  let res = pool.getHighestBidForSlotAndParent(
    state.slot, state.get_block_root_at_slot(state.slot - 1),
    payloadAvailability)
  if res.isNone:
    return static(Opt.none gloas.SignedExecutionPayloadBid)

  pool.dag.cfg.can_process_execution_payload_bid(
      state, res.unsafeGet, state.slot, {skipBlsValidation}).isOkOr:
    return static(Opt.none gloas.SignedExecutionPayloadBid)
  
  res
```

etan-status:
```suggestion
  result = pool.getHighestBidForSlotAndParent(
    state.slot, state.get_block_root_at_slot(state.slot - 1),
    payloadAvailability)
  result.isErrOr:
    pool.dag.cfg.can_process_execution_payload_bid(
        state, value, state.slot, {skipBlsValidation}).isOkOr:
      result.reset()
```

beacon_chain/spec/state_transition_block.nim
tersec: Sure

beacon_chain/consensus_object_pools/execution_payload_pool.nim
etan-status:
```suggestion
  if state.slot > GENESIS_SLOT:
    result = pool.getHighestBidForSlotAndParent(
      state.slot, state.get_block_root_at_slot(state.slot - 1),
      payloadAvailability)
    result.isErrOr:
      if pool.dag.cfg.can_process_execution_payload_bid(
          state, value, state.slot, {skipBlsValidation}).isErr:
        result.reset()
```

or like this, that at least does not mix "return" with "result", while avoiding the potential extra copy. hmm

tersec: Yeah this seems reasonable enough.

etan-status:
```suggestion
  if state.slot <= GENESIS_SLOT:
    return static(Opt.none gloas.SignedExecutionPayloadBid)

  let res = pool.getHighestBidForSlotAndParent(
    state.slot, state.get_block_root_at_slot(state.slot - 1),
    payloadAvailability)
  res.isErrOr:
    if pool.dag.cfg.can_process_execution_payload_bid(
        state, value, state.slot, {skipBlsValidation}).isErr:
      return static(Opt.none gloas.SignedExecutionPayloadBid)
  res
```

beacon_chain/spec/state_transition_block.nim
etan-status: some of the others can't be dropped as easily, e.g., 

> ./beacon_chain/spec/state_transition_block.nim(1089, 11) Error: undeclared identifier: 'SomeDenebBeaconBlockBody'

if one drops it. i.e., the "copy of datatypes/foo.nim" is fake news.

keeping this one single-concern, it becomes too messy. and dropping gloas but not others is weird.

tersec: Ok, also fine.

tersec: I did check that one, and it would be `deneb.SomeBeaconBlockBody`.

But yeah these can be done separately, it's fine.

PR #8609, opened 2026-06-16
Reject bids with invalid prev_randao during gossip validation

etan-status:
When someone sends us a bid with invalid prev_randao but high value, it kicks out the legitimate bid. The invalid one is also non-punishable as it fails subsequent state transition by peers.

Add an extra reject rule to filter out invalid prev_randao, there is only a single valid value for each parent block root, and we have efficient functionality in the blockchain_dag to obtain correct RANDAO without expensive state replays.

Also proposed for spec inclusion, but we'll need the rule regardless:

- https://github.com/ethereum/consensus-specs/pull/5360

Have moved this check to the end, because if the correct value has not yet been cached (i.e., no valid bid received), we may have to load the block from disk (once).

### DIFF
--- a/beacon_chain/consensus_object_pools/execution_payload_pool.nim
+++ b/beacon_chain/consensus_object_pools/execution_payload_pool.nim
@@ -130,9 +130,8 @@ proc getPrevRandao*(
     pool: ExecutionPayloadBidPool, slot: Slot,
     parentBid: BlockId): Opt[Eth2Digest] =
   for payloadAvailability in PayloadAvailability:
-    let bid = pool.getHighestBidForSlotAndParent(
-      slot, parentBid.root, payloadAvailability)
-    if bid.isSome:
-      return Opt.some bid.unsafeGet.message.prev_randao
+    pool.getHighestBidForSlotAndParent(
+        slot, parentBid.root, payloadAvailability).isErrOr:
+      return Opt.some value.message.prev_randao
 
   pool.dag.computeRandaoMix(parentBid)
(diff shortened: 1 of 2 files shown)


beacon_chain/consensus_object_pools/execution_payload_pool.nim
```diff
@@ -125,3 +125,14 @@ proc prune*(pool: var ExecutionPayloadBidPool, beforeSlot: Slot) =
...
     pool.slotBids.del(slot)
+
+proc getPrevRandao*(
+    pool: ExecutionPayloadBidPool, slot: Slot,
+    parentBid: BlockId): Opt[Eth2Digest] =
+  for payloadAvailability in PayloadAvailability:
+    let bid = pool.getHighestBidForSlotAndParent(
+      slot, parentBid.root, payloadAvailability)
+    if bid.isSome:
+      return Opt.some bid.unsafeGet.message.prev_randao
```
tersec: Can be done as efficiently and more quasi-staticly-safely with
```nim
    bid.isErrOr:
      return Opt.some value.message.prev_randao
```

tersec:
Merge conflict in beacon_chain/consensus_object_pools/execution_payload_pool.nim

beacon_chain/consensus_object_pools/execution_payload_pool.nim
etan-status:
```suggestion
    pool.getHighestBidForSlotAndParent(
        slot, parentBid.root, payloadAvailability).isErrOr:
      return Opt.some value.message.prev_randao
```

PR #8618, opened 2026-06-17
Derive should_extend_payload from fork choice instead of stored headPayload

Tomi-3-0:
Replace #8481's `dag.headPayload` proposer mechanism with the fork-choice `should_extend_payload`. 
Depends on #8572


ahshum:
`headPayload` is already a result by fork choice and saved into DAG, similar to `head`. Why do we need depending on fork choice?

ahshum:
And this is incorrect according to spec
https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/validator.md#executionpayload

> ... as decided by should_build_on_full(store, head) ...

That being said, the function name is borrowed from that in the fork choice. The idea stays but they are different.

PR #8620, opened 2026-06-17
Use EL and CL identifiers in default graffiti

### LINKED ISSUE #7580: Default to EL<version>CL<version> in graffiti
According to: https://github.com/ethereum/execution-apis/pull/517#issuecomment-1931404053

The EL provides a version number to the CL. Ideally Nimbus should also report this too. 

other cls: 

Image

akronim26:
Uses the EL and CL version to default graffiti to match the implementation with other CL clients

Resolves #7580

### DIFF
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -117,10 +117,6 @@ type
     clientVersion: Opt[ClientVersionV1]
       ## The version reported through `engine_getClientVersionV1`.
 
-    clientVersionExchanged: bool
-      ## Whether `engine_getClientVersionV1` has been attempted for this
-      ## connection.
-
     state: ELConnectionState
     hysteresisCounter: int
 
@@ -300,16 +296,20 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
   connection.web3.get.provider
 
 func consensusClientVersion(): ClientVersionV1 =
+  const commit = Bytes4(hexToByteArray[4](gitRevisionBytes4))
+
   ClientVersionV1(
-    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: default(Bytes4)
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: commit
   )
 
 func makeClientVersionGraffiti*(elVersion: ClientVersionV1): GraffitiBytes =
   let
     elCommit = toHex(distinctBase(elVersion.commit))
-    graffiti = elVersion.code & elCommit[0 ..< 4] & "NB" & gitRevision[0 ..< 4]
+    graffiti = elVersion.code & elCommit[0 ..< 4] & "NB" & gitRevisionBytes4[0 ..< 4]
 
-  distinctBase(result)[0 ..< graffiti.len] = toBytes(graffiti)
+  var res: GraffitiBytes
+  distinctBase(res)[0 ..< graffiti.len] = toBytes(graffiti)
+  res
 
 func getClientVersion*(m: ELManager): Opt[ClientVersionV1] =
   for connection in m.elConnections:
@@ -1106,8 +1106,7 @@ proc checkChainId(
 ) {.async: (raises: [CancelledError]).} =
   let rpcClient = await connection.connectedRpcClient()
 
-  if not connection.clientVersionExchanged:
-    connection.clientVersionExchanged = true
+  if connection.clientVersion.isNone:
     try:
       let versions = await connection.engineApiRequest(
         rpcClient.getClientVersion(consensusClientVersion()),
@@ -1124,7 +1123,6 @@ proc checkChainId(
           version = versions[0].version
     except CancelledError as exc:
       debug "Client version exchange was interrupted"
-      connection.clientVersionExchanged = false
       raise exc
     except CatchableError as exc:
       debug "Failed to obtain EL client version", reason = exc.msg
--- a/beacon_chain/el/el_manager.nim
+++ b/beacon_chain/el/el_manager.nim
@@ -16,10 +16,11 @@ import
   eth/common/eth_types,
   results,
   kzg4844/[kzg_abi, kzg],
-  stew/objects,
+  stew/[byteutils, objects],
   # Local modules:
   ../spec/[engine_authentication, forks, helpers_el],
   ../networking/network_metadata,
+  ../version,
   "."/[el_conf, engine_api_conversions]
 
 from std/sequtils import anyIt, filterIt, mapIt
@@ -112,6 +113,13 @@ type
 
     chainIdStatus: ChainIdStatus
       ## The latest status of the `checkChainId` exchange.
+
+    clientVersion: Opt[ClientVersionV1]
+      ## The version reported through `engine_getClientVersionV1`.
+
+    clientVersionExchanged: bool
+      ## Whether `engine_getClientVersionV1` has been attempted for this
+      ## connection.
 
     state: ELConnectionState
     hysteresisCounter: int
@@ -290,6 +298,25 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
       await sleepAsync(chronos.seconds(10))
 
   connection.web3.get.provider
+
+func consensusClientVersion(): ClientVersionV1 =
+  ClientVersionV1(
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: default(Bytes4)
+  )
+
+func makeClientVersionGraffiti*(elVersion: ClientVersionV1): GraffitiBytes =
+  let
+    elCommit = toHex(distinctBase(elVersion.commit))
+    graffiti = elVersion.code & elCommit[0 ..< 4] & "NB" & gitRevision[0 ..< 4]
+
+  distinctBase(result)[0 ..< graffiti.len] = toBytes(graffiti)
+
+func getClientVersion*(m: ELManager): Opt[ClientVersionV1] =
+  for connection in m.elConnections:
+    if connection.clientVersion.isSome:
+      return connection.clientVersion
+
+  Opt.none(ClientVersionV1)
 
 template retryUntilCancelled(body: untyped) =
   ## Perform the same request in a loop until it is explicitly cancelled,
@@ -1078,6 +1105,29 @@ proc checkChainId(
     connection: ELConnection
 ) {.async: (raises: [CancelledError]).} =
   let rpcClient = await connection.connectedRpcClient()
+
+  if not connection.clientVersionExchanged:
+    connection.clientVersionExchanged = true
+    try:
+      let versions = await connection.engineApiRequest(
+        rpcClient.getClientVersion(consensusClientVersion()),
+        "getClientVersion",
+        Moment.now(),
+      )
+
+      if versions.len > 0:
+        connection.clientVersion = Opt.some(versions[0])
+        debug "Exchanged EL client version",
+          url = connection.engineUrl,
+          code = versions[0].code,
+          name = versions[0].name,
+          version = versions[0].version
+    except CancelledError as exc:
+      debug "Client version exchange was interrupted"
+      connection.clientVersionExchanged = false
+      raise exc
+    except CatchableError as exc:
+      debug "Failed to obtain EL client version", reason = exc.msg
 
   if m.eth1Network.isSome and
      connection.chainIdStatus == ChainIdStatus.notExchangedYet:
--- a/beacon_chain/validators/beacon_validators.nim
+++ b/beacon_chain/validators/beacon_validators.nim
@@ -210,10 +210,18 @@ proc getValidatorForDuties*(
   node.attachedValidators[].getValidatorForDuties(
     key.toPubKey(), slot, slashingSafe)
 
-proc getGraffitiBytes*(
-    node: BeaconNode, validator: AttachedValidator): GraffitiBytes =
-  getGraffiti(node.config.validatorsDir, node.config.defaultGraffitiBytes(),
-              validator.pubkey)
+proc getGraffitiBytes*(node: BeaconNode, validator: AttachedValidator): GraffitiBytes =
+  let defaultGraffiti =
+    if node.config.graffiti.isSome:
+      node.config.defaultGraffitiBytes()
+    else:
+      let elClientVersion = node.elManager.getClientVersion()
+      if elClientVersion.isSome:
+        makeClientVersionGraffiti(elClientVersion.get)
+      else:
+        node.config.defaultGraffitiBytes()
+
+  getGraffiti(node.config.validatorsDir, defaultGraffiti, validator.pubkey)
 
 proc isSynced*(node: BeaconNode, head: BlockRef): bool =
   ## TODO This function is here as a placeholder for some better heurestics to
--- a/tests/test_el_manager.nim
+++ b/tests/test_el_manager.nim
@@ -18,6 +18,7 @@ import
   eth/common/eth_types,
   ../beacon_chain/el/[el_conf, el_manager],
   ../beacon_chain/spec/[digest, engine_authentication, forks],
+  ../beacon_chain/version,
   ../beacon_chain/networking/network_metadata,
   ./testutil
 
@@ -37,20 +38,23 @@ type
     ## Tracks the state of a mock execution client for testing scenarios, mainly
     ## for the purpose of testing failure, timeout and error scenarios (rather
     ## than specific payload properties)
-    chainId: UInt256
-    shouldFailNewPayload: bool
-    shouldFailForkchoice: bool
-    shouldFailGetPayload: bool
-    shouldFailChainId: bool
-    newPayloadCallCount: int
-    newPayloadV5CallCount: int
-    forkchoiceCallCount: int
-    forkchoiceV4CallCount: int
-    getPayloadCallCount: int
-    getPayloadV6CallCount: int
-    chainIdCallCount: int
-    responseDelay: Duration
-    blockNumber: uint64
+    chainId*: UInt256
+    shouldFailNewPayload*: bool
+    shouldFailForkchoice*: bool
+    shouldFailGetPayload*: bool
+    shouldFailChainId*: bool
+    shouldFailClientVersion*: bool
+    newPayloadCallCount*: int
+    newPayloadV5CallCount*: int
+    forkchoiceCallCount*: int
+    forkchoiceV4CallCount*: int
+    getPayloadCallCount*: int
+    getPayloadV6CallCount*: int
+    chainIdCallCount*: int
+    getClientVersionCallCount*: int
+    clientVersion*: ClientVersionV1
+    responseDelay*: Duration
+    blockNumber*: uint64
 
   MockSetup = ref object
     state: MockEngineState
@@ -67,6 +71,7 @@ func createMockEngineState(
     shouldFailForkchoice: false,
     shouldFailGetPayload: false,
     shouldFailChainId: false,
+    shouldFailClientVersion: false,
     newPayloadCallCount: 0,
     newPayloadV5CallCount: 0,
     getPayloadCallCount: 0,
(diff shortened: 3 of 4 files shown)


beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +299,25 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
...
+  ClientVersionV1(
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: default(Bytes4)
+  )
+
+func makeClientVersionGraffiti*(elVersion: ClientVersionV1): GraffitiBytes =
+  let
+    elCommit = toHex(distinctBase(elVersion.commit))
+    graffiti = elVersion.code & elCommit[0 ..< 4] & "NB" & gitRevision[0 ..< 4]
+
+  distinctBase(result)[0 ..< graffiti.len] = toBytes(graffiti)
```
tersec: https://status-im.github.io/nim-style-guide/language.result.html
> Use expression-based return

tests/test_el_manager.nim
```diff
@@ -37,20 +38,23 @@ type
...
+    newPayloadV5CallCount*: int
+    forkchoiceCallCount*: int
+    forkchoiceV4CallCount*: int
+    getPayloadCallCount*: int
+    getPayloadV6CallCount*: int
+    chainIdCallCount*: int
+    getClientVersionCallCount*: int
+    clientVersion*: ClientVersionV1
+    responseDelay*: Duration
+    blockNumber*: uint64
```
tersec: Nothing here needs to be exported (`*`), and especially all those which didn't change at all, but even the new ones. See also https://status-im.github.io/nim-style-guide/language.proc.html

There is no public API here.

tests/test_el_manager.nim
```diff
@@ -350,11 +376,45 @@ suite "EL Manager - Helpers":
...
+        code: "GE",
+        name: "go-ethereum",
+        version: "v1.14.0-super-long-version-string",
+        commit: Bytes4([0x16'u8, 0x8d, 0xbe, 0xef]),
+      )
+    )
+
+    check $graffiti == "GE168dNB" & gitRevision[0 ..< 4]
+
+proc createELManager*(
```
tersec: For no particular reason exports something which doesn't have to be exported, https://status-im.github.io/nim-style-guide/language.proc.html

beacon_chain/el/el_manager.nim
```diff
@@ -1079,6 +1106,29 @@ proc checkChainId(
...
+      if versions.len > 0:
+        connection.clientVersion = Opt.some(versions[0])
+        debug "Exchanged EL client version",
+          url = connection.engineUrl,
+          code = versions[0].code,
+          name = versions[0].name,
+          version = versions[0].version
+    except CancelledError as exc:
+      debug "Client version exchange was interrupted"
+      connection.clientVersionExchanged = false
```
tersec: It does this here but not as part of the `CatchableError`? So in some circumstances, it can fail to get a client version and the result will be `connection.clientVersionExchanged == true`, and in some other circumstances, it can fail to get the client version and the result will be `connection.clientVersionExchanged == false`.

beacon_chain/el/el_manager.nim
```diff
@@ -113,6 +114,13 @@ type
     chainIdStatus: ChainIdStatus
       ## The latest status of the `checkChainId` exchange.
 
+    clientVersion: Opt[ClientVersionV1]
+      ## The version reported through `engine_getClientVersionV1`.
+
+    clientVersionExchanged: bool
```
tersec: This doesn't seem used anywhere? It (correctly) checks for `clientVersion.isSome`, but this is then redundant.

tests/test_el_manager.nim
```diff
@@ -383,6 +443,32 @@ suite "EL Manager - Async Operations":
...
+    let engineUrl = createEngineApiUrl(mockPort)
+    let manager = createELManager(@[engineUrl])
+    manager.start()
+
+    check waitFor waitForGetClientVersionRequest(mockState)
+
+    let version = manager.getClientVersion()
+    check:
+      mockState.getClientVersionCallCount == 1
+      version.isSome
```
tersec: This is a redundant check; `version.get` will `Defect` if this isn't true.

beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +299,25 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
 
   connection.web3.get.provider
 
+func consensusClientVersion(): ClientVersionV1 =
+  ClientVersionV1(
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: default(Bytes4)
```
tersec: `default(Bytes4)`? `makeClientVersionGraffiti` already accesses it, for example, via `gitRevision`.

akronim26: Currently `gitRevision` spans 6 hex characters, but according to the [spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#clientversionv1), `commit` should be the latest commit of the build spanning 8 hex characters. Hence, we need to create `gitRevisionBytes4` to span 8 characters while keeping the existing `gitVersion` to be compatible with the default graffiti

beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +295,29 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
...
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: commit
+  )
+
+func makeClientVersionGraffiti*(elVersion: ClientVersionV1): GraffitiBytes =
+  let
+    elCommit = toHex(distinctBase(elVersion.commit))
+    graffiti = elVersion.code & elCommit[0 ..< 4] & "NB" & gitRevisionBytes4[0 ..< 4]
+
+  var res: GraffitiBytes
+  distinctBase(res)[0 ..< graffiti.len] = toBytes(graffiti)
```
etan-status, line 311: Graffiti has limited length, so if EL sends some value that's > 22 bytes this could trigger `RangeDefect`

beacon_chain/validators/beacon_validators.nim
```diff
@@ -210,10 +210,18 @@ proc getValidatorForDuties*(
   node.attachedValidators[].getValidatorForDuties(
     key.toPubKey(), slot, slashingSafe)
 
-proc getGraffitiBytes*(
-    node: BeaconNode, validator: AttachedValidator): GraffitiBytes =
-  getGraffiti(node.config.validatorsDir, node.config.defaultGraffitiBytes(),
-              validator.pubkey)
+proc getGraffitiBytes*(node: BeaconNode, validator: AttachedValidator): GraffitiBytes =
```
etan-status, line 213: Nimbus uses a different flow for validators connected via Validator Client.
The one here only applies to validators hosted within the beacon node itself.

For the Validator Client based flow, the VC typically provides its own graffiti preference (in the case on Nimbus VC, it _always_ sends its own preference, but the API is optional and other VCs may not indicate a preference). If a VC preference is given, not sure how to best handle that. It would be weird if the VC supplied graffiti would be randomly changed by the BN just to extract some questionable telemetry; would be a question for the spec authors how to handle that case.

beacon_chain/el/el_manager.nim
```diff
@@ -1079,6 +1106,27 @@ proc checkChainId(
 ) {.async: (raises: [CancelledError]).} =
   let rpcClient = await connection.connectedRpcClient()
 
+  if connection.clientVersion.isNone:
+    try:
+      let versions = await connection.engineApiRequest(
```
etan-status, line 1111: If engineApiRequest fails for whatever reason (is this endpoint always available, etc), this risks closing down the entire connection to the EL and marking it "degraded"

https://github.com/status-im/nimbus-eth2/blob/404a0001561d1d83c5b5bf35dcbedbcb5fb86572/beacon_chain/el/el_manager.nim#L223-L230

beacon_chain/el/el_manager.nim
```diff
@@ -1079,6 +1106,27 @@ proc checkChainId(
 ) {.async: (raises: [CancelledError]).} =
   let rpcClient = await connection.connectedRpcClient()
 
+  if connection.clientVersion.isNone:
```
etan-status, line 1109: Even if it's known, it might change over time, if the EL is upgraded without restarting the BN.

In the Websocket mode, one has a sense of connection, but in the HTTP mode, such an upgrade can happen between every request.

Actually, even in Websocket mode if one has a proxy in-between the BN and a multi-EL setup, because then we only connect through the proxy.

Probably can't get around periodically fetching the latest version information with the way how the spec is proposed.

beacon_chain/el/el_manager.nim
```diff
@@ -1079,6 +1106,27 @@ proc checkChainId(
...
+    try:
+      let versions = await connection.engineApiRequest(
+        rpcClient.getClientVersion(consensusClientVersion()),
+        "getClientVersion",
+        Moment.now(),
+      )
+
+      if versions.len > 0:
+        connection.clientVersion = Opt.some(versions[0])
```
etan-status, line 1118: Not sure if picking [0] is the best choice. Maybe better to pick based on the slot number, so that each EL gets a proportional choice on getting included (picked across all. Nimbus supports multiple --el arguments, and each --el might be a proxy that contains even more ELs)

beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +295,29 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
 
   connection.web3.get.provider
 
+func consensusClientVersion(): ClientVersionV1 =
+  const commit = Bytes4(hexToByteArray[4](gitRevisionBytes4))
```
etan-status, line 299: What if gitRevisionBytes4 is not a hex string, e.g., because someone compiles without the .git folder? should have a fallback for that probably.

beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +295,29 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
...
+
+  ClientVersionV1(
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: commit
+  )
+
+func makeClientVersionGraffiti*(elVersion: ClientVersionV1): GraffitiBytes =
```
etan-status, line 305: What about the cases that the spec lists if the user-specified graffiti is a bit longer? they have different versions for up to 30 bytes of user graffiti.

Also would be great to have tests for the various use cases. Certain alternative implementations don't implement the spec correctly (spec says to prepend, but I saw it being appended by others)

tests/test_el_manager.nim
```diff
@@ -350,11 +376,45 @@ suite "EL Manager - Helpers":
 
       gethWsUrl == "ws://localhost:8545"
 
-func createELManager(
+  test "client version graffiti includes EL and CL versions":
+    let graffiti = makeClientVersionGraffiti(
+      ClientVersionV1(
+        code: "GE",
+        name: "go-ethereum",
```
etan-status, line 383: Is this an official test vector from the specs? Otherwise, probably should stick to nimbus instead of advertising geth here

tests/test_el_manager.nim
```diff
@@ -350,11 +376,45 @@ suite "EL Manager - Helpers":
...
+    )
+
+    check $graffiti == "GE168dNB" & gitRevisionBytes4[0 ..< 4]
+
+  test "client version graffiti ignores long semver versions":
+    let graffiti = makeClientVersionGraffiti(
+      ClientVersionV1(
+        code: "GE",
+        name: "go-ethereum",
+        version: "v1.14.0-super-long-version-string",
```
etan-status, line 396: does the version even get included anywhere? I guess only the commit is relevant, and the code (so the tests should probably try invalid codes instead of invalid versions)

tests/test_el_manager.nim
```diff
@@ -350,11 +376,45 @@ suite "EL Manager - Helpers":
...
 ): ELManager =
   ELManager.new(engines, eth1Network)
 
+proc waitForGetClientVersionRequest(
+    state: MockEngineState
+): Future[bool] {.async: (raises: [CancelledError]).} =
+  for _ in 0 ..< 50:
+    if state.getClientVersionCallCount > 0:
+      return true
+    await sleepAsync(50.milliseconds)
```
etan-status, line 414: These kinds of timing based tests tend to get flaky when the CI server is randomly overloaded. Could we just use sth like an AsyncEvent instead?

tests/test_el_manager.nim
```diff
@@ -350,11 +376,45 @@ suite "EL Manager - Helpers":
...
+    check $graffiti == "GE168dNB" & gitRevisionBytes4[0 ..< 4]
+
+proc createELManager(
     engines: seq[EngineApiUrl], eth1Network: Opt[Eth1Network] = Opt.none(Eth1Network)
 ): ELManager =
   ELManager.new(engines, eth1Network)
 
+proc waitForGetClientVersionRequest(
+    state: MockEngineState
+): Future[bool] {.async: (raises: [CancelledError]).} =
```
etan-status, line 410: Opt[void] is better than bool, because with bool you never know if true means success or error (it's inconsistent across APIs)

tests/test_el_manager.nim
```diff
@@ -383,6 +443,31 @@ suite "EL Manager - Async Operations":
...
+
+    check waitFor waitForGetClientVersionRequest(mockState)
+
+    let version = manager.getClientVersion()
+    check:
+      mockState.getClientVersionCallCount == 1
+      version.get.code == "GE"
+      version.get.version == "v1.14.0"
+
+  test "ELManager ignores failed client version exchange":
```
etan-status, line 459: but it doesn't ignore it, it pushes the EL into "degraded" mode if it fails too often

beacon_chain/el/el_manager.nim
```diff
@@ -1079,6 +1106,27 @@ proc checkChainId(
...
+          url = connection.engineUrl,
+          code = versions[0].code,
+          name = versions[0].name,
+          version = versions[0].version
+    except CancelledError as exc:
+      debug "Client version exchange was interrupted"
+      raise exc
+    except CatchableError as exc:
+      debug "Failed to obtain EL client version", reason = exc.msg
+
```
etan-status, line 1129: Might be better to try the chain ID check first, as that one is important to raise to the user if it's wrong (keep in midn that this proc runs under a somewhat tight timeout)

beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +295,29 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
 
   connection.web3.get.provider
 
+func consensusClientVersion(): ClientVersionV1 =
+  const commit = Bytes4(hexToByteArray[4](gitRevisionBytes4))
+
+  ClientVersionV1(
+    code: "NB", name: "Nimbus", version: "v" & versionAsStr, commit: commit
```
etan-status, line 302: the NB / Nimbus things should probably be constants in version.nim where it could be adjusted centrally, rather than inlined here. otherwise, if we ever get a new code assigned or want to change product names it becomes annoying

beacon_chain/validators/beacon_validators.nim
```diff
@@ -210,10 +210,18 @@ proc getValidatorForDuties*(
...
+  let defaultGraffiti =
+    if node.config.graffiti.isSome:
+      node.config.defaultGraffitiBytes()
+    else:
+      let elClientVersion = node.elManager.getClientVersion()
+      if elClientVersion.isSome:
+        makeClientVersionGraffiti(elClientVersion.get)
```
etan-status, line 220: well, the spec doesn't say to replace the entire graffiti

beacon_chain/validators/beacon_validators.nim
```diff
@@ -210,10 +210,18 @@ proc getValidatorForDuties*(
...
     key.toPubKey(), slot, slashingSafe)
 
-proc getGraffitiBytes*(
-    node: BeaconNode, validator: AttachedValidator): GraffitiBytes =
-  getGraffiti(node.config.validatorsDir, node.config.defaultGraffitiBytes(),
-              validator.pubkey)
+proc getGraffitiBytes*(node: BeaconNode, validator: AttachedValidator): GraffitiBytes =
+  let defaultGraffiti =
+    if node.config.graffiti.isSome:
+      node.config.defaultGraffitiBytes()
```
etan-status, line 216: So, in the default case where the user specifies a graffiti, the feature doesn't even work?

https://hackmd.io/@wmoBhF17RAOH2NZ5bNXJVg/BJX2c9gja#DEFAULT-CASE-user-does-not-set-graffiti-or-elects-to-customize-lt-20-graffiti-characters

Once again, not sure if this should be default-on, at the very least the documentation should be updated to suggest user who wants to opt-out to specify a full 32-byte graffiti that contains trailing 0 bytes. Note graffitis are not strings, they are opaque byte arrays

beacon_chain/el/el_manager.nim
```diff
@@ -291,6 +295,29 @@ proc connectedRpcClient(connection: ELConnection): Future[RpcClient] {.
...
+    graffiti = elVersion.code & elCommit[0 ..< 4] & "NB" & gitRevisionBytes4[0 ..< 4]
+
+  var res: GraffitiBytes
+  distinctBase(res)[0 ..< graffiti.len] = toBytes(graffiti)
+  res
+
+func getClientVersion*(m: ELManager): Opt[ClientVersionV1] =
+  for connection in m.elConnections:
+    if connection.clientVersion.isSome:
+      return connection.clientVersion
```
etan-status, line 317: Might be better to pick a random one (e.g., slot number mod available versions)

etan-status (commented):
Thanks for giving this a shot.

I'm still not sure how acceptable the proposed feature is for users. As is, the [specs](https://hackmd.io/@wmoBhF17RAOH2NZ5bNXJVg/BJX2c9gja#DEFAULT-CASE-user-does-not-set-graffiti-or-elects-to-customize-lt-20-graffiti-characters) also require modifying user-provided graffiti values (and many users have custom --graffiti). The PR currently doesn't do that, and only affects the default path. But that would leave Nimbus at this weird stage where the spec is halfway implemented.

Maybe it could be made opt-in with an additional flag. Or opt-out by having the user choose an explicit 32-byte graffiti value (trailing 0 bytes).

If this is kept so that only the default graffiti is affected, keep in mind that the Nimbus VC always sends its own graffiti preference when calling BN api, so even if the BN is started without a --graffiti, the blocks would still end up having the VC provided one. Given that VC deployment is the default, this PR wouldn't end up having an impact there. Maybe the VC could be updated to _not_ send a graffiti preference, if the user doesn't specify one. But that would drop the nimbus default-graffiti when using a nimbus VC with a non-nimbus BN, also not ideal given that it was the VC signing over the block.

Further, this would also need to be extended to work with the signing_node's `proven_block_properties` feature: https://nimbus.guide/web3signer.html#verifying-web3signer

The way how the spec suggests it, the BN would arbitrarily prefix the graffiti with some runtime specific telemetry, but the signing node would have to know what values it is supposed to accept as valid. Otherwise, signing would fail and the block is lost.

akronim26:
Thanks for the detailed feeback on this! 

I read how other clients (Teku, Lighthouse, Lodestar, Prysm) do this. 

- All of them implement it on the BN-side and all of them append after user graffiti.
- All of them has this feature on by default with a opt-out flag.
- Except Lighthouse (with its `graffiti_policy` quiery param), other clients don't have any API involvement. 

For Nimbus, according the me, the approach should be:

- If the graffiti is disabled, then we use the default graffiti, otherwise client tag with appended user graffiti (if applicable).
- If there are multiple EL versions, we only use the CL client tag.
- The client tag should include BN, not the VC, as implemented by other clients too.
- The EL version fecth should move out of `checkChainId` into its own loop with a timestamp and staleness check similar to [Lighthouse](https://github.com/sigp/lighthouse/blob/e423a66763bb1bd780492d635123f208d80c3538/beacon_node/beacon_chain/src/graffiti_calculator.rs#L198).
- We can make the feature on by default with a opt-out flag.
- We can stop sending graffiti preferences from Nimbus VC to BN.

One thing I went against your instinct on is that the tag names the BN rather than the signing VC. All four clients do it that way, and `produceBlockV3` has no field carrying VC identity, so the BN can't know it even if we wanted to.

PR #8625, opened 2026-06-18
explicitly disable nim stack traces with libbacktrace

arnetheduck:
They are enabled by default in non-release builds resulting in two stack trace mechanisms when compiling using regular nim

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -146,10 +146,11 @@ switch("define", "withoutPCRE")
 
 when not defined(disable_libbacktrace):
   --define:nimStackTraceOverride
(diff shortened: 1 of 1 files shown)


tersec:
Have you tested this, to make sure we don't lose stack traces entirely?

arnetheduck:
> Have you tested this, to make sure we don't lose stack traces entirely?

https://github.com/nim-lang/Nim/blob/f8e470eb579c96515693d76ad393dbd41b393f26/config/nim.cfg#L87 - `-d:release` already sets it this way and we set -d:release in the Makefile

PR #8628, opened 2026-06-18
Rename dataColumnQuarantine and add verification optimization to Gloas quarantine too.

cheatfate:


beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -592,11 +613,20 @@ proc enqueueFromDb(self: ref BlockProcessor, root: Eth2Digest) =
     var sidecarsOk = true
 
     let sidecarsOpt =
-      when consensusFork >= ConsensusFork.Gloas:
+      when consensusFork > ConsensusFork.Gloas:
```
agnxsh: maybe a `debugHezeComment` stating there's need to confirm if heze column sidecars are same as gloas column sidecars, and in general, a heze comment regardless, often useful to track and revisit.

tersec: Well, currently, they are the same, and there's no indication yet they'll change. But sure, it's something to verify. But I'm not sure why this change at all; it seems a bit disconnected from the overall purpose of this PR.

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -1020,7 +1056,6 @@ proc storePayload(
     slot = signedBlock.message.slot
 
   # Store sidecars into db.
```
tersec, line 1022: This comment doesn't really make sense anymore?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -479,6 +487,18 @@ proc enqueueQuarantine(self: ref BlockProcessor, parent: BlockRef) =
 
           discard quarantine[].addSidecarless(dag.finalizedHead.slot, forkyBlck)
           continue
+      elif consensusFork == ConsensusFork.Gloas:
```
tersec: It seems like this is textually identical to the `when consensusFork == ConsensusFork.Fulu:` block preceding it; why separate/copied?

beacon_chain/gossip_processing/block_processor.nim
```diff
@@ -700,7 +729,20 @@ proc storeBlock(
...
+        if toVerify.len > 0:
+          ?verifySidecars(signedBlock, Opt.some(toVerify))
+    debug "block_processor verifySidecars completed",
+      verifySidecarsDur = Moment.now() - newPayloadTick,
+      blck = shortLog(signedBlock.message),
+      blockRoot = shortLog(signedBlock.root)
+  elif (consensusFork == ConsensusFork.Gloas) and
+       (sidecarsOpt is not NoSidecars):
+    let pendingVerify =
+      self.gloasColumnQuarantine[].popPendingVerify(signedBlock.root)
```
tersec: The logic in these `Fulu` and `Gloas` blocks is exactly duplicated except for `fuluColumnQuarantine` vs `gloasColumnQuarantine and not NoSidecars`, and it's not 100% trivial. It's pretty straightforward to have `when` block which picks up on that one change specifically rather than copying the entire thing.

ahshum: _Data columns_

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.11/specs/gloas/p2p-interface.md#modified-data_column_sidecar_subnet_id

> [IGNORE] A valid block for the sidecar's slot has been seen ...

_Envelope_

https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.11/specs/gloas/p2p-interface.md#new-execution_payload

> [REJECT] block passes validation.

So in gloas, block should be processed without any sidecars always.

PR #8632, opened 2026-06-19
fix: use gloas proposer dependent root for ProposerPreferences

Tomi-3-0:


beacon_chain/spec/beaconstate.nim
```diff
@@ -3180,6 +3180,15 @@ func dependent_root*(state: ForkyHashedBeaconState, epoch: Epoch): Eth2Digest =
     else:
       Eth2Digest() # "don't know"
 
+# https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.10/specs/gloas/p2p-interface.md#proposer_preferences
+func get_proposer_dependent_root*(
```
nflaig, line 3184: we are about to remove this function from the spec in https://github.com/ethereum/consensus-specs/pull/5374, just fyi

Tomi-3-0: thanks

